PR #456, opened 2012-06-26
Cascade stop

tardyp:
A long running request from users, when having a process heavily using
Trigger steps, is to be able to cascade stopping the builds

There are 2 cases here:
- Either the buildrequest is pending:
  then we simply cancel the buildrequest
- Or the builds are already started,
  then, we cancel the builds if possible (i.e on the same master)

This stuff definitively would need unit tests.

Doing early pull request to get first feedback on whether I'm doing something sane.

### DIFF
--- a/master/buildbot/db/buildrequests.py
+++ b/master/buildbot/db/buildrequests.py
@@ -96,6 +96,37 @@ class BuildRequestsConnectorComponent(base.DBConnectorComponent):
                      for row in res.fetchall() ]
         return self.db.pool.do(thd)
 
+
+    def getChildrenBuildRequests(self, buildername, buildnumber):
+        def thd(conn):
+            reqs_tbl = self.db.model.buildrequests
+            props_tbl = self.db.model.buildset_properties
+            props_aliases = {}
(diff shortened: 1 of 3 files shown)


gtmacdonald:
For those like myself that need to have this implemented and can't wait for Nine; I was able to drop in this code into 0.8.12, without too much trouble. Mind the return in stopTriggeredBuilds; it needs to be out a few scopes or only one triggered build will stop. And if the properties stuff gets hairy with your version (like it did for me) you can always set the two properties needed directly. The db code worked as is for me.

Thx @tardyp this would've taken me much longer to get going. :)

jaredgrubb:
Could you possibly offer the patch as a pull request? I'd love to see it :)

sa2ajj:
Interesting, I also implemented stop/cancel builds in eight :)  (GH-1686)  What's missing in that?

tardyp:
@sa2ajj What you implemented is different from what is needed here.

Here this is more about the trigger step which shall interrupt the triggered buildrequest when it is itself interrupted.

PR #859, opened 2013-09-03
Fix bug in parameter validation, empty arg should be remove from args li...

ylebon:
When using a forcescheduler Stringparameter the option "required" option is not taken account as the args  length is always bigger than 0.

### DIFF
--- a/master/buildbot/schedulers/forcesched.py
+++ b/master/buildbot/schedulers/forcesched.py
@@ -88,6 +88,12 @@ class BaseParameter(object):
            the optional regex field and calls 'parse_from_args' for the final conversion.
         """
         args = kwargs.get(self.fullName, [])
+
+        # delete white space for args
+        for arg in args:
+            if not arg.strip():
+                args.remove(arg)
+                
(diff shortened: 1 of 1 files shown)


dcoshea:
I always used `regex="."` to verify that StringParameters were non-blank.  I went back through my notes/old master.cfg revisions and couldn't find any record of why I did this, but I assume it might have been because I noticed this bug where `required=True` didn't actually work.

This fix breaks using `regex="."` to verify that the parameter is at least one character long, because any blank parameters are thrown away before regex checking is performed.

I'm happy to change my master.cfg, but would it be worth including in the next release notes that are published a little more detail about this change?  The 0.8.9 release notes mention it, but it wasn't obvious to me that the `regex` was now used to check the value of the *parameter* only *if* it was specified, rather than in a way being a restriction on the final value of the *property* (in my simple case where I only expect the parameter to be specified once, I admit to ignorance of what happens if a parameter is specified more than once).  I understand if you figure that probably nobody else would be concerned about this level of detail for a change that happened over a year ago, though!

PR #1077, opened 2014-02-26
(WIP) keep HTML logs out of buildstatus pickles

jpommerening:
Trac: http://trac.buildbot.net/ticket/739

The existing `HTMLLogFile` class worked pretty well for the occasional stack-trace (`err.html`), but is not really useful for storing large HTML logs from custom buildsteps, because it keeps the whole log in the buildstatus pickle.

I tried to solve the problem by making the class inherit from `LogFile`, so it serializes and compresses the logfile, but added a little bit of magic to store small log files in the pickles and keep compatible with earlier versions.

In it's current implementation, `HTMLLogFile` will keep anything that is smaller than 10k in memory & behave (mostly) just like before. Everything beyond that threshold will be passed right through to the `LogFile` class.

WIP:
- Passing very large HTML-logs into the constructor works, but is probably not a good idea, performance wise. Especially if you first have to upload the whole thing to the master and keep it in memory until you call `addHtmlLog`.
- I currently can't run all tests on my box (TestRealDB segfaults) – hope this was not caused by my change.

### DIFF
--- a/master/buildbot/test/unit/test_status_logfile.py
+++ b/master/buildbot/test/unit/test_status_logfile.py
@@ -382,14 +382,6 @@ class TestHTMLLogFile(unittest.TestCase, dirs.DirsMixin):
         self.logfile.master = self.master
         step.build.builder.basedir = self.basedir
 
-    def delete_logfile(self):
-        if self.logfile.openfile:
-            try:
-                self.logfile.openfile.close()
-            except:
-                pass  # oh well, we tried
-        os.unlink(os.path.join('basedir', '123-error_html'))
-
     def test_unpickle(self):
         self.pickle_and_restore()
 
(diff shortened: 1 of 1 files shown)


master/buildbot/status/logfile.py
```diff
@@ -657,56 +657,28 @@ def __setstate__(self, d):
...
-        d = self.__dict__.copy()
-        del d['step']
-        if "master" in d:
-            del d['master']
-        return d
+        # buildbot <= 0.8.8 stored all html logs in the html property
+        if 'html' in self.__dict__:
+            buf = "%d:%d%s," % (len(self.html) + 1, STDERR, self.html)
+            self.openfile = StringIO(buf)
+            del self.__dict__['html']
```
djmitche: Looking back at this with @mariangemarcano I notice that this doesn't write the HTML-free pickle back to disk.  That might be a nice improvement for someone else (Maria?) to work on.

PR #1103, opened 2014-03-21
Builder tags: Adjust the name 'categories' to 'tags', as applies to builders

jaredgrubb:
This is mostly a change from using the word "categories" for builders to using the word "tags". But it also adds the ability to have multiple of them.

Note that this does not change where that tags are strored. Pierre mentioned it should go in the db, but I'm curious about that. So, this is not the final patch to fix this issue, but it's a stepping stone.

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -57,6 +57,9 @@ def error(error):
     else:
         raise ConfigErrors([error])
 
+def warnDeprecated(version, msg):
+    # for now just log the deprecation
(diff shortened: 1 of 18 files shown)


sa2ajj:
@jaredgrubb is this the PR you mentioned on [IRC](http://irclogs.jackgrigg.com/irc.freenode.net/buildbot/2015-01-20#i_3551324)?

If you have a version for eight, I'd be happy to have it included.

jaredgrubb:
Yes, this was the PR I mentioned.

PR #1167, opened 2014-06-24
git: force submodule update so local changes/files are overwritten

codyps:
Without this, we sometimes get "Your local changes to the following
files would be overwritten by checkout" when submodules are updates.

On the non-submodule side, we're already using reset instead of
checkout to avoid this.

Signed-off-by: Cody P Schafer dev@codyps.com

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -530,7 +530,7 @@ class Git(Source):
     def _updateSubmodule(self, _=None):
         if self.submodules:
             return self._dovccmd(['submodule', 'update',
-                                  '--init', '--recursive'])
+                                  '--init', '--recursive', '--force'])
         else:
             return defer.succeed(0)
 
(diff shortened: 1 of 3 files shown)


krf:
I just cherry-picked this change and noticed that Git version 1.7.1 (from a CentOS6 slave) doesn't know about the `--force` parameter. This will make the Git step fail on this Git version.

Not sure about the minimum support Git version by Python but I do see version checks against 1.6.5 in `git.py`. Worth fixing?

jmesmon:
Probably. Looking through git's logs, it appears 01d4721565f1562172149ebb91d29d6a7e7f922d (v1.7.11.6, v1.7.12.1, v1.8.0) introduces some of the behavior of `submodule update --force` that these changes expect to have, while the option itself is introduced by 0d405d72f5ffb499b40dd3a95886ae5a1fc13b30 (first included in v1.7.6).

If we just care about not breaking due to the added option, the second versions is appropriate. If we want to try to prevent users from running into git submodule errors, the first set of versions would be a better choice.

krf:
Thanks for looking through git's history. So, I see several solutions to this issue:
- For Git <v1.7.6: Don't add --force (trivial to implement in git.py, I think)
- State in docs, that Git >=v1.7.6 is required

Maybe a question for the BB maintainers.

sa2ajj:
I'd say conditional support in git.py is preferred: people use Buildbot on all kind of old setups (since it just works for them :)), so impose a new requirement to the external environment does not seem reasonable.

djmitche:
Agreed, and we already do some special-casing of older git versions, so
adding a little more doesn't hurt.

krf:
Fixed by #1880, fwiw.

PR #1217, opened 2014-09-02
Make user name in GerritChangeSource optional.

nierob:
The user name is not required and defaults to the .ssh/config
file content.


djmitche:
This hasn't been updated for ages!  I'm going to close it to clear up the pull request queue, but of course the code is still here for you or someone else to pick up and run with!

PR #1272, opened 2014-10-20
Allow user to specify baseURL for GithubStatus

kchange:
Looks like there was an issued open for this a few months ago. <http://trac.buildbot.net/ticket/2876>. The changes were made to txgithub (tomprince/txgithub#7), but never carried over to buildbot. 

I just added the parameter to ```__init__``` to take in a custom url and passed it to txgithub's GithubApi. Relevent code: https://github.com/tomprince/txgithub/blob/master/txgithub/api.py#L37-L47


sa2ajj:
Will be continued in GH-1487

PR #1426, opened 2014-12-08
Issue 3019.5

kassemsandarusi:
This should add the property 'maxcpus' to a buildslaves configuration. If 'maxcpus' is not specified it defaults to 'None' which uses 'multiprocessing.cpu_count()' to receive the max amount of CPUs.

### DIFF
--- a/slave/buildslave/bot.py
+++ b/slave/buildslave/bot.py
@@ -17,6 +17,7 @@ import os.path
 import signal
 import socket
 import sys
+import multiprocessing
 
 from twisted.application import internet
 from twisted.application import service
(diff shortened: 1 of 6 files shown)


slave/buildslave/bot.py
```diff
@@ -13,6 +13,7 @@
 #
 # Copyright Buildbot Team Members
 
+import multiprocessing
```
sa2ajj: I believe, we just dropped support for Python 2.4

djmitche: Correct, this objection no longer applies.

sa2ajj:
and a rebase

jaredgrubb:
This has been pending a while. Is it just waiting for a rebase and docs update? Or were there other issues?

sa2ajj:
Only a rebase and the docs, that's it.

PR #1458, opened 2014-12-18
Instead of hard-coding a list of modes in svn.startVC(), use getattr instead

dburkart:
This patch allows users of buildbot to subclass the SVN class and create a new
mode. The only way to do this currently is by copying the entire startVC() class
to ensure that the new mode callback is added in the proper place.


dburkart:
@benallard: I'll get to implementing your suggestions for this PR, and look into the broader stuff later. Thanks!

sa2ajj:
Superseded by #1573

PR #1468, opened 2014-12-28
Improve api for trigger step

tardyp:
Trigger step was still difficult to subclass in order to add dynamic scheduler and property choosing.
Now it is much easier

While documenting this new API, it also improves a lot the doc for this step.

With this, you don't need to copy paste the entire step in order to implement travis.yml controlled parallel builds:
https://github.com/tardyp/buildbot_travis/blob/master/buildbot_travis/steps/spawner.py

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -84,33 +84,27 @@ class Trigger(BuildStep):
             self.ended = True
 
     # Create the properties that are used for the trigger
-    def createTriggerProperties(self):
+    def createTriggerProperties(self, properties):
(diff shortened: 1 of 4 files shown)


Stibbons:
Agree with MultiTriggerStep.

tardyp:
We had internal discussion with @Stibbons , and I think there is an agreement that it will be complicated to address the complex trigger use cases we have without changing radically the API.
Hence we will be working on proposing a MultiTriggerStep similar to what we did we ShellSequence.

Meanwhile, these changes improve implementation and documentation of the current Trigger step, while being compatible with documented API. So I think this is a good step forward and should be merged.

PR #1476, opened 2015-01-06
Dynamic build

thedylman:
One of the things I am trying to do is move the logic for building a project out of Buildbot, so that when you run on a build locally on a command line or on Buildbot they are as close as possible. Also I don't need to restart Buildbot every time the build commands change.

You could just have a single shell script which is run by Buildbot in a single stage, but it makes it difficult to see where something has gone wrong. This change allows you to add steps during the build, which allows the list of steps to change per changeset.

The workflow would be something like:
* Build starts
* clone a tree
* Run a command to generate the list of stages/commands
* Use the output from the command to generate and add steps to the build
* Run newly added commands
* Run upload/cleanup steps

Still need to add test, but I would like your comments on if this is the right way to do it.

Demo config:
```python
from buildbot.plugins import buildslave, schedulers, util, steps
c = BuildmasterConfig = {}
c['slaves'] = [buildslave.BuildSlave("example-slave", "pass")]
c['schedulers'] = []
c['schedulers'].append(schedulers.ForceScheduler(name="force", builderNames=["runtests"]))

from buildbot.process.build import DynamicBuild
class DynamicBuildFactory(util.BuildFactory):
    buildClass = DynamicBuild

factory = DynamicBuildFactory()
factory.addStep(steps.ShellCommand(name="first step", command=["echo", "first"]))

from buildbot.status.results import SUCCESS
class GetStagesCommand(steps.BuildStep):
    def run(self):
        # e.g. run a command to get list of stages from source,
        # then build the list of commands. (hard coded here for demo)
        self.build.addStepsAfterCurrentStep([
            steps.ShellCommand(name="command1", command=["sleep", "5"]),
            steps.ShellCommand(name="command2", command=["sleep", "5"]),
            steps.ShellCommand(name="commandx", command=["sleep", "5"]),
            steps.ShellCommand(name="commandx", command=["sleep", "5"]),
        ])
        return SUCCESS

factory.addStep(GetStagesCommand(name="dynamic step"))
factory.addStep(steps.ShellCommand(name="last step", command=["echo", "last"]))

c['builders'] = []
c['builders'].append(
    util.BuilderConfig(name="runtests",
      slavenames=["example-slave"],
      factory=factory))

c['status'] = []
c['title'] = "Pyflakes"
c['titleURL'] = "https://launchpad.net/pyflakes"
c['buildbotURL'] = "http://localhost:8010/"
c['www'] = {"port": 8020, "plugins" : {"waterfall_view": {"num_builds": 50}}}
c['db'] = {"db_url" : "sqlite:///state.sqlite"}
```
dynamic-stages


tardyp:
I like this!

Implementation is good and straighforward.

I'll go even further and add those APIs to the base Build class, so that we don't need the DynamicBuildFactory. The new API does not break the old one, so its okay.

Needs docs and unit tests also.
You can add most of the PR description in a "Dynamic BuildFactories" section of http://docs.buildbot.net/latest/manual/cfg-buildfactories.html

thedylman:
Changes:
* Removed DynamicBuildFactory and add functions to base class, updated demo config to match.
* Added some unit tests

In the unit tests is it ok to use the step names to check the order? Is checking the order enough?
Is everyone ok with the function names addStepsAfterCurrentStep and addStepsAfterLastStep?

Docs still to do ...

jaredgrubb:
This is pretty good! I've wanted something like this a long time and have brainstormed on how to do it .. and I think you nailed it! I'm tempted to try to backport this to eight.... :)

master/buildbot/process/build.py
```diff
@@ -323,35 +323,38 @@ def acquireLocks(self, res=None):
...
-        sps = []
+    def setUniqueStepName(self, step):
+        # If there are any name collisions, we add a count to the loser
+        # until it is unique.
+        name = step.name
+        if name in self.stepnames:
+            count = self.stepnames[name]
+            count += 1
+            self.stepnames[name] = count
+            name = step.name + "_%d" % count
```
sa2ajj: [nit] `name = '%s_%d' % (step.name, count)`

sa2ajj:
LGTM

Since one of the original goals is "Use the output from the command to generate and add steps to the build" and example how this is achieved would be very handy.

sa2ajj:
I think the docs are good, and with such a doc relnotes just need to point to the relevant section.

tardyp:
@jaredgrubb FYI, buildbot_travis has a working implementation of dynamic buildstep https://github.com/isotoma/buildbot_travis/blob/master/buildbot_travis/steps/create_steps.py#L160

thedylman:
fixed the docs and added a note in the release notes.

sa2ajj:
@thedylman it looks good now.  Can you do a favour?  Put relnotes update in a separate commit?  (It helps with backporting...)

thedylman:
I've split the relnotes into a separate commit.

PR #1484, opened 2015-01-14
Buildbot UI: fixes related to config.url not existing anymore

tardyp:
- Replace config.url logic by use of $location
- modernize the buildrequest page to use breadcrumb, and buildsummary. This page is purely for debug, as there is nothing that points there, but we can still have something nice enough. 

image


djmitche:
Looks good, but the status.py changes are causing some failures for the trigger step, seen in travis.

tardyp:
tests were right. The issue was in the routing config.

PR #1486, opened 2015-01-16
Systemsettings

MaximeSangoi:
Provider for generating settings
- Ability to add checkbox type settings
- Ability to add radio type settings
- Ability to add number type settings
- Ability to add text type settings
- Saving in LocalStorage


www/base/src/app/settings/service/settings.service.coffee
```diff
@@ -0,0 +1,44 @@
+class bbSettings extends Provider
+    constructor: ->
+        @groups = {}
+
+
+    mergeNewGroup: (oldGroup, newGroup) ->
```
tardyp: this is private method, so this should be prefixed with _

www/base/src/app/settings/service/settings.service.spec.coffee
```diff
@@ -0,0 +1,288 @@
+describe 'settingsService', ->
+    bbSettingsServiceProviderRef = ""
+    beforeEach module "guanlecoja.ui", (bbSettingsServiceProvider) ->
```
tardyp: I think the module should be "app", not 'guanlecoja.ui' I dont know how this can work like this..

www/base/src/app/settings/settings.tpl.jade
```diff
@@ -0,0 +1,25 @@
+.container
+  .row
+    .well
+        h2 Settings
+        .row(ng-repeat="group in ::settingsGroups")
+            h3 {{group.caption}}
+            form(name="{{ group.name }}")
+            div(ng-repeat="item in group.items")
+                div(ng-switch="", on="item.type" class="col-md-12")
```
tardyp: please avoid class=, and prefer jade shortcut.
```
 div.col-md-12(ng-switch="", on="item.type")
```

tardyp:
fix for http://trac.buildbot.net/ticket/3087

www/base/src/app/settings/service/settings.service.coffee
```diff
@@ -0,0 +1,44 @@
...
+    $get: [ ->
+        self = @
+        return {
+            getSettingsGroups: -> 
+                self.groups
+            save: ->
+                localStorage.setItem('settings', angular.toJson(self.groups))
+                null
+            getSetting: (settingSelector) ->
+                groupName = settingSelector.split('.')[0]
```
yetanotherion: I'd store the result of settingSelector.split('.') in a variable

www/base/src/app/settings/settings.route.coffee
```diff
@@ -0,0 +1,27 @@
...
+
+        # Register new state
+        state =
+            controller: "#{name}Controller"
+            templateUrl: "views/#{name}.html"
+            name: name
+            url: '/settings'
+            data: cfg
+
+        $stateProvider.state(state)
```
yetanotherion: maybe a newline here (so that github does not complain)

www/base/src/app/settings/service/settings.service.spec.coffee
```diff
@@ -0,0 +1,288 @@
...
+        userSetting1 = bbSettingsService.getSetting('User.checkbox1')
+        userSetting2 = bbSettingsService.getSetting('User.whatever')
+        userSetting3 = bbSettingsService.getSetting('UserAA.User_checkbox1')
+        expect(userSetting1).toBeDefined()
+        expect(userSetting2).toBeUndefined()
+        expect(userSetting3).toBeUndefined()
+
+    it 'should save correct settings', inject (bbSettingsService) ->
+        checkbox = bbSettingsService.getSetting('User.checkbox1')
+        expect(checkbox.value).toBeFalsy()
```
omonnier: expect(checkbox.value).toBe(false) is more strict.
if checkbox.value is undefined, expect(checkbox.value).toBeFalsy() is true

www/base/src/app/settings/service/settings.service.spec.coffee
```diff
@@ -0,0 +1,288 @@
...
+                value: false
+                default_value: false
+            ,
+                type:'bool'
+                name:'radio2'
+                value: false
+                default_value: false
+            ]
+
+    it 'should merge groups when new group has no name', inject (bbSettingsService) ->
```
omonnier: it 'should not add a group without name ' is maybe a better description ? (merge method is not called)

www/base/src/app/settings/service/settings.service.coffee
```diff
@@ -0,0 +1,45 @@
+class bbSettings extends Provider
```
omonnier: the class is injected below as bbSettingsService. As I'm quite new to AngularJS, I'm curious to understand what's the magic that make the link between bbSettings and bbSettingsService

www/base/src/app/settings/settings.controller.coffee
```diff
@@ -0,0 +1,11 @@
+class Settings extends Controller
+    constructor: ($scope, config, bbSettingsService) ->
```
omonnier: config is unused

www/base/src/app/settings/service/settings.service.coffee
```diff
@@ -0,0 +1,45 @@
...
+            return newGroup
+        else 
+            for newItem in newGroup.items
+                newItem.value = newItem.default_value
+                for oldItem in oldGroup.items
+                    if newItem.name is oldItem.name and oldItem.value?
+                        newItem.value = oldItem.value
+            return newGroup
+
+    _addSettingsGroup: (group) ->
```
omonnier: why nobody is calling this private method ?

www/base/src/app/settings/service/settings.service.coffee
MaximeSangoi: That magic is actually ng-classify magic. See here : https://github.com/CaryLandholt/ng-classify#provider

But in Guanlecoja, the default renaming of 'blabla' in 'blablaProvider' is actually 'blablaServiceProvider', so all the services are named and accessed with 'blablaService'.

www/base/src/app/settings/service/settings.service.coffee
MaximeSangoi: It's called in _addSettingsGroup few lines underneath

MaximeSangoi: _addSettingsGroup can be used in every module config (in the route.coffee). It's not called here because there is no relevant data settings to include here

MaximeSangoi:
Applied last patch yesterday. I think I answered all the questions.

djmitche:
It looks like this is ready to merge, then?

sa2ajj:
Thank you, @MaximeSangoi

PR #1488, opened 2015-01-17
Handle `\r\n` and `\r` in LineBoundaryFinder

djmitche:
Including careful handling of the case where `\r\n` is split along a
buffer boundary.  Fixes #3144.


tardyp:
cool! :+1:

PR #1489, opened 2015-01-17
make sure version is always defined before use

sa2ajj:


djmitche:
..just waiting for Travis

PR #1492, opened 2015-01-19
Fix build URL for the builder page

delanne:
Fix regression introduced in 159b5696fbc17cbd42ee61cde0940aa72f8eed9a

### DIFF
--- a/www/base/src/app/builders/builder/builder.tpl.jade
+++ b/www/base/src/app/builders/builder/builder.tpl.jade
@@ -32,7 +32,7 @@
         td Status
       tr(ng-repeat='build in builds | orderBy:"-started_at"')
           td(width='20px')
-            a(href='#/builders/{{ builder.id }}/build/{{build.number}}',title='{{build}}')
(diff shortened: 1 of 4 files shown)


sa2ajj:
LGTM.

Should we have some test cases for this?

delanne:
no idea. ptardy ?

tardyp:
I saw the issue as well this morning. Sorry I messed up yesterday.

in order to avoid such issues, I fixed the url to use ui-sref instead of hardcoded routes.

https://github.com/buildbot/buildbot/issues/1493

In order to validate such thing we need e2e tests with protractor, which is more complicated to setup.

tardyp:
:-1 in favor of my solution

djmitche:
Is 0484a57 necessary after merging #1493?

sa2ajj:
What do we do with this PR?  Looks like it just needs to be closed.

sa2ajj:
I'll close it as I think it's not needed any more.

PR #1494, opened 2015-01-22
[WIP/DO NOT MERGE] improve SVN poller & checkout step documentation

sa2ajj:
The seemingly common problem is: how do I track all branches in my SVN project and then build the detected changes?

The current examples are not complete and I spent some time trying to figure out how to do it after a user on IRC channel asked for help.  Another user said he uses old-style svn checkout step because it has `defaultBranch` option, which is not present in the new style step.

### DIFF
--- a/master/docs/guides/svn.rst
+++ b/master/docs/guides/svn.rst
@@ -1,0 +1,160 @@
+================
+Subversion Guide
+================
+
+`Apache Subversion <https://subversion.apache.org/>`_ is one of the version control systems that is quite popular in both open sources world as well as in enterprise setting.
+Since there are different ways Subversion can be used, we cover the most popular approaches in this guide:
+
+.. contents::
+   :local:
+   :depth: 2
+
+Following Trunk
+===============
+
+Let's assume you have a Subversion repository that keeps history of one project only, it is accessible as ``http://example.org/svn`` and you use the `standard layout <http://svnbook.red-bean.com/en/1.7/svn.branchmerge.maint.html#svn.branchmerge.maint.layout>`_ in your repository:
+
+.. code-block:: none
+
+    /
+        trunk/
+        branches/
+        tags/
+
+Let's also assume that your project uses :command:`make`, running ``make`` builds the program and running ``make test`` executes the tests.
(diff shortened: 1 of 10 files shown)


sa2ajj:
The idea is create a "Guides" section where these SVN related things are covered (in this PR) and then populate the section with similar guides for other version control systems or other difficult to grasp concepts.

The first draft for SVN is available as http://bb.mawhrin.net/guides/svn.html

master/docs/guides/svn.rst
```diff
@@ -0,0 +1,92 @@
+=================================
+Subversion Version Control System
```
djmitche: Maybe use a more detailed description for this particular guide?

sa2ajj: I'm open for suggestions :)

That was the best I could come out with.

djmitche: Maybe split each of the sections here into a separate guide?  Those after the first could refer back to the first ("You should be familiar with..").  So the guide heading would be "Building Every Commit to Trunk with Subversion"

sa2ajj: I considered at least two other guides:

* git (various practices specific for git related components)
* github (changes, status, maybe something else)

In the light of these ideas, a guide for SVN looked logical.  If I split all these (and following guides) into really small parts, then it's gonna be something else...

djmitche: That does seem like a lot of guides, it's true.  Maybe I just need to adjust the expectation I have from the word "guide" -- these are each focused on a particular usage topic that might involve content from all over the documentation (particularly for version-control systems, which will reference both change sources and source steps).

sa2ajj: > these are each focused on a particular usage topic that might involve content from all over the documentation (particularly for version-control systems, which will reference both change sources and source steps).

This is exactly what I have in mind.

Let's return to the topic of title when I think this "guide" is ready.

djmitche:
I like "Cookbook", as you just suggested in irc.

djmitche:
Should we keep this open?

sa2ajj:
Yes, please do.  I hope to return to this RSN(tm).

sa2ajj:
@djmitche on the second though, I close this (I still have the branch, do not I :))

PR #1502, opened 2015-01-23
Tags: backport #928 to eight

jaredgrubb:
This is a backport to eight of the following change:
> commit 0d222614c5996b0b3bc3ca0dc30629c4199f50a4
> Author: Jared Grubb <jaredgrubb@gmail.com>
> Date:   Mon Feb 10 12:53:43 2014 -0800
> 
>     Fix #928: Allow builder to be associated with multiple tags

For the most part, this patch is identical to the one I contributed to nine. However, the Status parts of the patch (eg, waterfall) are all brand new, since Status does not exist in this form in Nine. 

I'm sure there's going to be some issues, as I had some questions on parts of this patch (I'll write them inline below). 
Also, docs arent ported and notes are not updated. I'll do that next, but wanted to get this up for comment.

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -636,7 +636,8 @@ class MasterConfig(object):
 class BuilderConfig:
 
     def __init__(self, name=None, slavename=None, slavenames=None,
-                 builddir=None, slavebuilddir=None, factory=None, category=None,
+                 builddir=None, slavebuilddir=None, factory=None,
+                 category=None, tags=None,
                  nextSlave=None, nextBuild=None, locks=None, env=None,
                  properties=None, mergeRequests=None, description=None,
                  canStartBuild=None):
@@ -688,10 +689,25 @@ class BuilderConfig:
         self.slavebuilddir = slavebuilddir
 
         # remainder are optional
-        if category is not None and not isinstance(category, str):
-            error("builder '%s': category must be a string" % (name,))
+        if category and tags:
+            error("builder '%s': category is being replaced by tags; you should only specify tags" % (name,))
 
-        self.category = category or ''
+        if category:
+            if not isinstance(category, str):
+                error("builder '%s': category must be a string" % (name,))
+            tags = [category]
+
+            self.category = category  # Set this for legacy reasons
+
+        if tags:
+            if not isinstance(tags, list):
+                error("builder '%s': tags must be a list" % (name,))
+            bad_tags = any((tag for tag in tags if not isinstance(tag, str)))
+            if bad_tags:
+                error("builder '%s': tags list contains something that is not a string" % (name,))
+
+        self.tags = tags
+
         self.nextSlave = nextSlave
         if nextSlave and not callable(nextSlave):
             error('nextSlave must be a callable')
@@ -721,8 +737,8 @@ class BuilderConfig:
             'builddir': self.builddir,
             'slavebuilddir': self.slavebuilddir,
         }
-        if self.category:
-            rv['category'] = self.category
+        if self.tags:
+            rv['tags'] = self.tags
         if self.nextSlave:
             rv['nextSlave'] = self.nextSlave
         if self.nextBuild:
--- a/master/buildbot/status/builder.py
+++ b/master/buildbot/status/builder.py
@@ -62,23 +62,23 @@ class BuilderStatus(styles.Versioned):
     I live in the buildbot.process.build.Builder object, in the
     .builder_status attribute.
 
-    @type  category: string
-    @ivar  category: user-defined category this builder belongs to; can be
+    @type  tags: None or list of strings
+    @ivar  tags: user-defined "tag" this builder has; can be
                      used to filter on in status clients
     """
 
     implements(interfaces.IBuilderStatus, interfaces.IEventSource)
 
-    persistenceVersion = 1
+    persistenceVersion = 2
     persistenceForgets = ('wasUpgraded', )
 
-    category = None
+    tags = None
     currentBigState = "offline"  # or idle/waiting/interlocked/building
     basedir = None  # filled in by our parent
 
-    def __init__(self, buildername, category, master, description):
+    def __init__(self, buildername, tags, master, description):
         self.name = buildername
-        self.category = category
+        self.tags = tags
         self.description = description
         self.master = master
 
@@ -134,6 +134,12 @@ class BuilderStatus(styles.Versioned):
             del self.slavename
         if hasattr(self, 'nextBuildNumber'):
             del self.nextBuildNumber  # determineNextBuildNumber chooses this
+        self.wasUpgraded = True
+
+    def upgradeToVersion2(self):
+        if hasattr(self, 'category'):
+            self.tags = self.category and [self.category] or None
+            del self.category
         self.wasUpgraded = True
 
     def determineNextBuildNumber(self):
@@ -322,12 +328,27 @@ class BuilderStatus(styles.Versioned):
             b = self.getBuild(-2)
         return b
 
-    def setCategory(self, category):
+    def getTags(self):
+        return self.tags
+
+    def setTags(self, tags):
         # used during reconfig
-        self.category = category
+        self.tags = tags
+
+    def matchesAnyTag(self, tags):
+        return self.tags and any((tag in self.tags) for tag in tags)
+
+    def matchesAllTags(self, tags):
+        return self.tags and all((tag in self.tags) for tag in tags)
+
+    def setCategory(self, category):
+        return self.setTags([category])
 
     def getCategory(self):
-        return self.category
+        tags = self.getTags()
+        if tags:
+            return tags[0]
+        return None
 
     def getBuildByRevision(self, rev):
         number = self.nextBuildNumber - 1
@@ -438,7 +459,7 @@ class BuilderStatus(styles.Versioned):
             # sourcestamps match, skip this build
             if branches and not branches & self._getBuildBranches(b):
                 continue
-            if categories and not b.getBuilder().getCategory() in categories:
+            if categories and not b.getBuilder().matchesAnyTag(tags=categories):
                 continue
             if committers and not [True for c in b.getChanges() if c.who in committers]:
                 continue
@@ -583,7 +604,7 @@ class BuilderStatus(styles.Versioned):
         # Constant
         # TODO(maruel): Fix me. We don't want to leak the full path.
         result['basedir'] = os.path.basename(self.basedir)
-        result['category'] = self.category
+        result['tags'] = self.tags
         result['slaves'] = self.slavenames
         result['schedulers'] = [s.name
                                 for s in self.status.master.allSchedulers()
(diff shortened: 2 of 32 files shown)


sa2ajj:
(_just_ 29 files... easy to review...)

master/buildbot/config.py
```diff
@@ -688,10 +689,25 @@ def __init__(self, name=None, slavename=None, slavenames=None,
...
-            error("builder '%s': category must be a string" % (name,))
+        if category and tags:
+            error("builder '%s': category is being replaced by tags; you should only specify tags" % (name,))
+
+        if category:
+            if not isinstance(category, str):
+                error("builder '%s': category must be a string" % (name,))
+            tags = [category]
+
+            self.category = category  # Set this for legacy reasons
```
jaredgrubb: This line was not part of the Nine review. I added it just in case subclasses had come to depend on it being there, but maybe that's not a concern.

master/buildbot/status/builder.py
```diff
@@ -62,23 +62,23 @@ class BuilderStatus(styles.Versioned):
...
-    @ivar  category: user-defined category this builder belongs to; can be
+    @type  tags: None or list of strings
+    @ivar  tags: user-defined "tag" this builder has; can be
                      used to filter on in status clients
     """
 
     implements(interfaces.IBuilderStatus, interfaces.IEventSource)
 
-    persistenceVersion = 1
+    persistenceVersion = 2
```
jaredgrubb: Note that this persistence is identical to the version done in Nine, so there is no version conflict (ie, both versions have a "version 2" and they are identical).

master/buildbot/status/web/builder.py
```diff
@@ -602,10 +602,13 @@ def content(self, req, cxt):
...
-        if showCategories == []:
-            showCategories = None
-        builders = req.args.get("builder", status.getBuilderNames(categories=showCategories))
+        showTags = req.args.get("tag", [])
+        if not showTags:
+            showTags = req.args.get("category", [])
+            if not showTags:
+                showTags = None
+
+        builders = req.args.get("builder", status.getBuilderNames(tags=showTags))
```
jaredgrubb: For the status objects, I decided it was probably better to not break existing bookmarks, so all the request checks will accept 'category' (if 'tag' wasnt there).

master/buildbot/status/web/console.py
```diff
@@ -284,24 +284,26 @@ def getAllBuildsForRevision(self, status, request, codebase, lastRevision,
...
             # This is a hack to support the chromium usecase where they have
-            # multiple categories for each slave. We use only the first one.
+            # multiple tags for each slave. We use only the first one.
             # TODO(nsylvain): Create another way to specify "display category"
             #     in master.cfg.
-            category = category.split('|')[0]
-            if not builderList.get(category):
-                builderList[category] = []
+            if len(tags)==1:
+                tags = tags[0].split('|')
```
jaredgrubb: I'm really not sure about these two lines (300, 301), because I dont fully understand what the intent was in the comment. But, I'm hoping I preserved functionality.

master/buildbot/status/builder.py
```diff
@@ -322,12 +328,27 @@ def getLastFinishedBuild(self):
...
+
+    def setCategory(self, category):
+        return self.setTags([category])
 
     def getCategory(self):
-        return self.category
+        tags = self.getTags()
+        if tags:
+            return tags[0]
+        return None
```
jaredgrubb: Note that I did not remove the {get,set}Category functions here. They shouldnt really be used, but I wasnt sure if that would unncecesarily break user code. So, I figured it was easier to just leave it in and have it return the first tag.

sa2ajj: I'd add a deprecation warning here and in nine these can be just removed (if they are not already).

master/buildbot/status/web/console.py
sa2ajj: Two questions:

* do we have tests for this special approach?
* if the intent is not clear, why keep it at all?

master/buildbot/status/builder.py
```diff
@@ -322,12 +328,27 @@ def getLastFinishedBuild(self):
...
+    def getTags(self):
+        return self.tags
+
+    def setTags(self, tags):
         # used during reconfig
-        self.category = category
+        self.tags = tags
+
+    def matchesAnyTag(self, tags):
+        return self.tags and any((tag in self.tags) for tag in tags)
```
sa2ajj: If either parameter `tags` or `self.tags` is empty, the result is `False`, this is expected, right?

master/buildbot/status/builder.py
```diff
@@ -322,12 +328,27 @@ def getLastFinishedBuild(self):
...
+
+    def matchesAllTags(self, tags):
+        return self.tags and all((tag in self.tags) for tag in tags)
```
sa2ajj: Side note: using sets would be faster here and in the previous function.

master/buildbot/status/mail.py
```diff
@@ -299,12 +301,15 @@ def __init__(self, fromaddr, mode=("failing", "passing", "warnings"),
...
-        @param categories: a list of category names to serve status
+        @type  tags: list of strings
+        @param tags: a list of tag names to serve status
                            information for. Defaults to None (all
-                           categories). Use either builders or categories,
+                           categories). Use either builders or tags,
                            but not both.
 
+        @type  categories: list of strings
+        @param categories: (this attribute is deprecated; use 'tags' instead)
```
sa2ajj: No deprecation warning in this constructor?

master/buildbot/status/web/console.py
sa2ajj: Chromium uses version 0.8.4p1 (with own patches).  If they upgraded to the latest release (or, at least, to 0.8.9), I'd see some reason to keep this hack in.

master/buildbot/status/web/console.py
```diff
@@ -316,26 +318,26 @@ def getAllBuildsForRevision(self, status, request, codebase, lastRevision,
...
+        for tag in builderList:
+            count += len(builderList[tag])
 
-        categories = sorted(builderList.keys())
+        tags = sorted(builderList.keys())
 
         cs = []
 
-        for category in categories:
+        for tag in tags:
```
sa2ajj: [nit] This loop calls for a list comprehension...

sa2ajj:
LGTM

sa2ajj:
pyflakes error:

```
master/buildbot/status/web/waterfall.py:325: import 'builder' from line 32 shadowed by loop variable
```

master/buildbot/status/builder.py
jaredgrubb: builder.matchesAnyTags(["foo"]) should be False if that builder has no tags (eg, looking for the "Linux" builders should not match things that never opted in to be "Linux" builders).

builder.matchesAnyTags([]) ... I am not sure what this should be. I could make arguments both ways.

master/buildbot/status/builder.py
jaredgrubb: The Status objects do not exist in Nine, so there's not really any deprecation on this object.

master/buildbot/status/builder.py
sa2ajj: Let's keep it the way you already have it then.

master/buildbot/status/web/console.py
jaredgrubb: Yes, this would definitely look better as a list-comp ... In general, I tried to not rewrite code that wasnt actually changing behavior. I just have been wrong too many times by "obviously the same thing" kinds of changes :) Since this code is dead-end in eight, I'm tempted to just let it be unless you have a strong objection.

sa2ajj: No strong objection here :)

master/buildbot/status/builder.py
sa2ajj: Ok.

sa2ajj:
@jaredgrubb do you think it will be mergeable by this weekend?

jaredgrubb:
Ok, I think I addressed all the comments. Also, I have updated the docs and release notes. (Note that when I did 'make docs', it failed, but it seemed to fail for other reasons).

master/buildbot/status/web/console.py
sa2ajj: @jaredgrubb so what about this functionality?  I'd rather see it gone...

jaredgrubb: I'm ok to remove it ... I dont fully understand why it's there. Looks like it was added in the first commit of Console (looks like 0.7.10) and hasnt been updated since. So, I would guess it's probably not used anymore? Plus, I think the "multiple tags" actually implements what this is trying to accomplish anyway. Shall I remove it?

sa2ajj: Yes, please :)

sa2ajj:
Looks like a pep-8 error:

```
master/buildbot/status/tinderbox.py:69:33: E261 at least two spaces before inline comment
categories=None # deprecated, use tags instead
^

master/buildbot/status/web/console.py:294:1: W293 blank line contains whitespace
^

master/buildbot/status/web/console.py:300:25: E225 missing whitespace around operator
if len(tags)==1:
^

master/buildbot/test/unit/test_status_mail.py:189:5: E303 too many blank lines (2)
def test_init_enforces_categories_and_builders_are_mutually_exclusive(self):
^

1 E225 missing whitespace around operator
1 E261 at least two spaces before inline comment
1 E303 too many blank lines (2)
1 W293 blank line contains whitespace
```

jaredgrubb:
Fixed two of the pep8 issues (the other two were removed in that chromium patch).

sa2ajj:
Thank you very much! :)

Waiting for Travis...

sa2ajj:
We seem to have an error: http://buildbot.buildbot.net/console

sa2ajj:
[TRAC-3179](http://trac.buildbot.net/ticket/3179) has a better traceback.

master/buildbot/status/web/console.py
```diff
@@ -404,21 +397,21 @@ def displayStatusLine(self, builderList, allBuilds, revision, debugInfo):
...
-        categories = sorted(builderList.keys())
+        # Sort the tags.
+        tags = sorted(builderList.keys())
 
         builds = {}
 
-        # Display the boxes by category group.
-        for category in categories:
+        # Display the boxes by tag group.
+        for tag in tags:
```
sa2ajj: Inside this `for..` loop there was a variable `tag` used, which overwrites the value assigned here.

master/buildbot/status/web/console.py
```diff
@@ -475,7 +468,7 @@ def displayStatusLine(self, builderList, allBuilds, revision, debugInfo):
                 b["color"] = resultsClass
                 b["tag"] = tag
```
sa2ajj: This tag has nothing to do with the `tag` assigned in the loop.

sa2ajj:
#1514 partially fixes things.

PR #1503, opened 2015-01-23
Add better support for modularized master.cfg

vlovich:
master.cfg might unload all modules on reconfig.
Unclear if this is all the spots or the best fix, but it's what I
encountered and seems to fix the issue.


vlovich:
Never mind.  The way I was reloading stuff was incorrect.

PR #1504, opened 2015-01-23
Sschanges

tardyp:
Some steps e.g Gerrit are using change properties in order to work.

Populate the TempSourceStamp class with TempChanges..
I am not sure what is the impact of doing things right here @djmitche maybe you have better idea?


djmitche:
I like this.  In theory, there should only ever be zero or one changes associated with a sourcestamp, but a list is a good way to express that and leaves the option of having multiple changes later.

PR #1506, opened 2015-01-25
use buildbot.tac from the current directory as default for checkconfig

thedylman:
When `buildbot checkconfig` is used without an argument default to using buildbot.tac in the current directory to find the config file.

This makes checkconfig work like `buildbot start` where it defaults to the current directory.

### DIFF
--- a/master/buildbot/scripts/runner.py
+++ b/master/buildbot/scripts/runner.py
@@ -487,13 +487,12 @@ class CheckConfigOptions(base.SubcommandOptions):
 
     def getSynopsis(self):
         return "Usage:\t\tbuildbot checkconfig [configFile]\n" + \
-            "\t\tIf not specified, 'master.cfg' will be used as 'configFile'"
+            "\t\tIf not specified, the config file specified in " + \
+            "'buildbot.tac' from the current directory will be used"
 
     def parseArgs(self, *args):
         if len(args) >= 1:
             self['configFile'] = args[0]
-        else:
-            self['configFile'] = 'master.cfg'
 
 
 class UserOptions(base.SubcommandOptions):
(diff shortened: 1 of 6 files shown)


master/buildbot/scripts/runner.py
```diff
@@ -487,13 +488,14 @@ class CheckConfigOptions(base.SubcommandOptions):
...
-            "\t\tIf not specified, 'master.cfg' will be used as 'configFile'"
+            "\t\tIf not specified, the config file specified in " + \
+            "'buildbot.tac' from the current directory will be used"
 
     def parseArgs(self, *args):
         if len(args) >= 1:
             self['configFile'] = args[0]
         else:
-            self['configFile'] = 'master.cfg'
+            self['configFile'] = os.getcwd()
```
sa2ajj: The name of the variable contradicts its value.

thedylman: yes checkconfig() expects configFile to contain a path to the config file or a directory.
How about setting it to None and then using os.getcwd() if it is not set in checkconfig()?

sa2ajj: I like your suggestion more.

`checkconfig` already magically uses `os.getcwd()`...

thedylman:
Moved default value into checkconfig.py and add a test for default value.

sa2ajj:
LGTM

djmitche:
Can you update docs and add a release note?

thedylman:
What bit of the docs need updating?

djmitche:
Wow, checkconfig isn't even in `cmdline.rst`.  This might be a good time to fix that omission?

thedylman:
Added docs and a release note.

sa2ajj:
Thank you, @thedylman :)

PR #1510, opened 2015-01-31
Try to call startBuild for all slaves, not just latent ones.

tomprince:


tomprince:
I'm using this [here](https://github.com/ClusterHQ/build.clusterhq.com/blob/master/flocker_bb/ec2_buildslave.py).

djmitche:
Looks good

PR #1513, opened 2015-02-01
Slave proto: introduce proxies for RemoteCommand, FileWriter and FileReader

tardyp:
These changes for http://trac.buildbot.net/ticket/2437 remove the pb.Referenceable in buildbot protocol.

Introducing proxies that are documented in the cls-protocols.rst file

Despite the size of the patch, there is so much new code actually, just a bunch of refactorization
A new integration test to make sure that transfers are actually working with the proxy.
Other integration tests where already taking care of RemoteCommand

### DIFF
--- a/master/buildbot/buildslave/protocols/base.py
+++ b/master/buildbot/buildslave/protocols/base.py
@@ -68,6 +68,9 @@ class Connection(object):
         raise NotImplementedError
 
 
+# RemoteCommand base implementation and base proxy
+
+
 class RemoteCommandImpl(object):
 
     def remote_update(self, updates):
@@ -88,3 +91,63 @@ class RemoteCommandProxy(object):
 
     def remote_complete(self, failure=None):
         return self.impl.remote_complete(failure)
+
+
+# FileWriter base implementation and base proxy
+
+class FileWriterImpl(object):
+
+    def remote_write(self, data):
+        raise NotImplementedError
+
+    def remote_utime(self, accessed_modified):
+        raise NotImplementedError
+
+    def remote_unpack(self):
+        raise NotImplementedError
+
+    def remote_close(self):
+        raise NotImplementedError
+
+
+class FileWriterProxy(object):
+
+    def __init__(self, impl):
+        assert isinstance(impl, FileWriterImpl)
+        self.impl = impl
+
+    def remote_write(self, data):
+        return self.impl.remote_write(data)
+
+    def remote_utime(self, accessed_modified):
+        return self.impl.remote_utime(accessed_modified)
+
+    def remote_unpack(self):
+        return self.impl.remote_unpack()
+
+    def remote_close(self):
+        return self.impl.remote_close()
+
+# FileReader base implementation and base proxy
+
+
+class FileReaderImpl(object):
+
+    def remote_read(self, maxLength):
+        raise NotImplementedError
+
+    def remote_close(self):
+        raise NotImplementedError
+
+
+class FileReaderProxy(object):
+
+    def __init__(self, impl):
+        assert isinstance(impl, FileReaderImpl)
+        self.impl = impl
+
+    def remote_read(self, maxLength):
+        return self.impl.remote_read(maxLength)
+
+    def remote_close(self):
+        return self.impl.remote_close()
--- a/master/buildbot/buildslave/protocols/base.py
+++ b/master/buildbot/buildslave/protocols/base.py
@@ -79,8 +79,6 @@ class Connection(object):
 
 
 # RemoteCommand base implementation and base proxy
-
-
 class RemoteCommandImpl(object):
 
     def remote_update(self, updates):
@@ -90,21 +88,7 @@ class RemoteCommandImpl(object):
         raise NotImplementedError
 
 
-class RemoteCommandProxy(object):
-
-    def __init__(self, impl):
-        assert isinstance(impl, RemoteCommandImpl)
-        self.impl = impl
-
-    def remote_update(self, updates):
-        return self.impl.remote_update(updates)
-
-    def remote_complete(self, failure=None):
-        return self.impl.remote_complete(failure)
-
-
-# FileWriter base implementation and base proxy
-
+# FileWriter base implementation
 class FileWriterImpl(object):
 
     def remote_write(self, data):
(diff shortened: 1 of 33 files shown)


jaredgrubb:
Looks good to me!

sa2ajj:
failing test:

```
[ERROR]
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_buildslave_protocols_pb.py", line 188, in test_remoteStartCommand
    conn.remoteStartCommand(RCInstance, builder_name, commandID, remote_command, args)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/buildslave/protocols/pb.py", line 192, in remoteStartCommand
    args = self.createArgsProxies(args)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/buildslave/protocols/base.py", line 40, in createArgsProxies
    for k, v in args.iteritems():
exceptions.AttributeError: 'str' object has no attribute 'iteritems'

buildbot.test.unit.test_buildslave_protocols_pb.TestConnection.test_remoteStartCommand
```

tardyp:
Added an example 'null' protocol for slaves leaving in the same reactor as the master.
I'm a bit disappointed that I only get 5% improvement of performance for the integration tests.

master/buildbot/process/buildstep.py
```diff
@@ -496,9 +496,6 @@ def setRenderable(res, attr):
         except BuildStepFailed:
             self.results = FAILURE
 
-        except BuildSlaveTooOldError:
```
tardyp: I spent a faire amount of time debugging why my step went in exception with no message. the generic except Exception catches and reports the error much more nicely

tardyp:
@djmitche @tomprince , this finishes the slave-proto gsoc (obvious appart from the amp proto). I'd appreciate some review to make sure this fits you vision

master/docs/developer/cls-protocols.rst
```diff
@@ -16,10 +19,21 @@ to know about protocol calls or handle protocol specific exceptions.
...
+    .. py:attribute:: proxies
+
+        Dictionary containing mapping between ``Impl`` classes and ``Proxy`` class for this protocol
+        This may be overridden by subclass to declare its proxy implementations
+
+    .. py:method:: createArgsProxies(args)
+
+        :returns: shallow copy of args dictionary with proxies instead of impls
+
+        Helper methods that will use :attr:`proxies`, and replace ``Impl`` objects by specific ``Proxy`` counterpart.
```
djmitche: `method` (singular)

djmitche:
This is really cool, and nicely implemented.

My *one* concern is that this continues the model of the protocol holding references to master-side objects (RemoteCommand, FileWriter, and FileReader).  Where in the past we inherited from `pb.Referencable`, we now inherit from `*Impl`.  So we get the level of abstraction we need to support non-PB protocols.  But are we then requiring protocol implementations to do some of the crazy reference-holding magic that PB itself does (where a master-side reference is kept until the reference to it on the slave goes away)?

I think the answer is "no", from what I see here, but figured I'd ask the question anyway :)

djmitche:
Also, are proxies useful for anything but PB?  Should those move to the PB protocol?

tardyp:
Le Wed Feb 04 2015 at 5:17:12 AM, Dustin J. Mitchell <
notifications@github.com> a écrit :

> This is really cool, and nicely implemented.
>
> My *one* concern is that this continues the model of the protocol holding
> references to master-side objects (RemoteCommand, FileWriter, and
> FileReader). Where in the past we inherited from pb.Referencable, we now
> inherit from *Impl. So we get the level of abstraction we need to support
> non-PB protocols.
>
I also had this concern. But I then had a bigger concern which is how we
support old slaves?
How do we implement the new buildslave package to be compatible with old
protocol, and a new protocol?

> But are we then requiring protocol implementations to do some of the crazy
> reference-holding magic that PB itself does (where a master-side reference
> is kept until the reference to it on the slave goes away)?
>
I actually think pb's referenceable concept is a good way of resolving the
problem of command id.
I did not realize the problem of the gc, but I think that can be solved.
If we don't do so, we'll anyway have to put put the problem of reference id in
the logic code.

Actually, we already have a command_id, for interrupt command, so it is
feasible.

Any way I think the problem is: do we keep legacy protocol support?

djmitche:
I think we do keep legacy protocol support -- from what I can tell, your patch does so!  Did I miss something there?

My concerns are really about whether this is a good design looking forward.  My last comment mused about that, but I don't actually see any problem.  The process code supplies some *Impl objects to the protocol implementation, and it's up to that protocol implementation to manage their lifetime appropriately.  For protocols tied to a TCP connection, that's pretty easy.  For more "ephemeral" uses, the protocol may have some work to do to be sure it doesn't keep references to those objects for too long -- but the protocol is probably the right place for that work.

So I think the only remaining concerns I have are:

 * Are proxies useful for anything but PB? Should those move to the PB protocol?
 * Does this break legacy protocol support in a way that I'm not seeing?

tardyp:
Yes my patch keeps the legacy protocol. The thing is I don't see how we can keep the legacy protocol, and not "model the protocol to hold references to master-side objects" 

Anyway, I think we have an agreement. I agree that the Proxy class are actually uneeded. I dont think I'll need them either for my wamp PoC.

So. rebased, and cleaned that part (https://github.com/buildbot/buildbot/commit/60a5917ee2354a55c7e5b932d363c4a59e44d241 is new)

master/buildbot/buildslave/protocols/base.py
```diff
@@ -66,3 +76,39 @@ def remoteStartBuild(self, builderName):
 
     def remoteInterruptCommand(self, commandId, why):
         raise NotImplementedError
+
+
+# RemoteCommand base implementation and base proxy
```
tomprince: subclasses :frowning: http://www.youtube.com/watch?v=3MNVP9-hglc

tardyp: Well, this file was already based on subclassing.. and you mentored this gsoc project, afaik :)

If we'd followed all google's best practices, we'd rewrite buildbot in golang. :->

sa2ajj:
Let's take it from here...

PR #1517, opened 2015-02-02
allow '.'s and '_'s in properties Interpolate understands

sa2ajj:
the current situation is bad since GerritChangeSource produces properties that contain '.'s and Interpolate does not allow to use them


tardyp:
LGTM

jaredgrubb:
LGTM

PR #1524, opened 2015-02-04
skip tests for the wheels. This saves 50% of the size of the wheel

tardyp:

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -30,9 +30,9 @@ from distutils.command.install_data import install_data
 from distutils.command.sdist import sdist
 
 if "bdist_wheel" in sys.argv:
-    NOT_WHEEL = False
+    BUILDING_WHEEL = True
 else:
-    NOT_WHEEL = True
+    BUILDING_WHEEL = False
 
 def include(d, e):
     """Generate a pair of (directory, file-list) for installation.
@@ -170,17 +170,17 @@ setup_args = {
         "buildbot.steps.package.deb",
         "buildbot.steps.package.rpm",
         "buildbot.steps.source",
-        "buildbot.test.util",  # test.util and test.fake are part of the framework, and can (must!) be used
-        "buildbot.test.fake",  # to write unit tests for custom steps
         "buildbot.util",
         "buildbot.www",
-    ] + (NOT_WHEEL and [  # skip tests for wheels (save 50% of the archive)
+    ] + ([] if BUILDING_WHEEL else [  # skip tests for wheels (save 50% of the archive)
         "buildbot.test",
+        "buildbot.test.util",
+        "buildbot.test.fake",
         "buildbot.test.fuzz",
         "buildbot.test.integration",
         "buildbot.test.regressions",
         "buildbot.test.unit",
-    ] or []),
+    ]),
     'data_files': [
         ("buildbot", [
             "buildbot/buildbot.png",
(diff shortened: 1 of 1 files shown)


master/setup.py
```diff
@@ -166,16 +170,17 @@ def define_plugin_entries(groups):
         "buildbot.steps.package.deb",
         "buildbot.steps.package.rpm",
         "buildbot.steps.source",
-        "buildbot.test",
+        "buildbot.test.util",
```
sa2ajj: If tests are not included, why do we want to still include `buildbot.test.{util,fake}`?

sa2ajj:
Beside the clarification I asked for, LGTM.

master/setup.py
tardyp: Because they are part of the framework. They can be used to write tests for other projects using buildbot (e.g buildbot_travis)

tardyp:
added clarification

sa2ajj:
While I agree this is a strong use case, I do not think we have enough documentation for things provided by those modules...

All in all, I'd like to hear at least one more opinion on this - @jaredgrubb, @djmitche, @tomprince?

tardyp:
indeed, we only have those http://docs.buildbot.net/latest/developer/tests.html

But imho, the test system is very well designed and strong. my team would have saved some time if they'd use it instead of rewritting their own.

master/setup.py
```diff
@@ -29,6 +29,10 @@
 from distutils.command.install_data import install_data
 from distutils.command.sdist import sdist
 
+if "bdist_wheel" in sys.argv:
+    NOT_WHEEL = False
+else:
+    NOT_WHEEL = True
```
djmitche: A boolean with "not" in the name is confusing!  Can you reverse this to "BUILDING_WHEEL" or something similar?

djmitche:
I think this is good.  I don't understand your concern, @sa2ajj -- the test utils are, I'll admit, sparsely documented, but that's isolated to the `buildbot.tests` package hierarchy and seems more-or-less par for the course for tests.

sa2ajj:
@djmitche the intention for leaving `buildbot.test.{util,fake}` in is to allow third-party packages to perform tests: since they are 3rd party, they'd need to know how to use stuff provided by these test packages, and we do not have elaborate docs for that.  That's my concern.

djmitche:
Ah, I see.  And I don't like that :)

I consider all of those to be private and only for use within Buildbot -- so if I refactor FakeFoo, I can look for all occurrences of FakeFoo and update them.

Buildbot has historically advertised a very broad and badly-controlled API surface, and that's a big part of the pain of upgrades and refactors.  I don't want to repeat that.

Even the fake DB and Data API components should, I think, be kept private.  While they mostly implement the well-defined and controlled DB and Data APIs, they also have interfaces for adding and examining test data which may change from version to version.  Sometime in the midst of 0.9.x for x > 0 I might be comfortable regularizing and documenting those interfaces, but at the moment they're still pretty fluid.

tardyp:
my opinion is that it is always better to use unsupported api rather than rewriting its own unit test framework.

Buildbot expectCommand framework is awesome for testing custom steps. Even if this is not supported, I would recommend everybody to use that rather than anything in-house.

I would also recommend to use the test.util.integration in order to create integration tests for custom steps.

Anyway If people want this they can still take the tarball.

switched to ternary if, and removed test.util and test.fake

sa2ajj:
LGTM

djmitche:
Let's get a 1.0.0 bug filed to make those test suites available later.

tardyp:
rebased

PR #1525, opened 2015-02-05
Use parent instead of self.parent in MailNotifier.setServiceParent()

ewongbb:
MailNotifier's setServiceParent() was setting self.master_status to self.parent, which doesn't exist and
I'm guessing the intention is to set it to the parameter parent.

### DIFF
--- a/master/buildbot/status/mail.py
+++ b/master/buildbot/status/mail.py
@@ -465,7 +465,7 @@ class MailNotifier(base.StatusReceiverMultiService, buildset.BuildSetSummaryNoti
                 "customMesg is deprecated; use messageFormatter instead")
 
     def setServiceParent(self, parent):
-        self.master_status = self.parent
+        self.master_status = parent
         self.master_status.subscribe(self)
         self.master = self.master_status.master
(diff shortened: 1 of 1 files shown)


tardyp:
I guess this would require a little bit more attention. Looking at the code, the master_status is still not fully mq and data aware.

We have the choice to either keep the master_status api, and make it baked by the data and mq api, or completely remove the master_status, and make all status plugin directly use data and mq api.

I would be more in favour of to the second option as I fear that the old status api is too much synchronous to really be backed by data.

tardyp:
note this is about http://trac.buildbot.net/ticket/2648

sa2ajj:
I'd say that this particula change is about a typo/overlooked change and I'm very much inclined to merge it as is.

tardyp:
OK

Le jeu. 5 févr. 2015 13:32, Mikhail Sobolev <notifications@github.com> a
écrit :

sa2ajj:
Thank you, @ewongbb :)

djmitche:
Nice work -- Pierre's right, we need to fix this stuff, but this patch is good.

PR #1526, opened 2015-02-05
[DO NOT MERGE] wamp poc

tardyp:
FYI, my poc on wamp.

I've been reading a lot of stuff lately about wamp, and crossbar. That sounds like a nice technology.

My plan was to use this as a basis for resolving http://trac.buildbot.net/ticket/3148

But this is also a serious option for master/slave protocol.

This even might be an option for mq, even if the wamp protocol looks like a bit short on mq usecases.


tardyp:
Rebased on my slave-proto branch.

I now got a WampBuildSlave running connected to the master via crossbar

Now I need to connect 200000 slaves via that crossbar..

tardyp:
FYI, this is making progress here. I have been able to run a full master + 10 slaves + crossbar setup on cloud foundry setup.

one master, 10 slaves:
cf scale buildslave -i 10
cf scale master -i 1

two masters, 20 slaves:
cf scale buildslave -i 20
cf scale master -i 2

sa2ajj:
Sounds good: what are the next steps w/ this PR (considering the recent licence chance)?

tardyp:
This branch is basically functional. I've run basic builds with it

There are still a lot of things todo..
- The tests are only about the nominal case
- The protocol is not well defined regarding slave connection/disconnection handling
- Needs doc describing the actual protocol
- Probably tons of corner case bugs

I've paused this dev because I need to make progress on buildbot_travis for work.

master/buildbot/test/integration/test_transfer.py
```diff
@@ -59,6 +59,10 @@ class TransferStepsMasterNull(TransferStepsMasterPb):
     proto = "null"
 
 
+class TransferStepsMasterWamp(TransferStepsMasterPb):
+    proto = "wamp"
```
djmitche: We should probably rename the parent class here..

master/buildbot/wamp/connector.py
```diff
@@ -0,0 +1,105 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright  Team Members
```
djmitche: What happened to " Buildbot" here?

master/buildbot/www/config.py
```diff
@@ -58,6 +58,8 @@ def renderIndex(self, request):
         config['title'] = self.master.config.title
         config['titleURL'] = self.master.config.titleURL
         config['multiMaster'] = self.master.config.multiMaster
+        if "wamp" in self.master.config.protocols:
+            config['wamp'] = self.master.config.protocols['wamp']
```
djmitche: Why does the web UI need to know if this maser supports slaves connecting via WAMP?

master/buildbot/www/service.py
```diff
@@ -144,19 +143,12 @@ def setupSite(self, new_config):
...
         root.putChild('sse', sse.EventResource(self.master))
 
         self.root = root
 
-        def either(a, b):  # a if a else b for py2.4
-            if a:
-                return a
-            else:
-                return b
+        either = lambda a, b: a if a else b
```
djmitche: I like this, but the new PEP8 forbids this form (E731)

master/buildbot/www/config.py
tardyp: Its to get the router url. Goal is to have the web ui is wamp as an optimised communication channel. REST + SSE does not scale well according to my tests. (Try using several tabs on nine.buildbot.net)

djmitche: Agreed re: scaling, but I think this confuses the issue -- `master.config.protocols` is about master/slave protocols.

tardyp: well, the pb port is also used for try, and pbchangesource :)

djmitche:
I like this quite a bit, but it does a *lot* of things at once:
 * basic WAMP support
 * WAMP support for MQ
 * WAMP support for master/slave proto (on both the master and the slave)

Since the Concept is Proven, I think the next step is to break those out into separate PRs that we can polish up before merging.

I do love the removal of direct support for WebSockets.  And I love that crossbar directly supports both pubsub and RPC (since we need RPC for a number of things).

I'm worried about the lack of persistent queues - that may lead to schedulers missing changes during restarts, which would be unfortunate.

Would it improve the implementation significantly to *require* crossbar, instead of allowing it as one implementation of MQ?  Perhaps only requiring it for multi-master configurations?  What I'm getting at is, adopting WAMP and crossbar wholesale may be a great way to take a lot of complexity out of Buildbot and put it in an externally maintained tool.  It may get us to Nine sooner.

master/buildbot/www/config.py
djmitche: True, but only as a default, and because it's a historical precedent.  I don't really like it.  Ideally this would be configured more like:

```
c['wamp'] = {
  'url': '...',
}

c['protocols'] = {
    'wamp': {},
    'pb': {'port': 9999},  # for slaves we haven't upgraded yet
}
```

I think this would be a bit clearer if the implementation is split into a few PRs as I suggested -- the first "add WAMP support" commit wouldn't touch `c['protocols']`, saving that for the slave PR.

sa2ajj:
Where are we with this PR?

djmitche:
@tardyp is there still value in leaving this open?  Should we move this to a Trac bug?

tardyp:
I rebased this branch as part of my update for wampmq PR.
There are still a bunch of work to do to finish this. not something that I can do in the next few month.

PR #1528, opened 2015-02-06
Change brid to buildrequestid.

ewongbb:
change brid to buildrequestid.


tardyp:
LGTM if this fixes stuff for you. Don't bother adding unit tests, this code is on death sentence.

PR #1531, opened 2015-02-08
dump a list of install python packages

sa2ajj:
This will help identify certain problems while using Travis (successful tests when they should have failed and failed things when they should have succeeded)


tardyp:
:+1:

PR #1532, opened 2015-02-08
fix api_proxy

tardyp:
plugins weren't properly configured


master/contrib/api_proxy.py
```diff
@@ -57,6 +57,7 @@ def reroute_cache(url):
 
 
 plugins = dict((k, {}) for k in master_service.apps.names if k != "base")
+print "found plugins:", plugins
```
sa2ajj: This is probably a forgotten debug print.

tardyp: no, this one is intended to be left here in order to tell people which plugin are installed

PR #1534, opened 2015-02-08
Fix bogus test for getChangesForBuild

djmitche:


master/buildbot/test/unit/test_db_changes.py
```diff
@@ -719,16 +719,15 @@ def test_getChangesForBuild(self):
                   "buildsetSourceStampid": 0,
                   "buildrequestid": 0,
                   "buildid": 0}
-        codebase_ss = {}
+
+        codebase_ss = {}  # shared state between addChange and addBuild
```
sa2ajj: side note: just excellent

PR #1535, opened 2015-02-09
collect all test related decorators in one module

sa2ajj:
It's a bit difficult to see what decorators are available since they are spread all over various `buildbot.test.unit.*` modules.


sa2ajj:
@djmitche and here we have a good example when we "test" against Twisted 11.x, while in reality the tests are performed against Twisted 15. :(

sa2ajj:
Ok, with GH-1536, `txgithub` is not a problem any more.

master/buildbot/test/util/decorators.py
```diff
@@ -20,6 +24,31 @@
...
+def flaky(bugNumber):
+    def wrap(fn):
+        if not os.environ.get(_FLAKY_ENV_VAR):
+            fn.skip = ("Flaky test (http://trac.buildbot.net/ticket/%d) "
+                       "- set $%s to run anyway" % (bugNumber, _FLAKY_ENV_VAR))
+        return fn
+    return wrap
+
+
 def usesFlushLoggedErrors(test):
```
djmitche: Maybe this is for another PR, but we don't support Twisted <= 9.0.0 anymore, so we can probably just get rid of this decorator and all of its uses -- at least on the master.

djmitche:
Looks good!

master/buildbot/test/util/decorators.py
sa2ajj: That's a very good point and I think that this PR is very suitable for this removal.

PR #1537, opened 2015-02-09
Add the StashStatusPush handler

sa2ajj:
* + pep8 is fixed at version 1.5.7


djmitche:
This will need to be forward-ported to nine

sa2ajj:
@djmitche that's on me.

PR #1539, opened 2015-02-11
skip test where sys.executable is used when sys.executable is empty

sa2ajj:


sa2ajj:
This is a totally wrong approach.

PR #1542, opened 2015-02-13
Changed check if patch info should be in blamelist

emoon:
When patch_info is being setup in build_request.py it's being set to (None, None) which makes this check incorrect but patch is being set to None. Now the check patch instead and that resolves the issue outlined in this ticket:

http://trac.buildbot.net/ticket/3194

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -140,7 +140,7 @@ class Build(properties.PropertiesMixin):
             if c.who not in blamelist:
                 blamelist.append(c.who)
         for source in self.sources:
-            if source.patch_info:  # Add patch author to blamelist
+            if source.patch:  # Add patch author to blamelist
                 blamelist.append(source.patch_info[0])
         blamelist.sort()
         return blamelist
--- a/master/buildbot/test/unit/test_process_build.py
+++ b/master/buildbot/test/unit/test_process_build.py
@@ -833,8 +833,8 @@ class TestBuildBlameList(unittest.TestCase):
         r.sources.extend([self.patchSource])
         build = Build([r])
         blamelist = build.blamelist()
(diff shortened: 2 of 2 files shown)


emoon:
So it seems that the unit-tests fails which this change:

The coded in build_request.py looks like this when setting up the patch data:

```
            if ssdata['patch']:
                patch = ssdata['patch']
                ss.patch = (patch['level'], patch['body'], patch['subdir'])
                ss.patch_info = (patch['author'], patch['comment'])
            else:
                ss.patch = None
                ss.patch_info = (None, None)
```

While the test sets it like this

```
        self.patchSource = FakeSource()
        self.patchSource.repository = "repoB"
        self.patchSource.codebase = "B"
        self.patchSource.changes = []
        self.patchSource.revision = "67890"
        self.patchSource.patch_info = ("jeff", "jeff's new feature")
```

In this case self.patch is default set to None but this isn't the case in the actual code that is being used.

djmitche:
That should be fixed in the tests, too.  Bonus points for a way to ensure the fake and reality match, too.

emoon:
Should I update the test(s) to make sure it works with my change also?

sa2ajj:
@emoon, yes, please.

emoon:
Alright. I will take a look at it!

emoon:
Sorry I totally forgot about this. I will have this sorted by tomorrow.

emoon:
I made a fix for the test. It would be good to add more elaborate testing in the future at least this PR fixes the issue that mails can't be sent to the user that caused the problem.

tardyp:
I found this in my MailNotifier branch, and fixed it the same.

tardyp:
let's merge it until I get review

emoon:
Thanks!

PR #1543, opened 2015-02-15
Cosmetic changes for buildbot/util/service.py

sa2ajj:
I'd like to ask questions and discuss various bits related to that.

We have a number of tickets like "convert this to a mere service", I'd like to better understand what we have, what we need, and how to proceed.

### DIFF
--- a/master/buildbot/util/service.py
+++ b/master/buildbot/util/service.py
@@ -195,22 +195,19 @@ class AsyncMultiService(AsyncService, service.MultiService):
 
     def startService(self):
         service.Service.startService(self)
-        l = []
-        for svc in self:
-            # handle any deferreds, passing up errors and success
-            l.append(defer.maybeDeferred(svc.startService))
-        return defer.gatherResults(l, consumeErrors=True)
+
+        return defer.gatherResults(
+            [defer.maybeDeferred(svc.startService) for svc in self],
+            consumeErrors=True)
 
     def stopService(self):
         service.Service.stopService(self)
-        l = []
-        services = list(self)
-        services.reverse()
-        for svc in services:
-            l.append(defer.maybeDeferred(svc.stopService))
+
         # unlike MultiService, consume errors in each individual deferred, and
         # pass the first error in a child service up to our caller
-        return defer.gatherResults(l, consumeErrors=True)
+        return defer.gatherResults(
+            [defer.maybeDeferred(svc.startService) for svc in reversed(list(self))],
+            consumeErrors=True)
 
     def addService(self, service):
         if service.name is not None:
@@ -235,11 +232,12 @@ class BuildbotService(AsyncMultiService, config.ConfiguredMixin,
     configured = False
 
     def __init__(self, *args, **kwargs):
-        name = kwargs.pop("name", None)
-        if name is not None:
-            self.name = name
-        if self.name is None:
+        name = kwargs.pop("name", self.name)
+        if name is None:
             raise ValueError("%s: must pass a name to constructor" % type(self))
+
+        self.name = name
+
         self.checkConfig(*args, **kwargs)
         self._config_args = args
         self._config_kwargs = kwargs
@@ -247,10 +245,12 @@ class BuildbotService(AsyncMultiService, config.ConfiguredMixin,
 
     def getConfigDict(self):
         _type = type(self)
-        return {'name': self.name,
-                'class': _type.__module__ + "." + _type.__name__,
-                'args': self._config_args,
-                'kwargs': self._config_kwargs}
+        return {
+            'name': self.name,
+            'class': '%s.%s' % (_type.__module__, _type.__name__),
+            'args': self._config_args,
+            'kwargs': self._config_kwargs
+        }
 
     def reconfigServiceWithBuildbotConfig(self, new_config):
         # get from the config object its sibling config


master/buildbot/util/service.py
```diff
@@ -186,6 +186,8 @@ class AsyncService(service.Service):
     def setServiceParent(self, parent):
         if self.parent is not None:
             yield self.disownServiceParent()
+        # XXX(sa2ajj): why this magic?  The twisted version of setServiceParent
+        # does not lookg this complicated (the line below)
         parent = service.IServiceCollection(parent, parent)
```
djmitche: Actually, it has [exactly that line](http://twistedmatrix.com/trac/browser/tags/releases/twisted-14.0.0/twisted/application/service.py#L184)

master/buildbot/util/service.py
```diff
@@ -195,23 +197,22 @@ class AsyncMultiService(AsyncService, service.MultiService):
...
+
         # unlike MultiService, consume errors in each individual deferred, and
         # pass the first error in a child service up to our caller
-        return defer.gatherResults(l, consumeErrors=True)
+        return defer.gatherResults(
+            [defer.maybeDeferred(svc.startService) for svc in reversed(list(self))],
+            consumeErrors=True)
 
+    # XXX(sa2ajj): This method seems to be a full duplicate of the one from
+    # service.MultiService w/ the only difference of returning a deferred.
```
djmitche: Correct -- but more importantly, it returns the deferred from `startService`, which the parent method discards.  So just wrapping the parent and returning `defer.succeed(None)` wouldn't be sufficient.

master/buildbot/util/service.py
```diff
@@ -236,10 +237,11 @@ class BuildbotService(AsyncMultiService, config.ConfiguredMixin,
...
     def __init__(self, *args, **kwargs):
         name = kwargs.pop("name", None)
-        if name is not None:
-            self.name = name
-        if self.name is None:
+        if name is None:
             raise ValueError("%s: must pass a name to constructor" % type(self))
+
+        self.name = name
+
```
djmitche: I think this change precludes subclasses setting `name` as a class variable.

master/buildbot/util/service.py
```diff
@@ -264,6 +268,7 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
                                     **config_sibling._config_kwargs)
 
     def setServiceParent(self, parent):
+        # XXX(sa2ajj): can we make 'master' a simple read-only property instead?
```
djmitche: That'd be great, although note that it's used *very* frequently so it should be quick.

master/buildbot/util/service.py
sa2ajj: Correct :(

master/buildbot/util/service.py
tardyp: This is the magic of zope.interface and the adapters.
It will verify that parent is a IServiceCollection or that it can be adapted to an IServiceCollection

sa2ajj: So many more things to learn...

master/buildbot/util/service.py
sa2ajj: I see.

master/buildbot/util/service.py
sa2ajj: Using `__slots__` is not an option?

master/buildbot/util/service.py
tardyp: This is the reason why I am not so found of zope.interface (probably also why it is not in the stdlib):
It feels too magical and confuse people.

If you are into learning things, a good exercice is to look at PropertyMixin, and try to understand how it really works.

master/buildbot/util/service.py
tardyp: I guess we should not try to optimize what has *not* shown to be a bottleneck. I'm not convinced a property will be so much a bottleneck.
I agree with sa2ajj that a property would be a good option for the problem of the master reference.

```
import time
class A(object):
   @property
   def master(self):
      return self.parent.parent

a = A()
b = A()
c = A()
c.parent = b
b.parent = a
t1 = time.time()
for i in xrange(10000000):
    d = c.master
t2 = time.time()
for i in xrange(10000000):
    d = c.parent
t3 = time.time()
print t2-t1, t3-t2
% python test.py
3.43862700462 1.08624792099
```
ok, this property is 3.4 times slower than direct access, but you can still make more than 2M per second...

djmitche: Maybe I misunderstand the proposal -- AIUI the use of the property is only to prevent incorrectly setting `self.master` -- but we've never had that problem.  I don't think it makes sense to try to dynamically navigate back to the master through parent relationships on every request -- those parent relationships are not fixed.

What *would* be cool is some kind of "magic" master reference that we could get from a mixin, which somehow knows the current master.  That needs to be magic rather than just a global variable because we should be able to support multiple masters running in the same process (for example, for a graceful restart).  In synchronous code, that's pretty straightforward - a thread-local variable - but I'm not sure how to do it with async code.

tardyp: Le Sun Feb 22 2015 at 2:56:34 AM, Dustin J. Mitchell <
notifications@github.com> a écrit :

> In master/buildbot/util/service.py
> <https://github.com/buildbot/buildbot/pull/1543#discussion_r25128140>:
>
> > @@ -264,6 +268,7 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
> >                                      **config_sibling._config_kwargs)
> >
> >      def setServiceParent(self, parent):
> > +        # XXX(sa2ajj): can we make 'master' a simple read-only property instead?
>
> Maybe I misunderstand the proposal -- AIUI the use of the property is only
> to prevent incorrectly setting self.master -- but we've never had that
> problem. I don't think it makes sense to try to dynamically navigate back
> to the master through parent relationships on every request -- those parent
> relationships are not fixed.
>
I think they are *per class* so that would be good enough.

> What *would* be cool is some kind of "magic" master reference that we
> could get from a mixin, which somehow knows the current master. That needs
> to be magic rather than just a global variable because we should be able to
> support multiple masters running in the same process (for example, for a
> graceful restart). In synchronous code, that's pretty straightforward - a
> thread-local variable - but I'm not sure how to do it with async code.
>
I'm not sure I understand your graceful restart usecase. What would be the
advantage over just reconfig? Having two masters on the same process will
give some other problems. They will share the same masterid, would run for
the same buildrequest?
What advantage would it be over just spawning another process?

I see another usecase for several masters in same reactor. It is for
multimaster integration tests. I'd like to start that way for my wamp test
strategy (multimaster, multislave, disconnection handling, etc).

djmitche: They are fixed in the current implementation, but I worry that we'll make change to the service hierarchy and go cross-eyed trying to change `self.parent.parent.parent.parent` to `self.parent.parent.parent` in the right subclasses.

You're right - there's not a good case for running multiple masters in the same process in production, but in testing it makes a lot of sense.

I guess I just don't see the issue -- we have `self.master` pretty much everywhere we need it now.  Why change?

sa2ajj:
I'll remove all XXXs (and leave the only cosmetic change :))

sa2ajj:
This PR does not add any value, so I'll close it.

PR #1544, opened 2015-02-16
Enhancement for master housekeeping

tardyp:
While working on slave scalability, I found out that configured slaves where definitely not managed properly.
It should be cleaned up upon master disappear, and reconfiguration
It makes sure master housekeeping is done at least once a day, and at reboot.
This enforce to run new house keeping stuff at upgrading

Im also improving the slave page to make sure it displays in finite time

### DIFF
--- a/master/docs/developer/database.rst
+++ b/master/docs/developer/database.rst
@@ -739,7 +739,7 @@ buildslaves
 
     .. py:method:: buildslaveConfigured(buildslaveid, masterid, builderids)
 
-        :param integer buildslaveid: the ID of the buildslave or None
+        :param integer buildslaveid: the ID of the buildslave
         :param integer masterid: the ID of the master to which it configured
         :param list of integer builderids: the ID of the builders to which it is configured
         :returns: Deferred
--- a/master/docs/developer/database.rst
+++ b/master/docs/developer/database.rst
@@ -739,12 +739,21 @@ buildslaves
 
     .. py:method:: buildslaveConfigured(buildslaveid, masterid, builderids)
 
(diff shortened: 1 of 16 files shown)


tardyp:
ok, this one deserves nodging. :)

sa2ajj:
LGTM

master/docs/developer/database.rst
```diff
@@ -739,13 +739,22 @@ buildslaves
 
     .. py:method:: buildslaveConfigured(buildslaveid, masterid, builderids)
 
-        :param integer buildslaveid: the ID of the buildslave
+        :param integer buildslaveid: the ID of the buildslave or None
```
djmitche: When is None OK?

djmitche:
The general idea here is that, on every reconfig, every buildslave is marked as no longer configured on the master, and then immediately marked as configured there again (modulo any actual changes to what is and is not configured there).

Is there a particular reason to do this, rather than calculating the differences and only un-configuring removed slaves?  If it's a matter of ease of implementation, then we could certainly get a bug on file to do a better job of this later.  I'm just curious if there's a deeper reason.

tardyp:
Indeed, that was just the easiest path. I realized later that Slave-Manager was doing much smarter thing, indeed.
I wont have time to look at this this week, so I'd prefer to let this like this for now.

master/docs/developer/database.rst
tardyp: a relique of old attempt. will remove.

djmitche:
Sounds good.  I filed bug 3209 for this.

djmitche:
Go ahead and merge once the relique is removed :)

PR #1545, opened 2015-02-16
Fix trac ticket 3207

vlovich:
The unicode output in a log is lost & is treated as an ASCII string when there are multiple chunks.
This can cause exceptions when mixing strings which will validate that the contents of an "ascii" string are ascii.

### DIFF
--- a/master/buildbot/process/logobserver.py
+++ b/master/buildbot/process/logobserver.py
@@ -113,16 +113,12 @@ class BufferLogObserver(LogObserver):
             self.stderr.append(data)
 
     def _get(self, chunks):
-        if chunks is None:
-            return [u'']
-        if len(chunks) > 1:
-            chunks = [''.join(chunks)]
-        elif not chunks:
-            chunks = [u'']
-        return chunks
+        if chunks is None or not chunks:
+            return u''
+        return u''.join(chunks)
 
     def getStdout(self):
-        return self._get(self.stdout)[0]
+        return self._get(self.stdout)
 
     def getStderr(self):
-        return self._get(self.stderr)[0]
+        return self._get(self.stderr)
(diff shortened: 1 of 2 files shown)


vlovich:
Not sure if this solves the issue.  Haven't had a chance to test it out.

tardyp:
I think you should first try to create a unit test that reproduces the issue.

Looking at the patch I think that data should be enforced in unicode in the higher layer code.

logobserver should never get anything else than unicode.

vlovich:
Then if it's always getting unicode, then it needs to join with u''

vlovich:
Updated review.  I would love to add tests but I've never done that before.  Any guidance would be appreciated.

tardyp:
run:
trial buildbot.test.unit.test_process_logobserver

then you can look at the test: test_process_logobserver.py
* Make sure the line you modify is executed.
* see how you con modify it to add one test for your issue

vlovich:
There wasn't an existing test so I added it.  Hopefully got it right.  Made sure that the test case failed before I made changes; simplified _get implementation.

djmitche:
To whoever merges this, note that it's against the release branch and not `eight`, so cherry-pick it :)

I'm worried that joining with `u''` will end up casting any logged *bytes* to unicode, with attendant exceptions.  I really don't know if logs are always unicode (they *are* always unicode in nine -- I don't know in eight).

vlovich:
Should I rebase the change onto 8 for easier review?

djmitche:
No, with just one or two commits cherry-picking isn't hard.  The review is fine, but the merge would pull undesirable changes into eight.

djmitche:
Actually, it looks like this is *only* used in SetPropertyFromCommand.  @vlovich can you verify that such a step will still work if the step outputs arbitrary bytes?

vlovich:
Is that necessary?  I'm willing to do it if necessary, but how this code works changes depending on how many bytes are sent.  Doesn't that seem like incorrect behaviour anyway?

djmitche:
True -- this only changes the behavior when there are multiple chunks, which is an unusual case for SetPropertyFromCommand, so it's unlikely to break anything.

djmitche:
merged in ebdc9faee4998498ccd38a30a0f5adb19b1187ee

PR #1551, opened 2015-02-19
Various cosmetics

sa2ajj:


master/buildbot/buildslave/libvirt.py
```diff
@@ -230,12 +230,12 @@ def _prepare_base_image(self):
 
         log.msg("Cloning base image: %s %s'" % (clone_cmd, clone_args))
 
+        d = utils.getProcessValue(clone_cmd, clone_args.split())
+
+        @d.addBoth
         def _log_result(res):
             log.msg("Cloning exit code was: %d" % res)
```
tardyp: This fails in case of errback. res would be a failure, not an integer

tardyp:
That is lot of work..
I would prefer most of this to be replaced by inlineCallbacks, but that is even more work

master/buildbot/buildslave/libvirt.py
sa2ajj: That was the original code, but now it's a good time to look at it.

master/buildbot/changes/bonsaipoller.py
```diff
@@ -63,6 +65,7 @@ def __cmp__(self, other):
                     or len(self.nodes[i].files) != len(other.nodes[i].files):
                 return -1
 
+                # XXX(sa2ajj): this code is DEAD DEAD DEAD (extra indentation?)
```
djmitche: And no tests.. I'd be OK with deleting BonsaiPoller -- I think it's unused.

sa2ajj: Yay! :)

master/buildbot/changes/gitpoller.py
```diff
@@ -314,6 +314,9 @@ def _convert_nonzero_to_failure(res,
                 raise EnvironmentError('command %s %s in %s on repourl %s failed with exit code %d: %s'
                                        % (command, args, path, self.repourl, code, stderr))
             return stdout.strip()
+        # XXX(sa2ajj): or should we use addCallback as a decorator and not pass
+        # `command`, `args`, and `path` but use them directly as we do in
+        # pretty much all other places?
```
djmitche: Yes, except _convert_nonzero_to_failure is used other places

sa2ajj: but not this particular instance: this is a local function.

master/buildbot/changes/bonsaipoller.py
sa2ajj: removed.

master/buildbot/schedulers/basic.py
```diff
@@ -47,6 +47,7 @@ class BaseBasicScheduler(base.BaseScheduler):
     class NotSet:
         pass
 
+    # XXX(sa2ajj): `properties` can be removed from this list of parameters (so they'd become part of kwargs)
```
djmitche: Agreed

master/buildbot/schedulers/dependent.py
```diff
@@ -26,6 +26,7 @@ class Dependent(base.BaseScheduler):
 
     compare_attrs = ('upstream_name',)
 
+    # XXX(sa2ajj): properties can be left alone in 'kwargs'
```
djmitche: :white_check_mark:

master/buildbot/schedulers/trysched.py
```diff
@@ -77,6 +77,7 @@ class Try_Jobdir(TryBase):
 
     compare_attrs = ('jobdir',)
 
+    # XXX(sa2ajj): any particular reason why Try_Jobdir has very specific parameters (unlike other schedulers?)
```
djmitche: Specifically, that it lacks `**kwargs`?  No, I don't think so.

sa2ajj: So I need to re-introduce them and pass to the parent, right?

master/buildbot/steps/subunit.py
```diff
@@ -32,6 +32,11 @@ class SubunitLogObserver(logobserver.LogLineObserver, TestResult):
     parser in the most direct fashion.
     """
 
+    # XXX(sa2ajj): what are these?  I could not find anything that does anything
+    # with these?
```
djmitche: This is the only place either identifier appears, so I think it's safe to kill both.

sa2ajj: Will do.

master/buildbot/test/unit/test_process_buildrequest.py
```diff
@@ -64,6 +64,7 @@ def test_collapseRequests_no_other_request(self):
         def collapseRequests_fn(builder, brdict1, brdict2):
             # Allow all requests
             self.fail("Should never be called")
+            # XXX(sa2ajj): Dead in the water?
```
djmitche: [The docs](https://docs.python.org/2/library/unittest.html#unittest.TestCase.fail) don't say that fail doesn't return..

master/buildbot/test/unit/test_process_builder.py
```diff
@@ -261,6 +261,7 @@ def canStartBuild(bldr, slave, breq):
         def canStartBuild_deferred(bldr, slave, breq):
             record.append((bldr, slave, breq))
             return (slave, breq) == ('slave', 100)
+            # XXX(sa2ajj): too many returns?
             return defer.succeed((slave, breq) == ('slave', 100))
```
djmitche: I suspect I (or whoever wrote this) was trying to test both deferred and non-deferred versions of `canStartBuild`.  It's probably fine to delete the first of the two `return` lines.

master/buildbot/test/unit/test_process_buildrequest.py
sa2ajj: True: it's safer to leave return here.

master/buildbot/www/websocket.py
```diff
@@ -169,6 +169,7 @@ def parse_hybi07_frames(buf):
         if header & 0x70:
             # At least one of the reserved flags is set. Pork chop sandwiches!
             raise WSException("Reserved flag in HyBi-07 frame (%d)" % header)
+            # XXX(sa2ajj): two lines below are DEAD DEAD DEAD
```
djmitche: Yes; but note that this code is from Twisted.  It may not be used at all, or it may be possible to use a committed version.

djmitche: Ah, it is used (`master/buildbot/www/ws.py`).

sa2ajj: But how it can be used?  I do not seem to know Python after all.

master/buildbot/schedulers/trysched.py
djmitche: Yes

master/buildbot/www/websocket.py
djmitche: You're correct that the code is dead -- I mentioned the origin of the code to indicate that, rather than improving it in place, we may be better off deleting the file and using the copy distributed with Twisted.  Sadly, [TW-4173](https://twistedmatrix.com/trac/ticket/4173), where the file originated, is still open.  So, perhaps best just left alone?

sa2ajj: Good to know we are not the only project that has such old tickets :)

master/buildbot/buildslave/libvirt.py
sa2ajj: fixed

sa2ajj:
@tardyp, @djmitche, @jaredgrubb unless you see problems with this PR, I'd like to have it merged ASAP: since it's biggish, there're good chances that some other PR would "interfere" :smile:

tardyp:
:+1:

sa2ajj:
Yay :)

jaredgrubb:
FWIW: LGTM!

sa2ajj:
@jaredgrubb it's always nice to hear that it looks good :)

PR #1552, opened 2015-02-20
Properties with value zero not rendered.

bekimb:
Since python evaluates 0 as False, properties with 0 are returned as
None.

This is a problem in some cases: For instance, setting properties via buildbot.steps.trigger.Trigger and its set_properties param, e.g. set_properties={"a_property": Property("buildnumber")}. For this first buildnumber 0, this will lead to a_property being None.

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -609,7 +609,7 @@ class Property(util.ComparableMixin):
 
             @d.addCallback
             def checkDefault(rv):
-                if rv:
+                if rv is not None:
                     return rv
                 else:
                     return props.render(self.default)
--- a/master/buildbot/test/unit/test_process_properties.py
+++ b/master/buildbot/test/unit/test_process_properties.py
@@ -1164,19 +1164,13 @@ class TestProperty(unittest.TestCase):
         return d
 
     def testIgnoreFalseValue(self):
-        self.props.setProperty("do-tests-string", "", "scheduler")
-        self.props.setProperty("do-tests-int", 0, "scheduler")
-        self.props.setProperty("do-tests-list", [], "scheduler")
         self.props.setProperty("do-tests-None", None, "scheduler")
 
-        value = [Property("do-tests-string", default="Hello!"),
-                 Property("do-tests-int", default="Hello!"),
-                 Property("do-tests-list", default="Hello!"),
-                 Property("do-tests-None", default="Hello!")]
+        value = [Property("do-tests-None", default="Hello!")]
 
         d = self.build.render(value)
         d.addCallback(self.failUnlessEqual,
-                      ["Hello!"] * 4)
+                      ["Hello!"])
         return d
 
     def testDefaultWhenFalse(self):


master/buildbot/process/properties.py
```diff
@@ -609,7 +609,7 @@ def getRenderingFor(self, props):
 
             @d.addCallback
             def checkDefault(rv):
-                if rv:
```
tardyp: I think this one would be a better logic:
```
if rv is not None:
```

bekimb:
Hmm, wouldn't that meant that empty string would return empty string instead of None?

tardyp:
yes, but empty string means default is not very pythonic

bekimb:
Hmm, I agree. I have updated the commit.

tardyp:
:+1:

sa2ajj:
Looks like a test needs attention:

```
[FAIL]
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/twisted/trial/unittest.py", line 271, in assertEqual
    % (msg, pformat(first), pformat(second)))
twisted.trial.unittest.FailTest: not equal:
a = ['', 0, [], 'Hello!']
b = ['Hello!', 'Hello!', 'Hello!', 'Hello!']


buildbot.test.unit.test_process_properties.TestProperty.testIgnoreFalseValue
```

tardyp:
So the original code comes from https://github.com/buildbot/buildbot/commit/eab918dfd2d7cfa812358b52252756665dc7d001 by @tomprince 

I think his change make sense and is documented (at the time of this commit)
Probably the default behaviour is confusing.

We might be willing to change defaultWhenFalse default to False

djmitche:
I suspect that `None` is almost NEVER the default value that a user wants.

I agree that changing the default is a decent option.

The other option is to make `defaultWhenFalse` false when neither it nor a default value is given, but make `defaultWhenFalse` true when a default value is given.

The second option is more complex, and I've often made the wrong call in cases like this -- sometimes it's better to have a simpler API that is easily understood, even if it's not as flexible as a more complex API (but where users have to keep checking the docs to see how it works).

tardyp:
Well, if no default value is specified, defaultWhenFalse has few purpose..
Am I missing something?

djmitche:
That's exactly my reasoning :)

djmitche:
This is related to http://trac.buildbot.net/ticket/2524 and http://trac.buildbot.net/ticket/3205.

To build upon my suggestion from a few days ago, I think `Property` should act like `dict.get` by default, which doesn't care about the truthiness of the value it looks up.

If prop 'foo' is False and prop 'nosuch' does not exist, then:

 * `Property('foo')` should evaluate to `False`
 * `Property('foo', 'bar')` should evaluate to `False`
 * `Property('foo', 'bar', defaulWhenFalse=False)` should evaluate to `False`
 * `Property('foo', 'bar', defaulWhenFalse=False)` should evaluate to `False`
 * `Property('foo', defaultWhenFalse=True)` should evaluate to `None`
 * `Property('foo', 'bar', defaultWhenFalse=True)` should evaluate to `"bar"`
 * `Property('nosuch')` should raise an exception
 * `Property('nosuch', 'bar')` should evaluate to `"bar"`
 * `Property('nosuch', 'bar', defaultWhenFalse=False)` should evaluate to `"bar"`
 * `Property('nosuch', 'bar', defaultWhenFalse=True)` should evaluate to `"bar"`

with documentation updated to correspond.

djmitche:
It looks like this should be reflected back into one of the bugs mentioned above.

djmitche:
moved back to the bug.

PR #1553, opened 2015-02-21
[eight] ShellMixin: workdir should be overridable

jaredgrubb:
I was playing with using ShellMixin and needed to be able to override the 'workdir' on makeRemoteShellCommand (I run multiple commands in various places).


djmitche:
Looks good.  Does this need a docs update?

sa2ajj:
Yes, it does need one.

jaredgrubb:
Added docs update.

PR #1555, opened 2015-02-22
[eight] ShellMixin: two minor logging fixes

jaredgrubb:
This has two fixes that I found useful:
* collectStdout/collectStderr args are never used. I think this is an oversight.
* passing in the same log name creates new logs each time (eg, a bunch of 'stdio' logs). I think it should reuse the same log if it already exists.

I havent done any unit tests for these. I'm not really sure if those parts are tested?

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -878,7 +878,11 @@ class ShellMixin(object):
                 kwargs['workdir'] = self.build.workdir
 
         # the rest of the args go to RemoteShellCommand
-        cmd = remotecommand.RemoteShellCommand(**kwargs)
+        cmd = remotecommand.RemoteShellCommand(
+            collectStdout=collectStdout,
+            collectStderr=collectStderr,
+            **kwargs
+        )
 
         # set up logging
         if stdio is not None:
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -844,7 +844,12 @@ class ShellMixin(object):
         kwargs.update(overrides)
         stdio = None
         if stdioLogName is not None:
-            stdio = yield self.addLog(stdioLogName)
+            # Reuse an existing log if possible; otherwise, create one.
+            try:
+                stdio = yield self.getLog(stdioLogName)
+            except KeyError:
+                stdio = yield self.addLog(stdioLogName)
+
         kwargs['command'] = flatten(kwargs['command'], (list, tuple))
 
         # check for the usePTY flag


master/buildbot/process/buildstep.py
```diff
@@ -878,7 +882,11 @@ def makeRemoteShellCommand(self, collectStdout=False, collectStderr=False,
                 kwargs['workdir'] = self.build.workdir
 
         # the rest of the args go to RemoteShellCommand
-        cmd = remotecommand.RemoteShellCommand(**kwargs)
+        cmd = remotecommand.RemoteShellCommand(
+            collectStdout=collectStdout,
```
tardyp: Right, but I'm not sure this can be used by the user, actually

master/buildbot/process/buildstep.py
```diff
@@ -844,7 +844,11 @@ def makeRemoteShellCommand(self, collectStdout=False, collectStderr=False,
         kwargs.update(overrides)
         stdio = None
         if stdioLogName is not None:
-            stdio = yield self.addLog(stdioLogName)
+            try:
```
sa2ajj: A comment that describes why the same operation is performed twice would be nice.

jaredgrubb:
Maybe I misunderstood the purpose of ShellMixin then? Basically I have a build step that runs a dozen commands, each one returning a JSON object, which I combine them all to set a property. So I do:

       cmd = yield self.makeRemoteShellCommand(
            collectStdout=True,
            stdioLogName=codebase,   # Runs two commands per sourcestamp 
            command=command,
            workdir=workdir
        )
        res = yield self.runCommand(cmd)
        stdout = res.stdout
        try:
            if not isinstance(stdout, unicode):
                stdout = res.stdout.decode('utf8')
            data = json.loads(stdout)
        except:
            twisted.python.log.err()
            data = {}
        defer.returnValue(data)

Without this patch, I get multiple log files with the same name, rather than one logfile with multiple invocations in it.

tardyp:
That looks cool. I'm not sure about returning json file via stdout.
In my ShellSequence, we usually communicate via json file as well, but via a file, and a transfer command. My approach is less efficient, but for me stdio is for logs (command progress status, errors, etc), not for computer output.
I think we should not be opinionated for that and let the user do what he prefers.

jaredgrubb:
I have a set of Python git tools we use to manage bugs in our project, and I added a "--json" option to them so I could build more tools, and that seemed easier than having the tools write things into a file.

But, regardless, I thought the purpose of ShellMixin was to make it easy for people to subclass Buildstep and run remote commands.

The 'collectStdout' and 'collectStderr' options are in the makeRemoteShellCommand API, but are not used anywhere in the function ... so I figured they were left out by mistake.

The other change, of reusing a log file if it already exists, is a bit gratuitous. It's not needed, but it looks odd to see a bunch of stdio files there with the same name.

tardyp:
:+1:

Le Sun Feb 22 2015 at 20:53:21, Jared Grubb <notifications@github.com> a
écrit :

djmitche:
Jared, you definitely have the right understanding, and re-using the existing logs is the intended behavior.

jaredgrubb:
Updated to add a short comment (as suggested by Mikhail).

PR #1556, opened 2015-02-22
Don't delete args while iterating

djmitche:
The unit tests don't operate in a configuration context (where
config.error returns), so it's hard to replicate the issue that way.
Fixes #3208.

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -1067,10 +1067,14 @@ class ShellMixin(object):
             else:
                 setattr(self, arg, constructorArgs[arg])
             del constructorArgs[arg]
+        badargs = set()
         for arg in constructorArgs.keys():
             if arg not in BuildStep.parms:
-                bad(arg)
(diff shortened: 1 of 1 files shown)


sa2ajj:
LGTM

master/buildbot/process/buildstep.py
```diff
@@ -1067,10 +1067,14 @@ def bad(arg):
             else:
                 setattr(self, arg, constructorArgs[arg])
             del constructorArgs[arg]
+        badargs = set()
         for arg in constructorArgs.keys():
```
tardyp: the fix is to use ```for arg in constructorArgs.keys()``` instead of ```for arg in constructorArgs:```

Difference is that keys returns a copy of the keys while constructorArgs return iterator

tardyp:
for me the bug has been fixed already in 508fe5c38e655dbf7a7f5c88d53fa4a330e65b40

djmitche:
indeed!

PR #1558, opened 2015-02-24
mention that Python with Py_DEBUG is not supported

djmitche:


master/docs/manual/installation.rst
```diff
@@ -29,6 +29,9 @@ Python: http://www.python.org
   Buildbot requires Python-2.6 or later on the master, although Python-2.7 is recommended.
   The slave runs on Python-2.5.
 
+  Note that this should be a "normal" build of Python.
```
sa2ajj: I'm for adding this warning, though I'm really curious why this non-exotic option makes such a difference :/

djmitche: It produces extra output on stdout, which confuses tests that examine that stdout.

PR #1559, opened 2015-02-24
fix status_gerrit

tardyp:
default buildsummary behaviour works

buildstarted, buildfinished behaviour does not because it needs more work in the status framework

Still needs a doc update.

We are changing the API here!
Supporting legacy status api if for me far too expensive, I think we should drop it, and document a data api based status handlers.

http://trac.buildbot.net/ticket/3185

### DIFF
--- a/master/buildbot/test/unit/test_status_gerrit.py
+++ b/master/buildbot/test/unit/test_status_gerrit.py
@@ -259,7 +259,7 @@ class TestGerritStatusPush(unittest.TestCase):
         build = {}
         build['builder'] = dict(name=u'dummyBuilder')
         build['results'] = buildResult
-        build['properties'] = {k: (v, 'test') for k, v in self.TEST_PROPS.items()}
+        build['properties'] = dict([(k, (v, 'test')) for k, v in self.TEST_PROPS.items()])
 
         # TODO: actually status api is never calling buildFinished()
         gsp.buildFinished(u'dummyBuilder', build, buildResult)
--- a/master/buildbot/test/unit/test_status_gerrit.py
+++ b/master/buildbot/test/unit/test_status_gerrit.py
@@ -21,7 +21,6 @@ from buildbot.status.status_gerrit import GerritStatusPush
 from buildbot.status.status_gerrit import makeReviewResult
 from buildbot.test.fake import fakedb
 from buildbot.test.fake import fakemaster
-from buildbot.test.fake.fakebuild import FakeBuildStatus
 from mock import Mock
 from twisted.internet import defer
 from twisted.trial import unittest
@@ -83,27 +82,10 @@ def legacyTestSummaryCB(buildInfoList, results, status, arg):
     return (str(buildInfoList), verified, 0)
 
 
-def _get_prepared_gsp(*args, **kwargs):
-    """
-    get an instance of GerritStatusPush prepared for testing
-
-    Hostname and username are "hardcoded", the rest is taken from the provided
-    parameters.
-    """
-    gsp = GerritStatusPush('host.example.com', 'username', *args, **kwargs)
-
-    gsp.master = fakemaster.make_master()
-    gsp.master_status = gsp.master.status
-
-    gsp.sendCodeReview = Mock()
-
-    return gsp
-
-
 class TestGerritStatusPush(unittest.TestCase):
 
-    TEST_PROJECT = 'testProject'
-    TEST_REVISION = 'd34db33fd43db33f'
+    TEST_PROJECT = u'testProject'
+    TEST_REVISION = u'd34db33fd43db33f'
     TEST_PROPS = {
         'gerrit_branch': 'refs/changes/34/1234/1',
         'project': TEST_PROJECT,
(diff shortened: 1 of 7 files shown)


sa2ajj:
I support dropping legacy status API.

If we had a good example of a status target using Data API, that'd help a lot.  I've been reading the source code and it looks a bit too complicated (probably because status API is implemented using Data API).

master/buildbot/test/unit/test_status_gerrit.py
```diff
@@ -111,37 +93,27 @@ class TestGerritStatusPush(unittest.TestCase):
     }
     THING_URL = 'http://thing.example.com'
 
-    def run_fake_summary_build(self, gsp, buildResults, finalResult,
-                               resultText, expWarning=False):
-        buildpairs = []
-        i = 0
-        for i in xrange(len(buildResults)):
-            buildResult = buildResults[i]
+    def _get_prepared_gsp(self, *args, **kwargs):
```
sa2ajj: I moved it out of the class since it does not use `self`.  And I oppose this change since it still does not make any use of `self`.

sa2ajj: Oh, `self` was very well hidden at the end of the line...  Sorry.

tardyp:
I was talking about the GerritStatusPush API. This class takes callbacks for customization of the messages. Those callbacks were passed BuildStatus objects. Now they are passed data api build dictionary.
Then there is the internal status api.

I changed BuildSetSummaryNotifierMixin, so that it passes data api dictionaries instead of BuildStatus object. This brakes MailNotifier unit tests which are already brokingly passing (they are faking disabled api)

Looking at the code, there are a lot of stuff to cleanup, and probably the whole internal working to rethink :-/

tardyp:
Adding need work. There are two approaches here.

1. Start from the beginning, and make a solid new status internal api, and then rewrite the status plugins correctly

2. Hack the basic functionality to make it report something. This is the what I did here to make my buildbot_travis service work with Gerrit. This should be feasible for mail status as well, and will allow us to ship a beta version

djmitche:
Wait, we have a good solid status API -- that's what the Data API is!

tardyp:
Isn't the data api a bit too low level?

I meant a factorized api for status reporting plugin. like BuildSetSummaryNotifierMixin.
a base class with methods like getBuilderStatus(build) (who will tell if the build is fixed, breaking, broken, etc)
All status reporter more or less needs the same stuff, but send the same info over different wire.
For now, I think they are all reinventing the same wheel in a differet manner (except for BuildSetSummaryNotifierMixin)

djmitche:
I don't want to get back to having a half-baked Python API -- particularly not one involving inheritance.  Eliminating that API has been a huge part of the motivation for the last five years of work.

Where there's common code between the status plugins, we can introduce some utilities -- explicitly private at first, and perhaps eventually public -- that wrap the data API.

tardyp:
If you don't want it half backed, we can then cook it correctly.

Now that we have much more experience of what a status plugins need, I think we can make a proper api that makes it simple to create new status api.

I think that would be an mistake to stick to very generic API because supporting legacy api has been such a trauma.

For the moment, our api is too complicated to use. @sa2ajj is the first to complain there is no good example of status plugin so he cannot help on those bugs. For both of us, it is easy because we know the data/mq api, but for the newcomer, I think it is too low level and needs too much data api call to get anything you need.

Then I dont know exactly what you have against inheritance. Its a simple paradygm that people are used to. Its easy to understand that people just have to subclass this base class, then override onBuilderResults() or onBuildsetResults() to push the status to their needs.

I thought about this since I wrote this PR, I have some ideas of a simple API that will make it simple and efficient to write status plugins.

djmitche:
Ideally any of these status plugins could be written in Go or JS or anything like that.  Introducing a Python API gives in-process plugins a distinct advantage over external plugins.

A few of the mistakes we've made in Python API design:
 * unclear documentation of what methods are private and what are part of the API
 * failure to return and handle deferreds
 * instantiating objects in the configuration makes it difficult to handle reconfigs
 * using positional arguments instead of keyword arguments
 * unclear what "drives" the operation -- for example, BuildStep subclasses can either define a bunch of methods to be called during step operation, *or* define a method to run the step.

I'm interested to see what your ideas are.  I think we could meet in the middle with a collection of mixins that factor out commonly-used functionality (similar, as you said, to BuildSetSummaryNotifierMixin).  Then the status plugin is still fundamentally built on the data API, we've just factored out some repeated code.  Users of those mixins could overcome any non-compatible changes between Buildbot versions by just copying the old mixins they were using.

tardyp:
I'm not sure I see any advantages in writing status plugins in Go or JS.

For me, a good api should make simple things trivial, and complex things possible.

Trivial would means callback to implement with all data populated
Possible would mean use the data api to get the data that is not pre populated.
See #1567 for my proposal, lets continue the discussion there.

Then for the list of mistakes history, I think they are very valuable, thank you for listing them!

sa2ajj:
Should we close this PR then?

PR #1560, opened 2015-02-25
builderpage: support tags

tardyp:
Supports tags and tag filtering in builder page

image


djmitche:
LGTM.  I still can't get gulp to build www so I can't test this out, but Travis succeeded (despite what it's reported here), so I'll merge it.

PR #1561, opened 2015-02-25
Rescale waterfall with ctrl (meta) + "-" & "+"

tothandras:
This is just a quick implementation. Usable, but not perfect.
I have problems with:
time = y.invert(containerParent.scrollTop)
and after rescale
y.invert(y(time))
should be the same Date, but it's not. :/


tardyp:
Why requiring ctrl key? just + - looks cool to me.
That would be great if it integrates with the new shiny bbSettingsProvider.

tothandras:
:+1: okay, I will change that and integrate it with the bbSettingsProvider

PR #1563, opened 2015-02-26
fix logic for performing login for CVS build step

sa2ajj:


holsta:
For what it's worth: I tested this against a local OpenBSD CVS repo and with login=None set in the factory, I can now do a full checkout and build.

tardyp:
Step 6 of debugging: How did that ever work?

tardyp:
:+1:

sa2ajj:
@tardyp I think this problem was introduced while re-write...

sa2ajj:
Tests fail, so it's time to fix them.

sa2ajj:
Right, after some investigation I found that:

* functionality was lost -- `login` parameter is expected to have password for login, it's never passed to the remote side
* the tests were "fixed" to work by setting `login` to `True`

I'll need to:
- [ ] re-implement passing of password
- [x] fix test cases (actually, almost done -- I need to check that password is really passed over to the slave side)

sa2ajj:
ok, it looks like I implemented the logging in again, fixed test cases and update the docs.

PR #1564, opened 2015-02-27
fixes for reverse proxies

tardyp:
When buildbot sits behind reverse proxies:

- fonts where fetched using /fonts instead of /proxypath/fonts
- selforigin was miscaculated. /proxypath/ should be dropped from buildbotURL=http://server/proxypath -> origin is http://server

### DIFF
--- a/master/buildbot/test/unit/test_www_rest.py
+++ b/master/buildbot/test/unit/test_www_rest.py
@@ -83,19 +83,19 @@ class V2RootResource(www.WwwTestMixin, unittest.TestCase):
     def test_default_origin(self):
         self.master.config.buildbotURL = 'http://server/path/'
         self.rsrc.reconfigResource(self.master.config)
-        self.assertEqual([r.pattern for r in self.rsrc.origins], ['http\\:\\/\\/server\\Z(?ms)'])
+        self.assertEqual([r.pattern for r in self.rsrc.origins], [r'http\:\/\/server\Z(?ms)'])
 
         self.master.config.buildbotURL = 'http://server/'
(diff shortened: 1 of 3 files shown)


master/buildbot/test/unit/test_www_rest.py
```diff
@@ -80,6 +80,23 @@ def test_invalid_http_method(self):
...
+        self.rsrc.reconfigResource(self.master.config)
+        self.assertEqual([r.pattern for r in self.rsrc.origins], ['http\\:\\/\\/server\\Z(?ms)'])
+
+        self.master.config.buildbotURL = 'http://server:8080/'
+        self.rsrc.reconfigResource(self.master.config)
+        self.assertEqual([r.pattern for r in self.rsrc.origins], ['http\\:\\/\\/server\\:8080\\Z(?ms)'])
+
+        self.master.config.buildbotURL = 'https://server:8080/'
+        self.rsrc.reconfigResource(self.master.config)
+        self.assertEqual([r.pattern for r in self.rsrc.origins], ['https\\:\\/\\/server\\:8080\\Z(?ms)'])
```
rutsky: I'm completely out of context, but why not to use Python's raw strings?

E.g.:
```python
r'https\:\/\/server\:8080\Z(?ms)'
```

tardyp: I usually just copy paste what assertEqual complains the value should be (after visually checking them)

tardyp:
addressed @rutsky comments

djmitche:
Thanks for the review @rutsky!

PR #1567, opened 2015-03-01
reporter API + MailNotifier

tardyp:
as discussed in #1559 this is my proposal for a new 'reporters' api.

I think it is better to just drop 'status' word to avoid confusion, and speak about reporting, as this is now what this is about. There is no more status api where steps will push some extracted data in status objects.

I implemented MailNotifier with this new API. All the tests have been ported + new integration test have been written.

    Most of the features are there except:
    - removed the users db support
    - removed deprecated stuff: categories, customMesgData stuff, non utf8 patches, python2.4

    - changed the messageFormatter to be outside of module, and to use jinja templating


master/buildbot/reporter/interfaces.py
```diff
@@ -0,0 +1,45 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+"""Interface documentation.
+
+Define the interfaces that are implemented reporter classes.
```
sa2ajj: Do we need to put these interfaces into a separate module instead of `buildbot.interfaces`?

sa2ajj:
This looks "biggish".  I really need to print it and contemplate on it not in vicinity of a computer.

master/buildbot/reporter/interfaces.py
tardyp: I put it here for clarity of review

sa2ajj: May I kindly suggest that you'd put it in the target location?  For the clarify of review?

tardyp:
moved interfaces to interfaces.py

This file is a bit of a mess imho, there is a lot of doc that should be in developer documentation, and I'm not really sure that all those interfaces are really used.

I created a deprecated space for the status api interfaces.

tardyp:
about biggish, this is what I meant with "data api is complicated to use". You have to know a lot of the data model to do as simple stuff as "getting all  build, properties and steps related to a buildset"

 I prefer to have such code factorized, and provided as a simplified api.

master/buildbot/interfaces.py
```diff
@@ -129,6 +129,236 @@ def getAddress(user):
         be reached by email), or a Deferred which will fire with the same."""
 
 
+class ITestResult(Interface):
```
djmitche: I think we should kill this interface and remove its use from subunit.  We're not storing test results in the data API (and I don't think anything actually uses them).  From what I can tell this commit is just rearranging `interfaces.py`, in which case it's fine to just leave it as-is.

master/buildbot/reporter/manager.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __init__(self):
+        self.buildsetReporters = []
+        self.buildersReporters = []
+        self._buildsetCompleteConsumer = None
+
+    @defer.inlineCallbacks
+    def reconfigServiceWithBuildbotConfig(self, new_config):
+
+        yield service.ReconfigurableServiceMixin.reconfigServiceWithBuildbotConfig(self,
```
djmitche: I think you mean to call the parent class's method here..

master/buildbot/util/service.py
```diff
@@ -272,3 +273,71 @@ def checkConfig(self, *args, **kwargs):
 
     def reconfigService(self, *args, **kwargs):
         return defer.succeed(None)
+
+
+class BuildbotServiceManager(AsyncMultiService, config.ConfiguredMixin,
+                             ReconfigurableServiceMixin, util.ComparableMixin):
```
djmitche: I like this idea -- can you fish it out into a PR that makes all of the existing service managers also use it?

djmitche:
Thanks for this -- having code to stare at makes the conversation a lot more concrete.

I have two general concerns:
 * I don't like the idea of adding support specifically for a set of "reporters" in the configuration.  Aside from the built-in options (MailNotifier, IRC), I suspect that a lot of users' plugins will not be reporters, per se -- for example, someone will want a consumer that automatically retries failing builds or archives build artifacts.  I'd rather see generic "attach your Twisted service (or reconfigurable service) here" functionality into which the MailNotifier, IRC, etc. fit nicely.
 * The inversion-of-control pattern is strong here: Buildbot calls specific methods on reporters, with specific arguments that have specific meanings.  Since (per point a) this is the only way to install a service that consumes from the data API, *every* reporter plugin will implement these methods, and changing them later will be impossible, fixing the API forevermore.  The alternative is a library-style approach.  For example, define a `BuildsetListener` class which users can instantiate, which gathers all of the data about each completed buildset together in a specific (hopefully extensible) format, then invokes a callback function.  Users/authors who need more flexibility than that provided by the BuildsetListener are welcome to look at its implementation and adapt it, gathering whatever additional data they need directly from the data API.  And if we change the data API in such a way that BuildsetLister's data structures can't be extended to match (e.g., something similar to the introduction of codebases), then we can hack things as best we can (e.g., only report on one codebase) and document the limitation, without actually *preventing* anyone from getting done what they need to do.  The API would be a utility, but once it ceases to be useful, the natural next step would be for users to stop using it.

Hopefully that makes sense?

master/buildbot/interfaces.py
tardyp: Having test case oriented stats is a useful thing.
But like you said on ml there are too much unfinished stuff in buildbot.

djmitche: Yeah, I'm fairly certain this is 100% non-functional, too.

tardyp:
As per remark by @djmitche, I moved the build detail fetch routine to an utility method

Got a long flight today, hopefully I can make some progress here.

tardyp:
Up. Updated the PR with fully functional MailNotifier.

See updated PR Message and title for details

master/buildbot/reporters/utils.py
```diff
@@ -0,0 +1,156 @@
...
+    blamelist = list(blamelist)
+    blamelist.sort()
+    defer.returnValue(blamelist)
+
+
+def getURLForBuild(master, builderid, build_number):
+    prefix = master.config.buildbotURL
+    return prefix + "#builders/%d/builds/%d" % (
+        builderid,
+        build_number)
```
jaredgrubb: Is this whole file just a rename, or are there changes inside it?

tardyp: No this one is entirely new code.

mail.py is arguably a rename, but has been heavily refactored

master/docs/manual/cfg-statustargets.rst
```diff
@@ -1,55 +1,49 @@
...
     :local:
 
 The Buildmaster has a variety of ways to present build status to various users.
-Each such delivery method is a `Status Target` object in the configuration's :bb:cfg:`status` list.
-To add status targets, you just append more objects to this list::
+Each such delivery method is a `Reporter Target` object in the configuration's ``services`` list.
+To add reporter targets, you just append more objects to this list::
 
-    c['status'] = []
+    c['services'] = []
```
djmitche: Hmm, we already have support for this key via #1378, so this section should reference that documentation.  Maybe the wording's just a bit confusing here.
```
Each is a Buildbot service (see :bb:cfg:`services`) which communicates with the rest of Buildbot to learn about build status.
```

master/docs/manual/cfg-statustargets.rst
```diff
@@ -397,75 +388,28 @@ Builder result as a string
         # one of 'success', 'warnings', 'failure', 'skipped', or 'exception'
 
 URL to build page
-    ``master_status.getURLForThing(build)``
+    ``reporters.utils.getURLForBuild(master, build['buildid'])``
```
djmitche: is this `self.reporters` above it's `self.master`..

djmitche:
:thumbsup: in general.  I like this.  Apologies for not replying for so long, and thanks for the ping.

master/buildbot/process/build.py
```diff
@@ -135,6 +135,7 @@ def __repr__(self):
         return "<Build %s>" % (self.builder.name,)
 
     def blamelist(self):
+        # FIXME: kill this. This belongs to reporter.utils
```
yetanotherion: erase might be better suited than kill :)

master/buildbot/reporters/message.py
```diff
@@ -0,0 +1,132 @@
...
+from buildbot.reporters import utils
+from buildbot.status.results import CANCELLED
+from buildbot.status.results import EXCEPTION
+from buildbot.status.results import FAILURE
+from buildbot.status.results import SUCCESS
+from buildbot.status.results import WARNINGS
+from buildbot.status.results import statusToString
+
+
+class MessageFormatter():
```
yetanotherion: I'd prefer class MessageFormatter(object)

master/buildbot/reporters/utils.py
```diff
@@ -0,0 +1,156 @@
...
+            [master.data.get(("builds", build['buildid'], 'steps'))
+             for build in builds])
+        if wantLogs:
+            for s in flatten(buildsteps, types=(list, UserList)):
+                s['logs'] = yield master.data.get(("steps", s['stepid'], 'logs'))
+                for l in s['logs']:
+                    l['content'] = yield master.data.get(("logs", l['logid'], 'contents'))
+
+    else:
+        buildsteps = range(len(builds))
```
yetanotherion: len(builds) ?
I don't see how buildsteps and builds could be related like that

yetanotherion:
I have a doubt regarding 
buildsteps = range(len(builds))

master/buildbot/reporters/message.py
tardyp: ok

master/buildbot/reporters/utils.py
tardyp: If we dont want the step, we still need a list of size len(build) in order to simplify the zip later

tardyp:
addressed @yetanotherion comments

yetanotherion:
:+1:

sa2ajj:
Needs a rebase...

tardyp:
rebased. I would really appreciate we go faster on this one.

sa2ajj:
Merge commits are no-no!

tardyp:
Rebasing 2k lines is no-no. This stuff has been ready for 3 weeks!

master/buildbot/reporters/mail.py
```diff
@@ -0,0 +1,506 @@
...
+
+        sender_factory = ESMTPSenderFactory(
+            self.smtpUser, self.smtpPassword,
+            self.fromaddr, recipients, StringIO(s),
+            result, requireTransportSecurity=self.useTls,
+            requireAuthentication=useAuth)
+
+        reactor.connectTCP(self.relayhost, self.smtpPort, sender_factory)
+
+        return result
```
jaredgrubb: Is this result supposed to be connected to the SMTP somehow?

master/buildbot/reporters/message.py
```diff
@@ -0,0 +1,132 @@
+import jinja2
```
jaredgrubb: I'm very glad this is based on jinja :)

master/docs/manual/cfg-statustargets.rst
```diff
@@ -1,55 +1,49 @@
...
-A full list of status targets is available in the :bb:index:`status`.
+The remainder of this section describes each built-in reporters.
+A full list of reporters is available in the :bb:index:`reporter`.
 
-.. bb:status:: MailNotifier
+.. bb:reporter:: MailNotifier
 
 .. index:: single: email; MailNotifier
 
 MailNotifier
```
jaredgrubb: Should it be "MailReporter" now? I know that would be a more disruptive change.

PR #1568, opened 2015-03-01
fix login functionality for master side CVS source checkout step

sa2ajj:
* perform login only when password is provided
* do actually use the password
* fix the test cases (for vast majority remove use of login parameter)
* update docs to document the `login` parameter


sa2ajj:
Port of #1563

PR #1569, opened 2015-03-01
Don't assert on step status

djmitche:
Step status is debounced, so it's not a great attribute to use for
tests.  Just use a global variable instead.

Fixes #3125.

### DIFF
--- a/master/buildbot/test/integration/test_customservices.py
+++ b/master/buildbot/test/integration/test_customservices.py
@@ -21,15 +21,18 @@ from twisted.internet import defer
 # The custom step is using a CustomService, in order to calculate its result
 # we make sure that we can reconfigure the master while build is running
 
+observed_num_reconfig = None
 
 class CustomServiceMaster(RunMasterBase):
 
     @defer.inlineCallbacks
     def test_customService(self):
+        global observed_num_reconfig
 
-        build = yield self.doForceBuild(wantSteps=True)
+        observed_num_reconfig = None
+        yield self.doForceBuild(wantSteps=True)
 
-        self.assertEqual(build['steps'][0]['state_string'], 'num reconfig: 1')
+        self.assertEqual(observed_num_reconfig, 1)
 
         myService = self.master.namedServices['myService']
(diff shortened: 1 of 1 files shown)


tardyp:
Hum that smells bad, there are other tests that does that kind of checks, and there is nothing that prevents some other guy to do similar tests. Is this debouncing really necessary?
Shouldn't getResultSumary *always* be the last word for step status?

djmitche:
It's debounced so that steps can update their status as often as they (e.g., on every test failure) like without generating many events per second.  I think that's a good design, but using step status to tunnel information out of a step to a wrapping integration test is *not* a good design.

If there are other places we're doing the same thing, point me to them and I'll fix them.

tardyp:
I may agree that my test is not super clean, but we have a real bug here with the debouncing.

Don't you agree that last update status should be the last status reported by API?
We should debounce messages probably, but flush in the end of the step.

djmitche:
We do flush at the end of the step.  I don't think there's an actual production bug here -- just that the integration test is asserting on something that it hasn't correctly waited for.  It should wait for that last update to complete a fraction of a second later, except we don't want tests waiting for timers to expire.

So I really don't see an issue here..

tardyp:
I can see this code in the end of BuildStep.startstep()
```
        # update the summary one last time, make sure that completes,
        # and then don't update it any more.
        self.realUpdateSummary()
        yield self.realUpdateSummary.stop()
```
I would assume that realUpdateSummary is synchronously stop is the end of the step. This integration waits for the end of the build. So the race really shouldn't happen.
What I am missing here?

djmitche:
If I could reproduce the bug, I'd be able to give a definite answer.  But I am more suspicious of a problem with the integration test framework than with the the `stop` call.

tardyp:
I run the test 30k times this night, and I am unable to reproduce :-(

djmitche:
OK, well, let's call off this chase, then.  If you can reproduce the step behaviors you've seen, let me know!

PR #1573, opened 2015-03-03
Make source checkout steps not rely on hard-coded list of modes

seankelly:
Updated all of the source checkout steps that have a `mode` option to query their methods to determine the available modes. I dubbed this an "attribute group". This pull request only touches the `mode` parameter; the `method` parameter is another that could benefit from this.

Note that I skipped fully converting the p4 step due to a different check of `self.mode`.

This pull request supersedes #1458.

### DIFF
--- a/master/buildbot/steps/source/base.py
+++ b/master/buildbot/steps/source/base.py
@@ -161,7 +161,7 @@ class Source(LoggingBuildStep, CompositeStepMixin):
         Returns a list of all members in the attribute group.
         """
         from inspect import getmembers, ismethod
-        methods = getmembers(self, lambda m: ismethod)
+        methods = getmembers(self, ismethod)
         group_prefix = attrGroup + '_'
         group_len = len(group_prefix)
         group_members = [method[0][group_len:]
--- a/master/buildbot/steps/source/base.py
+++ b/master/buildbot/steps/source/base.py
@@ -140,6 +140,35 @@ class Source(LoggingBuildStep, CompositeStepMixin):
         if isinstance(self.descriptionSuffix, str):
             self.descriptionSuffix = [self.descriptionSuffix]
 
+    def _hasAttrGroupMember(self, attrGroup, attr):
+        """
+        The hasattr equivalent for attribute groups: returns whether the given
+        member is in the attribute group.
+        """
+        method_name = '%s_%s' % (attrGroup, attr)
+        return hasattr(self, method_name)
+
+    def _getAttrGroupMember(self, attrGroup, attr):
+        """
+        The getattr equivalent for attribute groups: gets and returns the
+        attribute group member.
+        """
+        method_name = '%s_%s' % (attrGroup, attr)
+        return getattr(self, method_name)
+
+    def _listAttrGroupMembers(self, attrGroup):
+        """
+        Returns a list of all members in the attribute group.
+        """
+        from inspect import getmembers, ismethod
+        methods = getmembers(self, lambda m: ismethod)
+        group_prefix = attrGroup + '_'
+        group_len = len(group_prefix)
+        group_members = [method[0][group_len:]
+                         for method in methods
+                         if method[0].startswith(group_prefix)]
+        return group_members
+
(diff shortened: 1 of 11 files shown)


seankelly:
Of course I skipped running *all* of the tests and see some integration tests failed. It appears the test is wrong because it's using `mode="copy"` but `copy` is defined for `method`, not `mode`.

master/buildbot/steps/source/base.py
```diff
@@ -140,6 +140,35 @@ def __init__(self, workdir=None, mode='update', alwaysUseLatest=False,
...
+        """
+        method_name = '%s_%s' % (attrGroup, attr)
+        return getattr(self, method_name)
+
+    def _listAttrGroupMembers(self, attrGroup):
+        """
+        Returns a list of all members in the attribute group.
+        """
+        from inspect import getmembers, ismethod
+        methods = getmembers(self, lambda m: ismethod)
```
sa2ajj: Hmm? Parameter `m` is not used?

master/buildbot/steps/source/bzr.py
```diff
@@ -50,7 +50,9 @@ def __init__(self, repourl=None, baseURL=None, mode='incremental',
         if baseURL is not None and defaultBranch is None:
             raise ValueError("you must provide defaultBranch with baseURL")
 
-        assert self.mode in ['incremental', 'full']
+        if not self._hasAttrGroupMember('mode', self.mode):
```
sa2ajj: What do you think about moving this check to base class?

master/buildbot/steps/source/base.py
seankelly: Oops, yeah.

master/buildbot/steps/source/bzr.py
seankelly: I don't have a problem with it but have a few concerns about this.

1. This would need to be a method that gets called. The p4 step has an extra check on the if and the repo step is completely different.
2. It looks like most of the steps are slightly different in their implementations, especially error reporting. Moving this check to the base class would probably entail updating all of the source steps to be homogeneous. And I have no idea what the recommended way is
3. The same unification could apply to the `method` parameter too. My thought was make `mode` consistent, then `method`, then error reporting.

Expanding on point 2, I see three ways right now: raising `ValueError`, `ConfigError(errors)`, and `buildbot.config.error(error)`. There's also a few `assert`s. Any recommendations? (Bring in @djmitche too).

djmitche: It sounds like this is a nice place to break off this PR and start on a new one.  I don't have any problem with work to make the source steps more consistent!

sa2ajj:
Re `method`: what do you think about it being a parameter for the mode handling function?

seankelly:
@sa2ajj Not sure I understand your question. Which function should have a parameter?

sa2ajj:
currently you have `mode_<mode>(self, _)` as the handler signature.  I wonder what you think about this kind of signature `mode_<mode>(self, _, <method>)`

sa2ajj:
As [docs](http://docs.buildbot.net/current/manual/cfg-buildsteps.html#id3) states:

```
mode method

    These two parameters specify the means by which the source is checked out. mode specifies the type of checkout and method tells about the way to implement it.
```

seankelly:
My primary objection is the inability to verify `method`. The two modes accept different methods; that is only partially checked right now. I would like any future changes to `method` to involve more checking rather than less.

sa2ajj:
It was just a question.

Anyway, what else needs to be done in this PR so it can land?

seankelly:
I don't know. You are the one that added the "needs work" label :).

sa2ajj:
My label was about use of `ismethod`, but, I believe, I just missed your commit.

I'll have a look now.

sa2ajj:
Ah, wait, we've been discussing moving that functionality to the base class.  Do you think it's for the next PR?

seankelly:
I think so. The next PR can figure out what to do about `method` at the same time.

PR #1570, opened 2015-03-02
Fix GerritStatusPush 'startCB' not work

stanzgy:
Fix GerritStatusPush 'startCB' not work with Gerrit+Git.
GerritStatusPush 'startCB' triggers before buildsteps start. But the condition
'startCB' decide to send code reviews is the property 'gerrit_branch' which is
set during the buildbot.steps.source.gerrit.Gerrit buildstep. So use property
'event.change.id' instead of 'gerrit_branch' to verify Gerrit source to ensure
'startCB' works properly.

The property 'got_revision' here also has same problem when 'startCB' triggers.

PR #1575, opened 2015-03-03
Fix GerritStatusPush 'startCB' not work

stanzgy:
Fix GerritStatusPush 'startCB' not work with Gerrit+Git.
GerritStatusPush 'startCB' triggers before buildsteps start. But the condition
'startCB' decide to send code reviews is the property 'gerrit_branch' which is
set during the buildbot.steps.source.gerrit.Gerrit buildstep. So use property
'event.change.id' instead of 'gerrit_branch' to verify Gerrit source to ensure
'startCB' works properly.

The property 'got_revision' here also has same problem when 'startCB' triggers.

original proposed in #1570

### DIFF
--- a/master/buildbot/status/status_gerrit.py
+++ b/master/buildbot/status/status_gerrit.py
@@ -323,9 +323,9 @@ class GerritStatusPush(StatusReceiverMultiService, buildset.BuildSetSummaryNotif
             return
 
         # Gerrit + Git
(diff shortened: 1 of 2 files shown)


sa2ajj:
LGTM

@djmitche can you have a quick look?

djmitche:
I don't know the Gerrit-related code very well, but I don't see anything here that looks amiss.

sa2ajj:
Thank you, @stanzgy

PR #1576, opened 2015-03-04
Fix logic in buildbot.status.builder:BuilderStatus.matchesAnyTag

adiroiban:
Problem
======

I think that current logic for matchesAnyTag is wrong.

Changes
=======

I wrote a few tests and update the code.

How to test
=========

Create builders with tags. Go in web status and click on the tags. You should see only builders with have one of the selected tags.

Let me know if something needs to be changes.

Thanks!

### DIFF
--- a/master/buildbot/status/builder.py
+++ b/master/buildbot/status/builder.py
@@ -337,7 +337,7 @@ class BuilderStatus(styles.Versioned):
 
     def matchesAnyTag(self, tags):
         # Need to guard against None with the "or []".
-        return set(self.tags or []).isdisjoint(tags)
+        return not set(self.tags or []).isdisjoint(tags)
 
     def matchesAllTags(self, tags):
         # Need to guard against None with the "or []".
--- a/master/buildbot/status/builder.py
+++ b/master/buildbot/status/builder.py
@@ -337,7 +337,7 @@ class BuilderStatus(styles.Versioned):
 
     def matchesAnyTag(self, tags):
         # Need to guard against None with the "or []".
-        return not set(self.tags or []).isdisjoint(tags)
+        return bool(set(self.tags or []) & set(tags))
 
     def matchesAllTags(self, tags):
         # Need to guard against None with the "or []".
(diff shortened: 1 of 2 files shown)


adiroiban:
I will look at fixing this for py2.6

master/buildbot/status/builder.py
```diff
@@ -337,7 +337,7 @@ def setTags(self, tags):
 
     def matchesAnyTag(self, tags):
         # Need to guard against None with the "or []".
-        return set(self.tags or []).isdisjoint(tags)
+        return not set(self.tags or []).isdisjoint(tags)
```
tardyp: return bool(set(self.tags or []) & set(tags))

tardyp:
good catch!

master/buildbot/test/unit/test_status_builder.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self.assertFalse(sut.matchesAnyTag(set()))
+        self.assertFalse(sut.matchesAnyTag(set(('any-tag', 'tag'))))
+
+    def test_matchesAnyTag_no_match(self):
+        """
+        Return False when requested tags don't match.
+        """
+        sut = self.makeBuilderStatus()
+        sut.tags = {'one'}
```
djmitche: Set comprehensions aren't supported in 2.6

adiroiban: true... fixed. Maybe it would be nice to have a tox.ini file or some script to simplify testing on different python version.

tardyp: There is already one

adiroiban: Ok. It is not working on my system... will send a patch :)

in the end I hope I can use tox to run something like this
```
$ tox -e py26,py26,py27 buildbot/test/unit/test_status_builder.py
```
Thanks!

sa2ajj:
LGTM

sa2ajj:
Thank you, @adiroiban

PR #1577, opened 2015-03-06
Update Dockerfile

shanzi:
It's time to migrate to Ubuntu 14.04 LTS now :D

I have tested it in my environment described below and it works well:

    OS version: Ubuntu 14.04
    Docker version: 1.0.1
    Go version: go1.2.1

Also, I think get `pip install -v` is better as it outputs more verbose log when installing with pip.

### DIFF
--- a/master/contrib/Dockerfile
+++ b/master/contrib/Dockerfile
@@ -70,7 +70,7 @@ run /bin/echo -e "\
 run apt-get update
 run DEBIAN_FRONTEND=noninteractive apt-get install -y python-pip python-dev \
     supervisor git sudo ssh
-run pip install -v sqlalchemy==0.7.9 buildbot buildbot_slave
+run pip install buildbot buildbot_slave
 
 # Set ssh superuser (username: admin   password: admin)
 run mkdir /data /var/run/sshd
--- a/master/contrib/Dockerfile
+++ b/master/contrib/Dockerfile
@@ -54,7 +54,7 @@
 # ssh -p $(docker port $CONTAINER_ID 22 | cut -d: -f 2) admin@localhost
 #
 # Base docker image
-from ubuntu:12.04
+from ubuntu:14.04
 
 # Make dpkg happy with the upstart issue
 ## appears to not be necessary anymore --dustin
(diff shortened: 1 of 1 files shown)


master/contrib/Dockerfile
```diff
@@ -64,13 +64,13 @@ from ubuntu:12.04
...
-    deb http://archive.ubuntu.com/ubuntu precise main universe\n\
-    deb http://archive.ubuntu.com/ubuntu precise-updates main universe" > \
+    deb http://archive.ubuntu.com/ubuntu trusty main universe\n\
+    deb http://archive.ubuntu.com/ubuntu trusty-updates main universe" > \
     /etc/apt/sources.list
 run apt-get update
 run DEBIAN_FRONTEND=noninteractive apt-get install -y python-pip python-dev \
     supervisor git sudo ssh
-run pip install sqlalchemy==0.7.9 buildbot buildbot_slave
+run pip install -v sqlalchemy==0.7.9 buildbot buildbot_slave
```
rutsky: Are you sure that verbose output of pip is needed?

shanzi: When I am trying with the original docker file, it stops here a lot of time outputing nothing, so I think verbose with level one is better?

BTW: the `pip install` instructions in the traditional [firstrun](http://docs.buildbot.net/current/tutorial/firstrun.html) page installs `sqlalchemy==0.7.9`. Maybe we should updated here too?

shanzi: If `-v` is not desired here, I will change it back to the origin.
Should I change `sqlalchemy==0.7.9` to `sqlalchemy==0.7.10`

rutsky: @shanzi I just wanted to note, that your changes not only updates Ubuntu image version, but also changes verbosity. Initially I thought that you added `-v` in commit by mistake.
If `-v` doesn't add too much output and works as "progress bar" for end user, then it's ok.

rutsky:
@shanzi I can't comment about installing `sqlalchemy==0.7.9` or `sqlalchemy==0.7.10`, you need to ask other developers about it.

benallard:
I believe you shouldn't need to install sqlalchemy independently.

shanzi:
@benallard 
Are you meaning remove `sqlalchemy` from the `pip` command is ok? 

But both the instructions in the [First Run](http://docs.buildbot.net/current/tutorial/firstrun.html) page and the `Dockerfile` itself before my modification installs it by hand independently.

shanzi:
@benallard After reviewed the [setup.py](https://github.com/buildbot/buildbot/blob/master/master/setup.py#L375) of master and tested the Dockerfile without install sqlalchemy independently, I am convinced that `pip install sqlalchemy` is not necessary, so I dropped it. 

I think the [First Run](http://docs.buildbot.net/current/tutorial/firstrun.html) instructions may need update.

@rutsky  I also dropped the `-v` as I found the stopping and outputting nothing might be a temp network problem. The `pip`  without `-v` will output a proper amount of logs. :P

tardyp:
:+1:

djmitche:
The last run insructions have already been updated in the /latest/ docs (which are generated from master) -- http://docs.buildbot.net/latest/tutorial/firstrun.html

djmitche:
Nice work, thank you!

PR #1578, opened 2015-03-07
Issue 2464: Makes GitPoller show small error message and not a full traceback.

prasoon2211:
Ref: [Issue 2464](http://trac.buildbot.net/ticket/2464)

When the error occurs on Git's end, we should not show the full taceback to the users (since it's not a problem with buildbot). This PR takes the Exception's error message and pushes it to stderr.

### DIFF
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -16,6 +16,7 @@
 import itertools
 import os
 import re
+import time
 import urllib
 
 from twisted.internet import defer
@@ -26,6 +27,10 @@ from buildbot import config
 from buildbot.changes import base
 from buildbot.util import ascii2unicode
 from buildbot.util.state import StateMixin
+
+
+class GitError(Exception):
+    """Raised when git exits with code 128."""
 
 
 class GitPoller(base.PollingChangeSource, StateMixin):
@@ -145,7 +150,13 @@ class GitPoller(base.PollingChangeSource, StateMixin):
 
     @defer.inlineCallbacks
     def poll(self):
-        yield self._dovccmd('init', ['--bare', self.workdir])
+        while True:
+            try:
+                yield self._dovccmd('init', ['--bare', self.workdir])
+            except GitError, e:
+                log.msg(e)
+            else:
+                break
 
         branches = self.branches
         if branches is True or callable(branches):
@@ -159,8 +170,16 @@ class GitPoller(base.PollingChangeSource, StateMixin):
             '+%s:%s' % (self._removeHeads(branch), self._trackerBranch(branch))
             for branch in branches
         ]
-        yield self._dovccmd('fetch',
-                            [self.repourl] + refspecs, path=self.workdir)
+
+        while True:
+            try:
+                yield self._dovccmd('fetch', [self.repourl] + refspecs,
+                                    path=self.workdir)
+            except GitError, e:
+                log.msg(e)
+                time.sleep(self.pollInterval)
+            else:
+                break
 
         revs = {}
         log.msg('gitpoller: processing changes from "%s"' % (self.repourl,))
@@ -311,6 +330,9 @@ class GitPoller(base.PollingChangeSource, StateMixin):
             "utility to handle the result of getProcessOutputAndValue"
             (stdout, stderr, code) = res
             if code != 0:
+                if code == 128:
+                    raise GitError('command %s %s in %s on repourl %s failed with exit code %d: %s'
+                            % (command, args, path, self.repourl, code, stderr))
                 raise EnvironmentError('command %s %s in %s on repourl %s failed with exit code %d: %s'
                                        % (command, args, path, self.repourl, code, stderr))
             return stdout.strip()
--- a/master/buildbot/test/unit/test_changes_gitpoller.py
+++ b/master/buildbot/test/unit/test_changes_gitpoller.py
@@ -285,6 +285,19 @@ class TestGitPoller(gpo.GetProcessOutputMixin,
                 'master': '4423cdbcbb89c14e50dd5f4152415afd686c5241'
             })
 
+    def test_poll_GitError(self):
+        # Raised when git exits with status code 128
+        self.expectCommands(
+            gpo.Expect('git', 'init', '--bare', 'gitpoller-work')
+            .exit(128),
+        )
+
+        d = self.assertFailure(self.poller._dovccmd('init', ['--bare',
+                               'gitpoller-work']), gitpoller.GitError)
+
+        d.addCallback(lambda _: self.assertAllCommandsRan())
+        return d
+
     def test_poll_nothingNew(self):
         # Test that environment variables get propagated to subprocesses
         # (See #2116)


master/buildbot/changes/gitpoller.py
```diff
@@ -311,6 +312,16 @@ def _convert_nonzero_to_failure(res,
...
+                if code == 128:
+                    # 128 is the error code git returns when things are
+                    # wrong on its end
+                    error_message = ('RuntimeError: command %s %s in %s on '
+                                    'repourl %s failed with exit code %d: %s'
+                                    % (command, args, path, self.repourl,
+                                       code, stderr)).strip()
+                    sys.stderr.write(error_message)
+                    sys.stderr.flush()
+                    sys.exit(1)
```
sa2ajj: I am not sure about `sys.exit(1)` here -- this means that should something get broken while master is running, it will suddenly stop.

prasoon2211: It should, shouldn't it? We are just trying to replicate an exception without actually providing a traceback. Before the PR, the exception was already halting the build since the poller couldn't fetch the change. We're doing the same here, minus the long traceback.

sa2ajj: It somehow contradicts the current behaviour: if config is OK and build master started, in case of problems they just fill log file.

What about other similar components?  Two questions:
* can we detect this kind of "fatal" error?
* if yes, why do not we also stop in those cases?

Changing the behaviour for only one component feels incomplete to me.

@djmitche do we have any other considerations to take into account?

tardyp: It is not the case. Buildbot is an agregation of a lot of services. An exception in one of them dont make the master shutdown. Only this service would be shutdown, and if there are other source of changes in the same master, or forced build, the master can still be available.

Here, as it looks like this comes from a misconfigured git server, I would suggest to log the error (with log.msg), and then retry the poll after a delay.

sa2ajj: Thanks, @tardyp, this is a good description of my _feeling_ :)

djmitche:
Yes, the change in the patch is simply to log the error from 'git' without the full traceback.  I believe it will already retry the operation after a delay.  Exiting is definitely not the right choice.

master/buildbot/changes/gitpoller.py
prasoon2211: @tardyp Hmm yes. So, if there are other change sources, then master should stay running. Okay then, I'll log the error and retry the poller with a timeout.

prasoon2211: @sa2ajj I wanted to do as @tardyp said - to log the error and retry the polling. But, if I retry the poll command right within `_convert_nonzero_to_failure`, then the stack will keep growing until there's a stack overflow (since the function will never return and keep calling the poll function recursively. `_convert_nonzero_to_failure` -> `poll` -> `_convert_nonzero_to_failure` -> `poll` -> .... ).
What we *could* do is to raise an Exception (as was happening before) and catch it in the `poll` function. But then, we won't be able to get the exit code. For this, we can create a new Exception, maybe a `GitException` that will be raised when error code is 128. We can catch this exception in `poll`, log it and retry the command doing the poll after the `pollInterval`.
Should I go ahead with this? Or, would you like to have it done some other way?

sa2ajj:
Then we all agree that the only thing that needs to be done is to print message and not to raise an exception.

prasoon2211:
Indeed we do. Please let me know what should be done for the [comment I added](https://github.com/buildbot/buildbot/pull/1578#discussion_r26127247).

prasoon2211:
@sa2ajj So, I made the changes and things are working as expected now. Please review.

master/buildbot/changes/gitpoller.py
```diff
@@ -159,8 +168,16 @@ def poll(self):
             '+%s:%s' % (self._removeHeads(branch), self._trackerBranch(branch))
             for branch in branches
         ]
-        yield self._dovccmd('fetch',
-                            [self.repourl] + refspecs, path=self.workdir)
+
+        while True:
+            log.msg("Haha")
```
tardyp: Haha?

master/buildbot/changes/gitpoller.py
```diff
@@ -159,8 +168,16 @@ def poll(self):
...
+            try:
+                yield self._dovccmd('fetch', [self.repourl] + refspecs,
+                                    path=self.workdir)
+            except GitError, e:
+                log.msg(e)
```
tardyp: I'll make a delay here in order to avoid flooding the git server

master/buildbot/changes/gitpoller.py
prasoon2211: Sorry! I put in random log message for debugging :smile:

master/buildbot/changes/gitpoller.py
prasoon2211: Okay, will do.

prasoon2211:
@tardyp I've done as you asked. Also, I've squashed all the commits and rebased on top of master.

Travis is showing that `make docs` is failing on several builds \[1\]  (including this one) which is happening due to the [latest sphinx release on 10th March](http://sphinx-doc.org/changes.html#release-1-3-released-mar-10-2015). For this, changing `latex_show_urls = True` to `latex_show_urls = 'inline'` in `master/docs/conf.py` [fixes the problem](http://sphinx-doc.org/config.html). Should I make a separate PR for that issue or should I include it in this PR itself?

\[1\]
https://travis-ci.org/buildbot/buildbot/jobs/54057418#L5352
https://travis-ci.org/buildbot/buildbot/jobs/54057276#L5266
https://travis-ci.org/buildbot/buildbot/jobs/54052338#L5266

tardyp:
yes, that would be great if you'd fix http://trac.buildbot.net/ticket/3230#ticket in a separate PR!

master/buildbot/changes/gitpoller.py
```diff
@@ -159,8 +170,16 @@ def poll(self):
...
-        yield self._dovccmd('fetch',
-                            [self.repourl] + refspecs, path=self.workdir)
+
+        while True:
+            try:
+                yield self._dovccmd('fetch', [self.repourl] + refspecs,
+                                    path=self.workdir)
+            except GitError, e:
+                log.msg(e)
+                time.sleep(self.pollInterval)
```
tardyp: ah ah you felt in my trap!
http://stackoverflow.com/questions/9310350/twisted-non-blocking-method-how-to

```
        yield util.asyncSleep(self.pollInterval)

```

prasoon2211: Ah.. I hadn't thought that this might be a problem. I kept thinking that since we're using `@defer.inlineCallbacks` so everything can work as in the blocking case :smile: I'll need to read twisted docs some more, I guess.

master/buildbot/changes/gitpoller.py
prasoon2211: Okay, done.

master/buildbot/changes/gitpoller.py
```diff
@@ -24,10 +25,14 @@
 
 from buildbot import config
 from buildbot.changes import base
-from buildbot.util import ascii2unicode
+from buildbot.util import ascii2unicode, asyncSleep
```
tardyp: pep8

tardyp:
one pep8 issue left, and you're good to go!

prasoon2211:
Closing and reopening to trigger a travis build. Also, I think everything's good now.

prasoon2211:
In the latest build failure, a [project has migrated](https://code.google.com/p/sqlalchemy-migrate/) from Google Code to Github and so, the python setup cannot finish. See [link](https://travis-ci.org/buildbot/buildbot/jobs/54118215#L628).

This is happening because [Google Code is shutting down](http://google-opensource.blogspot.in/2015/03/farewell-to-google-code.html) and Google has asked people to migrate to other hosting services. So, in the coming days, there might be even more broken builds if (when) any other project migrates.

prasoon2211:
@tardyp Review please. Pyflakes, pep8 give no errors (travis build is passing). I don't know why the coverage is being reduced though.

master/buildbot/changes/gitpoller.py
tardyp: please make one import per line.
http://legacy.python.org/dev/peps/pep-0008/#imports

prasoon2211: Huh. Odd. I don't know why pep8 is passing in travis. I'll correct it.

prasoon2211: I just looked again at the link you gave me. It says that when using the `from foo import bar` type declarations, it is okay to make more than one import per line.

> It's okay to say this though:
```
from subprocess import Popen, PIPE
```

Should I still change it?

djmitche: Yes please -- it's more of a Buildbot style.

shanzi:
@prasoon2211 you may need to `git pull upstream` and `git rebase upstream/master` to adapt you codes up to latest code base.

prasoon2211:
I have rebased my changes on top of master. I believe that this PR in mergeable now.

tardyp:
coveralls says coverage decreased. We usually ask Gsoc pretendant to create unit tests for their bug fixes.
How would you unit test this code?

prasoon2211:
A simple test would be to use a non-existent repository as a change source (that would yield a 404 error) and cause git to exit with 128 code. Then, I can check whether a GitError is raised.

I'll familiarize myself with the framework, add the test and update the PR.

prasoon2211:
@tardyp I've written a test and it works. But, GitHub won't let me repoen this pull request after a force push. I'll create a new one.

(I had to close this one because I had accidentally pushed some old changes).

PR #1579, opened 2015-03-08
ShellMixin: workdir should be overridable

jaredgrubb:
This is a port from the patch I made for eight. 
I did remove the doc notice; if it's part of 0.8.11's release notes, I dont think it should be part of 0.9, or?


sa2ajj:
Pending travis results...

sa2ajj:
And I did not answer the question: it should not be part of 0.9 of release notes.

PR #1580, opened 2015-03-09
Issue 3175: Alow customization of versions info on the About page

shanzi:
This pull request completed the task described in [\#3175](http://trac.buildbot.net/ticket/3175).

Now we can add user-defined versions information by adding a new value to `c['www']` with key named `versions`. The value should be a dictionary described the versions with key as the name of component and value as the the corresponding version. 

The versions of buildbot, python and twisted are now automatically found by server in `config.py` instead of hard coding.

The results looks like:

custom_versions

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -414,21 +414,18 @@ class MasterConfig(util.ComparableMixin):
         import twisted 
         from buildbot import version as bbversion
 
-        py_version_info = (sys.version_info.major,
-                           sys.version_info.minor,
-                           sys.version_info.micro)
-        pyversion = '.'.join(map(str, py_version_info))
+        pyversion = '.'.join(map(str, sys.version_info[:3]))
 
         tx_version_info= (twisted.version.major,
                           twisted.version.minor,
                           twisted.version.micro)
         txversion = '.'.join(map(str, tx_version_info))
 
-        return {
-            'pyversion': pyversion,
-            'bbversion': bbversion,
-            'txversion': txversion,
-        }
+        return [
+            ('Python version', pyversion),
+            ('Buildbot version', bbversion),
+            ('Twisted version', txversion),
+        ]
 
     def load_db(self, filenamekk, config_dict):
         self.db = dict(db_url=self.getDbUrlFromConfig(config_dict))
@@ -620,12 +617,12 @@ class MasterConfig(util.ComparableMixin):
                   (', '.join(unknown),))
 
         if www_cfg.get('versions') is None:
-            www_cfg['versions'] = {}
+            www_cfg['versions'] = []
 
-        if not isinstance(www_cfg['versions'], dict):
-            error("'versions' must be a dictionary")
+        if not isinstance(www_cfg['versions'], list):
+            error("'versions' must be a list of tuples")
 
-        www_cfg['env_versions'] = self.getEnvironmentVersions()
+        www_cfg['versions'] = self.getEnvironmentVersions() + www_cfg['versions']
 
         self.www.update(www_cfg)
 
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -409,7 +409,28 @@ class MasterConfig(util.ComparableMixin):
 
         return DEFAULT_DB_URL
 
-    def load_db(self, filename, config_dict):
+    @staticmethod
+    def getEnvironmentVersions():
+        import twisted 
+        from buildbot import version as bbversion
+
+        py_version_info = (sys.version_info.major,
+                           sys.version_info.minor,
+                           sys.version_info.micro)
+        pyversion = '.'.join(map(str, py_version_info))
+
+        tx_version_info= (twisted.version.major,
+                          twisted.version.minor,
+                          twisted.version.micro)
+        txversion = '.'.join(map(str, tx_version_info))
+
+        return {
+            'pyversion': pyversion,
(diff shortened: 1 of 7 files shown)


master/buildbot/config.py
```diff
@@ -592,12 +613,20 @@ def load_www(self, filename, config_dict):
...
             error("unknown www configuration parameter(s) %s" %
                   (', '.join(unknown),))
 
+        if www_cfg.get('versions') is None:
+            www_cfg['versions'] = {}
+
+        if not isinstance(www_cfg['versions'], dict):
+            error("'versions' must be a dictionary")
+
+        www_cfg['env_versions'] = self.getEnvironmentVersions()
```
djmitche: I think that this bit, and the `getEnvironmentVerions` method, should be somewhere under `master/buildbot/www` instead of here in `master/buildbot/config.py`.

djmitche:
Nice work!  This looks pretty good to me!  @tardyp may have other comments.

This would need documentation of the new configuration argument, and some Python unit tests for the config code and the new `getEnvironmentVersions` method.  But neither of those should be too hard.

master/buildbot/config.py
tardyp: env_versions and versions should not be separated when passed to front-end.

you can just use the dict.update() method to merge the two dictionaries

tardyp:
yep, nice work! I agree with @djmitche, this should be in buildbot.www.service module

master/buildbot/config.py
shanzi: @tardyp I separated two type of versions in order to keep a neat order of versions displayed on the page as the key order of a dictionary in python may be unexpected.

tardyp: Then we an forget about the dict, and rather use a list of tuples.

We an also use OrderedDict, but I think that will need some attention when serializing to json.

shanzi:
@djmitche @tardyp  I put `getEnvironmentVersions` here in case there are somewhere else need to access env versions other than `www`. But if you all think `buildbot.www.service` is better, I do not resist moving them.

hmmm. I will add documentation and tests in one or two days, now it's time to go to sleep :zzz:

sa2ajj:
If it's front end specific thing (and at the moment it is), I'd be inclined to move it to `buildbot.www.service`.

If somebody finds another use for this, then it could be moved depending on the use.

shanzi:
Wow, tests failed.

Hmmm.... It seems python 2.6's version_info has a different behavior with python 2.7

shanzi:
@tardyp should I omit `versions` when displaying `config` ?

shanzi:
@tardyp @djmitche @sa2ajj I think `www/config` is a better place than `www/service` so I move the code there. I have added some tests, please have a look at them.

Now we use a list of tuples to define the custom versions. I will update the document soon :)

master/buildbot/www/config.py
```diff
@@ -53,11 +72,20 @@ def renderIndex(self, request):
             config.update({"user": session.user_info})
         else:
             config.update({"user": {"anonymous": True}})
+
+        www = self.master.config.www
+        versions = []
+        versions += self.env_versions
```
tardyp: even if this is not a lot of work for master, I would prefer if the dictionary would be calculated inside the reconfigResource.

master/buildbot/www/config.py
```diff
@@ -37,6 +38,24 @@ def reconfigResource(self, new_config):
...
+
+        pyversion = '.'.join(map(str, sys.version_info[:3]))
+
+        tx_version_info= (twisted.version.major,
+                          twisted.version.minor,
+                          twisted.version.micro)
+        txversion = '.'.join(map(str, tx_version_info))
+
+        return [
+            ('Python version', pyversion),
```
tardyp: indeed, you can say just python, and add ' version' in the UI.
This will allow i18n in the future, as version is a translatable string.

shanzi:
It should be ok now :) Any other questions?

www/base/src/app/about/about.tpl.jade
```diff
@@ -5,15 +5,14 @@
         .row
           .col-sm-4
             ul
-              li Buildbot version: {{config.bbversion}}
-              li Twisted version: {{config.txversion}}
               li Project Infos:
-                a(ng-href="{{config.projecturl}}") {{config.projectname}}
+                a(ng-href="{{config.titleURL}}") {{config.title}}
+              li(ng-repeat="v in config.versions" ng-bind-template="{{v[0]}} version: {{v[1]}}")
```
tardyp: please dont use ng-bind-template, this is of no use in this case:
```
li(ng-repeat="v in config.versions")  {{v[0]}} version: {{v[1]}}
```

shanzi: @tardyp `ng-bind-template` has an advantage over using template directly inner the HTML tag. The tag will shows nothing before the rendering finished.

tardyp: What you say is only valid for main html page, not for templates.
Angularjs is parsing the templates before pushing them to the DOM, and the controller is also instanciated before the page is displayed, so everything is okay for using the simplified templating.

shanzi: @tardyp Okay~ changing it now~

tardyp:
:+1: good work!

djmitche:
Agreed, nice job!

PR #1581, opened 2015-03-10
Add support for extraRecipient to be renderable

vlovich:
Make extraRecipients in MailNotifier renderable. Useful for builds where the set of recipients may change between builds (e.g. if a builder is building multiple codebases).

Edit by @sa2ajj: copy the message from the original PR.

### DIFF
--- a/master/buildbot/status/mail.py
+++ b/master/buildbot/status/mail.py
@@ -30,6 +30,9 @@ from twisted.internet import reactor
 from twisted.python import log as twlog
 from zope.interface import implements
 
+from buildbot.interfaces import IRenderable
+from buildbot.util import flatten
+
 try:
     from twisted.mail.smtp import ESMTPSenderFactory
     ESMTPSenderFactory = ESMTPSenderFactory  # for pyflakes
@@ -266,7 +269,7 @@ class MailNotifier(base.StatusReceiverMultiService, buildset.BuildSetSummaryNoti
                                       of the Interested Users. If False, only
                                       send mail to the extraRecipients list.
 
-        @type  extraRecipients: tuple of strings
+        @type  extraRecipients: array or tuple of strings/renderables, or a renderable itself
         @param extraRecipients: a list of email addresses to which messages
                                 should be sent (in addition to the
                                 InterestedUsers list, which includes any
@@ -275,7 +278,9 @@ class MailNotifier(base.StatusReceiverMultiService, buildset.BuildSetSummaryNoti
                                 mailing list and deliver to that, then let
                                 subscribers come and go as they please.  The
                                 addresses in this list are used literally (they
-                                are not processed by lookup).
+                                are not processed by lookup).  If using renderables,
+                                the renderable may return a list of recipients.
+                                extraRecipients itself can be a renderable
 
         @type  subject: string
         @param subject: a string to be used as the subject line of the message.
@@ -384,14 +389,19 @@ class MailNotifier(base.StatusReceiverMultiService, buildset.BuildSetSummaryNoti
         """
         base.StatusReceiverMultiService.__init__(self)
 
-        if not isinstance(extraRecipients, (list, tuple)):
+        if IRenderable.providedBy(extraRecipients):
+            pass
+        elif not isinstance(extraRecipients, (list, tuple)):
             config.error("extraRecipients must be a list or tuple")
         else:
             for r in extraRecipients:
+                if IRenderable.providedBy(r):
+                    # renderable that we'll call later
+                    continue
                 if not isinstance(r, str) or not VALID_EMAIL.search(r):
                     config.error(
-                        "extra recipient %r is not a valid email" % (r,))
-        self.extraRecipients = extraRecipients
+                        "extra recipient %r is not a valid email nor a renderable" % (r,))
+        self._extraRecipients = extraRecipients
         self.sendToInterestedUsers = sendToInterestedUsers
         self.fromaddr = fromaddr
         if isinstance(mode, basestring):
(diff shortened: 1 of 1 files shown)


tardyp:
LGTM. needs tests!!! and docs!!

sa2ajj:
Travis fails:

```
FAILED (skips=8, failures=1, errors=18, successes=3060)
```

vlovich:
How can I figure out what the exact failures/errors are?  I can't really figure out what Travis is complaining about...

sa2ajj:
Oh, sorry.  Start looking at https://travis-ci.org/buildbot/buildbot/jobs/53756983#L886

Alternatively, checkout https://s3.amazonaws.com/archive.travis-ci.org/jobs/53756983/log.txt and search for FAIL.

vlovich:
Hmmm... Are the tests broken?  The interface for extraRecipients is defined to be a tuple/list but here we're getting a Mock object?

Can you provide guidance on what the best fix is?  It's also strange since the only way for _extraRecipients to be filled in is via the constructor which means somehow the mock object got by the isinstance check, was fine to iterate on in the constructor, but now we're having problems?

  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_status_mail.py", line 532, in test_buildFinished_mode_all_for_failure
    self.run_simple_test_sends_email_for_mode("all", FAILURE)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_status_mail.py", line 514, in run_simple_test_sends_email_for_mode
    mn.buildFinished('dummyBuilder', build, result)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/status/mail.py", line 558, in buildFinished
    self._renderRecipients(build)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/status/mail.py", line 547, in _renderRecipients
    for recipient in flatten([build.render(recipient) for recipient in build.render(self._extraRecipients)]):
exceptions.TypeError: 'Mock' object is not iterable

sa2ajj:
Tests are not broken (I checked them out :smile:).

I believe the problem comes from the fact that you don't declare `extraRecipients` as renderable and store the value in a property that is not "rendered".

I'd suggest to look at other instances of using renderable attributes and amend the patch accordingly (I really like its idea! :+1: ).

Please let me know if you'd like further guidance.

tardyp:
@sa2ajj This is a status plugin, so there is no such thing as automatically interpolation  of the renderable with status plugins.
I would believe the implementation is correct, as far as I see. Vitall, you can fix the tests if they are just mocking the extraRecipients instead of giving something viable.

vlovich:
Via visual inspection, I'm still not sure I see how to fix it.  Nothing tries to introspect or mock out the extra recipients.

I'll have to add some debug statements in the code & run the unit tests to see what's going on.

Is it possible that the rendering isn't correct - don't I have to yield build.render(recipient)?
I'm not sure I fully understand how the rendering & deferred interplay.

vlovich:
Hmmm... I just ran it locally.

Ran 2975 tests in 11.840s

PASSED (skips=38, successes=2937)

vlovich:
Maybe I ran it wrong?  trial buildbot.test

vlovich:
Never mind - the PYTHONPATH was wrong.  Was picking up the installed buildbot.

vlovich:
The mock build doesn't have render implemented.

vlovich:
FakeBuildStatus render() returns a MockObject but I'm not sure what's going on.  fakebuild.py is trying to register FakeBuildStatus as an adapter of IProperties.

I tracked it down as far as I can - I'm not familiar with the python interface library that's being used to know if the FakeBuildStatus is correct or not.

djmitche:
The fakes used for MailNotifier aren't very good.  They need to be fixed up in bug 3184.

I feel vaguely uncomfortable with the idea of rendering a renderable outside of the build process.  What do you think about configuring MailNotifier to just take a property name, and use the value of that property as the extra recipients?  `extraRecipientsFromProperty=".."`

sa2ajj:
@djmitche this is for eight, no Data API exists yet.

vlovich:
Hmmm.. I swear I replied to this.  I have a mild dislike from reading it from a property.

1. I think it likely suffers the same problem as the properties will be read from FakeBuildStatus unless I'm mistaken & if FakeBuildStatus is broken, then this could be a problem there anyway.

2.  I would need to have a step compute the set of interested recipients & set a property to it which feels much kludgier than having the notifier execute the renderable to get back the result.  It's also error-prone: what happens if a user forgets to make their step that sets this property alwaysRun=True?

sa2ajj:
Where are we with this PR?

djmitche:
@vlovich you're correct that using properties directly wouldn't help with the bad fakes -- you're on your own there :(

My concern is that the rendering process was designed to occur during step startup, and there may be hidden assumptions based on that design which will make it malfunction when performed after the step and build are complete, have been serialized to the DB, and unserialized on another master.  But I can't think of a specific assumption that would be invalidated (aside from the presence of a `buildbot.status.build.Build` object, which is already documented as a deprecated part of the rendering process).

So I'd say, go ahead and fix up the tests, and let's see what happens.  At worst, we end up backing out this feature because it can't be made to work reliably -- not the end of the world.

vlovich:
I'm assuming the multiple-master scenario is something new possible in 0.9?

Can I get guidance on how to fix the tests?  I'm still at a loss on exactly what the problem is & how I should fix it.

djmitche:
Yes; in eight MailNotifier will only send mail for builds that occur on the master where it's configured, and that bug will be fixed in nine.

jaredgrubb:
Another approach would be to allow 'messageFormatter' to return recipients in its dictionary. Or is that mixing concerns?

sa2ajj:
Where do we go from here?

vlovich:
Ping?  Still waiting on guidance for a workaround/fix for the tests.

djmitche:
Maybe I misunderstood the question.  The problem is that the tests don't pass, and you should fix it by either adjusting the tests so that they correctly validate the new (correct) behavior, or fixing the code if its behavior is incorrect.

vlovich:
As I said, the problem is that, for whatever reason, is that the fakes aren't rendering things properly.  I am waiting on guidance on how to fix that as I don't know how to begin to dig into FakeBuildStatus.

vlovich:
To clarify: there's nothing wrong with the tests nor the code.  The fake is broken & doesn't render properly.  I don't know how to fix that.

djmitche:
Looking back over this PR, I think maybe we should abandon it:

 * It renders outside of the step execution context, which is not an intended use of rendering and my gut tells me will fail in unexpected ways; @jaredgrubb had a good suggestion here, allowing the rendering to occur in the step context
 * It needs work on the tests; Vitali is looking for guidance, but I don't think there's any source for such guidance.
 * It's a patch against eight which someone else would need to forward-port to nine, or lose the feature on upgrade.

It sucks to turn away code, but sometimes it's the better choice..

djmitche:
Closing per my comments in April.  The code's still here, though!

PR #1582, opened 2015-03-10
added 3.3+ compatibility

kybrdbnd:
change to line 14  in module test_fiximport.py

### DIFF
--- a/common/test_fiximports.py
+++ b/common/test_fiximports.py
@@ -2,12 +2,16 @@ from fiximports import FixImports
 from textwrap import dedent
 from twisted.trial import unittest
 
+#returning functions
+def d_setUp():
+    return None
 
 class TestFixImports(unittest.TestCase):
 
     def setUp(self):
         self.fiximports = FixImports()
-        self.fiximports.printErrorMsg = lambda *_: None
+        ret_msg = d_setUp()
+        self.fiximports.printErrorMsg = ret_msg
 
     def oneTest(self, src, expected):
         src = dedent(src)
--- a/slave/buildslave/commands/base.py
+++ b/slave/buildslave/commands/base.py
@@ -132,8 +132,16 @@ class Command:
     # builder methods:
     #  sendStatus(dict) (zero or more)
     #  commandComplete() or commandInterrupted() (one, at end)
+    # 3.3+ functions
+    def list_args(x):
+        global args
+        l=[]
+        if x not in args:
+            l.append(x)
+        return l
+            
 
-    requiredArgs = []
+    #requiredArgs = []
     debug = False
     interrupted = False
     # set by Builder, cleared on shutdown or when the Deferred fires
@@ -147,7 +155,7 @@ class Command:
         self.args = args
         self.startTime = None
 
-        missingArgs = filter(lambda arg: arg not in args, self.requiredArgs)
+        missingArgs = list_args(arg)
         if missingArgs:
             raise ValueError("%s is missing args: %s" %
                              (self.__class__.__name__, ", ".join(missingArgs)))


common/test_fiximports.py
```diff
@@ -2,12 +2,16 @@
...
+#returning functions
+def d_setUp():
+    return None
 
 class TestFixImports(unittest.TestCase):
 
     def setUp(self):
         self.fiximports = FixImports()
-        self.fiximports.printErrorMsg = lambda *_: None
+        ret_msg = d_setUp()
```
djmitche: so `ret_msg` is None..

common/test_fiximports.py
```diff
@@ -2,12 +2,16 @@
...
+        self.fiximports.printErrorMsg = ret_msg
```
djmitche: which is not what `printErrorMessage` should be

slave/buildslave/commands/base.py
```diff
@@ -147,7 +155,7 @@ def __init__(self, builder, stepId, args):
         self.args = args
         self.startTime = None
 
-        missingArgs = filter(lambda arg: arg not in args, self.requiredArgs)
```
djmitche: How is this not Python-3.3 compatible?  Or PEP8?

djmitche:
You seem to have mixed up PEP8 and Python-3.3 compatibility here.  As far as I know, `lambda` is still supported in Python-3.3.  PEP8 prohibits assigning lambdas directly to a variable, so the first hunk is a reasonable PEP8 violation to fix.

common/test_fiximports.py
kybrdbnd: Yes

slave/buildslave/commands/base.py
kybrdbnd: this is also 3.3+ compatible but not according to PEP8 standards. That's why i have replaced the lambdas with a def function.

djmitche: What PEP8 standard does it violate?

tardyp:
Is this *all* what we need for python 3.3 is buildbot slave module?

If yes, please add a .travisrc change so that python 3.3 is tested, and run the buildslave unit tests

kybrdbnd:
my commit 'Python3.3 compatibility' means that the changes i have made are according to PEP8. That is i have replaced the lamdas with the def functions

common/test_fiximports.py
djmitche: And that's incorrect..

djmitche: Note that the existing code assigned a function of any number of positional arguments that always returns None.  That's very different from the simple value None.

djmitche:
That is a confusing name for a commit!  On that topic, there's also no need to indicate line numbers in the commit message -- the diff itself makes those clear.

kybrdbnd:
@tardyp i have just checked some of the modules because in the mail that i have received you mentioned that you have to look some of the modules and fix the lambas, pull the requests so that we can evaluate it. That's all i have done it now

slave/buildslave/commands/base.py
kybrdbnd: the use of lambas ofcourse

common/test_fiximports.py
kybrdbnd: for that i can change def setUp(**krgs): return None

slave/buildslave/commands/base.py
djmitche: PEP8 doesn't forbid lambdas!  It forbids the assignment of a lambda directly to a local variable - the fifth bullet point in https://www.python.org/dev/peps/pep-0008/#programming-recommendations

So actually, neither of the lambdas you're addressing here violate that recommendation -- in `test_fiximports.py` the lambda is assigned to an instance attribute (not a bare identifier), while in `base.py` the lambda is not assigned to anything.

common/test_fiximports.py
djmitche: It sounds like that would be more correct; however, I don't think that PEP8 actually requires this change.

djmitche:
This has gotten a little involved -- sorry about that!

I'd suggest that we close this pull request, and you begin a new one described as fixing PEP8-violating lambda expressions rather than Python-3.3 compatibility.  Here are a few examples you could fix (but start by addressing just one):
```
dustin@euclid ~/code/buildbot/t/buildbot [master] $ git  grep -E ' [a-z0-9_]+ = lambda'
master/buildbot/locks.py:    debuglog = lambda m: None
master/buildbot/process/buildstep.py:    _run_finished_hook = lambda self: None  # for tests
master/buildbot/process/buildstep.py:                callback = lambda cmd_arg, local_logname=logname: self.addLog(
master/buildbot/process/buildstep.py:                callback = lambda cmd_arg, logname=logname: self.addLog(
master/buildbot/status/status_gerrit.py:        callback = lambda gerrit_version: self.processVersion(gerrit_version, func)
master/buildbot/steps/source/svn.py:        quote = lambda uri: urllib.quote(uri, "!$&'()*+,-./:=@_~")
master/buildbot/steps/transfer.py:    callback = lambda arg: step.step_status.addLog('stdio')
master/buildbot/test/unit/test_config.py:        func = lambda _: "dummy"
master/buildbot/test/unit/test_config.py:        callable = lambda: None
master/buildbot/test/unit/test_config.py:        callable = lambda: None
master/buildbot/test/unit/test_config.py:        callable = lambda: None
master/buildbot/test/unit/test_config.py:        ns = lambda: 'ns'
master/buildbot/test/unit/test_config.py:        nb = lambda: 'nb'
master/buildbot/test/unit/test_process_builder.py:        cble = lambda: None
master/buildbot/test/unit/test_steps_trigger.py:        make_fake_br = lambda brid, builderid: fakedb.BuildRequest(
master/buildbot/test/unit/test_steps_trigger.py:        make_fake_build = lambda brid: fakedb.Build(
master/docs/manual/cfg-buildslaves-openstack.rst:        cmp_fn = lambda x,y: cmp(x.created, y.created)
```

kybrdbnd:
okey

PR #1587, opened 2015-03-12
update www/README.txt

djmitche:


djmitche:
It's a README, so I'll just merge it.

PR #1598, opened 2015-03-15
Test for mixed failure of classifyChanges

djmitche:
Test for an insert failure midway through a classification, in which
case the rollback shouldn't roll back any successful inserts.

Fixes #2696.

(Backport of d8d375352687b6d88ae960416a3d25bce1de4854)


djmitche:
failure here is due to #1599

sa2ajj:
This needs to be rebased and pushed again.  Otherwise looks good.

sa2ajj:
(Or I could just merge it, it's just not having green Travis makes me feel a bit uneasy :))

PR #1584, opened 2015-03-11
Made the REST 404 message more verbose.

prasoon2211:
Ref: [Ticket 3054](http://trac.buildbot.net/ticket/3054).

PR #1602, opened 2015-03-16
fix for bad PR #1584

tardyp:

### DIFF
--- a/master/buildbot/data/base.py
+++ b/master/buildbot/data/base.py
@@ -88,6 +88,9 @@ class Endpoint(object):
     def startConsuming(self, callback, options, kwargs):
         raise NotImplementedError
 
+    def __repr__(self):
+        return "endpoint for" + self.pathPatterns
+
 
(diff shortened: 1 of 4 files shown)


yetanotherion:
:+1:

master/buildbot/data/base.py
```diff
@@ -88,6 +88,9 @@ def control(self, action, args, kwargs):
     def startConsuming(self, callback, options, kwargs):
         raise NotImplementedError
 
+    def __repr__(self):
+        return "endpoint for" + self.pathPatterns
```
sa2ajj: don't we try to avoid string summation?  and use interpolation instead?

[nit] missing space

djmitche: I don't have a preference of one approach over the other, but yes -- the missing space needs to be added.  With that fixed, I think this is ready to merge.

djmitche:
I'll add the space while merging..

PR #1603, opened 2015-03-17
Websocket based event source

tardyp:
As an alternate for the sse protocol, this is a websocket based event source.

I hope this can fix some of the blockers we currently have with buildbot nine's realtime UI.


tothandras:
Wow, I haven't seen this coming. I was just reading the autobahn
documentatin. :D

tardyp:
This is only a tiny little first step before breaking everything with a new buildbotservice.

I think this protocol can be extended to resolve the race conditions and latency overheads inherent to REST + event protocol. 
typically adding a cmd_bind api to ws who would:
1) call data api to get first version of the data
2) verify if the data is complete (needs per endpoint knowledge to know if its complete, and wont evolve anymore)
3) if data isn't complete, then startConsuming updates until the data is finished
4) call data api to get updated version of the data now we are listening
5) finally report those decision to the front end

If we do this algo in the front end each step is another serialised roundtrip to the server (3 round trips)

tardyp:
After more tests, I'm pretty happy with this version. This at least solve http://trac.buildbot.net/ticket/3225
in a better way, than what I originally proposed.
Still not optimised we dont share the data between tabs, but at least the UI is not flaky.

I tested the UI survives:
- Wifi shutdown
- laptop sleep mode
- nginx reverse proxy restart
- buildbot master restart

The ui shows an alert saying it was disconnected, and inform user when reconnection happens

This also fixes http://trac.buildbot.net/ticket/3233 which I think was related, but which is not.

www/base/src/styles/animations.less
```diff
@@ -38,6 +38,19 @@ tl;dr is:
   }
 }
 
+/* animation for  */
+.anim-fade {
+    -webkit-transition: .5s linear all;
```
shanzi: Only `-webkit` ? No `-moz`, `-o` ?

In fact I think `-webkit-transition` is not needed either. Currently any modern browser that supports `transition` seems has already removes the prefix.

tardyp: -webkit-transition is the only supported way by a number of non auto-upgrading mobile browsers

while -moz, -o are auto-upgrading, and are fully supporting css3 transition

oberstet:
@tardyp fwiw, if you want multiple tabs running a single app that uses WS, that can also be done by running the WS on a shared background worker .. at least we want to add that to AutobahnJS for WAMP https://github.com/tavendo/AutobahnJS/issues/95

tardyp:
@oberstet you can see my plan in tavendo/AutobahnJS#95 about sharing websocket.

Here the goal is to fix the bug without reafactoring everything, so that we can at least make a release. So websocket sharing is only an optimization, and wont be part of first nine release (hopefully we can release before its ready)

sa2ajj:
(restarted Travis just to make sure)

djmitche:
This looks great to me -- no objections :)

tardyp:
shall I label it merge-me ?

djmitche:
If you're happy with it, go ahead and merge.

PR #1604, opened 2015-03-18
Fix sigtermTime race.

tomprince:
If twisted notices that a process dies between when SIGTERM is sent,
then RunProcess.process.pid will be None, causing checkDead to fail.

In any case, the timer will be leaked, and it will be cancelled after having
been called.


tomprince:
It isn't obvious how to easily setup a test for this.

tardyp:
LGTM, welcome back!

PR #1605, opened 2015-03-19
Issue 2464: Makes GitPoller write git based error to log file.

prasoon2211:
* Added GitError: Called when git returns with status code 128.
* The error message is logged to twistd.log
* Added test.

### DIFF
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -28,6 +28,10 @@ from buildbot.util import ascii2unicode
 from buildbot.util.state import StateMixin
 
 
+class GitError(Exception):
+    """Raised when git exits with code 128."""
+
+
 class GitPoller(base.PollingChangeSource, StateMixin):
 
     """This source will poll a remote git repo for changes and submit
@@ -145,7 +149,11 @@ class GitPoller(base.PollingChangeSource, StateMixin):
 
     @defer.inlineCallbacks
     def poll(self):
-        yield self._dovccmd('init', ['--bare', self.workdir])
+        try:
+            yield self._dovccmd('init', ['--bare', self.workdir])
+        except GitError, e:
+            log.msg(e.args[0])
+            return
 
         branches = self.branches
         if branches is True or callable(branches):
@@ -159,8 +167,13 @@ class GitPoller(base.PollingChangeSource, StateMixin):
             '+%s:%s' % (self._removeHeads(branch), self._trackerBranch(branch))
             for branch in branches
         ]
-        yield self._dovccmd('fetch',
-                            [self.repourl] + refspecs, path=self.workdir)
+
+        try:
+            yield self._dovccmd('fetch', [self.repourl] + refspecs,
+                                path=self.workdir)
+        except GitError, e:
+            log.msg(e.args[0])
+            return
 
         revs = {}
         log.msg('gitpoller: processing changes from "%s"' % (self.repourl,))
@@ -311,6 +324,9 @@ class GitPoller(base.PollingChangeSource, StateMixin):
             "utility to handle the result of getProcessOutputAndValue"
             (stdout, stderr, code) = res
             if code != 0:
+                if code == 128:
+                    raise GitError('command %s %s in %s on repourl %s failed with exit code %d: %s'
+                                   % (command, args, path, self.repourl, code, stderr))
                 raise EnvironmentError('command %s %s in %s on repourl %s failed with exit code %d: %s'
                                        % (command, args, path, self.repourl, code, stderr))
             return stdout.strip()
--- a/master/buildbot/test/unit/test_changes_gitpoller.py
+++ b/master/buildbot/test/unit/test_changes_gitpoller.py
@@ -22,6 +22,7 @@ from buildbot.changes import gitpoller
 from buildbot.test.util import changesource
 from buildbot.test.util import config
 from buildbot.test.util import gpo
+from buildbot.test.util import logging
 from twisted.internet import defer
 from twisted.trial import unittest
 
@@ -153,6 +154,7 @@ class GitOutputParsing(gpo.GetProcessOutputMixin, unittest.TestCase):
 
 class TestGitPoller(gpo.GetProcessOutputMixin,
                     changesource.ChangeSourceMixin,
+                    logging.LoggingMixin,
                     unittest.TestCase):
 
     REPOURL = 'git@example.com:foo/baz.git'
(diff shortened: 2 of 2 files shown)


master/buildbot/test/unit/test_changes_gitpoller.py
```diff
@@ -285,6 +285,19 @@ def cb(_):
...
             })
 
+    def test_poll_GitError(self):
+        # Raised when git exits with status code 128. See issue 2468
+        self.expectCommands(
+            gpo.Expect('git', 'init', '--bare', 'gitpoller-work')
+            .exit(128),
+        )
+
+        d = self.assertFailure(self.poller._dovccmd('init', ['--bare',
```
tardyp: please use inlineCallbacks

master/buildbot/changes/gitpoller.py
```diff
@@ -159,8 +170,16 @@ def poll(self):
             '+%s:%s' % (self._removeHeads(branch), self._trackerBranch(branch))
             for branch in branches
         ]
-        yield self._dovccmd('fetch',
```
tardyp: this loop is not tested

prasoon2211: Is there a way to capture the twisted log output in the test suite?

prasoon2211: I did a few grep searches and found a utility Mixin for this - logging.LoggingMixin.

prasoon2211:
I've added tests for the loop @tardyp

master/buildbot/test/unit/test_changes_gitpoller.py
```diff
@@ -285,6 +287,42 @@ def cb(_):
...
             })
 
+    def test_poll_GitError(self):
+        # Raised when git exits with status code 128. See issue 2468
+        self.expectCommands(
+            gpo.Expect('git', 'init', '--bare', 'gitpoller-work')
+            .exit(128),
+        )
+
+        @defer.inlineCallbacks
```
tardyp: this is not what I expected. :-) try again!

master/buildbot/test/unit/test_changes_gitpoller.py
```diff
@@ -285,6 +287,42 @@ def cb(_):
...
+        def run_command():
+            yield self.poller._dovccmd('init', ['--bare',
+                                       'gitpoller-work'])
+
+        d = self.assertFailure(run_command(), gitpoller.GitError)
+
+        d.addCallback(lambda _: self.assertAllCommandsRan())
+        return d
+
+    def test_poll_GitError_loop(self):
```
tardyp: Also here, need inline callbacks

I guess you also want to patch asyncSleep so that the test does not unneedly wait 1sec

prasoon2211:
@tardyp Done.

master/buildbot/changes/gitpoller.py
```diff
@@ -159,8 +170,16 @@ def poll(self):
...
-                            [self.repourl] + refspecs, path=self.workdir)
+
+        while True:
+            try:
+                yield self._dovccmd('fetch', [self.repourl] + refspecs,
+                                    path=self.workdir)
+            except GitError, e:
+                log.msg(e)
+                asyncSleep(self.pollInterval)
```
djmitche: The Deferred returned by asyncSleep is being ignored, so this will run `git fetch` repeatedly with no pause between runs.

master/buildbot/changes/gitpoller.py
djmitche: This would be great -- I think you'll want `log.msg(e.args[0])`, but I'd like to see the resulting log message in the test to be sure.

master/buildbot/changes/gitpoller.py
prasoon2211: I actually yielded this line before. But then, I messed up with a force push, did it whole PR all again from memory and forgot the `yield`. I'll fix it asap.

master/buildbot/changes/gitpoller.py
prasoon2211: Well, I tried using a non-existent change source and the error message was being logged correctly. But this does seem better to me as well. Will do.

prasoon2211: Okay done.

master/buildbot/test/unit/test_changes_gitpoller.py
djmitche: The rest of this file uses `d`, so I think it's OK here too.

master/buildbot/changes/gitpoller.py
djmitche: For me this test logged:

```
2015-03-22 12:23:57-0400 [-] --> buildbot.test.unit.test_changes_gitpoller.TestGitPoller.test_poll_GitError_loop <-- 
2015-03-22 12:23:57-0400 [-] command init ['--bare', 'gitpoller-work'] in None on repourl git@example.com:foo/baz.git failed with exit code 128:⋅
```
with no attending exception, so that looks great

master/buildbot/test/unit/test_changes_gitpoller.py
prasoon2211: So should I remove `inlineCallbacks`? (And, am I using `inlineCallbacks` correctly?)

djmitche:
The logging bit of this patch looks perfect.  However, in #1578 we discussed whether to retry -- the code *already* retries on errors, and that retry logic should be allowed to continue.  The retry logic you've added is incorrect in two ways: for `git init` your code loops infinitely with no delay, while for `git fetch` you have a delay but still loop forever.  For non-transient errors, both of these would cause the GitPoller to hang, forever retrying something that will never succeed, and never trying any other operations.

As things stand in the master branch, retries occur when an `EnvironmentError` is raised.  It is caught by the `doPoll` method in `buildbot.changes.base.PollingChangeSource`, which also logs the error, then returns (without an exception) to the caller.  At the next polling interval, the poller gets invoked again.

Since we want to avoid logging a traceback, the right solution here is to simply return from the `poll` method when a `GitError` occurs, as if everything had gone fine.  Then at the next polling interval, we'll get another chance.

prasoon2211:
Okay then. I'll just return instead of `break`ing. Also, I'll likely need to modify tests as well. I'm currently writing my proposal so, I'll fix it tomorrow if that's okay.

prasoon2211:
@djmitche I've made the changes you asked for. So, we don't have a loop in the `poll` method now and we just return if there's a `GitError`. I've also removed `inlineCallbacks` and instead returned `d` as is done in all other tests in the tests file. Tests were updated too. Also, this PR has been rebased on top of master.
I just checked it with a bad URL and the logging works just fine as well. So, I suppose if travis passed the build, then it should be good to merge.

djmitche:
Yes, looks great -- thanks for your perseverance!

PR #1606, opened 2015-03-22
2745 Factor duplicate code in master/buildbot/db/users.py

rafalkowalski:
Simple refactoring

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -589,6 +589,9 @@ class Git(Source):
 
             @d.addCallback
             def checkWithListdir(_):
+                if 'files' not in cmd.updates:
+                    # no files - directory doesn't exist
+                    return "clone"
                 files = cmd.updates['files'][0]
                 if '.git' in files:
                     return "update"


master/buildbot/db/users.py
```diff
@@ -130,23 +132,10 @@ def thd(conn):
             if not users_row:
                 return None
 
-            # make UsDict to return
-            usdict = UsDict()
-
             # gather all attr_type and attr_data entries from users_info table
             q = tbl_info.select(whereclause=(tbl_info.c.uid == users_row.uid))
             rows = conn.execute(q).fetchall()
```
djmitche: You could factor this a little more, so that you only pass a uid and not `rows`

master/buildbot/db/users.py
```diff
@@ -99,26 +99,28 @@ def thd(conn):
...
-            usdict['identifier'] = users_row.identifier
-            usdict['bb_username'] = users_row.bb_username
-            usdict['bb_password'] = users_row.bb_password
-
-            return usdict
+            return self.createUsDict(users_row, rows)
         d = self.db.pool.do(thd)
         return d
 
+    def createUsDict(self, users_row, rows):
```
djmitche: Within the DB API, we are careful to distinguish functions which are called in a sub-thread (and can't use asynchronous features like Deferreds) from those which are called in the main thread (and can use such features).  We do this with a naming convention: "thd_" in the function name.  So, please rename this to `thd_createUsDict`.

djmitche:
Please also be careful in how you name your commits.  We often look back in the history of a project to find out why some piece of code says what it does.  "Function create us dict" doesn't tell us much (and doesn't parse as an English phrase).  In this case, I might suggest something like "factor common code into thd_createUsDict".

PR #1608, opened 2015-03-24
fix encoding header validation

tardyp:
some browsers use:

content-type: application/json; Charset=UTF-8

So we use a real content-type parser.

fixes http://trac.buildbot.net/ticket/3238


yetanotherion:
travis seems to have failed, :+1 else

tardyp:
its a coveralls false positive.

PR #1612, opened 2015-03-26
Don't hardcode TLS version to SSLv3

SpecLad:
SSL 3.0 is now considered insecure (because of the POODLE attack), and servers are starting to turn it off. In any case, hardcoding a particular version is a bad idea - let Twisted figure out which version to use.

### DIFF
--- a/master/buildbot/status/mail.py
+++ b/master/buildbot/status/mail.py
@@ -36,13 +36,6 @@ try:
 except ImportError:
     ESMTPSenderFactory = None
 
-have_ssl = True
-try:
-    from twisted.internet import ssl
-    from OpenSSL.SSL import SSLv3_METHOD
-except ImportError:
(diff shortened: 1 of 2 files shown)


SpecLad:
I'm not sure about how to go about submitting patches to eight - do I also need to port and submit this to master?

djmitche:
We can cherry-pick this one commit without too much trouble.

SpecLad:
> We can cherry-pick this one commit without too much trouble.

So... I can see that this hasn't happened yet. Is anything the matter?

djmitche:
http://trac.buildbot.net/ticket/3246 is making it hard for me to test things locally.

djmitche:
Cherry-pick merged in 0f01dc0b19c7f9198ec75dd2cfd1aa033cd322de

PR #1617, opened 2015-03-31
Waterfall scaling with + and - keys

tothandras:
@tardyp I could not try this out, api_proxy doesn't work anymore due to moving from sse to ws.


sa2ajj:
LGTM.

@tardyp ?

www/waterfall_view/src/module/main.module.coffee
```diff
@@ -104,6 +107,18 @@ class Waterfall extends Controller
             # Bind scroll event listener
             angular.element(containerParent).bind 'scroll', onScroll
 
+            @$window.onkeydown = (e) =>
+                # +
+                if e.keyCode is 107
```
tardyp: I would expect this to update the settings as well, so that you adjust the settings with keyboard

www/waterfall_view/src/module/waterfall.config.coffee
```diff
@@ -0,0 +1,57 @@
+class Waterfall extends Config
+    constructor: (bbSettingsServiceProvider) ->
+
+        bbSettingsServiceProvider.addSettingsGroup
+            name: 'Waterfall'
+            caption: 'Waterfall related settings'
+            items: [
```
tardyp: wow, that is a lot of settings. I'm not sure these are really user friendly.

www/waterfall_view/src/module/waterfall.config.coffee
```diff
@@ -0,0 +1,57 @@
...
+                type: 'integer'
+                name: 'margin_top_waterfall'
+                caption: 'Margin top (px)'
+                default_value: 15
+            ,
+                type: 'integer'
+                name: 'margin_right_waterfall'
+                caption: 'Margin right (px)'
+                default_value: 20
```
tardyp: why a user would want to change that?

tothandras:
@tardyp I have addressed your remarks.

tardyp:
👍

PR #1619, opened 2015-04-02
Don't use custom font for log view

krf:
Possibly a controversial change, here's my reasoning:
- Let's not force a specific monospace font on the user,
  \<pre\> already renders the text accordingly
- This is an almost raw plain text view, so use the browser defaults
- Courier New is just an awful font

### DIFF
--- a/master/buildbot/status/web/files/default.css
+++ b/master/buildbot/status/web/files/default.css
@@ -555,7 +555,6 @@ li {
 /* log view */
 .log * {
 	vlink: #800080;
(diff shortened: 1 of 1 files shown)


tardyp:
as this is for buildbot eight, I dont think a cosmetic patch should be accepted.

krf:
Other opinions? I'm considering buildbot eight to be a well-established branch which lives from such small improvements / bug fixes.

sa2ajj:
I'm inclined to accept it.

PR #1620, opened 2015-04-02
Minor refactoring in db/users.py

prasoon2211:
* Pulled out all tbl and tbl_info declarations into self
* Made thd_createUsDict a staticmethod.

### DIFF
--- a/master/buildbot/db/users.py
+++ b/master/buildbot/db/users.py
@@ -28,21 +28,23 @@ class UsDict(dict):
 class UsersConnectorComponent(base.DBConnectorComponent):
     # Documentation is in developer/db.rst
 
+    def __init__(self, *args, **kwargs):
+        super(UsersConnectorComponent, self).__init__(*args, **kwargs)
+        self._tbl = self.db.model.users
+        self._tbl_info = self.db.model.users_info
+
     def findUserByAttr(self, identifier, attr_type, attr_data, _race_hook=None):
         # note that since this involves two tables, self.findSomethingId is not
         # helpful
         def thd(conn, no_recurse=False, identifier=identifier):
-            tbl = self.db.model.users
-            tbl_info = self.db.model.users_info
-
-            self.checkLength(tbl.c.identifier, identifier)
-            self.checkLength(tbl_info.c.attr_type, attr_type)
-            self.checkLength(tbl_info.c.attr_data, attr_data)
+            self.checkLength(self._tbl.c.identifier, identifier)
+            self.checkLength(self._tbl_info.c.attr_type, attr_type)
+            self.checkLength(self._tbl_info.c.attr_data, attr_data)
 
             # try to find the user
-            q = sa.select([tbl_info.c.uid],
-                          whereclause=and_(tbl_info.c.attr_type == attr_type,
-                                           tbl_info.c.attr_data == attr_data))
+            q = sa.select([self._tbl_info.c.uid],
+                          whereclause=and_(self._tbl_info.c.attr_type == attr_type,
+                                           self._tbl_info.c.attr_data == attr_data))
             rows = conn.execute(q).fetchall()
 
             if rows:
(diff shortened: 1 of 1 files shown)


prasoon2211:
Minor refactoring according to [issue 2745](http://trac.buildbot.net/ticket/2745). Haven't tested yet and leaving the test to travis.

master/buildbot/db/users.py
```diff
@@ -28,21 +28,23 @@ class UsDict(dict):
 class UsersConnectorComponent(base.DBConnectorComponent):
     # Documentation is in developer/db.rst
 
+    def __init__(self, *args, **kwargs):
+        super(UsersConnectorComponent, self).__init__(*args, **kwargs)
+        self.tbl = self.db.model.users
```
sa2ajj: are these two part of the API?  if not, make them private (prepend the names with `_`).

prasoon2211:
@sa2ajj I've made the changes you asked for. Tests are passing. Coverage is decreased in db/users.py because of the refactoring. Coverage has also decreased by 3 lines in `/master/buildbot/steps/python_twisted.py` (see the coveralls report). But, how do I know which 3 lines? Can this change be tracked in coveralls?

sa2ajj:
I'd suggest comparing https://coveralls.io/builds/2283707 and https://coveralls.io/builds/2260041

prasoon2211:
I took a look a a couple of recent builds and in [every](https://coveralls.io/builds/2283707) [single](https://coveralls.io/builds/2307131) [one](https://coveralls.io/builds/2305700) [of them](https://coveralls.io/builds/2305591), the coverage for `/master/buildbot/steps/python_twisted.py` has covered -3 and missed +3.

So, perhaps this change in coverage of `/master/buildbot/steps/python_twisted.py` is due to external factors? If so, then this PR can be merged.

djmitche:
I think part of the coverage changes are due to different Travis jobs completing first for the same commit, with different coverage for each job (for example, due to tests not running on different python versions, or perhaps in this case due to different twisted versions).  I'll dial back Coveralls' sensitivity to changes a little bit, so don't take the coveralls "failure" here as problematic.

sa2ajj:
LGTM

I'd appreciate another couple of prying eyes though.

master/buildbot/db/users.py
```diff
@@ -28,21 +28,23 @@ class UsDict(dict):
 class UsersConnectorComponent(base.DBConnectorComponent):
     # Documentation is in developer/db.rst
 
+    def __init__(self, *args, **kwargs):
+        super(UsersConnectorComponent, self).__init__(*args, **kwargs)
+        self._tbl = self.db.model.users
```
tardyp: we are never using this design pattern in any db modules, so I dont think we should do it here.

tardyp:
I am not sure exactly what @djmitche meant with http://trac.buildbot.net/ticket/2745
I dont think putting the table models is a good idea, and it dont really factorize many thing.

:-1: for me

master/buildbot/db/users.py
djmitche: Yes, it was mainly the code duplication I was referring to -- which has been addressed here.  We do sometimes use local variables for tables, but not class variables, so I agree with Pierre that this particluar hunk, and uses of `self._tbl*`, should not be merged.

prasoon2211: Hey all. My end-of-term exams are going on right now. I'll look into this PR as soon at I'm free (3rd May).

sa2ajj:
@djmitche could you please elaborate?

djmitche:
I could have sworn I did elaborate :(

What I meant was just reducing the large swaths of duplicated code.  I agree that putting putting the table models in instance attributes is unusual and not a good idea (using local variables isn't a problem)

tardyp:
so I think we can close it then.

PR #1629, opened 2015-04-10
Non-blocking latent slaves, fix for ticket 3169

caktux:
Taken from https://github.com/ClusterHQ/build.clusterhq.com/blob/master/flocker_bb/monkeypatch.py


tardyp:
This delays all build requests by 10s? Looks like everybody will not agree..

djmitche:
Agree with Pierre -- there must be a better way to solve this..

PR #1630, opened 2015-04-11
fix for 3246

tardyp:
not a good idea to reuse the same BuildbotServices between tests
They wont be reconfigured between tests


sa2ajj:
LGTM

jaredgrubb:
LGTM!

master/buildbot/test/integration/test_try_client.py
```diff
@@ -89,13 +89,16 @@ def remote_startBuild(self):
 class Schedulers(dirs.DirsMixin, www.RequiresWwwMixin, unittest.TestCase):
 
     def setUp(self):
+        global BuildmasterConfig
```
Stibbons: also clean in the tearDown

PR #1631, opened 2015-04-12
Allow function callback for setting codebase in github hook

esben:
Allow using a function for setting codebase for github change_hook events.  This makes it possible to use it for multiple codebase setups.

### DIFF
--- a/master/buildbot/test/unit/test_status_web_hooks_github.py
+++ b/master/buildbot/test/unit/test_status_web_hooks_github.py
@@ -28,6 +28,7 @@ from buildbot.test.util import compat
 
 from StringIO import StringIO
 from twisted.trial import unittest
+from twisted.internet import defer
 
 # Sample GITHUB commit payload from http://help.github.com/post-receive-hooks/
 # Added "modfied" and "removed", and change email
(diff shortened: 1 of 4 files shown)


djmitche:
Can you update docs and a release note, too?

esben:
Updated with docs and release note.

djmitche:
Thanks!!

I'm sorry; I should have asked for docs, release note, and *tests*.  I hate when people move the review goalposts for my patches, so please forgive me.. I blame being at PyCon US and being distracted.

esben:
What level of testing is expected for this change?  Is there any test of the current codebase implementation in the github change hook (I cannot find it)?

djmitche:
The tests are in `master/buildbot/test/unit/test_status_web_hooks_github.py`.  This should at least test that a callable is accepted and its output handled appropriately.

esben:
Updated with a fix for some of the existing github change hook unit tests also.

esben:
Is it possible to get travis to retry  a failed test run (fx. when pip fails with network timeout)?

sa2ajj:
Yes, it is

master/buildbot/test/unit/test_status_web_hooks_github.py
```diff
@@ -229,10 +229,10 @@ def check_changes(_):
         return d
 
     def test_ping_encoded(self):
-        self._check_ping(['{}'])
+        return self._check_ping(['{}'])
```
sa2ajj: Why all these returns?  What's the reason behind?

sa2ajj:
Beside this unclear reason for using `return`, LGTM.

esben:
The reason for these returns are that (fx.) _check_ping() returns d, which is a Deferred.  The actual checks are attached to the Deferred as callbacks, so unless the Deferred is executed, the checks are not either.

Twisted is completely new to me, but I found that writing any random broken assert* statements in the callbacks did not have any effect.  Returning the d through the testcases fixes that, but I haven't investigated the trial executor for what exactly it does with the return.  From trial documentation:

> The main feature is the ability to return Deferreds from tests and fixture methods and to have the suite wait for those Deferreds to fire.

tardyp:
You should rather use @defer.inlinecallbacks to write asynchronous code.
It's much easier to read (and write)

Le mar. 14 avr. 2015 06:40, Esben Haabendal <notifications@github.com> a
écrit :

esben:
If there is interest in it, I can come with a new PR where I wholesale rewrite the tests in test_status_web_hooks_github.py to use inlinecallbacks.  But shouldn't we rather do that in a new PR instead of mixing things up?

tardyp:
OK that would be great!

Le mar. 14 avr. 2015 12:51, Esben Haabendal <notifications@github.com> a
écrit :

esben:
The rewrite to use inlineCallbacks is in PR #1633.

sa2ajj:
@esben shame on me for writing tests that are not executed.  Thank you for spotting and fixing it!

PR #1597, opened 2015-03-15
GitHub pull requests support

caktux:

PR #1632, opened 2015-04-14
Add pull_request support for github change hook handler

esben:


tardyp:
Did you see:

https://github.com/buildbot/buildbot/pull/1597

I think it had a little bit more features (eg file list)
But yours has test which is great!

Le mar. 14 avr. 2015 11:53, Esben Haabendal <notifications@github.com> a
écrit :

esben:
No, I seems to have missed that PR #1597.

The PR #1597 mixes changes to GitHubStatus and the GitHub change_hook dialect of WebStatus.

In this PR, I have both avoided mixing different features, and have avoided making changes to existing behaviours, and it should be possible to merge without breaking backwards compatibility.

I have chosen to see a PR as a single change, instead of injecting all the commits that might be included in that PR.  You can argue both for and against this, but the implementation comes out much simpler than what is needed in PR #1597.

It should be possible to add the other approach as an optional behavior on top of this PR, controllable by a dialect parameter.

tardyp:
I agree it's better to have a single change. I think it would be a good
feature to have the list of files though. But this can be a further topic.
👍

Le mar. 14 avr. 2015 12:40, Esben Haabendal <notifications@github.com> a
écrit :

sa2ajj:
(a rebase is required)

I'd appreciate if you split relnotes update into a separate commit: this helps a lot while porting the change (we have this requirement in [submitting patches page](http://trac.buildbot.net/wiki/SubmittingPatches)).

sa2ajj:
(and I should have asked for the same thing in the previous PR: I was just too ashamed by what I had done.)

master/buildbot/status/web/hooks/github.py
```diff
@@ -111,6 +111,43 @@ def handle_push(self, payload):
 
         return changes, 'git'
 
+    def handle_pull_request(self, payload):
+        changes = []
+        number = payload['number']
+        refname = 'refs/pull/%d/head' % (number)
```
sa2ajj: `(number)` is no different from `number`.  If you'd like to make sure it's a tuple use `(number,)`

esben: fixed

sa2ajj:
LGTM (waiting for Travis)

sa2ajj:
@esben thank you!

PR #1633, opened 2015-04-14
Rewrite github change hook unit tests to use @defer.inlineCallbacks

esben:
This also fixes the tests to actually fail when things are wrong.


sa2ajj:
I think this is still a worthy update, however a rebase is required.

sa2ajj:
@esben thank you!

PR #1634, opened 2015-04-15
Correct GithubStatus usage example

nicolai86:
The example usage for GithubStatus was wrong in the documentation.

### DIFF
--- a/master/docs/manual/cfg-statustargets.rst
+++ b/master/docs/manual/cfg-statustargets.rst
@@ -804,12 +804,13 @@ GitHubStatus
 
 ::
 
-    from buildbot.plugins import status, util
+    from buildbot.process.properties import Interpolate
+    from buildbot.status import github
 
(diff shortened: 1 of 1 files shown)


tardyp:
The doc is okay on how it should work with the new plugin mechanism. Perhaps github status is not correctly declared as a plugin in setup.py?

nicolai86:
Not sure, using the code from the docs / buildbot 0.8.10 fails due to undefined modules/ classes.

sa2ajj:
The documentation you change is for `master`, not for 0.8.10.  And 0.8.10 did not have plugins implemented, 0.8.11 will though.

I'l check the exact list of `status` plugins, but without looking at it I'm inclined to not accept this patch.

sa2ajj:
Well, I loked at both `master` and `eight` (upcoming 0.8.11) and they both define `GitHubStatus` as part of `buildbot.plugins.status`

PR #1635, opened 2015-04-19
[eight] FIx sphinx errors

jaredgrubb:
When I ran "make docs" I got a bunch of warning-as-errors. This patch fixes them.

Was anyone else getting these? Maybe my sphinx is too old/new and that's why no one else sees them?


tardyp:
:+1:

PR #1636, opened 2015-04-19
[eight] Backport the "dynamic trigger" from nine

jaredgrubb:
I had a use-case where this would be very helpful, so I backported a change made in nine by @tardyp that added "dynamic scheduler".

This is mostly the same patch, but I did have to modify it a bit to make it work with eight's scheduler stuff.

Also, I added a note about renderable properties.

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -69,6 +69,9 @@ class Trigger(LoggingBuildStep):
         self.ended = False
         LoggingBuildStep.__init__(self, **kwargs)
 
+        # Scheduler name cache
+        self._all_schedulers = None
+
     def interrupt(self, reason):
         if self.running and not self.ended:
             self.step_status.setText(["interrupted"])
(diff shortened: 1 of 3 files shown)


tardyp:
:+1:

master/buildbot/steps/trigger.py
```diff
@@ -80,33 +80,28 @@ def end(self, result):
...
-                    invalid_schedulers.append(scheduler)
-            else:
-                invalid_schedulers.append(scheduler)
-
-        return (triggered_schedulers, invalid_schedulers)
+    def getSchedulerByName(self, name):
+        # Use a quick cache to avoid generating this dict every time.
+        try:
+            all_schedulers = self._all_schedulers
+        except AttributeError:
```
sa2ajj: using exceptions as a flow control?

why not set `self._all_schedulers` to `None` and check against that?

jaredgrubb:
Adjusted the patch to address sa2ajj's comment.

Stibbons:
+1

PR #1638, opened 2015-04-21
remove logfile info from docs

elmirjagudin:
As far as I understand the logs are now stored in the database, thus we should remove outdated section about log files from the docs.


tardyp:
:+1:

PR #1639, opened 2015-04-22
MD: the about page

shanzi:
Here is the about page, half finished.

The page is responsive to the width of page. When it is thin:

aboutview-thin

when it is fat:

aboutview-fat

I wrote a directive to displaying objects like the old `rawdata` directive, but is more elegant and clean. I omitted the nested displaying function, but using `JSON.stringify` for a better view of the value. The field can be collapsed or expanded displaying JSON in different indention.

Field collapsed:

field-collapsed

Field expanded:

field-expanded

Live demo at: http://buildbot.apps.io-meter.com/#/about

**NOTE**: as displaying the api specs depend on `buildbotService`, it has not been added yet.

### TODO
- [x] tests for `inpectdata` directive
- [x] use arrows to expand and collapse fields
- [x] displaying API spec


shanzi:
Added tests for `inspectdata` and some styles improvements.

Followed @tothandras 's suggestions about `bindToController` for the directive.

shanzi:
@tothandras as suggested that all of our directives should have a prefix as namespace like `bb-`, I think it is a good idea. What do you think about this? cc @tardyp

tardyp:
Well, I think this is a very important for bower libraries. but for the main application, this is not really necessary.

www/md_base/src/app/common/directives/inspectdata/inspectdata.directive.coffee
```diff
@@ -0,0 +1,53 @@
...
+                value: value_short
+                value_long: value_long
+            }
+
+    addItem: (k, v) ->
+        @items.push
+            key: k
+            value: @processValue(v)
+
+    constructor: ($scope) ->
```
tardyp: we usually put the constructor first.

www/md_base/src/styles/inspectdata.less
```diff
@@ -0,0 +1,46 @@
+.inspect-data {
```
tardyp: please put the styles for the directive in the same directory as the directive

www/md_base/src/styles/about.less
```diff
@@ -0,0 +1,35 @@
+.about{
+  padding: 20px;
```
tardyp: That is a lot of custom styles for one page.

I'd prefer we stick with a common style

www/md_base/src/app/common/directives/inspectdata/inspectdata.tpl.jade
```diff
@@ -0,0 +1,13 @@
+div.inspect-data
+  div.row(layout="row", ng-repeat="item in inspectdata.items")
+    div.key(flex="33") {{ item.key }}
+    div.value(flex="66")
+      span(ng-if="item.value.type == 'string'") {{ item.value.value }}
+      a(ng-if="item.value.type == 'link'", ng-href="{{ item.value.value }}") {{ item.value.value }}
+      div.object-value(
+        ng-if="item.value.type == 'object'",
+        ng-init="collapsed = 1",
+        ng-click="collapsed = !collapsed",
```
tardyp: the problem with putting ng-click here is that you cannot easily copy paste the value, and you actually dont know that this is even clickable. So I would prefer you stick with something like this for expansion:

image

shanzi: Added a sticker will make the html a little complex. I prefer not to add one. 

The mouse cursor over the field will become pointer to indicate it is clickable and I think ng-click is smart enough to distinguish drag and select with click. I met no problem trying select the text there.

www/md_base/src/styles/about.less
shanzi: The styles for section will be moved to be a general style as I think it could be used at settings page too. 

Will make this change soon :)

www/md_base/src/app/common/directives/inspectdata/inspectdata.tpl.jade
shanzi: Although saying that, I will try to implement what you said and see if it is workable with small modification.

shanzi:
Please have a look at current animation effect of expanding and collapsing the field at http://buildbot.apps.io-meter.com/#/about 

Is it good enough? May be no animation is better? Add animation for different fields that have different height is a little complex.

www/md_base/src/app/about/about.tpl.jade
```diff
@@ -1 +1,37 @@
-h1 About
+section.card(layout-md="column", layout-gt-md="row")
+  div.meta(flex-gt-md="33")
+    h2 Runtime informations
+    p
+      | Informations, configrations and statuses about this running instance of buildbot.
```
sa2ajj: configurations

www/md_base/src/app/about/about.tpl.jade
```diff
@@ -1 +1,37 @@
...
+  div.content(flex-gt-md="66")
+    h3 Project Info
+    inspect-data(data="about.projectInfo")
+    h3 Version Info
+    inspect-data(data="about.versions")
+    h3 Configrations
```
sa2ajj: Configurations

sa2ajj:
It'd be great for a native English speaker to go through the text.

www/md_base/src/app/about/about.tpl.jade
jaredgrubb: /Information/  ... I dont think you can pluralize that word.

www/md_base/src/app/common/directives/inspectdata/inspectdata.spec.coffee
```diff
@@ -0,0 +1,103 @@
...
+
+    $rootScope = $compile = null
+
+    injected = ($injector) ->
+        $compile = $injector.get('$compile')
+        $rootScope = $injector.get('$rootScope')
+
+    beforeEach inject injected
+
+    it 'should inspect data correctly', ->
```
tardyp: nornally you should test only one thing for each test.
you can factorize in beforeeach the initialization of the directive, and then test each feature in a different 'it'

www/md_base/src/app/common/directives/inspectdata/inspectdata.tpl.jade
tardyp: I think html complexity is no argument for directives. I indeed really would like you try with the sticker

shanzi:
Just removed the animation and added arrows to expand and collapse the field

arrows

shanzi:
Latest code has deployed to my live demo site: http://buildbot.apps.io-meter.com

The last part of the about page is displaying API spec, this will require the presence of `buildbotService`. This service is seems to be rewrite. Should I add the old one in push forward with it now? Or we wait for the new service code is ready?

tardyp:
[nit] Is there a way we could expand *after* the rotation?

Looks good anyway.

Yes, you'll need to take the buildbotService as-is waiting for andras to have something workable. I dont believe it would be hard to switch api. The new api should be similar principle, just done right.

shanzi:
Displaying api references:

api-references

www/md_base/src/app/common/services/buildbot/buildbot.service.coffee
```diff
@@ -0,0 +1,203 @@
+class BuildbotService extends Factory
```
tardyp: I would prefer that you symlink those files. As you have no intention to modify them for md_base, I would prefer to avoid duplication.

www/md_base/src/app/about/about.tpl.jade
```diff
@@ -1 +1,53 @@
...
+  div.meta(flex-gt-md="33")
+    h2 API references
+    p
+      | References and description of the API endpoints that provided by this application.
+  div.content
+    div(ng-if="!about.specs.length") Please wait...
+    div.api(ng-if="about.specs.length")
+      dl(ng-repeat="spec in about.specs|orderBy:'path'", ng-init="collapsed = 1")
+        dt
+          a(ng-click="collapsed = !collapsed", href="") /{{ spec.path }}
```
tardyp: please always add .no-select when you add ng-click on a text thing. else this mess up with double click

www/md_base/src/styles/styles.less
```diff
@@ -4,3 +4,7 @@
 [ng-cloak] {
   display: none !important;
 }
+
+a {
+  text-decoration: none;
```
tardyp: does not look right to remove text-decoration for all As

www/md_base/src/app/about/about.tpl.jade
shanzi: `.no-select`? Is this an existing css class?

www/md_base/src/styles/styles.less
shanzi: hmmm... many websites like Google has removed underline for all links these days. So I tried removing them and found the appearance is good. Maybe we can display underline only when `hover`?

tardyp: We need to differenciate clickable from unclickable.
image

here, "API Reference" is not clickable, "/application.spec" is clickable, but are in the same color, its hard to make a difference.

www/md_base/src/app/about/about.tpl.jade
tardyp: its in styles.less
```
.no-select {
  -webkit-touch-callout: none;
  -webkit-user-select: none;
  -khtml-user-select: none;
  -moz-user-select: none;
  -ms-user-select: none;
  user-select: none;
}
```

www/md_base/src/app/common/services/buildbot/buildbot.service.coffee
shanzi: It seems that gulp cannot follow symlinks for folders when compiling, unless we make symlinks for the files one by one. It makes the things a little complex.

shanzi:
@tardyp
Done. Now links have underlines.

In my opinion, it looks less good than that without underline, the clickable indicating may not be that important as the mouse cursor will change on clickable elements after all. Maybe we should override the theme's default link color of angular-material to blue?

shanzi:
@tardyp 
Any more comments? If not, I think this PR is ready to be merged :)

tardyp:
ok. I was waiting for you to remove the [WIP] tag.

shanzi:
@tardyp :)

Have you received my emails about Hangouts? Maybe we should make sure we can access to the other through that.

tardyp:
I have tried to run the thing locally, and I see the following error in js console as well as in unit tests:
```
Uncaught Error: [$injector:modulerr] Failed to instantiate module app due to:
Error: [$injector:unpr] Unknown provider: bbSettingsServiceProvider
http://errors.angularjs.org/1.3.15/$injector/unpr?p0=bbSettingsServiceProvider
    at http://localhost:5000/scripts.js?_1430308546951:1:27804
    at http://localhost:5000/scripts.js?_1430308546951:2:13850
```

Are you sure you rebuilt?

shanzi:
@tardpy this maybe a problem of the symlink as the settings service in md is symlinked to the old files. It seems ok here.

tardyp:
ok, the problem was with the last minute symlinks. I fixed them, and verified it builds
I also changed the make prebuilt not to build md_base.

PR #1640, opened 2015-04-23
Git: Use RC_SUCCESS instead of 0

jaredgrubb:
I'm not a fan of magic numbers in code ... I have some patches in the Git module coming up and this is just the start of that. 

I went through and replaced any use of "0", where it represented an exit code, and replaced it with RC_SUCCESS. There were two other small magic numbers I think look better as a constant.

This patch is a non-functional change.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -23,6 +23,10 @@ from buildbot.interfaces import BuildSlaveTooOldError
 from buildbot.process import buildstep
 from buildbot.steps.source.base import Source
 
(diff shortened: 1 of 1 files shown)


sa2ajj:
```
master/buildbot/steps/source/git.py:30:1: E302 expected 2 blank lines, found 1
def isTrueOrIsExactlyZero(v):
^
1       E302 expected 2 blank lines, found 1
```

jaredgrubb:
Fixed and updated.

sa2ajj:
Thanks!

PR #1643, opened 2015-04-26
Git fixes

wentasah:
Hi,

these are two fixes for problems that I experience with git submodules.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -125,6 +125,7 @@ class Git(Source):
         self.getDescription = getDescription
         self.config = config
         self.supportsBranch = True
+        self.supportsSubmoduleCheckout = True
         self.srcdir = 'source'
         Source.__init__(self, **kwargs)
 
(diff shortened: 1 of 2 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -534,14 +535,16 @@ def _syncSubmodule(self, _=None):
 
     def _updateSubmodule(self, _=None):
         if self.submodules:
-            return self._dovccmd(['submodule', 'update',
-                                  '--init', '--recursive', '--force'])
+            vccmd = ['submodule', 'update', '--init', '--recursive', '--force']
```
tardyp: looks like this one is not unit tested. We would apreciate if you add a unit test for the submodule use case.

wentasah: On Mon, Apr 27 2015, Pierre Tardy wrote:
>> @@ -534,14 +535,16 @@ def _syncSubmodule(self, _=None):
>>  
>>      def _updateSubmodule(self, _=None):
>>          if self.submodules:
>> -            return self._dovccmd(['submodule', 'update',
>> -                                  '--init', '--recursive', '--force'])
>> +            vccmd = ['submodule', 'update', '--init', '--recursive', '--force']
>
> looks like this one is not unit tested. We would apreciate if you add
> a unit test for the submodule use case.

It is - see https://github.com/wentasah/buildbot/commit/198f44f7241d3ec72a418bd9cca0b29aab3b488d

Or do you have in mind some other test?

-Michal

wentasah: Oh, you probably mean a test for pre-1.7.8 behavior. I'll add that.

wentasah:
Is it better now? (https://github.com/wentasah/buildbot/commit/a310e30afab77fd6b04c9f728bf73ff7d8f4d847)

tardyp:
:+1:

PR #1644, opened 2015-04-28
Dev/api help urlencoding

catharsis:
We have a lot of builders with slashes in their names. This PR contains commits that makes links to such builders work in the web status/JSON API/help sections.

### DIFF
--- a/master/buildbot/status/web/status_json.py
+++ b/master/buildbot/status/web/status_json.py
@@ -353,7 +353,7 @@ class HelpResource(HtmlResource):
     def content(self, request, cxt):
         cxt['level'] = self.parent_level
         cxt['text'] = ToHtml(self.text)
-        cxt['children'] = [n for n in self.parent_children if n != 'help']
+        cxt['children'] = [(urllib.quote(n, ''), n) for n in self.parent_children if n != 'help']
         cxt['flags'] = ToHtml(FLAGS)
         cxt['examples'] = ToHtml(EXAMPLES).replace(
             'href="/json',
--- a/master/buildbot/status/web/status_json.py
+++ b/master/buildbot/status/web/status_json.py
@@ -750,7 +750,7 @@ For help on any sub directory, use url /child/help
                 break
         if not builder:
             return
-        EXAMPLES = EXAMPLES.replace('<A_BUILDER>', builder.getName())
+        EXAMPLES = EXAMPLES.replace('<A_BUILDER>', urllib.quote(builder.getName(), ''))
         build = builder.getBuild(-1)
         if build:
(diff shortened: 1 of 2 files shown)


sa2ajj:
I'd appreciate a test case or two for this PR.

Otherwise, LGTM

catharsis:
Certainly, I'll look into it during the week. Any suggestions on a suite where it might fit?

krf:
Hey, I just saw this PR by incidence. This change would actually make this PR here somewhat redundant: https://github.com/buildbot/buildbot/pull/1651. Gerrit project names contain slashes, too, and Buildbot's behavior with project names with slashes wasn't perfect. Thus, I fixed the issue by translating the Gerrit project names into something more suitable for Buildbot.

Can you confirm that the rest of Buildbot works fine with project names containing slashes?

tardyp:
I myself cannot confirm, but I think this is a good idea to make sure it does.

In buildbot nine, we are using either ids, or slug fore referencing stuff in URLs. "slug" is like a name, but cleaned-up from all stuff that would force urlencode, because urlencode is ugly.

http://en.wikipedia.org/wiki/Semantic_URL#Slug

catharsis:
Well... I'd be lying if I said we haven't had problems like this before (this isn't my first patch). But on the whole, our build system, sporting a couple of thousand builders works like a charm. The things that do break tend to be far from mission critical, ime (like this PR, dealing with links in the help text of the JSON API).

We do have a couple of local patches applied to deal with a few of the similar bugs we've found that I don't think have made their way upstream. If I recall correctly, that's because they're fixed or unapplicable in nine.

From what I gather, however, eight is supposed to be a rather long-lived project with its own roadmap, so I'd be happy to go through the ones that we haven't pushed yet and make sure to do so if that's desirable.

catharsis:
Oh, and for the sake of reference, I created https://github.com/mitsuhiko/jinja2/issues/444 apropos jinja2's urlencode filter mentioned in one of the commit messages.

djmitche:
@sa2ajj is this good to merge to eight?  It seems tests are not forthcoming..

sa2ajj:
I'll review right after the meeting

sa2ajj:
Looks reasonable for inclusion.  I'll test it on my installation.

djmitche:
@sa2ajj how did the tests work?

sa2ajj:
@djmitche damn, I totally forgot about it.

Sending a reminder to self at office.

sa2ajj:
sorry, a busy week. tomorrow is going to be quiet at office.

sa2ajj:
Sorry, it took so long for me.

Thanks for the contribution!

PR #1646, opened 2015-04-30
MD: Settings page

shanzi:
Here is the settings page. Based on a similar style to about page, this page does not need much coding.

settings-preview

settings-preview-thin

**NOTE**: As other pages that providing settings have not been written yet, so I added a testing settings group to preview the effect. If merge with this testing group is acceptable? Please review and give your advices.

### DIFF
--- a/www/base/src/app/common/services/mq/mq.service.coffee
+++ b/www/base/src/app/common/services/mq/mq.service.coffee
@@ -146,6 +146,7 @@ class State extends Config
         bbSettingsServiceProvider.addSettingsGroup
             name:'Websockets'
             caption: 'Websocket'
+            description: 'Websocket related settings'
             items:[
                 type:'bool'
                 name:'debug_websockets'
--- a/www/md_base/src/app/index.jade
+++ b/www/md_base/src/app/index.jade
@@ -17,7 +17,7 @@ html(ng-app="app", ng-controller="appController as app")
       div(layout="column", role="main", flex)
         md-toolbar.topbar
           div.md-toolbar-tools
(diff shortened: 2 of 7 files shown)


shanzi:
live demo updated

tardyp:
pretty cool!
waiting for @tothandras +1

jaredgrubb:
Definitely cool :)

tothandras:
Really nice! :)
I would keep the styles used by the view in the same folder, and when other view starts to use it then move to the shared folder.

shanzi:
@tothandras you mean put the `setting-` related styles in a separated file under `/settings` folder?
These styles seems to be to little to be put in standalone file. Hmmmm....

tothandras:
I think it's not about the size. If I would like to modify the styles related to settings, the card.less will be the last where I would look for it.

shanzi:
@tothandras OK.

Which do you think is better? Put them in `/app/settings/settings.less` or `/styles/settings.less`?

tothandras:
I prefer /app/settings/settings.less

tardyp:
Le sam. 2 mai 2015 à 13:11, András Tóth <notifications@github.com> a écrit :

> I prefer /app/settings/settings.less
>

Agreed. We should document this in coffee-style.rst


>

tardyp:
actually, card.less is already shared, and make sense. I'll merge this tomorrow, if no other blockers

shanzi:
I think his pr is ready to be merged :)

sa2ajj:
Thanks!

PR #1647, opened 2015-04-30
Add toast notification exception interceptor

tothandras:
Show errors and uncatched exceptions (eg. `throw 'message'`) in a toast notification.

image

### DIFF
--- a/www/md_base/src/app/error/error.less
+++ b/www/md_base/src/app/error/error.less
@@ -1,0 +1,8 @@
+md-toast.error {
+  background-color: #f44336;
+  color: #ffffff;
+
+  p {
+    margin-right: 15px;
+  }
+}
--- a/www/md_base/src/app/error/error.controller.coffee
+++ b/www/md_base/src/app/error/error.controller.coffee
@@ -1,0 +1,5 @@
+class Error extends Controller
+    constructor: (@$mdToast, @message) ->
+
+    closeToast: ->
+        @$mdToast.hide()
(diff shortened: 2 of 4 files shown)


sa2ajj:
I'm not sure I follow the example: I think the things presented in the example should not be shown to the user.

tothandras:
The users will never see this type of DI errors, only devs.

www/md_base/src/app/error/error.less
```diff
@@ -0,0 +1,8 @@
+md-toast.error {
+  background-color: #f44336;
```
tardyp: I would prefer buildbot's red, which is less shiny than your screen shot

tardyp:
I think we should extend this to exceptions, and also put it as an option in the setting page.

For DI, I dont know how this can work, as angularjs will not be instanciated properly if there is a DI exception, so you'll endup with blankscreen anyway.

www/md_base/src/app/error/error.less
sa2ajj: In light of users vs developers, their reds should be different: developers would want to know went south vs when things just fail because of circumstances...

tothandras:
@tardyp this decorates, extends the built in exception handler, it catches all uncatched exceptions. I chose this red from the material color palette, I can change it for something else (currently #EE8888 is the failure color, or we can choose new colors for md).
@shanzi any comments?

shanzi:
hmmm, why we need this other than just inspecting the error in a brower's debug console?

I agree tha the color is too bright. Consider pick a color from the deep pink color palette of material design

yetanotherion:
:+1:

sa2ajj:
how do we proceed with this PR?

tardyp:
This is for md_base so lets go ahead

tothandras:
oh, I forgot to change the color :/ 
@shanzi please choose change it to something else you like

shanzi:
@tothandras Never mind. Current appearance is also ok.

PR #1649, opened 2015-05-02
Git: inlineCallback updates

jaredgrubb:
Two things here:
 * move a couple functions to inlineCallbacks
 * remove "fetchcount", which is not used at all in this file

This should be a non-functional change, and passes tests, pylint, pep8. Will backport to eight as well.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -608,7 +608,7 @@ class Git(Source):
         files = cmd.updates['files'][0]
         if '.git' in files:
             defer.returnValue("update")
-        elif len(files) > 0:
+        elif files:
             defer.returnValue("clobber")
         else:
             defer.returnValue("clone")
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -122,7 +122,6 @@ class Git(Source):
         self.retryFetch = retryFetch
         self.submodules = submodules
         self.shallow = shallow
-        self.fetchcount = 0
         self.clobberOnFailure = clobberOnFailure
         self.mode = mode
         self.getDescription = getDescription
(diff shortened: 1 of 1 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -587,34 +584,31 @@ def applyAlready(res):
...
 
-            @d.addCallback
-            def checkWithListdir(_):
-                if 'files' not in cmd.updates:
-                    # no files - directory doesn't exist
-                    return "clone"
-                files = cmd.updates['files'][0]
-                if '.git' in files:
-                    return "update"
-                elif len(files) > 0:
```
sa2ajj: I like that this is removed :)

master/buildbot/steps/source/git.py
```diff
@@ -587,34 +584,31 @@ def applyAlready(res):
...
+        cmd.useLog(self.stdio_log, False)
+        yield self.runCommand(cmd)
+
+        if 'files' not in cmd.updates:
+            # no files - directory doesn't exist
+            defer.returnValue("clone")
+        files = cmd.updates['files'][0]
+        if '.git' in files:
+            defer.returnValue("update")
+        elif len(files) > 0:
```
sa2ajj: However I do not like that this is added :)

```
elif files:
```

should be enough, no?

jaredgrubb: I take no credit for what's there, it's actually just de-dented, the diff isnt showing it well... but yeh that could certainly just be "elif files"

sa2ajj: I understand that, though I see no reason to take a credit for improving the code :)

jaredgrubb: Updated :)

sa2ajj:
LGTM

tardyp:
:+1:

PR #1650, opened 2015-05-03
RFC: shell.py: Also match on errors from stdio

krf:
This is a almost 1:1 copy from the patch attached to the bug report.
Applied onto current code base.

Partly fixes #339

TODO: Rename class?

### DIFF
--- a/master/buildbot/steps/shell.py
+++ b/master/buildbot/steps/shell.py
@@ -396,6 +396,10 @@ class WarningCountingShellCommand(ShellCommand):
 
     warnCount = 0
     warningPattern = '.*warning[: ].*'
+
+    errorCount = 0
+    errorPattern = '.*error[: ].*'
+
     # The defaults work for GNU Make.
     directoryEnterPattern = (u"make.*: Entering directory "
                              u"[\u2019\"`'](.*)[\u2019'`\"]")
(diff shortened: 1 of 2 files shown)


jaredgrubb:
Looks ok to me! It would be nice if there were tests for the new error stuff.

tardyp:
Looks good. I think we could abstract the concept a little bit more, and create PatternCountingShellCommand(patterns=dict(errors="Error:?.+", warnings="warning:?.*")

you would have to rename the class, and create a WarningCountingShellCommand as a preconfigured subclass (to keep compatibility)

sa2ajj:
This does not seem to cleanly apply to nine.  Maybe it's better to first submit it to nine and then backport to eight?

(@sa2ajj looks at the current set of changes in eight and is a bit concerned...)

yetanotherion:
:+1:

sa2ajj:
@yetanotherion I agree, but I'd like to see it in master first.

jaredgrubb:
Are we just waiting on a nine pull-reqeuest to satisfy this?

djmitche:
Yes, otherwise we'll be adding functionality to eight that's not available in nine.

krf:
Note: Just checked. Simply cherry-picking this to nine isn't possible; code has changed to much in the meanwhile. I won't have time to port this to Nine (no way to test either, as we don't use Nine yet).

djmitche:
OK, I'm going to close this up per http://trac.buildbot.net/wiki/StalePullRequestPolicy.  Of course, the code is still here if someone wants to forward-port to nine.  And this can be cherry-picked into an eight installation.

PR #1651, opened 2015-05-03
RFC: GerritChangeSource: Translate project names

krf:
There are cases where GerritChangeSource returns project names such as
"foo/bar". We need to be able to translate these project names into ones
that Buildbot can handle.

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -22,6 +22,9 @@ from twisted.internet import reactor
 from twisted.internet.protocol import ProcessProtocol
 from twisted.python import log
 
+def defaultProjectNameTranslator(gerritProjectName):
+    """Just return the last part of the project if it contains slashes"""
+    return gerritProjectName.rpartition('/')[-1]
 
 class GerritChangeFilter(ChangeFilter):
 
@@ -81,7 +84,8 @@ class GerritChangeSource(base.ChangeSource):
                  username,
                  gerritport=29418,
                  identity_file=None,
-                 handled_events=("patchset-created", "ref-updated")):
+                 handled_events=("patchset-created", "ref-updated"),
+                 projectNameTranslator=defaultProjectNameTranslator):
         """
         @type  gerritserver: string
         @param gerritserver: the dns or ip that host the gerrit ssh server,
@@ -97,6 +101,9 @@ class GerritChangeSource(base.ChangeSource):
 
         @type  handled_events: list
         @param handled_events: event to be handled (optional).
+
+        @type  projectNameTranslator: func
+        @param projectNameTranslator: Function translating a Gerrit project name to the Buildbot project name
         """
         # TODO: delete API comment when documented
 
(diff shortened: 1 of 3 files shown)


jaredgrubb:
A unit test testing the new functionality would be good.
Also a release-notes comment about the change.

master/buildbot/changes/gerritchangesource.py
```diff
@@ -22,6 +22,9 @@
 from twisted.internet.protocol import ProcessProtocol
 from twisted.python import log
 
+def defaultProjectNameTranslator(gerritProjectName):
```
tardyp: I think we should rather encourage subclassing rather than passing function. This might not be a shared opinion, as there are tons of callback configurability in buildbot. I really dislike those.

master/buildbot/changes/gerritchangesource.py
```diff
@@ -22,6 +22,9 @@
...
+    """Just return the last part of the project if it contains slashes"""
+    return gerritProjectName.rpartition('/')[-1]
```
tardyp: should we change default behaviour? Or put this as an optional basenameProjectNameTranslator()

tardyp:
even more important than the release note, please add documentation with example.

sa2ajj:
PEP-8 error:

```
pep8 --config=common/pep8rc master slave
master/buildbot/changes/gerritchangesource.py:25:1: E302 expected 2 blank lines, found 1
def defaultProjectNameTranslator(gerritProjectName):
^
master/buildbot/changes/gerritchangesource.py:29:1: E302 expected 2 blank lines, found 1
class GerritChangeFilter(ChangeFilter):
^
2       E302 expected 2 blank lines, found 1
2
make: *** [pep8] Error 1
```

Other things to do:
[ ] documentation with example
[ ] unit test case(s) for the new functionality

Stibbons:
I would recommend to use full path, just replacing "/" by "_" for example. So you won't have collision between "a/foo" and "another/foo"

krf:
Well, I'm aware of the possibility there might be clashes. However, I think just using the last part is probably what the user would expect here. I'm happy to be convinced otherwise, though.

djmitche:
Nine should be able to handle slashes properly, so I think this is moot there.  @sa2ajj do you want this for eight?

sa2ajj:
Well, it can't be merged as it is.

I'll try to experiment during this week with project names with slashes and see what kind of bad behaviour I'll observe (fortunately, I have a Gerrit instance :))

krf:
Thanks for looking into it. I'm really sorry, but I have very limited amount of time scheduled for helping out upstreaming this kind of patches.

Personally I'd be fine if we just wait for Nine to have proper support for project names with slashes. No need to merge this to Eight from my side. This patch is more a work-around than anything else unfortunately.

krf:
@sa2ajj: Feel free to close this if nine has proper support for project names with slashes.

sa2ajj:
@krf do you think you could rebase it before i start feeling? :wink:

krf:
Done

PR #1654, opened 2015-05-06
Add a proxy to use the API entries of nine.buildbot.net

shanzi:
I'm now starting to develop for Home page, and we need some data like build history to be displayed on the home page to preview the appearance. As it is not that convenient to fill test data to my server, a proxy is added to forward all API request to http://nine.buildbot.net , which has many data available.

This idea comes from @tardyp , thanks for this good idea! 

I believe both HTTP request and Websocket is able to be forwarded with `http-proxy`. Perform `npm install` before you run `gulp dev` next time.

Modifications of this commit is expected to be removed after the development of our new front end finished.

This pull request is intended to draw comments and suggestions. And @tothandras's project may also need this function.

### DIFF
--- a/master/buildbot/test/unit/test_scripts_processwwwindex.py
+++ b/master/buildbot/test/unit/test_scripts_processwwwindex.py
@@ -28,6 +28,7 @@ from buildbot.www.service import WWWService
 class TestUsersClient(unittest.TestCase):
 
     def setUp(self):
+        # un-do the effects of @in_reactor
         self.patch(processwwwindex, 'processwwwindex', processwwwindex.processwwwindex._orig)
 
     def test_no_input_file(self):
--- a/master/buildbot/test/unit/test_scripts_processwwwindex.py
+++ b/master/buildbot/test/unit/test_scripts_processwwwindex.py
@@ -20,9 +20,6 @@ from twisted.trial import unittest
 
 from buildbot.util import json
 from buildbot.scripts import processwwwindex
-from buildbot.test.fake import fakemaster
-from buildbot.www.config import IndexResource
-from buildbot.www.service import WWWService
 
 
 class TestUsersClient(unittest.TestCase):
@@ -58,25 +55,7 @@ class TestUsersClient(unittest.TestCase):
             self.assertEqual(ret, 0)
             with open(tmpf.name) as f:
                 config = json.loads(f.read())
-                config['versions'] = [(v[0], v[1]) for v in config['versions']]
-                correct = self.get_correct_config()
-                self.assertEqual(config, correct)
+                self.assertTrue(isinstance(config, dict))
 
         d.addCallback(check)
         return d
-
-    def get_correct_config(self):
-        master = fakemaster.make_master()
-        master_service = WWWService(master)
-
-        plugins = dict((k, {}) for k in master_service.apps.names if k != "base")
-
-        fakeconfig = {"user": {"anonymous": True}}
-        fakeconfig['buildbotURL'] = master.config.buildbotURL
-        fakeconfig['title'] = master.config.title
-        fakeconfig['titleURL'] = master.config.titleURL
-        fakeconfig['multiMaster'] = master.config.multiMaster
-        fakeconfig['versions'] = IndexResource.getEnvironmentVersions()
-        fakeconfig['plugins'] = plugins
-
-        return fakeconfig
--- a/master/buildbot/test/unit/test_scripts_processwwwindex.py
+++ b/master/buildbot/test/unit/test_scripts_processwwwindex.py
@@ -1,0 +1,81 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+
+import tempfile
+
+from twisted.trial import unittest
+
+from buildbot.util import json
+from buildbot.scripts import processwwwindex
+from buildbot.test.fake import fakemaster
+from buildbot.www.config import IndexResource
+from buildbot.www.service import WWWService
+
+
+class TestUsersClient(unittest.TestCase):
+
+    def setUp(self):
+        self.patch(processwwwindex, 'processwwwindex', processwwwindex.processwwwindex._orig)
+
+    def test_no_input_file(self):
+        d = processwwwindex.processwwwindex({})
+
+        def check(ret):
+            self.assertEqual(ret, 1)
+        d.addCallback(check)
+        return d
+
+    def test_invalid_input_file(self):
+        d = processwwwindex.processwwwindex({'index-file': '/some/no/where'})
+
+        def check(ret):
+            self.assertEqual(ret, 2)
+        d.addCallback(check)
+        return d
+
+    def test_output_config(self):
+        tmpf = tempfile.NamedTemporaryFile(suffix=".html")
+        with open(tmpf.name, 'w') as f:
+            f.write('{{ configjson|safe }}')
+
+        d = processwwwindex.processwwwindex({'index-file': tmpf.name})
+
+        def check(ret):
+            self.assertEqual(ret, 0)
+            with open(tmpf.name) as f:
+                config = json.loads(f.read())
+                config['versions'] = [(v[0], v[1]) for v in config['versions']]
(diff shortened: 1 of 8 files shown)


www/md_base/guanlecoja/config.coffee
```diff
@@ -88,4 +88,16 @@ gulp.task 'icons', ->
         ))
         .pipe(gulp.dest(path.join(config.dir.build, 'icons')))
 
+gulp.task 'apiproxy', ->
+    # this is a task for developing, it proxy api request to http://nine.buildbot.net
+    httpProxy = require 'http-proxy'
+    proxy = httpProxy.createProxyServer({target: 'http://nine.buildbot.net'}).listen(8021)
```
tardyp: Please put nine.buildbot.net as a option.

```
argv = require('minimist')(process.argv.slice(2))

argv.apiserver ?= 'http://nine.buildbot.net'
```

www/md_base/src/app/common/common.constant.coffee
```diff
@@ -5,17 +5,27 @@ invert_constant = (constant_name, inverted_constant_name) ->
 
 class Baseurlapi extends Constant('common')
     constructor: ->
-        return 'api/v2/'
```
tardyp: I would prefer you can make a better proxy solution that would not involve such hack.

With proxy, you are supposed to serve all the static frontend thing directly, and only proxy api/ sse/ and ws/

So your workflow is supposed to use http://localhost:8021/ as the test server, and you don't lunch any python server

This is problematic with the detection of plugins.
We have chat with andras about that yesterday evening, please synchronise.

We should find a long term solution for this problem. This is not only for development of this feature. For all frontend development, proxy is a very usefull feature. It has to be resolved with backend solution only not front-end hacks. The backend solution might eventually involve launching a python script that will report a json file with the plugin configuration available.

shanzi: But some resources such as HTML template seems won't easily served without
python server. hmmm...
Give me some time to study on this.

shanzi: @tardyp The biggest challenge is that we cannot get the `configjson` without python backend.

shanzi:
@tardyp OK now, I added local python server at `localhost:8020` as a proxy backend too. So we can have the pages rendered by jinja2 with api proxied to other place at the same time without any hack now.

You can access the proxied page at `localhost:8021`. `localhost:8020` will be left to be the original page without build history data.

It is not possible to run the front end independently by now. Maybe in the future we may make it by removing Jinja2 rendering step.

shanzi:
@tardyp Would you like to make the front end totally independent now? This may concerns about modification of `master` and will break the function of the old `www/base`. I suggest that we put it aside as one of tasks in the end of finishing `md_base`.

Though current solution need the python server running in background, I believe it is good enough with out any hacking on path constants.

www/md_base/guanlecoja/config.coffee
```diff
@@ -88,4 +88,33 @@ gulp.task 'icons', ->
...
+        proxyRes.headers['Access-Control-Allow-Origin'] = '*'
+        console.log "[Proxy] #{req.method} #{req.url}"
+
+    server = http.createServer (req, res) ->
+        if req.url.match /^\/api/
+            proxy.web req, res, {target: 'http://' + argv.apiserver}
+        else if req.url.match /^\/ws/
+            proxy.ws req, res, {target: 'ws://' + argv.apiserver}
+        else
+            proxy.web req, res, {target: argv.server}
```
tardyp: I think the serve could serve directly the static files here. This will simplify everything.

You can add a script called

buildbot getwwwplugins
that would output such json as:
{
   "base": "/path/to/buildbot/www/md-base/static"
   "waterfall": "/path/to/buildbot/www/waterfall/static"
}

You can look at : 
https://github.com/buildbot/buildbot/blob/master/master/buildbot/scripts/runner.py#L634
https://github.com/buildbot/buildbot/blob/master/master/buildbot/scripts/dataspec.py
on how to create a new buildbot script entrypoint.

You can look at:
https://github.com/buildbot/buildbot/blob/master/master/contrib/api_proxy.py#L22
to see how to get the list of plugins.

shanzi: @tardyp 

I didn't quite get your idea. You mean using a script command to generate json of plugins and `configjson` and embed them into the page like what we do to `window.dataspec`?

It works if we are compiling the front end with the backend we are using, but it may not work when we'd like to precompile them into wheels. And how can we get the runtime `configjson`? Doesn't it need to modify the backend to provide an API entry?

BTW: how to specify `apiproxy` as a dev-only task in `config.coffee` to avoid the test failed due to this task will cause a timeout as the server will always running in the background.

shanzi: @tardyp Could you provide more explanation?

tardyp: You perferctly rephrased what I meant.

The only thing is that we should require this only for the dev mode. The proxy server shall not be even launched for default build, and prod build.
we could also add a proxy task that is not a dependancy of dev.

gulp dev proxy --url http://nine.buildbot.net/

This will address your concerns I think.

www/md_base/guanlecoja/config.coffee
```diff
@@ -88,4 +88,33 @@ gulp.task 'icons', ->
         ))
         .pipe(gulp.dest(path.join(config.dir.build, 'icons')))
 
+gulp.task 'apiproxy', ->
+    # this is a task for developing, it proxy api request to http://nine.buildbot.net
+    argv = require('minimist')(process.argv)
+    argv.apiserver ?= 'nine.buildbot.net'
```
Stibbons: are you sure you want to hardcode this url?
one might also want to use its own server on localhost

shanzi: `?=` operator makes sure that 'nine.buildbot.net` only take place when the corresponding option is undefined.

shanzi:
@tardyp Done, please have a look at it :)

A unit test for the new script command might be needed.

www/md_base/guanlecoja/config.coffee
Stibbons: ok

www/md_base/guanlecoja/config.coffee
```diff
@@ -88,4 +88,47 @@ gulp.task 'icons', ->
...
+    proxy.on 'proxyRes', (proxyRes, req, res) ->
+        proxyRes.headers['Access-Control-Allow-Origin'] = '*'
+        console.log "[Proxy] #{req.method} #{req.url}"
+
+    server = http.createServer (req, res) ->
+        if req.url.match /^\/(api|sse)/
+            proxy.web req, res, {target: 'http://' + argv.host}
+        else if req.url.match /^\/ws/
+            proxy.ws req, res, {target: 'ws://' + argv.host}
+        else
```
tardyp: Great! But you dont fully manage the plugins. Each plugin have it's own directory of static files with one script and one CSS file. You need to tell the proxy where the file are in order to serve them.

shanzi:
@tardyp I symlinked the files. But currently the old plugins such as `waterfall_view` and `console_view` can not be loaded by `md_base` because they are depending on `guanlecoja-ui`, which is not added in the new base.

I suggest we put this problem aside and solve it after `md_base` is basically finished.

tardyp:
ok for me. I want to reuse this proxy concept for the guanlecoja-ui version :)

shanzi:
I added a test for the new script command. Please review before merge this :D

master/buildbot/test/unit/test_scripts_processwwwindex.py
```diff
@@ -0,0 +1,81 @@
...
+from buildbot.scripts import processwwwindex
+from buildbot.test.fake import fakemaster
+from buildbot.www.config import IndexResource
+from buildbot.www.service import WWWService
+
+
+class TestUsersClient(unittest.TestCase):
+
+    def setUp(self):
+        self.patch(processwwwindex, 'processwwwindex', processwwwindex.processwwwindex._orig)
```
tardyp: you should add a comment on what this does

master/buildbot/test/unit/test_scripts_processwwwindex.py
```diff
@@ -0,0 +1,81 @@
...
+                self.assertEqual(config, correct)
+
+        d.addCallback(check)
+        return d
+
+    def get_correct_config(self):
+        master = fakemaster.make_master()
+        master_service = WWWService(master)
+
+        plugins = dict((k, {}) for k in master_service.apps.names if k != "base")
```
tardyp: well.. you are testing by copy pasting the exact same code. does not look very worth it.

shanzi: hmmm. I didn't think of a better way to test it. Can you give some advice?

shanzi: I think the main goal of this test should be guarantee the command correctly replace the content of the file with that has been rendered by Jinja. And also make sure the rendered json is able to be interpreted. :)

tardyp: Yes. It should not crash is a good enough test when like this the content
is dependent of the installation

tardyp:
:+1: waiting for travis to merge

shanzi:
You can merge it now :)

PR #1656, opened 2015-05-07
stop

yetanotherion:
following the POC from tardyp's commit.

### DIFF
--- a/master/buildbot/data/buildrequests.py
+++ b/master/buildbot/data/buildrequests.py
@@ -18,6 +18,8 @@ from buildbot.data import types
 from buildbot.db.buildrequests import AlreadyClaimedError
 
 from buildbot.db.buildrequests import NotClaimedError
+from buildbot.status import results
+
 from twisted.internet import defer
 from twisted.internet import reactor
 
@@ -61,6 +63,39 @@ class BuildRequestEndpoint(Db2DataMixin, base.Endpoint):
         buildrequestid = kwargs.get('buildrequestid')
         return self.master.mq.startConsuming(callback,
                                              ('buildrequests', buildrequestid, None))
+
+    @defer.inlineCallbacks
+    def control(self, action, args, kwargs):
+        if action != "cancel":
+            raise ValueError("action: {} is not supported".format(action))
+        brid = kwargs['buildrequestid']
+        # first, try to claim the request; if this fails, then it's too late to
+        # cancel the build anyway
+        try:
+            b = yield self.master.db.buildrequests.claimBuildRequests(brids=[brid])
+        except AlreadyClaimedError:
+            # XXX race condition
+            # - After a buildrequest was claimed, and
+            # - Before creating a build,
+            # the claiming master still
+            # needs to do some processing, (send a message to the message queue,
+            # call maybeStartBuild on the related builder).
+            # In that case we won't have the related builds here. We don't have
+            # an alternative to letting them run without stopping them for now.
+            builds = yield self.master.data.get(("buildrequests", brid, "builds"))
+
+            # Don't call the data API here, as the buildrequests might have been
+            # taken by another master. We just send the stop message and forget about those.
+            for b in builds:
+                self.master.mq.produce(("control", "builds", str(b['buildid']), "stop"),
+                                       dict(reason=kwargs.get('reason')))
+            defer.returnValue(None)
+
+        # then complete it with 'CANCELLED'; this is the closest we can get to
+        # cancelling a request without running into trouble with dangling
+        # references.
+        yield self.master.data.updates.completeBuildRequests([brid],
+                                                             results.CANCELLED)
 
 
 class BuildRequestsEndpoint(Db2DataMixin, base.Endpoint):
--- a/master/buildbot/data/buildrequests.py
+++ b/master/buildbot/data/buildrequests.py
@@ -86,9 +86,10 @@ class BuildRequestEndpoint(Db2DataMixin, base.Endpoint):
 
             # Don't call the data API here, as the buildrequests might have been
             # taken by another master. We just send the stop message and forget about those.
+            mqKwargs = {'reason': kwargs.get('reason', 'no reason')}
             for b in builds:
                 self.master.mq.produce(("control", "builds", str(b['buildid']), "stop"),
-                                       dict(reason=kwargs.get('reason')))
+                                       mqKwargs)
             defer.returnValue(None)
 
         # then complete it with 'CANCELLED'; this is the closest we can get to
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -237,6 +237,7 @@ class Build(properties.PropertiesMixin):
         # the Deferred returned by this method.
 
         log.msg("%s.startBuild" % self)
+
         self.build_status = build_status
         # TODO: this will go away when build collapsing is implemented; until
         # then we just assign the bulid to the first buildrequest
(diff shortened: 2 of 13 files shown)


master/buildbot/data/buildrequests.py
```diff
@@ -62,6 +64,31 @@ def startConsuming(self, callback, options, kwargs):
...
+    def control(self, action, args, kwargs):
+        if action != "cancel":
+            raise ValueError("action: {} is not supported".format(action))
+        brid = kwargs['buildrequestid']
+        # first, try to claim the request; if this fails, then it's too late to
+        # cancel the build anyway
+        try:
+            b = yield self.master.db.buildrequests.claimBuildRequests(brids=[brid])
+        except AlreadyClaimedError:
+            builds = yield self.master.data.get(("buildrequests", brid, "builds"))
```
tardyp: There is a race condition here. The buildrequest might be claimed, but the build are not yet created. The claiming master has to find a slave, substanciate it, etc before creating the build.
As ee don't have a proper solution for this race condition, I would at least create a comment here stating this, and create a trac to follow up on this problem

master/buildbot/process/build.py
```diff
@@ -89,6 +90,7 @@ def __init__(self, requests):
         self.terminate = False
 
         self._acquiringLock = None
+        self.result = SUCCESS   # overall result, may downgrade after each step
```
tardyp: In this file,
self.results is the list of step results, and self.result is the final build result.

We use the word results in plural everywhere else to refer to a single integer result.

I am not sure if results in english is more apropriate than result to store the result integer. Can the native reviewers provided a hint here?

tardyp: @jaredgrubb @djmitche ?

master/buildbot/test/integration/test_stop_triggered_not_started.py
```diff
@@ -0,0 +1,97 @@
...
+
+from buildbot.test.util.integration import RunMasterBase
+from buildbot.status.results import CANCELLED
+
+# This integration test creates a master and slave environment,
+# with two builders and a trigger step linking them. the triggered build never ends
+# so that we can reliabily stop it recursively
+
+
+class TriggeringMaster(RunMasterBase):
```
tardyp: I would prefer you try to factorize those two tests in one. (maybe do a reconfig for the second test)

master/buildbot/steps/trigger.py
```diff
@@ -89,11 +89,20 @@ def __init__(self, schedulerNames=None, sourceStamp=None, sourceStamps=None,
...
-        #        (which was a bad idea in the first place)
-        # we should claim the self.brids, and CANCELLED them
-        # if they were already claimed, stop the associated builds via data api
-        # then the big deferredlist will automatically be called
+        # We claim the self.brids and CANCEL them.
+        # Sending cancel on a buildrequest through the data api,
+        # makes sure that if the requests were already claimed, i.e.
+        # the builds were started,
+        # a stop message will be sent on the messsage bus.
+        # However, even though the stop messages are sent, note that
```
tardyp: I dont see how this can happen. If waitForFinish is set to True, then the parent should wait for buildrequest finished event.

master/buildbot/process/build.py
djmitche: You're right that we usually use "results" and should keep that up.  It's not quite right in English, but better to be consistent with the rest of the codebase.

master/buildbot/steps/trigger.py
yetanotherion: My mistake, indeed those line of codes proove your point
        if self.waitForFinish:
            rclist = yield defer.DeferredList(dl, consumeErrors=1)
            # we were interrupted, don't bother update status
            if self.ended:
                defer.returnValue(CANCELLED)
I'll update the comments

yetanotherion:
@tardyp @djmitche comments addressed

master/buildbot/test/integration/test_stop_trigger.py
```diff
@@ -0,0 +1,148 @@
...
+                      slavenames=["local1"],
+                      factory=f),
+        BuilderConfig(name="triggered",
+                      slavenames=["local1"],
+                      factory=f2)]
+    return c
+
+
+def triggeredBuildIsNotCreated():
+    c = {}
```
tardyp: you can still factorize those two functions. only 2 lines are different if I read correctly

master/buildbot/test/util/integration.py
```diff
@@ -86,6 +103,11 @@ def setUp(self):
             s = LocalBuildSlave("local1", self.basedir, False)
         s.setServiceParent(m)
 
+    def setUp(self):
```
tardyp: good idea

yetanotherion:
@tardyp coment about test_stop_trigger factorization addressed

tardyp:
:+1:  Please avoid rebase for the review comments addressing. This helps the review (can only review new patches)

yetanotherion:
sorry for that, but a unit test failed in remove result.  preferred to integrate the fix in the related commit (2f4f24e) rather than modifying the git history (which seems like polluting it to me)

yetanotherion:
(fyil tests are now ok in my station, let's see what travis says)

yetanotherion:
Some errors/timeout seem to have made the build fail
(apt-get install nodejs and git fetch on github)
https://travis-ci.org/buildbot/buildbot/jobs/62226384
https://travis-ci.org/buildbot/buildbot/jobs/62226388

The others are ok, does someone have rights to rebuild these two only ?

yetanotherion:
:+1: from travis

master/buildbot/data/buildrequests.py
```diff
@@ -62,6 +64,39 @@ def startConsuming(self, callback, options, kwargs):
...
+            # needs to do some processing, (send a message to the message queue,
+            # call maybeStartBuild on the related builder).
+            # In that case we won't have the related builds here. We don't have
+            # an alternative to letting them run without stopping them for now.
+            builds = yield self.master.data.get(("buildrequests", brid, "builds"))
+
+            # Don't call the data API here, as the buildrequests might have been
+            # taken by another master. We just send the stop message and forget about those.
+            for b in builds:
+                self.master.mq.produce(("control", "builds", str(b['buildid']), "stop"),
```
nam4dev: Don't you need to set a default value while looking up for "reason" key into kwargs as you did below?

ie. dict(reason=kwargs.get('reason', 'no reason'))

master/buildbot/data/buildrequests.py
```diff
@@ -62,6 +64,39 @@ def startConsuming(self, callback, options, kwargs):
...
+            # - Before creating a build,
+            # the claiming master still
+            # needs to do some processing, (send a message to the message queue,
+            # call maybeStartBuild on the related builder).
+            # In that case we won't have the related builds here. We don't have
+            # an alternative to letting them run without stopping them for now.
+            builds = yield self.master.data.get(("buildrequests", brid, "builds"))
+
+            # Don't call the data API here, as the buildrequests might have been
+            # taken by another master. We just send the stop message and forget about those.
```
nam4dev: It might be faster, if you set your "self.master.mq.produce" kwargs dict before looping, avoiding to looking up into kwargs at each iteration

yetanotherion:
@nam4dev comments addressed

master/buildbot/process/build.py
```diff
@@ -533,22 +538,21 @@ def buildFinished(self, text, results):
         state.
 
         It takes two arguments which describe the overall build status:
-        text, results. 'results' is one of SUCCESS, WARNINGS, or FAILURE.
+        text, results. 'results' is one of the possible results (see buildbot.status.results).
```
bastien31: Somewhere in the code you use @type and @param to describe params, why not doing it here also?
With sphynx, you should be also able to do :ref:`buildbot.status.results`

bastien31:
:+1:

master/buildbot/process/build.py
yetanotherion: I actually don't know which doc generation tool is used in buildbot, (and thus
the @type @param are more aimed at being understood by the programmer reading the code rather than a programmer reading the doc). What do you think @tardyp  @djmitche @sa2ajj ?

tardyp:
:+1:

PR #1663, opened 2015-05-14
provide build itself for those who need it

sa2ajj:
Summary callback differs from the review build callback in lack of the
actual build instance: the former only gets particular information about
builds.

In some situations, the build itself is required to do some magic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,10 +48,6 @@ install:
   - (cd master; python setup.py develop)
   - (cd slave;  python setup.py develop)
   # mock is preinstalled on Travis
-  # txgithub requires Twisted >= 12.3.0
-  - pip install txgithub
-  # txrequests support only Python 2.6 and 2.7.
-  - pip install txrequests
 
   # Determine if current configuration is latest
   - |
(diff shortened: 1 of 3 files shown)


tardyp:
:+1: I can't see why travis has error.

sa2ajj:
Tests need to be reviewed

sa2ajj:
Obviously tests did not expect a new item in the dictionary.

sa2ajj:
This is interesting:

```
DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python.
  DeprecationWarning
```

sa2ajj:
This is a bit cheating...

Proper fix would be to run Python with `-W ignore` parameter, me thinks.

@tardyp, @djmitche your opinion?

tardyp:
python 2.6.9 was the last release of python and is 2 years old:
https://www.python.org/download/releases/2.6.9/

More and more python package will be dropping support for py2.6, so this will end-up more and more regression like this which are quite time consuming.

see answer from cryptography dev:
https://github.com/pyca/cryptography/commit/0857894306d26a655fe06326cd244b4e182c94bf#commitcomment-11225891

not installing txgithub for python 2.6 is a solution (which is not IS_LATEST).

tardyp:
I did not realize IS_LATEST is python 2.7 for eight. I'm porting this behaviour to nine.

PR #1665, opened 2015-05-16
Panel system for Home page's dashboard

shanzi:
Here is an elementary implementation of dashboard system on the homepage. The dashboard is based on several panel directives. The panel can be expanded or collapsed to half width and the order of the panel can be changed by drag and drop.

Currently, the expand and collapse state of the panel is recorded with `bbSettingsService`. I plan to record the order of panel to that service too, but it has not been done yet.

The original arrangement of panels looks like:

arrangement_of_panels

Click on the button at the top right of panel will toggle expanding state. Those expanding states can also be customized at the Settings page.

Please give your comments and suggestions on the design.

**TODO**

- [x] record panel order with `bbSettingsService`
- [x] tests for panel directive


www/md_base/guanlecoja/config.coffee
```diff
@@ -56,6 +56,9 @@ config =
             'underscore.string':
                 version: "~2.3.3"
                 files: 'lib/underscore.string.js'
+            'sortable.js':
```
tardyp: Could you explain why did you choose this lib over the other available?

www/md_base/src/app/common/directives/panel/panel.directive.coffee
```diff
@@ -0,0 +1,30 @@
...
+            scope:
+                title: '@'
+                isCollapsed: '='
+        }
+
+
+class _Panel extends Controller
+
+    constructor: (@$element, @$scope) ->
+        @$scope.$watch 'panel.isCollapsed', (=> @updateCollapse())
```
tardyp: no need for parens: 

    @$scope.$watch 'panel.isCollapsed', => @updateCollapse()

www/md_base/src/app/home/home.route.coffee
```diff
@@ -17,3 +17,24 @@ class State extends Config
             templateUrl: "views/#{name}.html"
             name: name
             url: '/'
+
+        bbSettingsServiceProvider.addSettingsGroup
+            name: 'home'
+            caption: 'Home'
+            description: 'Settings to customize the display of Home page'
+            items: [
```
tardyp: That is fine enough. Eventually, I would have liked a plugin system, where people could have developed their own widget for the home page.

www/md_base/src/icons/collapse-panel.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
tardyp: more svgs :(

tardyp:
looks good so far, I need to run it to see the look and feel. Can you point us to the demo server?

shanzi:
Live demo at: http://buildbot.apps.io-meter.com

www/md_base/src/icons/collapse-panel.svg
shanzi: These two SVGs are drawing by myself because I did't find other icon set
that provides the icons I think express the function well.

This might be an advantage of using SVG :)

www/md_base/guanlecoja/config.coffee
shanzi: This lib does not depend on jQuery, and it supports drag and sort on `div`s
with `float: left` even if they are all in different sizes. What's more, it
uses HTML5 drag API and works better than that doesn't.

It officially ships with an angular binding and best conforms to what I
need. That's why I finally chose it.

tardyp:
got a weird rendering on mac:
image

Is this a scrollbar?

shanzi:
@tardyp the scrolling bar might be a result of `overflow-y: scroll`. I did not expected that some browser will show it even it is not needed, it is a bug and I will find a way to solve it.

I am considering the feature of enabling users write their own panel. It seems is too complex and I think we'd better put that aside and come up with a usable home page first.

Persisting the order  of panels is expected to be done before this weekend. I need some time to consider more and I am a little out of form these days. So sorry the progress might be a little slow.

tardyp:
overflow-y: scroll forces the display of a scrollbar.
overflow-y: auto make it automatic.

tardyp:
overflow-y: scroll forces the display of a scrollbar.
overflow-y: auto make it automatic.

tardyp:
overflow-y: scroll forces the display of a scrollbar.
overflow-y: auto make it automatic.

shanzi:
@tardyp 

I finished a more flexible solution for the panels. The state of panels to show is saved in the settings by `bbSettingsService`, including the order and collapsed states. The content is displayed through loading by templates `ng-include` where the urls to the templates are set in the settings too.

I think this gives a basic framework for enabling user self-defined panel. May be we should write a separate panels service provider to handling this logic in the future.

Newly demo: http://buildbot.apps.io-meter.com/#/

shanzi:
test for `panel` is here.

Do you have any suggestions about the organization of panels related files? Would you like to put them all flat in `home/panels` or each in its own folder? Or put them some where else?

www/md_base/guanlecoja/config.coffee
tardyp: could you add this explaination in the source code?

www/md_base/src/app/common/directives/panel/panel.directive.coffee
tardyp: needs to be addressed

www/md_base/src/app/home/home.less
```diff
@@ -0,0 +1,17 @@
+.project-info {
```
tardyp: You should namespace those rules to .homepage

www/md_base/src/app/home/home.route.coffee
```diff
@@ -17,3 +17,33 @@ class State extends Config
...
+                type: 'bool'
+                name: 'lock_panels'
+                caption: 'Lock panels'
+                default_value: false
```
tardyp: I think this would be better to have it locked by default, and having a small button in order to unlock them directly in the home page.
ADF does it, and I find it pretty straigforward: http://sdorra.github.io/angular-dashboard-framework/#/sample/01

www/md_base/src/app/home/home.route.coffee
```diff
@@ -17,3 +17,33 @@ class State extends Config
...
+            ,
+                type: 'hidden'
+                name: 'panels'
+                default_value: [
+                    name: 'overview'
+                    title: 'overview'
+                    collapsed: false
+                    template: 'views/overview_panel.html'
```
tardyp: If you just use a directive, you would not need to encode the template.

Using a directive will allow to define new panels by plugins. (like I did for force fields https://github.com/buildbot/buildbot/blob/master/www/base/src/app/common/directives/forcefields/forcefields.directive.coffee#L7)

tardyp:
I think this design is a good start. Please fix the small nits, and we will merge it. Then you can open new PRs for content, and for directive abstraction of panels

www/md_base/src/app/home/home.route.coffee
shanzi: You'd like every panels' content to be a directive? I have considered this solution but felt this will a little make no sense.

tardyp: On the contrary, this is making a lot of sense for me, and is very much in the spirit of angularJS 2.0, where everything is a directive.

shanzi: `common/directives` might not be a good place for panels' directive. I think panels' content is not that common and each will be used only once.

shanzi:
I changed the `ng-include` solution to directive solution for panel contents. I put the panel related directives under `home` folder as I don't think they are `common` directives.

The default lock and add a button like ADF has not been done. I am inclined to keep current arrangement  because I have not find a elegant way to put such a button on the dashboard. Maybe we should put this aside for the future.

tardyp:
Hi shanzi, please look at https://github.com/buildbot/buildbot/blob/master/www/base/src/app/builders/builders.controller.coffee#L18 to find example on how to create a checkbox that reflect the state of a setting directly on the page

tardyp:
cool stuff!

shanzi:
@tardyp I was not worried about how to put a checkbox on the page, but about the appearance of doing so. Give some time to think it over.

shanzi:
Added a checkbox on the homepage, the appearance is acceptable. 

lock_panels

Live demo at http://buildbot.apps.io-meter.com.

I think we may be able to merge this now :)

tardyp:
I think you should reverse the semantic. And make it: edit panels

shanzi:
@tardyp but that won't suitable to express with a checkbox... Edit panels might be more an action than a state of the panels.

shanzi:
@tardyp We could use `UNLOCK PANELS` but it won't make a big difference, is it?
I inclined to keep it like this and we should go on for further development. :)

shanzi:
hmm... I thought of a better solution with buttons. Let's do it.

shanzi:
Current solution with md-button. Live demo has been updated.

screen shot 2015-05-24 at 1 20 57 am
screen shot 2015-05-24 at 1 21 41 am


There still a bug on the dashboard. It's too late and I'm going to solve it tomorrow.
Please don't merge this PR now.

shanzi:
@tardyp do you think which is better? Using `md-checkbox` or `md-button`?

tothandras:
what about an open/closed lock svg? :)

shanzi:
@tothandras I thought about using icons. But may be the icon is too thin and we only have one icon. So ...

shanzi:
@tardyp Could this PR be merged now?

tardyp:
OK for merge but im away for keyboard

shanzi:
@tardyp @tothandras I now feel `md-checkbox` is a little better than button.

hmm...... What about using a checkbox with text `edit panels`. Although I think edit panels is more an action than a state, but checkbox might look a little better. Hard to decide. What do you think? :-/

shanzi:
@tardyp I talked with @tothandras and we both agree that we should merge this PR like this. About the choice between `md-checkbox` and `md-button`, we think we'd better update it when we have a better idea in the future.

I hope we can merge this PR as soon as possible and turn to the content of home now :)

tardyp:
ok then I merge

PR #1668, opened 2015-05-16
Rewrite the flatten function to be even more flatenning

vlovich:
This fixes the default flattening to apply to tuples too.  Additionally, it also now flattens any nested lists too.  When given something that's not a list, it converts it to a list.  That way the result of the flatten operation is always guaranteed to be a list regardless of input.

### DIFF
--- a/master/buildbot/test/unit/test_util.py
+++ b/master/buildbot/test/unit/test_util.py
@@ -250,6 +250,7 @@ class Flatten(unittest.TestCase):
     def test_string(self):
         self.assertEqual(util.flatten("abc"), "abc")
 
+
 class Ascii2Unicode(unittest.TestCase):
 
     def test_unicode(self):
--- a/master/buildbot/test/unit/test_util.py
+++ b/master/buildbot/test/unit/test_util.py
@@ -236,9 +236,19 @@ class Flatten(unittest.TestCase):
         self.assertEqual(util.flatten([[1, 2], 3, [[4]]]),
                          [1, 2, 3, 4])
 
-    def test_tuples(self):
-        self.assertEqual(util.flatten([(1, 2), 3]), [(1, 2), 3])
+    def test_deeply_nested(self):
+        self.assertEqual(util.flatten([5, [6, (7, 8)]]),
+                         [5, 6, 7, 8])
 
+    def test_tuples(self):
+        self.assertEqual(util.flatten([(1, 2), 3]), [1, 2, 3])
+
+    def test_dict(self):
+        d = {'a': [5, 6, 7], 'b': [7, 8, 9]}
+        self.assertEqual(util.flatten(d), d)
+
+    def test_string(self):
+        self.assertEqual(util.flatten("abc"), "abc")
 
 class Ascii2Unicode(unittest.TestCase):
 
--- a/master/buildbot/test/unit/test_util.py
+++ b/master/buildbot/test/unit/test_util.py
@@ -236,12 +236,12 @@ class Flatten(unittest.TestCase):
         self.assertEqual(util.flatten([[1, 2], 3, [[4]]]),
                          [1, 2, 3, 4])
 
-    def test_deeply_nested(self):
-        self.assertEqual(util.flatten([5, [6, (7, 8)]]),
-                         [5, 6, 7, 8])
+    # def test_deeply_nested(self):
+    #     self.assertEqual(util.flatten([5, [6, (7, 8)]]),
+    #                      [5, 6, 7, 8])
 
-    def test_tuples(self):
-        self.assertEqual(util.flatten([(1, 2), 3]), [1, 2, 3])
+    # def test_tuples(self):
+    #     self.assertEqual(util.flatten([(1, 2), 3]), [1, 2, 3])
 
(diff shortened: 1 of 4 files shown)


master/buildbot/util/__init__.py
```diff
@@ -49,17 +49,34 @@ def key_func(item):
...
-        rv = []
-        for e in l:
-            if isinstance(e, types):
-                rv.extend(flatten(e))
-            else:
-                rv.append(e)
-        return rv
-    else:
-        return l
+def flattened_iterator(l, types=(list, tuple)):
```
sa2ajj: If this function is expected to be used only by `flatten` below, I'd suggest to mark it private `_flattened_iterator`; and in this case default value for `types`` does not seem to be necessary.

sa2ajj:
(Somehow this change is mergeable, needs a rebase)

jaredgrubb:
I think this change is ok .. it does change the way that function works in two ways, but I think those two ways are going to work ok.

However, there should be some unit tests and release note on this.

sa2ajj:
TODO:

- [ ] unit test cases
- [ ] release notes

vlovich:
Where do the release notes go?

vlovich:
Figured it out.  Added release notes.

Added the extra unit test that the old implementation failed + fixed up the existing unit test that failed due to a change in the default behaviour.

This was already based on the latest master so not about that comment.  Perhaps you thought this was done against the eight branch?  Not sure how to evaluate the risk so I decided not to put it there.  Even if I tried to maintain backwards compatibility (result of non-list type & using only list as the default type), the flattening behaviour is extra-flattening so still has the potential to surprisingly break existing code.

As for flattened_iterator, I left it as a public function because it feels like sometimes you do want a generator.

vlovich:
Bah - figured out your comment Mikhael.  I had forgotten to pull from upstream so my master was far behind.  Rebased onto real master.

master/buildbot/util/__init__.py
tardyp: agree.

master/buildbot/util/__init__.py
```diff
@@ -49,17 +49,34 @@ def key_func(item):
...
+    This is safe to call on something not a list/tuple - the original input is yielded.
+    """
+    if not isinstance(l, types):
+        yield l
+        return
+
+    for element in l:
+        for sub_element in flattened_iterator(element, types):
+            yield sub_element
+
```
tardyp: I don't understand why the old implementation did not work for you usecases. It was recursive the same, and supported different types.

I like the iterator implementation better however.

master/buildbot/util/__init__.py
tardyp: ok, you documented it, then that's fine for me. :+1:

tardyp:
good work! :+1:

tardyp:
oups. This fails some unit tests. A possible culprit would be the change of default types from list to (list,tuple)

vlovich:
K.  I'll look at the failing unit tests.  If you look, I added an example test-case that fails with the previous implementation, although it looks like master now passes due to changes you made in f148cdfabf009fa6565864bc52a52b06e2460543.

Namely the types weren't being propagated to the recursive function.  Other than that, it's just a re-write to be use a generator version

vlovich:
Looks like something else that's unhappy.  Even with what I believe to be full backwards compatibility to the old behaviour, tests are failing.

seankelly:
I can't get the Travis results to load so this is based on cursory examination. The tests need to be updated to no longer flatten tuples. I addition, I think tests need to be added that check list and tuple flattening by modifying `types`.

master/buildbot/test/unit/test_util.py
```diff
@@ -202,8 +202,12 @@ def test_deep(self):
         self.assertEqual(util.flatten([[1, 2], 3, [[4]]]),
                          [1, 2, 3, 4])
 
+    def test_deeply_nested(self):
+        self.assertEqual(util.flatten([5, [6, (7, 8)]]),
+                         [5, 6, 7, 8])
+
     def test_tuples(self):
-        self.assertEqual(util.flatten([(1, 2), 3]), [(1, 2), 3])
+        self.assertEqual(util.flatten([(1, 2), 3]), [1, 2, 3])
```
jaredgrubb: I would recommend also adding:
  flatten(tuple)
  flatten(dict)
  flatten(string)

It's good to test each type at the top-level (not inside any other type) as well as inside other types, just to make sure we get the type preservation at each level. Also, adding strings will make sure that any changes to the implementation in the future dont accidentally listify strings ("abc"->['a','b','c']).

sa2ajj:
(once again a rebase is required!)

vlovich:
Can someone provide guidance around what's failing?  Locally a lot of tests fail even on master which makes it difficult to compare what's failing on master vs failing on my branch.  Looking at the travis log the failures don't seem related to this change...

seankelly:
It looks like it's a doc error: line 110 needs to be indented. Line 127 looks to have the same issue.

tardyp:
I fixed the doc, and merge this PR. thanks!

PR #1671, opened 2015-05-17
Add Codecov coverage

stevepeak:
Hey,

I'm the founder of Codecov and had some customers ask to integrate with your awesome product. Thank you for building a great CI tool. Thought I should add Codecov. Checkout all our features at https://codecov.io

Cheers,

Steve

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,8 @@ install:
       export IS_LATEST=false
     fi;
 
+  # codecov is the interface to codecov.io; see after_success
+  - "echo 'travis_fold:start:codecov'; pip install codecov; echo 'travis_fold:end:codecov'"
   # coveralls is the interface to coveralls.io; see after_success
   - "echo 'travis_fold:start:coveralls'; pip install coveralls; echo 'travis_fold:end:coveralls'"
   - "echo 'travis_fold:start:tw'; [ $TWISTED = latest ] || pip install Twisted==$TWISTED; echo 'travis_fold:end:tw'"
@@ -117,4 +119,5 @@ notifications:
   email: false
 
 after_success:
+  - "[ $IS_LATEST = false ] || codecov"
   - "[ $IS_LATEST = false ] || coveralls"
--- a/README.rst
+++ b/README.rst
@@ -27,5 +27,7 @@ Related repositories:
 
 .. |travis-badge| image:: https://travis-ci.org/buildbot/buildbot.svg?branch=master
 .. _travis-badge: https://travis-ci.org/buildbot/buildbot
+.. |codecov-badge| image:: https://img.shields.io/codecov/buildbot/buildbot.svg
+.. _codecov-badge: https://codecov.io/buildbot/buildbot?branch=master
 .. |coveralls-badge| image:: https://img.shields.io/coveralls/buildbot/buildbot.svg
 .. _coveralls-badge: https://coveralls.io/r/buildbot/buildbot?branch=master


stevepeak:
Is there documentation on global environment variables? I'm looking for something [like this](https://github.com/codecov/codecov-bash/blob/068242c2959873f153739f1466089a822547dd63/codecov#L139-L147). Thanks!

sa2ajj:
@stevepeak could you please elaborate a bit what kind of documentation you are looking for?

stevepeak:
Thanks @sa2ajj I'm looking for something like this:

```
BRANCH=master
PROJECT_SLUG=myowner/myrepo
```

The documentation does not layout specific example values which would help for testing purposes. In the [uploader](https://github.com/codecov/codecov-bash/blob/068242c2959873f153739f1466089a822547dd63/codecov#L139-L147) you can get and idea of the variables I'm looking for: branch, commit, slug, pr#, build id, etc.

Thanks.

sa2ajj:
@stevepeak sorry, it took so long to follow up.

As such, Builldbot does not have environment variables -- it has build properties.  These can be exposed as environment variables, but it requires some parameters to the build step.

I'd suggest you describe the workflow you expect to work and then I could tell you how this could be achieved.

stevepeak:
@sa2ajj no problem, thanks for working with me on this.

### Method
There are two different ways on handling the workflow with Codecov. One is the traditional method used at all CI companies. The other is an embedded method (more integrated) so the customer can just add Codecov as a plugin (per-say) by toggling it on.

**The traditional method** is using an uploader, likely the [Global Bash Uploader](https://github.com/codecov/codecov-bash). You can read more about how that works on the repo [codecov/codecov-bash](https://github.com/codecov/codecov-bash#more-options)

**The embedded method** would be adding an upload method directly to Buildbot code base to make the transaction between our products seamless. We would work on this together if this method interests you.

### What Codecov needs
Both methods require Codecov to recognize some variables to understand what repo, commit, etc. are being uploaded. [Checkout this example](https://github.com/codecov/codecov-bash/blob/master/codecov#L122-L134)

tardyp:
Hi stevepeak.
Buildbot is a venerable Ci framework that existed before the CI companies built a "tradition" of using bash environment to transmit build information :)

Our framework uses properties, which are documented here:
http://docs.buildbot.net/latest/manual/cfg-properties.html
You will also need to understand the concepts of sourcestamps:
http://docs.buildbot.net/latest/manual/concepts.html#source-stamps

so probably the best option is to use the codecov -e option, and pass the environment variable via a properly configured Interpolate().
So you can either contribute a doc update explaining how to configure a shell step for codecov.

f.addStep(steps.ShellCommand(command=['codecov', '-e',
                                      util.Interpolate('BUILDBOT=true,BRANCH=%(prop:branch)s, etc')]))

or you can contribute a full featured step that would download codecov automatically, and the run the shellcommand.

You can look at the demo buildbot instance, which show the list of common properties that you can use.
http://nine.buildbot.net/#/builders/3/builds/160

sa2ajj:
@tardyp thanks for your comment: I still did not understand the question :)

sa2ajj:
@stevepeak regarding your change: what will it give Buildbot project?  (we already use coveralls)

stevepeak:
Codecov has plenty of advantages vs Coveralls. It's a decision you may make to switch products, but just my bias suggestion :) I'm happy to show you the many ways our product is far superior.

I will be integrating with Buildbot to provide a more streamline opportunity for your clients to upload reports to Codecov. My original question was simply what are some example variables provided, but it looks like there is more then just some environment variables at play. I'll need to spend some more time looking over how to fully integrate with Buildbot.

djmitche:
It's interesting that you began this PR with a patch to make Buildbot submit its own coverage information to Codecov, but this does not appear to be what you really want to do.

If Codecov provides advantages to our developers over coveralls -- or better, if we can use both without either one spamming our email or pull requests -- then I don't see any reason not to merge this PR.

To the bigger issue, I think the right fix is to write a build step that runs the codecov uploader command (assuming it's already been installed, or given a path to it in a virtualenv), somewhat like what @tardyp has suggested.  I think that would make a great second pull request.

sa2ajj:
@stevepeak I'll see if I could submit a PR for codecov/codecov-python

stevepeak:
@sa2ajj Thanks ;) Both [codecov/codecov-python](https://github.com/codecov/codecov-python) and [codecov/codecov-bash](https://github.com/codecov/codecov-bash) operate the same and are very suitable for the integration. I appreciate your help!

sa2ajj:
Regarding this PR: I looked at the http://codecov.io and saw the differences between coveralls and this service.  I'll merge the PR to see how all those features really compare :)

sa2ajj:
@stevepeak I'd really appreciate a better API docs:
* what parameters can be supplied?
* what types those parameters have?

stevepeak:
@sa2ajj I have not updated our API docs for the v2 endpoint which is a much better one to utilize. Are you planning on using the Bash or Python uploader or would you prefer the APIs directly, I can give you more details on them accordingly. Thanks!

sa2ajj:
@stevepeak Python uploader :)  Would you expect a Buildbot developer to use bash? :)

stevepeak:
Sounds good. Well in that case here is some more information about it.

Codecov is looking for general information about the build that includes the following:

|     Name    |                    Expected                   |     Required     |                                         Description                                         |
| ----------- | ------------------------------------------- | ---------------- | ------------------------------------------------------------------------------------------- |
| `commit`    | git sha                | Yes              | The commit sha being built                                                                  |
| `branch`    | git branch | Yes              | The branch being buit                                                                       |
| `build_url` | string                                      | No               | A target url where the build can be found                                                   |
| `build`     | string                                      | No               | The build number for matrics builds ex. 4.2                                                 |
| `token`     | uuid                                        | Private only     | A uuid token for uploadeing. Found on Codecov website                                       |
| `service`   | "buildbot" (assigned)                       | Yes              | A service name assigned by Codecov                                                          |
| `pr`        | string                                      | No               | If the build is a pull request merge build then this would be the pull request being tested |
| `slug`      | string                                      | No (recommended) | A "owner/repo" pair that states this repos name                                             |

I can add the **buildbot** section into the python uploader that can retrieve these values. Just need know how. See the [Jenkins section for reference](https://github.com/codecov/codecov-python/blob/66046c4187c65fdb17d776f34811b034ee50ed0b/codecov/__init__.py#L134-L143).

sa2ajj:
@stevepeak what about other VCS?

stevepeak:
Codecov requires your repo being hosted on Github, Bitbucket or Gitlab. So we support git and mercurial. A mercurial commit and branch is appropriate too.

sa2ajj:
Sorry, I've been distracted.  I'll submit the PR eventually :)

One question: since you added commands to submit coverage information to codecov.io, is it supposed to add any statuses of PRs?

stevepeak:
We do update the status of the PR for min coverage and relative coverage (incremental). However, this feature needs to be enabled in the UI under **Features -> Github Status**. Thanks

adiroiban:
This is linked from https://github.com/codecov/codecov-python as coming-soon but I think that a separate ticket is needed.

What @stevepeak want is to have a list of env vars which are exported by default by Buildbot and which contain information about branch , github , pr... etc.

The answer is that Buildbot does not expose them by default as Buildbot is way too flexible to handle this :) 

so @stevepeak rather than chasing after the env vars of each CI provider, I think that for codecov you should define a set of required env variables and then someone who want to integrate with codecov will need to make sure that those env variables are exposed.

I see the code from https://github.com/codecov/codecov-python/blob/master/codecov/__init__.py#L260-L430 but I don't see any generic lookup for those environment variables

So we need a last branch which is triggered with `elif os.getenv('CI') == 'true':` and which will write "Generic CI detected" and which will look after a standard set of env vars .

I suggest to prefix all of them with "CI_" ... so that we have "CI_BRANCH" , "CI_BUILD"

stevepeak:
@adiroiban 👏 Agreed. Happy to elaborate on details if necessary.

adiroiban:
See

Stibbons:
<joke>Try to repeat 10 times really fast the title of this pull request</joke>

J-Vin (approved):

PR #1657, opened 2015-05-08
SringDownload: fix a missing "None" check

jaredgrubb:
Simple 'None' check is required.
Modernized the unit tests to be a steps.BuildStepMixin and adjust them to that model.
Adding new unit test that does fail without the patch.

PR #1672, opened 2015-05-17
[nine] StringDownload: fix an exception that happens if the transfer fails

jaredgrubb:
This is a port to Nine (where I should have done it first) of GH-1657.

Simple 'None' check is required.
Adding new unit test that does fail without the patch.

### DIFF
--- a/master/buildbot/test/unit/test_steps_transfer.py
+++ b/master/buildbot/test/unit/test_steps_transfer.py
@@ -51,6 +51,19 @@ def uploadString(string, timestamp=None):
     return behavior
 
 
+def downloadString(memoizer, timestamp=None):
+    def behavior(command):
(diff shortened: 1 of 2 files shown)


master/buildbot/test/unit/test_steps_transfer.py
```diff
@@ -563,7 +615,6 @@ def testBasic(self):
         ss.asDict.return_value = dict(revision="12345")
         s.build.getSourceStamp.return_value = ss
 
-        s.step_status = Mock()
```
jaredgrubb: This wasnt really part of this patch, but had noticed that "step_status" was still being referenced here, so I removed it. Tests still pass.

tardyp:
:+1:

PR #1675, opened 2015-05-17
Git: add 'origin' option

jaredgrubb:
Add an "origin" option to allow the name of the remote repository to be changed. This is renderable, so it can be dynamic.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -66,11 +66,11 @@ class Git(Source):
 
     """ Class for Git with all the smarts """
     name = 'git'
-    renderables = ["repourl", "reference", "branch", "codebase", "mode", "method"]
+    renderables = ["repourl", "reference", "branch", "codebase", "mode", "method", "origin"]
 
     def __init__(self, repourl=None, branch='HEAD', mode='incremental', method=None,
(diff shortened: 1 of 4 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -422,21 +426,23 @@ def _fetchOrFallback(self, _=None):
...
         if shallowClone:
-            args += ['--depth', '1']
+            command += ['--depth', '1']
         if self.reference:
-            args += ['--reference', self.reference]
-        command = ['clone'] + args + [self.repourl, '.']
+            command += ['--reference', self.reference]
+        if self.origin:
+            command += ['--origin', self.origin]
+        command += [self.repourl, '.']
```
jaredgrubb: Note, the change above is mostly non-functional except for the origin line. This removes the variable 'args' and uses 'command' only. I think this improves readability.

tardyp:
:+1:

sa2ajj:
A thought just crossed my mind -- it might be a good idea to be able to specify `origin` to the `GitPoller`: it would help to troubleshoot situations when `GitPoller` says one thing and for some reason `Git` source step does something different.

(In light of that 'richer' experience with git)

PR #1676, opened 2015-05-18
docs: docker: make example code match text

olof:
The username/password didn't match what the text said it should be.


sa2ajj:
Thank you!

PR #1677, opened 2015-05-18
Port from 0.8.12 part2

sa2ajj:
Changes from 0.8.12 that do not touch status targets (which were re-factored in master).


yetanotherion:
:+1:

PR #1682, opened 2015-05-20
Port from 0.8.12 part1

sa2ajj:


master/buildbot/util/__init__.py
```diff
@@ -209,6 +209,35 @@ def datetime2epoch(dt):
         return calendar.timegm(dt.utctimetuple())
 
 
+# TODO: maybe "merge" with formatInterval?
+def human_readable_delta(start, end):
```
tardyp: I think this would belong to reporter.util as I dont see how it could be useful appart from that.

tardyp:
:+1: good work!

master/buildbot/util/__init__.py
sa2ajj: I agree.

Though let's move it when `buildbot/status/github.py` moves to `buildbot/reporters/github.py`

sa2ajj:
Looks like I managed to fix test cases (let's see what Travis says).

PR #1684, opened 2015-05-20
Bug3260 document webhook scheme picky

dankegel:
See http://trac.buildbot.net/ticket/3260

This just documents that webhook repourl for gitlab is git@, and Git step has to match, else changes will be ignored.  Also fixes a couple typos.


tardyp:
:+1: does this mean you validated the gitlab hook with nine?

dankegel:
No, alas, I'm still on 0.8.8.mumble.   But maybe I could try nine in my copious spare time.

sa2ajj:
Thank you :)

PR #1688, opened 2015-05-24
test/util/integration: fix 'new' change race condition

yetanotherion:
This commit permits to reproduce the integration test
race condition. At first glance, the "new" change event
is sent, before the associated scheduler's service has
finished starting (or more precisely activating).

### DIFF
--- a/master/buildbot/test/util/integration.py
+++ b/master/buildbot/test/util/integration.py
@@ -21,14 +21,63 @@ import textwrap
 
 from twisted.internet import defer
 from twisted.internet import reactor
+from twisted.internet import task
 from twisted.trial import unittest
 
 from buildbot.master import BuildMaster
 from buildbot.status.results import SUCCESS
 from buildbot.status.results import statusToString
 from buildbot.test.util import dirs
+from buildbot.schedulers.basic import BaseBasicScheduler
 from buildslave.bot import BuildSlave
 from buildslave.bot import LocalBuildSlave
+
+
+class WaitForStartConsumingSchedulers(object):
+    """
+    startService does not return a deferred. So, when
+    we call yield master.reconfigServiceWithBuildbotConfig
+    we cannot know whether services, like schedulers, have finished their
+    setup.
+    This creates an issue regarding the SingleBranch, AnyBranch schedulers.
+    These schedulers are able to trigger builds only after they have
+    subscribed to the 'new' change event. If we try to start a build with a new change
+    before they subscribed to the event (what can happen as startService does not return a deferred),
+    the build will never start, and the test will timeout.
+    This class implements a poller to ensure that these schedulers have subscribed to the event,
+    before any test tries to create a build with a new change.
+    """
+    POLL_INTERVAL_SEC = 1
+
+    def __init__(self, master):
+        self.ds = [(defer.Deferred(), s)
+                   for s in master.allSchedulers()
+                   if isinstance(s, BaseBasicScheduler) and s._change_consumer is None]
+        self.task = None
+
+    def startPolling(self):
+        self.task = task.LoopingCall(self.callbackSchedulers)
+        self.task.start(self.POLL_INTERVAL_SEC, now=True)
+
+    def stopPolling(self):
+        self.task.stop()
+
+    def getAllDeferreds(self):
+        return [d for (d, _) in self.ds]
+
+    def callbackSchedulers(self):
+        for (d, s) in self.ds:
+            if s._change_consumer is not None:
+                d.callback(None)
+        self.ds = [(d, s) for (d, s) in self.ds
+                   if s._change_consumer is None]
+
+    @defer.inlineCallbacks
+    def waitForSchedToBeActive(self):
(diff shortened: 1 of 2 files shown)


yetanotherion:
@tardyp @sa2ajj I think there is a race condition on any integration test that triggers a build
by means of doForceBuild(useChange=<something not None>). I'll see how we can ensure that the setupConfig of test/util/integration returns only after all services are activated. I'll keep you posted

yetanotherion:
@tardyp @sa2ajj I did not find a better way than this one for now. I let the first commit to ensure
that the travis build pass. If so, I'll rebase the branch to remove that reactor.callLater(10, ).
Reviews are welcome

yetanotherion:
@tardyp @sa2ajj the test are ok, reviews are very welcome

master/buildbot/test/util/integration.py
```diff
@@ -21,16 +21,65 @@
...
 from buildslave.bot import BuildSlave
 from buildslave.bot import LocalBuildSlave
 
 
+class WaitForStartConsumingSchedulers(object):
+    """
+    startService does not return a deferred. So, when
+    we call yield master.reconfigServiceWithBuildbotConfig
+    we cannot know whether services, like schedulers, have finished their
+    setup.
```
tardyp: yes it does. Our version of startService in asyncService returns a deferred.

master/buildbot/test/util/integration.py
```diff
@@ -21,16 +21,65 @@
...
+    This creates an issue regarding the SingleBranch, AnyBranch schedulers.
+    These schedulers are able to trigger builds only after they have
+    subscribed to the 'new' change event. If we try to start a build with a new change
+    before they subscribed to the event (what can happen as startService does not return a deferred),
+    the build will never start, and the test will timeout.
+    This class implements a poller to ensure that these schedulers have subscribed to the event,
+    before any test tries to create a build with a new change.
+    """
+    POLL_INTERVAL_SEC = 1
```
tardyp: 1 sec is supposed to be the total duration of the test (on fast harddrive).
I think we can do better job by turning AnyBranch scheduler into proper asyncservice.

master/buildbot/test/util/integration.py
yetanotherion: Indeed, it does on asyncService but not on ClusteredService (AnyBranchScheduler and SingleBranchScheduler inherit from this).
Are you suggesting to change ClusteredService's startService ?

master/buildbot/test/util/integration.py
yetanotherion: My point of view is the following:
- this commit should be merged like that, a.s.a.p, to stop disturbing buildbot developers with unit test false positives. In each PR, there is a debate on whether the test fails due to the code in the PR, which is not comfortable at all.
- then I propose we work on a solution, like changing BasicBaseScheduler into an asyncService (if it's a ClusteredService, it may be for a reason we have forgotten), in another PR.

tardyp: ClusteredService is for multimaster. I dont see a reason why is startService could not be made async. 
I have no strong opinion about rushing into fixing this racecondition, as this is a known flaky test.

Our current way about flaky tests is to mark them flaky with a decorator, which effectively put them out of the CI. I am not a big fan of this latter option as we ends up just forgotting about those test, and eventually make them fail 100%.

so go ahead and merge, I would prefer merging a workaround rather than disabling the test.

yetanotherion: CusteredService: 
Thanks for the information.

Regarding the flaky decorator solution, I agree with you.
Moreover, even if we try to implement it, note that we should put it on all integration tests that trigger builds my means of <AnyBranch or SingleBranch>Scheduler + add change mechanism, which does not seem very feasible.
@sa2ajj do you agree with the proposal ?
- merge this PR,
- make the CulsteredService startService become async in another PR, that will contain a revert to one of the two commits of this one: https://github.com/yetanotherion/buildbot/commit/866e90b5bdd0d4a820c1e6b4bfa2be61cb2b9218

tardyp: Trac?

Le lun. 25 mai 2015 13:26, yetanotherion <notifications@github.com> a
écrit :

tardyp:
http://trac.buildbot.net/ticket/3263#ticket

master/buildbot/test/util/integration.py
sa2ajj: Yes, I think it's a good idea to file a ticket about `ClusteredService` and its `startService`.

As soon as it's done, I'll merge this PR (I do not like the idea to mark test flaky either).

yetanotherion: @sa2ajj, @tardyp created
http://trac.buildbot.net/ticket/3263#ticket (Thanks Pierre !)
I just set the assignee to myself, so we might be ready to merge this one :) have a good evening !

tardyp:
merged after running common/validate.sh Please dont forget to run it, in order to get autopep8 + fiximport. There is a hook in common/hooks

PR #1693, opened 2015-05-26
Overhaul the monotone source step.

mwanner:
This is an overhaul of the monotone source step to improve database handling and correct the errors I found working with the buildbot setup for monotone itself. Given that buildbot itself is still eight-ish, I plan to backport, but would appreciate some review, first.

### DIFF
--- a/master/buildbot/steps/source/mtn.py
+++ b/master/buildbot/steps/source/mtn.py
@@ -70,44 +70,39 @@ class Monotone(Source):
         if errors:
             raise ConfigErrors(errors)
 
+    @defer.inlineCallbacks
     def startVC(self, branch, revision, patch):
         self.revision = revision
         self.stdio_log = self.addLogForRemoteCommands("stdio")
 
-        d = self.checkMonotone()
-
-        @d.addCallback
-        def checkInstall(monotoneInstalled):
+        try:
+            monotoneInstalled = yield self.checkMonotone()
             if not monotoneInstalled:
                 raise BuildSlaveTooOldError("Monotone is not installed on slave")
-            return 0
-        d.addCallback(lambda _: self._checkDb())
-        d.addCallback(lambda _: self._retryPull())
 
-        # If we're not throwing away the workdir, check if it's
-        # somehow patched or modified and revert.
-        if self.mode != 'full' or self.method not in ('clobber', 'copy'):
-            d.addCallback(lambda _: self.sourcedirIsPatched())
+            yield self._checkDb()
+            yield self._retryPull()
 
-            @d.addCallback
-            def checkPatched(patched):
+            # If we're not throwing away the workdir, check if it's
+            # somehow patched or modified and revert.
+            if self.mode != 'full' or self.method not in ('clobber', 'copy'):
+                patched = yield self.sourcedirIsPatched()
                 if patched:
-                    return self.clean()
-                else:
-                    return 0
+                    yield self.clean()
 
-        # Add a mode specific callback
-        d.addCallback(self._getAttrGroupMember('mode', self.mode))
+            # Call a mode specific method
+            fn = self._getAttrGroupMember('mode', self.mode)
+            yield fn()
 
-        if patch:
-            d.addCallback(self.patch, patch)
-        d.addCallback(self.parseGotRevision)
-        d.addCallback(self.finish)
-        d.addErrback(self.failed)
-        return d
+            if patch:
+                yield self.patch(None, patch)
+            yield self.parseGotRevision()
+            self.finish()
+        except Exception, e:
+            self.failed(e)
 
     @defer.inlineCallbacks
-    def mode_full(self, _):
+    def mode_full(self):
         if self.method == 'clobber':
             yield self.clobber()
             return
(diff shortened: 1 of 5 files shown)


master/buildbot/steps/source/mtn.py
```diff
@@ -83,14 +82,21 @@ def checkInstall(monotoneInstalled):
...
-            if patched:
-                return self.clean()
-            else:
-                return 0
+        # If we're not throwing away the workdir, check if it's
+        # somehow patched or modified and revert.
+        if self.mode != 'full' or self.method not in ('clobber', 'copy'):
+            d.addCallback(lambda _: self.sourcedirIsPatched())
+
+            @d.addCallback
```
tardyp: for big patches like this, we usually ask contributors to change the code to inlineCallbacks, which are so much easier to write and review

tardyp:
I have no idea how monotone works, so cannot comment on the why, but for the how, appart from the addCallback, which need to be translated to inlineCallbacks, the tests and code looks good! thanks!

master/buildbot/steps/source/mtn.py
mwanner: I tried to keep the patch as small as possible, but am happy to change more of it to inlineCallbacks.

sa2ajj: @mwanner that'd be great

bastien31:
needs work? will the inlineCallbacks translation be done?

master/buildbot/steps/source/mtn.py
```diff
@@ -216,52 +220,39 @@ def removeFiles(self, files):
...
     def _pull(self, abandonOnFailure=False):
-        command = ['mtn', 'pull', self.sourcedata,
-                   '--db=%s' % (self.database)]
+        command = ['mtn', 'pull', self.sourcedata, '--db', self.database]
         if self.progress:
             command.extend(['--ticker=dot'])
         else:
             command.extend(['--ticker=none'])
-        d = self._dovccmd(command, abandonOnFailure=abandonOnFailure)
+        d = self._dovccmd('', command, abandonOnFailure=abandonOnFailure)
```
sa2ajj: why this `''`?

sa2ajj: right.  why not make `workdir` parameter and option keyword one?

mwanner: On 06/02/2015 08:07 PM, Mikhail Sobolev wrote:
> In master/buildbot/steps/source/mtn.py
> <https://github.com/buildbot/buildbot/pull/1693#discussion_r31552308>:
> 
>>          if self.progress:
>>              command.extend(['--ticker=dot'])
>>          else:
>>              command.extend(['--ticker=none'])
>> -        d = self._dovccmd(command, abandonOnFailure=abandonOnFailure)
>> +        d = self._dovccmd('', command, abandonOnFailure=abandonOnFailure)
> 
> right. why not make |workdir| parameter and option keyword one?

I'm sorry, I'm unclear about what you mean, here.

I had to use different working directories for the various _dovccmd()
calls, so I added it as an argument to the _dovccmd function. For
consistency with RemoteShellCommand, I also put it before the command
argument.

Regards

Markus Wanner

master/buildbot/steps/source/mtn.py
mwanner: I finally got around to do this. See c49d8d5.

master/buildbot/steps/source/mtn.py
sa2ajj: I re-read what I wrote and I wonder how I did write it :)

What I meant is: why not make `workdir` a keyword parameter with default value to `None`?  Calling `_dovccmd` with `''` looks a bit strange.

mwanner: On 06/03/2015 07:11 AM, Mikhail Sobolev wrote:
> What I meant is: why not make |workdir| a keyword parameter with 
> default value to |None|? Calling |_dovccmd| with |''| looks a bit 
> strange. 
It's not None, it's an empty string, which is a valid relative path. Do 
you prefer '.'? That might clarify it.

WRT a default: I see no reason to prefer '.' over self.workdir as a 
default value (or vice versa), but prefer to keep it as an explicit 
argument I have to think about for every invocation of _dovccmd. I think 
it makes the code clearer to read. (Note that before, self.workdir was 
the default, as the mtn process was only ever executed in that directory.)

Regards

Markus Wanner

sa2ajj: Talking about readability, let's compare two snippets:

```python
d = self._dovccmd('', command, abandonOnFailure=abandonOnFailure)
```

and

```python
d = self._dovccmd(command, workdir='', abandonOnFailure=abandonOnFailure)
```

mwanner: Okay, I changed _dovccmd to allow workdir to be passed as a keyword argument after the command (no default, though). Please go ahead and change RemoteShellCommand to match and to get rid of the exact same issue in steps/source/svn.py.

Plus it's now a '.' (dot) to make it a bit more obvious that it's a relative path.

sa2ajj: Thank you.

I will.

sa2ajj:
Thank you!

PR #1694, opened 2015-05-27
Introduce master log compression.

bastien31:
Log in the db is stored as chunck of text.
The implementation expected a compress operation to be done at the end
of the step. But it will be costly in time/db operations with db
fragmentation risk.

This patch suggest to do live compression when chunck size to insert
is smaller once compressed.
The compress operation will "just" have then to optimize/regroup
all contiguous text chunk, compress them and replace them in the db.

Running that algorithm 50x on the "pyflakes" job, shows that:
- we stored 0.583 MB of log in db instead of 3.079 MB
- space saving ration: 81.05%

Signed-off-by: Sebastien Fusilier <sebastien.fusilier@intel.com>

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -12,6 +12,7 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
+import zlib
 
 import sqlalchemy as sa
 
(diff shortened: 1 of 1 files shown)


master/buildbot/db/logs.py
```diff
@@ -68,22 +69,27 @@ def thd(conn):
...
-                        idx = content.rindex('\n', 0, idx)
-                    content = content[:idx]
-                rv.append(content)
+                if row.compressed:
+                    # Uncompress the payload
+                    rv.append(zlib.decompress(row.content).decode('utf-8'))
+                else:
+                    # Extract all chunks
+                    content = row.content.decode('utf-8')
+                    if row.first_line < first_line:
```
tardyp: I think you need to untab this code.
Once the chunk is uncompressed, you still have to filter the lines requested.

master/buildbot/db/logs.py
```diff
@@ -176,6 +199,9 @@ def thd(conn):
 
     def compressLog(self, logid):
         # TODO: compression not supported yet
+
+        # As live compression is done in appendLog this method shall 'just' analyze if there are benefits
```
tardyp: needs a little bit more explaination of what is planned, as we discussed.

tardyp:
Todo:
- [ ] Update the unit tests
- [ ] support logcompression methods: http://docs.buildbot.net/latest/manual/cfg-global.html#cfg-logCompressionMethod

tardyp:
looks like this compression method is very efficient:
https://pypi.python.org/pypi/lz4

so this could be a good option for live compression

sa2ajj:
`lz4` would introduce another dependency

tardyp:
lz4 would be optional for high bandwith setup

master/buildbot/db/logs.py
```diff
@@ -115,24 +121,41 @@ def thd(conn):
...
+                #  - less "post-processing" operation
+                #  - more time efficient to compress "big" log ASAP
+                #  - avoid DB fragmentation risk
+                last_line = chunk_first_line + remaining.count('\n')
                 conn.execute(self.db.model.logchunks.insert(),
                              dict(logid=logid, first_line=chunk_first_line,
-                                  last_line=last_line, content=chunk,
-                                  compressed=0))
-                chunk_first_line = last_line + 1
+                                  last_line=last_line, content=compressed_remaining,
```
tardyp: we need to ensure compressed_remaining is smaller than MAX_CHUNK_SIZE

bastien31:
To support log [compression method] (http://docs.buildbot.net/latest/manual/cfg-global.html#cfg-logCompressionMethod) and lz4 support (both are related), I would suggest another patch.
We have already benefits with this one, it will be sad to wait for theses features; and one feature per patch is good habits....

master/buildbot/db/logs.py
bastien31: Correct.

master/buildbot/db/logs.py
bastien31: Correct, good catch!

tardyp:
We at least need to prepare the support for several compression method. In the db. compressed in a smallinteger, so we can use this as an enum.

0 -> raw
1 -> gzip
2 -> bz2
3 -> lz4

and only implement gzip as of now.

bastien31:
Doing requested changed in another pull request.

PR #1659, opened 2015-05-11
Some clean-up of buildslave unit tests,

elmirjagudin:
* use @defer.inlineCallbacks style
* added proper skip decarators for win32 only tests
* drop python 2.4 hacks

PR #1695, opened 2015-05-29
Remove slave source commands

elmirjagudin:
As discussed in PR https://github.com/buildbot/buildbot/pull/1659, here is PR that drops the slave-side source commands.

The 'trial buildslave.test' passes and I did some light testing with Git/Hg based checkouts. Nothing obvious breaks.


sa2ajj:
While I appreciate this kind of patches (0 additions, NNN deletions), there's a small problem: build slave reports its API version and build master decides whether a particular feature is available based on that.

Consider a scenario: build slave 0.9 (with the slave side VCS steps removed), a build master 0.8.x (or earlier) and somebody still uses slave side VCS steps.  Did you check what would be the outcome?

tardyp:
/me nods on @sa2ajj request for a special message when those commands are sent from a master

elmirjagudin:
Good point, I'm looking into generating an error message.

As as side note, I have tested using old masters, down to version 0.8.7, with a git checkout step. They all worked without any problems.

sa2ajj:
@elmirjagudin which one git checkout step? :)  If you remove these build slave steps, then the corresponding steps would fail.  For example, this is what I see in `buildbot.steps.source.oldsource.Git`:

```python
        slavever = self.slaveVersion("git")
        if not slavever:
            raise BuildSlaveTooOldError("slave is too old, does not know "
                                        "about git")
        cmd = RemoteCommand("git", self.args)
        self.startCommand(cmd)
```

Essentially, it checks that the command is understood (with your patch, it's not), so the result will be `slave is too old, does not know about git`.

elmirjagudin:
@sa2ajj I was using the 'new' git checkout step, buildbot.steps.source.git.Git, the one which is used in the generated master.cfg.sample file.

Was not aware of the buildbot.steps.source.oldsource, thanks for the tip! Now I finally have a good test case for error message :)

sa2ajj:
I'm glad to be of help :)  Though do not hesitate to ask for help either: what you do is really appreciated and if we can take it to a good state, it'd be a very useful improvement for master.

sa2ajj:
@elmirjagudin in case you need help, do not hesitate to ask for it :)

elmirjagudin:
Not sure what is a good way to generate the error message. As far as I can tell it's not possible to signal that a slave command have been removed within current protocol.

Here is one way I managed to come up with: https://github.com/elmirjagudin/buildbot/commit/4fa4e18119a04bc76e34f4e43d53973a337b5f7e

I have added a list of removed commands to the slave. Whenever master tries to run a command that the slave does not recognize, the list will be checked, and a RemoveCommand exception will be raised.

(This is still work in progress, I plan on adding unittests)

What do you think?

slave/buildslave/commands/registry.py
```diff
@@ -21,16 +21,7 @@
     "uploadFile": "buildslave.commands.transfer.SlaveFileUploadCommand",
     "uploadDirectory": "buildslave.commands.transfer.SlaveDirectoryUploadCommand",
     "downloadFile": "buildslave.commands.transfer.SlaveFileDownloadCommand",
-    "svn": "buildslave.commands.svn.SVN",
```
sa2ajj: What do you think about grouping all these commands at the end of the registry and changing corresponding class to something that produces an error?

elmirjagudin:
Sure: https://github.com/elmirjagudin/buildbot/commit/810daa6f9c9c53b84fc0e73d99c9219b3114a0d1

The only problem I run into is the when generating the error message:

 slave command 'XXX' is no longer supported

I did not want to go with a generic error message that does not mention the actual command, as it would be way less helpful.

The actual slave command name is not passed on anywhere from the SlaveBuilderBase.remote_startCommand() method. That's the reason I added all the subclasses to to RemovedCommand, so we can keep track of the command issued.

elmirjagudin:
I have now made the changes Mikhail suggested. Made the removed commands class to derive from SourceBaseCommand and reworded the error message.

(Not sure what happend to Mikhail's original comment)

sa2ajj:
(Sometimes comments are for commits and GitHub does not show them in the PR view.)

:+1:

tardyp:
arg! you are removing code that is more tested than the average :)

:+1:

sa2ajj:
@elmirjagudin Thank you!

PR #1697, opened 2015-05-30
Fix permission errors with mock

plfiorini:
mock builds fail with a permission error, even if the user running the build
slave was added to the mock group and packages can be built from
the console.

Using /usr/bin/mock as the following error message says fix the problem:

ERROR: [Errno 1] Operation not permitted
ERROR: The most common cause for this error is trying to run /usr/sbin/mock as an unprivileged user.
ERROR: Check your path to make sure that /usr/bin/ is listed before /usr/sbin, or manually run /usr/bin/mock to see if that fixes this problem.

### DIFF
--- a/master/buildbot/steps/package/rpm/mock.py
+++ b/master/buildbot/steps/package/rpm/mock.py
@@ -79,7 +79,7 @@ class Mock(ShellCommand):
         if not self.root:
             config.error("You must specify a mock root")
 
-        self.command = ['mock', '--root', self.root]
+        self.command = ['/usr/bin/mock', '--root', self.root]
         if self.resultdir:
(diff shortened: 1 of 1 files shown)


tardyp:
I think you can better fix this issue by configuring a proper $PATH for your slave.

Please put /usr/bin before /usr/sbin in your $PATH

sa2ajj:
:-1:

bastien31:
Agree with Pierre, hardcoding "/usr/bin/" path is very dangerous. You shall let the user configure properly the execution environment.
I did the mistake for pip in another script/project (as you can find it in different location according to the installation procedure/os) and it brings much more issues/supports than it solves.

plfiorini:
OK

PR #1699, opened 2015-06-01
[Material Design] Recent builds panel's content

shanzi:
Here is the contents of recent builds panel.

Along with this PR are also two new directives to inspect builds. One is `build-status` that display build status as icons with different colors. The other one is `build-item` that displays a build in a bar with some text such as builder name and build number. The latter directive should also link to the corresponding page but as those pages and routers has not yet finished, so I haven't added it now.

recent_builds

In addition to the recent builds panel, I also make several other important modifications:

1. Updated the version of Angular Material to latest version and resolved some compatible problems.
2. Added `angular-moment` for convenient date time formatting

Tests for `build-status` and `build-item` has already been finished but I didn't include a test for `recent-builds` yet. As that directive is just simply using `build-item` to display builds, it is hard to make substantial tests. Do you have any suggestions on this?

There are some other problems too. For example, the panel can not update its content when new builds become completed, I see no elegant solution under current `service` framework, maybe we may need to wait for @tothandras 's new services that could provide a shortcuts for this. Also, each `build-item` will fetch the builder themselves, I added a switch for this but there might be a better solution.

Please review and give your comments :)


tardyp:
I think it looks good. People will request to add information about build failures. So we need to show the status_text in that table

shanzi:
@tardyp The `build-status` icon is not enough?

tardyp:
People need to have a hint on why, without having to dig on the detailled patch. this is the purpose of state_string

sa2ajj:
@shanzi can you check 'console' view for 0.8.x?  One of the best features of it that it shows a summary of all failures: so you do not have to dig through unless you have a really good reason to do that.

bastien31:
need works? 
Agree with @tardyp, it's important to see all the build failures on a single page. (It's a strong request from our users)

shanzi:
showing_reason

@tardyp now showing status or reason such as `EXCEPTION`, `FAILED` or `CANCELED'. Along with corresponding tests are finished.

sa2ajj:
I might have missed the discussion about these big numbers, however I do not really find this very useful (though it takes a lot of screen space)

shanzi:
@sa2ajj hmm.... I think those big numbers are what an overview should have. Do you have any other suggestions?

sa2ajj:
I thought a bit and I do not have any suggestions for overview that'd look useful (you have that 'current' and 'recent' builds panels, which I find more useful).  So my question is do we need at all this kind of overview panel?  I'd rather see this information on the about page...

shanzi:
You lose nothing put this on the dashboard, isn't it?

And yes, maybe the most important function of the overview panel is to take
some space on the screen, but I think it is necessary as it makes the page
looks better. Although it will cost some http request, I'd rather keep it
there. We may try to find a way to reduce the HTTP request cost, update it
with more useful information such as displaying a total number of builds,
or we replace it with other panel if we finally have a better idea. :)

shanzi:
@tardyp showing state_string now. Also applied old color mappings for each different finish state.

state_string

tothandras:
looks good :+1:

bastien31:
:+1:

nam4dev:
:+1:

tardyp:
LGTM, merging.

PR #1700, opened 2015-06-01
Switch git completely to inline callbacks

vlovich:
Switch all the remaining interfaces to use inline callbacks.  Simplifies logic in a bunch of places


master/buildbot/steps/source/git.py
```diff
@@ -238,14 +233,26 @@ def incremental(self):
...
+        rc = yield self._syncSubmodule()
+        if rc != RC_SUCCESS:
+            raise buildstep.BuildStepFailed
+        rc = self._updateSubmodule()
+        if rc != RC_SUCCESS:
+            raise buildstep.BuildStepFailed
+        rc = self._cleanSubmodule()
+        if rc != RC_SUCCESS:
+            raise buildstep.BuildStepFailed
+        defer.returnValue(0)
```
sa2ajj: 0?

vlovich: Fixed - thanks for catching that.

sa2ajj:
I think it's worth implementing in master first, sorry.

vlovich:
That's not an easy request to satisfy.  It looks like eight hasn't been merged into master, at least not for a while.  For example, 9faa0325 rewrites to use runRmdir (among others).

I have probably made incorrect assumptions about the development process.  Can you clarify the policy about regarding what needs to be implemented in master first & when/how things from eight get merged into master?

master/buildbot/steps/source/git.py
```diff
@@ -147,40 +147,35 @@ def __init__(self, repourl=None, branch='HEAD', mode='incremental', method=None,
...
+
+            if self.mode == 'incremental':
+                yield self.incremental()
+            elif self.mode == 'full':
+                yield self.full()
+            if patch:
+                yield self.patch(patch=patch)
+            yield self.parseGotRevision()
+            res = yield self.parseCommitDescription()
+            yield self.finish(res)
```
jaredgrubb: I think you need:
   res = yield self.finish(res)
   defer.returnValue(res)

vlovich: Technically I guess yes.  However, finish (& consequently startVC) doesn't return anything AFAICT.  So omitting this is equivalent.

master/buildbot/steps/source/git.py
```diff
@@ -267,42 +274,37 @@ def fresh(self):
...
-            log.msg("Closing log, sending result of the command %s " %
-                    (self.cmd))
-            return results
-        d.addCallback(_gotResults)
-        d.addCallback(self.finished)
-        return d
+        self.setStatus(self.cmd, res)
+        log.msg("Closing log, sending result of the command %s " %
+                (self.cmd))
+        yield self.finished(res)
```
jaredgrubb: res= ... ; returnValue(res)

vlovich: Same point as before.  finished in BuildStep doesn't return anything.  Propagating None manually seems unnecessary since that's the default behaviour for inlineCallback.

sa2ajj:
@vlovich regarding policy, this is the best we have: https://github.com/buildbot/botherders/blob/master/policies/eight.md

vlovich:
Ahh.  OK.  Thanks for that.  It looks like the fix for ticket 3055 isn't in master which prevents a rebase.  Thoughts?

sa2ajj:
I think I tried to backport it...  I'll try again.

master/buildbot/steps/source/git.py
Stibbons: why not aligning base.Source with other BuildStep, ie:
- startVC returns SUCCESS/FAILURE
- base.Source.start should return self.startVC instead of just calling it

    return self.startVC(branch, revision, patch)

vlovich: I agree that's maybe a good idea (although the source steps still don't use new-style steps so it's less clear exactly what it means for start() to return anything if it even has any meaningful concept).

However, I think that would be out-of-scope of this change.  For instance, mercurial (& I suspect all source steps) would need to be updated in this way too.  This is a purely non-functional change that maintains existing semantics.

vlovich:
Cool.  Let me know when it's merged in & then I can rebase onto master & submit that first.

jaredgrubb:
Vitali, can you post another pull-request with these changes on nine? That should satisfy the request to do nine first.

Also, I see the Travis build failed but I'm not seeing why ... maybe I'm not reading it right?

vlovich:
I will - I was waiting on Mikhail to forward-port the change that was only in eight .  However, I've spent the past two days trying to write unit tests for switch-git-to-sigterm.  I don't think it was appreciated how much effort that actually would take but I'm hoping to finish that today.

sa2ajj:
@vlovich sorry, I've been busy lately (polishing Buildbot at work)

vlovich:
No problem.  I'm busy too.  I'll try this again on the weekend.

djmitche:
@vlovich any chance you could find time to wrap this up?

PR #1702, opened 2015-06-04
Prefer to SIGTERM git instead of killing it outright.

vlovich:
On slaves that support it, send the SIGTERM command first & then SIGKILL
after 30 seconds.  On slaves that don't support it, just try to SIGTERM.

This solves the problem where interrupting a git step (or it timing out)
can leave stale lockfiles around.

Resolves http://trac.buildbot.net/ticket/3269

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -352,11 +352,35 @@ class Git(Source):
                 full_command.append('-c')
                 full_command.append('%s=%s' % (name, value))
         full_command.extend(command)
+
+        # check for the interruptSignal flag
+        sigtermTime = None
+        interruptSignal = None
+
+        # If possible prefer to send a SIGTERM to git before we send a SIGKILL.
+        # If we send a SIGKILL, git is prone to leaving around stale lockfiles.
+        # By priming it with a SIGTERM first we can ensure that it has a chance to shut-down gracefully
+        # before getting terminated
+        if not self.slaveVersionIsOlderThan("shell", "2.16"):
+            # git should shut-down quickly on SIGTERM.  If it doesn't don't let it
+            # stick around for too long because this is on top of any timeout
+            # we have hit.
+            sigtermTime = 1
+        else:
+            # Since sigtermTime is unavailable try to just use SIGTERM by itself instead of
+            # killing.  This should be safe.
+            if self.slaveVersionIsOlderThan("shell", "2.15"):
+                log.msg("NOTE: slave does not allow master to specify interruptSignal. This may leave a stale lockfile around if the command is interrupted/times out\n")
+            else:
+                interruptSignal = 'TERM'
+
         cmd = remotecommand.RemoteShellCommand(self.workdir,
                                                full_command,
                                                env=self.env,
                                                logEnviron=self.logEnviron,
                                                timeout=self.timeout,
+                                               sigtermTime=sigtermTime,
+                                               interruptSignal=interruptSignal,
                                                collectStdout=collectStdout,
                                                initialStdin=initialStdin)
         cmd.useLog(self.stdio_log, False)
--- a/master/buildbot/test/util/steps.py
+++ b/master/buildbot/test/util/steps.py
@@ -322,6 +322,18 @@ class BuildStepMixin(object):
     def _validate_expectation(self, exp, command):
         got = (command.remote_command, command.args)
 
+        for child_exp in exp.nestedExpectations():
+            try:
+                yield self._validate_expectation(child_exp, command)
+                exp.expectationPassed(exp)
+            except AssertionError as e:
+                # log this error, as the step may swallow the AssertionError or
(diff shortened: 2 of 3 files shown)


jaredgrubb:
Tests and notes?

vlovich:
Sure.  What kind of tests are you looking for?

jaredgrubb:
I mean unit tests. We ask that any new features have, at a minimum, test(s) that fail if your code is removed but pass when present, with reasonable coverage. Sometimes that's not possible, but it's what we ask you to try for.

vlovich:
I meant could you elaborate on how I could write such a check.  I'm not sure how to exercise this codepath in unit tests.

master/buildbot/steps/source/git.py
```diff
@@ -352,11 +352,35 @@ def _dovccmd(self, command, abandonOnFailure=True, collectStdout=False, initialS
...
+
+        # check for the interruptSignal flag
+        sigtermTime = None
+        interruptSignal = None
+
+        # If possible prefer to send a SIGTERM to git before we send a SIGKILL.
+        # If we send a SIGKILL, git is prone to leaving around stale lockfiles.
+        # By priming it with a SIGTERM first we can ensure that it has a chance to shut-down gracefully
+        # before getting terminated
+        if not self.slaveVersionIsOlderThan("shell", "2.16"):
```
sa2ajj: You can mock this function to make it return different values.

master/buildbot/steps/source/git.py
```diff
@@ -352,11 +352,35 @@ def _dovccmd(self, command, abandonOnFailure=True, collectStdout=False, initialS
...
+            sigtermTime = '1'
+        else:
+            # Since sigtermTime is unavailable try to just use SIGTERM by itself instead of
+            # killing.  This should be safe.
+            if self.slaveVersionIsOlderThan("shell", "2.15"):
+                log.msg("NOTE: slave does not allow master to specify interruptSignal. This may leave a stale lockfile around if the command is interrupted/times out\n")
+            else:
+                interruptSignal = 'TERM'
+
         cmd = remotecommand.RemoteShellCommand(self.workdir,
```
sa2ajj: and mock this class to check that it's called with correct arguments.

sa2ajj:
HTH

:+1: otherwise!

vlovich:
Is there some example code you can point me at that you would like me to follow the pattern for?
I'm afraid I'd need a lot more time to figure this out or get too frustrated & give up if there's nothing similar.

If there's no existing pattern, is there someone that could write the unit tests for me?  Sorry if I need a lot of hand holding.  I haven't gotten ramped up on the unit tests & they seem very daunting to write.

sa2ajj:
This is how I tested `GerritStatusTarget` in `eight`: [test_status_gerrit.py](https://github.com/buildbot/buildbot/blob/eight/master/buildbot/test/unit/test_status_gerrit.py)

vlovich:
That's probably a good starting point when writing a new test.  However, I'm looking to extend the git tests which seem to follow a very different pattern of writing said tests.  Is there a way to tell expectCommand "hey - I don't care about the remaining commands"?

vlovich:
Added some unit tests.  This was much harder than expected because sigterm & interrupt signal were not being tested anywhere so tests broke all over the place.  I also tried to implement ExpectIgnoreUntil which would be like Expect but ignore any intervening steps but the actual implementation turned out harder than anticipated.  I can split that out into a separate branch if you want.

master/buildbot/test/util/steps.py
```diff
@@ -29,6 +29,50 @@
...
+            missing_in_d2.extend(['{}.{}'.format(k, m) for m in missing_in_v2])
+            for child_k, left, right in different_in_v:
+                different.append(('{}.{}'.format(k, child_k), left, right))
+            continue
+        if d1[k] != d2[k]:
+            different.append((k, d1[k], d2[k]))
+    for k in set(d2.keys()) - both:
+        missing_in_d1.append(k)
+    for k in set(d1.keys()) - both:
+        missing_in_d2.append(k)
```
jaredgrubb: missing_in_d1.extend(set(d2.keys()) - both)
missing_in_d2.extend(set(d1.keys()) - both)

vlovich: Thanks!

vlovich:
I can't tell.  Is the test failure spurious?  Looks to be a network outage.  Not sure where the "recheck" button is to try again.

djmitche:
"zero length field name in format" occurs when using .format with `{}` placeholders on Python-2.6.  The fix is to use `{0}`, `{1}`, etc.

vlovich:
Thanks Dustin.  Updated with this fixed (hopefully).

jaredgrubb:
Your current push only has changes to tests. Is that intentional? I thought there would be a change to Git here?

djmitche:
I think that's everything, then.  I'll have one more look through and then merge.

PR #1707, opened 2015-06-06
Make ClusteredScheduler AsyncService

yetanotherion:
http://trac.buildbot.net/ticket/3263

I wonder whether we should not make SchedulerManager become a BuildbotService, 
so that the reconfigService works similarly,
what do you think @tardyp @djmitche @sa2ajj ?

### DIFF
--- a/master/buildbot/process/measured_service.py
+++ b/master/buildbot/process/measured_service.py
@@ -13,10 +13,20 @@
 #
 # Copyright Buildbot Team Members
 
-from buildbot.process.measured_service import MeasuredBuildbotServiceManager
+from twisted.internet import defer
+
+from buildbot.process import metrics
+from buildbot.util.service import BuildbotServiceManager
 
 
-class SchedulerManager(MeasuredBuildbotServiceManager):
-    name = "SchedulerManager"
-    managed_services_name = "schedulers"
-    config_attr = "schedulers"
+class MeasuredBuildbotServiceManager(BuildbotServiceManager):
+    managed_services_name = "services"
+
+    @defer.inlineCallbacks
+    def reconfigServiceWithBuildbotConfig(self, new_config):
+        timer = metrics.Timer("{}.reconfigServiceWithBuildbotConfig".format(self.name))
+        timer.start()
+        yield super(MeasuredBuildbotServiceManager, self).reconfigServiceWithBuildbotConfig(new_config)
+        metrics.MetricCountEvent.log("num_{}".format(self.managed_services_name),
+                                     len(list(self)), absolute=True)
+        timer.stop()
--- a/master/buildbot/process/measured_service.py
+++ b/master/buildbot/process/measured_service.py
@@ -24,9 +24,9 @@ class MeasuredBuildbotServiceManager(BuildbotServiceManager):
 
     @defer.inlineCallbacks
     def reconfigServiceWithBuildbotConfig(self, new_config):
-        timer = metrics.Timer("{}.reconfigServiceWithBuildbotConfig".format(self.name))
+        timer = metrics.Timer("{0}.reconfigServiceWithBuildbotConfig".format(self.name))
         timer.start()
         yield super(MeasuredBuildbotServiceManager, self).reconfigServiceWithBuildbotConfig(new_config)
-        metrics.MetricCountEvent.log("num_{}".format(self.managed_services_name),
+        metrics.MetricCountEvent.log("num_{0}".format(self.managed_services_name),
                                      len(list(self)), absolute=True)
         timer.stop()
--- a/master/buildbot/schedulers/base.py
+++ b/master/buildbot/schedulers/base.py
@@ -320,9 +320,3 @@ class BaseScheduler(ClusteredBuildbotService, StateMixin):
             waited_for=waited_for, properties=properties_dict, builderids=builderids,
             external_idstring=external_idstring, **kw)
         defer.returnValue((bsid, brids))
-
-    @defer.inlineCallbacks
-    def reconfigService(self, *args, **kwargs):
-        if self.running:
-            yield self.stopService()
-        yield self.startService()
--- a/master/buildbot/schedulers/base.py
+++ b/master/buildbot/schedulers/base.py
@@ -17,7 +17,7 @@ from buildbot import config
 from buildbot import interfaces
 from buildbot.changes import changes
 from buildbot.process.properties import Properties
-from buildbot.util.service import ClusteredService
+from buildbot.util.service import ClusteredBuildbotService
 from buildbot.util.state import StateMixin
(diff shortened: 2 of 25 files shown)


tardyp:
Agreed

Le sam. 6 juin 2015 12:16, yetanotherion <notifications@github.com> a
écrit :

djmitche:
That makes sense to me -- reconfig for schedulers has not been handled very well so far.

yetanotherion:
It's still not finished, but I'll stop on that state for today.
(The test_scheduler_managers fails, 
due to the handling of reconfig of Schedulers). The commit is kind of big, but I did not find a way to split it more. Anyway, opinions about the direction are welcome :)

master/buildbot/util/service.py
```diff
@@ -44,8 +45,129 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
...
+class MasterService(AsyncMultiService):
+    # master service is the service that stops the master property recursion
+
+    @property
+    def master(self):
+        return self
+
+
+class BuildbotService(AsyncMultiService, config.ConfiguredMixin, util.ComparableMixin,
+                      ReconfigurableServiceMixin):
```
Stibbons: mixin should be placed first

master/buildbot/process/measured_service.py
```diff
@@ -0,0 +1,32 @@
...
+#
+# Copyright Buildbot Team Members
+
+from twisted.internet import defer
+
+from buildbot.process import metrics
+from buildbot.util.service import BuildbotServiceManager
+
+
+class MeasuredBuildbotServiceManager(BuildbotServiceManager):
```
tardyp: I think we can include this feature in the buildbotservice

master/buildbot/schedulers/base.py
```diff
@@ -122,6 +122,14 @@ def getPendingBuildTimes(self):
     @defer.inlineCallbacks
     def startConsumingChanges(self, fileIsImportant=None, change_filter=None,
                               onlyImportant=False):
+        # wait for 10s to start consuming changes
```
tardyp: why? looks racy

tardyp:
looks good for the most part

master/buildbot/process/measured_service.py
yetanotherion: that was my first shot, but as buildbot.process.metrics depends on buildbot.util.service
we need this other file

master/buildbot/util/service.py
yetanotherion: Indeed

master/buildbot/schedulers/base.py
yetanotherion: it's the patch with the [DO NOT MERGE] prefix, it ensures the race condition is reproduced (and that this fix corrects it too). (once done that patch will be removed from the history)

yetanotherion:
I wrote the concerns that rose while correcting the last unit tests as commentaries on the DO NOT MERGE
commit. There may be a design issue in BuildbotService, that I'd like to discuss before going into a wrong direction. @tardyp @djmitche @sa2ajj, what do you say ?

yetanotherion:
Unless, we expect that the old services are kept alive, and that they're reconfigured
by the object parameters of the new service object, which must not be started in that case.

tardyp:
Was writting an answer to your comment, then you forced pushed..

sch1_new should not stop sch1. sch1 should reconfig itself with the config of sch1_new. The Manager is responsible of making sure the running services are given their corresponding new config object.
Then the starting and stopping of clustered service should be independant from the reconfig. If sch1 was not started, this is because another master already took the responsability of that scheduler. Then that master should have been reconfigured the same way.
There is nothing to ensure all the masters have their scheduler configured the same. Only their name is important.
Hope this clarifies

master/buildbot/util/service.py
```diff
@@ -329,6 +359,11 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
             if reflect.qual(old.__class__) != reflect.qual(new.__class__):
                 removed_names.add(n)
                 added_names.add(n)
+            # compare using ComparableMixin if they don't support reconfig
```
tardyp: I think all the services should support being reconfigured.

yetanotherion:
I had an issue with git rebase that did not show me where the conflicts were.
I ended up making cherry-picks on a new branch. However,
the new code (since the last review) should be on the last commit only

yetanotherion:
These new commit should be reviewed too (https://github.com/yetanotherion/buildbot/commit/ba8a158fce5a0c11dcf87283a352c6827d012054)

master/buildbot/util/service.py
yetanotherion: just tried it, but it broke unit tests :S

yetanotherion:
I'll do 'non trial based' tests with this PR with a 'real' buildbot instance, before we can merge it in master.
Meanwhile reviews are welcome @tardyp @sa2ajj @Stibbons @djmitche

master/buildbot/schedulers/manager.py
```diff
@@ -13,87 +13,10 @@
...
-
-        metrics.MetricCountEvent.log("num_schedulers", len(list(self)),
-                                     absolute=True)
-
-        # reconfig any newly-added schedulers, as well as existing
-        yield util_service.ReconfigurableServiceMixin.reconfigServiceWithBuildbotConfig(self,
-                                                                                        new_config)
-
-        timer.stop()
+class SchedulerManager(MeasuredBuildbotServiceManager):
```
sa2ajj: (my kind of changes :))

yetanotherion:
The 'real' test validated the PR. So reviews are welcome. More precisely the commits added since the last review are:
- https://github.com/yetanotherion/buildbot/commit/ddd410342714a23daa2baeecec9188ff1485a804
- https://github.com/yetanotherion/buildbot/commit/7f3bff9f63aa60208aea16689c1eca115275ab74
- https://github.com/yetanotherion/buildbot/commit/16ec6f508eec3adc713125b7565a89298d31565a
- https://github.com/yetanotherion/buildbot/commit/fd1c3ad2e93e5f746ec4f6a22eca7aa08a83884b
- https://github.com/yetanotherion/buildbot/commit/f31a19a116677d5ea1c5724ddb182e0abb2f81cc

sa2ajj:
We need a rebase here

yetanotherion:
@sa2ajj rebase done. This PR is indeed big and 'hard' to review, but as reviews on shorter patches* developed later are most likely to preempt this one (and thus increasing the chances to need another rebase), reviews are very (very) welcome :)

*: Unfortunately, I already did my best to make the PR as short as possible

tardyp:
I think this is good work, and is ready for merge :+1:

PR #1712, opened 2015-06-08
Updating to use isinstance(str, basestring). using types.StringTypes …

jvlomax:
…is depreciated

### DIFF
--- a/slave/buildslave/runprocess.py
+++ b/slave/buildslave/runprocess.py
@@ -24,7 +24,6 @@ import stat
 import subprocess
 import sys
 import traceback
-import types
 
 from collections import deque
 from tempfile import NamedTemporaryFile
@@ -443,7 +442,7 @@ class RunProcess:
         self.pp = RunProcessPP(self)
 
         self.using_comspec = False
-        if type(self.command) in types.StringTypes:
+        if isinstance(self.command, basestring):
             if runtime.platformType == 'win32':
                 argv = os.environ['COMSPEC'].split()  # allow %COMSPEC% to have args
                 if '/c' not in argv:
@@ -581,7 +580,7 @@ class RunProcess:
         tf = NamedTemporaryFile(dir='.', suffix=".bat", delete=False)
         # echo off hides this cheat from the log files.
         tf.write("@echo off\n")
-        if type(self.command) in types.StringTypes:
+        if isinstance(self.command, basestring):
             tf.write(self.command)
         else:
             tf.write(win32_batch_quote(self.command))
--- a/slave/buildslave/commands/fs.py
+++ b/slave/buildslave/commands/fs.py
@@ -42,7 +42,7 @@ class MakeDirectory(base.Command):
             if not os.path.isdir(dirname):
                 os.makedirs(dirname)
             self.sendStatus({'rc': 0})
-        except OSError, e:
+        except OSError as e:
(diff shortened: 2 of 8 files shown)


jvlomax:
From the types.py file:

# StringTypes is already outdated.  Instead of writing "type(x) in
# types.StringTypes", you should use "isinstance(x, basestring)".  But
# we keep around for compatibility with Python 2.2.

tardyp:
please remove the 'import types' to make pyflakes happy.

slave/buildslave/runprocess.py
```diff
@@ -443,7 +442,7 @@ def _startCommand(self):
         self.pp = RunProcessPP(self)
 
         self.using_comspec = False
-        if type(self.command) in types.StringTypes:
+        if isinstance(self.command, basestring):
             if runtime.platformType == 'win32':
```
Stibbons: are there plan to switch to python 3?

does it make sense to use six?
```
from six import string_types
isinstance(s, string_types)
```

jvlomax:
Yes, this is in preparation to porting to python3. I'n not sure if we want to make buildbot have six as a requirement. a simple solution to make it compatible with both 2 and 3 without using six is to use:
```
try:
  basestring
except NameError:
  basestring = str
```

Not saying it's a better solution. I'm going to ask @djmitche or @jaredgrubb and see what they think before I commit to either pattern

bastien31:
:+1:

sa2ajj:
@jvlomax is this the only place where it's used?

jvlomax:
@sa2ajj It's the only place in buildslave, which is going to be ported before buildmaster

sa2ajj:
Then :+1: for your suggestion. :)

jvlomax:
Since I'm new to the pull request process, I'm wondering if I can submit other patches or push to my fork while this PR is still pending?

dblotsky:
If you push to this branch on your fork, it will show up on this PR. If you push to another branch on your fork, it will not show up on the PR.

jvlomax:
No, but the commits in this PR are also committed on the branch, so they will show up on a PR request from a separate branch as long as the commits are not merged. 

Just want to make sure this is ok, because there will be quite a few PR through summer

dblotsky:
@jvlomax To do what you mean, you will probably want to do you work on (and make PRs from) feature branches on your fork. That will let you keep different sets of work separate from each other and/or keep several PRs in flight at the same time.

djmitche:
If it's easy to avoid requiring six on the buildslave, that will be good.  But it's also no big deal to require it.

As for branches, generally you shouldn't make a pull request for your "master" branch. Instead, for each bit of work that will turn into a pull request, make a new topic branch.  For example, I created https://github.com/djmitche/buildbot-infra/tree/supybot for my work to set up supybot in the buildbot infrastructure repository.  These branches are called "topic branches".

When review takes a while and you have branches depending on other branches, things can get a little tricky, but hopefully things will be reviewed quickly for you.

djmitche:
Another minor note - commit messages should generally use the imperative tense: "update the foo..", 'remove the bar..", etc..

djmitche:
This looks good to me.  I thought "except foo as bar" wasn't supported in Py2.6, but apparently it is!

jaredgrubb:
This looks good to me!

sa2ajj:
@jvlomax Thanks!

PR #1715, opened 2015-06-10
[Material Design] Current Builds Panel

shanzi:
Here is the current builds panel.

current_panel

This panel is not complex, just bind running builds on the scope and display them. I added `enter` and `leave` animation for the `build-item` so there will be a animation when the builds shows or disappears.

There is a problem now. After a build disappeared in the current builds panel, it does not show in the recent panel automatically. I will dig into the `buildbotService` later to see if this problem is able to be solved.

### DIFF
--- a/www/md_base/src/app/common/directives/builditem/builditem.less
+++ b/www/md_base/src/app/common/directives/builditem/builditem.less
@@ -3,6 +3,7 @@ build-item {
   margin: 5px 0;
   padding: 5px;
   background: rgba(0, 0, 0, 0.05);
+  overflow: hidden;
 
   .inner {
     line-height: 30px;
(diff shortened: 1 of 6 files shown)


shanzi:
@tardyp @tothandras Please review this.

I have scanned over the `buildbotService`, I found if I'd like to let newly finished builds appears in recent  builds panel, I have to mix the update of collection and single object. I am not sure should I modify directly or wait for @tothandras 's new service. It seems that will be a complex change. Or may this is not worth to do?

tardyp:
yes, please dont try to fix buildbotservice. Just assume it will work eventually with @tothandras work

tardyp:
:+1:

shanzi:
Is this OK to be merged?  :)

tothandras:
:+1:

PR #1716, opened 2015-06-11
Replace print with log.msg().

jvlomax:
also 2 more compatibility fixes: 
(.func_name -> .__name__), 
(exception, e -> exception as e)

### DIFF
--- a/slave/buildslave/test/fake/slavebuilder.py
+++ b/slave/buildslave/test/fake/slavebuilder.py
@@ -12,7 +12,7 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-
+from __future__ import print_function
 import pprint
 
 
@@ -33,7 +33,7 @@ class FakeSlaveBuilder(object):
 
     def sendUpdate(self, data):
         if self.debug:
-            print "FakeSlaveBuilder.sendUpdate", data
+            print("FakeSlaveBuilder.sendUpdate", data)
         self.updates.append(data)
 
     def show(self):
--- a/slave/buildslave/test/unit/test_commands_utils.py
+++ b/slave/buildslave/test/unit/test_commands_utils.py
@@ -12,6 +12,7 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
+from __future__ import print_function
 
 import os
 import shutil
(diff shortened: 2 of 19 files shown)


sa2ajj:
While I appreciate the change, I'd rather see all those `print`s to be replaced with `log.msg`

jaredgrubb:
This looks good ... sa2ajj does raise a good point though. It's how we do it on the buildbot side. If we change to log.msg does it fail?

bastien31:
Agree with @sa2ajj, it should not be a big rework to replace print by log.msg and much more cleaner.

jvlomax:
I have updated all the prints to use log.msg, but the tests fail (24 failures in total). One example of a test in test_scripts_upgrade_slave.py:110:


```
self.assertStdoutEqual("buildbot.tac updated\n")
```

Should I update the tests, or should the replacement be considered a bigger job?

sa2ajj:
I'd appreciate very much if you updated the tests.

jvlomax:
I used the log checking function from buildbot. Theas meant that i also had to change the tests a little. I hope this is an ok way to do it

slave/buildslave/test/fake/slavebuilder.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, usePTY=False, basedir="/slavebuilder/basedir"):
 
     def sendUpdate(self, data):
         if self.debug:
-            print "FakeSlaveBuilder.sendUpdate", data
+            print("FakeSlaveBuilder.sendUpdate", data)
```
sa2ajj: why do we still use `print()` here?

slave/buildslave/test/unit/test_commands_utils.py
```diff
@@ -108,7 +109,7 @@ def tearDown(self):
             if os.path.exists(self.target):
                 shutil.rmtree(self.target)
         except Exception:
-            print "\n(target directory was not removed by test, and cleanup failed too)\n"
+            print("\n(target directory was not removed by test, and cleanup failed too)\n")
```
sa2ajj: and here?

sa2ajj:
LGTM.

I'd appreciate answers though.

slave/buildslave/test/unit/test_commands_utils.py
jvlomax: The prints are still in the tests, because I'm assuming they are there to tell the tester if something goes wrong. I don't see any benefit to making it go into a log

jaredgrubb:
LGTM!

slave/buildslave/test/unit/test_commands_utils.py
sa2ajj: Sounds reasonable.

sa2ajj:
Waiting for Travis.

jvlomax:
Restarted build after fixing pep8 stuff

PR #1719, opened 2015-06-12
Replace StringIO with io

jvlomax:
More python3 compatibility fixes

### DIFF
--- a/slave/buildslave/test/unit/test_commands_transfer.py
+++ b/slave/buildslave/test/unit/test_commands_transfer.py
@@ -13,7 +13,7 @@
 #
 # Copyright Buildbot Team Members
 
-import StringIO
+import io
 import os
 import shutil
 import sys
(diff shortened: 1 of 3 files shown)


sa2ajj:
:+1:

ssavrim:
It seems that there is other imports of StringIO in the source code.
Maybe you could replace all the occurences. What do you think ?

jvlomax:
In buildslave of buildbot?

sa2ajj:
I did not see any other occurences in build slave.  Hence :+1: :)

jvlomax:
@ssavrim I'm currently working on porting the buildslave only, buildbot will be ported in july/august. So the remaning StringIO will have to wait :)

ssavrim:
Thanks for the context :) @jvlomax 
You're right it was in buildbot.
:+1:

jaredgrubb:
LGTM!

PR #1720, opened 2015-06-12
py3 compatability misc fixes

jvlomax:
Updates dictionary iteration, range/xrange, filter, map, etc.

Also updated travis to include the future package. Not sure if this is the right way to do it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,6 +102,8 @@ install:
   - "[ $IS_LATEST = false ] || pip install pep8==1.5.7"
   # Install lz4 for log compression utest
   - pip install lz4
+  # Install python-future for py2/3 compatability
+  - pip install future
 
 # Tests running commands
 script:
--- a/slave/buildslave/monkeypatches/bug4881.py
+++ b/slave/buildslave/monkeypatches/bug4881.py
@@ -14,6 +14,8 @@
 #
 # Copyright Buildbot Team Members
 
+from builtins import range
+
 import os
 
 from twisted.internet import process
@@ -189,7 +191,7 @@ class _FDDetector(object):
         # to close
         if maxfds > 1024:
             maxfds = 1024
-        return xrange(maxfds)
+        return range(maxfds)
 
     def _fallbackFDImplementation(self):
         """
@@ -197,7 +199,7 @@ class _FDDetector(object):
         close 256 FDs.
         """
         maxfds = 256
-        return xrange(maxfds)
+        return range(maxfds)
 
 
 detector = _FDDetector()
--- a/slave/buildslave/runprocess.py
+++ b/slave/buildslave/runprocess.py
@@ -17,6 +17,8 @@
 Support for running 'shell commands'
 """
 
+from future.utils import iteritems
+
 import os
 import re
 import signal
@@ -317,7 +319,7 @@ class RunProcess(object):
         if not os.path.exists(workdir):
             os.makedirs(workdir)
         if environ:
-            for key, v in environ.iteritems():
+            for key, v in iteritems(environ):
                 if isinstance(v, list):
                     # Need to do os.pathsep translation.  We could either do that
                     # by replacing all incoming ':'s with os.pathsep, or by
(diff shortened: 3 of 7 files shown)


slave/buildslave/monkeypatches/bug4881.py
```diff
@@ -189,7 +191,7 @@ def _resourceFDImplementation(self):
         # to close
         if maxfds > 1024:
             maxfds = 1024
-        return xrange(maxfds)
+        return list(range(maxfds))
```
tardyp: looks overkill I would prefer to make an utility function which does the right thing according to the python version.

You are also changing the functionality. xrange returns an iterator in 2.7, range returns a list in 2.7 and an iterator in 3.0

.travis.yml
```diff
@@ -102,6 +102,8 @@ install:
   - "[ $IS_LATEST = false ] || pip install pep8==1.5.7"
   # Install lz4 for log compression utest
   - pip install lz4
+  # Install python-future for py2/3 compatability
+  - pip install future
```
tardyp: why not six?

jvlomax:
python future has much better documentation and seems to have a better syntax. It just seems like the better option

slave/buildslave/monkeypatches/bug4881.py
jvlomax: When using python future, range will return an iterator in 2.7.
you can read more about it in the [documentation](http://python-future.org/compatible_idioms.html#xrange) for python future

jvlomax: I might have interpreted the documentation wrong. I'll double check and then update it to be the correct type of ragnge

jaredgrubb: Yeh, I think you have it backwards ... based on my understanding of that doc: 
  xrange should become range()
  range should become list(range())
That sound right?

slave/buildslave/monkeypatches/bug4881.py
```diff
@@ -14,6 +14,8 @@
 #
 # Copyright Buildbot Team Members
 
+from builtins import range
```
sa2ajj: why do we need this import?

slave/buildslave/runprocess.py
```diff
@@ -335,11 +338,11 @@ def to_str(cmd):
             def subst(match):
                 return os.environ.get(match.group(1), "")
             newenv = {}
-            for key in os.environ.keys():
+            for key in iterkeys(os.environ):
```
sa2ajj: why can't we just use `for key in os.environ`?

slave/buildslave/runprocess.py
```diff
@@ -317,7 +320,7 @@ def to_str(cmd):
         if not os.path.exists(workdir):
             os.makedirs(workdir)
         if environ:
-            for key, v in environ.iteritems():
+            for key, v in iteritems(environ):
```
sa2ajj: why do we need `iteritems` here?

slave/buildslave/monkeypatches/bug4881.py
jvlomax: It allows us to use xrange in py2, and range in py3. It "backports" py3 range to py2. You can read more about it in the [python-future documentation](http://python-future.org/imports.html?highlight=builtins)

slave/buildslave/monkeypatches/bug4881.py
jvlomax: @jaredgrubb , Yes, you are correct

slave/buildslave/runprocess.py
jvlomax: We don't. but I like having the code do one thing one way. since ```iteritems(environ)``` is the syntax used in other places, I also did it here

.travis.yml
jvlomax: From the python-future docs: 
>python-future is a higher-level compatibility layer than six that includes more backported functionality from Python 3, more forward-ported functionality from Python 2, and supports cleaner code, but requires more modern Python versions to run.

slave/buildslave/runprocess.py
jvlomax: Yes we can :) I'll fix that

slave/buildslave/runprocess.py
jvlomax: I was mistaken. We **do** have to use 
```
iteritems(environ)
```



dict.iteritems()  does not work in python3

sa2ajj:
Why do we see those other commits here?  Is it rebased against last master?

jvlomax:
I'm wondering the same thing. I believe I rebased against master, and those other commits are already merged into master

jvlomax:
Rebased against master again. Seems to have fixed it

jvlomax:
Rebased and all tests pass. I fixed all the initial issues and hopefully answered any questions. Anything else that should be fixed?

PR #1724, opened 2015-06-15
Enhance forcescheduler button

ssavrim:
Currently the button of schedulers are base on the scheduler name.
This name could be sometimes not user friendly.

This patch proposes to :
    - use the attribute 'buttonName' from the instance of ForceScheduler class
    - use class 'btn-primary' for the button

### DIFF
--- a/master/buildbot/data/forceschedulers.py
+++ b/master/buildbot/data/forceschedulers.py
@@ -24,6 +24,7 @@ from twisted.internet import defer
 def forceScheduler2Data(sched):
     ret = dict(all_fields=[],
                name=unicode(sched.name),
+               button_name=unicode(sched.buttonName),
                label=unicode(sched.label),
                builder_names=map(unicode, sched.builderNames))
     ret["all_fields"] = [field.getSpec() for field in sched.all_fields]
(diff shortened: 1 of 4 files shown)


master/buildbot/data/forceschedulers.py
```diff
@@ -24,6 +24,7 @@
 def forceScheduler2Data(sched):
     ret = dict(all_fields=[],
                name=unicode(sched.name),
+               button_name=unicode(sched.buttonName),
```
tardyp: it should default to the name for backward compatibility

and be documented (and this will probably break the tests)

ssavrim: buttonName is already documented in forcescheduler section :)
I will fix the unit tests and update the default value of buttonName attribute

tardyp: OK, I wasn't aware of it!
Looks good, but we need to fix the default value to other thing as "Force Build"

ssavrim: The default value is set to the name of the scheduler.

master/buildbot/schedulers/forcesched.py
```diff
@@ -669,7 +668,7 @@ def __init__(self, name, builderNames,
         self.all_fields.extend(self.forcedProperties)
 
         self.reasonString = reasonString
-        self.buttonName = buttonName
+        self.buttonName = name if buttonName is None else buttonName
```
bastien31: Why not "buttonName or name"?

bastien31:
:+1:

tardyp:
:+1:

PR #1725, opened 2015-06-15
Statistics service for Buildbot

prasoon2211:
This PR adds support for statistics service. The aim of this project is to provide support for storing and visualizing data generated by builds gathered from within Buildbot (as graphs and charts). See: https://gist.github.com/prasoon2211/cbe93436acc436be61c9 for introduction.

As a first step, this PR adds support for storage of build properties in the InfluxDB storage database.

Here's a brief outline of the the working:

* The user initializes storage backends in the config file:
```
captures = [stats.CaptureProperty('runtests', 'tree-size-KiB')]
c['statsServices'] = [stats.InfluxStorageService('localhost', 8086, 'root', 'root', 'test', captures)]
```
* `StatsService` gets started as a twisted service when Buildbot master is started.
* A build is started (scheduled or forced).
* When the build ends, all the build properties are sent to the running `StatsService`.
* `StatsService.postProperties` then filters out the properties as defined in the config file and sends these properties to the proper storage backend (InfluxDB, in this case). For example, the above configuration will store the value of `tree-size-KiB` for the builder `runtests` in the storage backend (in this case the InfluxStorageService).
* `InfluxStorageService`, which is a Python client for the InfluxDB database, is then responsible for sending this data to the actual running InfluxDB instance on the user's machine. The data is stored as a time-series in InfluxDB. This data is then accessible via the InfluxDB user frontend (that runs in a browser), via an HTTP API to InfluxDB or by using any number of language-specific clients for InfluxDB.

We (me and the mentors @Stibbons and @benallard) plan for the user to be able to use several storage backends. So, `c['statsServices']` will be a list of all the storage backends. Currently, we only support InfluxDB but we'll add support for other storage backends (e.g. Elasticsearch) later on. Even now, with only InfluxDB support, `c['statsServices']` can contain several `InfluxStorageServices`.

Division of work:

The code is divided into three logically separate commits:
* The first commit adds the actual stats service ( + tests + docs).
* The second commit adds support for storage backends - InfluxStorageService, in this case ( + tests + docs).
* The third commit adds support for the capture module - it's very small right now but we plan to add all the capturing support to this (e.g. capturing build/start duration, start, end time etc.). This commit also contains the user documentation (the Buildbot manual).

At first, we wanted to make three separate PRs, one for each of the commits above. But, each commit is not feature-complete, i.e., a single commit doesn't introduce a full feature. So, if the first PR was merged, then Buildbot would have been effectively broken util the third PR was merged. Therefore, we decided to make a single PR but the commits still show the division of code very well.

Instructions for use:
* Install Buildbot with for development with `pip -e`, fetch and apply this PR to your code.
* Install InfluxDB (Google), go to web interface (typically localhost:8083) and create user `root` with pass `root`. Also create a DB called test.
* In the virtualenv for Buildbot, install influxdb (https://github.com/influxdb/influxdb-python)
* Open master.cfg and add:
```
captures = [stats.CaptureProperty('runtests', 'tree-size-KiB')]
c['statsServices'] = [stats.InfluxStorageService('localhost', 8086, 'root', 'root', 'test', captures)]
```
Replace `runtests` with your own builder name and `tree-size-KiB` with a property generated in the builds of this builder.

Then, run Buildbot as usual (maybe force a build or two). To see the data, go to influx's web interface, log in with username and password. Select the database test. This will bring you to a input prompt for a sql like language. Run `show series;` to get the name of the series. It will be of the form `<builder_name>-<property_name>`. Then, run a select *:
`select * from "series_name";`
(Keep the double quotes). This will show you a table of values that Influx registered.

Note: All tests are pasing. pep8/pyflakes passes, pylint gives some warnings but no errors. Rebased on top of master.

**EDIT**:

List of things to correct in this PR:

- [ ] Style-based changes:
  * Dictionary intialization
  * Extra newlines
  * Private memebers prepended with `_`
  * Change some method names
  * Minor refactoring
    * Changes to some tests
    * storage_backends.py to be divided into files.
  * Some error handling for graceful exits in case of failure.

- [ ] Making all superclasses use ABC.
  * Capture
  * CaptureBuildTimes
  * StatsStorageBase

- [ ] Add helper methods for quickly capturing data:
  * Capture a specific data for all builders (properties/times).
  * Capture all properties mathing a regex on all builders.
  * Capture all properties mathing a regex on a list of builders.
  * Capture times (start/end/duration) on a list of builders.

- [ ] Changes to the docs:
  * Fix the format (one sentence per line).
  * Document the changes from the above TODOs.
    * Write about ABC usage and helper methods
  * Add a new tutorial/chapter on the usage of this service.


master/buildbot/statistics/stats_service.py
```diff
@@ -0,0 +1,66 @@
...
+        for svc in self.registeredStorageServices:
+            for cap in svc.captures:
+                for prop_name in properties.properties:
+                    if builder_name == cap.builder_name and \
+                       prop_name == cap.property_name:
+                        context = {
+                            "builder_name": builder_name
+                        }
+                        series_name = builder_name + "-" + prop_name
+                        return threads.deferToThread(svc.postStatsValue, prop_name,
```
Stibbons: simply use @deferToThread decorator on ```def postStatsValue```

I would prefer if postProperties is an inlinecallbakcs

master/buildbot/test/fake/fakestats.py
```diff
@@ -0,0 +1,71 @@
...
+        self.doSomething()
+        return SUCCESS
+
+
+class FakeStatsService(stats_service.StatsService):
+    """
+    Fake StatsService for use in fakemaster
+    """
+    @defer.inlineCallbacks
+    def postProperties(self, properties, builder_name):
```
Stibbons: because here it is an inlinecallbacks

master/docs/developer/config.rst
```diff
@@ -136,6 +136,11 @@ described in :ref:`developer-Reconfiguration`.
         The metrics configuration from :bb:cfg:`metrics`, or an empty
```
Stibbons: on the graph, there could be as many graphical visualization than storage (even more)

master/buildbot/statistics/stats_service.py
prasoon2211: I actually removed that since it was just a wrapper (to reduce the size of the PR). Also, I did a grep search to see how deferToThread is use elsewhere. This is the convention used in Buildbot elsewhere (since deferToThread just returns a Deferred). But okay, we can use inlinecallbacks.

prasoon2211:
It seems some tests are failing on Python 2.6 (but passing on 2.7). I'll fix them and do a force push to preserve commits.

Stibbons:
Dont force push. Commit In your branch the pl will be updated

Le mardi 16 juin 2015, Prasoon Shukla <notifications@github.com> a écrit :

> It seems some tests are failing on Python 2.6 (but passing on 2.7). I'll
> fix them and do a force push to preserve commits.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/buildbot/buildbot/pull/1725#issuecomment-112385102>.
>


-- 
-----
Gaetan

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class InfluxStorageService(StatsStorageBase):
+    """
+    Delegates data to InfluxDB
+    """
+    def __init__(self, url, port, user, password, db, captures,
+                 name="InfluxStorageService"):
+        try:
+            from influxdb import InfluxDBClient
```
benallard: You should put the include on top of the file, and set ``influxdb`` to None when the module is not found. At this place, if influxdb is None, raise the configerror.

prasoon2211:
@djmitche Everything looks good here (the tests + the coverage) and I have the go-ahead of the mentors. Can I get a review please?

Stibbons:
+1

master/buildbot/master.py
```diff
@@ -132,6 +133,9 @@ def create_child_services(self):
         self.metrics = metrics.MetricLogObserver()
         self.metrics.setServiceParent(self)
 
+        self.stats_service = stats_service.StatsService(self)
```
tardyp: If stats_service has nothing particular, I would prefer not particularize it, and put it in the genereic services bucket.

c['services'].append(StatsService(config, for, stats))

master/buildbot/statistics/stats_service.py
```diff
@@ -0,0 +1,67 @@
...
+
+from twisted.internet import defer
+from twisted.internet import threads
+from twisted.python import log
+
+from buildbot.util import service
+from buildbot.statistics.storage_backends import StatsStorageBase
+
+
+class StatsService(service.ReconfigurableServiceMixin, service.AsyncMultiService):
```
tardyp: Please make it a BuildbotService

master/buildbot/statistics/stats_service.py
```diff
@@ -0,0 +1,67 @@
...
+    A middleware for passing on statistics data to all storage backends.
+    """
+    def __init__(self, master):
+        service.AsyncMultiService.__init__(self)
+        self.setName('StatsService')
+        log.msg("Creating StatsService")
+        self.master = master
+        self.registeredStorageServices = []
+
+    def reconfigServiceWithBuildbotConfig(self, new_config):
```
tardyp: please look at the new reporters services to see how buildbotService can be made

master/buildbot/process/build.py
```diff
@@ -579,6 +580,14 @@ def getSummaryStatistic(self, name, summary_fn, initial_value=_sentinel):
...
             return reduce(summary_fn, step_stats_list, initial_value)
 
+    @defer.inlineCallbacks
+    def postProperties(self):
+        # we get all properties, post to MetricsService to filter out
+        # the one that a user wants to post.
+        properties = interfaces.IProperties(self)
+        yield self.master.stats_service.postProperties(properties,
+                                                       self.builder.name)
+
```
tardyp: I think this could be better handled with mq api.
Just register to ("builds", None, "properties", None)
You will need to actually make that event work. because it is not yet handled.
see data/properties.py

master/docs/developer/config.rst
tardyp: Please put mq between builders and statistic service

tardyp:
I realize, I'm asking those stuff a bit late, but this is important to integrate correctly. 

Actually StatsService must eventually be a ClusteredService, but we should wait for @yetanotherion PR to be finished. Waiting for that you will have more beniefits making it a BuildbotService.

master/docs/developer/config.rst
prasoon2211: Okay.

master/buildbot/master.py
prasoon2211: `master.stats_service` is used in the code to call the `StatsService.post*` methods.

I think that you might be confusing the StatsService and storage backends. StatsService isn't user controlled - so, a user needn't touch the config file to have to do anything with the StatsService. Also, it runs reference to the running service instance is stored in `master`. On the other hand, the config file is needed to configure and add a storage backend, the way you mentioned in the comment. This list of storage backends is stored in the config.

master/buildbot/statistics/stats_service.py
prasoon2211: I'll take a look.

master/buildbot/process/build.py
prasoon2211: I am taking a look at how mq works right now. In the meantime, can you please explain what is wrong with this approach?

tardyp: mq to write service like statistics service without having to hardocde hook in the code like you do
It works also better with multi master, as you will only need one master to do the stats, while with your approach, you need all master to have a stat service, and make sure they dont step in each other toe.

djmitche:
@prasoon2211: The reviews from others here are good -- you don't need to wait for my review.

Stibbons:
For the mq stuff, i've though we can dig it but later in this gsoc. I am not sure mq will be always deployed in every buildbot installation, for me, it is just another storage service.

master/buildbot/process/build.py
Stibbons: is mq always available in every buildbot nine installation?

master/buildbot/master.py
Stibbons: ok if we can use the c['services'] registration point. At the end it is just another service. Storage services are actually subservices, there maybe some thing to tune here (using setParent, etc)?

master/buildbot/process/build.py
djmitche: It's always available, just like the DB.  You just can't do a multi-master deployment without some external service like rabbitmq

tardyp:
The mq stuff needs to be handled properly from the beginning. statService should not require any specific hook like there is now.

mq is a pub/sub mechanism made especially for this kind of stuff. statService is nothing else than another reporter plugin. mq has a simple backend for intra process communication, and allow clean separation between components.

http://docs.buildbot.net/latest/developer/mq.html?highlight=mq#module-buildbot.mq.simple

Stibbons:
My personal opinion is that is will be more flexible to have something like:
```
[ buildbot steps/builds ] --> [statService] --> [ amqpstorage ] --> dedicated stat amqp messages
                                                              --> [influxdb]
                                                              --> [mysql]
```

this will open us more doors, allow to yield any data we might need, than if we reuse the buildbot messages:
```
[ buildbot steps/builds ] --> [ buildbot amqp messages] --> [influx db]
```

Also, because if we are only after amqp listener, we can only listen to what buildbot sends us (for me statService will send much more data type than what buildbot even might currently send, the current status of this PL, ie we send only properties and some build status values, are only the first stage of this project). But more important, if we only listen to amqp event, we are actually no more a buildbot code, it is like an external project. Will it still be covered by the gSoc?

There are several paths that might be followed from now, consider this as a first draft of something that might be more flexible later, or we can choose to restrict the scope of this subject to only "listen" what we can receive by amqp (buildbot builder events).

Or we can enhance the Data API message to include our use cases (ex: the disk space used by a given step). If this can be possible, this choice might be the better option, but we will still need hook points in the code to "yields" step specific data (other than step status/duration and properties). In such case the architecture becomes:
```
[ buildbot steps/builds ] --> [statService] --> [ amqpstorage ] --> Data API enhancements
```

There is also the filtering feature to look. We don't want to floor our storages, not even the amqp with message we don't want. So the administrator should be able to configure its installation such as:
- for all builders, I want to the the properties A, B and C sent
- for builders A and C, I want to have the disk space used by steps S1 and S2 (S1 and S2 being in the same builder and being the same step BuildSize, we cannot use properties to send the data)
- let's imagine we have a step that perform actions on a bug tracker referenced by the commit messages, we might want to yields the url or referenced bugs to which patch. I am not sure if we would be able to do it at the end, but if would be very powerful.
- for builders A, B, and D, I want to send all step duration, builders start, end, wait time, and result. This is what Data API currently exposes. 

and so on. I am not sure we want to send all properties of all builders each time! or if we have steps that exposes, for instance, the code coverage of different modules, we may not need to send all these data.

We have weekly meeting on tuesday at 13h30 Toulouse time. There is also the buildbot irc chat on tuesday we can have a talk all together.

Stibbons:
Prasoon, you can start digging into the Data API and see what we can do with it, and how we can enhance it. Indeed in a multimaster environment, it might be mandatory to use it.

master/buildbot/master.py
prasoon2211: But I read the documentation and c['services'] is for allowing advanced users to add custom services to Buildbot at runtime - not for a permanent service like the stats service, I believe.

tardyp: The documentation is not entirely up to date on that part. See the cfg-statustarget.rst documentation.

tardyp: We are extending the concept of 'services' to include more and more configured independant services.
There is no need to have a statsService if it is not configured. c['services'] very well deals with those.

prasoon2211: Okay then. I'll pull it out of the master. Where should I put the initialization in? (which file)

tardyp: You would configure it in the master.cfg like:
```
c['services'] = StatService(backends= [
      InfluxStorageService(...)
])
```

prasoon2211:
Okay, so since I didn't know anything about the mq layer or the data layer, I took a bit of time to read some of the code and documentation to see how thing actually work. As far as I understand, right now, stats service works just fine on a single master setup.
Now, here's what I understand about the multi-master setup (please correct me if I am wrong) - if we have a set of build slaves that work on the same queue of buildrequests, then we can have more than one masters claim the buildrequest and work with a subset of those build slaves. So, I suppose it is just a distributed system that should have problems of any other distributed system - synchronization, concurrecny, race conditions etc.
What I don't understand is what data is Buildbot trying to synchronize? I mean, when a build on a particular build slave ends, Buildbot master just writes some data to the db. If we have 2 masters, each of them can write the data to the db and it wouldn't be a problem since they're not trying to overwrite data (or are they? The two masters have their own build slaves to work with and so, their shouldn't be a conflict). Anyway, that's beside the point at the moment but I'll appreciate it if someone (@tardyp? @Stibbons?) can clarify this doubt.

Anyway, getting back to the topic at hand, all the data that stats service writes to external storage systems is build data - so, it's stuff that is produced during a build. So, build properties, build/step start/end/duration etc. Basically, anything that is accessible via `master.data.get` (If later we need something that isn't in the db, we can make a produceEvent call to mq and pass it the required object as message). As such, we only need to call stats service when the build ends. I stepped through a build and discovered a hook in the process/build.py that calls `finishBuild` in data/builds.py that eventually calls `produceEvent(build, "finished")`. So, this means that we need the "finished" event as as a filter in `startConsumption`. Also, we receive the build instance as the message. The rest of the data we can get from `master.data.get`.

Now, I have a question, @tardyp. If we make stats service work with mq and if we install RabbitMQ and add it in the config, will the stats service work on the multi-master setup? Since I don't really understand how buildbot is using mq to deal with the problem of several masters, I don't know whether this will work.

Also @Stibbons , if merely using the mq layer would make it work for the multi-master setup, then I think that what you proposed adds an unnecessary complexity. First, we'll need the users to install not one but two different dependencies to get stats service to work (e.g, influxDB + RabbitMQ) along with their python clients even for a single master setup. Second, our code for getting data from `[amqp-storage]` will need to keep track of breaking changes in the release cycle of that software (RabiitMQ + Python bindings) along with the breaking changes in the storage backend (InfluxDB). So, I believe that letting the mq layer deal with the amqp service is better.:

```
[ buildbot steps/builds ] --> [statService] --> [ data layer ] --> [mq layer] --> [stats service]
                                                                        |               +--> [influxdb]
                                                      [ amqp storage]<--+               +--> [mysql]
```
@tardyp, you asked me before to make this into a BuildbotService. Later, you asked for it to be a ClusteredService (which I suppose it should since only single instance should be running to avoid duplication of posted data) but asked me wait till https://github.com/buildbot/buildbot/pull/1707 is merged (which doesn't be have any activity in the last half month). So, I think that we can make stats service as a BuildbotService and when that PR is merged, we can make it a ClusteredService. Would that be okay for this PR?

Stibbons:
I wonder if we really want to use mq in every situation. But we follow this path, we definitely need to enhance the data api to support our use case.

I reformulate the drawing for the first proposal: 

[ buildbot steps/builds ] --> [statService] --> [ amqpstorage ] --> dedicated stat amqp messages
                                                  --> [influxdb]
                                                  --> [mysql]

prasoon2211:
Hey all. @Stibbons @tardyp @benallard 

I've made the changes that were suggested before. Now, the whole architecture is based on the MQ and Data layers. Here's how things are functioning now:

1. User adds to master.cfg something like this:
```
captures = [stats.CaptureProperty('runtests', 'tree-size-KiB'),
            stats.CaptureBuildDuration('runtests')]
c['services'] = []
c['services'].append(stats.StatsService(
    storage_backends=[
        stats.InfluxStorageService('localhost', 8086, 'root', 'root', 'test', captures)
    ], name="StatsService"))
```
2. StatsService is reconfigured:
    * All the Capture* instance get a reference to their parent services (the storage backends in which the captured data is to go) and a reference to `self.master`.
    * All the consumers in the Capture* instance are registered to the mq layer.

3. StatsService gets initialized and added to `master.namedServices`.

4. A build is executed. It produces events that go to the mq layer. The mq layer then calls all the Capture* consumers.

5. The CaptureSomething.consumer method runs, gives a call to self.callback to post-process some data, and sends it to storage backend's post method.

6. Storage backend's post method prepares the data in the proper format and send it so the storage backend's server.

That's pretty much it. Let me know what changes to make.

master/buildbot/process/build.py
```diff
@@ -301,7 +301,6 @@ def startBuild(self, build_status, expectations, slavebuilder):
         yield self.master.data.updates.setBuildStateString(self.buildid,
                                                            u'finished')
         yield self.master.data.updates.finishBuild(self.buildid, self.results)
-
```
prasoon2211: Will correct this.

prasoon2211:
Also, the last commit it new. It says an older date because I squashed the newer commits into the older commit.

Stibbons:
@sa2ajj @djmitche it's review time!

master/buildbot/plugins/__init__.py
```diff
@@ -22,9 +22,10 @@
 from buildbot.interfaces import IChangeSource
 from buildbot.interfaces import IScheduler
 from buildbot.plugins.db import get_plugins
+from buildbot import statistics as stats
 
 
-__all__ = ['changes', 'schedulers', 'buildslave', 'steps', 'util', 'reporters']
+__all__ = ['changes', 'schedulers', 'buildslave', 'steps', 'util', 'reporters', 'stats']
```
sa2ajj: any particular reason to use a shorter name?

as the comment below says
```
# Names here match the names of the corresponding Buildbot module, hence
# 'changes', 'schedulers', but 'buildslave'
```

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+        self.parent_svcs = []
+        self.master = None
+
+    def defaultContext(self, msg):
+        return {
+            "builder_name": self.builder_name,
+            "build_number": str(msg['number'])
+        }
+
+    def consumer(self, routingKey, msg):
```
sa2ajj: since our requirement is Python >= 2.6, i'd suggest to use [abstract base classes](https://docs.python.org/2/library/abc.html)

what others think?

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+from twisted.internet import defer
+from twisted.internet import threads
+
+
+class Capture(object):
+    """
+    Base class for all Capture* classes.
+    """
+    def __init__(self, routingKey, callback):
+        self.routingKey = routingKey
```
sa2ajj: which of these class members are expected to be used outside of this hierarchy of classes?

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+        raise NotImplementedError
+
+
+class CaptureProperty(Capture):
+    """
+    Convenience wrapper for getting statistics for filtering.
+    Filters out build properties specifies in the config file.
+    """
+    def __init__(self, builder_name, property_name, callback=None):
+        self.builder_name = builder_name
```
sa2ajj: same here: which of these class members are expected to be used outside of this hierarchy of classes?

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+            context = self.defaultContext(msg)
+            series_name = self.builder_name + "-" + self.property_name
+            post_data = {
+                "name": self.property_name,
+                "value": ret_val
+            }
+            for svc in self.parent_svcs:
+                yield threads.deferToThread(svc.postStatsValue, post_data, series_name,
+                                            context)
+
```
sa2ajj: any reason for an empty line here?

master/buildbot/statistics/capture.py
sa2ajj: it somehow looks that `builder_name` is present in every class.  why not put it into `Capture` class?

master/buildbot/statistics/capture.py
sa2ajj: suggesion on the name -- `consume`

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+        Capture.__init__(self, routingKey, callback)
+
+    @defer.inlineCallbacks
+    def consumer(self, routingKey, msg):
+        """
+        Consumer for CaptureBuildStartTime. Gets the build start time.
+        """
+        builder_info = yield self.master.data.get(("builders", msg['builderid']))
+        if self.builder_name == builder_info['name']:
+            ret_val = self.callback(*self.retValParams(msg))
```
sa2ajj: You expect `retValParams` method to be defined by children, please add it to this class and document expectations.

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+class CaptureBuildStartTime(CaptureBuildTimes):
+    """
+    Capture methods for capturing build start times.
+    """
+    def __init__(self, builder_name, callback=None):
+        def default_callback(start_time):
+            return start_time.isoformat()
+        if not callback:
+            callback = default_callback
+        self._time_type = "start-time"
```
sa2ajj: I do not quite understand how this `_time_type` is used.  In the parent class the value is generated base on the value `retValParams` return, however the latter only returns the value if I understand correctly.

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+    Convenience wrapper for getting statistics for filtering.
+    Filters out build properties specifies in the config file.
+    """
+    def __init__(self, builder_name, property_name, callback=None):
+        self.builder_name = builder_name
+        self.property_name = property_name
+        routingKey = ("builders", None, "builds", None, "finished")
+
+        def default_callback(props, property_name):
+            return props[property_name][0]  # index: 0 - prop_value, 1 - prop_source
```
sa2ajj: should we somehow gracefully handle situation when property is not present?

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+        if self.builder_name == builder_info['name']:
+            properties = yield self.master.data.get(("builds", msg['buildid'], "properties"))
+            ret_val = self.callback(properties, self.property_name)
+            context = self.defaultContext(msg)
+            series_name = self.builder_name + "-" + self.property_name
+            post_data = {
+                "name": self.property_name,
+                "value": ret_val
+            }
+            for svc in self.parent_svcs:
```
sa2ajj: these three lines appear in several classes (I'm not sure why not in every class, btw), I'd suggest to refactor it so this functionality is offered by a suitable parent class.

master/buildbot/statistics/stats_service.py
```diff
@@ -0,0 +1,90 @@
...
+
+        @params
+        data_name: (str) The unique name for identifying this data.
+        post_data: (dict) A dictionary of key-value pairs that'll be sent for storage.
+        buildid: The buildid of the current Build.
+        """
+        build_data = yield self.master.data.get(('builds', buildid))
+        routingKey = ("stats-yieldMetricsValue", "stats-yield-data")
+
+        msg = dict()
```
sa2ajj: I really prefer
```python
msg = {
    'data_name': data_name,
    'post_data': post_data,
    'build_data': build_data
}
```

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
```
sa2ajj: I'd suggest to make this file `buildbot/statistics/storage_backends/base.py` and move InfluxDB related code to `buildbot/statistics/storage_backends/influxdb.py` as we usually do.

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
...
+    InfluxDBClient = None
+
+
+class StatsStorageBase(object):
+    """
+    Base class for sub service responsible for passing on stats data to
+    a storage backend
+    """
+    def postStatsValue(self, name, value, series_name, context={}):
+        return defer.succeed(None)
```
sa2ajj: Do we need a default behaviour here? I'd say it's an abstract method.

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
...
+except ImportError:
+    InfluxDBClient = None
+
+
+class StatsStorageBase(object):
+    """
+    Base class for sub service responsible for passing on stats data to
+    a storage backend
+    """
+    def postStatsValue(self, name, value, series_name, context={}):
```
sa2ajj: using `{}` as a default for a function parameter is a big NO

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
...
+
+class InfluxStorageService(StatsStorageBase):
+    """
+    Delegates data to InfluxDB
+    """
+    def __init__(self, url, port, user, password, db, captures,
+                 name="InfluxStorageService"):
+        if not InfluxDBClient:
+            config.error("Python client for InfluxDB not installed.")
+        self.url = url
```
sa2ajj: Do we plan to use all of these class data members outside of the class?

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
...
+        self.db = db
+        self.name = name
+
+        self.captures = captures
+        self.inited = False
+        self.client = InfluxDBClient(self.url, self.port, self.user,
+                                     self.password, self.db)
+        self.inited = True
+
+    def postStatsValue(self, post_data, series_name, context={}):
```
sa2ajj: Is this method synchronous??

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
...
+        if not self.inited:
+            log.err("Service {0} not initialized".format(self.name))
+            return
+        log.msg("Sending data to InfluxDB")
+        log.msg("post_data: {0!r}".format(post_data))
+        log.msg("context: {0!r}".format(context))
+
+        data = {}
```
sa2ajj: see above comment about initialising a dictionary.

master/buildbot/statistics/storage_backends.py
```diff
@@ -0,0 +1,70 @@
...
+        data['name'] = series_name
+        data['fields'] = post_data
+        data['tags'] = context
+        points = [data]
```
sa2ajj: and, I think, you can just initialise `points` as
```python
points = [{
   ...
}]
```

master/buildbot/test/fake/fakestats.py
```diff
@@ -0,0 +1,81 @@
...
+
+class FakeInfluxDBClient(object):
+    """
+    Fake Influx module for testing on systems that don't have influxdb installed.
+    """
+    def __init__(self, *args, **kwargs):
+        self.points = None
+
+    def write_points(self, points):
+        self.points = points
```
sa2ajj: maybe `self.points.extend(points)` and change the initial value above to `[]`?

master/docs/developer/stats-service.rst
```diff
@@ -0,0 +1,255 @@
...
+Here is a diagram demonstrating the working of the stats service:
+
+.. image:: _images/stats-service.png
+
+Stats Service
+-------------
+
+.. py:class:: StatsService
+
+   An instance of this class functions as a :class:`BuildbotService`. The instance of the running
```
sa2ajj: One sentence per line!

sa2ajj: (and I am not going to repeat this for every instance of wrong formatting)

master/docs/developer/stats-service.rst
```diff
@@ -0,0 +1,255 @@
...
+
+   .. py:method:: reconfigService(self, storage_backends)
+
+     :param storage_backends: A list of storage backends.
+
+     This mehtod is called automatically to reconfigure the running service.
+
+   .. py:method:: registerConsumers(self)
+
+      Internal method called to register all consumers (methods from Capture classes) to the MQ layer.
```
sa2ajj: "Internal" for who??

master/docs/developer/stats-service.rst
```diff
@@ -0,0 +1,255 @@
...
+
+.. _storage-backend:
+
+
+Storage backends
+----------------
+
+Storage backends are responsible for storing any stats-data sent to them.
+A storage backend will generally be some sort of a database-server running on a machine.
+(*Note*: This machine may be different from the one running :class:`BuildMaster`)
```
sa2ajj: sphinx supports
```rst
.. note::

   text of note
```

master/docs/developer/stats-service.rst
```diff
@@ -0,0 +1,255 @@
...
+   * ``routingKey``: (tuple) The routing key to be used by :class:`StatsService` to register consumers to the MQ lasyer for the subclass of this class.
+   * ``callback``: The callback registered with the MQ layer for the consumer of a subclass of this class. Each subclass must provide a default callback for this purpose.
+
+   .. py:method:: defaultContext(self, msg):
+
+      A method for providing default context to the storage backends.
+
+   .. py:method:: consumer(self, routingKey, msg):
+
+      Raises ``NotImplementedError``. Each subclass of this method should implement its own consumer.
```
sa2ajj: Instead of "Raises `NotImplementedError`", I'd suggest to write "each subclass of class `Capture` must implement this method, which is expected to do ..."

master/docs/developer/stats-service.rst
```diff
@@ -0,0 +1,255 @@
...
+
+   * ``builder_name``: The name of builder in which the property is recorded.
+   * ``property_name``: The name of property needed to be recorded as a statistic.
+   * ``callback=None``: The callback function that is used by ``CaptureProperty.consumer`` to post-process data before formatting it and sending it to the appropriate storage backends. A default callback is provided for this.
+
+   **The default callback:**
+
+     .. py:function:: default_callback
+
+     Defined in ``CaptureProperty.__init__``. Receives:
```
sa2ajj: I do not think "Defined in ..." is really necessary.  Those who want to look at the implementation will see where it is defined.  However the list of anticipated parameters is due.

master/docs/developer/stats-service.rst
```diff
@@ -0,0 +1,255 @@
...
+
+   It takes the following arguments:
+
+   * ``builder_name``: The name of builder in which the property is recorded.
+   * ``property_name``: The name of property needed to be recorded as a statistic.
+   * ``callback=None``: The callback function that is used by ``CaptureProperty.consumer`` to post-process data before formatting it and sending it to the appropriate storage backends. A default callback is provided for this.
+
+   **The default callback:**
+
+     .. py:function:: default_callback
```
sa2ajj: why not write `default_callback(props, property_name)`?

master/buildbot/statistics/capture.py
```diff
@@ -0,0 +1,200 @@
...
+    def consumer(self, routingKey, msg):
+        raise NotImplementedError
+
+
+class CaptureProperty(Capture):
+    """
+    Convenience wrapper for getting statistics for filtering.
+    Filters out build properties specifies in the config file.
+    """
+    def __init__(self, builder_name, property_name, callback=None):
```
sa2ajj: what if we want to capture data from _all_ builders?

sa2ajj:
I like the way it goes, however there are things to think about and to fix.

master/buildbot/statistics/capture.py
Stibbons: indeed:
- how to filter a data on all builders
- how to filter all data on some builder
- how to capture all data from all builders

master/buildbot/plugins/__init__.py
prasoon2211: I hadn't read that. I will change it to the module name (`statistics`).

master/buildbot/statistics/capture.py
prasoon2211: Well, we should be able to provide helper methods for all these cases that will, internally, register the proper captures with the proper storage backends.

We have a meeting tomorrow - we'll discuss it more in detail then.

master/buildbot/statistics/capture.py
prasoon2211: routingKey is., rest aren't. Should I add an underscore to the names?

master/buildbot/statistics/capture.py
prasoon2211: Nothing is used outside CaptureProperty.

As for `builder_name`, we're not moving it into the Capture class because we plan on adding more capture classes later on which might not need `builder_name`.

master/buildbot/statistics/capture.py
sa2ajj: Yes, please.  We have had a bit of maintenance problems when things that are not supposed to be used outside of the class were exposed.

master/buildbot/statistics/capture.py
sa2ajj: For private things, please prepend the names with `_`

sa2ajj: As for `builder_name`, let's wait for those other classes then.

master/buildbot/statistics/capture.py
prasoon2211: I'm not sure. The user should be made aware that the there's a problem with the build. If we catch the error here and not send the data over to InfluxDB, the user might not know there a problem with their build for a long time, even with all the logging. This might cause them to miss build data of several builds. I'll discuss it with the mentors tomorrow (Friday).

sa2ajj: I should extend my  "gracefully": the current code will just fail without providing any useful information.  My suggestion is to "fail" while providing useful information, such as:
* builder name
* build number
* capturer

master/buildbot/statistics/capture.py
sa2ajj: on unrelated note, does it mean that the user would have to create an instance of `CaptureProperty` class for each property she wants to log?

master/buildbot/statistics/capture.py
prasoon2211: That sounds like a good idea.

master/buildbot/statistics/capture.py
prasoon2211: You mean 75-77? They only appear inside the consumer method. I'll refactor them away.

sa2ajj: Yes, 75-77.

master/buildbot/statistics/capture.py
prasoon2211: Just some neatness. It separates two branches of logic the method can follow. I checked with pep8 and there wasn't a problem. Is this against Buildbot style guide?

master/buildbot/statistics/capture.py
prasoon2211: Okay.

master/buildbot/statistics/capture.py
prasoon2211: `_time_type` is just the name of the column in the database under which the value will go.

master/buildbot/statistics/stats_service.py
prasoon2211: I'll change this.

master/buildbot/statistics/storage_backends.py
prasoon2211: Okay.

master/buildbot/statistics/capture.py
prasoon2211: I personally think that we should just use `NotImplementedError`. It would be more idiomatic Python since Python follows more of a 'easier to ask for forgiveness than permission' principle.

master/buildbot/statistics/storage_backends.py
prasoon2211: Again, I would prefer to use a simple `NotImplementedError` rather than using abstract classes.

master/buildbot/statistics/storage_backends.py
prasoon2211: No. I'll prepend them with a `_`.

master/buildbot/statistics/storage_backends.py
prasoon2211: Yes, but it is called in a different thread using `defer.deferToThread` so it behaves asynchronously to the rest of the system. See any consumer in capture.py - this method is called at the end of the method.

master/buildbot/statistics/storage_backends.py
prasoon2211: Will do.

master/buildbot/test/fake/fakestats.py
prasoon2211: I'm not sure why that would be helpful.

master/docs/developer/stats-service.rst
prasoon2211: Will do.

master/docs/developer/stats-service.rst
prasoon2211: "Internal method for this class" would probably be a better phrasing.

master/docs/developer/stats-service.rst
prasoon2211: Okay.

master/docs/developer/stats-service.rst
prasoon2211: I'll make the changes.

master/buildbot/statistics/capture.py
sa2ajj: The problem with `NotImplementedError` is that it's raised when a particular method is _used_, while abstract base classes would cause the error to be raised when a class is instantiated.  The latter is usually more helpful to understand what the problem is.

However, as I said, I wonder what others think about this.

Stibbons: I vote for abc if allowed by buildbot

master/buildbot/statistics/capture.py
sa2ajj: I re-read the code and it became a bit more obvious this time :smile:

I suggest that `_time_type` is initialised in `CaptureBuildTimes`, otherwise it's not really clear where it is used.

master/buildbot/statistics/capture.py
Stibbons: maybe support some kind of regex?

master/buildbot/statistics/capture.py
sa2ajj: Since there's a continuation in the previous line, the whole construct looked strange to _me_.  It's a personal preference, so feel free to ignore :)

Stibbons: pep8 removes all ambiguity for this kind of thing. autopep8 will do the job (use the validate.sh script provided by the buildbot code, there is also a post commit hook)

master/buildbot/statistics/storage_backends.py
sa2ajj: In any case, there's no need for a default behaviour.

master/buildbot/test/fake/fakestats.py
sa2ajj: The tests may become a bit more complicated and it would be possible to check that all data is collected.

master/docs/developer/stats-service.rst
sa2ajj: Yes, sounds better.

master/buildbot/statistics/storage_backends.py
benallard: method deferred to other threads should have their name prepended with ``thd_``

benallard:
Thank you very much @sa2ajj for that first round of style-comments, hopefully that will make it easier for other to look at it !
I'd really welcome some comments about the design itself before we can plan our next actions there.

master/buildbot/statistics/capture.py
prasoon2211: So, I'm holding off on this change for a bit.

master/buildbot/statistics/capture.py
prasoon2211: I discussed this with Ben today. So, we're going to have a few helper method that would allow easier method for this. And yes, a regex seems like a good idea. However, I'm not sure whether such helpers should go in the next PR or in this one.

master/buildbot/statistics/capture.py
sa2ajj: I wonder what your mentors said about this item.

prasoon2211: Well, I talked to Ben today and he was okay with this. Gaetan has already given his assent above.

sa2ajj: Okay with what? :)  Current implementation or the suggested change?

prasoon2211: With the suggested implementation. I didn't talk about this in so many words - I just said to him that I'll be changing the implementation to all as suggested and he said okay. So, maybe we should take a second opinion.

djmitche: Buildbot uses `NotImplementedError` in a number of other places, but largely out of historical necessity.  Especially if this is a class which users will be subclassing, I think there's a big benefit in usability to earlier, clearer errors -- which `abc` would supply.

tardyp: I was not familiar with abc as well. I think we should start to use it. Problem I can see is that it is fairly more complex concepts than just raise NotImplemented. So a note in the developer documentation will be more than welcome, as long with a link to a tutorial a little bit more complete than just the doc pointed by misha.

I'm particularly worried by the notes of the docs:

- Note Unlike Java abstract methods, these abstract methods may have an implementation. This implementation can be called via the super() mechanism from the class that overrides it. This could be useful as an end-point for a super-call in a framework that uses cooperative multiple-inheritance.

- C++ programmers should note that Python’s virtual base class concept is not the same as C++’s.

master/docs/manual/cfg-global.rst
```diff
@@ -553,6 +553,127 @@ This value can also be changed via a reconfig.
 
 Read more about metrics in the :ref:`Metrics` section in the developer documentation.
 
+.. bb:cfg:: stats-service
+
+Statistics Service
+~~~~~~~~~~~~~~~~~~
+
+The Statistics Service (or stats service) is a new service being introduced in Buildbot Nine.
```
tardyp: please remove ref to buildbot nine (only reference it in the release notes)

master/docs/manual/cfg-global.rst
```diff
@@ -553,6 +553,127 @@ This value can also be changed via a reconfig.
...
+
+::
+
+    captures = [stats.CaptureProperty('runtests', 'tree-size-KiB'),
+                stats.CaptureBuildDuration('runtests')]
+    c['services'] = []
+    c['services'].append(stats.StatsService(
+        storage_backends=[
+            stats.InfluxStorageService('localhost', 8086, 'root', 'root', 'test', captures)
+        ], name="StatsService"))
```
tardyp: I like the API!

benallard: Thanks !

A small note there, while it's pretty clear that 'tree-size-KiB' is a property, it might not be obvious that 'runtests' is the name of the builder, choosing a more explicit name here might help.

master/buildbot/statistics/capture.py
benallard: For the particular use-case of capturing a property on all builders, we agreed on making the ``buildername`` parameter optional => When not provided, capture for all builders.

benallard: @sa2ajj: Yes, you have to create one instance of the class for each property you want to capture.

master/buildbot/statistics/capture.py
benallard: For the record, abc looks like a natural fit here: +1

master/buildbot/statistics/capture.py
tardyp: See the authz PR, It introduces an optional fnmatch or regex match for string matching. For now it is only in the authz namespace, but it can be in the global util namespace

tardyp:
I am *very* worried that there is too much comments on this thread. We will either miss things or never merge anything.

My proposal is to:
- Create a todo list in the design document
- have the reviewers of this thread approve the todo list
- merge this code as is
- submit one easy to review PR for each of the things in the todo list

Until all those PRs are finished, no new features.

sa2ajj:
I agree with @tardyp: to make it easier to track TODO items progress, let's add it to the first message (the one provided for PR submission).

master/buildbot/statistics/capture.py
prasoon2211: Very well. I'll change the implementation to use abc and update the docs to reflect this change.

sa2ajj: But first, please, check out two last "global" comments @tardyp and I.

master/docs/manual/cfg-global.rst
prasoon2211: Okay.

master/docs/manual/cfg-global.rst
prasoon2211: Yes, I agree. I'll switch the name to something like 'Builder1'.

prasoon2211:
This idea is fine by me. I'll make a TODO list and submit it for approval.

prasoon2211:
@tardyp @sa2ajj I have made a list. It's in the original PR message in the top. Please take a look.

sa2ajj:
Please check out GitHub's markup

prasoon2211:
I made the lists into boxes.

prasoon2211:
@tardyp can you guys please approve the todo list and merge this PR? I'd like to move on with that list.

Stibbons:
I agree to consolidate an external todo list and merge this first pl when a few fixes left are done. Please update the slides we use to follow this work for the next weeks.

sa2ajj:
@prasoon2211 the TODO list looks good to me.

bhearsum:
I'm not sure why I got assigned to this...

prasoon2211:
Hey guys. I am in moving and so, I will not be available for comment for
the next day or two. Thanks.

tardyp:
@bhearsum Sorry, probably misclick.

tardyp:
The todo list looks good to me. we have +1 from @Stibbons and @sa2ajj. merging.

tardyp:
@prasoon2211 Please make sure you run "make hooks" once before commiting, there are a lot of autopep8 issues in this code (I'll fix them this time)

tardyp:
please also make sure you run common/validate.sh

There are 6 "yield defer.returnValue(..)" in this code!

prasoon2211:
Thanks @tardyp ! I have been a little busy moving last few days but I'll start working on the new PRs now. Also, I'll make sure to tun the hooks.

PR #1729, opened 2015-06-19
do the hustle (2)

ravench:
Corrected a mistake in the previous pull request and added a test.

This is the outcome of a conversation on #qutebrowser, where someone mentioned that qutebrowser-bb needed 'to learn to do the hustle'... (inspired by ```command_DANCE```)

This should do exactly that.


bastien31:
Essential :p, but no way to have better hustle movement?

PR #1730, opened 2015-06-19
Authz

tardyp:
lieter on irc asked me how we could just make simple authz working.
I unburied this very old branch, and connected the remaining things together, so that it actually works for a simple configuration:
```
authz = Authz(
  allowRules=[
    StopBuildEndpointMatcher(role="admins", defaultDeny=True),
    ForceBuildEndpointMatcher(role="admins", defaultDeny=True)
  ],
  roleMatchers=[
    RolesFromEmails(admins=["my@email.com"])
  ]
)
auth=util.UserPasswordAuth({'my@email.com': 'mypass'})
c['www']['auth'] = auth
c['www']['authz'] = authz

```

This PR also sets the owner of forced builds to the email of the authenticated user.


tardyp:
@sa2ajj @jaredgrubb any reviews?

sa2ajj:
@tardyp will do later today (after the meeting)

tardyp:
@sa2ajj I got review on uicustomization, but not on authz :)

jaredgrubb:
LGTM!

master/buildbot/data/forceschedulers.py
```diff
@@ -38,22 +38,29 @@ class ForceSchedulerEndpoint(base.Endpoint):
...
     def control(self, action, args, kwargs):
         if action == "force":
-            for sched in self.master.allSchedulers():
-                if sched.name == kwargs['schedulername'] and isinstance(sched, forcesched.ForceScheduler):
-                    try:
-                        res = yield sched.force("user", **args)
-                        defer.returnValue(res)
-                    except forcesched.CollectedValidationError as e:
-                        raise BadJsonRpc2(e.errors, JSONRPC_CODES["invalid_params"])
+            sched = yield self.findForceScheduler(kwargs['schedulername'])
```
sa2ajj: do we always have `schedulername` here? I can see that the previous code assumed that as well, so my question would be for the old code as well :)

master/buildbot/data/forceschedulers.py
```diff
@@ -38,22 +38,29 @@ class ForceSchedulerEndpoint(base.Endpoint):
...
+            if "owner" not in args:
+                args['owner'] = "user"
+            try:
+                res = yield sched.force(**args)
+                defer.returnValue(res)
```
sa2ajj: I'm curious if `defer.returnValue` also acts as a `return`..

tardyp: yep. It is actually raising an exception that is caught by the inlineCallbacks wrapper.

master/buildbot/data/forceschedulers.py
tardyp: I think it might be worth to return proper error in case the param is not present

sa2ajj:
I would like two answers (one is already give :)), otherwise LGTM

master/buildbot/data/forceschedulers.py
tardyp: Actually, after looking at the code:

```
    pathPatterns = """
        /forceschedulers/i:schedulername
    """
```
This garanties that the control is always called with parameter schedulername present and being an identifier.

sa2ajj: good

PR #1731, opened 2015-06-20
Added requirement for future

jvlomax:


jaredgrubb:
As soon as Travis responds ... I'll merge this is so we can unbreak our buildbots.

jvlomax:
Assuming this is the correct fix. There wouldn't be anywhere else that I need to at a dependency?

jaredgrubb:
Not that I know of!

PR #1734, opened 2015-06-21
buildbot.status is not available in master any more

sa2ajj:
* remove unnecessary interface
* make sure reporters are exported for 'import *'

### DIFF
--- a/master/buildbot/plugins/__init__.py
+++ b/master/buildbot/plugins/__init__.py
@@ -21,11 +21,10 @@ from buildbot.interfaces import IBuildSlave
 from buildbot.interfaces import IBuildStep
 from buildbot.interfaces import IChangeSource
 from buildbot.interfaces import IScheduler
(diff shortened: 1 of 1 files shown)


bastien31:
In the commit message you mention that interface was removed, but /master/buildbot/interfaces.py still contains "IStatusReceiver".

sa2ajj:
Good point.  I'll revise it within next couple of hours.

sa2ajj:
After the codebase, I'd say it's too early to remove the interface description: there's old `status` code that uses it and there are test cases for that code.  I'd rather not remove the code as we want it ported to reporters API, it's just not happening now.

When the last bit is ported then it will be time to remove the interface.

ssavrim:
:+1:

bastien31:
So if this patch is just to remove "status"  and add "reporters" from __init__.__all__ => :+1: (maybe update the commit message accordingly)

PR #1735, opened 2015-06-22
git: skip fetch if the revision already exists

jaredgrubb:
Background: I am working on a "richer" git experience with buildbot, where I can reliably get tags, work with multiple remote servers (for the same repo), and share git-objects on the slaves to avoid multiple builders doing redundant network fetches.

The current Git step is not quite sufficient because it seems (correct me if I'm wrong) to take a conservative view of working with a git-server: to only fetch the minimal amount needed. 

So, this patch doesnt actually fix anything, so maybe it's not worth taking. But I thought I would post it.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -216,23 +216,7 @@ class Git(Source):
             yield self._fullCloneOrFallback()
             return
 
-        # test for existence of the revision; rc=1 indicates it does not exist
-        if self.revision:
-            rc = yield self._dovccmd(['cat-file', '-e', self.revision],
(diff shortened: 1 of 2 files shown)


sa2ajj:
PEP8:

```
master/buildbot/steps/source/git.py:368:18: E225 missing whitespace around operator
            if rc==RC_SUCCESS:
                 ^
1       E225 missing whitespace around operator
1
make: *** [pep8] Error 1
```

sa2ajj:
This looks a bit cleaner for me.

:+1:

jaredgrubb:
Pushed the pep8 fix.

PR #1738, opened 2015-06-23
display builds/buildrequests status in builder page

ssavrim:
This PR aims at displaying the status of builds in text format when user mouse over a build/buildrequest number in the page of a builder.

Here what has been done:
- display build status in text format on mouse over build number
- display buildrequest status in text format on mouse over buildrequest number
- add link to buildrequest page when user click on buildrequest number

I attached an example
build_status

### DIFF
--- a/www/base/src/app/builders/builder/builder.tpl.jade
+++ b/www/base/src/app/builders/builder/builder.tpl.jade
@@ -8,40 +8,45 @@
       | None
     table.table.table-hover.table-striped.table-condensed(ng-show='buildrequests.length>0')
       tr
-        td #
-        td Submitted At
-        td Owner
+        td(width='100px') #
+        td(width='150px') Submitted At
+        td(width='150px') Owner
         td Properties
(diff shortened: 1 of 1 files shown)


nam4dev:
I just thought about it, but why don't put both at the same time giving the information without any user effort ?

See below for example:

build_text_status

tardyp:
Because it is ugly?

Le mer. 24 juin 2015 12:10, Namgyal Brisson <notifications@github.com> a
écrit :

ssavrim:
Agree with @tardyp, displaying the build number and the status together is too overloaded.

bastien31:
According to me, displaying the verdict in tooltip is just for new users that are not yet familiar with the color code. After some working hours on this section, it will be "cristal clear" and so no need to overload the UI for that.

tardyp:
:+1: for the mouseover

bastien31:
:+1:

tardyp:
taking this ui improvement in beta

PR #1740, opened 2015-06-24
Betarelease

tardyp:
fix the version matching code to include b1 or p1 according to pep440.

So that we can generate sdist without having to force a export BUILDBOT_VERSION

for review, because I wonder why this wasn't already working like that. I might miss something


jaredgrubb:
LGTM

ssavrim:
:1:

PR #1745, opened 2015-06-26
Redirect status

kviktor:
Works great with PNG status (for example: https://git.ik.bme.hu/circle/cloud)

### DIFF
--- a/master/buildbot/status/web/baseweb.py
+++ b/master/buildbot/status/web/baseweb.py
@@ -103,7 +103,7 @@ class WebStatus(service.MultiService):
      /builders/BUILDERNAME/builds/NUM/steps/STEPNAME/logs/LOGNAME: a StatusLog
      /builders/_all/{force,stop}: force a build/stop building on all builders.
      /buildstatus?builder=...&number=...: an embedded iframe for the console
-     /png?buildername=...&number=...&size=... a png resource with build info
+     /png?builder=...&number=...&size=... a png resource with build info
      /redirect?builder=...&revision=... a redirect to the build that built the
                                         given revision
      /changes : summarize all ChangeSources
--- a/master/buildbot/status/web/baseweb.py
+++ b/master/buildbot/status/web/baseweb.py
@@ -38,6 +38,7 @@ from buildbot.status.web.grid import GridStatusResource
 from buildbot.status.web.grid import TransposedGridStatusResource
 from buildbot.status.web.olpb import OneLinePerBuild
 from buildbot.status.web.pngstatus import PngStatusResource
+from buildbot.status.web.redirectstatus import RedirectStatusResource
 from buildbot.status.web.root import RootPage
 from buildbot.status.web.slaves import BuildSlavesResource
 from buildbot.status.web.status_json import JsonStatusResource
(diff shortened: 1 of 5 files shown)


master/buildbot/status/web/redirectstatus.py
```diff
@@ -0,0 +1,47 @@
...
+                number = build.getNumber()
+                url = "/builders/%(builder)s/builds/%(number)d" % {
+                    'builder': builder,
+                    'number': number,
+                }
+                return url
+        return "/waterfall"
+
+    def render(self, request):
+        url = self.get_url(request)
```
sa2ajj: [nit] is this really necessary? why not put `self.get_url(request)` directly to the call below?

master/buildbot/status/web/redirectstatus.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        revision = request.args.get("revision", [None])[0]
+
+        builder = request.args.get("builder", [None])[0]
+
+        if revision and builder in self.status.getBuilderNames():
+            build = self.status.getBuilder(builder).getBuild(0, revision)
+            if build:
+                number = build.getNumber()
+                url = "/builders/%(builder)s/builds/%(number)d" % {
```
sa2ajj: i'd suggest using `buildbot.status.web.path_to_build` here

master/buildbot/status/web/redirectstatus.py
```diff
@@ -0,0 +1,47 @@
...
+                    'builder': builder,
+                    'number': number,
+                }
+                return url
+        return "/waterfall"
```
sa2ajj: i do not think this is correct

sa2ajj:
Looks good otherwise

master/buildbot/status/web/baseweb.py
```diff
@@ -102,7 +103,9 @@ class WebStatus(service.MultiService):
      /builders/BUILDERNAME/builds/NUM/steps/STEPNAME/logs/LOGNAME: a StatusLog
      /builders/_all/{force,stop}: force a build/stop building on all builders.
      /buildstatus?builder=...&number=...: an embedded iframe for the console
-     /png?buildername=...&number=...&size=... a png resource with build info
+     /png?builder=...&number=...&size=... a png resource with build info
+     /redirect?builder=...&revision=... a redirect to the build that built the
+                                        given revision
```
jaredgrubb: "... to the *last* build that built ..." -- technically there could be more than one. (I assume it's last, maybe it's first?)

kviktor: True, in case of force build it redirects to the forced one.

sa2ajj:
@kviktor a few changes and we take it in! :)

sa2ajj:
@kviktor any updates?

And a rebase is required.

djmitche:
Is this forward-compatible to nine?  If not, I'm inclined to close, to avoid adding features in eight that aren't replicated in nine.

djmitche:
In fact, given that and over a month with no response, per http://trac.buildbot.net/wiki/StalePullRequestPolicy I'm going to close the PR.  The code is still here!

kviktor:
I don't know about forward-compatibility but function wise it's very similar to pngstatus and can probably be implemented the same way (last time I checked it's not in nine yet tho)

PR #1746, opened 2015-06-28
Fix for trac #3299

vlovich:
I tested this live that it resolves the bug but I do not understand why.  The test example written passes with & without this change.

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -253,6 +253,7 @@ class BuildStep(results.ResultComputingConfigMixin,
         'descriptionSuffix',
         'doStepIf',
         'hideStepIf',
+        'workdir',
     ]
 
     # 'parms' holds a list of all the parameters we care about, to allow
--- a/master/buildbot/test/unit/test_process_buildstep.py
+++ b/master/buildbot/test/unit/test_process_buildstep.py
@@ -962,6 +962,21 @@ class TestShellMixin(steps.BuildStepMixin,
         yield self.runStep()
 
     @defer.inlineCallbacks
+    def test_example_step_renderable_workdir(self):
+        @renderer
+        def rendered_workdir(_):
+            return '/alternate'
+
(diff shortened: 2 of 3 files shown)


vlovich:
OK.  Found a test that it fails without this fix.  Still unclear what the cause is but this fixes my problem.

tardyp:
:+1:

jaredgrubb:
Go Go GO!

vlovich:
Hmm.... I was hoping someone could provide some deeper insight into what's going wrong.  This doesn't seem to affect a basic BuildStep.  Only ones that inherit from ShellCommand.  I'm fine if this is the best fix but if there's something else going wrong I'd hate to paper over it...

tardyp:
I dont know exactly what is your misunderstanding. If you want a shell's attribute to be renderable you have to put it in the renderable list.
It looks like for some reason it was put out of that list by mistake

vlovich:
Well I've done a diff of process/buildstep.py & steps/shell.py & I don't see any changes to workdir (i.e it wasn't previously on a renderable list).  So I don't understand why it was working in 0.8 but is broken now in 0.9.

vlovich:
(diff between 0.8 & master)

tardyp:
You should see a lot of changes in workdir for those steps. I did sanify a
lot of the default workdir behaviour.


Le dim. 28 juin 2015 à 13:18, Vitali Lovich <notifications@github.com> a
écrit :

vlovich:
Yes, but I didn't find anything that would explain why it was render-able in the 0.8 branch but not anymore in the 0.9 branch.

sa2ajj:
@vlovich In eight `workdir` was an allowed parameter for the `ShellMixin` and the latter had renderables equal to the list of its allowed parameters.  In master `workdir` moved up and it was not included in any renderables.

vlovich:
Ah.  Cool.  That explains it.  Thanks.

PR #1747, opened 2015-06-28
comment out copyright notice for the template in use

sa2ajj:
I still want to keep it in the sourcre, while it does not have to go the resulting HTML.


tardyp:
Why not remove? + add a note about original author of template

sa2ajj:
That sounds good: add a note where?

tardyp:
I think a README inside qtile them would be enough

sa2ajj:
Good suggestion.  Updated the PR.

tardyp:
:+1:

PR #1749, opened 2015-06-29
Enhance slave status by displaying Not Connected/Idle/Running in buil…

cbocta:
…dslaves page

### DIFF
--- a/www/base/src/app/buildslaves/buildslaves.controller.coffee
+++ b/www/base/src/app/buildslaves/buildslaves.controller.coffee
@@ -17,9 +17,34 @@ class Buildslaves extends Controller
         buildbotService.all('masters').bind $scope,
             onchild: (master) ->
                 $scope.mastersById[master.masterid] = master
-        buildbotService.all('buildslaves').bind($scope)
+        buildbotService.all('buildslaves').bind $scope,
+            onchild: (buildslave) ->
+                buildslave.some('builds', order: '-started_at', complete: false).bind $scope,
+                dest:buildslave
 
         $scope.settings = bbSettingsService.getSettingsGroup("Slaves")
         $scope.$watch('settings', ->
             bbSettingsService.save()
         , true)
+
+        $scope.slavestatus = {}
+        # Updates the status of a given slave
+        updateSlaveStatus = (slave) ->
+            status = if slave.connected_to.length > 0 then "online" else "not connected"
+            # Looking in our scope fields if there's a match
+            if status == "online"
+                if slave.builds.length > 0
+                    status = "running"
+                else
+                    status = "idle"
+            $scope.slavestatus[slave.name] = status
+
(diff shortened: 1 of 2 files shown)


www/base/src/app/buildslaves/buildslaves.controller.coffee
```diff
@@ -17,9 +17,34 @@ class Buildslaves extends Controller
...
+                else
+                    status = "idle"
+            $scope.slavestatus[slave.name] = status
+
+        # Updates the status of all slaves
+        updateSlavesStatus = (a, b) ->
+            for slave in $scope.buildslaves
+                updateSlaveStatus(slave)
+
+        $scope.$watch('buildslaves', updateSlavesStatus, true)
```
tardyp: We should rather put a watch on connected_to for each slaves, rather than this implementation that will watch everything, and refresh everything all the time

cbocta: I understand the need to narrow watching to only what need to be actually watch, but connected_to is not sufficient because we need to differentiate idle from running

tardyp: ahh I did not see that. hum I still would prefer to have a watch per slave, and not for the whole list.

www/base/src/app/buildslaves/buildslaves.controller.coffee
```diff
@@ -17,9 +17,34 @@ class Buildslaves extends Controller
         buildbotService.all('masters').bind $scope,
             onchild: (master) ->
                 $scope.mastersById[master.masterid] = master
-        buildbotService.all('buildslaves').bind($scope)
+        buildbotService.all('buildslaves').bind $scope,
+            onchild: (buildslave) ->
+                buildslave.some('builds', order: '-started_at', complete: false).bind $scope,
+                dest:buildslave
```
tardyp: $scope.$watch(buildslave, updateSlaveStatus, true)

www/base/src/app/buildslaves/buildslaves.tpl.jade
```diff
@@ -9,7 +9,7 @@
             tr(ng-repeat='buildslave in buildslaves',
                ng-show="settings.show_old_slaves.value || buildslave.configured_on.length > 0")
                 td {{buildslave.name}}
-                td {{buildslave.connected_to.length}} connection
+                td {{slavestatus[buildslave.name]}}
```
tardyp: buildslave.status

djmitche:
@cbocta any chance you'll have some time to wrap this up?

cbocta:
not this week, I 'd try to update code next week

cbocta:
Unfortunatelly not enough time! :-(, I prefere to give an ETA... so really stalled

sa2ajj:
Let's give it a month starting today.

PR #1751, opened 2015-06-29
Schedulers: let codebases be specified as a list

jaredgrubb:
I've always been annoyed that schedulers require a list of dictionaries for the 'codebase' parameter. This is espcially bad in eight (I'll do another patch there), but it's just kinda dumb that users must do this:

    SingleBranchScheduler(..., 
                          codebases={ 'foo': {}, 'bar': {} }) # nine
    SingleBranchScheduler(..., 
                          codebases={ 'foo': { 'repository':'' }, 
                                      'bar': { 'repository':'' }}) # eight

rather than just:

    SingleBranchScheduler(..., codebases=['foo', 'bar'])

### DIFF
--- a/master/buildbot/schedulers/base.py
+++ b/master/buildbot/schedulers/base.py
@@ -66,19 +66,20 @@ class BaseScheduler(ClusteredService, StateMixin):
         known_keys = set(['branch', 'repository', 'revision'])
         if codebases is None:
             config.error("Codebases cannot be None")
+        elif isinstance(codebases, list):
+            codebases = dict((codebase, {}) for codebase in codebases)
+        elif not isinstance(codebases, dict):
+            config.error("Codebases must be a dict of dicts, or list of strings")
         else:
-            if not isinstance(codebases, dict):
-                config.error("Codebases must be a dict of dicts")
-            else:
-                for codebase, attrs in codebases.iteritems():
-                    if not isinstance(attrs, dict):
-                        config.error("Codebases must be a dict of dicts")
-                    else:
-                        unk = set(attrs) - known_keys
-                        if unk:
-                            config.error(
-                                "Unknown codebase keys %s for codebase %s"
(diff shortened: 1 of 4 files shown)


jaredgrubb:
Hmm.. there was supposed to be a test here. Let me find that.

sa2ajj:
I thought that `repository` key must be present...  On the other hand, it's usually specified for the source step...

jaredgrubb:
In eight, it must be (and the default value has it set), but in nine it doesnt seem to be needed (and the default value does _not_ have it set). I'll be doing a eight-patch after this that WILL have it set.

sa2ajj:
Good

bastien31:
You shall update /master/docs/manual/cfg-schedulers.rst to mention this simplest usage.

jaredgrubb:
Alright, sorry for the delay. Docs updated. Tests pass and new one for the new form.

bastien31:
:+1:

sa2ajj:
Do we have a problem with Python 2.6?

master/buildbot/schedulers/base.py
```diff
@@ -66,19 +66,23 @@ def __init__(self, name, builderNames, properties=None,
         known_keys = set(['branch', 'repository', 'revision'])
         if codebases is None:
             config.error("Codebases cannot be None")
+        elif isinstance(codebases, list):
+            codebases = {
```
sa2ajj: Looks like this construct is not compatible with Python 2.6.

jaredgrubb: Yeh, looks like dict-comprehensions are 2.7+. I'll patch it.

jaredgrubb:
Pushed a fix; should be good on Py 2.6 now.

jaredgrubb:
Anything left to do here or? I have rebased this.

PR #1756, opened 2015-07-04
Jgrubb nine docs

jaredgrubb:
I just (barely) started to get one of my masters ported up to nine and noticed some inaccurate and out-dated information in the docs.

Also, I started a section on eight-to-nine migration notes to start a porting guide. There's only one bullet there, but it's a start. Take a look and see if you like that, or whether you think it should go elsewhere (or even as a separate section/doc)

### DIFF
--- a/master/buildbot/scripts/sample.cfg
+++ b/master/buildbot/scripts/sample.cfg
@@ -18,9 +18,8 @@ c = BuildmasterConfig = {}
 c['slaves'] = [buildslave.BuildSlave("example-slave", "pass")]
 
 # 'protocols' contains information about protocols which master will use for
-# communicating with slaves.
-# You must define at least 'port' option that slaves could connect to your master
-# with this protocol.
+# communicating with slaves. You must define at least 'port' option that slaves 
+# could connect to your master with this protocol.
 # 'port' must match the value configured into the buildslaves (with their
 # --master option)
 c['protocols'] = {'pb': {'port': 9989}}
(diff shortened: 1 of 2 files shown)


master/docs/manual/cfg-www.rst
```diff
@@ -119,10 +119,10 @@ The available classes are described here:
 
     ::
 
-        from buildbot.plugins import util
+        from buildbot.www import auth
         c['www'] = {
             # ...
-            'auth': util.UserPasswordAuth({"homer": "doh!"}),
```
tardyp: It should really be in the plugin namespace

jaredgrubb: "should be" meaning it is and my patch is wrong, or that we should move it? Right now it's in "master/buildbot/www/auth.py"

tardyp: http://docs.buildbot.net/latest/manual/plugins.html

https://github.com/buildbot/buildbot/blob/master/master/setup.py

jaredgrubb: Ok, but then you go to this page:
http://docs.buildbot.net/latest/manual/cfg-www.html#web-authentication

And it says the class is "buildbot.www.auth.UserPasswordAuth", so that's what I was basing my patch on... As I was reading the docs, the "util" form never was explained and I assumed the text snippets (that I put in this patch) were all wrong.

jaredgrubb: Note that I personally dont care which way it is, but the docs are confusing.

tardyp: Yes, the class is in www.auth, but it is also available in plugin.util as per setup.py. This allows us to move the class and be happy knowing the users are supposed to use the plugin namespace to important it.

I agree this can be confusing.

master/docs/manual/cfg-www.rst
```diff
@@ -119,10 +119,10 @@ The available classes are described here:
...
+            'auth': auth.UserPasswordAuth({"homer": "doh!"}),
         }
 
 .. py:class:: buildbot.www.auth.HTPasswdAuth(passwdFile)
```
jaredgrubb: So should we change this to "buildbot.plugins.util.HTPasswdAuth" and so on?

tardyp: could be.
@sa2ajj  as you introduced plugin namespace you might have an opinion about it?

sa2ajj: We have certain _entitites_ in Buildbot, those are exposed as `buildbot.plugins.entity`, the rest goes under `buildbot.plugins.utils`...

So in this case it should be `buildbot.plugins.utils.HTPasswdAuth` and so on...

jaredgrubb: Ok let me prepare a docs change like that then...

sa2ajj: After a bit of thinking (sorry), I believe the best will be to put it in a separate namespace: `buildbot.plugins.utils.www.HttPasswdAuth` etc...

sa2ajj:
How do we proceed with this PR?  Any help needed?

jaredgrubb:
Alright I removed the changes to the class paths ... sounds like there's work that needs to be done there outside of docs, so I'll leave it be and we can get the easy part of this patch in.

master/buildbot/scripts/sample.cfg
```diff
@@ -73,24 +72,22 @@ c['builders'].append(
 
 # 'status' is a list of Status Targets. The results of each build will be
 # pushed to these targets. buildbot/status/*.py has a variety to choose from,
-# including web pages, email senders, and IRC bots.
+# like IRC bots.
 
 c['status'] = []
```
sa2ajj: shouldn't we use `reporters` somehow??

(I need to check current `master`)

sa2ajj:
Let's start :)

PR #1758, opened 2015-07-06
Update exceptions to "as"

jvlomax:
This will break python 2.5 compatibility if that is an issue


djmitche:
Our current requirements are:
```
Python: http://www.python.org

  Buildbot requires Python-2.6 or later on the master, although Python-2.7 is recommended.
  The slave runs on Python-2.5.
```

sa2ajj:
:+1:

jaredgrubb:
LGTM

PR #1759, opened 2015-07-06
Triggerable: should allow reason to be set

jaredgrubb:
Will forward-port this to nine after merge.


sa2ajj:
:+1:

tardyp:
shouldn't the reason be inherited from the parent?

jaredgrubb:
Maybe, but it never has been. I wanted an override such that I could associate the triggered build with the parent build, but Triggerable unconditionally overwrites the reason

tardyp:
looking at the impl, I dont think it should be complicated to implement the inherit. 
Trigger step would have to pass its reason.
Anyway, I think we can merge this like this. 

:+1:

PR #1763, opened 2015-07-09
fix for #3223

byilmazturk:
added identifier checks for slaves while running checkconfig http://trac.buildbot.net/ticket/3223

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -538,13 +538,20 @@ class MasterConfig(util.ComparableMixin):
                 error(msg)
                 return
 
-            if sl.slavename in ("debug", "change", "status"):
-                msg = "slave name '%s' is reserved" % sl.slavename
-                error(msg)
-            
-            if not util.identifiers.isIdentifier(50, sl.slavename):
-                msg = "Buildslave name %r is not a 50-character identifier" % (sl.slavename,)
-                error(msg)
+            def validate(slavename):
+                if sl.slavename in ("debug", "change", "status"):
+                    yield "slave name %r - is reserved" % slavename
+                if not isinstance(slavename, unicode):
+                    yield "slave name %r - is not a unicode string" % slavename
+                if not util.identifiers.ident_re.match(slavename):
+                    yield "slave name %r - is not an identifier" % slavename
+                if len(slavename) < 1:
+                    yield "slave name %r - identifiers cannot be an empty string" % slavename
+                if len(slavename) > 50:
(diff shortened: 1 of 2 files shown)


master/buildbot/config.py
```diff
@@ -541,6 +541,10 @@ def load_slaves(self, filename, config_dict):
             if sl.slavename in ("debug", "change", "status"):
                 msg = "slave name '%s' is reserved" % sl.slavename
                 error(msg)
+            
+            if not util.identifiers.isIdentifier(50, sl.slavename):
+                msg = "Buildslave name %r is not a 50-character identifier" % (sl.slavename,)
+                error(msg)
```
jaredgrubb: This error makes it sound like it must be 50 chars, not something at most 50 chars... maybe "Buildslave name %r must be an identifier of at most 50 chars" or something?

byilmazturk: Actually I copied error messages from https://github.com/buildbot/buildbot/blob/master/master/buildbot/data/buildslaves.py#L115 . I can change both of them, but I see that isIdentifier function does not control only length of slavename. Slavename can also break some other rules. May it be good to print all of those rules, if name breaks one of them?

sa2ajj: Sounds like a good idea: guessing over a cryptic error message is not really a productive way of spending time. :smile:

djmitche: You could also link to the docs..

djmitche:
Any chance you'll have time to finish this up?

byilmazturk:
sorry, I do not look and answer previous week because of holiday. I am going to check this now.

master/buildbot/config.py
```diff
@@ -538,8 +539,19 @@ def load_slaves(self, filename, config_dict):
...
-            if sl.slavename in ("debug", "change", "status"):
-                msg = "slave name '%s' is reserved" % sl.slavename
+            def validate(slavename):
+                if slavename in ("debug", "change", "status"):
+                    yield "slave name '%s' is reserved" % slavename
+                if not isinstance(slavename, unicode):
+                    yield "slave name %r - is not a unicode string" % slavename
+                if not util_identifiers.ident_re.match(slavename):
+                    yield "slave name %r - is not an identifier" % slavename
+                if len(slavename) < 1:
```
sa2ajj: should not this be 'if not slavename:` instead?

sa2ajj:
LGTM otherwise

tardyp:
:+1:

PR #1764, opened 2015-07-09
Add python3 compatability docs

jvlomax:
This is an early version to cover the work done on buildbot so far. More will be added

### DIFF
--- a/master/docs/developer/py3-compat.rst
+++ b/master/docs/developer/py3-compat.rst
@@ -1,0 +1,164 @@
+Python3 compatibility
+=====================
+
+
+A good place to start looking for advice to ensure that any code is compatible with both Python-3.x and Python2.6/2.7 is too look at the python-future `cheat sheet <http://python-future.org/compatible_idioms.html>`_ .
+Buildbot uses the python-future library to ensure compatibility with both Python2.6/2.7 and Python3.x.
+
+Imports
+-------
+All ``__future__`` import have to happen at the top of the module, anything else is seen as a syntax error.
+All imports from the python-future package should happen below ``__future__`` imports, but before any other.
+
+Yes::
+
+    from __future__ import print_function
+    from builtins import basestring
+
+No::
+
+    from twisted.application import internet
+    from twisted.spread import pb
+    from builtins import str
+    from __future__ import print_function
+
+Dictionaries
+------------
+In python3, ``dict.iteritems`` is no longer available.
+While ``dict.items()`` does exist, it can be memory intensive in python2.
+For this reason, please use the python.future function ``iteritems()``.
+
+Example::
+
+    d = {"cheese": 4, "bread": 5, "milk": 1}
+    for item, num in d.iteritems():
+        print("We have {} {}".format(num, item))
+
+should be written as::
+
+    from future.utils import iteritems
+    d = {"cheese": 4, "bread": 5, "milk": 1}
+    for item, num in iteritems(d):
+        print("We have {} {}".format(num, item))
+
+This also applies to the similar methods ``dict.itervalues()`` and ``dict.values()``, which have an equivalent ``itervalues()``.
+
+If a list is required, please use ``list(iteritems(dict))``.
+This is for compatibility with the six library.
+
+For iterating over dictionary keys, please use ``for key in dict:``.
+For example::
+
+    d = {"cheese": 4, "bread": 5, "milk": 1}
+    for item in d:
+        print("We have {}".format(item))
+
+Similarly when you want a list of keys::
+
+    keys = list(d)
+
+New-style classes
+-----------------
+All classes in Python3 are newstyle, so any classes added to the code base must therefore be new-style.
+This is done by inheriting ``object``
+
+Old-style::
+
+    class Foo:
+        def __init__(self, bar)
+            self.bar = bar
+
+new-style::
+
+    class Foo(object):
+        def __init__(self, bar)
+            self.bar = bar
+
+When creating new-style classes, it is advised to import ``object`` from the builtins module.
+The reasoning for this can be read `in the python-future documentation <http://python-future.org/changelog.html#newobject-base-object-defines-fallback-py2-compatible-special-methods>`_
+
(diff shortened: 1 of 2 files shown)


sa2ajj:
So this PR is not to merge?  If so, please add `[WIP]` to the title (there's a button called `Edit` above :))

master/docs/developer/py3-compat.rst
```diff
@@ -0,0 +1,74 @@
+Python3 compatibility
+=====================
+
+
+A good place to start looking for advice to ensure that any code is compatible with both python-3.x and python2.7/2.6
```
sa2ajj: I believe Python is spelled with capital *P*

master/docs/developer/py3-compat.rst
```diff
@@ -0,0 +1,74 @@
+Python3 compatibility
```
sa2ajj: General comment "one sentence per line" is our policy.

master/docs/developer/py3-compat.rst
```diff
@@ -0,0 +1,74 @@
...
+    b'This is a bytestring, a string for computers to read'
+
+Print statements
+----------------
+Print statements are gone in python3. Please import ``from __future__ import print_function`` at the very top
+of the module to enable use of python3 style print functions
+
+Exceptions
+----------
+All exceptions should be written with the ``as`` statement. Before::
```
sa2ajj: "should be" -> "must"?

master/docs/developer/py3-compat.rst
```diff
@@ -0,0 +1,74 @@
...
+
+    try:
+        number = 5/0
+    except ZeroDivisionError, e:
```
sa2ajj: I'd suggest `err` instead of `e` here.

master/docs/developer/py3-compat.rst
```diff
@@ -0,0 +1,74 @@
...
+----------------
+Print statements are gone in python3. Please import ``from __future__ import print_function`` at the very top
+of the module to enable use of python3 style print functions
+
+Exceptions
+----------
+All exceptions should be written with the ``as`` statement. Before::
+
+    try:
+        number = 5/0
```
sa2ajj: this expression should be `number = 5 / 0`

master/docs/developer/py3-compat.rst
```diff
@@ -0,0 +1,74 @@
...
+    except ZeroDivisionError, e:
+        print(e.msg)
+
+After::
+
+    try:
+        number = 5/0
+    except ZeroDivisionError as e:
+        print(e.msg)
+
```
sa2ajj: No need for an empty line at the end of the file.

master/docs/developer/py3-compat.rst
jvlomax: I'm mirroring the python official documentation, but I would be happy to change it :)

sa2ajj:
Otherwise, LGTM

master/docs/developer/py3-compat.rst
jvlomax: Actually there is at the moment. All code blocks should end with a newline :). But if I change it to have more text, I'll make sure that there is no newline at the end

sa2ajj: Thanks

master/docs/developer/py3-compat.rst
sa2ajj: That's what I do not like about Python: there's PEP-8, but they themselves do not follow it in core library nor in the docs :/

master/docs/developer/py3-compat.rst
sa2ajj: as @djmitche pointed out, https://github.com/buildbot/buildbot/blob/master/master/docs/developer/style.rst

jaredgrubb:
This is a good start... why WIP? Why not merge it?

master/docs/developer/py3-compat.rst
jvlomax: I'm a little unsure about the entire section to be honest. While we do want a clear distinction between unicode and bytestrings, I'm not sure we are ready to enforce it yet. I might remove the entire section

jvlomax:
@jaredgrubb I'm wondering that a little myself. I'm very unsure about the bytestring/unicode section, and I might remove it before I request a merge. But there is no reason to have it as WIP I guess, it only covers work that is already done, and they should be done that way from now on

master/docs/developer/py3-compat.rst
sa2ajj: I think it's a good idea to keep it.

As for "should" vs "must", I'd say that _new_ code it can be 'must'.

As for enforcing, I wonder if there's a tool to perform the check.

sa2ajj:
@jvlomax what doubts do you have about bytestring/unicode section?

jvlomax:
Guess It won't make any difference while buildbot is still py2 only. I'm just wondering if it might be smart to put it into the documentation after the existing code has been made to handle bytestring and unicode properly

sa2ajj:
I'd rather add a note: this way it will not be forgotten and it certainly provides useful information.

master/docs/developer/py3-compat.rst
jvlomax: There is a tool in the python-future package that might help. If you run ``futurize -u`` it will ensure that ``from __ future__ import unicode literals`` is in all applicable files. This will then make all string not prefaced with ``b``, become unicode strings.

I don't know of any tools that will ensure all string have a designation though

jvlomax:
I have made the documentation cover the work done so far. Would this be ok to merge now?

sa2ajj:
LGTM, except for the empty line at the end of the file :)

I'd suggest to put a comment
```python

# empty line above is required to make the code example properly handled.
```

And please change the subject of the PR: remove `[WIP]`

jvlomax:
Empty newline is gone, there is now text **without** a trailing newline

sa2ajj:
Yay! :)

PR #1765, opened 2015-07-09
Update git_gerrit.cfg

shaurin:
fixed returning True on Code-Review=-2 in Approvals (of comment-added)


master/docs/examples/git_gerrit.cfg
```diff
@@ -54,7 +54,7 @@ c['change_source'] = changes.GerritChangeSource(gerrit_url, gerrit_user,
 def change_code_review_plus_2(change):
     if "event.approvals" in change.properties:
         for a in change.properties["event.approvals"]:
-            if "Code-Review" in a["type"] and "2" in a["value"]:
+            if "Code-Review" in a["type"] and a["value"] == "2":
```
tardyp: I think some gerrit version might say "+2" in a["value"]. probably we should do 

int(a["value"]) == 2

djmitche:
@shaurin any chance you'll find some time to finish this up?

PR #1766, opened 2015-07-10
Use 'Build' as property source for owners

Riziero:
f owners:
    self.setProperty('owners', owners, self.reason)

Is this intended? The source of the property is the reason?
Should not it be 'Build' since it's happening in class Build?
It has not changed since Eight so I might be wrong, but right now the result on the web UI is horrible.

@tardyp :  I agree it makes few sense


sa2ajj:
:+1:

jaredgrubb:
Looks good!

benallard:
looks good to me.

tardyp:
:+1:

PR #1767, opened 2015-07-11
Uicustomization

tardyp:
In eight, you could customize the templates via the template directory. This is similar implementation for angularjs's templates

``custom_templates_dir``
    This directory will be parsed for custom angularJS templates to replace the one of the original website templates.
    if the directory string is relative, it will be joined to the master's basedir.
    Either ``*.jade`` files or ``*.html`` files can be used, and will be used to override ``views/<filename>.html`` templates in the angularjs templateCache.
    Unlike with the regular nodejs based angularjs build system, python only jade interpreter is used to parse the jade templates.
    ``pip install pyjade`` is be required to use jade templates.
    You can also override plugin's directives, but they have to be in another directory.

    .. code-block:: none

        # replace the template whose source is in:
        # www/base/src/app/builders/build/build.tpl.jade
        build.jade

        # replace the template whose source is in
        # www/console_view/src/module/view/builders-header/buildersheader.tpl.jade
        console_view/buildersheader.html


master/buildbot/www/config.py
```diff
@@ -39,11 +41,51 @@ def reconfigResource(self, new_config):
...
+                template_name = os.path.basename(root) + "/views/%s.html"
+            for f in files:
+                fn = os.path.join(root, f)
+                basename, ext = os.path.splitext(f)
+                if ext not in allowed_ext:
+                    continue
+                if ext == ".html":
+                    with open(fn) as f:
+                        html = f.read().strip()
+                if ext == ".jade":
```
sa2ajj: maybe just `else`?

sa2ajj:
:+1:

master/buildbot/test/unit/test_www_config.py
```diff
@@ -38,7 +39,7 @@ def test_render(self):
         rsrc.jinja = mock.Mock()
         template = mock.Mock()
         rsrc.jinja.get_template = lambda x: template
-        template.render = lambda configjson, config: configjson
+        template.render = lambda configjson, config, custom_templates: configjson
```
sa2ajj: I wonder if we really need a `lambda` here...

sa2ajj: On the other hand, it's a test, so no worries

master/buildbot/test/unit/test_www_config.py
```diff
@@ -60,3 +61,27 @@ def test_render(self):
         exp = '{"titleURL": "http://buildbot.net", "versions": %s, "title": "Buildbot", "auth": {"name": "NoAuth"}, "user": {"anonymous": true}, "buildbotURL": "h:/a/b/", "multiMaster": false, "port": null}'
         exp = exp % vjson
         self.assertIn(res, exp)
+
+    def test_parseCustomTemplateDir(self):
+        exp = {'views/builds.html': json.dumps('<div>\n</div>')}
+        try:
+            import pyjade
```
sa2ajj: why do we import `pyjade` here at all?

master/buildbot/www/config.py
```diff
@@ -39,11 +41,51 @@ def reconfigResource(self, new_config):
         if isinstance(self.config.get('versions'), list):
             versions += self.config['versions']
 
+        template_dir = self.config.get('custom_templates_dir')
+        if template_dir is not None:
```
sa2ajj: [nit] reverse the condition and `then` and `else` parts?

master/buildbot/www/config.py
```diff
@@ -39,11 +41,51 @@ def reconfigResource(self, new_config):
...
+            import pyjade
+            allowed_ext.append(".jade")
+        except ImportError:  # pragma: no cover
+            log.msg("pyjade not installed. Ignoring .jade files from %s" % (template_dir,))
+            pyjade = None
+        for root, dirs, files in os.walk(template_dir):
+            if root == template_dir:
+                template_name = "views/%s.html"
+            else:
+                template_name = os.path.basename(root) + "/views/%s.html"
```
sa2ajj: hmm... `os.path.join`?

master/buildbot/www/config.py
```diff
@@ -39,11 +41,51 @@ def reconfigResource(self, new_config):
...
+        allowed_ext = [".html"]
+        try:
+            import pyjade
+            allowed_ext.append(".jade")
+        except ImportError:  # pragma: no cover
+            log.msg("pyjade not installed. Ignoring .jade files from %s" % (template_dir,))
+            pyjade = None
+        for root, dirs, files in os.walk(template_dir):
+            if root == template_dir:
+                template_name = "views/%s.html"
```
sa2ajj: I'm not 100% sure here, but `os.path.join('views', '%s.html')` might be better here.

master/buildbot/www/config.py
```diff
@@ -94,5 +136,6 @@ def toJson(obj):
 
         tpl = self.jinja.get_template('index.html')
         tpl = tpl.render(configjson=json.dumps(config, default=toJson),
+                         custom_templates=self.custom_templates,
```
sa2ajj: I seem to have missed how this part works

nam4dev:
:+1:

master/buildbot/www/config.py
```diff
@@ -39,11 +41,51 @@ def reconfigResource(self, new_config):
         if isinstance(self.config.get('versions'), list):
```
nam4dev: A bit of optimization ;)
```
vs = self.config.get('versions')
if isinstance(vs, list):
    versions += vs
```

This avoid to access twice the dict

master/buildbot/www/config.py
```diff
@@ -39,11 +41,51 @@ def reconfigResource(self, new_config):
...
             versions += self.config['versions']
 
+        template_dir = self.config.get('custom_templates_dir')
```
nam4dev: What about something more *dry & kiss* ?

```
self.custom_templates = {}
template_dir = self.config.pop('custom_templates_dir', None)
if template_dir:
    template_dir = os.path.join(self.master.basedir, template_dir)
    ...
```

*Avoiding `else` and `del` statements :)*

master/buildbot/www/config.py
tardyp: explicit is better than implicit

master/buildbot/www/config.py
tardyp: we really want '/' in those strings

master/buildbot/www/config.py
tardyp: indeed, this is a bit tricky. I added some comments

tardyp:
Hopefully, I have addressed your comments

master/buildbot/www/config.py
sa2ajj: ok

sa2ajj:
Yes, we are good to go.

PR #1770, opened 2015-07-13
Allow boto to handle ec2 latent slave auth, deprecate custom credentials file

apilloud:
The Buildbot EC2 latent slave code currently uses a custom '.ec2/aws_id' file to store credentials and asserts that credentials are provided explicitly or via this file. This change deprecates the custom file and delegates authentication to boto rather than asserting. This adds support for authenticating via [IAM Roles for EC2 Instances](https://aws.amazon.com/about-aws/whats-new/2012/06/11/Announcing-IAM-Roles-for-EC2-instances/), which was added to Boto in 2.5.1 (boto/boto#811).  It also adds support for the [standardized AWS Credentials file](http://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs) which was added to Boto in 2.29.0 (boto/boto#2292).

Will follow up with a backport to eight once merged.

### DIFF
--- a/master/buildbot/buildslave/ec2.py
+++ b/master/buildbot/buildslave/ec2.py
@@ -122,7 +122,7 @@ class EC2LatentBuildSlave(AbstractLatentBuildSlave):
                 aws_id_file_path = os.path.join(home, '.ec2', 'aws_id')
                 if not os.path.exists(aws_id_file_path):
                     aws_id_file_path = None
-            if aws_id_file_path and os.path.exists(aws_id_file_path):
+            if aws_id_file_path:
                 log.msg('WARNING: EC2LatentBuildSlave is using deprecated '
                         'aws_id file')
                 with open(aws_id_file_path, 'r') as aws_file:
--- a/master/buildbot/buildslave/ec2.py
+++ b/master/buildbot/buildslave/ec2.py
@@ -119,9 +119,9 @@ class EC2LatentBuildSlave(AbstractLatentBuildSlave):
                 'supply both or neither of identifier, secret_identifier')
             if aws_id_file_path is None:
                 home = os.environ['HOME']
-                aws_id_file_path = os.path.join(home, '.ec2', 'aws_id')
-                if not os.path.exists(aws_id_file_path):
-                    aws_id_file_path = None
+                default_path = os.path.join(home, '.ec2', 'aws_id')
+                if os.path.exists(default_path):
+                    aws_id_file_path = default_path
             if aws_id_file_path:
                 log.msg('WARNING: EC2LatentBuildSlave is using deprecated '
                         'aws_id file')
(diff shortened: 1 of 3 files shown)


tardyp:
:+1:

master/buildbot/buildslave/ec2.py
```diff
@@ -120,15 +120,14 @@ def __init__(self, name, password, instance_type, ami=None,
...
-                    "Please supply your AWS access key identifier and secret "
-                    "access key identifier either when instantiating this %s "
-                    "or in the %s file (on two lines).\n" %
-                    (self.__class__.__name__, aws_id_file_path))
-            with open(aws_id_file_path, 'r') as aws_file:
-                identifier = aws_file.readline().strip()
-                secret_identifier = aws_file.readline().strip()
+                if not os.path.exists(aws_id_file_path):
+                    aws_id_file_path = None
+            if aws_id_file_path and os.path.exists(aws_id_file_path):
```
sa2ajj: I think the warning should be issued even if the specified file does not exist.

sa2ajj:
LGTM otherwise.

master/buildbot/buildslave/ec2.py
apilloud: @sa2ajj I dropped the os.path.exists. This will trigger the warning if the path is set or the default file exists. It will also error out if the path is set to a non-existent file.

master/buildbot/buildslave/ec2.py
```diff
@@ -120,15 +120,14 @@ def __init__(self, name, password, instance_type, ami=None,
...
-                    "access key identifier either when instantiating this %s "
-                    "or in the %s file (on two lines).\n" %
-                    (self.__class__.__name__, aws_id_file_path))
-            with open(aws_id_file_path, 'r') as aws_file:
-                identifier = aws_file.readline().strip()
-                secret_identifier = aws_file.readline().strip()
+                if not os.path.exists(aws_id_file_path):
+                    aws_id_file_path = None
+            if aws_id_file_path:
+                log.msg('WARNING: EC2LatentBuildSlave is using deprecated '
```
sa2ajj: You are still producing a warning only when the `aws_id_file_path` is provided and the corresponding file exists.

More importantly the code at line 120, adds the value if it's not present.

I would put after line 122
```python
else:
     log.msg(BIG FAT WARNING)
```

sa2ajj:
(sorry for a slow follow up)

master/buildbot/buildslave/ec2.py
apilloud: @sa2ajj I believe this already does what you want, but it might be clearer if I wasn't using `aws_id_file_path` as temporary storage between 120 and 124. I've added a new variable, the actual logic hasn't changed. The warning will trigger if `aws_id_file_path` is set or the default file exists.

sa2ajj:
@apilloud Thank you

PR #1773, opened 2015-07-15
Wampmq

tardyp:
This is a MQ implementation using [wamp](http://wamp.ws/)  protocol.
This implementation uses [Python Autobahn](http://autobahn.ws) wamp client library, and is fully asynchronous (no use of threads)
To use this implementation, you need a wamp router like [Crossbar](http://crossbar.io).
The implementation does not yet support wamp authentication yet.
This MQ allows buildbot to run in multi-master mode.

Note that this implementation also does not support message persistence across a restart of the master.
For example, if a change is received, but the master shuts down before the schedulers can create build requests for it, then those schedulers will not be notified of the change when the master starts again.

### DIFF
--- a/master/buildbot/mq/wamp.py
+++ b/master/buildbot/mq/wamp.py
@@ -22,6 +22,7 @@ from twisted.internet import defer
 from buildbot.mq import base
 from buildbot.util import service
 from buildbot.util import toJson
+from twisted.python import log
 
 import json
 
@@ -33,7 +34,8 @@ class WampMQ(service.ReconfigurableServiceMixin, base.MQBase):
         base.MQBase.__init__(self, master)
 
     def produce(self, routingKey, data):
-        self._produce(routingKey, data)
+        d = self._produce(routingKey, data)
+        d.addErrback(log.err, "Problem while producing message on topic " + repr(routingKey))
 
     @classmethod
     def messageTopic(cls, routingKey):
@@ -43,7 +45,8 @@ class WampMQ(service.ReconfigurableServiceMixin, base.MQBase):
     def _produce(self, routingKey, data):
         _data = json.loads(json.dumps(data, default=toJson))
         options = PublishOptions(exclude_me=False)
-        return self.master.wamp.publish(self.messageTopic(routingKey), dict(topic=routingKey, data=_data), options=options)
+        return self.master.wamp.publish(self.messageTopic(routingKey),
+                                        dict(topic=routingKey, data=_data), options=options)
 
     def startConsuming(self, callback, _filter, persistent_name=None):
         if persistent_name is not None:
@@ -78,7 +81,8 @@ class QueueRef(base.QueueRef):
             yield self.stopConsuming()
 
     def invoke(self, msg):
-        return base.QueueRef.invoke(self, msg[u'topic'], msg[u'data'])
+        # json will transform tuple into list, so we must convert it back to tuple
+        return base.QueueRef.invoke(self, tuple(msg[u'topic']), msg[u'data'])
 
     @defer.inlineCallbacks
     def stopConsuming(self):
--- a/master/buildbot/mq/wamp.py
+++ b/master/buildbot/mq/wamp.py
@@ -37,12 +37,12 @@ class WampMQ(service.ReconfigurableServiceMixin, base.MQBase):
 
     @classmethod
     def messageTopic(cls, routingKey):
-        routingKey = [key if key is not None else u"1" for key in routingKey]
-        return cls.NAMESPACE + u"." + u".".join(routingKey)
+        ifNone = lambda v, default: default if v is None else v
+        return cls.NAMESPACE + u"." + u".".join([ifNone(key, "") for key in routingKey])
 
     def _produce(self, routingKey, data):
         _data = json.loads(json.dumps(data, default=toJson))
-        options = PublishOptions(excludeMe=False)
+        options = PublishOptions(exclude_me=False)
         return self.master.wamp.publish(self.messageTopic(routingKey), dict(topic=routingKey, data=_data), options=options)
 
     def startConsuming(self, callback, _filter, persistent_name=None):
--- a/master/buildbot/mq/wamp.py
+++ b/master/buildbot/mq/wamp.py
@@ -12,7 +12,7 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-
+from __future__ import unicode_literals
 from autobahn.wamp.exception import TransportLost
 from autobahn.wamp.types import PublishOptions
 from autobahn.wamp.types import SubscribeOptions
@@ -28,7 +28,7 @@ import json
 
 
 class WampMQ(service.ReconfigurableServiceMixin, base.MQBase):
-    NAMESPACE = u"org.buildbot.mq"
+    NAMESPACE = "org.buildbot.mq"
 
     def __init__(self):
         base.MQBase.__init__(self)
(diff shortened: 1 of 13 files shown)


sa2ajj:
This is interesting :smile:

master/buildbot/wamp/connector.py
```diff
@@ -0,0 +1,140 @@
...
+            url=self.router_url,
+            extra=dict(master=self.master, parent=self),
+            realm=wamp.get('realm', 'buildbot'),
+            make=make,
+            debug=wamp.get('debug_websockets', False),
+            debug_wamp=wamp.get('debug_lowlevel', False),
+            debug_app=wamp.get('debug', False)
+        )
+
+        self.app.setServiceParent(self)
```
yetanotherion: 1. this will not trigger a stopService/startService ?
2. shouldn't you yield it ? (I'm kind of lost with the overall API for
startService
stopService
setServiceParent
disownParent
as some service Class return deferreds, and others don't.

master/buildbot/wamp/connector.py
```diff
@@ -0,0 +1,140 @@
...
+def make(config):
+    if config:
+        return MasterService(config)
+    else:
+        # if no config given, return a description of this WAMPlet ..
+        return {'label': 'Buildbot master wamplet',
+                'description': 'This contains all the wamp methods provided by a buildbot master'}
+
+
+class WampConnector(service.ReconfigurableServiceMixin, service.AsyncMultiService):
```
yetanotherion: Why not making it a BuildbotService ?

master/buildbot/wamp/connector.py
```diff
@@ -0,0 +1,140 @@
...
+from autobahn.wamp.exception import TransportLost
+
+from twisted.internet import defer
+from twisted.python import failure
+from twisted.python import log
+
+from buildbot.util import service
+
+
+class MasterService(ApplicationSession, service.AsyncMultiService):
```
yetanotherion: multiple inheritance ?

master/buildbot/mq/wamp.py
```diff
@@ -0,0 +1,96 @@
...
+
+from buildbot.mq import base
+from buildbot.util import service
+from buildbot.util import toJson
+from twisted.python import log
+
+import json
+
+
+class WampMQ(service.ReconfigurableServiceMixin, base.MQBase):
```
yetanotherion: why not a BuildbotService ?

yetanotherion:
Most of my concern comes from seeing 3 new services
created in 3 different ways.
It is quite painful to fix after (https://github.com/buildbot/buildbot/pull/1707)
Besides that it is fine for me :)

master/buildbot/wamp/connector.py
tardyp: good catch. It will trigger a startService

master/buildbot/wamp/connector.py
tardyp: Because it does not take its config from a peer object

It takes it from c['mq'], which is originaly designed to be a simple dictionary.

master/buildbot/wamp/connector.py
tardyp: BuildbotService is really for Services configured in the master.cfg. Here this is a internal service.

AsyncMultiService allows to manage the list of childrens, and lifecycle capability to the ApplicationSession.

We could separate it in attributes, but I think this would just make the thing more complicated.

master/buildbot/mq/wamp.py
tardyp: Same, this is the original design of the mq configuration. I think this is indeed not a bad idea to change it to BuildbotService eventually, but this will break a number of unit tests, so this should be seperated.

tardyp:
> It is quite painful to fix after (#1707)
I dont understand exactly what you mean.

sa2ajj:
(this needs to be rebase!)

master/docs/developer/mq.rst
```diff
@@ -105,11 +105,50 @@ Simple
...
+....
+
+.. py:module:: buildbot.mq.wamp
+
+.. py:class:: WampMQ
+
+    The :py:class:`WampMQ` class implements message-queueing using a wamp router.
+    This class translates the semantics of the buildbot mq api to the semantics of the wamp messaging system.
+    A notable difference is that autobahn API does not send the original topic to the callback, so we need to include it inside the data part, for the case of the wildcard subscriber.
+    Example message that is sent via wamp is:
```
meejah: You can get the topic passed to your callback via an optional kwarg.

You can get this kwarg passed to your callback by specifiying ``details_arg='details'`` in a SubscribeOptions -- this would cause a "details=" kwarg to be passed to your callback (which will be an ``EventDetails`` instance, which has a ``.topic`` member)

master/buildbot/wamp/connector.py
```diff
@@ -0,0 +1,140 @@
...
+        self.parent.service = self
+        self.parent.serviceDeferred.callback(self)
+
+    @defer.inlineCallbacks
+    def onLeave(self, details):
+        # XXX We don't handle crossbar reboot, or any other disconnection well.
+        # this is a tricky problem, as we would have to reconnect with expononential backoff
+        # re-subscribe to subscriptions, queue messages until reconnection.
+        # This is quite complicated, and I believe much better handled in autobahn
+        # It is possible that such failure is practically non-existant
```
meejah: FWIW, re-connection logic *is* coming in Autobahn, after an API for arbitrary transports for ApplicationRunner is figured out. If you're interested in this, see https://github.com/tavendo/AutobahnPython/issues/409 and also issue 295.

master/buildbot/wamp/connector.py
yetanotherion: the SchedulerManager takes the config from c['schedulers'], and we made it a BuildbotServiceManager which permitted to find/fix bugs.
I think uniformizing the way we handle services (all are either BuildbotService or BuildbotServiceManager's) has a non negligible ROI

master/buildbot/mq/wamp.py
yetanotherion: Ok, trac ?

tardyp: http://trac.buildbot.net/ticket/3316

master/buildbot/wamp/connector.py
tardyp: agreed. Should be done also for Pb configuration, and also probably other dict based config.

master/buildbot/mq/wamp.py
yetanotherion: great, thx

tardyp:
Thanks @meejah for review. I managed to make it work with details_arg.

I can't wait for reconnecting stuff. I've looked at the dev-branch, but I was a little bit too afraid by the lots of XXX in the code, to try it.

I think we can merge this version anyway.

master/buildbot/mq/wamp.py
```diff
@@ -0,0 +1,104 @@
...
+        base.MQBase.__init__(self)
+
+    def produce(self, routingKey, data):
+        d = self._produce(routingKey, data)
+        d.addErrback(log.err, "Problem while producing message on topic " + repr(routingKey))
+
+    @classmethod
+    def messageTopic(cls, routingKey):
+        ifNone = lambda v, default: default if v is None else v
+        return cls.NAMESPACE + u"." + u".".join([ifNone(key, "") for key in routingKey])
```
sa2ajj: [nit] this is not really readable :/

master/buildbot/mq/wamp.py
```diff
@@ -0,0 +1,104 @@
...
+        base.QueueRef.__init__(self, callback)
+        self.unreg = None
+
+    @defer.inlineCallbacks
+    def subscribe(self, service, _filter):
+        self.filter = _filter
+        self.emulated = False
+        options = dict(details_arg='details')
+        if None in _filter:
+            options["match"] = u"wildcard"
```
sa2ajj: [nit] Can we enable unicode literals and forget these `u'...'`?

sa2ajj:
Let me know if you are not going to address those nits and I'll merge it.

tardyp:
@se2ajj I think I addressed your comments.

sa2ajj:
Yay! :)

PR #1775, opened 2015-07-17
simplify the travis.rc

tardyp:
several goals:

- accelerate the travis builds by doing more parallelization
  Currently everything is in latest, which makes it very slow
- use the container based system (so we remove all sudo)
   nodejs is actually installed in all travis VMs, so we dont need to install it
- removed the hack to fetch more history, there is already a git: depth option in travisrc.
- removed the test of sqlalchem-migrate versions.The setup.py has a fixed version so it will force the upgrade
- Have a simpler config that runs on buildbot_travis (which does not support exporting shell variable from travisrc)

- run py3 tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,12 +60,19 @@ install:
   # Install python-future for py2/3 compatability
   # Install pyjade for custom templates tests
   - |
+      [ $TESTS = py3 ] || \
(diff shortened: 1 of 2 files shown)


sa2ajj:
I like it.

PR #1777, opened 2015-07-17
[Eight] Allow boto to handle ec2 latent slave auth, deprecate custom credentials file

apilloud:
This is a cherrypick of #1770 to eight.

The Buildbot EC2 latent slave code currently uses a custom '.ec2/aws_id' file to store credentials and asserts that credentials are provided explicitly or via this file. This change deprecates the custom file and delegates authentication to boto rather than asserting. This adds support for authenticating via [IAM Roles for EC2 Instances](https://aws.amazon.com/about-aws/whats-new/2012/06/11/Announcing-IAM-Roles-for-EC2-instances/), which was added to Boto in 2.5.1 (boto/boto#811).  It also adds support for the [standardized AWS Credentials file](http://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs) which was added to Boto in 2.29.0 (boto/boto#2292).

### DIFF
--- a/master/buildbot/buildslave/ec2.py
+++ b/master/buildbot/buildslave/ec2.py
@@ -110,6 +110,8 @@ class EC2LatentBuildSlave(AbstractLatentBuildSlave):
                 if not os.path.exists(aws_id_file_path):
                     aws_id_file_path = None
             if aws_id_file_path and os.path.exists(aws_id_file_path):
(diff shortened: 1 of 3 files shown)


sa2ajj:
thank you :)

PR #1780, opened 2015-07-21
base.py: support earlier buildbot.tac

yetanotherion:
Fixes
http://trac.buildbot.net/ticket/3318
that created a regression introduced by
201f0d9cbcffefb4062668c03ca95558f6a6b8c7

the regression being buildbot not supporting
buildbot.tac files containing
print someString
lines anymore.

### DIFF
--- a/master/docs/relnotes/0.9.0b1.rst
+++ b/master/docs/relnotes/0.9.0b1.rst
@@ -257,6 +257,11 @@ Deprecations, Removals, and Non-Compatible Changes
   Previously, it would return the original value.
   Instead, it now returns an array with the original value as the sole element.
 
+
+* ``buildbot.tac`` does not support ``print`` statements anymore. Such files should now use ``print``
+  as a function instead (see https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function
+  for more details). Note that this applies to both python2.x and python3.x runtimes.
+
 WebStatus
 .........
 


jaredgrubb:
As an advocate for the devil here, maybe asking people to update these files is a good thing? It may be this is the only Py3-compatible change we will make here, but if there are others, letting them not update could be a problem.

But I dont have an objection to this.

djmitche:
Agree, looking forward, `buildbot.tac` should be python-3 compatible too.  The error seems to be fairly self-explanatory, so perhaps just a release note would do the trick?

I suspect most users don't have a `print` statement in their tac files.

yetanotherion:
@djmitche @jaredgrubb as per your reviews, the code was removed and the release note updated

master/docs/relnotes/0.9.0b1.rst
```diff
@@ -257,6 +257,11 @@ Deprecations, Removals, and Non-Compatible Changes
   Previously, it would return the original value.
   Instead, it now returns an array with the original value as the sole element.
 
+
+* ``buildbot.tac`` does not support ``print`` statements anymore. Such files should now use ``print``
+  as a function instead (see https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function
+  for more details)
```
tardyp: I'd add a specific comment that this is even if you are in a python 2.x runtime

yetanotherion:
@tardyp note about python2.x added

tardyp:
:+1:

PR #1781, opened 2015-07-22
Update hgbuildbot.py

sscherfke:
I updated, cleaned up and extend the Mercurial hook for buildbot.

I tested it with Mercurial 2.8.2 (shipped with Ubuntu 14.04 LTS), Python 2.7 and buildbot 0.9.0b1. It should also work with Python 3 (if buildbot and twisted to).

Changes:

- It seems like it is no longer necessary to fork.  I get no errors and the hook works as expected.
- You can now extend PYTHONPATH via the hook config which allows you to use a buildbot-venv instead of installing buildbot to the system site-packages.
- Added documentation to the hook (because it was removed from the official docs).
- Misc. clean-up and reorganization in order to make the code a bit more readable.

### DIFF
--- a/master/contrib/hgbuildbot.py
+++ b/master/contrib/hgbuildbot.py
@@ -13,138 +13,228 @@
 #
 # Portions Copyright Buildbot Team Members
 # Portions Copyright 2007 Frederic Leroy <fredo@starox.org>
-
-# hook extension to send change notifications to buildbot when a changeset is
-# brought into the repository from elsewhere.
 #
-# See the Buildbot manual for configuration instructions.
-
-# WARNING
-# This code does not work with recent versions of Twisted and Mercurial.  It
-# was never a good idea to try to run Twisted code within Mercurial, and now it
-# doesn't work.  Use this code with caution.
 #
+# Documentation
+# =============
+#
+# Mercurial "changegroup" hook that notifies Buildbot when a number of
+# changsets is brought into the repository from elsewhere.
+#
+# Copy this file to ".hg/hgbuildbot.py" in the repository that should notify
+# Buildbot.
+#
+# Add it to the "[hooks]" section of ".hg/hgrc".  Also add a "[hgbuildbot]"
+# section with additional parameters, for example:
+#
+#     [hooks]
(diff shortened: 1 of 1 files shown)


master/contrib/hgbuildbot.py
```diff
@@ -13,138 +13,228 @@
...
+#
+# You can alternatively strip some of the repo path's components and prepend
+# a custom base URL instead.  For example, if you want to create an URL like
+# "https://code.company.com/myrepo", you must specify the following parameters:
+#
+#     baseurl = https://code.company.com/
+#     strip = 4
+#
+# This would strip everything until (and including) the 4th "/" in the repo's
+# path leaving only "myrepo" left.  This would then be append to the base URL.
```
sa2ajj: Thank you for providing an extensive documentation.

I believe one little thing is missing: what needs to be configured on the Buildbot side.  I believe a reference to [PBChangeSource](http://docs.buildbot.net/latest/manual/cfg-changesources.html#pbchangesource) would be appropriate.

master/contrib/hgbuildbot.py
```diff
@@ -156,24 +246,24 @@ def _printFailure(why):
...
-        if c == -1:
-            break
-        path = path[c + 1:]
-        count -= 1
-    return path
+    # and strip the *count* first slash
+    return path.split('/', count)[-1]
+
+
+def send_cs(res, s, c):
```
sa2ajj: We try to avoid this kind of variable names: `s` and `c`.  They are not descriptive.

If you'd replace them with `sender` and `changeset` respectively, the doc string is redundant.

sa2ajj:
:+1: otherwise

master/contrib/hgbuildbot.py
sscherfke: I left them as they were in the original script, but I can change them.

master/contrib/hgbuildbot.py
sscherfke: Good idea.

sa2ajj:
Let's wait for Travis...

sa2ajj:
Thank you!

PR #1783, opened 2015-07-23
remove coveralls from travis.yml

sa2ajj:
codecov seems to provide more information, so I think we'd stick with that.


tardyp:
👍

Le jeu. 23 juil. 2015 18:43, Mikhail Sobolev <notifications@github.com> a
écrit :

PR #1785, opened 2015-07-24
Include log contents in failed integration tests

djmitche:
This may help deciphering some flaky integration tests.


sa2ajj:
:+1:

(will merge after Travis has its say)

PR #1787, opened 2015-07-24
Implement retries for p4 sync command.

samhocevar:
Most of the code is similar to the implementation of `retry` in other source providers.

### DIFF
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -23,6 +23,7 @@ from buildbot.process import buildstep
 from buildbot.process.properties import Interpolate
 from buildbot.steps.source import Source
 from twisted.internet import defer
+from twisted.internet import reactor
 from twisted.python import log
 from types import StringType
 
@@ -141,7 +142,7 @@ class P4(Source):
         yield self._createClientSpec()
 
         # Then p4 sync #none
-        yield self._dovccmd(['sync', '#none'])
+        yield self._sync(['#none'])
 
         # Then remove directory.
         yield self.runRmdir(self.workdir)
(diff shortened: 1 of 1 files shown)


tardyp:
:+1:

tardyp:
btw. no tests?

master/buildbot/steps/source/p4.py
```diff
@@ -150,11 +151,11 @@ def full(self, _):
         if self.revision:
             if debug_logging:
                 log.msg("P4: full() sync command based on :base:%s changeset:%d", self._getP4BaseForLog(), int(self.revision))
-            yield self._dovccmd(['sync', '%s...@%d' % (self._getP4BaseForCommand(), int(self.revision))], collectStdout=True)
+            yield self._sync(['%s...@%d' % (self._getP4BaseForCommand(), int(self.revision))])
```
sa2ajj: Hmm... I thought revision would be specified with `#revision`.  From `p4 help revisions`:

```
            #rev - A revision number or one of the following keywords:
            #none - A nonexistent revision (also #0).
            #head - The current head revision (also @now).
            #have - The revision on the current client.
            @change - A change number: the revision as of that change.
            @client - A client name: the revision on the client.
            @label - A label name: the revision in the label.
            @date - A date or date/time: the revision as of that time.
                    Either yyyy/mm/dd or yyyy/mm/dd:hh:mm:ss
                    Note that yyyy/mm/dd means yyyy/mm/dd:00:00:00.
                    To include all events on a day, specify the next day.
```

sa2ajj:
I'd appreciate new tests as well.

Otherwise, LGTM.

master/buildbot/steps/source/p4.py
samhocevar: Perforce’s `#rev` is a per-file revision number, whereas `@change` is a global, per-commit revision number (also called changeset or changelist). It’s inherited from CVS. Since the code syncs whole directories, the `@` notation is correct here.

samhocevar:
I’ll try to write some tests after I learn how they work in Buildbot, but the current P4 module doesn’t have much coverage to get inspiration from.

sa2ajj:
We all will appreciate that.  Please do not hesitate to ask for help though.

master/buildbot/steps/source/p4.py
sa2ajj: Thanks for the clarification.

djmitche:
By the way, many of the other version-control systems have extensive tests.  P4 does not because we don't have a P4 installation with which to generate realistic example data.  If you can add that, it will be a great boon to Buildbot!

sa2ajj:
Side note: nowdays Perforce offers free installation for 10 users, I believe.

samhocevar:
Okay, I’ll try to get some company time to work on unit tests (which we’ll eventually need anyway since we have more Perforce-related fixes to contribute).

PR #1788, opened 2015-07-24
Implement retries for p4 sync command.

samhocevar:
This is the same patch as in pull request #1787. The difference is that we do not have a working `nine` version yet, so we could not test this version of the patch. I’m providing the code nonetheless for anyone else to review/test, and as soon as we deploy `nine` I’ll follow up on this.

### DIFF
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -23,6 +23,7 @@ from buildbot.process import buildstep
 from buildbot.process.properties import Interpolate
 from buildbot.steps.source import Source
 from twisted.internet import defer
+from twisted.internet import reactor
 from twisted.python import log
 from types import StringType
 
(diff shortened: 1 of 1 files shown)


sa2ajj:
[According to the policy](https://github.com/buildbot/botherders/blob/master/policies/eight.md) we first need to make it working for `master`.

master/buildbot/steps/source/p4.py
```diff
@@ -165,14 +166,37 @@ def mode_incremental(self, _):
...
         if debug_logging:
-            log.msg("P4:incremental() command:%s revision:%s", command, self.revision)
-        yield self._dovccmd(command)
+            log.msg("P4:incremental() args:%s revision:%s", args, self.revision)
+
+        yield self._sync(args)
+
+    def _sync(self, args):
+        abandonOnFailure = (self.retry[1] <= 0) if self.retry else True
+        d = self._dovccmd(['sync', *args],
```
sa2ajj: looks like you need to use `['sync'] + args`

djmitche:
We should be able to forward-port, and test, well enough once #1787 lands.  I'll close this for now, to keep conversation focused in one place.

PR #1789, opened 2015-07-25
Changed statistics service classes to use ABC.

prasoon2211:
Second item on the todo list. This is a very small commit.

I haven't added any tests here because those tests belong in the ABC module itself - testing that an ABC class will raise error when abstract methods aren't implemented in child classes is something that ABC would check in their test own test suite.

@sa2ajj @Stibbons @benallard 

Updated TODO list:

- [x] Style-based changes:
  * Dictionary intialization
  * Extra newlines
  * Private memebers prepended with `_`
  * Change some method names
  * Minor refactoring
    * Changes to some tests
    * storage_backends.py to be divided into files.
  * Some error handling for graceful exits in case of failure.

- [x] Making all superclasses use ABC.
  * Capture
  * CaptureBuildTimes
  * StatsStorageBase

- [ ] Add helper methods for quickly capturing data:
  * Capture a specific data for all builders (properties/times).
  * Capture all properties mathing a regex on all builders.
  * Capture all properties mathing a regex on a list of builders.
  * Capture times (start/end/duration) on a list of builders.

- [ ] Changes to the docs:
  * Fix the format (one sentence per line).
  * Document the changes from the above TODOs.
    * Write about ABC usage and helper methods
  * Add a new tutorial/chapter on the usage of this service.

### DIFF
--- a/master/buildbot/statistics/capture.py
+++ b/master/buildbot/statistics/capture.py
@@ -13,6 +13,8 @@
 #
 # Copyright Buildbot Team Members
 
+import abc
+
 from twisted.internet import defer
 from twisted.internet import threads
 
@@ -24,6 +26,8 @@ class Capture(object):
     """
     Base class for all Capture* classes.
     """
+
+    __metaclass__ = abc.ABCMeta
 
     def __init__(self, routingKey, callback):
         self.routingKey = routingKey
@@ -38,6 +42,7 @@ class Capture(object):
             "build_number": str(msg['number'])
         }
 
+    @abc.abstractmethod
     def consume(self, routingKey, msg):
         raise NotImplementedError
 
@@ -134,6 +139,10 @@ class CaptureBuildTimes(Capture):
         msg = "%s failed on build %s on builder %s." % (self.__class__.__name__,
                                                         build_data['number'], self._builder_name)
         return msg
+
+    @abc.abstractmethod
+    def _retValParams(self, msg):
+        pass
 
 
 class CaptureBuildStartTime(CaptureBuildTimes):
--- a/master/buildbot/statistics/capture.py
+++ b/master/buildbot/statistics/capture.py
@@ -44,7 +44,7 @@ class Capture(object):
 
     @abc.abstractmethod
     def consume(self, routingKey, msg):
-        raise NotImplementedError
+        pass
 
     @defer.inlineCallbacks
     def _store(self, post_data, series_name, context):
--- a/master/buildbot/test/unit/test_stats_service.py
+++ b/master/buildbot/test/unit/test_stats_service.py
@@ -154,13 +154,6 @@ class TestInfluxDB(TestStatsServicesBase, logging.LoggingMixin):
         svc.thd_postStatsValue("test", "test", "test")
         self.assertLogged("Service.*not initialized")
 
-    def test_storage_backend_base(self):
-        # Just initlize and run thd_postStatsValue
-        svc = StatsStorageBase()
-        r = svc.thd_postStatsValue("test", "test", "test")
-        assert isinstance(r, defer.Deferred)
-        assert r.result == None
-
 
 class TestStatsServicesConsumers(steps.BuildStepMixin, TestStatsServicesBase):
 
--- a/master/buildbot/test/unit/test_stats_service.py
+++ b/master/buildbot/test/unit/test_stats_service.py
@@ -55,6 +55,16 @@ class TestStatsServicesBase(unittest.TestCase):
         self.stats_service.stopService()
 
 
+class DummyStatsStorageBase(StatsStorageBase):
+
+    """
+    A dummy class to test intialization of StatsStorageBase.
+    """
+
+    def thd_postStatsValue(self, *args, **kwargs):
+        return defer.succeed(None)
+
+
 class TestStatsServicesConfiguration(TestStatsServicesBase):
 
     def test_reconfig_with_no_storage_backends(self):
@@ -153,6 +163,13 @@ class TestInfluxDB(TestStatsServicesBase, logging.LoggingMixin):
         svc._inited = False
         svc.thd_postStatsValue("test", "test", "test")
         self.assertLogged("Service.*not initialized")
+
+    def test_storage_backend_base_failure_on_init(self):
+        svc = DummyStatsStorageBase()
+
+        r = svc.thd_postStatsValue("test", "test", "test")
+        assert isinstance(r, defer.Deferred)
+        assert r.result == None
 
 
 class TestStatsServicesConsumers(steps.BuildStepMixin, TestStatsServicesBase):
--- a/master/buildbot/statistics/storage_backends/base.py
+++ b/master/buildbot/statistics/storage_backends/base.py
@@ -13,6 +13,8 @@
 #
 # Copyright Buildbot Team Members
 
+import abc
+
 from twisted.internet import defer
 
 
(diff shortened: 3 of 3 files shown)


tardyp:
ok for not creating unit tests. but not ok for breaking them :)

prasoon2211:
@tardyp I removed the test that was trying to instantiate a ABC and the tests pass now on my system. However, travis seems to be acting up again. The last commit, c18158b, is failing tests on this PR. But, this same commit is passing on my personal travis account: https://travis-ci.org/prasoon2211/buildbot/builds/72608684

So, the tests are actually passing.

sa2ajj:
I do not really like removing a test case without good explanation why it was done.

sa2ajj:
@jvlomax what about your patch regarding new style classes?

jvlomax:
@sa2ajj It is still in limbo. I was having some issues I could not figure out. Last I heard of it was a [buildbot meeting](https://supybot.buildbot.net/meetings/buildbot/2015/buildbot.2015-07-14-16.00.html) where @djmitche or someone with spare time would look at the patch and try to figure out what was making it fail. I don't see this patch creating any issues with it though

sa2ajj:
@jvlomax it's not about issues, I believe new style classes are a pre-requisite for [ABC](https://docs.python.org/2/library/abc.html)

prasoon2211:
@sa2ajj I removed the test because it was instantiating the base class - this I had done simply for coverage. Now that this base class is an ABC, it cannot be instantiated. Therefore, I removed the test.

jvlomax:
afik, it does not require new-style classes

sa2ajj:
@jvlomax: some tests _I_ run were complaining about old style classes :/

master/buildbot/statistics/capture.py
```diff
@@ -38,6 +42,7 @@ def _defaultContext(self, msg):
             "build_number": str(msg['number'])
         }
 
+    @abc.abstractmethod
     def consume(self, routingKey, msg):
```
nam4dev: When using **abc** module, you don't need to use `NotImplementedError` exception as it is the purpose of that module :)
Just add a docstring or a `pass` statement as you did below (_retValParams)

prasoon2211: Yes, of course. This method seems to have slipped my attention.

prasoon2211:
@tardyp @sa2ajj Is there anything left to do in this?

sa2ajj:
@prasoon2211 regarding the test, I believe it's the best if you keep it while handling the situation properly: I think it's the best to make sure that the best class can't be instantiated.

prasoon2211:
@sa2ajj Okay. So I tried adding the test to check that ABC initialization fails. It worked. But, pylint won't let me even commit such a change - here's the message:

```
(sandbox)[prasoon@jimmypage src]$ git commit -m "Re-added test for StatsStorageBase"
Validating the following commits:
c3f9ce8f2be48adab60147532d3681a994684851 Re-added test for StatsStorageBase
** Skipping JavaScript Tests **
** Skipping Python Tests **
-- checking formatting --
-- checking for release notes --
** HEAD~..HEAD does not add release notes **
-- checking import module convention in modified files --
-- running autopep8 --
-- running pep8 --
-- running pyflakes --
-- running pylint --
Warning: option include-ids is deprecated and ignored.
Warning: option symbols is deprecated and ignored.
************* Module buildbot.test.unit.test_stats_service
E:160,47: Abstract class 'StatsStorageBase' with abstract methods instantiated (abstract-class-instantiated)
** pylint failed **
-- not building docs, because it was not changed --

NO GOOD!
** Skipping JavaScript Tests (warning)
** Skipping Python Tests (warning)
** HEAD~..HEAD does not add release notes (warning)
** pylint failed
[todo-pr-2 c3f9ce8] Re-added test for StatsStorageBase
 1 file changed, 5 insertions(+)
```

Abstract classes can simply never be instantiated - we don't need to test this as I mentioned in the original message of this PR. Anyway, I think besides this, this is ready for a merge.

sa2ajj:
You can disable this particular check for this particular class.

We need a test since we rely on this particular functionality.

However, I wonder what others have to say.

Stibbons:
Indeed, abstract class should not be instantiated. That's why they are abstract.
Here is what google says: http://stackoverflow.com/questions/8086826/cant-instantiate-abstract-class-with-abstract-methods-on-class-that-shouldn

master/buildbot/test/unit/test_stats_service.py
```diff
@@ -154,13 +154,6 @@ def test_influx_service_not_inited(self):
         svc.thd_postStatsValue("test", "test", "test")
         self.assertLogged("Service.*not initialized")
 
-    def test_storage_backend_base(self):
-        # Just initlize and run thd_postStatsValue
-        svc = StatsStorageBase()
```
Stibbons: not possible, since it is an abstract class. Use a lower level of class, or create a dummy class inherited from StatsStorageBase directly

sa2ajj: actually, a dummy class inherited directly from `StatsStorageBase` sounds like a good idea!  Thanks :)

prasoon2211:
@sa2ajj Can you please have a look again? I have added a dummy subclass and reintroduced the removed test.

master/buildbot/test/unit/test_stats_service.py
```diff
@@ -154,9 +164,9 @@ def test_influx_service_not_inited(self):
...
+    def test_storage_backend_base_failure_on_init(self):
+        svc = DummyStatsStorageBase()
```
sa2ajj: So this is a positive test, not a negative one, right?  You defined a class that defines the required abstract method.  What about one more test where the class does not define the required abstract method?

prasoon2211: Because that will be no different from initializing the parent class itself. Just for testing, I tried doing that and again, pylint raises a post-commit error:

```
(sandbox)[prasoon@jimmypage src]$ git commit -m "Added negative check test"
Validating the following commits:
4be7f823f7f6d01a9641987054947c79cdaa7db4 Added negative check test
** Skipping JavaScript Tests **
** Skipping Python Tests **
-- checking formatting --
-- checking for release notes --
** HEAD~..HEAD does not add release notes **
-- checking import module convention in modified files --
-- running autopep8 --
-- running pep8 --
-- running pyflakes --
-- running pylint --
Warning: option include-ids is deprecated and ignored.
Warning: option symbols is deprecated and ignored.
************* Module buildbot.test.unit.test_stats_service
E:184,47: Abstract class 'DummyStatsStorageBaseNoMethod' with abstract methods instantiated (abstract-class-instantiated)
** pylint failed **
-- not building docs, because it was not changed --

NO GOOD!
** Skipping JavaScript Tests (warning)
** Skipping Python Tests (warning)
** HEAD~..HEAD does not add release notes (warning)
** pylint failed
[todo-pr-2 4be7f82] Added negative check test
 1 file changed, 12 insertions(+)
```

Stibbons: there are abstract method left. I use quite heavily abc and pylint only complains for the base class instanciation. Derived class should not complain if all abstractmethod are overriden in the inherited class

nam4dev: I use it too and fully agree with @Stibbons

prasoon2211: Yes, of course. I agree too. In the last commit, I subclassed the base ABC and overrode the abstract method. That worked just fine. In this new commit, I subclassed it again and *didn't* override the abstract method because that's was @sa2ajj as far as I could tell. Hence the error.

sa2ajj:
Ok, let's proceed with the current PR, I think it's good enough from other people point of view :)

@prasoon2211 thank you for addressing the comments

PR #1790, opened 2015-07-25
fix race condition with appendlog and finishLog

tardyp:
I finally found why we had so much failure with the test_trigger integration test..

log could be finished before the insertion and compression
of the last chunk is finished

This lead to instability of integrations tests.
We can see this more often since we implemented compression


ssavrim:
:+1:

nam4dev:
:+1:

PR #1792, opened 2015-07-27
add a bundle in setup.py

tardyp:
thus you can install full buildbot using:

    pip install 'buildbot[bundle]'

attempt to fix http://trac.buildbot.net/ticket/3311

Having a metapackage would only change the message to

pip install buildbot-bundle

and requires to generate and maintain another package.

@vlovich would that work for you?

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -391,6 +391,10 @@ else:
             'pylint==1.1.0',
             'pyflakes',
         ],
+        'bundle': [
+            'buildbot-www',
+            'buildbot-slave'
+        ]
     }
 
     if os.getenv('NO_INSTALL_REQS'):


master/setup.py
```diff
@@ -391,6 +391,10 @@ def define_plugin_entries(groups):
             'pylint==1.1.0',
             'pyflakes',
         ],
+        'bundle': [
+            'buildbot-www',
+            'buildbot-slave'
```
sa2ajj: do we really want to install `buildbot-slave` as well?

master/setup.py
```diff
@@ -391,6 +391,10 @@ def define_plugin_entries(groups):
             'pylint==1.1.0',
             'pyflakes',
         ],
+        'bundle': [
+            'buildbot-www',
```
sa2ajj: with `buildbot-www` we have a bit of issue: it depends on `buildbot_pkg`

master/setup.py
tardyp: its made for very early installs, in that case probably you want to run a slave on the master.
If we only depend on buildbot-www, another fix for that bug would be to make buildbot-www depend on buildbot, and ask people to install only buildbot-www if they need web, or buildbot if they dont.

master/setup.py
tardyp: it does not really as far as pip is involved. this is only a build dependancy, so it is not in the deps.
I verified buildbot-www.whl depends on nothing.

sa2ajj: side note: I usually do not use wheels :)

vlovich:
Sure.  I'm not concerned about the particular solution.  As long as I can setup my setup.py to use this & the obvious documentation that users of buildbot reach for first mentions pip install 'buildbot[bundle]' that would be fine for me I think.

master/setup.py
sa2ajj: I think this approach would make sense.

tardyp:
no objection since 6 days.. I'm merging

PR #1794, opened 2015-07-29
Fix: Instead of pull everything should it be pull --rev (branch or rev)

mariangemarcano:
Fix: Instead of pull everything should it be pull --rev (branch or rev), hg pull (everything) will create a problem with any branch that has multiple heads (which is valid workflow in Mercurial). Depending on whether you pull a specific revision first, then pull everything, or pull everything first and then get a specific revision, the 'most recently changed head' can be different on the branch, and that can lead to inconsistent results between buildslaves.

Command tested on a katana buildmaster.

### DIFF
--- a/master/buildbot/steps/source/mercurial.py
+++ b/master/buildbot/steps/source/mercurial.py
@@ -148,7 +148,7 @@ class Mercurial(Source):
         @defer.inlineCallbacks
         def _cmd(updatable):
             if updatable:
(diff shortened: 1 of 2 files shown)


sa2ajj:
:+1:

tardyp:
:+1:

nam4dev:
:+1:

PR #1795, opened 2015-07-29
Install buildbot.statistics master package

olof:
The new `buildbot.statistics` package hasn't been added to `setup.py` yet, leading to issues when doing `buildbot create-master` on a system install. This should fix that :)


sa2ajj:
Thanks for the update!

PR #1796, opened 2015-07-30
cosmetic changes to the Trigger build step

sa2ajj:
* the parameters are listed in a different way
* we usually do not show the signature of the class


tardyp:
:+1:

PR #1801, opened 2015-08-04
Improve GitHub status code and update contrib's GitHub webhook

apilloud:
Whole bunch of changes all lumped together here:

1. Updates the structure be closer to the git_buildbot changesource.
2. Adds flag for only reporting the head of a push rather than all commits.
3. Updates fields pulled from GitHub JSON to match the latest APIs
4. Adds support for Pull Requests.
5. Modifies GitHub status reporter to work with GitHub webhook changes by default. 
6. Fixes various bugs.

### DIFF
--- a/master/contrib/github_buildbot.py
+++ b/master/contrib/github_buildbot.py
@@ -101,12 +101,9 @@ class GitHubBuildBot(resource.Resource):
         event_type = request.getHeader("X-GitHub-Event")
         logging.debug("X-GitHub-Event: %r", event_type)
 
-        if event_type == "ping":
-            request.setResponseCode(OK)
-            return json.dumps({"result": "pong"})
+        handler = getattr(self, 'handle_%s' % event_type, None)
 
-        # Reject events not involving code changes
-        if event_type not in ("push", "pull_request"):
+        if handler is None:
             logging.info(
                 "Rejecting request. Received unsupported event %r.",
                 event_type)
@@ -114,7 +111,6 @@ class GitHubBuildBot(resource.Resource):
             return json.dumps({"error": "Bad Request."})
 
         try:
-
             content_type = request.getHeader("Content-Type")
 
             if content_type == "application/json":
@@ -131,11 +127,7 @@ class GitHubBuildBot(resource.Resource):
             logging.debug("Payload: %r", payload)
             repo = payload['repository']['full_name']
             repo_url = payload['repository']['html_url']
-            changes = None
-            if event_type == "push":
-                changes = self.process_changes(payload, repo, repo_url)
-            elif event_type == "pull_request":
-                changes = self.process_changes_pr(payload, repo, repo_url)
+            changes = handler(payload, repo, repo_url)
             self.send_changes(changes, request)
             return server.NOT_DONE_YET
 
@@ -169,7 +161,10 @@ class GitHubBuildBot(resource.Resource):
              'project': repo,
              'branch': branch}
 
-    def process_changes(self, payload, repo, repo_url):
+    def handle_ping(self, *_):
+        return None
+
+    def handle_push(self, payload, repo, repo_url):
         """
         Consumes the JSON as a python object and actually starts the build.
 
@@ -201,7 +196,7 @@ class GitHubBuildBot(resource.Resource):
                     change, branch, repo, repo_url))
         return changes
 
-    def process_changes_pr(self, payload, repo, repo_url):
+    def handle_pull_request(self, payload, repo, repo_url):
         """
         Consumes the JSON as a python object and actually starts the build.
 
--- a/master/contrib/github_buildbot.py
+++ b/master/contrib/github_buildbot.py
@@ -129,14 +129,13 @@ class GitHubBuildBot(resource.Resource):
                 return json.dumps({"error": "Bad Request."})
 
             logging.debug("Payload: %r", payload)
-            user = payload['sender']['login']
             repo = payload['repository']['full_name']
             repo_url = payload['repository']['html_url']
             changes = None
             if event_type == "push":
-                changes = self.process_changes(payload, user, repo, repo_url)
+                changes = self.process_changes(payload, repo, repo_url)
             elif event_type == "pull_request":
-                changes = self.process_changes_pr(payload, user, repo, repo_url)
+                changes = self.process_changes_pr(payload, repo, repo_url)
             self.send_changes(changes, request)
             return server.NOT_DONE_YET
 
@@ -145,7 +144,7 @@ class GitHubBuildBot(resource.Resource):
             request.setResponseCode(INTERNAL_SERVER_ERROR)
             return json.dumps({"error": e.message})
 
-    def process_change(self, change, branch, user, repo, repo_url):
+    def process_change(self, change, branch, repo, repo_url):
         files = change['added'] + change['removed'] + change['modified']
         who = ""
         if 'username' in change['author']:
@@ -165,7 +164,7 @@ class GitHubBuildBot(resource.Resource):
              'project': repo,
              'branch': branch}
 
-    def process_changes(self, payload, user, repo, repo_url):
+    def process_changes(self, payload, repo, repo_url):
         """
         Consumes the JSON as a python object and actually starts the build.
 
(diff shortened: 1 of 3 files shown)


master/contrib/github_buildbot.py
```diff
@@ -104,10 +105,10 @@ def render_POST(self, request):
             request.setResponseCode(OK)
             return json.dumps({"result": "pong"})
 
-        # Reject non-push, non-ping events
-        if event_type != "push":
+        # Reject events not involving code changes
+        if event_type not in ("push", "pull_request"):
```
yetanotherion: an enum for even_type may be welcome
https://docs.python.org/3/library/enum.html

yetanotherion:
minor comment

master/contrib/github_buildbot.py
apilloud: @yetanotherion Enum is new to Python 3.4. Wouldn't that break compatibility with older versions of python (such as 2.7)?

apilloud:
Rebased to latest passing build and fixed PR build breaks.

master/contrib/github_buildbot.py
yetanotherion: https://pypi.python.org/pypi/enum ?

2015-08-04 22:57 GMT+02:00 Andrew Pilloud <notifications@github.com>:

> In master/contrib/github_buildbot.py
> <https://github.com/buildbot/buildbot/pull/1801#discussion_r36240773>:
>
> > @@ -104,10 +105,10 @@ def render_POST(self, request):
> >              request.setResponseCode(OK)
> >              return json.dumps({"result": "pong"})
> >
> > -        # Reject non-push, non-ping events
> > -        if event_type != "push":
> > +        # Reject events not involving code changes
> > +        if event_type not in ("push", "pull_request"):
>
> @yetanotherion <https://github.com/yetanotherion> Enum is new to Python
> 3.4. Wouldn't that break compatibility with older versions of python (such
> as 2.7)?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/buildbot/buildbot/pull/1801/files#r36240773>.
>



-- 
Ion

sa2ajj:
I was a bit confused by the PR title.

Actually, it changes GitHub status code and updates contrib's GitHub webhook.

master/contrib/github_buildbot.py
sa2ajj: This kind of event handling does not really allow to easily extend the list of events to handle.  Would you consider implementing something like [this](https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/hooks/github.py#L51)?

master/contrib/github_buildbot.py
```diff
@@ -143,7 +144,27 @@ def render_POST(self, request):
             request.setResponseCode(INTERNAL_SERVER_ERROR)
             return json.dumps({"error": e.message})
 
-    def process_change(self, payload, user, repo, repo_url, project, request):
+    def process_change(self, change, branch, repo, repo_url):
+        files = change['added'] + change['removed'] + change['modified']
+        who = ""
+        if 'username' in change['author']:
```
sa2ajj: Side note: we seem to have a number of places where similar code is present, it feels to be worth to be extracted to function...

master/contrib/github_buildbot.py
```diff
@@ -143,7 +144,27 @@ def render_POST(self, request):
...
+        else:
+            who = change['author']['name']
+        if 'email' in change['author']:
+            who = "%s <%s>" % (who, change['author']['email'])
+
+        return \
+            {'revision': change['id'],
+             'revlink': change['url'],
+             'who': who,
+             'comments': change['message'][:1024],
```
sa2ajj: I do not think just trimming message after 1K chars is a good idea, I'd prefer to have an indicator that it was trimmed.  Something like `super useful description ... (trimmed as the length of commit messages exceeds 1024 characters)` (and yes, this will lead to a few characters from the original commit message to be stored).

master/contrib/github_buildbot.py
```diff
@@ -153,28 +174,68 @@ def process_change(self, payload, user, repo, repo_url, project, request):
                 Hook.
         """
         changes = None
-        branch = payload['ref'].split('/')[-1]
+        refname = payload['ref']
+
+        m = re.match(r"^refs\/(heads|tags)\/(.+)$", refname)
+        if not m:
+            logging.info("Ignoring refname `%s': Not a branch or tag", refname)
```
sa2ajj: `Not a branch or a tag`

master/contrib/github_buildbot.py
```diff
@@ -153,28 +174,68 @@ def process_change(self, payload, user, repo, repo_url, project, request):
...
+            return changes
+
+        branch = m.group(2)
```
sa2ajj: this is misleading in the light of the above regular expression: it's either a branch or a tag.

master/contrib/github_buildbot.py
```diff
@@ -265,6 +326,10 @@ def setup_options():
                            "perspective broker.",
                       default="change:changepw", dest="auth")
 
+    head_commit_help = ("If set, only trigger builds for commits at head")
```
sa2ajj: Is there any reason for introducing `head_commit_help`?

master/contrib/github_buildbot.py
apilloud: How about I call it "refname"?

master/contrib/github_buildbot.py
apilloud: I think this might be a more general bug, as the sqlite database doesn't enforce column length limits. Changed to include a message.

master/contrib/github_buildbot.py
apilloud: I like that method, it's a great improvement. Thanks for the suggestion!

apilloud:
Made changes in response to feedback.

apilloud:
Found a bug in the fields being used to determine the head commit on a push, fixed.

master/contrib/github_buildbot.py
sa2ajj: sounds good to me: it better reflects the nature of the value.

sa2ajj:
We have this nasty problem with certain recent versions of Twisted.  Could you please rebase against master so the tests would show the actual situation?

master/contrib/github_buildbot.py
```diff
@@ -100,20 +101,16 @@ def render_POST(self, request):
...
-            request.setResponseCode(OK)
-            return json.dumps({"result": "pong"})
+        handler = getattr(self, 'handle_%s' % event_type, None)
 
-        # Reject non-push, non-ping events
-        if event_type != "push":
+        if handler is None:
             logging.info(
-                "Rejecting request.  Expected a push event but received %r instead.",
+                "Rejecting request. Received unexpected event %r.",
```
sa2ajj: Rather 'unsupported'

master/contrib/github_buildbot.py
```diff
@@ -153,28 +174,69 @@ def process_change(self, payload, user, repo, repo_url, project, request):
...
+                Python Object that represents the JSON sent by GitHub Service
+                Hook.
+        """
+        changes = None
+
+        branch = "refs/pull/{}/head".format(payload['number'])
+
+        if payload['action'] not in ("opened", "synchronize"):
+            logging.info("PR %r %r, ignoring",
+                         payload['number'], payload['action'])
```
sa2ajj: I think an explicit return would be good here.

sa2ajj:
Looks good to me otherwise.

apilloud:
Made changes and rebased.

apilloud:
Rebased again.

apilloud:
Rebased again, cleaned up style issues found by pylint, added tests for github status changes.

sa2ajj:
Looks like we have failed tests:
```
[ERROR]
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_status_github.py", line 497, in test_repoOverDefined
    self.assertRaisesConfigError(
exceptions.AttributeError: 'TestGitHubStatus' object has no attribute 'assertRaisesConfigError'

buildbot.test.unit.test_status_github.TestGitHubStatus.test_repoOverDefined
[ERROR]
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_status_github.py", line 508, in test_repoUnderDefined
    self.assertRaisesConfigError(
exceptions.AttributeError: 'TestGitHubStatus' object has no attribute 'assertRaisesConfigError'

buildbot.test.unit.test_status_github.TestGitHubStatus.test_repoUnderDefined
```

sa2ajj:
It looks like you need to add `ConfigErrorsMixin` as, for example, in [here](https://github.com/buildbot/buildbot/blob/917f34b53a5a8706aa4003f065883d93e7b1bec0/master/buildbot/test/unit/test_config.py#L131)

apilloud:
The ./common/validate.sh script doesn't appear to run the tests I added. Fixed the test failures, rebased (due to odd failures), and fixed a typo in one of the config error messages.

sa2ajj:
Thank you very much!

PR #1802, opened 2015-08-05
[WIP] new style classes everywhere in master

sa2ajj:


sa2ajj:
@jvlomax is it something that you have in works?

jvlomax:
yes, I have a patch [here](https://github.com/jvlomax/buildbot/tree/new_classes2). I have not submitted it yet do to the tests failing when one particular class is made to be new-style. I asked for some help on it a while back, but no one got back to me. It was maybe discussed at a [weekly meeting](https://supybot.buildbot.net/meetings/buildbot/2015/buildbot.2015-07-14-16.00.html). @djmitche might know more about what is causing the issue

djmitche:
Twisted helpfully upgraded all of our services to new-style classes sometime after 15.0.0, and yeah -- the problems are numerous!

http://trac.buildbot.net/ticket/3329

For 0.9.0b2, I was considering just requiring <15.0.0, but obviously that precludes changing all of our classes to new-style.  Another option is to revert the `master` property and go back to setting self.master everywhere, like we used to: arguably the `master` property is only partially implemented, although that partial implementation was OK with old-style classes since properties weren't enforced.

sa2ajj:
I wonder what would be the best to _properly_ fix this assignment to master property...

tardyp:
what happens if we rebase this over #1808?

sa2ajj:
Let's see :smile:

djmitche:
Yikes, not pretty.

We can drop the unpickling tests, as well as the import of changes in DB version 001, as we are (right?) committed to not supporting migration of 0.8.x data to nine.  That should fix a lot of the issues in the travis errors.

sa2ajj:
This seem to require much deeper investigation.

PR #1804, opened 2015-08-05
Added capture classes to capture on all builders, regexes.

prasoon2211:
This is the third PR in the series of changes. This adds new capture classes that can be used to capture properties/build times/data on all builders. Also, this class adds supports for capturing properties using regular expressions.

I've added the tests. Docs will go in PR-4.

@sa2ajj @Stibbons @benallard

### DIFF
--- a/master/buildbot/statistics/__init__.py
+++ b/master/buildbot/statistics/__init__.py
@@ -14,14 +14,21 @@
 # Copyright Buildbot Team Members
 
 from buildbot.statistics.capture import CaptureBuildDuration
+from buildbot.statistics.capture import CaptureBuildDurationAllBuilders
 from buildbot.statistics.capture import CaptureBuildEndTime
+from buildbot.statistics.capture import CaptureBuildEndTimeAllBuilders
 from buildbot.statistics.capture import CaptureBuildStartTime
+from buildbot.statistics.capture import CaptureBuildStartTimeAllBuilders
 from buildbot.statistics.capture import CaptureData
+from buildbot.statistics.capture import CaptureDataAllBuilders
 from buildbot.statistics.capture import CaptureProperty
+from buildbot.statistics.capture import CapturePropertyAllBuilders
 from buildbot.statistics.stats_service import StatsService
 from buildbot.statistics.storage_backends.influxdb_client import InfluxStorageService
 
 __all__ = [
-    'StatsService', 'InfluxStorageService', 'CaptureProperty', 'CaptureBuildDuration',
-    'CaptureBuildStartTime', 'CaptureBuildEndTime', 'CaptureData'
+    'StatsService', 'InfluxStorageService', 'CaptureProperty', 'CapturePropertyAllBuilders',
+    'CaptureBuildDuration', 'CaptureBuildDurationAllBuilders', 'CaptureBuildStartTime',
+    'CaptureBuildStartTimeAllBuilders', 'CaptureBuildEndTime', 'CaptureBuildEndTimeAllBuilders',
+    'CaptureData', 'CaptureDataAllBuilders'
 ]
--- a/master/buildbot/statistics/__init__.py
+++ b/master/buildbot/statistics/__init__.py
@@ -27,8 +27,16 @@ from buildbot.statistics.stats_service import StatsService
 from buildbot.statistics.storage_backends.influxdb_client import InfluxStorageService
 
 __all__ = [
-    'StatsService', 'InfluxStorageService', 'CaptureProperty', 'CapturePropertyAllBuilders',
-    'CaptureBuildDuration', 'CaptureBuildDurationAllBuilders', 'CaptureBuildStartTime',
-    'CaptureBuildStartTimeAllBuilders', 'CaptureBuildEndTime', 'CaptureBuildEndTimeAllBuilders',
-    'CaptureData', 'CaptureDataAllBuilders'
+    'CaptureBuildDuration',
+    'CaptureBuildDurationAllBuilders',
+    'CaptureBuildEndTime',
+    'CaptureBuildEndTimeAllBuilders',
+    'CaptureBuildStartTime',
+    'CaptureBuildStartTimeAllBuilders',
+    'CaptureData',
+    'CaptureDataAllBuilders',
+    'CaptureProperty',
+    'CapturePropertyAllBuilders',
+    'InfluxStorageService',
+    'StatsService'
 ]
--- a/master/buildbot/statistics/capture.py
+++ b/master/buildbot/statistics/capture.py
@@ -344,6 +344,10 @@ class CaptureDataBase(Capture):
             context = self._defaultContext(build_data, builder_info['name'])
             yield self._store(post_data, series_name, context)
 
+    @abc.abstractmethod
+    def _builder_name_matches(self, builder_info):
+        pass
+
 
 class CaptureData(CaptureDataBase):
 
@@ -365,9 +369,6 @@ class CaptureDataAllBuilders(CaptureDataBase):
     """
     Capture methods for arbitraty data that may not be stored in the Buildbot database.
     """
-
-    def __init__(self, data_name, callback=None):
-        CaptureDataBase.__init__(self, data_name, callback)
 
     def _builder_name_matches(self, builder_info):
         return True
--- a/master/buildbot/statistics/capture.py
+++ b/master/buildbot/statistics/capture.py
@@ -14,6 +14,7 @@
 # Copyright Buildbot Team Members
 
 import abc
+import re
 
 from twisted.internet import defer
 from twisted.internet import threads
@@ -36,9 +37,9 @@ class Capture(object):
         self.parent_svcs = []
         self.master = None
 
-    def _defaultContext(self, msg):
+    def _defaultContext(self, msg, builder_name):
         return {
-            "builder_name": self._builder_name,
+            "builder_name": builder_name,
             "build_number": str(msg['number'])
         }
 
@@ -53,16 +54,15 @@ class Capture(object):
                                         context)
 
 
-class CaptureProperty(Capture):
+class CapturePropertyBase(Capture):
 
     """
-    Convenience wrapper for getting statistics for filtering.
-    Filters out build properties specifies in the config file.
+    A base class for CaptureProperty* classes.
     """
 
-    def __init__(self, builder_name, property_name, callback=None):
-        self._builder_name = builder_name
+    def __init__(self, property_name, callback=None, regex=False):
         self._property_name = property_name
+        self._regex = regex
         routingKey = ("builders", None, "builds", None, "finished")
 
         def default_callback(props, property_name):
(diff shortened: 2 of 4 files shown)


master/buildbot/statistics/capture.py
```diff
@@ -80,25 +80,74 @@ def consume(self, routingKey, msg):
...
+
+
+class CapturePropertyAllBuilders(CapturePropertyBase):
+
+    """
+    Capture class for filtering out build properties for all builds.
+    """
+
+    def __init__(self, property_name, callback=None, regex=False):
+        CapturePropertyBase.__init__(self, property_name, callback, regex)
```
Stibbons: useless, let the default __init__

Stibbons:
Minor remark. Overall this is quite good for me. Top code quality :)

master/buildbot/statistics/capture.py
```diff
@@ -80,25 +80,74 @@ def consume(self, routingKey, msg):
...
-            series_name = '%s-%s' % (self._builder_name, self._property_name)
-            post_data = {
-                "name": self._property_name,
-                "value": ret_val
-            }
-            yield self._store(post_data, series_name, context)
+            filtered_prop_names = []
+            if self._regex:
+                prop_names = properties.keys()
+                for pn in prop_names:
```
jvlomax: If you want prop_names as a list of keys, please use ```list(properties)``` or (less preferably) ```list(properties.keys())```.
If you only want to use it for iteration, please use ```for pn in properties:```. 

This is to ensure python3 compatibility

Stibbons: for name in properties.keys()

is enough for iterating through all the name. No need to convert to list (in python 2.7 this is a list, in 3 I guess this is a generator)

master/buildbot/statistics/capture.py
```diff
@@ -80,25 +80,74 @@ def consume(self, routingKey, msg):
...
+                    if re.match(self._property_name, pn):
+                        filtered_prop_names.append(pn)
```
Stibbons: you can rewrite it:

    filtered_prop_names = [pn for pn in properties if re.match(self._property_name, pn]

master/buildbot/statistics/capture.py
```diff
@@ -80,25 +80,74 @@ def consume(self, routingKey, msg):
...
+
+    def __init__(self, builder_name, property_name, callback=None, regex=False):
+        self._builder_name = builder_name
+
+        CapturePropertyBase.__init__(self, property_name, callback, regex)
+
+    def _builder_name_matches(self, builder_info):
+        if self._builder_name == builder_info['name']:
+            return True
+        return False
```
Stibbons: return self._builder_name == builder_info['name']

directly

master/buildbot/statistics/capture.py
```diff
@@ -178,6 +250,25 @@ def default_callback(end_time):
     def _retValParams(self, msg):
         return [msg['complete_at']]
 
+    def _builder_name_matches(self, builder_info):
+        if self._builder_name == builder_info['name']:
+            return True
+        return False
```
Stibbons: return self._builder_name == builder_info['name']

master/buildbot/statistics/capture.py
```diff
@@ -206,16 +297,34 @@ def default_callback(start_time, end_time):
     def _retValParams(self, msg):
         return [msg['started_at'], msg['complete_at']]
 
+    def _builder_name_matches(self, builder_info):
+        if self._builder_name == builder_info['name']:
+            return True
+        return False
```
Stibbons: return self._builder_name == builder_info['name']

master/buildbot/statistics/capture.py
```diff
@@ -234,15 +345,46 @@ def consume(self, routingKey, msg):
...
+    def __init__(self, data_name, builder_name, callback=None):
+        self._builder_name = builder_name
+
+        CaptureDataBase.__init__(self, data_name, callback)
+
+    def _builder_name_matches(self, builder_info):
+        if self._builder_name == builder_info['name']:
+            return True
+        return False
+
```
Stibbons: same

master/buildbot/test/unit/test_stats_service.py
```diff
@@ -389,6 +439,27 @@ def test_capture_data(self):
         )], self.fake_storage_service.stored_data)
 
     @defer.inlineCallbacks
+    def test_capture_data_all_builders(self):
+        self.setup_fake_storage([capture.CaptureDataAllBuilders('test')])
```
Stibbons: use same convension => camelCase for methodNames

setup_fake_storage => setupFakeStorage

master/buildbot/statistics/__init__.py
```diff
@@ -14,14 +14,21 @@
...
+from buildbot.statistics.capture import CaptureBuildStartTimeAllBuilders
 from buildbot.statistics.capture import CaptureData
+from buildbot.statistics.capture import CaptureDataAllBuilders
 from buildbot.statistics.capture import CaptureProperty
+from buildbot.statistics.capture import CapturePropertyAllBuilders
 from buildbot.statistics.stats_service import StatsService
 from buildbot.statistics.storage_backends.influxdb_client import InfluxStorageService
 
 __all__ = [
-    'StatsService', 'InfluxStorageService', 'CaptureProperty', 'CaptureBuildDuration',
```
Stibbons: i would prefer 1 item per line so you can easily insert new one, and sort them

sa2ajj: /me agrees

master/buildbot/statistics/capture.py
```diff
@@ -53,16 +54,15 @@ def _store(self, post_data, series_name, context):
...
-    Convenience wrapper for getting statistics for filtering.
-    Filters out build properties specifies in the config file.
+    A base class for CaptureProperty* classes.
     """
 
-    def __init__(self, builder_name, property_name, callback=None):
-        self._builder_name = builder_name
+    def __init__(self, property_name, callback=None, regex=False):
         self._property_name = property_name
+        self._regex = regex
```
sa2ajj: should some sanity check be performed here?  like it's either a string or an instance of compiled regex?

master/buildbot/statistics/capture.py
sa2ajj: and this list comprehension is much easier to read.

master/buildbot/statistics/capture.py
```diff
@@ -80,25 +80,74 @@ def consume(self, routingKey, msg):
...
+    """
+    Capture class for filtering out build properties for all builds.
+    """
+
+    def __init__(self, property_name, callback=None, regex=False):
+        CapturePropertyBase.__init__(self, property_name, callback, regex)
+
+    def _builder_name_matches(self, builder_info):
+        """
+        Since we need to match all builders, we simply return True here.
```
sa2ajj: This is better as a comment, not as a doc string.

master/buildbot/statistics/capture.py
Stibbons: list comprehension is great

dict comprehension is even greater!

master/buildbot/statistics/capture.py
```diff
@@ -118,32 +167,36 @@ def consume(self, routingKey, msg):
...
         """
         builder_info = yield self.master.data.get(("builders", msg['builderid']))
-        if self._builder_name == builder_info['name']:
+        if self._builder_name_matches(builder_info):
             try:
                 ret_val = self._callback(*self._retValParams(msg))
             except Exception as e:
                 # catching generic exceptions is okay here since we propagate it
-                config.error(self._err_msg(msg) + " Exception raised: " + type(e).__name__ +
+                config.error(self._err_msg(msg, builder_info['name']) + " Exception raised: " + type(e).__name__ +
```
sa2ajj: Open question: is use of `config.error` warranted here?

And it's better to use '%s ... %s' % (...)` for formatting the error message.  This way it's easier to understand what is where.

master/buildbot/statistics/capture.py
```diff
@@ -80,25 +80,74 @@ def consume(self, routingKey, msg):
...
+                    if re.match(self._property_name, pn):
+                        filtered_prop_names.append(pn)
+            else:
+                filtered_prop_names.append(self._property_name)
+
+            for pn in filtered_prop_names:
+                try:
+                    ret_val = self._callback(properties, pn)
+                except KeyError:
+                    config.error("CaptureProperty failed."
```
sa2ajj: Open question: is use of `config.error` warranted here?

master/buildbot/statistics/capture.py
```diff
@@ -118,32 +167,36 @@ def consume(self, routingKey, msg):
...
-                config.error(self._err_msg(msg) + " Exception raised: " + type(e).__name__ +
+                config.error(self._err_msg(msg, builder_info['name']) + " Exception raised: " + type(e).__name__ +
                              " with message: " + str(e))
-            context = self._defaultContext(msg)
+            context = self._defaultContext(msg, builder_info['name'])
             post_data = {
                 self._time_type: ret_val
             }
-            series_name = self._builder_name + "-build-times"
+            series_name = builder_info['name'] + "-build-times"
```
sa2ajj: I should've commented on this earlier: '%s-build-times' % (builder_info['name'],)` would be better.

master/buildbot/statistics/capture.py
```diff
@@ -161,11 +214,30 @@ def default_callback(start_time):
     def _retValParams(self, msg):
         return [msg['started_at']]
 
+    def _builder_name_matches(self, builder_info):
+        if self._builder_name == builder_info['name']:
```
sa2ajj: `return self._builder_name == builder_info['name']`

sa2ajj:
And I think it's a good idea to rebase against master to include @djmitche's fix for `self.master = master`

master/buildbot/statistics/capture.py
Stibbons: look like error is a custom defined error handler.

I would recommend to use logging so you can do something like:
```
errLog = logging.getLogger(...)

errLog.error("blablabla %s, %d", val1, val2)
```

And it will do a lazy evaluation of the string.

But in this case I recommend to use '{0} {1},...' formatting, which is better than concatenating strings with '+'

('{0}' for support of python 2.6)

Stibbons:
indeed look like there is an issue with ```self.master```

master/buildbot/statistics/capture.py
prasoon2211: Ok.

master/buildbot/statistics/capture.py
prasoon2211: Okay. I'm changing it to `for pn in properties`.

master/buildbot/statistics/capture.py
prasoon2211: Ok.

master/buildbot/statistics/capture.py
prasoon2211: `regex` is a bool (default=False) based on which we invoke the regex filtering. I don't understand what kind of check we should do here.

master/buildbot/statistics/capture.py
prasoon2211: Ok.

master/buildbot/statistics/capture.py
sa2ajj: We are not using `logger` in Buildbot: we are using Twisted's `log`

master/buildbot/statistics/capture.py
sa2ajj: Sorry, it was not obvious for me.

master/buildbot/statistics/capture.py
prasoon2211: Well, it is a config error because `self._callback` can be set by the user in the config file. If done improperly, this will cause an exception. Also, just to be clear about exactly what the error is, we are showing actual exception and the error message.

sa2ajj: The thing is that this call is way too late: usually `config.error` is used in constructors.

master/buildbot/statistics/capture.py
Stibbons: So lazy string Eval should work :)

benallard:
To be completely honest, I'm a bit worried about the multiplication of classes, and the apparent duplication of code in this PR ... Unfortunately, I haven't taken the time to look deeper into the content to see if there is some trivial fix to that.

sa2ajj:
@benallard even if you did not look deep enough, could you please suggest what are your first thoughs/reactions?

master/buildbot/statistics/capture.py
prasoon2211: Are there any other exceptions defined in Buildbot that would work here?

master/buildbot/statistics/capture.py
prasoon2211: I don't get this - we need to raise some kind of exception so execution halts and not just record the error. I don't get why changing the logger here would be useful at all.

I took a look at some other instances of `config.error` usage and it is as @sa2ajj says - `config.error` is used mostly inside constructors and sometimes outside but it isn't used for runtime errors. So I suppose we need a better exception here.

prasoon2211:
Please have a look at this. I haven't dealt with the `config.error` issue but all else has been dealt with.

prasoon2211:
Just added a minor change that I had forgot earlier!

master/buildbot/statistics/capture.py
Stibbons: agree to stick with current usage. If buildbot uses this error class, let's use it.

Stibbons:
mostly look good

benallard:
The ``config.error`` is just meant for initialisation-time error (i.e. when we can still catch them and fix them: when someone is still looking at the terminal). Once buildbot is running, there is no way to stop the execution and make the user aware of the issue. best we do is report the exception in the log, and hope someone will look at it sometime. Competent system administrators do that from time to time ...

So in your case, if you have no way to detect the error early (and hence use the ``config.error``), you should just log it.

prasoon2211:
Okay. So, in the end, what should I do about the config.error issue? Should I raise the error or just log it? Ben says that I should just report it but feel that raising an error is important here - if there is an undetected error, a lot of build data may be lost.

@Stibbons What's your take on this?

sa2ajj:
Please use `config.error` only in the constructor.  Otherwise raise some other exception.

If you do not feel any available exceptions meet your use, add one in `buildbot.errors`.

Stibbons:
ok for me

prasoon2211:
Okay then. I've added a new exception.

sa2ajj:
:+1:

prasoon2211:
Can we please get this merged? I have two more PRs to make after this.

sa2ajj:
we'll take from here.

Stibbons:
great!

PR #1809, opened 2015-08-10
Add a name to the link to the logs.

benallard:
Full disclosure: this is only a wild guess on what should be added there, not tested.


tardyp:
👍 how could that ever work? ™

The step page is but very useful...

PR #1810, opened 2015-08-12
Displaying build step and logs (a counterpart to debug #1760)

shanzi:
Here is my latest work. It is a counterpart to @tothandras 's latest `buildbot-data` (see also #1760 ). So it needs latest dataService to get working. This pull request is not only for showing my progress by now, but also for helping debug new dataService.

With a dataService tweaked by me from the old one, we can achieve effects like:

screen shot 2015-08-12 at 9 33 33 pm

and

screen shot 2015-08-12 at 9 33 51 pm

But currently, it exposed some bugs in new dataService (so it is in fact not 100% working with new dataService):
1. `builder.tags`'s type is a piece of JSON string (which should parsed into a list)
2. get `buildstep.logs` will return an empty list, though it seems the correctly HTTP request is performed and returned JSON is valid too.
3. new dataService causes a lot of `websocketService` not found errors.

Sorry for still have not a workable live demo now. It needs both master and slave and I have not come up with a good way to run slave on dokku platform yet. I will try to overcome this once main body of new UI is finished.


tardyp:
here is a buildbot.tac which runs a master and a slave in the same process.

https://gist.github.com/tardyp/e2af6270cc9563cca197

sa2ajj:
by looking at the screenshots, I have a suggestion of not expanding stdout by defaut, unless it's the only log (I'm not sure how it's handled).

/me is not very good at UI stuff review, sorry :(

shanzi:
@sa2ajj 
The stdout will be showed once you click on a step to expand it. I find the real problem is if the log is very long that may slow down the page and not convenient to view. I'm considering of ways to address this problem such as only show some tail lines by default. Show nothing before user's action is also an option. Thanks for your suggestion and I will look into ways to come up with a better interaction design.

www/md_base/src/app/builds/builder/builder.controller.coffee
```diff
@@ -3,6 +3,38 @@ class Builder extends Controller
...
     autoSelect: true
 
+    constructor: ($scope, @$state, @$mdDialog, @dataService) ->
+        @info = {}
+        @builds = []
+        @lastBuild = {}
+        @forceschedulers = []
+        @buildTabs = []
+        
+        opened = dataService.open()
```
tothandras: you can write `dataService.open($scope)` instead of these two lines

www/md_base/src/app/builds/builder/builder.route.coffee
```diff
@@ -19,15 +19,21 @@ class State extends Config
             name: 'builds.builder.buildstab'
             url: '/buildstab'
             templateUrl: "views/#{name}.buildstab.html"
-            controller: ($scope) ->
-                $scope.builder.selectTab('buildstab')
+            controller: [
```
tothandras: use the `### @ngInject ###` annotation instead

www/md_base/src/app/common/directives/buildlog/buildlog.directive.coffee
```diff
@@ -0,0 +1,58 @@
...
+        @contentsLength = 0
+        @codeTable.html('')
+
+        # Load and process contents
+        @log = @raw_log
+        @contents = @log.contents or @log.loadContents().getArray()
+        @updateContents()
+
+    addLine: (lineno, line) ->
+        @codeTable.append """
```
tothandras: `$sanitize` just in case?

tardyp:
please try https://github.com/buildbot/buildbot/pull/1815
for running the slave inside the master process.

www/md_base/src/app/builds/builder/builder.route.coffee
shanzi: It doesn't work, coffeescript doesn't compile

master/buildbot/scripts/dataspec.py
```diff
@@ -28,7 +28,8 @@
 @defer.inlineCallbacks
 def dataspec(config):
     master = yield fakemaster.make_master()
-    data = connector.DataConnector(master)
+    data = connector.DataConnector()
+    data.setServiceParent(master)
```
shanzi: @tardyp I added a line to fix the error occurs when performing `gulp dev` here as `DataConnector` no longer accept a master as init params.

www/md_base/src/app/common/directives/buildlog/buildlog.directive.coffee
shanzi: :+1: done

shanzi:
@tardyp what if we merge this and move on? Although it still doesn't work with new `dataService`, I think we can move on to the directive of building force build dialog form. That seems to be some work.

After that form is finished the whole UI will be close to usable. So I plan to finish that before deadline of GSoC. There still are many add aspects need handling left. I will continue perfecting it beyond summer.

tothandras:
:+1:

www/md_base/src/app/builds/builder/builder.route.coffee
tothandras: Did you try?:
```coffeescript
### @ngInject ###
controller: ($scope) -> $scope.builder.selectTab('buildstab')
```

sa2ajj:
Let's move on :)

PR #1815, opened 2015-08-14
LocalBuildSlave

tardyp:
as per request from @shanzi, I tried to run a slave within the master process with the null protocol, and figured out it is not very practical.

This change makes it very simple to run one or several slaves within a master.


shanzi:
:+1:

sa2ajj:
LGTM.

A line in release notes would be nice though :)

PR #1816, opened 2015-08-14
use cctrial to run selected tests on post-commit hook

tardyp:
I updated cctrial https://pypi.python.org/pypi/cctrial/1.2.0 to have a hook mode, which will detect and run tests that are appropriate for the modified files.

### DIFF
--- a/common/validate.sh
+++ b/common/validate.sh
@@ -142,8 +142,10 @@ fi
 if ! $quick; then
     status "running Python tests"
     run_tests || not_ok "Python tests failed"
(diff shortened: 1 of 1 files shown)


sa2ajj:
I think this needs to be reflected in some developer documentation: it'd be not so nice to have failed because the tool is not available.

tardyp:
The presence of the tool is tested.

Le lun. 17 août 2015 17:56, Mikhail Sobolev <notifications@github.com> a
écrit :

sa2ajj:
oh, sorry, I missed that.

PR #1812, opened 2015-08-13
Bug3254 gitlab try gateway, Bug3333 'buildbot try' should output URL of failed builds

dankegel:
For your reviewing pleasure.
Lightly tested.

PR #1820, opened 2015-08-15
Bug3254 gitlab try gateway bis

dankegel:
Changes since https://github.com/buildbot/buildbot/pull/1812 :
- added relnotes
- autopep8'd
- hyperlink to builder on success, too
- better error handling and messages
- hoisted builder selection code out into function to make it easier to customize if needed
- more complete doc

I tried to write a test for my fix for bug 3333, but bug 2762 got in the way;
RUN_FLAKY_TESTS=1 PYTHONPATH=. trial buildbot.test.integration.test_try_client
fails every time for me.

### DIFF
--- a/master/buildbot/clients/tryclient.py
+++ b/master/buildbot/clients/tryclient.py
@@ -685,7 +685,7 @@ class Try(pb.Referenceable):
         self.outstanding = []
 
         # self.results holds the list of build results. It holds a tuple of
-        # (result, text)
+        # (result, text, buildnumber)
         self.results = {}
 
         # self.currentStep holds the name of the Step that each build is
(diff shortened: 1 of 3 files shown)


tardyp:
I am not a fan of creating a webserver that checks-out a git tree in order to generate a patch for buildbot try.

- The branch is available has a git branch, why not let the buildslave fetch it and merge it in master via the existing change and changehook mechanism (like git gerrit does)?
- maintaining a synchronous webserver in our tree makes me nervous, this could have been implemented with the hook service.
- the try interface is deprecated, and is destinated to be replaced by a REST api call.

So even if it might work perfectly for you, I am not for including it in the upstream branch (and thus makes it the recommended way of doing this).

I am not against the `try --wait` part of this PR, though, which would deserve a test.

dankegel:
Thanks for the review.
I'll see if I can get gitlab to cough up the patch instead of computing it in the hook.
And I'll try to port it to twistd, like the standalone github_buildbot.py; once that's working I'd feel more comfortable trying to add it to the existing gitlab hook.

xeor:
@dankegel did you end up creating an issue with gitlab? Can't find it. If you didn't, no worries :)

dankegel:
Pierre, thanks for your feedback.  I'm trying to do it properly inside buildbot/status/web/hooks/gitlab.py now.   It's going to be a lot simpler.

PR #1822, opened 2015-08-16
Fix #3248: BuildStep: fix up interrupt logic

jaredgrubb:
I'm still testing this patch on my own bot, but wanted to put up for comments. It seems that all the logic in the derived Interrupt function belongs in the base:
 - self.cmd is checked and set in the base
 - 'setWaitingForLocks' is set by the base

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -389,11 +389,24 @@ class BuildStep(object, properties.PropertiesMixin):
         raise NotImplementedError("your subclass must implement run()")
 
     def interrupt(self, reason):
+        # TODO: consider adding an INTERRUPTED or STOPPED status to use
+        # instead of FAILURE, might make the text a bit more clear.
+        # 'reason' can be a Failure, or text
         self.stopped = True
         if self._acquiringLock:
             lock, access, d = self._acquiringLock
             lock.stopWaitingUntilAvailable(self, access, d)
             d.callback(None)
+
+        if self._step_status.isWaitingForLocks():
+            self.addCompleteLog(
+                'interrupt while waiting for locks', str(reason))
+        else:
+            self.addCompleteLog('interrupt', str(reason))
+
+        if self.cmd:
+            d = self.cmd.interrupt(reason)
+            d.addErrback(log.err, 'while interrupting command')
 
     def releaseLocks(self):
         log.msg("releaseLocks(%s): %s" % (self, self.locks))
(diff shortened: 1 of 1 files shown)


master/buildbot/process/buildstep.py
```diff
@@ -389,12 +389,25 @@ def start(self):
         raise NotImplementedError("your subclass must implement run()")
 
     def interrupt(self, reason):
+        # TODO: consider adding an INTERRUPTED or STOPPED status to use
```
tardyp: there is already a CANCELLED status, that could work.

master/buildbot/process/buildstep.py
```diff
@@ -389,12 +389,25 @@ def start(self):
...
             lock.stopWaitingUntilAvailable(self, access, d)
             d.callback(None)
 
+        if self._step_status.isWaitingForLocks():
+            self.addCompleteLog(
+                'interrupt while waiting for locks', str(reason))
+        else:
+            self.addCompleteLog('interrupt', str(reason))
+
+        if self.cmd:
```
tardyp: do we have self.cmd in BuildStep? I dont think we do.

tardyp:
I am all for making BuildStep base more command aware. I think there are a lot of steps that are reimplementing cmd awareness in a different way.

master/buildbot/process/buildstep.py
jaredgrubb: Yes, "cmd = None" is in the base class, and there is a BuildStep.runCommand.

master/buildbot/process/buildstep.py
sa2ajj: that might, though there's no CANCELLED in console (this is for eight, btw)

djmitche:
The base BuildStep class is aware of RemoteCommand and how to run it; it is *not* aware of RemoteShellCommand and all of the machinery that goes around supporting it.

master/buildbot/process/buildstep.py
```diff
@@ -683,21 +696,6 @@ def setupLogfiles(self, cmd, logfiles):
                 # and tell the RemoteCommand to feed it
                 cmd.useLog(newlog, True)
 
-    def interrupt(self, reason):
```
djmitche: This `interrupt` is in `LoggingBuildStep`, which is in fact deprecated.  So I think it makes a lot of sense to move it to the base class.  In short, I like this PR :)

jaredgrubb:
Removed the "Do Not Merge" since I think this patch is good.. I've been using it a while.

djmitche:
I already +1'd this.  @tardyp did you have any objections or should we get this merged?

sa2ajj:
Let's proceed with this.

PR #1823, opened 2015-08-20
dataService bug fixes and documentation

tothandras:
@shanzi has found a few bugs, this PR contains fixes for those.
I also added a simple documentation for dataService.

I have started to migrate the base module to the dataService, I will include that in my last PR before GSoC ends.

In this semester I can actively contribute to Buildbot, I feel like there is still a lot to do to in terms of optimisation. Sharing the WebSocket connection and using IndexedDB as a single data store introduced a lot of unexpected problems, it took more time to solve those than expected.

### DIFF
--- a/master/docs/developer/www.rst
+++ b/master/docs/developer/www.rst
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
     Call the control data api.
     This builds up a POST with jsonapi encoded parameters
 
+DataService
+.............
+
+DataService is the future replacement of BuildbotService for accessing the Buildbot data API.
+It has a modern interface for accessing data. It uses IndexedDB for storing cached data as a single data store,
+and LocalStorage for broadcasting events between browser tabs. DataService works in a master/slave architecture.
+The master browser tab is responsible for keeping the requested data up to date in the IndexedDB and notify slaves when a data is ready to be used or it is updated.
+It handles both the Rest API calls and the WebSocket subscriptions globally.
+
+It uses the following libraries:
+
+* Dexie.js (https://github.com/dfahlander/Dexie.js) - Minimalistic IndexedDB API with bulletproof transactions
+* Tabex (https://github.com/nodeca/tabex) - Master election and in browser message bus
+
+The DataService is available as a standalone AngularJS module.
+Installation via bower:
+
+  .. code-block:: none
+
+      bower install buildbot-data --save
+
+Inject the ``bbData`` module to your application:
+
+  .. code-block:: javascript
+
+      angular.module('myApp', ['bbData'])
+
+Methods:
+
+* ``.get(endpoint, [id], [query])``: returns a promise<Collection>, when the promise is resolved, the Collection contains all the requested data
+
+  * call .getArray() on the returned promise to get the updating Collection before it's filled with the initial data
+
+  .. code-block:: coffeescript
+
+      # assign builds to $scope.builds once the Collection is filled
+      dataService.get('builds', builderid: 1).then (builds) ->
+          $scope.builds = builds
+          # load steps for every build
+          builds.forEach (b) -> b.loadSteps()
+
+      # assign builds to $scope.builds before the Collection is filled using the .getArray() function
+      $scope.builds = dataService.get('builds', builderid: 1).getArray()
+
+* ``.getXXX([id], [query])``: returns a promise<Collection>, when the promise is resolved, the Collection contains all the requested data
+
+  * it's higly advised to use these instead of the ``.get('string')`` function
+  * XXX can be the following: Builds, Builders, Buildrequests, Buildsets, Buildslaves, Changes, Changesources, Forceschedulers, Masters, Schedulers, Sourcestamps
+  * call .getArray() on the returned promise to get the updating Collection before it's filled with the initial data
+
+  .. code-block:: coffeescript
+
+      # assign builds to $scope.builds once the Collection is filled
+      dataService.getBuilds(builderid: 1).then (builds) ->
+          $scope.builds = builds
+          # load steps for every build
+          builds.forEach (b) -> b.loadSteps()
+
+      # assign builds to $scope.builds before the Collection is filled using the .getArray() function
+      $scope.builds = dataService.getBuilds(builderid: 1).getArray()
+
+* ``.open(scope)``: returns a DataAccessor, handles bindings, open a new accessor every time you need updating data in a controller
+
+  .. code-block:: coffeescript
+
+      # open a new accessor every time you need updating data in a controller
+      class DemoController extends Controller
+          constructor: ($scope, dataService) ->
+              # automatically closes all the bindings when the $scope is destroyed
+              opened = dataService.open($scope)
+              # alternative syntax:
+              #   opened = dataService.open()
+              #   opened.closeOnDestroy($scope)
+              # closing it manually is also possible:
+              #   opened.close()
+
+              # request new data, it updates automatically
+              @builders = opened.getBuilders(limit: 10, order: '-started_at').getArray()
+
+* ``.control(url, method, [params])``: returns a promise, sends a JSON RPC2 POST request to the server
+
+  .. code-block:: coffeescript
+
+      # open a new accessor every time you need updating data in a controller
+      dataService.control('Forceschedulers/force', 'force').then (response) ->
+          $log.debug(response)
+      , (reason) ->
+          $log.error(reason)
+
+* ``.clearCache()``: clears the IndexedDB tables and reloads the current page
+
+  .. code-block:: coffeescript
+
+      class DemoController extends Controller
+          constructor: (@dataService) ->
+          onClick: -> @dataService.clearCache()
+
+Methods on the Wrapper classes:
+
+* ``.getXXX([id], [query])``: returns a promise<Collection>, when the promise is resolved, the Collection contains all the requested data
+
+  * same as dataService.getXXX, but with relative endpoint
+
+  .. code-block:: coffeescript
+
+      # assign builds to $scope.builds once the Collection is filled
+      dataService.getBuilds(builderid: 1).then (builds) ->
+          $scope.builds = builds
+          # get steps for every build
+          builds.forEach (b) ->
+              b.getSteps().then (steps) ->
+                  # assign completed test to every build
+                  b.complete_steps = steps.map (s) -> s.complete
+
+* ``.loadXXX([id], [query])``: returns a promise<Collection>, the Collection contains all the requested data, it is also assigned to wrapperInstance.xxx
+
+  .. code-block:: coffeescript
+
+      $q (resolve) ->
+          # get builder with id = 1
+          dataService.getBuilders(1).then (builders) ->
+              builders.forEach (builder) ->
+                  # load all builds
+                  builder.loadBuilds().then (builds) ->
+                      builds.forEach (build) ->
+                          # load all buildsteps
+                          build.loadSteps().then -> resolve(builders[0])
+      .then (builder) ->
+          # builder has a builds field, and the builds have a steps field containing the corresponding data
+          $log.debug(builder)
+
+* ``.control(method, params)``: returns a promise, sends a JSON RPC2 POST request to the server
+
 RecentStorage
 .............
 
--- a/master/docs/developer/www.rst
+++ b/master/docs/developer/www.rst
@@ -634,7 +634,7 @@ It uses the following libraries:
 The DataService is available as a standalone AngularJS module.
 Installation via bower:
 
-  .. code-block:: none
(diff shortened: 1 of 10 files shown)


master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
     Call the control data api.
     This builds up a POST with jsonapi encoded parameters
 
+DataService
+.............
+
+DataService is the future replacement of BuildbotService for accessing the Buildbot data API.
```
tardyp: I dont think it should be marked as future. We should really drop the BuildbotService ASAP

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+It has a modernized interface for accessing data. It uses IndexedDB for storing cached data as a single data store,
```
tardyp: So no need to compare. ` it has a moden interfact for`

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+  .. code-block:: none
+
+      bower install buildbot-data --save
+
+Inject the ``bbData`` module to your application:
+
+  .. code-block:: javascript
+
+      angular.module('myApp', ['bbData'])
```
tardyp: Is it really worth to abreviate the module name, while it is only written once per plugin?

master/docs/developer/www.rst
tothandras: I am currently rewriting the base and the plugins to the DataService, but I wanted this to be correct. So this is independent from that and I can remove the 'future' in that PR.

master/docs/developer/www.rst
tothandras: done

master/docs/developer/www.rst
tothandras: It's not that significant, I can change this. Everyone use abbreviated module names so I thought that I do too.

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+          $scope.builds = builds
+          # load steps for every build
+          builds.forEach (b) -> b.loadSteps()
+
+      # assign builds to $scope.builds before the Collection is filled using the .getArray() function
+      $scope.builds = dataService.get('builds', builderid: 1).getArray()
+
+* ``.getXXX([id], [query])``: returns a promise<Collection>, when the promise is resolved, the Collection contains all the requested data
+
+  * it's higly advised to use these instead of the ``.get('string')`` function
```
tardyp: highly

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+Inject the ``bbData`` module to your application:
+
+  .. code-block:: javascript
+
+      angular.module('myApp', ['bbData'])
+
+Methods:
+
+* ``.get(endpoint, [id], [query])``: returns a promise<Collection>, when the promise is resolved, the Collection contains all the requested data
```
tardyp: I think you should describe this API later if this is not the preferred way.

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+      # assign builds to $scope.builds once the Collection is filled
+      dataService.getBuilds(builderid: 1).then (builds) ->
+          $scope.builds = builds
+          # load steps for every build
+          builds.forEach (b) -> b.loadSteps()
+
+      # assign builds to $scope.builds before the Collection is filled using the .getArray() function
+      $scope.builds = dataService.getBuilds(builderid: 1).getArray()
+
+* ``.open(scope)``: returns a DataAccessor, handles bindings, open a new accessor every time you need updating data in a controller
```
tardyp: please explain that this method automatically registers to scope.$on('$destroy') in order to release all the references and watches that this data accessor has requested.

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+              # request new data, it updates automatically
+              @builders = opened.getBuilders(limit: 10, order: '-started_at').getArray()
+
+* ``.control(url, method, [params])``: returns a promise, sends a JSON RPC2 POST request to the server
+
+  .. code-block:: coffeescript
+
+      # open a new accessor every time you need updating data in a controller
+      dataService.control('Forceschedulers/force', 'force').then (response) ->
```
tardyp: this is 'forcescehedulers/force'.

This sounds a bit weird to have the control api use a REST style endpoint, while the get api is query like.

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+It uses the following libraries:
+
+* Dexie.js (https://github.com/dfahlander/Dexie.js) - Minimalistic IndexedDB API with bulletproof transactions
+* Tabex (https://github.com/nodeca/tabex) - Master election and in browser message bus
+
+The DataService is available as a standalone AngularJS module.
+Installation via bower:
+
+  .. code-block:: none
```
sa2ajj: [nit] I'd say it's shell...

master/docs/developer/www.rst
sa2ajj: Do we need `* ` here?

master/docs/developer/www.rst
sa2ajj: ditto

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+  .. code-block:: coffeescript
+
+      # open a new accessor every time you need updating data in a controller
+      dataService.control('Forceschedulers/force', 'force').then (response) ->
+          $log.debug(response)
+      , (reason) ->
+          $log.error(reason)
+
+* ``.clearCache()``: clears the IndexedDB tables and reloads the current page
```
sa2ajj: ditto

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+  .. code-block:: coffeescript
+
+      class DemoController extends Controller
+          constructor: (@dataService) ->
+          onClick: -> @dataService.clearCache()
+
+Methods on the Wrapper classes:
+
+* ``.getXXX([id], [query])``: returns a promise<Collection>, when the promise is resolved, the Collection contains all the requested data
```
sa2ajj: ditto

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+      # assign builds to $scope.builds once the Collection is filled
+      dataService.getBuilds(builderid: 1).then (builds) ->
+          $scope.builds = builds
+          # get steps for every build
+          builds.forEach (b) ->
+              b.getSteps().then (steps) ->
+                  # assign completed test to every build
+                  b.complete_steps = steps.map (s) -> s.complete
+
+* ``.loadXXX([id], [query])``: returns a promise<Collection>, the Collection contains all the requested data, it is also assigned to wrapperInstance.xxx
```
sa2ajj: ditto

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+                  # load all builds
+                  builder.loadBuilds().then (builds) ->
+                      builds.forEach (build) ->
+                          # load all buildsteps
+                          build.loadSteps().then -> resolve(builders[0])
+      .then (builder) ->
+          # builder has a builds field, and the builds have a steps field containing the corresponding data
+          $log.debug(builder)
+
+* ``.control(method, params)``: returns a promise, sends a JSON RPC2 POST request to the server
```
sa2ajj: ditto

sa2ajj:
Otherwise, LGTM

master/docs/developer/www.rst
```diff
@@ -617,6 +617,139 @@ Several methods are added to each "restangularized" objects, aside from get(), p
...
+
+* ``.clearCache()``: clears the IndexedDB tables and reloads the current page
+
+  .. code-block:: coffeescript
+
+      class DemoController extends Controller
+          constructor: (@dataService) ->
+          onClick: -> @dataService.clearCache()
+
+Methods on the Wrapper classes:
```
tardyp: What is a wrapper class? I think we should not discuss of this implementation detail. Just say that the object returned by the getXXX methods have the following utility methods.

master/docs/developer/www.rst
tardyp: I have no idea by reading this what is the difference between loadXXX and getXXX

tardyp: after reading twice I finally understand, but this should be made clearer

tardyp: just say that `o.loadXs()` is equivalent to `o.getXs().then xs -> o.xs = xs`

www/data_module/src/mock/generator.service.coffee
```diff
@@ -0,0 +1,31 @@
+# types for generating test data: null, number, string, boolean, timestamp, <array>[], <object>, <objectName in Specification>
+class Generator extends Service
+    self = null
+    constructor: ->
+        self = @
+
+    number: (min = 0, max = 100) ->
+        random = Math.random() * (max - min) + min
```
tardyp: Its usually not very cool to use random for generating faked data.

It means that your data is not reproducible

www/data_module/src/services/dataUtils/dataUtils.service.coffee
```diff
@@ -1,52 +1,68 @@
...
-        else if angular.isString(arrayOrString)
-            # split the string to get an array
-            arrayOrString.split('/')
-        else
-            throw new TypeError("Parameter 'arrayOrString' must be a array or a string, not #{typeof arrayOrString}")
-
-    unWrap: (data, path) ->
-        type = @type(path)
-        data[type]
+            # returns the type of the endpoint
```
tardyp: curse github for such bad review experience for whitespace changes

www/data_module/src/specification.constant.coffee
```diff
@@ -1,272 +1,280 @@
...
+                    'slaveinfo'
+                ]
+                root: true
+                paths: []
+                static: true
+            changes:
+                id: 'changeid'
+                fields: [
+                    'changeid'
+                    'author'
```
tardyp: Eventually that will be cool to add a computed_fields map here. For example to add author_email field

www/data_module/src/specification.constant.coffee
```diff
@@ -1,272 +1,280 @@
...
-            paths: [
-                'changes'
-            ]
-            static: true
+            builds:
+                id: 'buildid'
+                fields: [
+                    'buildid'
+                    'builderid'
+                    'buildrequestid'
```
tardyp: Why is it important to describe all fields in this spec?

We should have a way to make sure the spec here is coherent with the dataspec generated by buildbot dataspec.
We should not reproduce the mistake of verifying it within the normal build, but at least make sure it is automated in the CI (.travisrc)

tardyp:
I am having hardtime to see what bug you have actually resolved. I'd like to understand what are the level of trickyness you are currently suffering.
Would be cool to separate bug fixes from codestyle changes

www/data_module/src/specification.constant.coffee
tothandras: I use these fields to create the IndexedDB tables, it's possible to put an object with additional fields beside these, but those won't be query-able. In our cases it doesn't really matters, because I filter the data with a similar algorithm to the server side filtering, so it's not working with the IndexedDB API directly but on all the data in a table. I wan't to use these to create mocked data, that we can use to help the offline development, without a proxy.

www/data_module/src/mock/generator.service.coffee
tothandras: I'd like to use this generator to help the offline development. In unit tests I find myself to write exact  test data for that scenario where a generator won't be really useful even if it were reproducible.

tothandras:
I fixed issues where a certain data type doesn't contain ids for the parent. Joining these table on a request was almost impossible before.

tardyp:
I think we are good to go here.
:+1:

sa2ajj:
we have failed tests, are they flaky?

PR #1825, opened 2015-08-24
Material UI : generate force build form according to fields definition returned from server

shanzi:
This PR is to enable the front end to generate the form for force build according to definitions that is returned by the server (which in return can be customized in config). And we are able to trigger a build with formal ways from new `dataService`. So the whole UI is now close to usable.

The only thing left undone is the tests for the new directive `forcebuild-form`. The reason is that the new dataService of @tothandras removed the dependency of `websocketService`, which is a dependency for request mocking and spying in tests. So many of the tests will be failed and it is hard to write and debug the tests. We may have to wait for some time for new HTTP and web socket mocking structure under new `dataService`.

@tardyp  The tests for this directive need not HTTP mocking, so I'm wondering if there is a way to run only part of the tests instead of all of them.

TextField, TextareaField, IntField, BoolField and ListField are all supported. The form generated looks like:
screen shot 2015-08-24 at 9 39 35 pm

### DIFF
--- a/www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
+++ b/www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
@@ -1,6 +1,6 @@
 class ForceDialog extends Controller
-
-    fields: {}
+    data: {}
+    fields: []
     field_errors: null
 
     cancel: ->
@@ -8,7 +8,7 @@ class ForceDialog extends Controller
 
     confirm: ->
         param = builderid: @builder.builderid
-        _.extend param, @fields
+        _.extend param, @data
         @forceBuild param
 
     forceBuildSuccess: ->
@@ -31,11 +31,13 @@ class ForceDialog extends Controller
         res.then (=> @forceBuildSuccess()), (data) => @forceBuildFail(data.error)
 
     constructor: (@builder, @scheduler, @dataService, @restService, @$mdDialog) ->
+        @fields = @scheduler.all_fields
+
         parseFields = (field) =>
             if field.nested
                 parseFields(subfield) for subfield in field.fields
             else if field.name
-                @fields[field.name] = field.default
+                @data[field.name] = field.default
 
-        parseFields(field) for field in @scheduler.all_fields
+        parseFields(field) for field in @fields
 
--- a/www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
+++ b/www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
@@ -8,7 +8,7 @@ class ForceDialog extends Controller
 
     confirm: ->
         param = builderid: @builder.builderid
-        _.extend param, @data
+        angular.extend param, @data
         @forceBuild param
 
     forceBuildSuccess: ->
--- a/www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
+++ b/www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
@@ -16,21 +16,17 @@ class ForceDialog extends Controller
         @$mdDialog.hide()
 
     forceBuildFail: (error) ->
-        if error.code == -32602
+        if error.code == -32602 # Non-networking error
             @field_errors = data.error.message
         else
             alert "Unexpected error occurs, please try again."
             @field_errors = null
 
     forceBuild: (param) ->
-        res = @restService.post 'forceschedulers/force',
-            id: @dataService.getNextId()
-            jsonrpc: '2.0'
-            method: 'force'
-            params: param
+        res = @dataService.control 'forceschedulers/force', 'force', param
         res.then (=> @forceBuildSuccess()), (data) => @forceBuildFail(data.error)
 
-    constructor: (@builder, @scheduler, @dataService, @restService, @$mdDialog) ->
+    constructor: (@builder, @scheduler, @dataService, @$mdDialog) ->
         @fields = @scheduler.all_fields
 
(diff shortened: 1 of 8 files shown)


sa2ajj:
I'm curious how it would look for multiple codebases?  Is `CodebaseParameter` supported?

www/md_base/src/app/common/directives/forcebuildform/forcebuildform.directive.coffee
```diff
@@ -0,0 +1,158 @@
...
+                when 'textarea'
+                    return @textareaField(field)
+                when 'int'
+                    return @intField(field)
+                when 'list'
+                    return @listField(field)
+                when 'bool'
+                    return @boolField(field)
+                else
+                    return @textField(field)
```
tardyp: This method is obviously much more simple than the way I used previously.
It however does not address the requirement that people can define custom field for their own need.
See for example the CodeParameter instance.
The idea is that the form is filled with custom directives whose name is "#{field.type}Field". So my plugin just has to register a new directive in order to create a new field.


I think you can achieve this while keeping the simpler plugins like this.
You just have to change the default case.
```
when 'text'
    return @textField(field)
else
    return @customField(field)
```
and customField will just generate

```
    customField: (field) ->
          elem = @element field.type + "Field"
          elem.$scope.field = field # I am sure this wont work, but you'll have to find a way to pass field as a parameter to the custom field
          return
```

I'd like you to eventually make it work with codeparameter plugin, without having to modify it.
I realize that this require to work first on making the plugin system work.
maybe that is too much work.

tardyp:
@sa2ajj for what I understand, CodebaseParameter is a fake parameter which translate to 4 text fields project, repository, branch, revision.

sa2ajj:
and `codebase` name that is used as a title for the group...

shanzi:
OK, I will dig into this :)

tardyp:
I think this can be done in subsequent PR however, so :+1: from me. Waiting for :+1: from @tothandras

sa2ajj:
Sounds like a plan :)

sa2ajj:
Though tests fail:
```
[FAIL]
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/twisted/internet/defer.py", line 1045, in _inlineCallbacks
    result = g.send(result)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/integration/test_customservices.py", line 45, in test_customService
    self.assertEqual(build['steps'][0]['state_string'], 'num reconfig: 2')
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/twisted/trial/unittest.py", line 270, in assertEqual
    % (msg, pformat(first), pformat(second)))
twisted.trial.unittest.FailTest: not equal:
a = u"'echo hei'"
b = 'num reconfig: 2'


buildbot.test.integration.test_customservices.CustomServiceMaster.test_customService
[ERROR]
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/twisted/internet/defer.py", line 1045, in _inlineCallbacks
    result = g.send(result)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/util/integration.py", line 133, in tearDown
    raise self.failureException(dump.getvalue())
twisted.trial.unittest.FailTest: FAILED! dumping build db for debug
*** BUILD 1 *** ==> finished (success)
    *** STEP shell *** ==> num reconfig: 1 (success)
        log:stdio (88)
        *********** LOG: stdio *********
        ********************************
*** BUILD 2 *** ==> finished (success)
    *** STEP shell *** ==> 'echo hei' (success)
        log:stdio (88)
        *********** LOG: stdio *********
        ********************************


buildbot.test.integration.test_customservices.CustomServiceMaster.test_customService
```

www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
```diff
@@ -8,7 +8,7 @@ class ForceDialog extends Controller
 
     confirm: ->
         param = builderid: @builder.builderid
-        _.extend param, @fields
+        _.extend param, @data
```
tothandras: How many times do you use underscore? I think angular.extend/.merge is just fine here.

www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
```diff
@@ -16,26 +16,24 @@ class ForceDialog extends Controller
         @$mdDialog.hide()
 
     forceBuildFail: (error) ->
-        if error.code == -32602
+        if error.code == -32602 # Non-networking error
```
tothandras: maybe put it in a constant?

tothandras:
I am not sure what I have removed, we can talk about it in chat. 
Here is my :+1:

www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
shanzi: put it at `common.constant.coffee` ? As this constant won't be used at other place, I think a comment might be ok enough :)

www/md_base/src/app/builds/builder/forcedialog/forcedialog.controller.coffee
shanzi: ok

tardyp:
I think Andras should provide those Constants in buildbot data.

tothandras:
I will, good idea!

shanzi:
How about this pull request now? Could it be merged?

tardyp:
I forgot about it.

PR #1826, opened 2015-08-24
ws.py: call parent constructor

tardyp:
should fix http://trac.buildbot.net/ticket/3332


benallard:
Applied the patch locally and the issue has yet to be seen !  :+1:

PR #1828, opened 2015-08-27
Fix try scheduler, mark test as no longer flaky

djmitche:
This flaky test wasn't being run, and as a result we weren't seeing
the fact that the scheduler was subscribing to a nonexistent message
stream.

This also fixes a loop that would bail out on seeing just one
duplicate build, rather than simply skipping that build.  I doubt this
ever caused problems, but this is more correct.

### DIFF
--- a/master/buildbot/data/builds.py
+++ b/master/buildbot/data/builds.py
@@ -143,6 +143,14 @@ class BuildsEndpoint(Db2DataMixin, base.Endpoint):
             buildscol.append(data)
         defer.returnValue(buildscol)
 
+    def startConsuming(self, callback, options, kwargs):
+        builderid = kwargs.get('builderid')
+        if builderid is not None:
+            path = ('builders', str(builderid), 'builds', None, None)
+        else:
+            path = ('builds', None, None)
+        return self.master.mq.startConsuming(callback, path)
+
 
 class Build(base.ResourceType):
 
--- a/master/buildbot/test/integration/test_try_client.py
+++ b/master/buildbot/test/integration/test_try_client.py
@@ -23,7 +23,6 @@ from buildbot.clients import tryclient
 from buildbot.schedulers import trysched
 from buildbot.test.util import dirs
 from buildbot.test.util import www
-from buildbot.test.util.decorators import flaky
 from twisted.cred import credentials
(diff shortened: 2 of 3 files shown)


master/buildbot/data/builds.py
```diff
@@ -143,6 +143,16 @@ def get(self, resultSpec, kwargs):
             buildscol.append(data)
         defer.returnValue(buildscol)
 
+    def startConsuming(self, callback, options, kwargs):
```
sa2ajj: `**kwargs`?

master/buildbot/data/builds.py
```diff
@@ -143,6 +143,16 @@ def get(self, resultSpec, kwargs):
...
+        builderid = kwargs.get('builderid')
+        if builderid is not None:
+            return self.master.mq.startConsuming(callback,
+                                                 ('builders', str(builderid),
+                                                  'builds', None, None))
+        else:
+            return self.master.mq.startConsuming(callback,
```
sa2ajj: it's a personal preference, but I'd rather calculate the routing key and then use _one_ call to `self.master.mq.startConsuming`

master/buildbot/test/integration/test_try_client.py
```diff
@@ -133,7 +132,7 @@ def getSourceStamp(vctype, treetop, branch=None, repository=None):
 
         # stub out printStatus, as it's timing-based and thus causes
         # occasional test failures.
-        self.patch(tryclient.Try, 'printStatus', lambda: None)
+        self.patch(tryclient.Try, 'printStatus', lambda self: None)
```
sa2ajj: probably `lambda _: None` would be better.

master/buildbot/data/builds.py
djmitche: No, the kwargs have already been gathered - they're all spelled like this

djmitche:
amended accordingly

sa2ajj:
:+1:

djmitche:
For some reason this is no longer passing tests for me :(

djmitche:
Huh, I had a too-old version of something - Twisted-15.3.0 maybe

PR #1830, opened 2015-08-30
Add a `maxcpus` buildslave property

djmitche:
This should add the property 'maxcpus' to a buildslaves configuration.
If 'maxcpus' is not specified it defaults to 'None' which uses
'multiprocessing.cpu_count()' to receive the max amount of CPUs.

(rebased from https://github.com/buildbot/buildbot/pull/1426, +pep8)

This is an update of #1426.

This still needs more work, though: the value is totally ignored on the master side.  In fact, I don't think any slave information is easily available to custom steps.

Am I missing something?

### DIFF
--- a/master/docs/developer/master-slave.rst
+++ b/master/docs/developer/master-slave.rst
@@ -61,6 +61,8 @@ The slave-side Bot object has the following remote methods:
         OS the slave is running (extracted from Python's os.name)
     ``basedir``
         base directory where slave is running
+    ``numcpus``
+        number of CPUs on the slave, either as configured or as detected (since buildbot-slave version 0.9.0)
 
 :meth:`~buildslave.bot.Bot.remote_getVersion`
     Returns the slave's version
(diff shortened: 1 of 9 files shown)


jaredgrubb:
Looks good so far!

tardyp:
:+1: 

I think we can add the possibility to access slave info with interpolate, like we do for accessing sourcestamp.
Second simpler option is to just populate a numcpu property at startbuild time.

djmitche:
@tardyp just to be sure: you're suggesting merging this, with work to make slave info available master-side as a follow-up project?

djmitche:
(not that I'm promising to work that project!)

tardyp:
Correct

Le lun. 31 août 2015 15:33, Dustin J. Mitchell <notifications@github.com> a
écrit :

@tardyp <https://github.com/tardyp> just to be sure: you're suggesting
merging this, with work to make slave info available master-side as a
follow-up project?


—
Reply to this email directly or view it on GitHub
<https://github.com/buildbot/buildbot/pull/1830#issuecomment-136372533>.

PR #1793, opened 2015-07-27
Switch git to inline callbacks

vlovich:
Rebase off master

PR #1833, opened 2015-08-30
Switch git to inline callbacks (update #1793)

djmitche:
This updates #1793 by merging to master and resolving some conflicts and quite a bit of bitrot.

Issues:
 * `_updateSubmodule` and `_cleanSubmodule` were asynchronous but called from `clean` without `yield`
 * `copy` called `mode_incremental` and expected a return code, but that method returns nothing
 * `startVC` called the parent `patch` method without supplying a value for `_`.

@vlovich what do you think?

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -150,40 +150,35 @@ class Git(Source):
         if not isinstance(self.getDescription, (bool, dict)):
             bbconfig.error("Git: getDescription must be a boolean or a dict.")
 
+    @defer.inlineCallbacks
     def startVC(self, branch, revision, patch):
         self.branch = branch or 'HEAD'
         self.revision = revision
         self.method = self._getMethod()
         self.stdio_log = self.addLogForRemoteCommands("stdio")
 
-        d = self.checkBranchSupport()
+        try:
+            gitInstalled = yield self.checkBranchSupport()
 
-        @d.addCallback
-        def checkInstall(gitInstalled):
             if not gitInstalled:
                 raise BuildSlaveTooOldError("git is not installed on slave")
-            return RC_SUCCESS
 
-        d.addCallback(lambda _: self.sourcedirIsPatched())
+            patched = yield self.sourcedirIsPatched()
 
-        @d.addCallback
-        def checkPatched(patched):
             if patched:
-                return self._dovccmd(['clean', '-f', '-f', '-d', '-x'])
-            else:
-                return RC_SUCCESS
+                yield self._dovccmd(['clean', '-f', '-f', '-d', '-x'])
 
-        d.addCallback(self._getAttrGroupMember('mode', self.mode))
-        if patch:
-            d.addCallback(self.patch, patch)
-        d.addCallback(self.parseGotRevision)
-        d.addCallback(self.parseCommitDescription)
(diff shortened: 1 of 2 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -222,14 +217,26 @@ def mode_incremental(self, _):
...
-        d.addCallback(self._updateSubmodule)
-        d.addCallback(self._cleanSubmodule)
-        return d
+        rc = yield self._dovccmd(command)
+        if rc != RC_SUCCESS:
+            raise buildstep.BuildStepFailed
+
+        rc = yield self._fetchOrFallback()
+        if rc != RC_SUCCESS:
+            raise buildstep.BuildStepFailed
```
Stibbons: What about refactor in a loop:

for clean_meth in ['_fetchOrFallback', '_syncSubmodule', ...]:
    rc = yield getattr(self, clean_meth)
    if rc != RC_SUCCESS:
        raise buildstep.BuildStepFailed

tardyp: I don't like the loop think, but I dont like the lots of if either.

In repo, there is a abandonOnFailure in _dovccmd likes. I think this is a better method for handling this problem.

master/buildbot/steps/source/git.py
```diff
@@ -251,46 +258,35 @@ def fresh(self):
...
-        cmd = remotecommand.RemoteCommand('rmdir', {'dir': self.workdir,
-                                                    'logEnviron': self.logEnviron,
-                                                    'timeout': self.timeout, })
-        cmd.useLog(self.stdio_log, False)
-        d = self.runCommand(cmd)
+        yield self.runRmdir(self.workdir, abandonOnFailure=False,
+                            timeout=self.timeout)
 
         old_workdir = self.workdir
         self.workdir = self.srcdir
```
tardyp: smelly

master/buildbot/steps/source/git.py
```diff
@@ -251,46 +258,35 @@ def fresh(self):
...
-        def resetWorkdir(_):
+        try:
+            yield self.mode_incremental()
+            cmd = buildstep.RemoteCommand('cpdir',
+                                          {'fromdir': self.srcdir,
+                                           'todir': old_workdir,
+                                           'logEnviron': self.logEnviron,
+                                           'timeout': self.timeout, })
+            cmd.useLog(self.stdio_log, False)
+            yield self.runCommand(cmd)
```
tardyp: wouldn't it better to say here

    yield self.runCommand(cmd, workdir= self.srcdir)

tardyp:
Thanks for working on this.  The new code looks so much better!

I would expect 90% diff hit is a little bit low for a source step. especially that you found issues with the old code.

master/buildbot/steps/source/git.py
djmitche: Yeah, the idea is to re-use the full checkout code but pointed at the srcdir.

djmitche:
Well, I just rebased it and fixed some test failures.  I'm really not sure where those test failures came from, but I think it's just bitrot - the original patch passed its tests before rebase.

As for the suggested changes, I'll leave that to @vlovich

tardyp:
ok, lets merge it then :+1:

PR #1831, opened 2015-08-30
maybeStartBuildsForSlave after insubstantiating

djmitche:
(also handle the deferred from self.insubstantiate elsewhere)

PR #1834, opened 2015-08-31
Backport GH-1831 to 0.8.x: maybeStartBuildsForSlave after insubstantiating

sa2ajj:


devoid:
I can confirm that this patch fixes the issue I was seeing--my DockerLatentBuildSlave spawns a new container for queued builds after the first exits. Thanks for all of your help!

PR #1840, opened 2015-09-06
Rebuild

tardyp:
fixes: http://trac.buildbot.net/ticket/2639

as there is no add Buildrequest api, I tried to reuse as much as possible the addBuildSet api, thus we don't share the buildset, nor properties, but we share the sourcestamp


Stibbons:
What about rebuild with property edition (some talked this noon about this "killing" feature)?

tardyp:
Yes this is only valid for force build which are not so common.

It's actually a much easier story. Just ui..

Le lun. 7 sept. 2015 20:05, Gaetan Semet <notifications@github.com> a
écrit :

jaredgrubb:
Being able to Rebuild+Edit Properties would be amazing! (As would the ability to have a URL pre-fill a form).

master/buildbot/data/builds.py
```diff
@@ -100,12 +102,31 @@ def startConsuming(self, callback, options, kwargs):
                                                  ('builds', str(buildid), None))
 
     def control(self, action, args, kwargs):
-        if action != "stop":
+        # we convert the action into a mixedCase method name
+        action_method = getattr(self, "action" + action[0].upper() + action[1:])
```
jaredgrubb: FWIW, your code is fine, but you can do "action[0].capitalize()" for the same effect.

tardyp:
addressed review comments, and python 2.6 issues

ssavrim:
:+1:

PR #1841, opened 2015-09-08
interfaces: isRenderable

yetanotherion:
Method to compute whether an object
is renderable.


sa2ajj:
Do we already have users for this method?

jaredgrubb:
Looks good, but a unit test or two could be good.

tardyp:
after digging, it looks like this patch is needed by workdir renderable patch we have on our old buildbot eight. After discussing with the devs, I believe that the isRenderable is not really needed. @yetanotherion  if you agree, please close

yetanotherion:
closed, good we to all !

PR #1849, opened 2015-09-15
fix deprecated source step name: 'git' => 'Svn'

asashnov:
fix deprecated source step name: 'git' => 'Svn'

### DIFF
--- a/slave/buildslave/commands/removed.py
+++ b/slave/buildslave/commands/removed.py
@@ -22,8 +22,16 @@ class RemovedSourceCommand(base.SourceBaseCommand):
     def start(self):
         self.sendStatus(
             {"header":
-             "slave-side source checkout for '%s' is no longer supported by "
-             "build slave of version %s\n" % (self.name, buildslave.version)})
+             "slave-side source checkout for '{0}' is no longer supported by "
+             "build slave of version {1}\n"
+             "\n"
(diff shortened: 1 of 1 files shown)


slave/buildslave/commands/removed.py
```diff
@@ -28,7 +28,7 @@ def start(self):
 
 
 class Svn(RemovedSourceCommand):
-    name = "git"
+    name = "Svn"
```
tardyp: good catch!
Are you sure of the capital S?

sa2ajj: I think it should be 'svn' (vs 'Subversion')

asashnov: I changed it to capitalized.

And we may also provide more informative and instructive text like:

Since BuildBot 0.9 old slave-side-logic source checkout method
buildbot.steps.source.SVN was removed.
Please use buildbot.steps.source.svn.SVN instead.

tardyp:
I though the name was used in order to identify the command from the master, but it is not the case.
The identifier is in registry.py.

Please feel free to update the message line 25 of that file to something you find clearer

asashnov:
Now build log looks like this:
https://gist.github.com/asashnov/19f46f4b4b61453b133d#file-gistfile1-txt

PR #1852, opened 2015-09-18
fix gerrit change source verbosity

tardyp:
It is not very useful to log all the gerrit events, especially when like me you are connected to  a 2k users gerrit.

This PR fixes reconfig for this change source.

There are still the other change sources to make reconfigurable.


sa2ajj:
I feel your pain! :)

PR #1854, opened 2015-09-19
p4poller: do not rely on server time to get a list of changes

thedylman:
#head is equivalent to now but does not depend on the p4 server time.

This is to workaround an issue I have where the p4 mirror server I polls is an hour behind of the server I commit to and so changes are delayed by an hour (both are in the same timezone). Need to fix this server side, but Buildbot also should not depend on the server time for getting changes.

### DIFF
--- a/master/buildbot/changes/p4poller.py
+++ b/master/buildbot/changes/p4poller.py
@@ -214,7 +214,7 @@ class P4Source(base.PollingChangeSource, util.ComparableMixin):
             args.extend(['-P', self._getPasswd()])
         args.extend(['changes'])
         if self.last_change is not None:
-            args.extend(['%s...@%d,now' % (self.p4base, self.last_change + 1)])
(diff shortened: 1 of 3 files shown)


tardyp:
:+1: I would appreciate +1 from @bdbaddog , who wrote the most of this step

sa2ajj:
please rebase

sa2ajj:
@thedylman what version of P4 supports this and where this is documented?

I checked `p4 help revisions` and `p4 help changes` (2014.2/1045032) and did not see anything about '>' notation.

thedylman:
I got it from http://stackoverflow.com/a/12059203 I'll try and find it the docs. If I can't then #head should do the equivalent thing.

sa2ajj:
Unless it's clearly documented somewhere in P4's docs, I'd rather go with `#head`.

thedylman:
I could not find it in the docs so changed it to use #head

sa2ajj:
Thank you

PR #1856, opened 2015-09-21
Add a nine migration guide

djmitche:
..and also refactor the installation docs quite a bit

This links to http://trac.buildbot.net/wiki/NineMigrationGuide for the wiki bit of the documentation.  If that accumulates some good material, we can easily move it into the docs.


tardyp:
According to the commit message, there is not so much new things to review, appart a link, and split into several files.
:+1: based on that assumption

djmitche:
Well, `nine-upgrade.rst` is new.  But yeah, not much to review :)

PR #1857, opened 2015-09-21
Clean up buildrequest messages

djmitche:


master/buildbot/data/buildrequests.py
```diff
@@ -164,9 +164,9 @@ class EntityType(types.Entity):
 
     @defer.inlineCallbacks
     def generateEvent(self, brids, event):
-        for _id in brids:
+        for id in brids:
```
tardyp: even if this work, I'm not so found of using reserved words as identifiers. This is why I used _id, if you dont like the _ we can call it brid

tardyp:
W: 20, 0: Unused import ascii2unicode (unused-import)

LGTM apart from that

master/buildbot/data/buildrequests.py
djmitche: oh, good point - I'll adjust

jaredgrubb:
LGTM!

sa2ajj:
(all green)

PR #1860, opened 2015-09-21
Log debugging info instead of printing it

djmitche:
Logging doesn't appear in the summary, but is available afterward.
Printing appears when the test is run, which may be many lines above the
summary.

This also adjusts the assertion messages to include information about
the received value, rather than just stating that it differs from the
expected value.

### DIFF
--- a/master/buildbot/test/util/steps.py
+++ b/master/buildbot/test/util/steps.py
@@ -70,12 +70,15 @@ def _describe_cmd_difference(exp, command):
 
     missing_in_exp, missing_in_cmd, diff = _dict_diff(exp.args, command.args)
     if missing_in_exp:
-        print('Keys in cmd missing from expectation: {0}'.format(missing_in_exp))
+        log.msg(
+            'Keys in cmd missing from expectation: {0}'.format(missing_in_exp))
     if missing_in_cmd:
-        print('Keys in expectation missing from command: {0}'.format(missing_in_cmd))
+        log.msg(
+            'Keys in expectation missing from command: {0}'.format(missing_in_cmd))
     if diff:
-        formatted_diff = ['"{0}": expected {1}, vs actual {2}'.format(*d) for d in diff]
-        print('Key differences between expectation and command: {0}'.format(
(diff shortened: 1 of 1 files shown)


master/buildbot/test/util/steps.py
```diff
@@ -335,7 +356,7 @@ def check(result):
 
     @defer.inlineCallbacks
     def _validate_expectation(self, exp, command):
-        got = (command.remote_command, command.args)
+        got=(command.remote_command, command.args)
```
tardyp: pep8?

master/buildbot/test/util/steps.py
```diff
@@ -51,9 +51,12 @@ def _dict_diff(d1, d2):
 
     for k in both:
         if isinstance(d1[k], dict) and isinstance(d2[k], dict):
-            missing_in_v1, missing_in_v2, different_in_v = _dict_diff(d1[k], d2[k])
-            missing_in_d1.extend(['{0}.{1}'.format(k, m) for m in missing_in_v1])
-            missing_in_d2.extend(['{0}.{1}'.format(k, m) for m in missing_in_v2])
+            missing_in_v1, missing_in_v2, different_in_v = _dict_diff(
```
tardyp: not sure if those are really needed. I usually cut at 100. cutting at 72 will mean a lot of such refacto, and to me loss of readability.

tardyp:
some nits. obviously :+1: on the principle

master/buildbot/test/util/steps.py
djmitche: I think autopep8 did it (and the `got=(..` below)

tardyp: mmhh very weird. I think you have a wrong autopep8 version or badly configured editor?

the autopep8 from the hooks is enforcing 100 columns, not 72.
I just verified that I can modify that file, and autopep8 wont remove the spaces between = nor reduce lines

tardyp:
:+1: with the new version

PR #1861, opened 2015-09-23
Add a GerritChangeSource that does not produce unrelated changes

sa2ajj:


tardyp:
This means that of you have n projects, you need n ssh connections. I'm not sure that we want to keep this API for long term

sa2ajj:
I have a slightly different use case: one buildmaster looks for a particular branch of a particular project.

tardyp:
Yes, I understand, but the case 1 filter is just a particularization of the case n filters.
The api you are proposing may not be enhanced easily to the case n filters.

I don't think the version in buildbot_travis is suitable either: https://github.com/tardyp/buildbot_travis/blob/master/buildbot_travis/vcs/gerrit.py

This is why I did not upstream it. I think we should find a way to have a simple configuration *and* share the ssh process in case of multiproject master.

sa2ajj:
I am not sure what you mean by "share", however I agree about the API.

sa2ajj:
Let's do it for master first.

PR #1863, opened 2015-09-26
make it possible to filter Gerrit events before they land in the database

sa2ajj:


tardyp:
I like the api better. I think you should put this optional parameter in
the base class without the need for a new class.

Also don't override init and rather use checkConfig/reconfigService.
See my last pr about gerrit change source.

Le sam. 26 sept. 2015 11:07, Mikhail Sobolev <notifications@github.com> a
écrit :

sa2ajj:
I've been thinking about adding `filter` parameter to the `GerritChangeSource` that would basically do the same thing as `filter` in all those schedulers.  The only difference would be that those changes that are filtered here are not added to the database and, as a result, would not "clutter" console view.

sa2ajj:
(and I'll check the PR you mentioned, GH-1852, right?)

tardyp:
Humm, you are using changeFilter API.. I didn't though about that. Good idea!

I think it makes a lot of sense to have a filter at the source, and then a routing filter (that's what scheduler's changefilter are meant for)

I like a lot how it turns out!

sa2ajj:
Hmm... Looks my changes touch quite a few other bits :(  I'll need (for sure) to re-arrange commits and make them to be more sensible (a difficult part)

sa2ajj:
I really want to complete this, it's just a matter of time.

djmitche:
Closed per http://trac.buildbot.net/wiki/StalePullRequestPolicy -- this can always be re-opened or a new PR opened!

PR #1865, opened 2015-09-29
Add support for running npm commands on Windows

alexandrul:
This is the only form of ```npm``` commands that work on my PC:
 ```subprocess.call([...], shell=True)```

Enviroment:
* Windows 7 SP1 x64
* Python 2.7.10

[Review on Reviewable](https://reviewable.io/reviews/buildbot/buildbot/1865)

### DIFF
--- a/pkg/buildbot_pkg.py
+++ b/pkg/buildbot_pkg.py
@@ -104,8 +104,14 @@ def build_js(cmd):
         npm_bin = check_output("npm bin").strip()
         assert npm_version != "", "need nodejs and npm installed in current PATH"
         assert LooseVersion(npm_version) >= LooseVersion("1.4"), "npm < 1.4 (%s)" % (npm_version)
-        cmd.spawn(['npm', 'install'])
-        cmd.spawn([os.path.join(npm_bin, "gulp"), 'prod', '--notests'])
+        npm_cmd = ['npm', 'install']
+        gulp_cmd = [os.path.join(npm_bin, "gulp"), 'prod', '--notests']
+        if os.name == 'nt':
+            subprocess.call(npm_cmd, shell=True)
+            subprocess.call(gulp_cmd, shell=True)
+        else:
+            cmd.spawn(npm_cmd)
+            cmd.spawn(gulp_cmd)
 
     cmd.copy_tree(os.path.join(package, 'static'), os.path.join("build", "lib", package, "static"))
 


sa2ajj:
:+1:

tardyp:
I dont think we shall have a different thing for posix and for windows.
subprocess.call(list, shell=True) is not supposed to be working as shell=True you are supposed to pass a string.

alexandrul:
From 2.7.10 documentation at https://docs.python.org/2/library/subprocess.html#popen-constructor :
> On Windows, if args is a sequence, it will be converted to a string in a manner described in Converting an argument sequence to a string on Windows. This is because the underlying CreateProcess() operates on strings.
> The shell argument (which defaults to False) specifies whether to use the shell as the program to execute. If shell is True, it is recommended to pass args as a string rather than as a sequence.

The description is the same for 3.5.0: https://docs.python.org/3/library/subprocess.html#popen-constructor

tardyp:
shell=True also interpret the shell syntax. What if there is a space inside the path for npm?
or a %?

Are you really sure you can't make it work without shell=True

alexandrul:
>shell=True also interpret the shell syntax. What if there is a space inside the path for npm?
or a %?

That would be a problem for everything that invokes npm from a command line, wouldn't it? I'm not sure why buildbot would assume the responsibility to ensure a proper npm execution.

> Are you really sure you can't make it work without shell=True

Unlike git, which added a ```git.exe``` in the windows builds, npm only provides a ```npm.cmd```, and like I said in the first comment:
>This is the only form of npm commands that work on my PC

Are there any other windows users of buildbot? How are they installing it?

tardyp:
> shell=True also interpret the shell syntax. What if there is a space inside the path for npm?
or a %?
> That would be a problem for everything that invokes npm from a command line, wouldn't it? I'm not sure why buildbot would assume the responsibility to ensure a proper npm execution.
See second answer from this SO post.
http://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess


> Are you really sure you can't make it work without shell=True
>Unlike git, which added a git.exe in the windows builds, npm only provides a npm.cmd, and like I said in the first comment:

> This is the only form of npm commands that work on my PC
>Are there any other windows users of buildbot? How are they installing it?

You are supposed to use the wheels which are provided in pypi.
I agree that we should make the setup work on windows, but I really would like to avoid "if windows" as much as possible. These build scripts are already too hacky.

sa2ajj:
@alexandrul it's not Buildbot, it's underlying machinery (= Twisted)

alexandrul:
@tardyp everybody agrees that `shell=True` is dangerous if the command contains *user input*, which clearly is not the case here.

@sa2ajj not sure what Twisted has to do with my patch

Since I have limited free time, I'm not sure when I will be able to provide a patch with `shell=False`.
If my interest in using buildbot on Windows are bothering you, please close the issue (and maybe delete the first line from http://trac.buildbot.net/wiki/RunningBuildbotOnWindows ).

sa2ajj:
@alexandrul Plesae disregard my comment.

I'm OK with this change.

tardyp:
> Since I have limited free time, I'm not sure when I will be able to provide a patch with shell=False.
> If my interest in using buildbot on Windows is bothering you, please close the issue (and maybe delete the first line from http://trac.buildbot.net/wiki/RunningBuildbotOnWindows ).

Please be sure we have interest in running buildbot on windows, and continue to seek windows experts to help, but we also need to keep the high level of quality for our code, and imho q&d windows hacks have to be avoided.
If we accept this hack, I am pretty sure that nobody else will bother fixing npm build for windows, and we will stick with a ugly ```if windows``` forever.

alexandrul:
@tardyp no worries, I can patch it for myself, and this isn't the only tool that's using ugly windows hacks during the setup/installation part.

However, it would be nice to include into the documentation some workarounds for the ones that would like to improve buildbot on windows (you know, removing as many barriers to entry as possible).

After all, ensuring an easier setup on windows might actually enable more users to actively improve buildbot, and some day provide a better fix for this particular issue.

tardyp:
better than nothing..

PR #1866, opened 2015-10-02
Get results of build in StatusPush handler

burrows-labs:
The 'buildFinished' event indicates to the StatusPush callback that the build is complete.  Considering that all steps of the build have completed, it's useful to know whether they all completed successfully or otherwise.

This is easy to accomplish because StatusPush::buildFinished already has the success/error value available. Simply push it with the rest of the event onto the queue.


tardyp:
I think this is okay change, can you please add a not in relnotes/index.rst?

burrows-labs:
@tardyp release note added.

tardyp:
We usually link to the doc of the component using bb:ref. See the note just
before.

burrows-labs:
@tardyp ack

PR #1867, opened 2015-10-02
Results: move from status to process

jaredgrubb:
This patch moves the "results" stuff into the process submodule.

Note that I did remove that "status.results" module completely -- which could be a breaking change. An alternative is to add a "from ... import *" to that file so it still works, but maybe this is a good breaking change to make?


tardyp:
super cool! and there is a relnote in bonus
:+1: :+1: 

yeah, thats a double thumbs up !

djmitche:
I missed this PR but .. awesome!!

I think this is easy enough for users to handle in upgrading, so while it may be breaking, it's easily overcome.

PR #1871, opened 2015-10-09
[nine] Fix #3248: BuildStep: fix up interrupt logic

jaredgrubb:
THis is a forward-port of the fix on eight (9ee4dbc)

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -658,11 +658,24 @@ class BuildStep(results.ResultComputingConfigMixin,
         raise NotImplementedError("your subclass must implement run()")
 
     def interrupt(self, reason):
+        # TODO: consider adding an INTERRUPTED or STOPPED status to use
+        # instead of FAILURE, might make the text a bit more clear.
+        # 'reason' can be a Failure, or text
         self.stopped = True
         if self._acquiringLock:
             lock, access, d = self._acquiringLock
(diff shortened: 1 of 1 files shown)


sa2ajj:
LGTM

master/buildbot/process/buildstep.py
```diff
@@ -658,12 +658,25 @@ def start(self):
         raise NotImplementedError("your subclass must implement run()")
 
     def interrupt(self, reason):
+        # TODO: consider adding an INTERRUPTED or STOPPED status to use
```
tardyp: I would appreciate you remove those 3 lines, and make sure that CANCELLED is set as the status if self.stopped is True

jaredgrubb: Sorry, I missed this comment.

I think that would be valuable, but I'd rather do it in a new pull-request because that is a different issue from fixing the interrupt logic.

tardyp: fair enough :+1:

jaredgrubb:
Pushed a rebase. If the travis comes back green I'll merge.

djmitche:
Looks like the Travis failure was just a normal npm failure.  It's cute how it pretends it's a kernel with `EPEERINVALID`.

I re-triggered, but even if it still comes back red please feel free to merge.

jaredgrubb:
Ha thanks. Merging!

PR #1872, opened 2015-10-09
[WIP] RFC: Git Mirrors

jaredgrubb:
I am posting this as an early RFC on the approach. The basic idea is to create an object that represents a full-fidelity git-clone of a repo (not just a one-time, quick checkout that Git currently tries to do) that can be robustly shared across builders (in order to be economic with network activity and disk usage).

I want to solve a couple problems:

1. share git objects across many builders
  -- The current Git step does support "reference" but I dont think it's robust enough to really be shared across multiple builders (eg, there's no locking inherent)

2. support multiple remotes per git-repo
  --  eg, Git step can have both "buildbot/buildbot" and "jaredgrubb/buildbot" available, each named and selectable via the "repo" sourcestamp parameter. This kinda works today, but I have had issues flipping between them.

3. support full clones, including dangling tags
  -- Git step tries to be frugal and only clone exactly the revision needed. For example, Git step may not even interact with network if a revision is already present in the workdir. I want to be able to force fetching tags even if the revision already exists (I use "git describe" a lot for downstream steps, and that doesnt work so well if tags are not fetched).

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -64,12 +64,16 @@ git_describe_flags = [
 
 class GitMirrorRepo(object):
 """
-    [{
+Example:
+    git_mirror = GitMirrorRepo("my-repo", workdir, {
         'repourl': 'http://whereever/project.git',
         'origin':  'origin',    # required
         'alwaysUpdate': False,  # default True
         'fetchTags': True,      # default False
-    }]
+    })
+    
+    #...
+    steps.append(Git(mirrorRepo=git_mirror, ...))
 """
 
     def __init__(self, name, workdir=None, repo_config=None, lock=None):
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -62,6 +62,118 @@ git_describe_flags = [
 ]
 
 
+class GitMirrorRepo(object):
+"""
+    [{
+        'repourl': 'http://whereever/project.git',
+        'origin':  'origin',    # required
+        'alwaysUpdate': False,  # default True
+        'fetchTags': True,      # default False
+    }]
+"""
+
+    def __init__(self, name, workdir=None, repo_config=None, lock=None):
+        if not repo_config:
+            bbconfig.error("GitMirrorRepo: must provide a 'repo_config' dict")
+
+        if not lock:
+            # Access to the mirror should be exclusive on each slave.
+            # Git probably could handle concurrent accesses, but we might
+            # clobber the mirror repo if we think it's corrupt. The only
+            # way to protect that kind of thing is to make this all exclusive.
+            # Besides, making this one step exclusive should not be a big deal.
+            lock = SlaveLock("%s-GitMirrorRepo" % name, maxCount=1)
+        
+        self.name = name
+        self.workdir = workdir
+        self.repo_config = repo_config
+        self.lock = lock
+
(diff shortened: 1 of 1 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -62,6 +62,122 @@ def isTrueOrIsExactlyZero(v):
...
+        else:
+            command = ['config', remote_tags_key, expect_tags]
+            yield dovccmd(command, abandonOnFailure=False)
+
+    @defer.inlineCallbacks
+    def doCheckMirrorConfig(self, dovccmd):
+        """ Ensure the mirror's config is right. """
+
+        # Easiest to get the full config all at once.
+        git_config = yield dovccmd(['config', '--list', '-z'],
```
sa2ajj: how the config is expected to be configured?

From my point of view, every time you need to touch a slave, you have extra burden.

jaredgrubb: No, it's actually fairly easy. 

In your master config, you would define one GitRepoMirror instance (to describe what the clone represents), and then pass that same instance as a param to each Git step (one or more builders). 

The idea is that the Mirror represents _one_ directory on a slave, and then each Git step clones that to its builder's workdir. 

The Mirror class has the job of maintaining that one directory (checking its config, fetching, etc), which is a bare clone since it is not directly used. It then also has a SlaveLock to serialize things (in case two builders are allowed to run concurrently on the same slave; each of the Git steps that touch the mirror's directory should run independently).

sa2ajj:
@jaredgrubb could you please elaborate?

| I dont think it's robust enough to really be shared across multiple builders (eg, there's no locking inherent)

jaredgrubb:
I've personally never understood how to use the 'reference' parameter safely. I'm supposed to pass it a directory, but how is _that_ directory configured and managed? What makes sure it fetches, that its remote-config is valid (and handles a reconfig correctly), etc?

But let's assume you figure all _that_ out and it's managed... I assume that one of the decisions that management might make is "the reconfig is so different that we need to just 'rm -rf' and start over". Does that manager make sure any builders relying on that reference are not actively reading it? I think this means every Git step relying on it needs to be completely destroyed and remade (since the underlying reference links are gone and replaced by something it no longer knows). Etc.

This patch is about formalizing the management of that shared repository in a class-instance that can be shared explicitly to all your builders, such that they can coordinate and make sure things stay coherent and robust, especially across reconfigs. (Plus the other full-fidelity benetifs are big boons for my use case)

sa2ajj:
:+1:

tardyp:
Looks good to me. You should eventually make a statement in the doc this will not work if the mirror is share between masters.
I tend to agree that this should be better handled by git (I agree that --reference not so useful in the CI case)

tardyp:
Hi @jaredgrubb will you have the time to finish it?

jaredgrubb:
I have been playing with this, but I ran into some problems with this approach. Sharing a git repo (and making sure it can work with multiple remotes!) ends up being a lot harder than I thought and I havent been able to find a good solution. I think there is one, I just havent had time to flush it out. I think if we are done with eight, then we should close this. The ideas will transfer forward to nine, when I (or anyone) has time to do it right.

djmitche:
eight and nine have *substantial* differences now, so I think writing this to the eight branch is not a good use of time.  The idea is still sound, of course!

PR #1874, opened 2015-10-12
Add Transform, a renderable for combining other renderables

SpecLad:


tardyp:
I am not against it, its funny from a hacker point of view (in the noble mean)

I would highly suggest my fellow devs not to use it for maintainability prospective.
I mean, for me, it is too much a mind twister. rendereable concept is already giving a hard time to many people.

:+1:  high quality work!

master/buildbot/process/properties.py
```diff
@@ -743,3 +743,27 @@ def getRenderingFor(self, build):
...
+    A renderable that combines other renderables' results using an arbitrary function.
+    """
+
+    implements(IRenderable)
+
+    def __init__(self, function, *args, **kwargs):
+        if not callable(function) and not IRenderable.providedBy(function):
+            config.error("function given to Transform neither callable nor renderable")
+
+        self.function = function
```
sa2ajj: I know that we do not consistently follow this, but for new code (at least) I prefer those members to be "private":

```python
    self._function = function
    ...
```

In the past we had quite a number of nightmares related to API management (since all non-private data members are essentially part of the API).

master/buildbot/process/properties.py
```diff
@@ -743,3 +743,27 @@ def getRenderingFor(self, build):
...
+        self.args = args
+        self.kwargs = kwargs
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, iprops):
+        rfunction = yield iprops.render(self.function)
```
sa2ajj: Rendering function??

sa2ajj:
Otherwise, LGTM.

master/buildbot/process/properties.py
SpecLad: Good point; fixed.

master/buildbot/process/properties.py
SpecLad: The "r" stands for "rendered", if that's what you mean. Do you want me to change the name?

sa2ajj: No, it's not about the name, I do not quite understand the idea behind `yield iprops.render(self._function)`...

SpecLad: It's so that the function can be a renderable as well. So you can do stuff like this:

```python
@renderer
def slave_join(iprops):
    return ntpath.join if iprops.getProperty('os') == 'windows' else posixpath.join

subdir = Transform(slave_join, Property('builddir'), 'subdir')
```

sa2ajj: Thank you for the explanation.

sa2ajj:
I think this feature will find own users.  Let's see what Travis says.

sa2ajj:
Thank you :smile:

PR #1875, opened 2015-10-13
[nine] UI update of steps and buildsummary

ssavrim:
* Update ui of steps to display external urls in new browser tab
* Do same thing for ui of buildsummary


ssavrim:
The objective of this patch is to open external links into another browser tabs.
Currently when we click on an external link and then click on back the UI of bb9 lost its context

cbocta:
:+1:

tardyp:
:+1:

PR #1879, opened 2015-10-14
pass properties to buildset as got a change from PBChangeSource

blue119:
I found I am not able to get properties in canStartChange function when the builder is triggered by buildbot sendchange. And I saw it work when the builder triggered by "force build". So I look into the buildbot source and make the patch to let it can see the properties from sendchange in canStarChange.

### DIFF
--- a/master/buildbot/schedulers/basic.py
+++ b/master/buildbot/schedulers/basic.py
@@ -123,8 +123,13 @@ class BaseBasicScheduler(base.BaseScheduler):
             # unimportant changes
             if not important:
                 return defer.succeed(None)
+
+            properties = {}
+            if hasattr(change, 'properties'):
+                properties = change.properties
             # otherwise, we'll build it right away
             return self.addBuildsetForChanges(reason=self.reason,
+                                              properties=properties,
                                               changeids=[change.number])
 
         timer_name = self.getTimerNameForChange(change)
--- a/master/buildbot/schedulers/basic.py
+++ b/master/buildbot/schedulers/basic.py
@@ -125,9 +125,13 @@ class BaseBasicScheduler(base.BaseScheduler):
             if not important:
                 return defer.succeed(None)
 
+            properties = change.properties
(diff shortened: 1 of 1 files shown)


tardyp:
I think you got an interresting usecase.

The data model isn't meant to be used like this. This duplicates the properties data in the db.

I think the canStartBuild should use the data api to get the change properties as well.
Or caller to canStartBuild should pass a properties dict built like the in the same manner as the builder:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/build.py#L174

@djmitche what do you think?

blue119:
Hi @tardyp ,

I guess make the properties parameter to be Properties object is straightforward in caller before passing the properties to addBuildsetForChanges function. and I modified the first pull request patch at d28f1d2 that can pass travis check.

What do you think?

tardyp:
I agree it is straightfoward for your particular usecase, but I disagree this is the way to go for upstream.
We need to find the correct way to resolve your situation, as I describe in my first answer.

blue119:
got it. thank for your explanation.

sa2ajj:
@djmitche your input is awaited :D

djmitche:
Buildsets already have properties -- that's not new.   Not all buildsets refer to changes, yet they may still have properties.  In cases where the BS is as a result of a Change, it makes sense to copy the change's properties to the BuildSet.

However, per http://docs.buildbot.net/latest/manual/cfg-properties.html this should already be the case.  So maybe I'm missing something in the request here?

tardyp:
as per what is described in the doc, buildset properties is the "scheduler" properties. This document how the "build" properties are populated at build start. (the link to the source I pointed)

djmitche:
Ah, of course you're absolutely right.  The change properties get mixed into the build properties when the build starts, not via the buildset.

So I agree that, if you need information about the changes in `canStartBuild`, you should make a Data API request to get that information.

tardyp:
closing, as per discussion

PR #1880, opened 2015-10-15
Only use --force with "git submodules update" if supported

vadz:
Support for --force option was added only in [Git 1.7.6](https://github.com/git/git/blob/master/Documentation/RelNotes/1.7.6.txt#L112), so don't use it with earlier versions (such as 1.7.1 in CentOS 6.5).

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -130,6 +130,7 @@ class Git(Source):
         self.getDescription = getDescription
         self.config = config
         self.supportsBranch = True
+        self.supportsSubmoduleForce = True
         self.srcdir = 'source'
         self.origin = origin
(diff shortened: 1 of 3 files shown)


vadz:
Oops, and this simple change broke the tests because they expect `--force` in the output but hard code git version 1.7.5, which doesn't understand it. I think the best would be to simply increase 1.7.5 to 1.7.6, or is there some better way of fixing this?

vadz:
I've finally updated the tests to check the behaviour with both 1.7.5 and 1.7.6.

sa2ajj:
Looks good, thank you.

Could you please add a note into release notes? `master/docs/relnotes/index.rst`?

vadz:
Updated the release notes now.

sa2ajj:
Thank you.

PR #1884, opened 2015-10-19
Don't leak Github Target's builder status subscription

apilloud:
The github status target leaks builder subscriptions on reload. This can result in a huge multiplication on API calls to github, which is ratelimited at 5000 requests per hour. This change saves off a handle to every builder status, and unsubscribes on service stop.

### DIFF
--- a/master/buildbot/test/unit/test_status_github.py
+++ b/master/buildbot/test/unit/test_status_github.py
@@ -170,11 +170,28 @@ class TestGitHubStatus(unittest.TestCase, logging.LoggingMixin,
 
         self.status._status.subscribe.assert_called_with(self.status)
 
+    def test_stopService(self):
+        """
+        When stopped it will unsubscribe from all statuses.
+        """
+        builderstatus = FakeBuilderStatus()
+
+        result = self.status.builderAdded('builder-name', builderstatus)
+
+        self.status.stopService()
+
+        self.assertEqual(
+            0,
+            self.builderstatus.subscribe_count,
+            'There are still %d outstanding subscriptions'
+            % self.builderstatus.subscribe_count,
+        )
+
     def test_builderAdded(self):
         """
         Status is attached to every builder.
         """
-        result = self.status.builderAdded('builder-name', None)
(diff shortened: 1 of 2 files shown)


tardyp:
Did you test it on master branch?

I dont think this works like that in nine.

apilloud:
I didn't test this on master, only eight, but the same bug exists in both places. It looks like the mock interface the github target test implements is incomplete. I'll improve the existing test and add a new one that detects this leak.

tardyp:
@apilloud we would love to see this merged! will you have the time to finish it?

apilloud:
@tardyp Sorry, our buildbot deployment was on fire when I was working on this, so I didn't have time to write the tests. I then proceeded to forget about it. Rebased and added the test.

master/buildbot/test/unit/test_status_github.py
```diff
@@ -170,11 +170,28 @@ def test_startService(self):
 
         self.status._status.subscribe.assert_called_with(self.status)
 
+    def test_stopService(self):
+        """
+        When stopped it will unsubscribe from all statuses.
+        """
+        builderstatus = FakeBuilderStatus()
+
+        result = self.status.builderAdded('builder-name', builderstatus)
```
tardyp: wait what..

There is no builderStatus in buildbot nine. I guess this was a bug on eight?
github status is not yet ported to nine, so it needs to be ported first.

tardyp: we really need to port this. if you are volunteer that would be very cool
see the status, so that you can find example of ported code:
http://docs.buildbot.net/latest/manual/cfg-reporters.html#githubstatus-not-migrated

apilloud: The code in 'master/buildbot/status/builder.py' is just dead then? Our production and test environments are stuck on buildbot eight. Unfortunately there aren't any features that are compelling enough to push us to nine, and we've added a bunch of features on top of eight that don't port to nine, so moving isn't going to happen any time soon.

PR #1885, opened 2015-10-20
Added a missed instruction: c['schedulers'] = []

marco-buttu:
The example in the "Adding a 'try' scheduler" section did not work because c had not key 'schedulers':

$ buildbot reconfig master
2015-10-20 11:28:12+0200 [-] error while parsing config file:
      ...    
  exceptions.KeyError: 'schedulers'


jaredgrubb:
LGTM!

PR #1886, opened 2015-10-20
Fix issue with the table 'scheduler_masters'

delanne:
This table is populated by the SchedulersConnectorComponent::setSchedulerMaster, but deleted by the function MastersConnectorComponent::setMasterState even if the master is active.

If the master is active, the content of this table mustn't be deleted, otherwise schedulers aren't link to a master. As a consequence, when requesting
 - /api/v2/masters/<master_id>/schedulers/<sched_id>, the master field is always null. 
 - /api/v2/schedulers, the master field is always null.
- the Gui page about schedulers (http://127.0.0.1:8020/#/schedulers) the column Master is empty

### DIFF
--- a/master/buildbot/db/masters.py
+++ b/master/buildbot/db/masters.py
@@ -52,11 +52,12 @@ class MastersConnectorComponent(base.DBConnectorComponent):
                 return False  # can't change a row that doesn't exist..
             was_active = bool(rows[0].active)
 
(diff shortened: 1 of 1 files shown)


Stibbons:
I was wondering, maybe it is offtopic here, but someone may know the anwser: why does buildbot use SQLAlchemy Core and not ORM?

PR #1891, opened 2015-10-21
don't expose change_hook_dialects, as it may contain secrets

djmitche:


pieterlexis:
:+1: this fixes the issue.

PR #1892, opened 2015-10-24
Bug fixes for buildbot-data

tothandras:
- [x] fix the control function
- [x] log updates
- [x] replace `opened` with `data` where `dataService` is used


tardyp:
thanks! still need the breadcrumb fixes, but this is good start!

tothandras:
Okay, will do! I knew that I forgot something..

PR #1895, opened 2015-10-26
reload 'master.cfg' from thread when buildbot is reconfigured

delanne:
When buildbot is started, master.cfg is loaded from a thread (see 76c5b0ce6f64f7eeda5e590e1155fabece68812d).
When buildbot is reconfigured, master.cfg is loaded from the main thread. It should be loaded from a thread.


tardyp:
:+1: 
Actually https://github.com/buildbot/buildbot/commit/76c5b0ce6f64f7eeda5e590e1155fabece68812d was pretty useless, and I really meant this commit :)

PR #1897, opened 2015-10-26
fix for http://trac.buildbot.net/ticket/3369

tardyp:
If you change the slave configuration on the master.cfg, and do a buildbot reload then the slave list in the UI is not updated accordingly

This was because the db update algorithm was not properly maintaining the list of buildermasterids for a slave


djmitche:
lgtm

nam4dev:
+1

djmitche:
@nam4dev thanks for the reviews :heart:

PR #1898, opened 2015-10-27
Ticket3321

ewongbb:
This is a Work in progress and this PR exists only for Feedback/review.  Not for Merging (at least, not now).  This is against Eight.


tardyp:
I think this is good cleanup, but certainly must go nine first.

sa2ajj:
And tests fail...

ewongbb:
the tests were expected to fail..  I haven't finished the patch :)

djmitche:
I don't think this is worth doing against the eight branch.  If you want to continue this work, please do so against the nine branch in a new PR.  Thanks Edmund!

PR #1906, opened 2015-10-30
fixes for Webui

tardyp:
289d5b7 enhance the build page
d61b3bf remove verbose messages
0c0aed2 fix publicfields filter
4f891e3 build page: display slave without rawdata
4eb0860 builders page: order builds in reverse

### DIFF
--- a/www/base/src/app/builders/builders.tpl.jade
+++ b/www/base/src/app/builders/builders.tpl.jade
@@ -21,7 +21,7 @@
             a(ui-sref='builder({builder: builder.builderid})')
               | {{ builder.name }}
           td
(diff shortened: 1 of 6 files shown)


tothandras:
I like the verbose messages, but only in development. We could use [$logProvider.debugEnabled](https://docs.angularjs.org/api/ng/provider/$logProvider) option.

tardyp:
@tothandras those logs are really information that is available in the network tab (either rest and ws are easily debugable via the dev tools)

tothandras:
good point :)

tardyp:
if no objection, I merge this.

PR #1908, opened 2015-11-03
waterfall: fix  ReferenceError

ClemPi:
Currently waterfall is broken (http://nine.buildbot.net/#/waterfall) because of a ReferenceError.

This patch fix it.


tardyp:
:+1:

PR #1909, opened 2015-11-03
Fix workaround

tothandras:
@tardyp

### DIFF
--- a/www/base/src/app/builders/builds/build.controller.coffee
+++ b/www/base/src/app/builders/builds/build.controller.coffee
@@ -72,7 +72,7 @@ class Build extends Controller
         data = dataService.open($scope)
         data.getBuilders(builderid).then (builders) ->
             $scope.builder = builder = builders[0]
-            builder.getBuilds(number__lt: buildnumber + 2, limit: 3, order: '-number').then (builds) ->
+            builder.loadBuilds(number__lt: buildnumber + 2, limit: 3, order: '-number').then (builds) ->
                 $scope.prevbuild = null
                 $scope.nextbuild = null
                 build = null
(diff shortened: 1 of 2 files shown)


www/base/src/app/builders/builds/build.controller.coffee
```diff
@@ -72,7 +72,7 @@ class Build extends Controller
         data = dataService.open($scope)
         data.getBuilders(builderid).then (builders) ->
             $scope.builder = builder = builders[0]
-            builder.getBuilds(number__lt: buildnumber + 2, limit: 3, order: '-number').then (builds) ->
+            builder.loadBuilds(number__lt: buildnumber + 2, limit: 3, order: '-number').then (builds) ->
```
tardyp: we don't need builder.builds, so this is really a getBuilds

tardyp:
otherwise, looks good!

www/base/src/app/builders/builds/build.controller.coffee
tothandras: does the page switching works that way? this might be another bug, because the builder losts the reference to builds and can't close its connection

tardyp: mmh indeed, this does not look right, we should not rely on the fact that we store a reference of the the element to stop the listening.

I think each wrapped element should have a reference on the dataService.open handle, so that it can registers its sub registrations properly to be able to release them afterwards

tardyp: Note that I still see the deadlock sometimes, and it does not make sense for me if this is due to a subscribe leak.

tothandras:
I am closing this, because of #1911.

PR #1910, opened 2015-11-03
 improve tag filtering in builder page

tardyp:
when you have a lot of builders, its not handy to see what you want.

+tag means all tags with + must be present in the builder tags
-tag means no tags with - must be present in the builder tags
tag means at least one of the filtered tag might be present


www/data_module/src/specification.constant.coffee
```diff
@@ -73,7 +73,7 @@ class Specification extends Constant
                     'builds/n:number/steps/n:number/logs/i:slug/contents'
                     'builds/n:number/steps/n:number/logs/i:slug/raw'
                 ]
-                static: true
+                static: false
```
tardyp: @tothandras is this the correct way to make it always refresh?

PR #1911, opened 2015-11-03
Dataoptims

tardyp:
This is a attempt to fix the performance issues we currently have with data module.

It actually disables all the optimizations (tabex and indexeddb cache)
We need to sort out a more efficient way to implement this, but for now, its better without them..

long term solution analysis:
https://docs.google.com/document/d/1VRdY2lqyz0UnbD4kllkzXBDIGDDoxvVgmWjn6M3oK4A/edit?ts=563b583f


tardyp:
updated with disabling of the indexeddb.

tardyp:
lets merge this to fix the mainline..

tothandras:
:+1:

tardyp:
new version with unit test fixing, and use of the fake tabex (previous removal was putting lot of trash in LocalStorage)
I still find the build page too slow, but I can't profile the reason yet.

tardyp:
does not work. it actually request several time the same thing.

tardyp:
New version.. again.
based on the midterm solution.

I updated my travis prod with that, and every page looks good and fast.

tardyp:
if no objection, I'd like to merge it tonight to put http://nine.buildbot.net/ back online.

PR #1914, opened 2015-11-09
reactivate control api

tardyp:
It was broken with the midterm version of datamodule, thus rebuild/stop build feature was broken


tothandras:
:+1:

PR #1918, opened 2015-11-10
fix unnecessary escapes

sa2ajj:
some regular expressions are way too complicated.


tardyp:
looks good, and 100% coverage on the patch, so good to go.

PR #1923, opened 2015-11-16
py3 compatibility: fix for urllib

iblis17:


jaredgrubb:
Looks good to me!

iblis17:
But i have no idea about the failed test.
I am investigating on it ...

iblis17:
Hmm... I just try to make travis ci rebuild, then it seems that everything works fine.

PR #1924, opened 2015-11-25
Fix AttributeError: no attribute 'matchesAnyTag'

krf:
Regression caused by the backport of the tags support feature to eight.

This caused an AttributeError exception whenever the IRC command
'status' was performed.

Log (shortened):
2015-11-25 12:10:25+0100 [XmlStream,client] irc command status
...
".../buildbot/status/master.py", line 244, in getBuilderNames
            if bldr.matchesAnyTag(tags):
        exceptions.AttributeError: 'Builder' object has no attribute 'matchesAnyTag'

This fix makes the behavior equal to nine, using the 'getBuilder'
function.


tardyp:
:+1:

PR #1925, opened 2015-11-27
Don't leak slave passwords in the logs

pieterlexis:
By default, the latent Docker slaves leak the password of the slave when
the slave is created with `buildslave create-slave`. This commit adds 2
lines to the buildslave.tac file to scrub the password from the
environment, as is shown in the example buildslave.tac for docker
slaves.


tardyp:
:+1:

PR #1926, opened 2015-11-30
Add get() method to Properties

rdb:
The properties object as passed into a renderer function behaves a lot like a dictionary, so my first instinct has always been to call the get() method, which is ubiquitous on dict-like object in Python.  However, Properties does not provide this, instead providing the same functionality through "getProperty".

Adding a get() alias (similar to how there is already a has_key alias) will make it behave more like what people will expect of a dict-like object.

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -131,6 +131,7 @@ class Properties(util.ComparableMixin):
     def hasProperty(self, name):
         return name in self.properties
 
+    get = getProperty
     has_key = hasProperty
 
     def setProperty(self, name, value, source, runtime=False):


sa2ajj:
It looks like this simple change breaks test cases...

rdb:
Which test case are you referring to?  I looked through them and it didn't find any failures that weren't already occurring on the eight branch.

sa2ajj:
So do we have unrelated test failures here?

sa2ajj:
Considering that we use different versions of Twisted, it does not really seem to be so.

master/buildbot/process/properties.py
```diff
@@ -131,6 +131,7 @@ def getProperty(self, name, default=None):
     def hasProperty(self, name):
         return name in self.properties
 
+    get = getProperty
```
Stibbons: so you should also provide update()

sa2ajj:
No update for a long time.  Thanks for the PR, but it requires work.

PR #1928, opened 2015-12-03
Soft disconnect when 'build_wait_timeout==0'

behlendorf:
Insubstantiation of a latent build slave without cleanly disconnecting
can result in a ping failure if a new latent slave is immediately
substantiated.  To prevent this the slave should disconnect cleanly
before being insubstantiated.

  [Broker,2,52.25.93.92] ping finished: failure
  [Broker,2,52.25.93.92] slave ping failed; re-queueing the request

This is easily reproducible by running back-to-back builds.

Signed-off-by: Brian Behlendorf <behlendorf1@llnl.gov>

### DIFF
--- a/master/buildbot/buildslave/base.py
+++ b/master/buildbot/buildslave/base.py
@@ -773,7 +773,7 @@ class AbstractLatentBuildSlave(AbstractBuildSlave):
         self.building.remove(sb.builder_name)
         if not self.building:
             if self.build_wait_timeout == 0:
-                d = self.insubstantiate()
+                d = self._soft_disconnect()
                 # try starting builds for this slave after insubstantiating;
                 # this will cause the slave to re-substantiate immediately if
                 # there are pending build requests.


behlendorf:
I should mention that we're setting `build_wait_timeout==0` because we're using buildbot for testing and we require a clean reproducible environment for each run.  This change updates the `build_wait_timeout==0` behavior so calls the same _soft_disconnect() function which is used by the timer.  That works for me, but if there's a better solution that's fine too!

tardyp:
LGTM. this code would deserve more love with unit tests. :)

behlendorf:
@tardyp thanks for the review.  I completely agree this area of the code could benefit from better test coverage.  Related to this I've been using buildbot with EC2 to [test pull requests]( http://build.zfsonlinux.org) for a kernel module, it's proven to be a great way to expose problems.  Buildbot really doesn't like it when you regularly panic its build slaves.

djmitche:
See http://trac.buildbot.net/wiki/StalePullRequestPolicy regarding stalled pull requests

PR #1929, opened 2015-12-05
avoid setting attribute of undefined in getGroups

djmitche:


senyai:
Fixed the issue for me

tardyp:
cool, thanks

PR #1931, opened 2015-12-08
Bug 3387: fix tests on new Twisted versions

djmitche:


tardyp:
mmhh so new twisted have their response code in str?
sounds like a bug..

LGTM anyway

PR #1935, opened 2015-12-18
Source Base: add a 'do with retry' function

jaredgrubb:
There's a lot of duplicated 'retry' code across the source classes. I moved that logic into the Source base class and adjusted the derived classes to use it. (This is all preliminary to me making Git-fetch more retry-happy).

### DIFF
--- a/master/buildbot/steps/source/darcs.py
+++ b/master/buildbot/steps/source/darcs.py
@@ -174,32 +174,19 @@ class Darcs(Source):
         return d
 
     def _checkout(self):
+        @defer.inlineCallbacks
+        def action(tryCount, totalTries):
+            # Clean up after a prior attempt.
+            if tryCount:
+                yield self.runRmdir(self.workdir)
 
-        if self.retry:
-            abandonOnFailure = (self.retry[1] <= 0)
-        else:
-            abandonOnFailure = True
+            abandonOnFailure = (tryCount >= totalTries)
+            res = yield self._clone(abandonOnFailure)
 
-        d = self._clone(abandonOnFailure)
+            needRetry = res != 0
+            defer.returnValue((needRetry, res))
 
-        def _retry(res):
-            if self.stopped or res == 0:
-                return res
-            delay, repeats = self.retry
-            if repeats > 0:
-                log.msg("Checkout failed, trying %d more times after %d seconds"
-                        % (repeats, delay))
-                self.retry = (delay, repeats - 1)
-                df = defer.Deferred()
-                df.addCallback(lambda _: self.runRmdir(self.workdir))
-                df.addCallback(lambda _: self._checkout())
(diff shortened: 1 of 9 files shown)


master/buildbot/steps/source/darcs.py
```diff
@@ -174,32 +174,19 @@ def _clone(self, abandonOnFailure=False):
...
-        if self.retry:
-            d.addCallback(_retry)
-        return d
+            abandonOnFailure = (tryCount >= totalTries)
+            res = yield self._clone(abandonOnFailure)
+
+            needRetry = res != 0
+            defer.returnValue((needRetry, res))
+
+        return self.doWithRetry(action, 'checkout')
```
tardyp: I think this would make sense to make this a decorator of _checkout():

maybe this would make it a little bit too magical though

master/docs/relnotes/index.rst
```diff
@@ -34,6 +34,10 @@ Deprecations, Removals, and Non-Compatible Changes
 Changes for Developers
 ~~~~~~~~~~~~~~~~~~~~~~
 
+* :class:`~buildbot.steps.source.base.Source` now has a ``doWithRetry`` method to help make it easier
```
tardyp: why only for sources? wouldn't be better in a mixin or something that can apply everywhere?

tardyp:
LGTM anyway

master/docs/relnotes/index.rst
jaredgrubb: Yeh, that's a cool idea .. there's not much source-specific about it. Alternatives:
 - mixin
 - member variable (eg, self.retry = RetryHelper(retryParams)); this would let you have multiple "retry regimes"
 - move to Shell/BuildStep without utility.

What do you think?

master/buildbot/steps/source/darcs.py
jaredgrubb: @RetryHelper.wrapFunction(self.retry)
@RetryMixin.wrapFunction

It's an interesting idea! Do we have many other examples where we do that for somethign like this?

tardyp: util/debounce.py is a good example. It actually does very similar stuff excepts it debounces..

master/docs/relnotes/index.rst
tardyp: I would stay similar semantics as debounce

tardyp:
I would really love to see it in nine. Not sure if we want to make another eight release

jaredgrubb:
I tried to split it out as a utility class and had trouble because of the way it interacts with the "stop" feature ... I can get this going as-is for nine, but I'm not sure I'll have time to fully delve into making it a separate mixin.

sa2ajj:
why do we have failure?

if this feature is applicable to `nine`, the nine firest :/

PR #1938, opened 2015-12-23
add peer dependencies of dependent packages to dependencies

rutsky:
Starting from npm@3 peer dependencies are not installed by default.
I believe it's assumed, that if package A depends on package B that
have peer dependency C, A must install C in order to use B.

Discussion of that npm change can be found here: https://github.com/npm/npm/issues/5080

Not sure either `"*"` version of peer dependency should be specified in parent package or specific version.
I think `"*"` version will lead npm to use best matching version that satisfies dependent package.

To reproduce issue that this PR is fixing you can upgrade npm to version newer than 3.0 and try to do clean installation of `buildbot/www/*` packages (with removed `node_modules`).

Similar PR in guanlecoja: https://github.com/buildbot/guanlecoja/pull/3.

### DIFF
--- a/www/codeparameter/package.json
+++ b/www/codeparameter/package.json
@@ -5,6 +5,7 @@
         "npm": ">=1.4.0"
     },
     "devDependencies": {
-        "guanlecoja": "~0.4.0"
+        "guanlecoja": "~0.4.0",
+        "gulp": "*"
     }
 }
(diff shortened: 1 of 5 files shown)


tardyp:
I'd like to find a better option via guanlecoja trick

PR #1946, opened 2016-01-05
Allow user to launch a try build even when there are no changes in th…

tenuki:
Allow user to launch a try build even when there are no changes in the local repository.


tardyp:
thanks

PR #1948, opened 2016-01-08
Initial creation of logentries status plugin

jcftang:
This is a first attempt at creating a logentries status plugin.

This takes the build started and finished events and just sends them to logentries.

### DIFF
--- a/master/contrib/logentries.py
+++ b/master/contrib/logentries.py
@@ -1,0 +1,213 @@
+'''
+Logentries Status plugin for buildbot
+
+This plugin sends a subset of status updates to the Logentries service.
+
+Usage:
+
+    import logentries
+    c['status'].append(logentries.LogentriesStatusPush(api_token="da45d4be-e1e7-4de3-9861-45bdab564cb3", endpoint="data.logentries.com", port=10000))
+
+If you want SSL
+
+    import logentries
+    c['status'].append(logentries.LogentriesStatusPush(api_token="da45d4be-e1e7-4de3-9861-45bdab564cb3", endpoint="data.logentries.com", port=20000, tls=True))
+'''
+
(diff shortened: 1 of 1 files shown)


rutsky:
If you use Buildbot >= 0.8.11 you can convert your module to proper Python package and use it as an actual Buildbot plugin. See http://docs.buildbot.net/0.8.12/manual/plugins.html.

Your status receiver service is blocking, so your master will be blocked until all data is sent to the logentries server, which may break other Buildbot async services.
You need to use Twisted async API for network I/O to send data.

jcftang:
Ok, that sounds like I should packing it up as a python package for easier use/distribution.

I currently don't have any experience with twisted's API, so I won't likely be reworking any time soon. The blocking behaviour works for me as the volume of builds that I have is relatively low.

I'll close this PR for now.

jcftang:
Just for the record -- https://pypi.python.org/pypi/buildbot_status_logentries

I ended up making a python package and it now is a bit more twisted friendly in the way that it works

rutsky:
Thanks @jcftang!

I think your plugin may be added to the list of Buildbot plugins.
@djmitche do we have list of Buildbot plugins? :)
Perhaps create page in the Trac Wiki with list of them?
(I don't think including dynamically changed list in the Buildbot docs is a good idea.)

djmitche:
That's great!  Yes, let's add a wiki page, and point to it from the documentation.

djmitche:
http://trac.buildbot.net/ticket/3406

jcftang:
I've yet to split out the plugin into its own little repo, I'll probably split it out later today and put it somewhere on github.

djmitche:
http://trac.buildbot.net/wiki/Plugins

PR #1957, opened 2016-01-25
Fix: install default mail template

jhunold:


rutsky:
LGTM.

Related to http://trac.buildbot.net/ticket/3356.

sa2ajj:
What's interesting is that the ticket was already closed :/

sa2ajj:
@jhunold thanks for the PR! :)

rutsky:
@jhunold on the second though, could you use `include()` function to include all `*.txt` files from templates directory?

In `MANIFEST.in` all `*.txt` files are included:
https://github.com/buildbot/buildbot/blob/master/master/MANIFEST.in#L27

Example of how to include files by pattern few lines above of your code:
https://github.com/buildbot/buildbot/blob/master/master/setup.py#L202

PR #1964, opened 2016-01-28
data_module: remove old "dist" before copying new "dist" to be sure, that no old files are left.

rutsky:
@tardyp please review.


tardyp:
:+1:

PR #1967, opened 2016-01-30
Bug 3420: specify utf8 charset everywhere

djmitche:
This allows Buildbot to run in a DB created without 'DEFAULT CHARSET = utf8', as is the case for the default MySQL DB's.

I tried monkey-patching `sa.Table`, but SQLAlchemy is such deep Python magic that it just didn't work.

Instead, I just specified the charset everywhere, and then added some tests to make sure that it is actually set after every upgrade script.

### DIFF
--- a/master/buildbot/db/migrate/versions/001_initial.py
+++ b/master/buildbot/db/migrate/versions/001_initial.py
@@ -20,123 +20,154 @@ import sqlalchemy as sa
 from buildbot.db.migrate_utils import test_unicode
 from buildbot.util import json
 from buildbot.util import pickle
+from buildbot.util import sautils
 from twisted.persisted import styles
 
 metadata = sa.MetaData()
 
-last_access = sa.Table('last_access', metadata,
-                       sa.Column('who', sa.String(256), nullable=False),
-                       sa.Column('writing', sa.Integer, nullable=False),
-                       sa.Column('last_access', sa.Integer, nullable=False),
-                       )
+last_access = sautils.Table(
+    'last_access', metadata,
+    sa.Column('who', sa.String(256), nullable=False),
+    sa.Column('writing', sa.Integer, nullable=False),
+    sa.Column('last_access', sa.Integer, nullable=False),
+)
 
-changes_nextid = sa.Table('changes_nextid', metadata,
-                          sa.Column('next_changeid', sa.Integer),
-                          )
+changes_nextid = sautils.Table(
+    'changes_nextid', metadata,
+    sa.Column('next_changeid', sa.Integer),
+)
 
-changes = sa.Table('changes', metadata,
-                   sa.Column('changeid', sa.Integer, autoincrement=False, primary_key=True),
-                   sa.Column('author', sa.String(256), nullable=False),
-                   sa.Column('comments', sa.String(1024), nullable=False),
-                   sa.Column('is_dir', sa.SmallInteger, nullable=False),
-                   sa.Column('branch', sa.String(256)),
-                   sa.Column('revision', sa.String(256)),
-                   sa.Column('revlink', sa.String(256)),
-                   sa.Column('when_timestamp', sa.Integer, nullable=False),
-                   sa.Column('category', sa.String(256)),
-                   )
+changes = sautils.Table(
+    'changes', metadata,
+    sa.Column('changeid', sa.Integer, autoincrement=False, primary_key=True),
+    sa.Column('author', sa.String(256), nullable=False),
+    sa.Column('comments', sa.String(1024), nullable=False),
+    sa.Column('is_dir', sa.SmallInteger, nullable=False),
+    sa.Column('branch', sa.String(256)),
+    sa.Column('revision', sa.String(256)),
+    sa.Column('revlink', sa.String(256)),
+    sa.Column('when_timestamp', sa.Integer, nullable=False),
+    sa.Column('category', sa.String(256)),
+)
 
-change_links = sa.Table('change_links', metadata,
-                        sa.Column('changeid', sa.Integer, sa.ForeignKey('changes.changeid'), nullable=False),
-                        sa.Column('link', sa.String(1024), nullable=False),
-                        )
+change_links = sautils.Table(
+    'change_links', metadata,
+    sa.Column('changeid', sa.Integer, sa.ForeignKey(
+        'changes.changeid'), nullable=False),
+    sa.Column('link', sa.String(1024), nullable=False),
+)
 
-change_files = sa.Table('change_files', metadata,
-                        sa.Column('changeid', sa.Integer, sa.ForeignKey('changes.changeid'), nullable=False),
-                        sa.Column('filename', sa.String(1024), nullable=False),
-                        )
+change_files = sautils.Table(
+    'change_files', metadata,
+    sa.Column('changeid', sa.Integer, sa.ForeignKey(
+        'changes.changeid'), nullable=False),
+    sa.Column('filename', sa.String(1024), nullable=False),
+)
 
-change_properties = sa.Table('change_properties', metadata,
-                             sa.Column('changeid', sa.Integer, sa.ForeignKey('changes.changeid'), nullable=False),
-                             sa.Column('property_name', sa.String(256), nullable=False),
-                             sa.Column('property_value', sa.String(1024), nullable=False),
-                             )
+change_properties = sautils.Table(
+    'change_properties', metadata,
+    sa.Column('changeid', sa.Integer, sa.ForeignKey(
+        'changes.changeid'), nullable=False),
+    sa.Column('property_name', sa.String(256), nullable=False),
+    sa.Column('property_value', sa.String(1024), nullable=False),
+)
 
-schedulers = sa.Table("schedulers", metadata,
-                      sa.Column('schedulerid', sa.Integer, autoincrement=False, primary_key=True),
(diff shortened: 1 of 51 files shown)


master/docs/manual/cfg-schedulers.rst
```diff
@@ -808,39 +808,45 @@ What you need in your config file is something like::
...
-                 # property is the name of the parameter
-                 properties=[
-                    util.BooleanParameter(name="force_build_clean",
-                                          label="force a make clean",
-                                          default=False),
-                    util.StringParameter(name="pull_url",
-                                         label="optionally give a public Git pull url:<br>",
-                                         default="", size=80)
-                 ])
+        name="force",
```
rutsky: Is this part got in this PR accidentaly?

master/buildbot/db/migrate/versions/001_initial.py
```diff
@@ -28,10 +28,12 @@
                        sa.Column('who', sa.String(256), nullable=False),
                        sa.Column('writing', sa.Integer, nullable=False),
                        sa.Column('last_access', sa.Integer, nullable=False),
+                       mysql_character_set='utf8'
```
rutsky: I suggest to add trailing comma there and below.

rutsky:
@djmitche random guess, is specifying `charset=utf8` on connection string will help without these changes? As documented [here](http://docs.sqlalchemy.org/en/rel_1_0/dialects/mysql.html?highlight=mysql#mysql-unicode):

```
e = create_engine("mysql+pymysql://scott:tiger@localhost/test?charset=utf8")
```

master/buildbot/db/migrate/versions/001_initial.py
djmitche: Haha, OK, time to fire up sed :)

djmitche:
Buildbot already adds "charset=utf8" to the URL for MySQL.  But that affects SQLAlchemy's behavior and the client's behavior, not the server's behavior.

master/buildbot/db/migrate/versions/001_initial.py
djmitche: fixed

tardyp:
LGTM!

rutsky:
@djmitche, what will happen with tables that MySQL users already have?

For example if user uses MySQL with some old version of Buildbot (so that he has not latest Buildbot database version).
Then he upgrades Buildbot and runs migration scripts --- newly created tables will have set default charset to UTF-8, but old tables will be left intact.

Is this a problem to have different tables with different charset settings?
Maybe migration should be introduced that changes charset for all tables?
May change of charset break some functionality, may non-UTF-8 charset be expected and used somewhere?

djmitche:
Those are already charset=utf8 -- just those people had to make an extra effort to accomplish that.

rutsky:
It wasn't possible to use MySQL with Buildbot without (manually) settings default database charset to UTF-8?

master/buildbot/test/util/migration.py
```diff
@@ -82,5 +84,17 @@ def upgrade_thd(engine):
                 schema.runchange(version, change, 1)
         d.addCallback(lambda _: self.db.pool.do_with_engine(upgrade_thd))
 
+        def check_table_charsets(engine):
```
rutsky: I think it should be `check_table_charsets_thd` or something like this. It's being run in thread, right?

master/buildbot/test/util/migration.py
```diff
@@ -82,5 +84,17 @@ def upgrade_thd(engine):
...
+            # charsets are only a problem for MySQL
+            if engine.dialect.name != 'mysql':
+                return
+            dbs = [r[0] for r in engine.execute("show tables")]
+            for tbl in dbs:
+                r = engine.execute("show create table %s" % tbl)
+                create_table = r.fetchone()[1]
+                self.assertIn('DEFAULT CHARSET=utf8', create_table,
```
rutsky: What if non-Buildbot tables will be in the DB? Or it's unsupported configuration?
Maybe some external tools (phpMyAdmin?) will create additional tables, for example to add additional metadata?

rutsky:
`mysql_character_set='utf8'` everywhere looks ugly...
Maybe do `ALTER DATABASE CHARACTER SET utf8 COLLATE utf8_general_ci;` to change wrong default if current character set is not UTF8?

Though [ALTER DATABASE](https://dev.mysql.com/doc/refman/5.6/en/alter-database.html) requires additional ALTER privilege.
Also that change is unexpected if database is shared with other application.

rutsky:
If ALTER table is not an option and monkey-patching is too complex with project such as SQLAlchemy, we can define custom wrapper for `sa.Table`, e.g.:

```
def Table(*args, **kwargs):
    return sa.Table(*args, mysql_charset_set='utf8', **kwargs)
```

and use it instead of `sa.Table`.

djmitche:
> It wasn't possible to use MySQL with Buildbot without (manually) settings default database charset to UTF-8?

correct.

> If ALTER table is not an option and monkey-patching is too complex with project such as SQLAlchemy, we can define custom wrapper for sa.Table

I considered that, but then we've got to modify exactly the same files to import this Table replacement, and test that the replacement hasn't been missed somewhere -- that seems about the same level of ugly.  Do you feel that'd be sufficiently cleaner to warrant the change?

> What if non-Buildbot tables will be in the DB? Or it's unsupported configuration?
Maybe some external tools (phpMyAdmin?) will create additional tables, for example to add additional metadata?

Those won't be present when we're running tests -- the tests repeatedly delete all tables.

rutsky:
> > If ALTER table is not an option and monkey-patching is too complex with project such as SQLAlchemy, we can define custom wrapper for sa.Table

> I considered that, but then we've got to modify exactly the same files to import this Table replacement, and test that the replacement hasn't been missed somewhere -- that seems about the same level of ugly.  Do you feel that'd be sufficiently cleaner to warrant the change?

Use of Table wrapper requires changes of the same files, yes.
And I think probability to forget to use Table wrapper in new code is the same as probability to forget to include `mysql_charset_set='utf8'`.

What I'm actually worrying is:

1. It's much easier to make a typo in a new code with `mysql_charset_set='utf8'`, than with Table wrapper (e.g. I believe following will be silently eaten by SQLAlchemy: `_mysql_charset_set='utf8'`). 

2. Adding of other options in the future may be easier. E.g. what if we will need to also specify COLLATE in MySQL: `mysql_collate='utf8_general_ci'` (btw, do we need collation currently?)

> > What if non-Buildbot tables will be in the DB? Or it's unsupported configuration?
> > Maybe some external tools (phpMyAdmin?) will create additional tables, for example to add additional metadata?

> Those won't be present when we're running tests -- the tests repeatedly delete all tables.

Yes, right, I thought that check were done in migrations, and it actually in migrations tests.

djmitche:
That's a good point.  We have tests to ensure everything ends up in utf8, but it would be nice to not repeat this process for the next option.  I'll refactor.

djmitche:
K, fixed

master/buildbot/db/migrate/versions/001_initial.py
```diff
@@ -20,21 +20,22 @@
...
 from buildbot.util import json
 from buildbot.util import pickle
+from buildbot.util import sautils
 from twisted.persisted import styles
 
 metadata = sa.MetaData()
 
-last_access = sa.Table('last_access', metadata,
+last_access = sautils.Table('last_access', metadata,
                        sa.Column('who', sa.String(256), nullable=False),
```
rutsky: I think PEP8 will complain about invalid indentation of arguments here, it will expect:

```
last_access = sautils.Table('last_access', metadata,
                            sa.Column('who', sa.String(256), nullable=False),
```

Perhaps there is a way to format arguments without alignment with opening brace?

Something like:

```
last_access = sa.Table(
    'last_access', metadata,
    sa.Column('who', sa.String(256), nullable=False),
    ...
)
```

Alignment with opening brace is always bothered me with long list of arguments.

djmitche: Yeah, I was hoping to keep the diff minimal but I'll have to fix that.  More vimming :)

djmitche:
Hopefully that's better.

master/buildbot/util/sautils.py
```diff
@@ -50,3 +50,11 @@ def tryint(s):
                 return -1
         return tuple(map(tryint, sa.__version__.split('.')))
     return (0, 0, 0)  # "it's old"
+
+
+def Table(*args, **kwargs):
+    """Wrap table creation to add any necessary dialect-specific options"""
+    # work aorund the case where a database was created for us with
```
rutsky: typo: aorund

rutsky:
Small typo in a comment, beyond this LGTM.

It will conflict with bug2340 changes, but I will merge them.

djmitche:
OK, I fixed those two little bits.  I'm happy to wait for 2340 to land and merge this afterward, if you'd prefer -- otherwise, click the "merge pull requests" button :)

rutsky:
@sa2ajj, I expected to merge it before or after bug2340 depending on my investigations on other DB issues (since I'm the gatekeeper during these weeks of bug2340 merging).

sa2ajj:
@rutsky I'm really sorry :(  I did not see "do not merge" label and thought it was OK.  I'll refrain myself from merging stuff for the time being then.

PR #1970, opened 2016-01-31
validate.sh: don't clean docs builddir if built in quick mode

rutsky:
Docs rebuilt from scratch takes about 30 seconds on my machine, and only few seconds if only few files are being rebuilt (which is more appropriate for quick mode).


rutsky:
Building docs without cleaning builddir *may* miss invalid reference error, but this is acceptable IMO in quick mode.

sa2ajj:
LGTM

PR #1974, opened 2016-02-05
Checkartifact/halt on error

FLMFreddy:
Make sure we halt if we can't connect to halt if we encounter an error


master/buildbot/status/buildstep.py
```diff
@@ -375,12 +387,21 @@ def asDict(self):
...
-            self.build.builder.status.getURLForThing(l)]
+            self.build.builder.status.getURLForThing(l) + args]
                 for l in self.getLogs()]
+
+        result["urls"] = self.getURLs()
+
+
+        if request is not None:
+            from buildbot.status.web.base import path_to_step
+            result['url'] = path_to_step(request, self)
```
Stibbons: this is really weird

master/buildbot/status/buildstep.py
```diff
@@ -253,8 +261,11 @@ def logFinished(self, log):
         for w in self.watchers:
             w.logFinished(self.build, self, log)
 
-    def addURL(self, name, url):
+    def addURL(self, name, url, results=None):
```
Stibbons: this is weird to add this results arg... why is it needed?

master/buildbot/status/buildstep.py
FLMFreddy: Hi Stibbons, I actually created the PR by mistake, but do feel free to look it over

PR #1978, opened 2016-02-08
treat Row._next_id=None case as if next id should be 1

rutsky:
Otherwise first call to Row.nextId() returns None, which leads to failures of
some tests if they are being run first (before other tests initialized next id
to non-None value).


djmitche:
LGTM

PR #1966, opened 2016-01-30
Update ForceScheduler example to suit new config

djmitche:
Bug [3416](http://trac.buildbot.net/ticket/3416)

PR #1979, opened 2016-02-09
Fix forcescheduler page.

tardyp:
When the scheduler is not yet loaded, the forcefield directive will crash.

I don’t know exactly what triggered this issue, maybe an upgrade of angularjs changed the timing.

Solution is to load the directive only when the scheduler is present in scope

+ minor clean of use of deprecated data-module api

### DIFF
--- a/www/base/src/app/builders/buildrequest/forcedialog/forcedialog.controller.coffee
+++ b/www/base/src/app/builders/buildrequest/forcedialog/forcedialog.controller.coffee
@@ -1,6 +1,6 @@
 class forceDialog extends Controller
     constructor: ($scope, $state, modal, schedulerid, $rootScope, builderid, dataService) ->
-        dataService.getForceschedulers(schedulerid, subscribe: false).then (schedulers) ->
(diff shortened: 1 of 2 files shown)


tardyp:
if no objection, I think we should merge this, and publish a beta7.
This is for me a big issue, that I failed to properly prioritize :-}

rutsky:
I haven't testes these and #1980 changes yet, but if they are looks good to other --- let's merge them.

@bdbaddog, can you test that they are fixing issues that you observed?

If yes, lets release beta7 with them.

bdbaddog:
I'll try to look at this later today. On a plane a the moment. :)

rutsky:
@bdbaddog I think it would be enough to merge this PR and PR #1980 in a temporary branch, and then run `make frontend` in the Buildbot root (IMO it's easier for one-time check than `gulp dev` development setup).

Also, Bill, can you try #1966 in that temporary branch?

PR  #1966 depends on #1979 and #1980. And these three PR should fix following issues:
http://trac.buildbot.net/ticket/3416
http://trac.buildbot.net/ticket/3418
http://trac.buildbot.net/ticket/3422

PR #1981, opened 2016-02-09
rework database creation, cleanup + other db fixes

rutsky:

PR #1982, opened 2016-02-10
run tests in Travis using file-based SQLite

rutsky:


rutsky:
At this moment following test should fail:
`buildbot.test.unit.test_db_users.TestUsersConnectorComponent.test_updateUser_race`
This is due to the fact that multiple connections doesn't work with file-based SQLite by default (DB file is locked).

If other failures will occur --- PR #1981 should fix them, since I don't see other errors when changes from that PR are applied.

rutsky:
Yep, tests are failing. Pushed fix for `buildbot.test.unit.test_db_users.TestUsersConnectorComponent.test_updateUser_race`.

rutsky:
This PR is integrated in PR #1981

PR #1983, opened 2016-02-10
add Travis configuration for PostgreSQL + fix PostreSQL issues on psycopg driver

rutsky:


master/buildbot/test/fake/fakedb.py
```diff
@@ -90,6 +91,10 @@ def __init__(self, **kwargs):
         for k, v in iteritems(self.values):
```
rutsky: I don't like this part. If all strings should be unicode in the interface IMO it's better to add assert that they are, e.g.:

```
for k, v in iteritems(self.values):
    assert not isinstance(v, str), "All strings passed to DB layer must be unicode "
```

and fix tests to pass only unicode strings.

tardyp: well, as this is "only" test code, I dont think we should bother be annoying on that, I prefer we do like this. did you try to remove that code? I would be you have a few thousands lines to update..

master/buildbot/test/fake/fakedb.py
```diff
@@ -90,6 +91,10 @@ def __init__(self, **kwargs):
...
             if isinstance(v, str):
                 self.values[k] = unicode(v)
+        # encode columns that should be stored utf-8 encoded (e.g. LargeBinary
+        # columns) to utf-8
+        for col in self.utf8_columns:
+            self.values[col] = self.values[col].encode("utf-8")
```
tardyp: so the purpose of this is to raise in case this is not utf8 encodable?

master/buildbot/test/fake/fakedb.py
```diff
@@ -528,6 +533,7 @@ class LogChunk(Row):
         compressed=0)
 
     required_columns = ('logid', )
+    utf8_columns = ('content', )
```
tardyp: I am not sure exactly what this is about. fakedb is about faking the db. I dont see how this code will help fixing pg

master/buildbot/test/fake/fakedb.py
rutsky: No, this is emulation of Binary columns --- they are read and written as bytestrings.

master/buildbot/test/unit/test_db_migrate_versions_024_add_buildslaves_table.py
```diff
@@ -42,9 +42,15 @@ def verify_thd(conn):
             self.assertEqual(res.fetchall(), [])
 
             # and buildslave name is unique, so we'll get an error here
-            dialect = conn.dialect.name
-            exc = (sa.exc.ProgrammingError if dialect == 'postgresql'
-                   else sa.exc.IntegrityError)
+            if conn.dialect.name == 'postgresql':
```
tardyp: could we rather say:

```
exc = sa.exc.IntegrityError
if  conn.dialect.driver == 'pg8000':
   exc = sa.exc.ProgrammingError
```

master/buildbot/test/fake/fakedb.py
rutsky: FakeDB emulates models tables.
In real model's tables Binary fields are read and written only as bytestrings.
When unicode is passed from the outside into DB layer's Binary field writer it's being handled in the following way:
1) if driver is pg8000: value is encoded back by SQLAlchemy to bytestring, effectively as `str(value)` (which of course fails if `value` is not ascii)
2) if driver is psycopg2: passing unicode value into Binary column is an error.

IMO behaviour of psycopg2 is correct and it was luck that in pg8000 that code worked.
This PR fixes raised error when psycopg2 driver is used.

tardyp: Thanks it is clear now.

I think it is worth to put this explaination as a comment in the code.

master/buildbot/test/unit/test_db_migrate_versions_024_add_buildslaves_table.py
rutsky: It depends on answer to the following question: if unique constraint is being broken, is it more like ProgrammingError, or IntegrityError?

If it is semantically IntegrityError, then by default IntegrityError should be used, and special case pg8000 should be added (as you suggested).

master/buildbot/test/fake/fakedb.py
rutsky: Yes, comment is a good idea.
Do you think it's better to place it here, or in `Raw` class near `Raw.utf8_columns` definition?

master/buildbot/test/fake/fakedb.py
rutsky: I'll try and see is it worth removing: if there will thousands of places to changes, it's not worth it, but if almost everywhere unicode string are already passed, then it may worth it.

master/buildbot/test/unit/test_db_migrate_versions_024_add_buildslaves_table.py
rutsky: I believe constraint violation is [IntegrityError](https://www.python.org/dev/peps/pep-0249/#integrityerror):

> Exception raised when the relational integrity of the database is affected, e.g. a foreign key check fails.

[ProgrammingError](https://www.python.org/dev/peps/pep-0249/#programmingerror):

> Exception raised for programming errors, e.g. table not found or already exists, syntax error in the SQL statement, wrong number of parameters specified, etc.

master/buildbot/test/fake/fakedb.py
rutsky: > FAILED (skips=15, failures=2, errors=764, successes=3382)

It's not worth it atm :)

rutsky:
@tardyp I fixed your comments. Please take a look.

PR #1989, opened 2016-02-13
log SQL queries using Python's logging interface + add default logger

rutsky:
Now all unhandled Python log events are printed into `stderr`, instead of generating "No handlers could be found for logger" message.

### DIFF
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -79,7 +79,7 @@ class BuildbotEngineStrategy(strategies.ThreadLocalEngineStrategy):
         if u.database:
 
             # Use NullPool instead of the sqlalchemy-0.6.8-default
-            # SingletonThreadpool for sqlite to suppress the error in
+            # SingletonThreadPool for sqlite to suppress the error in
             # http://groups.google.com/group/sqlalchemy/msg/f8482e4721a89589,
             # which also explains that NullPool is the new default in
             # sqlalchemy 0.7 for non-memory SQLite databases.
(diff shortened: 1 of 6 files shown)


djmitche:
Looks good

rutsky:
@djmitche I've added destroying connection pool on the thread. This should fix warnings (and should not break other tests :) ).

djmitche:
Nice work!

rutsky:
Ugh, tests are hanging now...

rutsky:
@djmitche I decided to stick with simple solution: disable SQLite check for using DB from multiple threads.
Destroying pool from the thread doesn't work for some reason; for example in `buildbot.test.integration.test_upgrade.TestWeirdChanges.testUpgradeChangeNoRevision` test.

master/buildbot/db/enginestrategy.py
```diff
@@ -89,6 +89,16 @@ def special_case_sqlite(self, u, kwargs):
             if not os.path.isabs(u.database[0]):
                 u.database = os.path.join(kwargs['basedir'], u.database)
 
+        else:
+            # For in-memory database SQLAlchemy will SingletonThreadPool and
```
djmitche: "..will use SingletonThreadPool.."

djmitche:
There's a lot of black magic in the DB API!

LGTM.

rutsky:
@djmitche can you take a look for the last time? :)

rutsky:
@djmitche,  I decided to rework log catching enabling/disabling.
I'm trying to debug some stuff and noticed, that current method is not very convenient.

master/buildbot/db/enginestrategy.py
```diff
@@ -89,6 +89,16 @@ def special_case_sqlite(self, u, kwargs):
             if not os.path.isabs(u.database[0]):
                 u.database = os.path.join(kwargs['basedir'], u.database)
```
sa2ajj: [nit] do you really need to keep this empty line here?

sa2ajj:
LGTM

PR #1992, opened 2016-02-14
Remove remaining sa.Table invocations

djmitche:
With a check in validate to be sure!


tardyp:
hopefully there are no from sa import Table

:+1:

djmitche:
I don't see any

PR #1994, opened 2016-02-14
install service_identity in Travis to silence Twisted's warnings

rutsky:


tardyp:
I think we should rather depend on this module

Le dim. 14 févr. 2016 à 19:16, Vladimir Rutsky <notifications@github.com> a
écrit :

rutsky:
@tardyp in what cases it will be useful for Buildbot?

Does Buildbot support serving content over HTTPS?
Buildbot may download stuff over HTTPS in user steps (and may not).

AFAIK PB connection doesn't involve any cryptography.

tardyp:
Well buildbot has a few steps to work with distant http server.
http://docs.buildbot.net/latest/manual/cfg-buildsteps.html#http-requests

rutsky:
> Well buildbot has a few steps to work with distant http server.
> http://docs.buildbot.net/latest/manual/cfg-buildsteps.html#http-requests

`HTTPStep` explicitly states in docs, that it requires `txrequests` and `requests` Python modules to work, we can put `service_identity` dependency there for them (if `HTTPStep` uses it).

In code I see that only `IrcStatusBot` directly tries to import `twisted.internet.ssl` module.

And warning is produced when `twisted.web.client` module is imported.

rutsky:
From meeting in IRC:

```
[20:50] <rutsky> In short: Twisted complains about missing service_identity - should we add new dep to BB or should we ask end user to do so?
[20:50] <sa2ajj> the latter
[20:50] <djmitche> we do this for a number of other tools, yes
[20:51] <djmitche> we should have it installed for tests, and we should include something in requirements.rst about it
[20:51] <sa2ajj> rationale: more dependencies, more likely things might go wrong for a novice
[20:51] <djmitche> right
[20:51] <djmitche> and more for packagers
[20:51] <sa2ajj> indeed
[20:51] <rutsky> tardyp: do you agree with djmitche, sa2ajj?
[20:51] <sa2ajj> since i do not create packages for quite some time, i forgot about my sorrows :)
```

rutsky:
According to discussion in IRC I need to document this issue too (add note to the docs about how to fix this issue if it occurs).

rutsky:
@tardyp, @djmitche, @sa2ajj I decided to replace manual `service_identity` installation by adding `tls` extras to Buildbot: it adds dependency to Twisted's `tls` extras and should install all required by Twisted packages by itself.

PR #1995, opened 2016-02-15
Improve format of configuration file syntax errors

adityadivekar03:
Change the format of configuration file syntax errors to make them more
user friendly.

fixes [#3437](http://trac.buildbot.net/ticket/3437)

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -19,6 +19,7 @@ import os
 import re
 import sys
 import warnings
+import traceback
 
 from buildbot import interfaces
 from buildbot import locks
@@ -198,6 +199,11 @@ class MasterConfig(util.ComparableMixin):
             except ConfigErrors as e:
                 for err in e.errors:
                     error(err)
+                raise errors
+            except SyntaxError:
+                error("encountered a SyntaxError while parsing config file:\n%s "%
+                    (traceback.format_exc(),),
+                    )    
                 raise errors
             except Exception:
                 log.err(failure.Failure(), 'error while parsing config file:')
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -203,7 +203,7 @@ class MasterConfig(util.ComparableMixin):
             except SyntaxError:
                 error("encountered a SyntaxError while parsing config file:\n%s "%
                     (traceback.format_exc(),),
-                    )    
+                    )
                 raise errors
             except Exception:
                 log.err(failure.Failure(), 'error while parsing config file:')
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -18,8 +18,8 @@ from future.utils import itervalues
 import os
 import re
 import sys
-import warnings
 import traceback
+import warnings
 
 from buildbot import interfaces
 from buildbot import locks
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -201,7 +201,7 @@ class MasterConfig(util.ComparableMixin):
                     error(err)
                 raise errors
             except SyntaxError:
-                error("encountered a SyntaxError while parsing config file:\n%s "%
+                error("encountered a SyntaxError while parsing config file:\n%s " %
                     (traceback.format_exc(),),
                     )
                 raise errors
--- a/master/buildbot/test/unit/test_config.py
+++ b/master/buildbot/test/unit/test_config.py
@@ -237,7 +237,7 @@ class MasterConfig(ConfigErrorsMixin, dirs.DirsMixin, unittest.TestCase):
     def test_loadConfig_parse_error(self):
         self.install_config_file('def x:\nbar')
         self.assertRaisesConfigError(
-            re.compile("error while parsing.*traceback in logfile"),
+            re.compile("encountered a SyntaxError while parsing config file:"),
             lambda: config.MasterConfig.loadConfig(
                 self.basedir, self.filename))
         self.assertEqual(len(self.flushLoggedErrors(SyntaxError)), 1)
--- a/master/buildbot/test/unit/test_config.py
+++ b/master/buildbot/test/unit/test_config.py
@@ -240,7 +240,6 @@ class MasterConfig(ConfigErrorsMixin, dirs.DirsMixin, unittest.TestCase):
             re.compile("encountered a SyntaxError while parsing config file:"),
             lambda: config.MasterConfig.loadConfig(
                 self.basedir, self.filename))
-        self.assertEqual(len(self.flushLoggedErrors(SyntaxError)), 1)
 
     def test_loadConfig_eval_ConfigError(self):
         self.install_config_file("""\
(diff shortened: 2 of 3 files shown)


master/buildbot/config.py
```diff
@@ -201,8 +202,9 @@ def loadConfig(cls, basedir, filename):
                 raise errors
             except Exception:
                 log.err(failure.Failure(), 'error while parsing config file:')
-                error("error while parsing config file: %s (traceback in logfile)" %
-                      (sys.exc_info()[1],),
+                error("error while parsing config file"
```
rutsky: space is missed here, message will be like `... file(full ...`

adityadivekar03:
Sorry for the mistake. I've inserted the space now.

rutsky:
Thanks!

LGTM.

master/buildbot/config.py
```diff
@@ -201,8 +202,9 @@ def loadConfig(cls, basedir, filename):
                 raise errors
             except Exception:
                 log.err(failure.Failure(), 'error while parsing config file:')
-                error("error while parsing config file: %s (traceback in logfile)" %
-                      (sys.exc_info()[1],),
+                error("error while parsing config file "
+                      "(full traceback in logfile):\n%s" %
```
tardyp: I agree parse error should be in the stderr.

but it looks like now the full traceback is not only in the logfile

so I would remove (full traceback in logfile)

rutsky:
> so I would remove (full traceback in logfile)

Then there will be no traceback in the logfile, which is bad IMO, log must be self-sufficient.

Log is duplicated in the user console when user does `buildbot start ./master`, because `start` or `reconfig` commands follows and outputs log too, AFAIK.
But as I saw they not really duplicated because of Python/Twisted issues/misconfiguration:
traceback that goes into logfile looks like doesn't have frames from inside `eval` session, and traceback that is formatted by `traceback.format_exc()` doesn't include frames outside `eval`.

Here is a log that I saw with changes from this PR:

```
2016-02-02 12:02:54+0300 [-] Loading configuration from '/home/bob/stuff/buildbot/master/master.cfg'
2016-02-02 12:02:54+0300 [-] error while parsing config file:
	  File "/home/bob/stuff/buildbot/env/local/lib/python2.7/site-packages/twisted/python/threadpool.py", line 246, in inContext
	    result = inContext.theWork()
	  File "/home/bob/stuff/buildbot/env/local/lib/python2.7/site-packages/twisted/python/threadpool.py", line 262, in <lambda>
	  File "/home/bob/stuff/buildbot/env/local/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
	  File "/home/bob/stuff/buildbot/env/local/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
	    return func(*args,**kw)
	  File "/home/bob/stuff/buildbot/buildbot/master/buildbot/config.py", line 204, in loadConfig
	    exec(f, localDict)
	exceptions.SyntaxError: invalid syntax (master.cfg, line 4)
	
2016-02-02 12:02:54+0300 [-] Configuration Errors:
2016-02-02 12:02:54+0300 [-]   error while parsing config file (full traceback in logfile):
	  File "/home/bob/stuff/buildbot/buildbot/master/buildbot/config.py", line 204, in loadConfig
	    exec(f, localDict)
	  File "/home/bob/stuff/buildbot/master/master.cfg", line 4
	    c = BuildmasterConfig = {}}
	                              ^
	SyntaxError: invalid syntax
```

@adityadivekar03, can you investigate how to get complete in terms of stack frames and user-friendly looking stack trace?

adityadivekar03:
@rutsky , If I'm not wrong I should look into how to include the frames from inside eval into the logfile so as to make the logfile complete?

rutsky:
> @rutsky , If I'm not wrong I should look into how to include the frames from inside eval into the logfile so as to make the logfile complete?

Yes. This is not obligatory IMO, but if this issue can be easily solved either by adding flags to `traceback.format_exc()` or by not simply reraising exception, but doing something more advanced (e.g. by using Twisted's Failure), that would be nice to have.

adityadivekar03:
Could you tell me how can I check this issue locally, so that I can check my fix locally first?

rutsky:
Sure! You need to create master configuration and add an error to it.
Follow [this tutorial](http://trac.buildbot.net/wiki/RunningBuildbotWithVirtualEnv) until master start part:

```
buildbot create-master master-dir
mv master-dir/master.cfg.sample master-dir/master.cfg 
# add error to master.cfg
echo "XXX" >> master-dir/master.cfg
buildbot start ./master-dir
# take a look at traceback
```

adityadivekar03:
I tried many approaches using the functions from the Failure class from Twisted, but unfortunately couldn't come to a fix. 
I am not able to figure out how to include the line that's being executed, in the logged file.  
Any hints if possible?

rutsky:
I think Twisted simply doesn't handle `SyntaxError` in the same way, as Python does.
Python error handler treats `SyntaxError` in a [special way](https://docs.python.org/2.7/library/traceback.html?highlight=syntaxerror#traceback.print_exception):

>  if etype is SyntaxError and value has the appropriate format, it prints the line where the syntax error occurred with a caret indicating the approximate position of the error.

Lets handle `SyntaxError` in separate `except SyntaxError:` clause and output `traceback.format_exc()` there (and no changes in `except Exception` clause at all).

adityadivekar03:
I think the tests are failing because the tests are configured to assert against a predefined handling involving the earlier Exception clause, but according to the commit the syntax errors are raised separately, and hence the message displayed has changed. 
The log in my system on adding a syntax error in the master.cfg file is as follows -
<pre><code>
2016-02-20 17:17:27+0530 [-] Log opened.
2016-02-20 17:17:27+0530 [-] twistd 15.5.0 (/home/hp/buildbot-work/sandbox/bin/python 2.7.6) starting up.
2016-02-20 17:17:27+0530 [-] reactor class: twisted.internet.epollreactor.EPollReactor.
2016-02-20 17:17:27+0530 [-] Starting BuildMaster -- buildbot.version: 0.9.0b7-8-gc4423de
2016-02-20 17:17:27+0530 [-] Loading configuration from '/home/hp/buildbot-work/sandbox/master-dir/master.cfg'
2016-02-20 17:17:27+0530 [-] Configuration Errors:
2016-02-20 17:17:27+0530 [-]   encountered a SyntaxError while parsing config file:
	  File "/home/hp/buildbot-work/src/master/buildbot/config.py", line 198, in loadConfig
	    exec(f, localDict)
	  File "/home/hp/buildbot-work/sandbox/master-dir/master.cfg", line 11
	    c = BuildmasterConfig = {}}
	                              ^
	SyntaxError: invalid syntax
	 
2016-02-20 17:17:27+0530 [-] Halting master.
2016-02-20 17:17:27+0530 [-] BuildMaster is stopped
2016-02-20 17:17:27+0530 [-] Main loop terminated.
2016-02-20 17:17:27+0530 [-] Server Shut Down.
</code></pre>

rutsky:
@adityadivekar03 
>  Should I look into how the test cases are to be changed?

Tests can be changed if you reasonably change behavior. So please fix them.

Your log looks great! What do you in the console?

adityadivekar03:
The same log as above is displayed in the terminal. 
There is no duplication of logs as was the case in the earlier code. 
Also I think, the error displayed in terminal is apt now since it only displays those calls to files that the user can verify, i.e. only the concerned frames are displayed now. 
I'll start working on the test cases then. Thanks!

adityadivekar03:
@rutsky the tests are working fine now. Removed the flushLoggedErrors function call since we are not adding the SyntaxError to the list of errors, but simply creating a log event when it is encountered. The list of errors is used only for adding ConfigErrors and hence we do not add the SyntaxError to the list.

rutsky:
Thanks, @adityadivekar03, your changes LGTM!

adityadivekar03:
Thanks a lot for your help!

PR #1996, opened 2016-02-16
improve nine upgrade doc

tardyp:
Note that the '~' in py:class is removing the full path from the output text

As the classes have changed module path, its important to show them.


rutsky:
LGTM

PR #1998, opened 2016-02-16
make it possible to have waitForFinish callable

sa2ajj:
I encountered a situation when `waitForFinish` is to depend on values of particular properties.

This is a way how I think it could be addressed.

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -209,9 +209,14 @@ class Trigger(BuildStep):
         results = SUCCESS
         self.running = True
 
+        if isinstance(self.waitForFinish, bool):
+            waitForFinish = self.waitForFinish
+        else:
+            waitForFinish = yield self.waitForFinish(self)
+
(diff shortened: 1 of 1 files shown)


tardyp:
I think that waitForFinish is already renderable, if its not, it should rather than making it optionally callable

sa2ajj:
It's not about renderable, but callable

sa2ajj:
What I'd like to see is a possibility to decide whether to wait for the triggered builds to finish based on other conditions.  So I modelled it after `doStepIf`

tardyp:
With util.renderer, you can easily make a renderable from a callable:
http://docs.buildbot.net/latest/manual/cfg-properties.html#renderer
So callables is a subset of renderables modulo a decorator.
see #2005

Le ven. 19 févr. 2016 à 18:55, Mikhail Sobolev <notifications@github.com> a
écrit :

tardyp:
closed in favor of #2005 and cleaning up the huge list of PRs

PR #2001, opened 2016-02-18
#3292 - allow gitpoller to view changes in merge commits

yhager:


tardyp:
Hi thanks for your patch.
Please create a unit test for your change as well as a release note entry

tardyp:
Also we would prefer to have the PR done on matter branch first

yhagerjs:
I'm not using nine yet, but I'll try to see if it applies there first, and I will look into creating a test. Thanks.

yhager:
I'm not using nine yet, but I will check to see if it applies there. I will also look into writing a unit test. Thanks

sa2ajj:
@yhager any updates?

yhager:
Honestly, I'm not sure this is the right approach. I'm sure that without my patch gitpoller will miss files that were changed in a merge commit. I'm also sure that with the patch it doesn't miss those. However, it seems to me like I'm seeing more file in the change list than the actual list of changed files.

When using the very common pattern of issue branches, every time master is merged to the issue branch and then that issue branch back to master *all* the files changed in both merges are listed, which is too much IMO.

djmitche:
Please note that we'll need to get a version of this merged to the master branch before we can merge one to the eight branch -- we aren't adding functionality to eight that is not present in master.

PR #2002, opened 2016-02-18
Large build page

tardyp:
I got feedback from several people that the build properties details are not very important for most of the users, and they are using a lot of horizontal space in the current UI.

Here is a proposal on the build page that we could use in order to fix this issue:

The idea is simply to put the build steps as another tab, and make the tabcontainer a full width component. build properties are now one more click away, but you get a full vision on your build, and more spaces for build step names and statuses.

before:
image

after:
Build page:
image
Child Buildpage (with parent label)
image

Expand buttons:
image
image
image

Buildrequest page:
image

debug page:
image

### DIFF
--- a/www/base/src/app/builders/buildrequest/buildrequest.controller.coffee
+++ b/www/base/src/app/builders/buildrequest/buildrequest.controller.coffee
@@ -6,6 +6,11 @@ class Buildrequest extends Controller
                 findBuilds $scope,
                     $scope.buildrequest.buildrequestid,
                     $stateParams.redirect_to_build
+                # when a build is discovered, force the tab to go to that build
+                savedNew = $scope.builds.onNew
+                $scope.builds.onNew = (build) ->
+                    build.active = true
+                    savedNew(build)
 
         doCancel = ->
             $scope.is_cancelling = true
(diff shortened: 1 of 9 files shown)


sa2ajj:
LGTM

Re properties: I have implemented filtering properties for Gerrit -- some of them are really just using the space.

bdbaddog:
I like the change. Tabs for all the info so it's not so crowded.

tardyp:
I spent some more time working on this concept.
I added more screenshots demonstrating it

details as per commit message:
    Enhance build and buildrequests pages:

    - Now everything is in the tabs, so that the UI is less crowded, and the steps are display full width
    - rework display of the parent build to display it as a label in the right, colored according to the parent build status
    - now that we have more space, we can rework step expension so that it is more intuitive

    Enhance the buildrequest page so that it looks more like the build page
    - factorize properties display, so that the template can be shared

    Add Debug page to display raw debug info

@rutsky this may clash a little bit with the worker rework

jaredgrubb:
I like it!

tardyp:
lets merge bug2340 first, and I'll rebase

tardyp:
rebased over the latest master

nam4dev:
:+1:

rutsky:
LGTM

PR #2003, opened 2016-02-19
Install docs dependencies in setup.py

### LINKED ISSUE #3435: Documentation for Pushover reporter says "sends email"

adityadivekar03:
Remove the pip install command from the docs Makefile used for installing the
sphinx distribution. 

fixes #3435

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,7 +128,11 @@ install:
   # astroid > 1.4 is incompatible with pylint 1.1.0
   - "[ $TESTS != lint ] || pip install pylint==1.1.0 pep8==1.5.7 flake8  astroid==1.3.8"
   # TODO: Pin Sphinx version to workaround http://trac.buildbot.net/ticket/3408
-  - "[ $TESTS != docs ] || pip install sphinx==1.3.3"
+  - |
+      [ $TESTS != docs ] || \
+      pip install sphinx==1.3.3 \
+                  sphinxcontrib-blockdiag \
+                  docutils>=0.8 \
 
 before_script:
   # create real MySQL database for tests
--- a/master/docs/Makefile
+++ b/master/docs/Makefile
@@ -10,8 +10,6 @@ TAR_TRANSFORM := $(if $(filter bsdtar,$(TAR_VERSION)),-s /^html/$(VERSION)/,--tr
 docs.tgz: clean html singlehtml images-png
 	sed -e 's!href="index.html#!href="#!g' < _build/singlehtml/index.html > _build/html/full.html
 	tar -C _build $(TAR_TRANSFORM) -zcf $@ html
-pipdeps:
-	pip install sphinxcontrib-blockdiag 'docutils>=0.8'
 
 images:
 	$(MAKE) -C images all
--- a/master/setup.py
+++ b/master/setup.py
@@ -426,9 +426,16 @@ else:
             'buildbot-www',
             'buildbot-slave'
         ],
+<<<<<<< HEAD
         'tls': [
             'Twisted[tls] ' + twisted_ver,
         ],
+=======
+        'docs': [
+            'sphinxcontrib-blockdiag',
+            'docutils>=0.8',
+        ]
+>>>>>>> Install docs dependencies in setup.py
     }
 
     if os.getenv('NO_INSTALL_REQS'):
--- a/master/docs/conf.py
+++ b/master/docs/conf.py
@@ -21,6 +21,17 @@ import textwrap
 sys.path.append(os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
+# Check if sphinx is installed
+try:
+    import sphinxcontrib.blockdiag
+    import pkg_resources
+    pkg_resources.require('docutils>=0.8')
+except DistributionNotFound:
+    print("Please run this command to install the dependencies first "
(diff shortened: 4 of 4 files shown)


tardyp:
I am the culprit for that pip install.

We need to also update the metabuildbot, and maybe other documentation automation (I am not sure if there is an ansible script to build the doc)

so that they use pip install buildbot[docs] (isn't it even possible to run pip install -e . with optional deps)

master/setup.py
```diff
@@ -418,6 +418,9 @@ def define_plugin_entries(groups):
         'bundle': [
             'buildbot-www',
             'buildbot-slave'
+        ],
+        'docs': [
+            'sphinxcontrib-blockdiag>=0.8',
```
tardyp: this is not the same deps.
old version said 

sphinxcontrib-blockdiag (implicit latest)
'docutils>=0.8'  forces upgrade of docutils

tardyp:
travis fails

master/docs/Makefile
```diff
@@ -10,8 +10,6 @@ TAR_TRANSFORM := $(if $(filter bsdtar,$(TAR_VERSION)),-s /^html/$(VERSION)/,--tr
 docs.tgz: clean html singlehtml images-png
 	sed -e 's!href="index.html#!href="#!g' < _build/singlehtml/index.html > _build/html/full.html
 	tar -C _build $(TAR_TRANSFORM) -zcf $@ html
-pipdeps:
```
rutsky: Target `pipdeps` is used in other places in this Makefile.

rutsky:
You need to install Buildbot in `.travis.yml` with `[docs]` extras (in addition to installing `[docs]` in metabuildbot, as @tardyp noted).

Perhaps there is a way to check that Buildbot is installed with `[docs]` extras in `master/docs/Makefile`?

adityadivekar03:
@rutsky  I was thinking that we could use the ```test -e``` command along with the ```shell```  function and run it to check if the user has the ```[docs]``` dependencies installed or not.  We could run the shell command from the makefile, and if the ```[docs]``` dependencies are installed , we could allow the further execution of the makefile for sphinx documentation. This will however involve spawning a new shell, and I would like your opinion whether it can be used. 

About handling the rest of the pipdeps, once we make sure that the user has installed the ```[docs]```, we could remove the pipdeps commands from the rest of the makefile, as it will only be executed if the ```[docs]``` are installed (because of the condition we put in).

@tardyp I somehow skipped the docutils part. I will correct it in the next commit.

tardyp:
@adityadivekar03 we will be very glad if you can finish this PR! Every little helps!

adityadivekar03:
Yes I really do intend to finish it. I was actually waiting for @tardyp 's opinion on the approach I gave in the comments section! Could you give your opinion if possible? :)

tardyp:
I think we should rather modify conf.py to check for python deps, and fail with an explicit message, asking for install of [docs]
I prefer python code to makefile :)

adityadivekar03:
Yes, editing the makefile could be a bit cumbersome and difficult.
Your idea sounds better. I'll try implementing it and definitely revert soon!

adityadivekar03:
@tardyp I've added try-except block to check if the dependencies exist in the conf.py file
<pre><code>
try:
    import pkg_resources
    pkg_resources.require('docutils>=0.8')
except DistributionNotFound:
    print("Please install the [docs] dependencies first")
except VersionConflict: 
    print("The version of docutils is not supported. Please install a version >= 0.8")
</code></pre>

Is it okay to just print the above or do you want to catch the error differently?

tardyp:
As [] not well known feature. I'll recommend to print the full command pip
install

Le sam. 5 mars 2016 08:38, adityadivekar03 <notifications@github.com> a
écrit :

adityadivekar03:
@tardyp  There were conflicts in the original MR, so I am putting in a new MR.

master/setup.py
```diff
@@ -426,9 +426,16 @@ def define_plugin_entries(groups):
             'buildbot-www',
             'buildbot-slave'
         ],
+<<<<<<< HEAD
```
tardyp: you forgot to fix the merge conflict issue.

tardyp:
you can just amend your commit, and forcepush the result on the same branch. its better like this.

.travis.yml
```diff
@@ -128,7 +128,11 @@ install:
   # astroid > 1.4 is incompatible with pylint 1.1.0
   - "[ $TESTS != lint ] || pip install pylint==1.1.0 pep8==1.5.7 flake8  astroid==1.3.8"
   # TODO: Pin Sphinx version to workaround http://trac.buildbot.net/ticket/3408
-  - "[ $TESTS != docs ] || pip install sphinx==1.3.3"
+  - |
+      [ $TESTS != docs ] || \
+      pip install sphinx==1.3.3 \
```
tardyp: why not doing pip install -e ./master[docs] ?

tardyp: that would require also to add sphinx to [docs] and add more coherency

adityadivekar03:
I apologize for all the patchy work. And I have opened a new MR for this. I tried rebasing before, and applying my patch on top of it. It did not occur to me that I could have added the lines in setup.py myself.
And I've made the changes in travis.yml as per your recommendation, and added the sphinx package to ```[docs]```

PR #2004, opened 2016-02-19
[doc] Numbering sections

XonqNopp:
It helps the reader know where he is when the sections are numbered.

### DIFF
--- a/master/docs/index.rst
+++ b/master/docs/index.rst
@@ -15,6 +15,7 @@ Table Of Contents
 
 .. toctree::
    :maxdepth: 2
+   :numbered:
 
    tutorial/index
    manual/index


tardyp:
could you please copy paste a screenshot showing what this is actually doing?

djmitche:
My worry is that these numbers will change from version to version as sections are added and removed.  Does that cause a problem for readers?

sa2ajj:
I echo @djmitche's concern.

And I think that numbers **will** change.

djmitche:
To be clear, if that's OK -- if the numbers are just useful for navigation of a single version of the docs -- then I think this is fine.

XonqNopp:
Sorry if I was not clear. It is not meant to memorize where to find the doc. But if I am looking for something in section 2 and I click on "next" for five minutes, I must everytime check in the right pane if I am still in section 2. Having sections numbered means I only have to look at section titles to know if I am looking at the right page ;-)

rutsky:
@XonqNopp your changes created other issue: http://trac.buildbot.net/ticket/3468
Can you take a look at how it can be solved?

PR #2005, opened 2016-02-19
make waitForFinish a renderable

tardyp:
I think this should be a better option than #1998


tardyp:
@sa2ajj do this fix address your concerns?

tardyp:
@sa2ajj ping?

PR #2011, opened 2016-02-22
check that ShellCommand command in set in constructor

rutsky:
This change prevents misconfiguration issues when ShellCommand is
constructed without specifiying `command' argument.


jaredgrubb:
"ShellCommand's 'command' must be set"

rutsky:
@jaredgrubb yep, haven't run tests locally at all :)

rutsky:
I added check that `command` is specified only if `ShellCommand` or `WarningCountingShellCommand` is instantiated directly.

tardyp:
:+1:

PR #2013, opened 2016-02-23
Setup Circle CI to build docs

rutsky:


rutsky:
@djmitche if you enable Circle CI for buildbot/buildbot repo with this PR we will see if docs are built for pull requests.

tardyp:
That is a third CI system running for this project. Not sure it's worth

Le mar. 23 févr. 2016 12:05, Vladimir Rutsky <notifications@github.com> a
écrit :

rutsky:
@tardyp 
> That is a third CI system running for this project. Not sure it's worth 

It's worth to get browsable documentation for each Pull Request, this is a try to resolve this issue: http://trac.buildbot.net/ticket/3459

Also I don't think it's bad to have multiple CI, since we can get them for free and it's nothing more from our part that writing (and maintaining) configuration file, like `.travis.yml`.
CI have different abilities and scopes. Currently I'm trying Circle CI only to get browsable docs for each PR, since Travis doesn't allow that.

rutsky:
OK, this should work.

Now for all PRs Circle CI should build documentation and store result as build artifact.
Artifacts are publicly served (for example [this one](https://circle-artifacts.com/gh/rutsky/buildbot/7/artifacts/0/home/ubuntu/buildbot/master/docs/_build/html/index.html)), but to see builds artifacts on Circle CI you need to be logged in into Circle CI.

Here is how Circle CI build page looks like when you logged in, note "Artifacts" tab:

screenshot_2016-02-23_004

Here is how Circle CI build page looks like for anonymous user, note that there is no "Artifacts" tab:

screenshot_2016-02-23_005

So at this moment, to see built documentation user need to go to Circle CI build page ("Details" link in list of all checks), login with GitHub account, select "Artifacts" tab, and navigate in files to built documentation `index.html`:

screenshot_2016-02-23_006

Link on index.html can be then posted in PR discussion, so other users won't need to be logged in to Circle CI to find built documentation.

Posting direct link can be automated, e.g. by setting up hook on Circle CI that will trigger some bot when build finishes, bot will prepare direct link docs for just finished build and post in into PR, for example, as Status:

screenshot_2016-02-23_007

Here I manually posted status to PR with link on built documentation: https://github.com/rutsky/buildbot/pull/3

djmitche:
This is really cool.  And minimal difficulty to maintain.  So, I'm all for it, especially if we add the bot you mentioned.

rutsky:
Adding bot requires additional efforts and I'm not sure what is the best way to do this, so I can't fulfill this desire in this PR.

I see following options:

1. Setup (i.e. run in BB infra) bot like [Hubot](https://hubot.github.com/) which is designed to do such tasks (listen for hooks and call GitHub API methods).

2. Setup metabbot configuration to handle hook from Circle CI.

3. Use third-party services. In modern times looks like there are services for everything and perhaps service that is capable to do required task already exists. For example, maybe [Zapier](https://zapier.com/zapbook/webhook/github/6870/circle-ci-catch-hook-to-github-create-issue/) will be able to do this.

djmitche:
Another option is just to generate these docs in the metabuildbot.  After
all, we maintain a CI system!

It would be good for us to solve the question of building PRs with Buildbot.

But I realize that this is a bit more work -- for now, Circle CI is fine.

Dustin

rutsky:
Running build in our infra is preferred option, but we need to handle security-related issues with them first.
I believe even "just" docs building with Sphinx is not secure, for example autodoc plugin actually import and run arbitrary Python module code.

Perhaps easiest option would be to setup something like hardened with SELinux/seccomp throw-away Docker containers inside throw-away VMs (for cases of misconfiguration or bugs in Docker/Linux), *and* continuously maintain this setup (apply security updates, etc), but we need significant human and CPU resources to do that.
I would *love* to replace third party CI with native Buildbot, I just don't see an option to do it at this moment.

tardyp:
I think we agreed already to merge this. I think you should contact the circleCI to know if we can configure the status report message and include the link there.
Maybe they will agree to add the feature and that will save you some time.

PR #2015, opened 2016-02-23
Documenation note

Riziero:
about mysql 5.x and 'can't have a default value' errors.

### DIFF
--- a/master/docs/manual/cfg-global.rst
+++ b/master/docs/manual/cfg-global.rst
@@ -68,6 +68,8 @@ Setting this parameter ensures that connections are closed and re-opened after t
 If you see errors such as ``_mysql_exceptions.OperationalError: (2006, 'MySQL server has gone away')``, this means your ``max_idle`` setting is probably too high.
 ``show global variables like 'wait_timeout';`` will show what the currently configured ``wait_timeout`` is on your MySQL server.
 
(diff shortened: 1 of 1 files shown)


rutsky:
This is related to http://trac.buildbot.net/ticket/3451.

Is `sql_mode='MYSQL40'` this a "true way" for fixing warnings?

Why MySQL 5.* doesn't allow defaults on TEXT/BLOB columns? Is there are performance or other issues with it?
Maybe we should use another column type for it in MySQL?

Riziero:
@rutsky I think you're right. Since I moved from sqlite to mysql (5.7 with the above param) the whole system got slow, surfing the pages is a pain.

Kinda hard to say if it's due to sql_mode being set to MYSQL40.

tardyp:
:+1:

Riziero:
Hmm thinking the cause of the slow down was mysql I switched to postgres. Same thing. Gonna try sqlite however at this point I think the PR is fine.

rutsky:
This is bugreport on MySQL bugtracker for this issue: https://bugs.mysql.com/bug.php?id=21532
It dates back to 2006 and looks like it never will be fixed.

Lets just leave the note about it.

djmitche:
Haha, every MySQL bug report I've seen has been in the "will never fix" state.

PR #2016, opened 2016-02-23
fix problem where parent_changeids might be corrupted

tardyp:
In some case, parent_changeids will return an unknown changeid,
we should at least return the list of changes found.

### DIFF
--- a/master/buildbot/db/changes.py
+++ b/master/buildbot/db/changes.py
@@ -198,6 +198,9 @@ class ChangesConnectorComponent(base.DBConnectorComponent):
                        change['parent_changeids']):
                     # For the moment, a Change only have 1 parent.
                     change = yield self.master.db.changes.getChange(change['parent_changeids'][0])
+                    # http://trac.buildbot.net/ticket/3461 sometimes, parent_changeids could be corrupted
+                    if change is None:
+                        break
                     changes.append(change)
         defer.returnValue(changes)
 


tardyp:
shall I merge?

rutsky:
Is missing parent is normal behaviour?
If it's not, maybe we should at least output warning in log that something wrong with Buildbot/Buildbot configuration?

rutsky:
@tardyp do you have any idea why parent may be missing?

tardyp:
Its not config. It's probably corruption. There is not much to do. So
better not polute the log

Le mar. 1 mars 2016 12:06, Vladimir Rutsky <notifications@github.com> a
écrit :

rutsky:
@tardyp but who is responsible for corruption?

Is this is corrupted DB due to hard disk failure or other similar malfunction which we cannot control in any way and we are not responsible, 
or is this due to some bug in Buildbot or not handled case (race condition, unexpected slave disconnection)?

tardyp:
This is hard for me to debug the corruption. We got a stacktrace from a user without too much more information on what was the original cause.
So I figure we better not crash buildbot in such a case. If somebody can reproduce original corruption with better data, we can go further

rutsky:
But how we will now that corruption is occurred?
If it will be silently ignored every time there is possibility that we never discover source of this problem (and who knows which other parts are getting corrupted, this may be critical issue) - that's why I suggest to at least log this issue.

I'm +1 for merging this, just want to make sure that we are not "closing our eyes" on actual issue that will arise sooner or later.

tardyp:
The thing is: there is a foreign key for the parent change_id, I verified that. So the db should reject any attempt on setting non-existing parent_change_id.

Then, the only reason why this line could crash with this signature is change = None
and the only reason why getChange would return None is non existent changeid, and this case would have been catched by fk. This is why I say "because of corruption" 

But.. for sqlite, the fk are disabled... until we merge bug3472. So maybe this is indeed a bug in the parent_change_id insertion code, that would have been catched by proper configuration of db.

I dont think there are many instances of nine running against mysql or pg, so maybe we will get better reproduction of this bug  after we release a fk enforcement with sqlite.

rutsky:
Ah, violated FK constraint is indeed case we shouldn't consider.

PR #2017, opened 2016-02-23
XXX: test PR with Circle CI

rutsky:


rutsky:
Oh, Circle CI already started build on #2013

PR #2006, opened 2016-02-20
Nine: a few doc fixes and code fixes

jaredgrubb:
I have finally got some time to play with nine and had some suggestions on doc-fixes and fixed one bug on the module name for the ldap stuff.

PR #2020, opened 2016-02-24
add fixing of LDAP plugin to relnotes

rutsky:


rutsky:
Fixed in PR #2006

tardyp:
thanks

PR #2021, opened 2016-02-24
remove tox files

djmitche:
Aditya noticed that these don't work (they attempt to run `python setup.py test`, which `setup.cfg` aliases to `python setup.py trial -m buildbot`, but there is no "trial" subcommand, so this fails).

I think we should have one and only one way of running tests -- invoking trial.


rutsky:
IMO `python setup.py test` should work regardless of used test system. This is some kind of "standard" Python/setuptools way to run tests and typical Python developer will expect that it will work.

Tox is nice if we need to run test in multiple Python interpreters. While it's OK to run tests manually using Py27 virtualenv, it's tiresome to manually run tests for Py26, Py26 (and Py34, Py35, when Buildbot/Twisted will be Py3-ready).

But since none of it works, lets at least don't confuse newcomers with them, lets remove them.

BTW, `python setup.py test` should be that single entrypoint for running tests, that will be called from metabbot and Travis.

@djmitche, what will happen after your changes if I run `python setup.py test`?

djmitche:
I assume it will continue to fail as it does now..

tardyp:
I like tox because it allows to easily replicate the CI environment for our developers.

In ramlfication, they write their travis.yml to call directly tox, and eventually all the automation setup is described in tox.
https://github.com/spotify/ramlfications/blob/master/.travis.yml

running our test is *not* only running trial.
Its installing [test], running pep8, etc etc.

tox takes care of the virtualenv complication, and I think python developers are starting to think "tox" rather than setup.py test.

I think we underestimated tox. The real problem with tox is that it looks complicated to install several packages like we need

rutsky:
On the second though maybe we just fix `python setup.py test` and tox?

rutsky:
tox/`setup.py test` was broken in 74e75cf2aace6a1fdb3fde2cf1326017da822712, when [setuptools_trial](https://github.com/simplegeo/setuptools-trial) were removed from dependencies.

IMO having `setuptools_trial` in `setup_requires` is OK, but if it's not, lets inline that [TrialTest](https://github.com/simplegeo/setuptools-trial/blob/master/setuptools_trial/setuptools_trial.py#L6) class from `setuptools_trial` and use it for `setup.py trial` (as was when `setuptools_trial` were in deps list).

djmitche:
that sounds fine to me - do you want to close this PR then?

rutsky:
Do we really care that `setuptools_trial` is being installed when user runs `python setup.py test`/`python setup.py trial`/`python setup.py --help-commands` **if it is not yet installed**?

In case of developers environments this shouldn't bother us at all, since we can (and should) put `setuptools_trial` in `buildbot[test]` extras (or in probably upcoming `requirements-dev.txt`).

IMO instead of removing tox it's better to revert 74e75cf2aace6 and add `setuptools_trial` into `buildbot[test]` extras.

Second option is to copy-paste `setuptools_trial`'s TrialTest, which I don't really like (as any copy-pasting).

Perhaps @tomprince can comment on this issue, since he is the author of 74e75cf2aace6.

rutsky:
I'm trying to fix `python setup.py test`: #2022

PR #2022, opened 2016-02-25
fix `python setup.py test` invocation

rutsky:
Work in progress.
Should fix tox too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,11 @@ apidocs/reference.tgz
 common/googlecode_upload.py
 master/docs/tutorial/_build
 _build
-.tox
+# tox
+.tox/
+# downloaded by setup_requires eggs
+*.egg
+.eggs/
 master/setuptools_trial*
 master/buildbot/www/static/js/default/
 sandbox
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ matrix:
 
     # Configuration to run `python setup.py test` to check this test runner.
     - python: "2.7"
-      env: TESTS=setuppy_test
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=setuppy_test
 
 # Dependencies installation commands
 install:
@@ -96,7 +96,7 @@ install:
   # Install boto and moto for running EC2 tests
   # Install txgithub to run buildbot.status.github module tests
   - |
-      [ $TESTS != trial ] || \
+      [ $TESTS != trial -a $TESTS != coverage ] || \
       pip install -e pkg \
                   -e master[tls] \
                   -e slave \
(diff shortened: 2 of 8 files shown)


rutsky:
Looks like using `setuptools_trial` from PyPI is not an option at this moment — it doesn't work with recent setuptools due to minor error: https://github.com/simplegeo/setuptools-trial/pull/1
So I will inline `TestTrial` in `setup.py` of master and slave packages if upstream will not fix this issue soon.

`python setup.py test` for slave package works perfectly (with clean environment, package is not installed — all deps are downloaded during test step and cached locally).

`python setup.py test` for master package fails to install locally autobahn+Twisted due to setuptools imprefect sandbox magic: https://bitbucket.org/pypa/setuptools/issues/500/when-directorysandbox-is-active-isinstance
After installing autobahn+Twisted into virtualenv all tests runs perfectly.

tardyp:
m2c, I dont think we should use setup_tools trial
We should rather invoke trial from tox (and tox from setup.py test)

rutsky:
@tardyp 
> m2c, I dont think we should use setup_tools trial
> We should rather invoke trial from tox (and tox from setup.py test)

I think tox should call `setup.py` and not in reverse.

setuptools in `setup.py` manages package build and test dependencies and define how package should be build, tested and deployed *in the specific Python environment* (i.e. setup.py is called in specific Python environment).

Tox is "virtualenv-based automation of test activities" — it manages Python environments and can automate typical testing task, like "install my package in Py2.6, Py3.5 and run tests, build docs, prepare coverage report". Also tox may not build or install any package at all — we can ask to tox to just run some shell scripts or Makefiles in different environments  (in contrast `setup.py` by default only knowns how to build, install and test package, and nothing more).

Tox is on higher level of abstraction, than setuptools in `setup.py`.

If `setup.py` will call tox, next step that tox will do is to run `setup.py` again to either install the package in venv or to install it's dependencies.

rutsky:
Also this is how testing done in one of the most popular Python test-runners: [pytest](https://pytest.org/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner) and [nose](https://nose.readthedocs.org/en/latest/setuptools_integration.html) (and perhaps in other frameworks that I'm not familiar).

rutsky:
`setuptools_trial` should work now, and I expect autobahn+Twisted installation will still fail.

rutsky:
Here is issue that was cause of `setup.py test` handling removal: http://trac.buildbot.net/ticket/2369 (in case anyone interested).

rutsky:
`python setup.py test` should work with single exception: it can't install `autobahn` due to this issue: https://bitbucket.org/pypa/setuptools/issues/500/when-directorysandbox-is-active-isinstance

I suggest to ignore this issue and let user install `autobahn` it explicitly if he need it.
These changes still useful for running `python setup.py test` in worker and `python setup.py test` in master with preinstalled `autobahn`.

PR #2023, opened 2016-02-27
Select distinct buildermasterid's to save duplication

djmitche:
Fixes #3472.

PR #2024, opened 2016-02-28
Bug3472

tardyp:
improve version on #2023  from @djmitche 
- Fix the foreign key issue that pg catched
- add foreign key check on sqlite if it is supported
- Factorize the batching
- document all this

### DIFF
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -113,9 +113,6 @@ class BuildbotEngineStrategy(strategies.ThreadLocalEngineStrategy):
 
     def set_up_sqlite_engine(self, u, engine):
         """Special setup for sqlite engines"""
-        import sqlite3
-        if sqlite3.sqlite_version_info < (3, 6, 19):
-            raise RuntimeError("Please use sqlite version > 3.6.19")
         # try to enable WAL logging
         if u.database:
             def connect_listener(connection, record):
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -113,13 +113,16 @@ class BuildbotEngineStrategy(strategies.ThreadLocalEngineStrategy):
 
     def set_up_sqlite_engine(self, u, engine):
         """Special setup for sqlite engines"""
+        def connect_listener_enable_fk(connection, record):
+            # fk must be enabled for all connections
+            if not getattr(engine, "fk_disabled", False):
+                connection.execute('pragma foreign_keys=ON')
+
+        sa.event.listen(engine.pool, 'connect', connect_listener_enable_fk)
         # try to enable WAL logging
         if u.database:
             def connect_listener(connection, record):
                 connection.execute("pragma checkpoint_fullfsync = off")
-                # fk must be enabled for all connections
-                if not getattr(engine, "fk_disabled", False):
-                    connection.execute('pragma foreign_keys=ON')
 
             sa.event.listen(engine.pool, 'connect', connect_listener)
 
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -120,7 +120,9 @@ class BuildbotEngineStrategy(strategies.ThreadLocalEngineStrategy):
         if u.database:
             def connect_listener(connection, record):
                 connection.execute("pragma checkpoint_fullfsync = off")
-                connection.execute('pragma foreign_keys=ON')
+                # fk must be enabled for all connections
+                if not getattr(engine, "fk_disabled", False):
+                    connection.execute('pragma foreign_keys=ON')
 
             sa.event.listen(engine.pool, 'connect', connect_listener)
 
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -113,19 +113,16 @@ class BuildbotEngineStrategy(strategies.ThreadLocalEngineStrategy):
 
     def set_up_sqlite_engine(self, u, engine):
         """Special setup for sqlite engines"""
+        import sqlite3
+        if sqlite3.sqlite_version_info < (3, 6, 19):
+            raise RuntimeError("Please use sqlite version > 3.6.19")
         # try to enable WAL logging
         if u.database:
             def connect_listener(connection, record):
                 connection.execute("pragma checkpoint_fullfsync = off")
+                connection.execute('pragma foreign_keys=ON')
 
-            if sautils.sa_version() < (0, 7, 0):
-                class CheckpointFullfsyncDisabler(object):
-                    pass
-                disabler = CheckpointFullfsyncDisabler()
-                disabler.connect = connect_listener
-                engine.pool.add_listener(disabler)
-            else:
-                sa.event.listen(engine.pool, 'connect', connect_listener)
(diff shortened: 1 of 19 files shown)


master/buildbot/db/model.py
```diff
@@ -759,6 +759,17 @@ def thd(engine):
             return db_version == repo_version
         return self.db.pool.do_with_engine(thd)
 
+    def enableForeignKeys(self):
+        # foreign keys is a recent feature of sqlite3, it is still disabled by default
+        # https://www.sqlite.org/pragma.html#pragma_foreign_key_check
```
djmitche: Will this setting fail on older versions of sqlite?

djmitche: From the docs, it looks like yes.  In that case, should this function catch the exception and log it, if the support is not present?

master/docs/developer/database.rst
```diff
@@ -1856,6 +1880,25 @@ id.
...
+The C-api ``sqlite_limit`` is just not bound to the python.
+
+When you hit this problem, you will get error like the following:
+
+.. code-block:: none
+
+    sqlalchemy.exc.OperationalError: (OperationalError) too many SQL variables
+    u'DELETE FROM scheduler_changes WHERE scheduler_changes.changeid IN (?, ?, ?, ......tons of ?? and IDs .... 9363, 9362, 9361)
+
+You can use the method :py:meth:`doBatch` in order to write batching code in a consitent manner.
```
djmitche: "consistent"

djmitche:
Aside from the compatibility question, looks good!

master/buildbot/db/model.py
tardyp: I tried to run a typoed plasma and got no error. This is why I kept it like
this. Ill look at it more seriously

aelsabbahy:
Thank you both for the quick turnaround on this.

tardyp:
sqlite 3.6.19 is there since 2009. I cannot find an ubuntu version with older version, so I cannot even checks which exception to catch. so I just did a version check.
http://fr.archive.ubuntu.com/ubuntu/pool/main/s/sqlite3/

rutsky:
When I configured testing on AppVeyor (i.e. on Windows) noticed that in on AppVeyor's Python 2.7 SQLite has version 3.6.21, and I think same version is bundled with official Python 2.7 Windows installer.

I can assume, that Windows installer for Python 2.6 has SQLite <3.6.21 and most probably < 3.6.19 ("final" release of Python 2.6 was at October 1, 2008).

master/buildbot/db/enginestrategy.py
```diff
@@ -113,19 +113,18 @@ def special_case_sqlite(self, u, kwargs):
 
     def set_up_sqlite_engine(self, u, engine):
         """Special setup for sqlite engines"""
+        import sqlite3
+        if sqlite3.sqlite_version_info < (3, 6, 19):
+            raise RuntimeError("Please use sqlite version > 3.6.19")
```
djmitche: Se suggest 3.7.0 in requirements.rst, but indicate that earlier versions are also supported.  So we should remove that indication from the docs.

master/buildbot/util/sautils.py
```diff
@@ -58,3 +59,16 @@ def Table(*args, **kwargs):
     # a non-utf8 character set (mysql's default)
     kwargs['mysql_character_set'] = 'utf8'
     return sa.Table(*args, **kwargs)
+
+
+@contextmanager
+def withoutSqliteForeignKeys(engine):
+    if engine.dialect.name == 'sqlite':
+        engine.fk_disabled = True
```
rutsky: Maybe assert here that `engine.fk_disabled` was `False`?
I don't say that is worth it to make `withoutSqliteForeignKeys` reentrant, but if not at least assert that it's not called recursively should be added.

master/buildbot/test/util/migration.py
```diff
@@ -79,9 +79,10 @@ def upgrade_thd(engine):
                 schema = migrate.versioning.schema.ControlledSchema(
                     engine, self.db.model.repo_path)
                 changeset = schema.changeset(target_version)
-                for version, change in changeset:
-                    log.msg('upgrading to schema version %d' % (version + 1))
-                    schema.runchange(version, change, 1)
+                with sautils.withoutSqliteForeignKeys(engine):
```
rutsky: Is migrations with enabled FK check doesn't work in SQLite?

rutsky:
FYI, SQLite FK constraint support still is very limited comparing to PostgreSQL FK support.
For example, with enabled FK check in SQLite you can rename table on which other tables have FK references, and DB will become broken.

Also AFAIK MySQL's MyISAM storage engine doesn't support FK constraint at all, and I heard that Buildbot works with MySQL only with MyISAM storage engine.

master/docs/developer/database.rst
```diff
@@ -1779,16 +1799,22 @@ Postgres, add them to ``implied_indexes`` in
 
 Foreign key checking
 --------------------
-Non sqlite db backends are checking the foreign keys consistancy. As sqlite is much
-easier to install, most of the developer would only test against sqlite, and then get
-errors on metabuildbot. In order to avoid that the fakedb can check the foreign key
-consistancy of your test data. for this, just enable it with::
+All supported db backends are checking the foreign keys consistancy.
```
rutsky: What about MySQL with MyISAM storage engine?

master/buildbot/test/util/migration.py
tardyp: No :-(, I've documented why in database.rst
I tried hard to make it work, but eventually sqlalchemy-migrate does not help

master/docs/developer/database.rst
tardyp: Ah you are right. You can run buildbot with Innodb, but you have to configure it for accepting bigger keys. I made a PoC once, but never got much further.
We need to document and test it more heavily

master/buildbot/test/util/db.py
```diff
@@ -135,7 +135,11 @@ def __thd_clean_database(self, conn):
 
             # Drop all reflected tables and indices. May fail, e.g. if
             # SQLAlchemy wouldn't be able to break circular references.
-            meta.drop_all()
+            # Sqlalchemy fk support with sqlite is not yet perfect, so we must deactivate fk during that
+            # operation, even though we made our possible to use use_alter
+            with withoutSqliteForeignKeys(conn.engine):
+                meta.drop_all()
```
rutsky: Is without disabling FK this fails in migrations due to sqlalchemy-migrate corrupting FK constraints?

tardyp: as sqlite does not support alter table, migrate does  renames the table, and then the fk contraints are renamed too, and then you cannot destroy the renamed table.
The code is here. It supposed to do the correct thing, but this does not work for me, and gives a lot of "migration_tmp does not exist" errors
https://github.com/openstack/sqlalchemy-migrate/blob/master/migrate/changeset/databases/sqlite.py

tardyp:
As I dig into the db code, I also cleaned up workarounds for old pysqlite2 bugs.

Created a bug for sqlalchemy-migrate: https://bugs.launchpad.net/sqlalchemy-migrate/+bug/1552351
We will see how they will advise..

tardyp:
ok for merge?

PR #2025, opened 2016-03-01
reexport .ai files into SVG and remove .ai files

rutsky:
Fixes [bug 3427](http://trac.buildbot.net/ticket/3427).


rutsky:
As a downside: resulting SVG are 1.5 times larger. master.
`master.svg` is 490K now, was 320K.

rutsky:
Here is how these SVG looks like in docs: https://circle-artifacts.com/gh/rutsky/buildbot/45/artifacts/0/home/ubuntu/buildbot/master/docs/_build/html/manual/introduction.html

rutsky:
Here is how they looked before: https://circle-artifacts.com/gh/rutsky/buildbot/44/artifacts/0/home/ubuntu/buildbot/master/docs/_build/html/manual/introduction.html

tardyp:
Now that svg is supported by  all the browsers, I would be for removing the pngs. 
http://caniuse.com/#feat=svg

It is possible that all those schemas could be written in blockdiag. That could save you a couple of hours..
http://blockdiag.com/en/blockdiag/demo.html

rutsky:
@tardyp I believe PNGs are used not for HTML version of docs (on our docs site I see SVG), but for PDF: http://trac.buildbot.net/ticket/2238

I don't think PDF version of manual is actively used, but I don't think it should be removed either, since it's building cost almost nothing for us (it's done by ReadTheDocs).

While it would be nice to have diagrams generated by code (everything is better, when it's done by writing some code :) ), I don't this it's feasible to do so:
currently diagrams contains non-trivial layout and styling, take a look at [this one](https://circle-artifacts.com/gh/rutsky/buildbot/45/artifacts/0/home/ubuntu/buildbot/master/docs/_build/html/_images/slaves.svg) — either we need to drop all niceties, or we need to do real hard work with `blockdiag`.

PR #2026, opened 2016-03-02
Fix AttributeError for using asDict on None when codebases are being used

benallard:
``self.build.getSourceStamp()`` will return None when codebases are being used (different from ``''``)


tardyp:
glad to see you again :)
This fix breaks some unit tests

benallard:
I was never gone for real ;)
Should be fixed now.

tardyp:
thanks

PR #2028, opened 2016-03-05
Install [docs] dependencies in setup.py

### LINKED ISSUE #3435: Documentation for Pushover reporter says "sends email"

adityadivekar03:
Remove the pip install command from the docs Makefile used for installing the sphinx distribution. 
Add condition in the conf.py file to check if the [docs] dependencies are installed. If not, suitable error is displayed.

fixes #3435


master/setup.py
```diff
@@ -429,6 +429,11 @@ def define_plugin_entries(groups):
         'tls': [
             'Twisted[tls] ' + twisted_ver,
         ],
+        'docs': [
+            'sphinx',
```
tardyp: ==1.3.3
as specified in travisrc

tardyp:
you also need to fix circleci to also install [docs] in it

circle.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
   override:
     - pyenv global 2.7.11
     - pip install -e pkg -e master -e slave
+    - pip install -e master[docs]
```
rutsky: @adityadivekar03 add `[doc]` to `master` in the line above, please.

master/docs/conf.py
```diff
@@ -22,6 +22,17 @@
 
 # -- General configuration -----------------------------------------------------
 
+# Check if [docs] dependencies are installed
+try:
+    import sphinxcontrib.blockdiag
```
rutsky: This will raise `ImportError` that is not handled.
Also message is only printed to stdout, I suggest to raise exception instead.

I would do it in this way:

```python
try:
except ImportError:
    raise RuntimeError("sphinxcontrib.blockdiag is not installed. Please install documentation dependencies with `pip install buildbot[docs]`")

import pkg_resources
try:
    pkg_resources.require('docutils>=0.8')
except resources.ResolutionError:
    raise RuntimeError("docutils is not installed or has incompatible version. Please install documentation dependencies with `pip install buildbot[docs]`")
```

Also, I'm worried that we will implicitly list all docs dependencies there, so it will not be true that "all documentation dependencies specified in the single place", which is one of the goals of this PR.

master/docs/conf.py
```diff
@@ -22,6 +22,17 @@
...
+    import pkg_resources
+    pkg_resources.require('docutils>=0.8')
+except DistributionNotFound:
```
rutsky: `DistributionNotFound` is not imported.

master/docs/Makefile
```diff
@@ -10,8 +10,6 @@ TAR_TRANSFORM := $(if $(filter bsdtar,$(TAR_VERSION)),-s /^html/$(VERSION)/,--tr
 docs.tgz: clean html singlehtml images-png
 	sed -e 's!href="index.html#!href="#!g' < _build/singlehtml/index.html > _build/html/full.html
 	tar -C _build $(TAR_TRANSFORM) -zcf $@ html
-pipdeps:
```
rutsky: I think I already commented this change: if we remove `pipdeps` make target, lets remove it in whole Makefile (it's used in several places).

adityadivekar03: Removing pipdeps from the entire file - 
```
-html: conf.py pipdeps
+html:
```
I'm not sure of what should replace the ```pipdeps``` command. 
Frankly, I haven't been able to completely understand makefiles till now.

rutsky: In short, Makefiles describes set of targets (usually files), their dependencies and how to build them.

In this example:

```
.PHONY: tutorial manual pipdeps

pipdeps:                                                                                                                                          
        pip install sphinxcontrib-blockdiag 'docutils>=0.8'

html: conf.py pipdeps
        $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
```

two targets are defined: `html` and `pipdeps`.

To build them you can run `make html` and `make pipdeps`

To build `pipdeps` target make will issue commands that are specified below target: `pip install sphinxcontrib-blockdiag 'docutils>=0.8'`.

`html` target has two dependencies: `conf.py` file and `pipdeps` target.
They will be built first before invoking commands to build `html` target.

By default targets are files. `.PHONY` is used to specify that target is not a file (it's strange to me that `html` is not a phony target here).

So you need to remove from other targets `pipdeps` as dependency:

```
-html: conf.py pipdeps
+html: conf.py
```

tardyp:
lgtm

master/docs/conf.py
```diff
@@ -22,6 +22,17 @@
...
 
+# Check if [docs] dependencies are installed
+try:
+    import sphinxcontrib.blockdiag
+except ImportError:
+    raise RuntimeError("sphinxcontrib.blockdiag is not installed. Please install documentation dependencies with `pip install buildbot[docs]`") 
+import pkg_resources
+try:
+    pkg_resources.require('docutils>=0.8')
+except resources.ResolutionError:
```
rutsky: Maybe `pkg_resources`?

rutsky: My example is just example, I haven't tested it :)

master/docs/conf.py
```diff
@@ -22,6 +22,17 @@
 
 # -- General configuration -----------------------------------------------------
 
+# Check if [docs] dependencies are installed
+try:
+    import sphinxcontrib.blockdiag
+except ImportError:
+    raise RuntimeError("sphinxcontrib.blockdiag is not installed. Please install documentation dependencies with `pip install buildbot[docs]`")
```
rutsky: Please split such long lines over several lines.

master/docs/conf.py
adityadivekar03: Yes, you are right, the error being caught is from the pkg_resources module, and needs to caught as `pkg_resources.ResolutionError` and not `resources.ResolutionError`

master/docs/Makefile
adityadivekar03: Thanks a lot for the explanation. :smiley:

circle.yml
rutsky: @adityadivekar03 what about this comment?

rutsky:
@adityadivekar03 your changes looks almost complete, thanks!

Please fix last comment about `master[docs]` in `cicleci.yml`, and rebase your changes against master, looks like there are conflicts.

tardyp:
please rebase, as there are conflicts with recent PR

tardyp:
please avoid closing your PR when doing rebase, you just need to do git push -f.
This allows us to better manage history of a PR

PR #2029, opened 2016-03-05
Fix deprecation warning from docker_py

anish:
Squashes : "Passing host config parameters in start() is deprecated. Please use host_config in create_container instead."

### DIFF
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -22,7 +22,7 @@ class Client(object):
     def images(self):
         return self._images
 
-    def start(self, *args, **kwargs):
+    def start(self, container):
         pass
 
     def stop(self, id):
@@ -41,6 +41,9 @@ class Client(object):
             for line in logs:
                 yield line
             self._images.append({'RepoTags': [tag + ':latest']})
+
+    def create_host_config(self, *args, **kwargs):
+        pass
 
     def create_container(self, image, *args, **kwargs):
         return {'Id': '8a61192da2b3bb2d922875585e29b74ec0dc4e0117fcbf84c962204e97564cd7',
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -64,7 +64,7 @@ class DockerLatentWorker(AbstractLatentWorker):
                  **kwargs):
 
         if not client:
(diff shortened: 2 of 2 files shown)


anish:
This has been deprecated as of docker-py 1.2.3, current version is 1.7.2

appveyor.yml
```diff
@@ -10,6 +10,7 @@ environment:
 install:
   # Twisted requires pywin32 on Windows in order to spawn processes.
   - "%PYTHON%\\python.exe -m pip install pypiwin32"
+  - "%PYTHON%\\python.exe -m pip install docker-py"
```
rutsky: Don't forget to remove this line, since Fake Docker client should be used.

rutsky:
@anish, please add to message [here](https://github.com/anish/buildbot/blob/add_host_config/master/buildbot/worker/docker.py#L67) that `docker-py>=1.4.0` is required.

master/buildbot/test/fake/docker.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, base_url):
     def images(self):
         return self._images
 
-    def start(self, *args, **kwargs):
+    def start(self, instance):
```
rutsky: I think argument should be called `container`: https://docker-py.readthedocs.org/en/latest/api/#start

rutsky:
Other than `instance` -> `container` renaming this PR LGTM, thanks!

anish:
@rutsky but the code that starts this calls it instance : 
```
        self.instance = instance
        docker_client.start(instance)
        log.msg('Container started')
```

Either is okay with me, but please confirm

rutsky:
@anish, more important is that `container` is part of `start()` function definition in `docker-py`, and we should mimic it in our Fake implementation.

This is an example of `start()` usage from `docker-py` documentation:

```
>>> from docker import Client
>>> cli = Client(base_url='tcp://127.0.0.1:2375')
>>> container = cli.create_container(
...     image='busybox:latest',
...     command='/bin/sleep 30')
>>> response = cli.start(container=container.get('Id'))
>>> print(response)
None
```
and it explicitly uses `container` as keyword argument, which will not work in your implementation.

anish:
gotcha. Changed as requested.

rutsky:
LGTM!

PR #2030, opened 2016-03-06
use flake8 instead of pyflakes in slave Makefile

rutsky:


rutsky:
Also I fixed master's Makefile.

Now flake8 in the root Makefile is forwarded to the subdirectory Makefiles.

Also I enabled flake8 check on docs/conf.py.

tardyp:
looks good!

rutsky:
Somehow pylint reports that `UnhighlightedError` is not inherited from `BaseException`: https://travis-ci.org/buildbot/buildbot/jobs/113982097#L635

I can not reproduce this locally and code looks correct to me:
`UnhighlightedError` inherits from `SphinxWarning`, which inherits from `SphinxError`, which inherits from `Exception`.

tardyp:
Le dim. 6 mars 2016 à 12:52, Vladimir Rutsky <notifications@github.com> a
écrit :

> Somehow pylint reports that UnhighlightedError is not inherited from
> BaseException: https://travis-ci.org/buildbot/buildbot/jobs/113982097#L635
>
> I can not reproduce this locally and code looks correct to me:
> UnhighlightedError inherits from SphinxWarning, which inherits from
> SphinxError, which inherits from Exception.
>

I think this might be because sphinx is not available in travis
environment, so pylint cannot find out the ancestor tree

rutsky:
@tardyp, good idea, thanks!

I'll try to install sphinx for `lint`-ing.

PR #2036, opened 2016-03-07
Add docker host config support

anish:
Add ability to specify host configuration while creating a
docker latent worker. Currently regularly used options like --net and --dns cannot be provided while creating a docker latent worker, resulting in a broken worker as this often breaks network access

A new hostconfig parameter is added that accepts host config options as a dict, complete list of supported options can be found [here](https://docker-py.readthedocs.org/en/latest/hostconfig/).

PR #2038, opened 2016-03-08
add method createHostConfig for the class DockerLatentWorker.

delanne:
For my CI setup, I need a buildbot worker (DockerLatentWorker) which is able to build docker image or run docker-compose. In order to be able to use docker in docker, the container must be run with the privileged flag.

So, I add the method createHostConfig, in order to be able to create a custom HostConfig when needed.


tardyp:
looks good. need an update on the doc, and relnote. thanks!

delanne:
#2038 is a duplicate of #2036. *i haven't notice that #2036 exists when I created this PR.

tardyp:
funny that you two worked on the exact the thing at the same time

PR #2040, opened 2016-03-09
setup properties in time for LatentWorkers to use it

delanne:
The image parameter is renderable, but the build's property are the slave subtanciation. As a consequence, when the code render the "image" parameter of DockerLatentWorker, the properties aren't set.

The build properties must be set before worker substanciation


tardyp:
looks good. I could swear I did the exact same thing a while ago

anish:
solved some major issues for me, thanks for the work @delanne !

anish:
@delanne uhh, correction. Works, however properties are missing if you hit 'rebuild' on a build done on a latentslave.

PR #2041, opened 2016-03-11
Make github payload parsing for pull requests more consistent

anish:
Currently, for a push request the repository field is filled with the
value of `['repository']['url']` but a pull request uses
`['repository']['clone_url']`. The clone_url has a .git suffix, this
results in a non-regex filter for a git url working with only the push
or the pull request, not both.

Fix and also parse and fill repository field for pull requests, update documentation

Sample of the parsed payload for clarification :

Push :
```
 added change with revision ce9e9d44b2d090541099f540e706c8ae54b89a05 to database
 injected change Change(revision=u'ce9e9d44b2d090541099f540e706c8ae54b89a05', who=u'Anish Bhatt <xxxx@xxxxx.xxx>', branch=u'master', comments=u'Create TEST', when=1457647806, category=None, project=u'knightswhosaynee/testrepo', repository=u'https://github.com/knightswhosaynee/testrepo', codebase=u'')
```
Pull :
```
 added change with revision db35deb31691cd50617342718125ed1aa61303e3 to database
 injected change Change(revision=u'db35deb31691cd50617342718125ed1aa61303e3', who=u'anish-bhatt', branch=u'refs/pull/3/head', comments=u'GitHub Pull Request #3 (1 commit)', when=1457647899, category=u'pull', project=u'', repository=u'https://github.com/knightswhosaynee/testrepo.git', codebase=u'')
```
Note how the repository field doesn't match up due to suffix and
empty project field in the pull request.

### DIFF
--- a/master/buildbot/test/unit/test_www_hooks_github.py
+++ b/master/buildbot/test/unit/test_www_hooks_github.py
@@ -37,6 +37,7 @@ gitJsonPayload = """
   "before": "5aef35982fb2d34e9d9d4502f6ede1072793222d",
   "repository": {
     "url": "http://github.com/defunkt/github",
+    "html_url": "http://github.com/defunkt/github",
     "name": "github",
     "full_name": "defunkt/github",
     "description": "You're lookin' at it.",
@@ -84,6 +85,7 @@ gitJsonPayloadNonBranch = """
   "before": "5aef35982fb2d34e9d9d4502f6ede1072793222d",
   "repository": {
     "url": "http://github.com/defunkt/github",
+    "html_url": "http://github.com/defunkt/github",
     "name": "github",
     "full_name": "defunkt/github",
     "description": "You're lookin' at it.",
@@ -244,6 +246,7 @@ gitJsonPayloadEmpty = """
   "before": "5aef35982fb2d34e9d9d4502f6ede1072793222d",
   "repository": {
     "url": "http://github.com/defunkt/github",
+    "html_url": "http://github.com/defunkt/github",
     "name": "github",
     "full_name": "defunkt/github",
     "description": "You're lookin' at it.",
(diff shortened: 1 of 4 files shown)


anish:
@rutsky circleci needs another kicking

tardyp:
LGTM

anish:
I just tested this out with github enterprise vs github.com, for reasons unknown the url field does not behave the same way for both. For regular github, the domain is always api.github.com for pull requests but github.com for push requests, for GHE the domain is always the actual domain.

The html_url field on the other hand seems to behave consistently across the board. I'm gonna update to that instead.

tardyp:
@anish I've seen your IRC ping. not sure I understand. please let me know if you are happy with this. I am still not sure. I can only trust you on this as I can't test GHE

anish:
What I was saying was since the html_url key is lacking from sample json, I was expecting an error along the lines of "key not found". For reasons unknown, this showed as a mix of "no changes found", "excepted 0, found 2 changes", which made me think the issue was something else entirely.

This turned out to be a non-issue, adding the html_url keys allows all the tests to pass. I tested this on GHE (v2.3) and github on my side. Works for me in both cases. We actually upgraded to GHE 2.4 as of yesterday  and I plan to test with that asap.

Would appreciate if someone else could test this out as well, as the actual unit tests include a bunch of payloads that I don't know how to test. Based on the github provided [samples](https://developer.github.com/v3/activity/events/types/), my approach seems correct as well.

tardyp:
I think the best to let people try it more is to add a release note and doc saying that this has been tested against public github and GHE.
I let you add those, and I'll merge it!

anish:
Sounds good, will add that.

tardyp:
thanks!

PR #2042, opened 2016-03-11
Migrating GitHubStatusPush to new reporter API

sithhell:
This patch migrates the old GitHubStatusPush to the new reporter API
and reduces the number of needed parameters in favor of deducing the
parameters from the changeset and build


tardyp:
I'm very pleased somebody stepped up for this task!
And it looks good!

I would be delighted if on top of that unit tests were adapted :)
Please have a look at other test_reporters_*

and adapt/rename the one that is failing right now, and make travis unhappy:
buildbot/master/buildbot/test/unit/test_status_github.py


thank you so much!

sithhell:
Am 11.03.2016 5:16 nachm. schrieb "Pierre Tardy" <notifications@github.com>:
>
> I'm very pleased somebody stepped up for this task!
> And it looks good!
>
> I would be delighted if on top of that unit tests were adapted :)
> Please have a look at other test_reporters_*
>
> and adapt/rename the one that is failing right now, and make travis
unhappy:
> buildbot/master/buildbot/test/unit/test_status_github.py

Ha, totally missed the unit tests... Thanks for pointing out...

>
> thank you so much!

You are most welcome! I kind of needed it for my project... You can see it
in action here:

https://github.com/sithhell/hpx/pull/7

We are in the transition to nine, so I thought this might help!

tardyp:
btw, this is http://trac.buildbot.net/ticket/3285

anish:
FWIW, this did solve http://trac.buildbot.net/ticket/3489 for me

tardyp:
@sithhell , would you have time to create the unit tests?
I totally want this in next beta, which should get out as soon as @rutsky is done with the worker transition, and its nearly finished

master/buildbot/reporters/github.py
```diff
@@ -0,0 +1,158 @@
...
+            self.handledBuildIds.add(build['buildid'])
+
+        buildrequest = yield self.master.data.get(("buildrequests", build['buildrequestid']))
+        buildset = yield self.master.data.get(("buildsets", buildrequest['buildsetid']))
+        yield utils.getDetailsForBuilds(self.master, buildset, [build])
+        build['url'] = utils.getURLForBuild(self.master, build['builder']['builderid'], build['number'])
+
+        status = {}
+
+        props_dict = yield self.master.db.builds.getBuildProperties(build['buildid'])
```
tardyp: I would prefer not to use the db API. please use data api

anish: @sithhell  In this same block of code, needs to skip builds that haven't come from a webhook, else the whole thing crashes when parsing a non-existent sourcestamp. At my end I just skip if the scheduler is 'force' or one of the periodic schedulers and that seems to do the trick though I'm not sure if that's the best solution.

tardyp:
see https://github.com/buildbot/buildbot/pull/2057
for more example on how to use reporter and data api.

I dont think we should use Interpolate in those reporter plugins

master/buildbot/reporters/github.py
sithhell: @tardyp To be honest, I don't know how to achieve the same with the data API. The overall properties I need to interpolate here weren't there in the data API.

@anish good point! thanks!

sithhell:
@tardyp I think having the ability to use Interpolate for status reports has its benefits. Makes it so much more flexible. What is the downside?

sithhell:
I am still struggling with the unit tests ... sorry for making no progress on that front ... I am planning on pushing a simplified, but passing test later today.

master/buildbot/reporters/github.py
tardyp: getDetailsForBuild has an option to request population of Properties.

tardyp:
Hi @sithhell 
please have a look at:
https://github.com/buildbot/buildbot/pull/2057

This creates a base class for all the HttpStatusPush like. Stash can be seen as the simple example.
This also adds a set of tools to ease unit tests (as indeed its not very simple when you are new to the code to create such unit tests)

I believe the unit tests for http reporters are now very easy to do. Please let me know.

master/buildbot/reporters/github.py
```diff
@@ -0,0 +1,158 @@
...
+        self.handledBuildIds = set()
+
+    @defer.inlineCallbacks
+    def startService(self):
+        yield service.BuildbotService.startService(self)
+        startConsuming = self.master.mq.startConsuming
+
+        self._buildUpdatedConsumer = yield startConsuming(
+            self.sendGitHubStatus,
+            ('builds', None, 'update'))
```
tardyp: why didn't you use the "new" event?

tardyp:
@sithhell 
I've taken your code, and put it in #2057 

It is only unit tests, but I did not test it with github

@anish  I'd apreciate if you want to give it a try with your bot

master/buildbot/reporters/github.py
sithhell: Am 19.03.2016 12:03 nachm. schrieb "Pierre Tardy" <notifications@github.com
>:
>
> In master/buildbot/reporters/github.py:
>
> > +        self.context = context or
Interpolate('buildbot/%(prop:buildername)s')
> > +        self.startDescription = startDescription or 'Build started.'
> > +        self.endDescription = endDescription or 'Build done.'
> > +
> > +        self.github = GitHubAPI(oauth2_token=token, baseURL=baseURL)
> > +
> > +        self.handledBuildIds = set()
> > +
> > +    @defer.inlineCallbacks
> > +    def startService(self):
> > +        yield service.BuildbotService.startService(self)
> > +        startConsuming = self.master.mq.startConsuming
> > +
> > +        self._buildUpdatedConsumer = yield startConsuming(
> > +            self.sendGitHubStatus,
> > +            ('builds', None, 'update'))
>
> why didn't you use the "new" event?

For some reason or another, the new event didn't give me all the properties
in my case, all properties were empty.

anish:
@tardyp trying, will let you know as soon as possible

PR #2043, opened 2016-03-12
Install [docs] dependencies in setup.py

### LINKED ISSUE #3435: Documentation for Pushover reporter says "sends email"

adityadivekar03:
Remove the pip install command from the docs Makefile used for installing the sphinx distribution.
Add condition in the conf.py file to check if the [docs] dependencies are installed. If not, suitable error is displayed.

fixes #3435

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -132,7 +132,7 @@ install:
   # Install sphinx so that pylint will be able to import it
   - "[ $TESTS != lint ] || pip install sphinx==1.3.3"
   # TODO: Pin Sphinx version to workaround http://trac.buildbot.net/ticket/3408
-  - "[ $TESTS != docs ] || pip install sphinx==1.3.3"
+  - "[ $TESTS != docs ] || pip install -e ./master[docs]"
 
 before_script:
   # create real MySQL database for tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,7 +131,7 @@ install:
   - "[ $TESTS != lint ] || pip install pylint==1.1.0 pep8==1.5.7 flake8  astroid==1.3.8"
   # Install sphinx so that pylint will be able to import it
   - "[ $TESTS != lint ] || pip install sphinx==1.3.3"
-  # TODO: Pin Sphinx version to workaround http://trac.buildbot.net/ticket/3408
+  # Install docs dependencies for running the tests
   - "[ $TESTS != docs ] || pip install -e ./master[docs]"
 
 before_script:
--- a/master/docs/conf.py
+++ b/master/docs/conf.py
@@ -22,7 +22,7 @@ sys.path.append(os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
 try:
-    import sphinxcontrib.blockdiag  # noqa
+    import sphinxcontrib.blockdiag as _  # noqa
 except ImportError:
     raise RuntimeError("sphinxcontrib.blockdiag is not installed. "
         "Please install documentation dependencies with `pip install buildbot[docs]`")
--- a/master/docs/conf.py
+++ b/master/docs/conf.py
@@ -22,7 +22,7 @@ sys.path.append(os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
 try:
-    import sphinxcontrib.blockdiag
+    import sphinxcontrib.blockdiag  # noqa
 except ImportError:
     raise RuntimeError("sphinxcontrib.blockdiag is not installed. "
         "Please install documentation dependencies with `pip install buildbot[docs]`")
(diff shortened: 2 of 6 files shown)


tardyp:
please look at travis results:

docs/conf.py:25:5: F401 'sphinxcontrib' imported but unused
    ^
1       F401 'sphinxcontrib' imported but unused

adityadivekar03:
The import statement has been put in place to check if the user has the sphinxcontrib.blockdiag package installed. If not we need to raise an error, since we require it for documentation.
The import is thus used only for checking purposes. 
Will it be an issue to have an unused import ?

master/docs/conf.py
```diff
@@ -21,7 +21,18 @@
 sys.path.append(os.path.abspath('.'))
 
 # -- General configuration -----------------------------------------------------
-
+try:
+    import sphinxcontrib.blockdiag
```
rutsky: add `  # noqa` here to ignore flake8 linting errors (with two spaces before `#` sign):

```
    import sphinxcontrib.blockdiag  # noqa
```

.travis.yml
```diff
@@ -130,7 +130,7 @@ install:
   # Install sphinx so that pylint will be able to import it
   - "[ $TESTS != lint ] || pip install sphinx==1.3.3"
   # TODO: Pin Sphinx version to workaround http://trac.buildbot.net/ticket/3408
```
rutsky: Please move this comment (`TODO: Pin Sphinx ...`) to the corresponding line in `setup.py`.

rtd-requirements.txt
```diff
@@ -4,7 +4,4 @@
 -e master/
```
rutsky: Please, place `master[docs]` here, and remove whole paragraph below.

rutsky:
One more rebase needed...

Changes LGTM with the exceptions that I commented.
Good work, @adityadivekar03!

adityadivekar03:
Rebased and pushed :)
@rutsky  I apologize for the dragging of this issue for so long.

rutsky:
LGTM, lets wait for CI to finish and merge.

Thanks, @adityadivekar03 !

adityadivekar03:
The python 2.6 tests are failing it seems. I'm not able to understand the traceback. Do you have any ideas?

rutsky:
Yes, this is https://github.com/crossbario/autobahn-python/issues/614.
Lets ignore these py 2.6 errors for now.

oberstet:
@rutsky Yep, Autobahn does not any longer support Python 2.6 (we once did ..): https://github.com/crossbario/autobahn-python/blob/master/setup.py#L224 and https://github.com/crossbario/autobahn-python/blob/master/.travis.yml#L10

Is that a problem for BB?

rutsky:
> Is that a problem for BB?

I don't think so.

Buildbot still runs on Python  2.6, but to do so we already use quite old version of Twisted (Twisted also dropped support for 2.6 some time ago).
We will pin maximum autobahn version on Python 2.6, as we done with Twisted, so this is not a problem.

Honestly I don't think even Python 2.6 support is feasible, but while it's not (yet) time consuming I think we will support it.

rutsky:
Workaround for Python 2.6 (or whole Py 2.6 dropping) is discussed in #2048 .

adityadivekar03:
Seems both pylint and flake8 are raising import errors. 

How about we use a statement `assert sphinxcontrib-blockdiag` just to show that the import is being used in some way? We can place this statement in the try block.

rutsky:
flake8 doesn't report errors, only pylint does.

@adityadivekar03 can you try this?

```
    import sphinxcontrib.blockdiag as _  # noqa
```

I think pylint ignores unused `_` variable, e.g. here: https://github.com/buildbot/buildbot/blame/master/master/buildbot/test/__init__.py#L64

rutsky:
Also we could try to ignore unused imports for that specific line, like as done [here](http://stackoverflow.com/a/29562085/391865), but I'm not sure how to put both pylint's `# pylint: disable=unused-import` and flake8's `noqa` on the same line.

adityadivekar03:
I too came across the similar solution but couldn't figure out a way to include both the instructions in the same line. 
But using the statement `assert sphinxcontrib.blockdiag` solved the issue. Now it is no more an unused import. That should solve it :)

rutsky:
Personally I would prefer to be explicit: setting `# noqa` or `# pylint: ...` is explicit declaration that you intent to ignore some warning, and `assert` should be used for assertions.

But your approach is ok, since pylint + flake8 doesn't give us much choices, so lets merge it.

Thanks, @adityadivekar03 !

adityadivekar03:
Yes, I'll keep that practice in mind. 
@rutsky I should say thanks to you for being tolerant for so long!

master/docs/conf.py
tardyp: @rutsky what we usually do in the code is  
```
[sphinxcontrib.blockdiag]
```

problem with assert is that it evaluates the Trueness of the thing which could actually do some side effects or even return false

PR #2046, opened 2016-03-14
Install correct version in bundle

Singletoned:
Currently the instructions in the first run tutorial don't work as you have to install more packages than just buildbot in order to be able to run buildbot.  I've updated the [bundle] extras option to include the correct version of these.  I've also updated the docs to reflect the new command.

Unfortunately the docs are still slightly wrong, as you need to type `./bin/pip install buildbot[bundle]==0.9.0b7` at the moment (because pip still defaults to 0.8), but I didn't want to hardcode a version number in the docs, and I could work out how to get a variable to work in Sphinx (eg `./bin/pip install buildbot[bundle]=={{version}}`).

### DIFF
--- a/master/docs/tutorial/firstrun.rst
+++ b/master/docs/tutorial/firstrun.rst
@@ -65,7 +65,7 @@ Now that we are ready, we need to install buildbot:
 
 .. code-block:: bash
 
-  ./bin/pip install buildbot
+  ./bin/pip install buildbot[bundle]
 
 Now that buildbot is installed, it's time to create the master:
 
--- a/master/setup.py
+++ b/master/setup.py
@@ -425,10 +425,10 @@ else:
             'pyflakes',
         ],
         'bundle': [
-            "buildbot-www=={}".format(bundle_version),
-            "buildbot-slave=={}".format(bundle_version),
-            "buildbot-waterfall-view=={}".format(bundle_version),
-            "buildbot-console-view=={}".format(bundle_version),
+            "buildbot-www=={0}".format(bundle_version),
+            "buildbot-slave=={0}".format(bundle_version),
+            "buildbot-waterfall-view=={0}".format(bundle_version),
+            "buildbot-console-view=={0}".format(bundle_version),
         ],
         'tls': [
             'Twisted[tls] ' + twisted_ver,
--- a/master/setup.py
+++ b/master/setup.py
@@ -396,6 +396,8 @@ if sys.version_info[:2] == (2, 6):
 else:
     twisted_ver = ">= 12.1.0"
 
+bundle_version = version.split("-")[0]
+
 try:
     # If setuptools is installed, then we'll add setuptools-specific arguments
     # to the setup args.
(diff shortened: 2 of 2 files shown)


rutsky:
LGTM

rutsky:
@Singletoned thanks!

Perhaps you have other remarks about first run experience?

Singletoned:
@rutsky I've gone through the 0.8 and the 0.9 ones, and they seem good to me.

Between 0.8 and 0.9, you seem to have changed to cd'ing into the virtualenv and providing the commands as `./bin/buildbot create-master master`.  I didn't like this change as I use centrally managed virtualenvs (using virtualenvwrapper) and had to be careful not to copy the leading `./bin/` each time.  I would prefer just putting `buildbot create-master master` in the documentation as it is simpler and more intuitive.

Singletoned:
Another change from 0.8 is that the port that the web interface runs on is 8020 rather than 8010, but the docs still say 8010.  I can fix it easily enough, but I'm not sure whether to fix it in the code or in the docs?

rutsky:
Hm. I don't  like `./bin/pip` either.

Looks like we also create master and worker inside virtualenv directory in docs, which sound horrible to me: virtualenv directory should be opaque and touching their internals should be minimum.
Also these instructions will not work on Windows, because on windows `pip` is inside `Scripts/pip`, not `bin/pip`.

I opened ticket for this issue: http://trac.buildbot.net/ticket/3483

rutsky:
> Another change from 0.8 is that the port that the web interface runs on is 8020 rather than 8010, but the docs still say 8010. I can fix it easily enough, but I'm not sure whether to fix it in the code or in the docs?

Can you open PR for this one (with solution you think is correct, e.g. by fixing docs)?

It should be one-liner change and we can discuss this issue with other devs in that PR.

tardyp:
I like the fact to use the same version for the bundle.

FWIW, I usually recommend to use pip install --pre for beta version.
You actually need it if you have a recent version of pip.

To my knowledge, pip will refuse to install beta deps if you dont use --pre

rutsky:
`--pre` looks better than specifying exact versions of packages, thanks for the info, Pierre!

Not sure if we always want to have exact same versions: do we always release *all* components at the same time? What if there will be patch-level releases of some components?

I vote for using `pip install --pre buildbot[bundle]` in docs (until 0.9.0 will be released).

As for synchronized versions of components, perhaps we can come up with more smart constraints than exact version match?

rutsky:
Also, @Singletoned, tests are failing on Python 2.6: https://travis-ci.org/buildbot/buildbot/jobs/115849350
Looks like some problem with version string.

tardyp:
At least for the beta, where we got big incompatibilities bw version, its better to keep things to the same version.

tardyp:
we do release all packages at the same time, pointing to the same tag

rutsky:
Then lets pin all bundle packages versions to the master version, as @Singletoned done currently.

If we pin them, then there is no need to use `--pre` for beta versions, and when there will released 0.9.0 there will be no need to use `--pre` too.

So +1 for current solution with the exception, that it fails on Python 2.6 — @Singletoned, please take a look.

tardyp:
:+1:

rutsky:
LGTM

tardyp:
@Singletoned can you confirm the firstrun.rst is now correct for a complete newbie?

if so, we can close http://trac.buildbot.net/ticket/3324

Singletoned:
Yes, I think it is.

PR #2048, opened 2016-03-15
Fail with ConfigError without UnboundLocalError when parsing volumes for DockerLatentWorker fails

anish:
When volumes configuration has an error in it, loop does not break
correctly. A ConfigError is returned but the  succeeding code actual
fails with an UnboundLocalError as well since volume and bind were
never initialised.

Failure before this change :
```
Restarting buildmaster "FRB Buildmaster"
            self.volumes.append(volume)
        exceptions.UnboundLocalError: local variable 'volume' referenced
before assignment

2016-03-15 21:27:27+0000 [-] Configuration Errors:
2016-03-15 21:27:27+0000 [-]   Invalid volume definition for docker
/tmp/t/tmp/t. Skipping...
2016-03-15 21:27:27+0000 [-]   error while parsing config file: local
variable 'volume' referenced before assignment (traceback in logfile)
2016-03-15 21:27:27+0000 [-] Halting master.
```
Failure after this change :
```
Restarting buildmaster "FRB Buildmaster"
2016-03-15 21:27:19+0000 [-]
2016-03-15 21:27:19+0000 [-] Configuration Errors:
2016-03-15 21:27:19+0000 [-]   Invalid volume definition for docker
/tmp/t/tmp/t. Skipping...
2016-03-15 21:27:19+0000 [-] Halting master.
```

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -79,6 +79,7 @@ class DockerLatentWorker(AbstractLatentWorker):
             except ValueError:
                 config.error("Invalid volume definition for docker "
                              "%s. Skipping..." % volume_string)
+                continue
             self.volumes.append(volume)
 
             ro = False


master/buildbot/worker/docker.py
```diff
@@ -79,6 +79,7 @@ def __init__(self, name, password, docker_host, image=None, command=None,
             except ValueError:
                 config.error("Invalid volume definition for docker "
                              "%s. Skipping..." % volume_string)
+                break
```
rutsky: Perhaps `continue` will be better?
User will be able to see all incorrectly formatted volumes.

anish: @rutsky I had continue first, but I think failing at the point of parsing would be better :)  will switch

tardyp:
:+1:

rutsky:
New version of autobahn fails on Python 2.6: https://travis-ci.org/buildbot/buildbot/jobs/116316560#L1735

I reported issue about this: https://github.com/crossbario/autobahn-python/issues/614

rutsky:
@tardyp, @djmitche, autobahn-python dropped Python 2.6 support in 0.13.0.

Lets make [hack](https://github.com/buildbot/buildbot/blame/master/master/setup.py#L393) similar to Twisted version handling in setup.py?

djmitche:
I think that's a good idea.

rutsky:
@anish can you add this workaround for autobahn-python on Python 2.6?

tardyp:
Looks like python 2.6 support is starting to be quite expensive for us. How
many pr have to be reworked because of format {}?
Now we have 2 of our base libraries not supporting it anymore. Plus this
has a cost in term of CI time.

Le mer. 16 mars 2016 14:47, Vladimir Rutsky <notifications@github.com> a
écrit :

rutsky:
> Looks like python 2.6 support is starting to be quite expensive for us. How many pr have to be reworked because of format {}?

Rework Buildbot pull requests --- none, we just need to add workaround for this issue and merge in master (then rerun Travis tests on failed PR, currently there are 3 failing PR due to this issue).

Personally I would like to drop 2.6. Is there are any reasons to support 2.6?
Nine branch is a good place to remove deprecated stuff.

anish:
@rutsky fixed in #2053 for now even if python2.6 support is dropped later on

anish:
@rutsky Now that #2053 is merged, is more work still needed ?

tardyp:
thanks!

PR #2050, opened 2016-03-16
fix error messages in forcescheduler ui

tardyp:
dataservice returns the actually error object directly, unlike restangular


jaredgrubb:
I dont know this all that well, but LGTM :)

PR #2051, opened 2016-03-16
Make legible the hello.cfg example.

ava-dylang:
Note, **super untested**, I don't have a working buildbot to try this with ATM.

General code cleanup; get rid of single variables names, don't use * imports, use functions to reduce duplicated code, a focus on showing the user the meaningful part of the example.

I assume 'BuildmasterConfig' is some magic namespace that must be declared, so I excluded it from pylint's checks.

### DIFF
--- a/master/docs/examples/hello.cfg
+++ b/master/docs/examples/hello.cfg
@@ -1,72 +1,71 @@
 # -*- python -*-
 # ex: set syntax=python:
 
-from buildbot.plugins import *
+from buildbot.plugins import buildslave
+from buildbot.plugins import changes
+from buildbot.plugins import util
+from buildbot.plugins import steps
 
-BuildmasterConfig = c = {}
+SVN_REP = "file:///usr/home/warner/stuff/Projects/BuildBot/demo/SVN-Repository"
+CVS_REP = "/usr/home/warner/stuff/Projects/BuildBot/demo/Repository"
+DARCS_REP = "http://localhost/~warner/hello-darcs"
 
-c['slaves'] = [buildslave.BuildSlave("bot1", "sekrit")]
 
-c['change_source'] = changes.PBChangeSource(prefix="trunk")
-c['builders'] = []
+def _standard_build(checkout_step, test_step=None):
+    if test_step is None:
+        test_step = steps.Test(command=["make", "check"])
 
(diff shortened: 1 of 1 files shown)


tardyp:
I am not sure about this.

To me as an experienced python developer, its much more readable, indeed.
but remember our target audience here is more inexperienced python users.

I think this is why all our master.cfg are written this way.

re 'import *': this was in the design from @sa2ajj to write it that way as a mean to simplify usage of plugins.

I'll let other comment on this

master/docs/examples/hello.cfg
```diff
@@ -1,72 +1,71 @@
 # -*- python -*-
 # ex: set syntax=python:
 
-from buildbot.plugins import *
+from buildbot.plugins import buildslave
+from buildbot.plugins import changes
+from buildbot.plugins import util
+from buildbot.plugins import steps
```
marius-rizac: what about
```python
from buildbot.plugins import buildslave, changes, util, steps
```

tardyp: well the whole point was to just have everything in the global namespace, so that people can just copy paste samples from the doc, and not worry about what to import.

ava-dylang: Preface: You're all welcome to modify and integrate these changes however you like.

As to why I write imports this way:
@marius-rizac Because Google's Style Guidelines require [each import have its own line](https://google.github.io/styleguide/pyguide.html#Imports_formatting)
@tardyp Because Google's Style Guidelines [implicitly prohibit them](https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports) (Only concrete imports are allowed) and [star imports do bad things](http://stackoverflow.com/q/2386714/1695680). Not to mention that you can still copy-paste this example and it still "just-works":tm:, I consider there to be _less_ to worry about importing when what to import is actually documented. :bow:

sa2ajj: "Google's Style Guidelines" are not exactly followed here.

What is followed is [PEP-8](https://www.python.org/dev/peps/pep-0008/) with Twisted twisted names for methods and functions.

sa2ajj:
I am sorry, but we will not accept this PR.

PR #2053, opened 2016-03-16
Lock down autobahn version to < 0.13.0 when used wwith python2.6

anish:


rutsky:
LGTM, however we will probably just drop Python 2.6 support: http://trac.buildbot.net/ticket/3488

anish:
@rutsky I assumed as much, but this would help move the CI along till that decision is made. (And I am using 2.6 locally right now)

rutsky:
I would appreciate if you post you opinion on dropping Python 2.6 support in the ticket, since you are Python 2.6 user: what are the reasons you use Python 2.6? can you upgrade to Python 2.7/how hard it will be for your environment?

anish:
We're running RHEL6 boxes. The build environment on these is being recreated using containers. While the base system might move up from python2.6, the build environment might take longer to move up. Would be nice to have the ability to run against 2.6, even if it is unsupported.

tardyp:
thanks!

PR #2055, opened 2016-03-17
add URLForBuild util

tardyp:


master/buildbot/reporters/utils.py
```diff
@@ -155,3 +156,11 @@ def getURLForBuild(master, builderid, build_number):
     return prefix + "#builders/%d/builds/%d" % (
         builderid,
         build_number)
+
+
+@renderer
+@defer.inlineCallbacks
+def URLForBuild(props):
+    build = props.getBuild()
+    builder_id = yield build.builder.getBuilderId()
```
Stibbons: I suggest to add an accessor in the base class, so any inherited class can easily do ``self.getUrlForBuild()``.

tardyp:
if no other remarks, I will merge this trivial change

PR #2056, opened 2016-03-17
Optimise the recent changes page

tardyp:
when change list is becoming large, its very slow to use the soft pagination
As we have a recent change db api. We use it.

Of course we really should do something much nicer with automatic generation of
the sqlalchemy query, but this is *much* more work, so.. post 0.9.0

### DIFF
--- a/master/buildbot/data/changes.py
+++ b/master/buildbot/data/changes.py
@@ -78,8 +78,13 @@ class ChangesEndpoint(FixerMixin, base.Endpoint):
             else:
                 changes = []
         else:
-            changes = yield self.master.db.changes.getChanges()
-
+            # this special case is useful and implemented by the dbapi
+            # so give it a boost
+            if all((resultSpec.limit, resultSpec.offset is None,
+                    resultSpec.order == ['-changeid'])):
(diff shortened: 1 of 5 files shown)


master/buildbot/data/changes.py
```diff
@@ -78,8 +78,13 @@ def get(self, resultSpec, kwargs):
             else:
                 changes = []
         else:
-            changes = yield self.master.db.changes.getChanges()
-
+            # this special case is useful and implemented by the dbapi
+            # so give it a boost
+            if all((resultSpec.limit, resultSpec.offset is None,
+                    resultSpec.order == ['-changeid'])):
```
jaredgrubb: That seems much harder to read than:

if resultSpec.limit and resultSpec.offset is None and resultSpec.order == ['-changeid']):

jaredgrubb:
One minor comment .. Good though!

master/buildbot/data/changes.py
tardyp: yeah is because pep8 does not like if when they are too long (cause the indent is not distinguishable from the if content)

tardyp: looks like the pep8 bug is actually fixed now. switch to the more logical expression

PR #2058, opened 2016-03-18
master/steps: add a CMake step

sa2ajj:
This step provides a convenience interface for `cmake` command.

### DIFF
--- a/master/buildbot/steps/cmake.py
+++ b/master/buildbot/steps/cmake.py
@@ -1,0 +1,84 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+from future.utils import iteritems
+
+from twisted.internet import defer
+
+from buildbot import config
+from buildbot.process.buildstep import LoggingBuildStep, ShellMixin
+
+
+class CMake(ShellMixin, LoggingBuildStep):
+    DEFAULT_CMAKE = 'cmake'
+
+    name = 'cmake'
+    description = ['running', 'cmake']
+    descriptionDone = ['cmake']
+
+    renderables = (
+        'cmake',
+        'definitions',
+        'generator',
+        'options',
+        'path'
+    )
+
+    haltOnFailure = True
+
+    def __init__(self, path=None, generator=None, definitions=None,
+                 options=None, cmake=DEFAULT_CMAKE, **kwargs):
+
+        self.path = path
+        self.generator = generator
+
+        if not (definitions is None or isinstance(definitions, dict)):
+            config.error('definitions must be a dictionary')
+        self.definitions = definitions
+
+        if not (options is None or isinstance(options, (list, tuple))):
+            config.error('options must be a list or a tuple')
+        self.options = options
+
+        self.cmake = cmake
+
+        super(CMake, self).__init__(**kwargs)
+
+    @defer.inlineCallbacks
+    def run(self):
+        """
+        run CMake
+        """
+        command = [self.cmake]
+
+        if self.generator:
+            command.extend([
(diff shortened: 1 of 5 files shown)


tardyp:
looks good so far!

master/docs/manual/cfg-buildsteps.rst
```diff
@@ -1240,8 +1267,65 @@ The arguments are identical to :bb:step:`ShellCommand`.
...
+
+In addition to the parameters :bb:step:`ShellCommand` supports, this step accepts the following parameters:
+
+``path``
+    Either a path to a source directory to (re-)generate a build system for it in the current working directory.
+    Or an existing build directory to re-generate its build system.
+
+``generator``
+    A build system generator.
+    See `cmake-generators(7) <https://cmake.org/cmake/help/v3.5/manual/cmake-generators.7.html>`_ for available options.
```
Stibbons: Why would you run the generator in a CI ?

sa2ajj: I am not sure I understood the question: in CMake a generator is what produces the actual build system.  In my day job, it's Ninja.

Could you please clarify what you mean?

Stibbons: My mistake. I remember the way of using cmake and wasn't traditional...

I mean, there might be several strategy around cmake:
- indeed, you might want to run the generator in the ci to product the build system files (Makefile, Ninja,...). In the CI, you need to run the generator and then the build system. => for that, don't you want to add a way to run the build system automatically, or you want user to add another "build" step that would run the ninja command?
- in my case, we generated the makefiles once, committed them, and made heavy modifications on it, so we didn't run the generator in the CI... but, yes, it's not the canonic way of working with cmake !

sa2ajj:
Looks like I need to spend a bit more time on figuring out some class dependencies :smile:

master/docs/manual/cfg-buildsteps.rst
sa2ajj: Our  case is that we *require* users to first run CMake and then run ninja:
```
$ git clone remote:repo
$ cd repo
$ mkdir build
$ cd build
$ cmake -G Ninja .. <other parameters>
$ ninja
$ ninja unittest.run
```

We have 6 different target environments (4 hardware + 2 desktop) + a number of options that control, for example, static analysis target generation in `build.ninja` file.

All this somehow makes us feel that adding all those pre-configured environments does not make sense.

sa2ajj:
Any suggestions on negative test cases?

master/buildbot/test/unit/test_steps_cmake.py
```diff
@@ -0,0 +1,132 @@
...
+        self.expectOutcome(result=SUCCESS)
+        return self.runStep()
+
+    def test_definitions(self):
+        definition = {
+            'a': 'b'
+        }
+        self.setupStep(CMake(definitions=definition))
+        self.expect_command('-D%s=%s' % definition.items()[0])
+
```
tardyp: you are supposed to run your step!

return self.runStep()

tardyp:
not sure what you mean by negative test cases.

master/buildbot/test/unit/test_steps_cmake.py
sa2ajj: I should rename the function name since it's 'expect **and** run`

sa2ajj:
any suggestions of what **wrong** situations test cases should try?

tardyp:
LGTM. Testing the config parameters is good enough for me.

You got 100% hit on the patch

rutsky:
@sa2ajj I see that you reformatted some RST files, do you use some tools do this or do you reformat them manually?

sa2ajj:
@rutsky I do it manuall :)  I use ReST quite a lot and these kinds of problems stand out for me and my fingers automatically fix them :)

master/buildbot/steps/cmake.py
```diff
@@ -0,0 +1,88 @@
...
+            for item in iteritems(self.definitions):
+                command.append('-D%s=%s' % item)
+
+        if self.options is not None:
+            command.extend(self.options)
+
+        cmd = yield self.makeRemoteShellCommand(
+            command=command,
+            collectStdout=True,
+            collectStderr=True
```
djmitche: collect* mean that these outputs will be stored in memory, and I imagine they'll be quite large -- but they appear not to be used.  Is this necesary?

djmitche:
This looks good to me, with that one question answered.  I didn't look to hard at most of the .rst except the bit about the new step.

master/buildbot/steps/cmake.py
sa2ajj: Oh, in this case it's not really necessary.

I need to check the docs about those options and this note if it's not mentioned.

sa2ajj: Ok, it's somewhat mentioned :/

I updated this branch to not use these parameters.

PR #2060, opened 2016-03-19
Revert "revert-me: disable sqlite FK"

Rjaylyn:
Reverts buildbot/buildbot#2054


rutsky:
@Rjaylyn perhaps I'm missing something, but looking at http://trac.buildbot.net/ticket/3490 I see that it's not yet resolved, so this workaround should stay.
Why do you revert it?

tardyp:
I suspect this is an automated account which reverts reverts-me automatically

PR #2062, opened 2016-03-21
Remove isdir attribute from base change_hook for www endpoint

rayalan:
- isdir was removed elsewhere previously
- Updated unit tests to make sure attribute is not provided by any hook

--
See http://trac.buildbot.net/ticket/3494 for corresponding ticket. Note that this review doesn't address any lingering problems in the CSV mailer from isdir being set.


tardyp:
Looks good to me. 

I was not aware of cvs mailer! looks like crazy legacy stuff.

I would rather remove this old mailer code. It can still be unburied later if somebody cares.

PR #2063, opened 2016-03-21
model.py: restrict index lengths to 255

yetanotherion:
mysql/InnoDB tables have a limitation on index size,
regarding string elements.

http://dev.mysql.com/doc/refman/5.7/en/create-index.html
"...a prefix can be up to 767 bytes long for InnoDB tables..."

Some indexed attributes used to overflow that limit
on utf8 databases. This commit reduces the length of those
from 256 to 255 to fit that constraint.

### DIFF
--- a/master/docs/developer/database.rst
+++ b/master/docs/developer/database.rst
@@ -1840,15 +1840,13 @@ Index Length in MySQL
 
 .. index:: single: MySQL; limitations
 
-MySQL only supports about 330-character indexes.  The actual index length is
+MySQL only supports about 330-character indexes. The actual index length is
 1000 bytes, but MySQL uses 3-byte encoding for UTF8 strings.  This is a
 longstanding bug in MySQL - see `"Specified key was too long; max key
 length is 1000 bytes" with utf8 <http://bugs.mysql.com/bug.php?id=4541>`_.
 While this makes sense for indexes used for record lookup, it limits the
 ability to use unique indexes to prevent duplicate rows.
 
-InnoDB has even more severe restrictions on key lengths, which is why the MySQL
-implementation requires a MyISAM storage engine.
 
 Transactions in MySQL
 ~~~~~~~~~~~~~~~~~~~~~
--- a/master/docs/developer/database.rst
+++ b/master/docs/developer/database.rst
@@ -1847,6 +1847,8 @@ length is 1000 bytes" with utf8 <http://bugs.mysql.com/bug.php?id=4541>`_.
 While this makes sense for indexes used for record lookup, it limits the
 ability to use unique indexes to prevent duplicate rows.
 
+InnoDB only supports indexes up to 255 unicode characters, which is why
+all indexed columns are limited to 255 characters in Buildbot.
 
 Transactions in MySQL
 ~~~~~~~~~~~~~~~~~~~~~
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,12 +22,13 @@ env:
   - TWISTED=15.4.0 SQLALCHEMY=latest TESTS=trial
   - TWISTED=latest SQLALCHEMY=latest TESTS=trial
   # Configuration when SQLite database is persistent between running tests
-  # (by default in other tests in-memory SQLite database is used which is 
+  # (by default in other tests in-memory SQLite database is used which is
   # recreated for each test).
   # Helps to detect issues with incorrect database setup/cleanup in tests.
   - TWISTED=latest SQLALCHEMY=latest TESTS=trial BUILDBOT_TEST_DB_URL=sqlite:////tmp/test_db.sqlite
   # Configuration that runs tests with real MySQL database
   - TWISTED=latest SQLALCHEMY=latest TESTS=trial BUILDBOT_TEST_DB_URL=mysql+mysqldb://travis@127.0.0.1/bbtest
+  - TWISTED=latest SQLALCHEMY=latest TESTS=trial BUILDBOT_TEST_DB_URL=mysql+mysqldb://travis@127.0.0.1/bbtest&storage_engine=InnoDB
   # Configuration that runs tests with real PostgreSQL database with pg8000 and psycopg2 drivers
   - TWISTED=latest SQLALCHEMY=latest TESTS=trial BUILDBOT_TEST_DB_URL=postgresql+psycopg2:///bbtest?user=postgres
   - TWISTED=latest SQLALCHEMY=latest TESTS=trial BUILDBOT_TEST_DB_URL=postgresql+pg8000:///bbtest?user=postgres
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -141,7 +141,7 @@ class BuildbotEngineStrategy(strategies.ThreadLocalEngineStrategy):
 
         kwargs['pool_recycle'] = int(u.query.pop('max_idle', 3600))
 
-        # default to the MyISAM storage engine; InnoDB is not supported
+        # default to the MyISAM storage engine
         storage_engine = u.query.pop('storage_engine', 'MyISAM')
         kwargs['connect_args'] = {
             'init_command': 'SET default_storage_engine=%s' % storage_engine,
--- a/master/buildbot/db/migrate/versions/046_mysql_innodb_compatibility.py
+++ b/master/buildbot/db/migrate/versions/046_mysql_innodb_compatibility.py
@@ -84,6 +84,16 @@ def upgrade(migrate_engine):
             _has_incompatible_object_state(metadata, migrate_engine),
             _has_incompatible_users(metadata, migrate_engine)]):
         raise ValueError('cannot upgrade due to invalid data')
+    if migrate_engine.dialect.name == 'postgresql':
+        # Sql alchemy migrate does not apply changes on postgresql
+        def reduce_table_column_length(table, column):
+            return 'ALTER TABLE {} ALTER COLUMN {} TYPE character varying(255)'.format(table, column)
+        for table, columns in {'changes': ['author', 'branch', 'revision', 'category'],
+                               'object_state': ['name'],
+                               'users': ['identifier']}.items():
+            for column in columns:
+                migrate_engine.execute(reduce_table_column_length(table, column))
+        return
 
     changeset.alter_column(
         sa.Column('author', sa.String(255), nullable=False),
--- a/master/buildbot/db/migrate/versions/046_mysql_innodb_compatibility.py
+++ b/master/buildbot/db/migrate/versions/046_mysql_innodb_compatibility.py
@@ -1,0 +1,122 @@
+# This file is part of Buildbot. Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+import sqlalchemy as sa
+from sqlalchemy.sql import func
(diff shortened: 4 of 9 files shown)


tardyp:
Hi Ion,
thanks for working on innodb, it seems like a often requested feature.

There is a need for a migration script, which does all the alter table

djmitche:
If this is really all the change that's required, that's great -- let's do it!

Maybe the migration script should fail if there are any rows with exactly 256 characters?  Then users can rename or delete before re-running the upgrade.

djmitche:
As I recall, last time I looked at this, there was an index with *two* 256-character columns in it, which still wouldn't fit even at 255 characters each.  I don't recall what the index or columns were (this was years ago), but apparently they're gone now.

yetanotherion:
@djmitche I tried to implement the check in the migration script. However, even when the migration fails due to columns of length 256, the buildbot upgrade-master did not generate any log in my environment. How is the error usually pinpointed ? (right now I only checked the migrate_version was stuck in the previous version).

Regarding the two 256 character columns/index, I did not find any issue at the creation of the database. Do you recall the name/table having this configuration ?

tardyp:
hey @yetanotherion there is a pretty much powerful unit test framework for migration scripts. You may want to try that.

yetanotherion:
@tardyp unit tests are available in the last commit

tardyp:
looks good!

Did the unit test helped on your issue with failing migration handling?

yetanotherion:
Well, I won't say they didn't help, but I still don't know how the user making "buildbot upgrade-master" will see that he needs to change some 'changes, users, object_state having given changeids, objectid, uid' before trying again.

tardyp:
You could gather the issues rather than logging them, and put the in the exception.

I tend to agree that for me it is simpler for everybody if the migration script just truncates the data automatically.

maybe add an option via environment variable:

if you want to automatically truncate your data, please set the environment variable "FORCE_TRUNCATE_DATA=true" before calling upgrade-master

yetanotherion:
Let me rephrase my issue:
for now I do not see none of log.msg message, neither anyone inside ValueError, during the non unit test. Just to be sure, did any of you saw the output of an error after a 'buildbot upgrade-master' ?

tardyp:
Indeed, looks like a bug in upgrade-master. I am not sure if @rutsky saw the same issue?

rutsky:
> Indeed, looks like a bug in upgrade-master. I am not sure if @rutsky saw the same issue?

I haven't seen any issues with upgrade-master, but I use it rarely.
I saw issues with `start` command, but they shouldn't be critical (just wrong error messages).

yetanotherion:
I can propose you the following scenario to reproduce the issue:
- checkout the master branch
- create a database, you should have 45 in migrate_version
- download my pull request and add a ``raise ValueError('hey')`` just after the upgrade() in 
master/buildbot/db/migrate/versions/046_mysql_innodb_compatibility.py
- execute buildbot upgrade-master. 
In my environment I get

stdout:
``upgrading basedir
upgrading database (sqlite:///state.nine.sqlite)
upgrade complete
``
nothing in twisted.log

yetanotherion:
I just created and assigned http://trac.buildbot.net/ticket/3499 to myself (to fix it in another PR)

yetanotherion:
@tardyp @djmitche @rutsky the PR is now ok on Travis
https://travis-ci.org/buildbot/buildbot/builds/118066293

djmitche:
I like it -- just needs a release note.

Stibbons:
Just truncate without failing. What about the current inserts in actual builds? They would be truncated without asking the user what he wishes. So just truncate to 255 no matter the size of existing rows.

yetanotherion:
@stibbons there isn't any build related information in the truncated columns
https://github.com/buildbot/buildbot/pull/2063/commits/50f5b5f9a459c4c5c728e956cdf8d03eeccfc75a

I personally find Dustin's suggestion to be quite wise, at least that's the behavior I'd like as a user. 
(Note that the related code is already implemented in the PR)

Stibbons:
I mean, from now on, the new inserts will be truncated to 255 in normal builds, without warning the users (SQLAlchemy will just complain in a log, but accept the inserts)
So TMOH, this loss of data on previous builds is acceptable and should be the default, but indeed needs to be documented.
If we want to be sure to be able to revert this change, you need to back to a VARTEXT or BLOG backup column but that's ugly... in any case, user has no other choice that accep this truncate or do not upgrade this version

yetanotherion:
"in any case, user has no other choice that accep this truncate or do not upgrade this version"
I think we exactly agree on that. The code that checks for the possible truncate, 
permits to signal and let the user decide what she/he prefers, 
instead of taking the "truncate decision" for him/her.

yetanotherion:
@djmitche just to check I understood correctly, I should:
- create a master/docs/relnotes/0.9.0b8.rst, 
- add a section regarding the mysql innodb support, and 
- reference 0.9.0b8.rst in master/docs/relnotes/index.rst ?

rutsky:
@yetanotherion release notes should go to `master/docs/relnotes/index.rst`. They will be moved to `0.9.0b8.rst` or `0.9.0rc1` at the moment of a new release.

yetanotherion:
спасибо @rustky, I just uploaded a new commit for that.

rutsky:
@yetanotherion can you take a look at other places where InnoDB is mentioned, since you implemented support for InnoDB?

For example:

```
$ git grep -C 3 InnoDB                                                                                                                           
master/buildbot/db/enginestrategy.py-
master/buildbot/db/enginestrategy.py-        kwargs['pool_recycle'] = int(u.query.pop('max_idle', 3600))
master/buildbot/db/enginestrategy.py-
master/buildbot/db/enginestrategy.py:        # default to the MyISAM storage engine; InnoDB is not supported
master/buildbot/db/enginestrategy.py-        storage_engine = u.query.pop('storage_engine', 'MyISAM')
master/buildbot/db/enginestrategy.py-        kwargs['connect_args'] = {
master/buildbot/db/enginestrategy.py-            'init_command': 'SET default_storage_engine=%s' % storage_engine,
--
master/docs/developer/database.rst-While this makes sense for indexes used for record lookup, it limits the
master/docs/developer/database.rst-ability to use unique indexes to prevent duplicate rows.
master/docs/developer/database.rst-
master/docs/developer/database.rst:InnoDB has even more severe restrictions on key lengths, which is why the MySQL
master/docs/developer/database.rst-implementation requires a MyISAM storage engine.
master/docs/developer/database.rst-
master/docs/developer/database.rst-Transactions in MySQL
--
master/docs/manual/cfg-global.rst-
master/docs/manual/cfg-global.rst-MySQL defaults to the MyISAM storage engine, but this can be overridden with the ``storage_engine`` URL argument.
master/docs/manual/cfg-global.rst-
master/docs/manual/cfg-global.rst:Note that, because of InnoDB's extremely short key length limitations, it cannot be used to run Buildbot.
master/docs/manual/cfg-global.rst-See http://bugs.mysql.com/bug.php?id=4541 for more information.
master/docs/manual/cfg-global.rst-
master/docs/manual/cfg-global.rst-.. index:: Postgres
```

Also it would be nice to run tests on InnoDB engine on Travis.
You can add another driver configuration like done with PostgreSQL: https://github.com/buildbot/buildbot/blob/master/.travis.yml#L32
I believe you can specify `storage_engine` on SQLAlchemy URI: https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/enginestrategy.py#L145 (but please test that InnoDB is actually used).

yetanotherion:
@rutsky,
Regarding storage_engine=InnoDB, we're currently running quite heavy tests on a buildbot/master instance using MySQL/InnoDB tables. Based on these tests, I confirm you that: 
- the 'storage_engine' parameter is used, and that 
- the current PR permits to create and use InnoDB tables on the current buildbot database model, without having observed any errors during our tests.

master/docs/developer/database.rst
```diff
@@ -1840,15 +1840,13 @@ Index Length in MySQL
...
-MySQL only supports about 330-character indexes.  The actual index length is
+MySQL only supports about 330-character indexes. The actual index length is
 1000 bytes, but MySQL uses 3-byte encoding for UTF8 strings.  This is a
 longstanding bug in MySQL - see `"Specified key was too long; max key
 length is 1000 bytes" with utf8 <http://bugs.mysql.com/bug.php?id=4541>`_.
 While this makes sense for indexes used for record lookup, it limits the
 ability to use unique indexes to prevent duplicate rows.
 
-InnoDB has even more severe restrictions on key lengths, which is why the MySQL
-implementation requires a MyISAM storage engine.
```
rutsky: This sentence should rephrased, not removed, since this section describes MySQL limitations.

Something like: `InnoDB only supports key lengths up to 255 unicode characters, which is why all indexed columns are limited to 255 characters in Buildbot`.

yetanotherion:
@rutsky rebased & added InnoDB specific section.
FYI the previous travis build succeded (https://travis-ci.org/buildbot/buildbot/jobs/118212710)

rutsky:
Any idea why there are SQLite warnings in build that should use InnoDB?
https://travis-ci.org/buildbot/buildbot/jobs/118212722

tardyp:
@rutsky I guess all the unit tests does not use the BUILDBOT_DB_URL parameter

rutsky:
@tardyp Is it so? I thought `BUILDBOT_TEST_DB_URL` is exactly for overriding database in unit tests:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/util/db.py#L188

yetanotherion:
@rutsky correct me if I'm wrong, but when I browse other tests
- https://travis-ci.org/buildbot/buildbot/jobs/118212721, MySQL/MyIsam
- https://travis-ci.org/buildbot/buildbot/jobs/118212720: sqlite3
- ...
I see the same warnings

rutsky:
@yetanotherion yes, I think something has broken `BUILDBOT_TEST_DB_URL` usage either in Buildbot code, or on Travis.

I'm quite sure that I saw PostgreSQL related errors on Travis, so I think it worked before.

rutsky:
For example in build that was almost month ago there are only MySQL warnings: https://travis-ci.org/buildbot/buildbot/jobs/111442619 on MySQL configuration.

rutsky:
Ah, sorry, there *are* SQLite warnings even in tests with MySQL configuration: https://travis-ci.org/buildbot/buildbot/jobs/111442619#L853

Which is very suspicious, but not related to this PR.

tardyp:
Those sqlite warnings are fairly new to my knowledge. I think they are related to my sqlite PR, but I wasn't able to repro locally yet.

I would expect that BUILDBOT_TEST_DB_URL is only working for tests that actually use buildbot.test.util.db. I would expect some tests not to care about correctly initalizing the db for tests, and thus use sqlite. I am not sure this is a strong issue though.
Its important to run the db api, db migration, and integration tests with DB_URL, but the unit tests that incidentally *use* the db api, we can use the fastest implementation.

yetanotherion:
@rutsky 
I can't speak for all unit tests, but what could be a 'SQL migrate' bug was revealed by 
- https://travis-ci.org/buildbot/buildbot/builds/118039910, on postgresql
- when all the other tests were ok (besides flake8 https://travis-ci.org/buildbot/buildbot/builds/118039910)

PostgreSQL specific unit tests seems to still be functional, even though the sqlite3 warnings appear there too (https://travis-ci.org/buildbot/buildbot/jobs/118212714).

This seems coherent with
- your last sentence  "... very suspicious, but not related to this PR." and 
- @tardyp's root cause proposal.

rutsky:
@tardyp I agree, I didn't knew that some tests uses DB without `buildbot.test.util.db`. So SQLite warnings are OK at this moment.

rutsky:
> I can't speak for all unit tests, but what could be a 'SQL migrate' bug was revealed by 

@yetanotherion, great, thanks for pointing out on this.

Changes LGTM, lets wait for Travis to finish.

Thanks, @yetanotherion!

yetanotherion:
@rutsky the PR was correct. I added two commits to fix http://trac.buildbot.net/ticket/3499
in that PR (the bug was in the end very easy to fix)

rutsky:
@yetanotherion Haven't @tardyp fixed issue 3499 in #2068?

yetanotherion:
@rutsky indeed, commit reverted

yetanotherion:
@tardyp @rutsky the travis build passed
https://travis-ci.org/buildbot/buildbot/builds/118271404

tardyp:
:+1:  letting a little bit more time for other to get a chance to review

PR #2065, opened 2016-03-22
remove worker's "upgrade-slave" + rename "slave" in command line arguments

rutsky:


sa2ajj:
I *think* I understand the reason, though I'd really appreciate to have it listed in, e.g., the PR description.

rutsky:
Sorry, @sa2ajj, I haven't specify detailed description since this change were discussed on the IRC.

This is part of bug [2340](http://trac.buildbot.net/ticket/2340): "slave" to "worker" renaming.
In addition to command line arguments renaming, I remove "upgrade-slave" command, since its not (yet?) required in buildbot-worker.

It's intended that buildbot-worker will be installed on new systems, and *not* upgraded from buildbot-slave.
Users that don't want to reinstall buildbot-slave can and should still use it — it will work with nine master.
For new installations buildbot-worker should be used — buildbot-slave development will be frozen after release of buildbot-worker (though I expect that critical errors of buildbot-slave, if they will be any, still may be fixed in buildbot-slave).

Not implementing upgrade from buildbot-slave to buildbot-worker in buildbot-worker makes buildbot-worker completely "slave"-free, which is one of the goals of bug 2340.

If we really want to have some upgrade tool from buildbot-slave to buildbot-worker it can be implemented it as separate contrib — do we want it?

tardyp:
LGTM

PR #2066, opened 2016-03-22
rename worker Twisted Application name from 'buildslave' to 'buildbot-worker'

rutsky:
This is just a string, right?


sa2ajj:
LGTM

PR #2068, opened 2016-03-23
report errors in the master-upgrade command line

tardyp:
fix for http://trac.buildbot.net/ticket/3499 

* [x] tests

### DIFF
--- a/master/buildbot/scripts/upgrade_master.py
+++ b/master/buildbot/scripts/upgrade_master.py
@@ -138,7 +138,7 @@ def _upgradeMaster(config, master_cfg):
         yield upgradeDatabase(config, master_cfg)
     except Exception:
         e = traceback.format_exc()
-        print("problem while upgrading!:\n", e, file=sys.stderr)
+        print("problem while upgrading!:\n" + e, file=sys.stderr)
         defer.returnValue(1)
     else:
         if not config['quiet']:
--- a/master/buildbot/scripts/upgrade_master.py
+++ b/master/buildbot/scripts/upgrade_master.py
@@ -107,32 +107,32 @@ def upgradeDatabase(config, master_cfg):
 
 
 @in_reactor
-@defer.inlineCallbacks
 def upgradeMaster(config, _noMonkey=False):
     if not _noMonkey:  # pragma: no cover
         monkeypatches.patch_all()
 
     if not base.checkBasedir(config):
-        defer.returnValue(1)
-        return
+        return defer.succeed(1)
 
     os.chdir(config['basedir'])
 
     try:
         configFile = base.getConfigFileFromTac(config['basedir'])
     except (SyntaxError, ImportError) as e:
-        print("Unable to load 'buildbot.tac' from '%s':" % config['basedir'])
-        print(e)
-
-        defer.returnValue(1)
-        return
+        print("Unable to load 'buildbot.tac' from '%s':" % config['basedir'], sys.stderr)
+        print(e, sys.stderr)
+        return defer.succeed(1)
     master_cfg = base.loadConfig(config, configFile)
     if not master_cfg:
-        defer.returnValue(1)
-        return
+        return defer.succeed(1)
+    return _upgradeMaster(config, master_cfg)
 
-    upgradeFiles(config)
+
+@defer.inlineCallbacks
+def _upgradeMaster(config, master_cfg):
+
     try:
+        upgradeFiles(config)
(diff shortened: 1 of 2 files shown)


tardyp:
@yetanotherion was an easy one :)

master/buildbot/scripts/upgrade_master.py
```diff
@@ -106,34 +107,38 @@ def upgradeDatabase(config, master_cfg):
...
 @in_reactor
-@defer.inlineCallbacks
 def upgradeMaster(config, _noMonkey=False):
     if not _noMonkey:  # pragma: no cover
         monkeypatches.patch_all()
 
     if not base.checkBasedir(config):
-        defer.returnValue(1)
-        return
+        return defer.succeed(1)
```
rutsky: Isn't 

```
defer.returnValue(1)
return
```

will lead to `upgradeMaster()` deferred to return `1` and  your version:

```
return defer.succeed(1)
```

will lead to waiting of completed deferred and to return `None`?

yetanotherion: @rutsky I wonder whether the return below
defer.returnValue(1)
was ever executed

rutsky: AFAIK `defer.returnValue` raises exception, so no, `return` will not be executed. But in this way it's obvious even for non-Twisted developer (or static analysis tool), that function will return here, so I don't this this is a problem.

@tardyp ah, sorry, I missed that you *removed* inlineCallbacks, it's all OK!

master/buildbot/scripts/upgrade_master.py
```diff
@@ -106,34 +107,38 @@ def upgradeDatabase(config, master_cfg):
...
 
     try:
         configFile = base.getConfigFileFromTac(config['basedir'])
     except (SyntaxError, ImportError) as e:
-        print("Unable to load 'buildbot.tac' from '%s':" % config['basedir'])
-        print(e)
-
-        defer.returnValue(1)
-        return
+        print("Unable to load 'buildbot.tac' from '%s':" % config['basedir'], sys.stderr)
```
rutsky: maybe `file=sys.stderr`?

master/buildbot/scripts/upgrade_master.py
```diff
@@ -106,34 +107,38 @@ def upgradeDatabase(config, master_cfg):
...
+        print(e, sys.stderr)
```
rutsky: Maybe use `traceback.print_exc()`?

master/buildbot/scripts/upgrade_master.py
```diff
@@ -106,34 +107,38 @@ def upgradeDatabase(config, master_cfg):
...
-    if not config['quiet']:
-        print("upgrade complete")
+@defer.inlineCallbacks
+def _upgradeMaster(config, master_cfg):
+
+    try:
+        upgradeFiles(config)
+        yield upgradeDatabase(config, master_cfg)
+    except Exception as e:
+        print("problem while upgrading!:", e, file=sys.stderr)
```
rutsky: Perhaps we should print stack trace here too (in addition to the current message)?

master/buildbot/scripts/upgrade_master.py
yetanotherion: @rutsky I have the same understanding regarding defer.returnValue. I'm personally in favor of not adding return's below returnValue, for two reasons:
- to have less code
- so that "non-Twisted developers" ramp-up faster into "Twisted developers"

But that may be out of topic

tardyp:
addressed review comments.

master/buildbot/test/unit/test_scripts_upgrade_master.py
```diff
@@ -223,4 +223,6 @@ def test_upgradeDatabaseFail(self):
             mkconfig(basedir='test', quiet=True),
             config_module.MasterConfig())
         self.assertEqual(ret, 1)
-        self.assertIn("problem while upgrading!: o noz", sys.stderr.getvalue())
+        self.assertIn("problem while upgrading!:\n Traceback (most recent call last):\n",
```
rutsky: I wonder why there is space between `\n Traceback`? I think this test passes, so there should be space, but why?

tardyp: print adds space between things when they are separated by a colon.

yetanotherion:
LGTM

PR #2069, opened 2016-03-23
Nextworker error doc

gracinet:
Should fix http://trac.buildbot.net/ticket/3498

I did test on my experimental buildmaster (don't imagine there are unit tests for that), and could see the traceback

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -934,6 +934,8 @@ class BuilderConfig(util_config.ConfiguredMixin, WorkerAPICompatMixin):
         if nextWorker and (nextWorker.func_code.co_argcount == 2 or
                            (isinstance(nextWorker, MethodType) and
                             nextWorker.func_code.co_argcount == 3)):
+            warnDeprecated(
+                "0.9", "nextWorker now takes a 3rd argument (build request)")
             self.nextWorker = lambda x, y, z: nextWorker(x, y)  # pragma: no cover
         self.nextBuild = nextBuild
         if nextBuild and not callable(nextBuild):
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -848,7 +848,7 @@ class BuilderConfig(util_config.ConfiguredMixin, WorkerAPICompatMixin):
         if nextSlave is not None:
             reportDeprecatedWorkerNameUsage(
                 "'nextSlave' keyword argument is deprecated, "
-                "use 'workerbuilddir' instead")
+                "use 'nextWorker' instead")
             assert workerbuilddir is None
             nextWorker = nextSlave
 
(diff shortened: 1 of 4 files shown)


tardyp:
indeed this was bad..

:+1:

gracinet:
The AppVeyor build is puzzling, the test seems to expect 2 arguments for nextWorker, but I didn't change that (that's the point).

tardyp:
I think the test was wrong, but as the error were swallowed, nobody noticed..

gracinet:
ah yes, obviously you must be right, ok I'll update the test, then.

tardyp:
looks like the blame is @nam4dev (and myself)

The test is modifying the builderconfig with:

        self.bldr.config.nextWorker = nextWorker

but the actual support of the 2 args signature is made in the constructore for builderconfig (in config.py)

gracinet:
I'm a bit confused here, do you want to add a backwards compat for 2 args signature ?

I do reproduce the test errors (took me a while to remember how to use trial), so udating them for 3 arguments should be doable on my side.

gracinet:
ok, got it, indeed the intent it to provide compatibility, and I do like the idea not to pollute the actual distributor code with that detail. So these tests should go to BuilderConfig tests, right ?

Still I don't understand why this simple logging ends up as an ERROR test result, and I wouldn't actually enforce that errors get ignored to build no matter what as in test_nextWorker_fails if it were me. If someone actually wants to use this nextWorker to actually run the build only where it makes sense, that would defeat the purpose (but that's your call obviously).

tardyp:
Log.error are catched by the unit test framework unless you tell him it's
normal.

Le mer. 23 mars 2016 19:37, Georges Racinet <notifications@github.com> a
écrit :

tardyp:
@gracinet do you need more help here?

gracinet:
@tardyp, if you agree that I move the testing of the fallback to 2 args to BuilderConfig tests and remove the test that asserts that RuntimeError('xx') should be swallowed, no I'm fine

(got a few other matters to process this morning)

gracinet:
Finally, decided to adapted the existing tests (instead of replacing them with tests for BuilderConfig).
It may be better convincing that the change actually works

But since now `BuilderConfig,__init__` does some rewriting for `nextWorker`, it would probably be overall simpler if the case where `nextWorker` is `None` (replaced by a default impl) would also be treated there. This would make BRD slightly simpler to always assume there is a correct `nextWorker`.

gracinet:
ah, the doc change of [9ba538eb287fc3d44604184a42b6c60df9ed6e3b] has become unnecessary.
what about adding a deprecation warning instead ?

tardyp:
looks good!. Agree about the deprecation warning.

tardyp:
:+1:

PR #2070, opened 2016-03-24
Fix bug where build did not have 'got_revision' property

ryansydnor:
If a build does not have the "got_revision" property, we hit the "or" path which seems to have missed an update to the new helper function getProperty(build, "prop") from the older build.getProperty('prop').


ryansydnor:
Messed up author. Reopening.

tardyp:
thanks!

PR #2071, opened 2016-03-24
Fix crash caused by incorrectly appended volume strings

anish:
Docker volumes are supposed to be a list in the form of
[ '<path>:<bind>', '<path>:<bind>' ]

Current parsing code parses this correctly but zeros out self.volumes
and initializes it again in the form of [ '<path>', '<path>', '<path' ]

This breaks the parser if volumes is parsed again as it is expecting
 <path>:<bind> (a specific example is rebuild)

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -80,7 +80,7 @@ class DockerLatentWorker(AbstractLatentWorker):
                 config.error("Invalid volume definition for docker "
                              "%s. Skipping..." % volume_string)
                 continue
-            self.volumes.append(volume)
+            self.volumes.append(volume_string)
 
             ro = False
             if bind.endswith(':ro') or bind.endswith(':rw'):


tardyp:
looks good. I am not sure why CIs did not start

anish:
I think a combination of latent slave + volumes + rebuild is not tested by the CI

anish:
Fixes http://trac.buildbot.net/ticket/3500

rutsky:
Looks like Travis ran tests, but GitHub missed status report. And tests are failing: https://travis-ci.org/buildbot/buildbot/jobs/118164622

anish:
Will take care of that

anish:
Uhh, I hadn't fixed the unit tests for this, probably why it's breaking everything.

tardyp:
I read the thread too fast. I though that the test passed. the fix looked so obvious..

PR #2073, opened 2016-03-25
ignore spurious logs

tardyp:
Lots of logs are seen recently in travis about sqlite threads:

Exception closing connection <sqlite3.Connection object at 0x6a80810>
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/sqlalchemy/pool.py", line 290, in _close_connection
    self._dialect.do_close(connection)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/sqlalchemy/engine/default.py", line 426, in do_close
    dbapi_connection.close()
ProgrammingError: SQLite objects created in a thread can only be used in that same thread.The object was created in thread id 140587768477440 and this is thread id 140595138430720

git bisect shows:
https://github.com/buildbot/buildbot/commit/96f09167ecb592b5a4e04d457bdffc3d37a16f28
as the culprit, but actually if you remove, you will still see one instance of following

     No handlers could be found for logger "sqlalchemy.pool.SingletonThreadPool"

when executing:

    trial buildbot.test.unit.test_db_pool.Basic.test_do

I did go back up to 0.8.5, and still see this warning log for this test
so I conclude this warning can be safely ignored

Here is the git bisect method I used:

git bisect run bash check.sh

0 % cat check.sh
make rmpyc >/dev/null
trial buildbot.test.unit.test_db_pool.BasicWithDebug.test_do >/dev/null 2>err
if grep 'SQLite objects created in a thread can only be used in that same thread' err >/dev/null
then
echo bad
exit 1
else
echo good
exit 0
fi

### DIFF
--- a/master/buildbot/test/__init__.py
+++ b/master/buildbot/test/__init__.py
@@ -36,8 +36,19 @@ if sys.version_info[:2] < (3, 2):
     # WARNINGs and ERRORs will be printed).
     import logging
     _handler = logging.StreamHandler()
-    logging.getLogger().addHandler(_handler)
 
+    # Ignore following Exception logs:
+    #       Traceback (most recent call last):
+    #       File "[..]site-packages/sqlalchemy/pool.py", line 290, in _close_connection
+    #       self._dialect.do_close(connection)
+    #       File "[..]/sqlalchemy/engine/default.py", line 426, in do_close
+    #       dbapi_connection.close()
+    #       ProgrammingError: SQLite objects created in a thread can only be used in that same thread.
+    #       The object was created in thread id 123145306509312 and this is thread id 140735272824832
+    # sqlalchemy is closing pool connections from the main thread, which sqlite does not like
+    # the warning has been there since forever, but would be catched by the next lastResort logger
+    logging.getLogger("sqlalchemy.pool.SingletonThreadPool").addHandler(None)
+    logging.getLogger().addHandler(_handler)
 # import mock so we bail out early if it's not installed
 try:
     import mock


tardyp:
@yetanotherion @rutsky 
the mystery of the sqlite warnings is resolved..

rutsky:
@tardyp honestly I would like to see this warning actually resolved by closing connections from the proper thread. Our current behavior is not quite correct and SQLite warns us about it.
I even tried to fix this: spent few hours and failed — looks like it's just too hard to do it properly with current implementation.

Hopefully this warning shouldn't be an error in our case: AFAICT this warning should warn us about potential race condition if we would use SQLite connection from multiple thread *simultaneously*, which is not the case in Buildbot — we only close connection object from other thread, so this **LGTM**.

Other drawback of ignoring this warning is when we *will* make an error and try to use SQLite connection we will be not noticed about the issue. Also, there may be other useful warnings in `sqlalchemy.pool.SingletonThreadPool` handler.

tardyp:
Yes, I also spent some time trying to find a better fix. Actually, the proper fix would be in sqlalchemy, and it looks like they did not bother fix it (they just ignore the error + log something)

https://github.com/zzzeek/sqlalchemy/commit/1b0ac7e43b715b9050cb5f45d4a180bfc203c8a9


The connection close use case is for me not so much important.
- Its used for the unit tests, so we might end up eventually to lot of open files to old db.
- Its used at buildbot stop, but the file will eventually be closed in the

PR #2074, opened 2016-03-25
Fix unit tests breaking as a result of #2071

anish:
Fix got merged without the unit tests accomodating the change


anish:
@tardyp

tardyp:
thanks for quick fix! sorry for my mistake

PR #2075, opened 2016-03-27
Protect Interpolate rendering from change of _notHasKey constant

gracinet:
Should fix http://trac.buildbot.net/ticket/3505#ticket for Nine

It's not the only solution possible, but that one does not require rewriting much.
The very same fix also resolves the issue on 0.8.12, but the test has been developed on current master.


master/buildbot/test/unit/test_process_properties.py
```diff
@@ -408,6 +408,29 @@ def test_property_colon_minus(self):
...
+        self.props.setProperty("buildername", "linux4", "test")
+        command = Interpolate("echo buildby-%(prop:buildername:-blddef)s")
+
+        from buildbot.process import properties
+        saved_notHasKey = properties._notHasKey
+
+        def restore(result):
+            """Avoid making other tests mad"""
+            properties._notHasKey = saved_notHasKey
+            return result
```
tardyp: there is self.patch to do patch + restore pattern.
http://twistedmatrix.com/documents/15.0.0/api/twisted.trial.unittest.SynchronousTestCase.html#patch

master/buildbot/test/unit/test_process_properties.py
```diff
@@ -408,6 +408,29 @@ def test_property_colon_minus(self):
...
+        properties._notHasKey = object()
+
+        d = self.build.render(command)
```
tardyp: please write the tests with inlineCallbacks, which are much more easy tor read

tardyp:
@tomprince I'd like to get your opinion on this, since you are the author of this code

I think we should make sure that the marker is really constant rather than working around

gracinet:
@tardyp, if a reload occurs, the marker can't be constant. Here's a demo:

cst.py
```
cst = object()

class L(object):

    def __init__(self, opt=cst):
        self.opt = opt

    def not_default(self):
        return self.opt is not cst
```

main.py:
```
import cst
cst1 = cst.cst
l = cst.L()
reload(cst)

print("Before: %r, after: %r" % (cst1, cst.cst))
assert(cst1 is not cst.cst)

print("Now the L instance believe its opt is not the default one")
assert(l.not_default())
```

master/buildbot/test/unit/test_process_properties.py
gracinet: I agree, but all surrounding tests are written in that style.

tardyp: we usually write new tests inlinecallbacks, and fix the old one when we can

master/buildbot/test/unit/test_process_properties.py
gracinet: thanks, done !

tardyp:
about constant marker:
yes of course if you write markers using 
marker = object()
then it can't be constant over reload.

I was wondering if there is a better method to do markers that would survive reloading

master/buildbot/test/unit/test_process_properties.py
gracinet: done

tardyp:
If I look even a little bit further, I don't see why we could not replace the whole _notHasKey by None.

- _Lookup is private, and there is not case where it is used with hasKey=None
- I have not sure of any use case that would make sense to put hasKey=None. But the code is not crystal clear to me.

gracinet:
Even strings behave that way as long as you insist on testing them with `is` / `is not`.
I tried to play with classes implementing ``__hash__`` to no result, it turns out that ``is`` is strict.

So, unless I missed something, it seems that the better way would be to change all tests for `_notHasKey` not to use this constant-based pattern.

tardyp:
also reloading buildbot.process.properties does not work because of adapters.

  File "/Users/ptardy/dev/bb/buildbot-heroku/sandbox/lib/python2.7/site-packages/twisted/python/components.py", line 87, in registerAdapter
    raise ValueError("an adapter (%s) was already registered." % (factory, ))
ValueError: an adapter (<class 'buildbot.process.properties._DefaultRenderer'>) was already registered.


replacing _notHasKey by None pleases all the tests except the one that tests that default is not None:

    def test_Lookup_hasKey(self):
        self.failIfEqual(
            _Lookup({'test': 5, 'other': 6}, 'test', hasKey=None),
            _Lookup({'test': 5, 'other': 6}, 'test'))

tardyp:
So unless @tomprince explains why this has to be done like this, I would be in favor of simplifying _notHasKey by None in all this code.

gracinet:
about adapters, I noticed that while writing the test, so that something else must change the marker value (or properly unregister adapters before that, I don't know).

Anyway, I wouldn't mind using `None` (wouldn't have done that alone by myself)

tardyp:
> about adapters, I noticed that while writing the test, so that something else must change the marker value (or properly unregister adapters before that, I don't know).

Indeed, so you cannot really reload that module in the test, which should imho be the actual test, as this is your bug.
btw, how could you trig that bug, if this module cannot be reloaded?

> Anyway, I wouldn't mind using None (wouldn't have done that alone by myself)

Thats why we are a community trying to find out best solution for our common code.

tomprince:
I don't think buildbot wants to go down the route of supporting reloading buildbot modules, that way lies madness. I would be inclined to close any bug asking for that as WONTFIX. I see from the ticket that apparently this also occurs without any reloading, but I think it would be valuable to determine what is actually going on in that case, before going off and changing the code in a way to support reloading as a fix for a possibly unrelated problem.

> So unless @tomprince explains why this has to be done like this, I would be in favor of simplifying _notHasKey by None in all this code.

It has been a while since I looked at this code, but I'm fairly certain that I intended `_Lookup` to be made public, in which case, `None` is definitely something that people might want to pass to `hasKey`, which shouldn't be broken.

gracinet:
Hi @tomprince !

I can confirm that the 0.9 case I have on hand does not involve any reload, at least not in master.cfg, nor in directly called libraries. It's also systematic, and happens as soon as the first build after buildmaster start.

I'll try to figure out why the value of the marker changes. Probably will have to do some dichotomy for that, as I am already sure that `buildbot.process.properties` does not get imported twice.

gracinet:
Ok, got it, with the explanation for why it doesn't appear so often.

First, it's not the constant that changes, it's actually the value inside _Lookup, so I started suspecting some kind of serialization/deserialization and set in place some tooling to introspect that kind of thing.
Turns out that was a `deepcopy()`, and it does come from some dispatching code of mine.

So, in short, the issue is that a deep copy of an `Interpolate` instance that has no `hasKey` value is broken. So would be an unpickled one. 

I can avoid that `deepcopy()` (it's not even intended to work on build steps), and stop being affected by this, even if this PR gets dropped.

But I still think that this is a nasty trap (it's been especially hard to debug), whose root cause is that the absence of data gets stored as something that's not so totally unambiguous as one might think. 
This PR avoids the problem entirely and makes `Interpolate` safer for deepcopy or serializations. Storing a simple boolean and testing for it would also work, but it would introduce more compatibility risks than a Python property.

tomprince:
A simpler and (I think) clearer solution would be something like
```python
class _notHasKey(util.ComparableMixin):
    compare_attrs = []
...
   def __init__(self, ..., hasKey=_notHasKey(), ...):
      ...
```
and then checking either with `==` or `isinstance` instead of `is`.

---
That is, make the default value survive copying, rather than working around it not working if it is copied.

gracinet:
@tomprince, yes it'd work also, it spares the property, but requires changing all the `if` occurrences. If everybody is comfortable with that, I can change it to this way of doing.

tomprince:
I think that that change will end up with better code, even if is bigger. The code in this PR is a workaround for something that works, when we could just have code that works.

tardyp:
I like @tomprince usage of ComparableMixin. There is only one if to change as far as I've seen.

We can reuse the same unit test, and create a new one to verify it also works with deepcopy of a Interpolate object..

gracinet:
Did as @tomprince suggested
The travis error seems to be a false positive: all tests pass, the only error I see in the log is a PyPI download

@tardyp maybe I should squash all these changes an make a smaller PR ?

tardyp:
:+1: 
@sa2ajj will indeed appreciate squashing in one patch

gracinet:
superseded by #2082

PR #2076, opened 2016-03-29
Added MIT License 

zozo123:


tardyp:
thanks zozo, but no thanks

PR #2078, opened 2016-03-29
optimize builds by buildrequests query

tardyp:
since rework of UI, UI search builds by buildrequests using:

https://bboturl/api/v2/builds?buildrequestid=58642
which takes very long, because filtering is implemented in rest layer

While this one is instant:
https://bboturl/api/v2/buildrequests/58642/builds

It is pretty easy to fix this in the data layer, so we fix it there.
We could also fix in the UI layer..


yetanotherion:
:+1:

PR #2079, opened 2016-03-29
Added GPL License 

zozo123:


rutsky:
@zozo123 please squash you commits for next pull requests.

There is no need to `license` file in the root of Buildbot, since Buildbot consists of multiple projects distributed over several directories, and they have appropriate license files there (or setuptools license specification).
E.g. master's GPL: https://github.com/buildbot/buildbot/blob/master/master/COPYING

PR #2081, opened 2016-03-31
Fix LdapUserInfo group search for UTF-8 DNs

gracinet:
This is about http://trac.buildbot.net/ticket/3510#ticket


djmitche:
This is the right fix, but I believe we are set up to use `future` and related utilities so that this would still work under python 3 (where `str` and `unicode` are the same thing).  Can you tweak the patch to use that pattern?

gracinet:
@djmitche will look into it (and check the behaviour of ldap3 under python3)

gracinet:
@djmitche should be ok (and ldap3 does the right thing under py3)

tardyp:
:+1:

PR #2083, opened 2016-03-31
Include build properties in gerrit summary callback

ryansydnor:
This allows a gerrit summary callback to do some more advanced filtering

### DIFF
--- a/master/buildbot/reporters/gerrit.py
+++ b/master/buildbot/reporters/gerrit.py
@@ -326,6 +326,7 @@ class GerritStatusPush(service.BuildbotService):
                 }.get(result, "completed with unknown result %d" % result)
 
                 return {'name': build['builder']['name'],
+                        'properties': build['properties'],
                         'result': result,
                         'resultText': resultText,
                         'text': build['state_string'],


sa2ajj:
in 'eight' I added a more generic 'build' property exactly for tihs kind of cases: that would provide informarmation for all possible properties of a build without further extending the dictionary.

ryansydnor:
@sa2ajj - what do you mean "in 'eight'"? v8? Is the change easily pulled forward?

A slightly broader question - why transform the build objects at all? Why not just pass a list of builds to the callback?

sa2ajj:
in version `0.8.x`, which is maintained in branch 'eight'.  as for easily, I'm looking at it: the change itself, yes, the test -- I need to contemplate a bit.  (for your reference https://github.com/buildbot/buildbot/blob/eight/master/buildbot/status/status_gerrit.py#L303)

Answering your question: that would **require** better understanding of internals of Buildbot for those who would want to use the corresponding reporter (or, as it was previously called, status target).

sa2ajj:
@ryansydnor please have alook at #2085

ryansydnor:
@sa2ajj - beautiful! closing in favor of that. thanks for your help :)

PR #2084, opened 2016-04-01
Change compare_attrs to use tuples instead of lists

ewongbb:
Based on Nine (master)

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -31,9 +31,9 @@ from buildbot.util import json
 
 class BitbucketPullrequestPoller(base.PollingChangeSource):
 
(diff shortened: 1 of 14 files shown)


ewongbb:
did I mention it's a WIP?

ewongbb:
review ping

rutsky:
Changes LGTM.

Is lists changes to tuples in all places? How did you checked that?

@sa2ajj suggested adding explicit check (to ComparableMixing?) that forces use of tuples.
While I don't think this is good for third party code (if this is part of public interface, which I'm unsure), explicit check can be used to see if anything is missed.

rutsky:
Is this PR still WIP?

ewongbb:
No. Not a WIP.

PR #2086, opened 2016-04-01
Change assert_ checks to assertEquals/NotEquals/True/False...

ewongbb:
This is for ticket 2521.

### DIFF
--- a/master/buildbot/test/unit/test_changes_p4poller.py
+++ b/master/buildbot/test/unit/test_changes_p4poller.py
@@ -147,7 +147,7 @@ class TestP4Poller(changesource.ChangeSourceMixin,
         self.add_p4_describe_result(3, encoded_p4change[3])
 
         # The first time, it just learns the change to start at.
-        self.assert_(self.changesource.last_change is None)
+        self.assertTrue(self.changesource.last_change is None)
         d = self.changesource.poll()
 
         def check_first_check(_):
--- a/master/buildbot/test/unit/test_changes_p4poller.py
+++ b/master/buildbot/test/unit/test_changes_p4poller.py
@@ -147,7 +147,7 @@ class TestP4Poller(changesource.ChangeSourceMixin,
         self.add_p4_describe_result(3, encoded_p4change[3])
 
         # The first time, it just learns the change to start at.
-        self.assertTrue(self.changesource.last_change is None)
+        self.assertIsNone(self.changesource.last_change)
         d = self.changesource.poll()
 
         def check_first_check(_):
--- a/master/buildbot/test/unit/test_changes_p4poller.py
+++ b/master/buildbot/test/unit/test_changes_p4poller.py
@@ -147,7 +147,7 @@ class TestP4Poller(changesource.ChangeSourceMixin,
         self.add_p4_describe_result(3, encoded_p4change[3])
 
         # The first time, it just learns the change to start at.
-        self.assertIsNone(self.changesource.last_change)
+        self.assertTrue(self.changesource.last_change is None)
         d = self.changesource.poll()
 
         def check_first_check(_):
--- a/master/buildbot/test/unit/test_process_build.py
+++ b/master/buildbot/test/unit/test_process_build.py
@@ -388,7 +388,7 @@ class TestBuild(unittest.TestCase):
         self.assertNotIn(('startStep', (self.workerforbuilder.worker.conn,), {}),
                      step.method_calls)
         self.assertEquals(claimCount[0], 1)
-        self.assertIsNone(b.currentStep)
+        self.assertTrue(b.currentStep is None)
         self.assertIsNotNone(b._acquiringLock)
 
     def testStopBuildWaitingForLocks(self):
@@ -419,7 +419,7 @@ class TestBuild(unittest.TestCase):
 
         self.assertNotIn(('startStep', (self.workerforbuilder.worker.conn,), {}),
                      step.method_calls)
-        self.assertIsNone(b.currentStep)
+        self.assertTrue(b.currentStep is None)
         self.assertEqual(b.results, CANCELLED)
         self.assertNotIn(('interrupt', ('stop it',), {}), step.method_calls)
 
@@ -451,7 +451,7 @@ class TestBuild(unittest.TestCase):
 
         self.assertNotIn(('startStep', (self.workerforbuilder.worker.conn,), {}),
                      step.method_calls)
-        self.assertIsNone(b.currentStep)
+        self.assertTrue(b.currentStep is None)
         self.assertEqual(b.results, RETRY)
         self.assertNotIn(('interrupt', ('stop it',), {}), step.method_calls)
         self.build.build_status.setText.assert_called_with(["retry", "lost", "connection"])
@@ -476,7 +476,7 @@ class TestBuild(unittest.TestCase):
         def acquireLocks(res=None):
             gotLocks[0] = True
             retval = LoggingBuildStep.acquireLocks(step, res)
-            self.assertIs(b.currentStep, step)
+            self.assertTrue(b.currentStep is step)
             b.stopBuild('stop it')
             return retval
         step.acquireLocks = acquireLocks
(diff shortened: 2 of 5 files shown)


master/buildbot/test/unit/test_changes_p4poller.py
```diff
@@ -147,7 +147,7 @@ def do_test_poll_successful(self, **kwargs):
         self.add_p4_describe_result(3, encoded_p4change[3])
 
         # The first time, it just learns the change to start at.
-        self.assert_(self.changesource.last_change is None)
+        self.assertTrue(self.changesource.last_change is None)
```
rutsky: Can we use `self.assertIsNone` here?

master/buildbot/test/unit/test_process_build.py
```diff
@@ -157,7 +157,7 @@ def testRunSuccessfulBuild(self):
         b.startBuild(FakeBuildStatus(), None, self.workerforbuilder)
 
         self.assertEqual(b.results, SUCCESS)
-        self.assert_(('startStep', (self.workerforbuilder.worker.conn,), {})
+        self.assertTrue(('startStep', (self.workerforbuilder.worker.conn,), {})
```
rutsky: Can we use `self.assertIn(..., ...)`?

master/buildbot/test/unit/test_steps_transfer.py
```diff
@@ -866,8 +866,8 @@ def testBasic(self):
                 self.assertEquals(
                     data, json.dumps(dict(sourcestamps=[ss.asDict()], properties={'key1': 'value1'})))
                 break
-        else:
-            self.assert_(False, "No downloadFile command found")
+            else:
+                self.assertRaise(ValueError, "No downloadFile command found")
```
rutsky: Is there `self.assertRaise` in `TestCase`?

I know about `self.assertRaises`, and in this place I think original code is more corresponds to `raise ValueError("No downloadFile command found")/`

rutsky:
Modern versions of `unittest` module have convenient methods like [`assertIn`](https://docs.python.org/3.5/library/unittest.html#unittest.TestCase.assertIn) and [`assertIsNone`](https://docs.python.org/3.5/library/unittest.html#unittest.TestCase.assertIsNone), however I'm not sure is they were backported to Twisted's `TestCase` or is Twisted's `TestCase` has proper alternative.

master/buildbot/test/unit/test_steps_transfer.py
rutsky: Also you changed logic here, when moved `else` to `if`.

tardyp:
twisted TestCase is a subclass of unittest TestCase..

master/buildbot/test/unit/test_steps_transfer.py
```diff
@@ -866,8 +866,8 @@ def testBasic(self):
                 self.assertEquals(
                     data, json.dumps(dict(sourcestamps=[ss.asDict()], properties={'key1': 'value1'})))
                 break
-        else:
-            self.assert_(False, "No downloadFile command found")
+            else:
+                raise ValueError("No downloadFile command found")
```
rutsky: Why do you move `else` here?

ewongbb: got confused.  Also didn't realize there was the for...else: syntax.

master/buildbot/test/unit/test_changes_p4poller.py
Stibbons: yes, it is advised

master/buildbot/test/unit/test_process_build.py
```diff
@@ -157,8 +157,8 @@ def testRunSuccessfulBuild(self):
         b.startBuild(FakeBuildStatus(), None, self.workerforbuilder)
 
         self.assertEqual(b.results, SUCCESS)
-        self.assert_(('startStep', (self.workerforbuilder.worker.conn,), {})
-                     in step.method_calls)
+        self.assertIn(('startStep', (self.workerforbuilder.worker.conn,), {}),
```
Stibbons: This is a good candidates for refactor in TestCase base classe (we use a tiny abstract class between unittest.TestCase and our tests, so we can implement our how missing assert* methods)...

  self.assertMethodCalledWith(step, ('startStep', (self.workerforbuilder.worker.conn,), {}),

master/buildbot/test/unit/test_process_build.py
```diff
@@ -385,11 +385,11 @@ def claim(owner, access):
...
         b.startBuild(FakeBuildStatus(), None, self.workerforbuilder)
 
-        self.assert_(('startStep', (self.workerforbuilder.worker.conn,), {})
-                     not in step.method_calls)
+        self.assertNotIn(('startStep', (self.workerforbuilder.worker.conn,), {}),
+                     step.method_calls)
         self.assertEquals(claimCount[0], 1)
-        self.assert_(b.currentStep is None)
-        self.assert_(b._acquiringLock is not None)
+        self.assertTrue(b.currentStep is None)
```
Stibbons: assertIsNone

master/buildbot/test/unit/test_process_build.py
```diff
@@ -385,11 +385,11 @@ def claim(owner, access):
 
         b.startBuild(FakeBuildStatus(), None, self.workerforbuilder)
 
-        self.assert_(('startStep', (self.workerforbuilder.worker.conn,), {})
-                     not in step.method_calls)
+        self.assertNotIn(('startStep', (self.workerforbuilder.worker.conn,), {}),
+                     step.method_calls)
         self.assertEquals(claimCount[0], 1)
-        self.assert_(b.currentStep is None)
-        self.assert_(b._acquiringLock is not None)
```
Stibbons: assertIsNotNone

master/buildbot/test/unit/test_process_metrics.py
```diff
@@ -182,13 +182,13 @@ def testReconfig(self):
         # enable log_interval
         new_config.metrics = dict(log_interval=10, periodic_interval=0)
         observer.reconfigServiceWithBuildbotConfig(new_config)
-        self.assert_(observer.log_task)
+        self.assertTrue(observer.log_task)
         self.assertEquals(observer.periodic_task, None)
```
Stibbons: assertIsNone

master/buildbot/test/unit/test_steps_transfer.py
```diff
@@ -867,7 +867,7 @@ def testBasic(self):
                     data, json.dumps(dict(sourcestamps=[ss.asDict()], properties={'key1': 'value1'})))
                 break
         else:
-            self.assert_(False, "No downloadFile command found")
+            raise ValueError("No downloadFile command found")
```
Stibbons: self.fail("No downloadFile command found")

master/buildbot/test/unit/test_process_build.py
gracinet: assertIsNone appears with python 2.7, I think (easy to backport in base test class)

gracinet:
By the way, `assertEquals` is deprecated in favor of `assertEqual`, if that matters
https://docs.python.org/2.7/library/unittest.html#deprecated-aliases

master/buildbot/test/unit/test_process_build.py
ewongbb: yes.. 2.7.  adding assertIsNone (and similar asserts) fail in 2.6, which is why I returned to assertTrue.

Stibbons: monkey patch :)

tardyp: buildbot master is now python2.7 only, so You have assertIsNone assertRaises, and all that fancy stuff.
Note that even when we were compatible 2.6, we added some of them via monkey patching (we might want to remove those btw)

rutsky: Is it 2.7 only? I think we agreed to not break 2.6 compatibility in master without special need on one of the last meetings.

tardyp: indeed, sorry. I was confused by the python2,6 PR from @sa2ajj

tardyp:
in this case, lets merge it.

PR #2088, opened 2016-04-01
Fixes CMake documentation example error and typo.

arianepaola:
This fixes an error in the CMake documentation example and also a typo when using definitions.


sa2ajj:
@arianepaola thank you very much :)

arianepaola:
You're welcome @sa2ajj :smile:

PR #2089, opened 2016-04-03
compress logs in a transaction

tardyp:
During the log compression process, there is a time where all the log chunks are deleted
If while that a process reads the log, it will find an empty log.

This happens on integrations tests making them flaky.

This can theorically also happen for reporters even if we did not get any bug reports on that

What we do is to make the delete + re-insert a transaction, so that any read in between will get the old un-compressed version of the chunks (which is an okay answer)

I can reproduce the bug easily with a combination of sleep 2 after delete, and sleep 1 at the begining of getLogLines's thd,
but I dont want to cripple the code to add the necessary syncrhonisation hooks required to unit test this bug.

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -177,25 +177,27 @@ class LogsConnectorComponent(base.DBConnectorComponent):
                      num_lines=last_line + 1)
         return first_line, last_line
 
-    def appendLog(self, logid, content):
+    def thdAppendLog(self, conn, logid, content):
         # check for trailing newline and strip it for storage -- chunks omit
         # the trailing newline
         assert content[-1] == u'\n'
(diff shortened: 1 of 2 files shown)


master/buildbot/db/logs.py
```diff
@@ -268,28 +270,24 @@ def thd(conn):
             for row in rows:
                 wholelog += self.COMPRESSION_BYID[row.compressed]["read"](row.content).decode('utf-8') + "\n"
 
+            if len(wholelog) == 0:
+                return 0
```
djmitche: `defer.returnValue`

djmitche: ugh, sorry, misread the indentation :)

djmitche:
Looks good!

PR #2090, opened 2016-04-03
Beta8

tardyp:
- relnotes for beta8

- remove all the tocs, which make sphinx very slow
- move test deps from .travisrc into [tests]


tardyp:
I went through the git log to look for fixes that weren't in relnotes.
I dont always ask for relnotes for simple fixes, as this often leads to rebase issues.

@rutsky @djmitche @sa2ajj let me know if I forgot something

djmitche:
:+1: .. not sure why travis is hanging..

master/docs/relnotes/0.9.0b8.rst
```diff
@@ -0,0 +1,556 @@
+Release Notes for Buildbot 0.9.0b8
+==================================
+
+
```
rutsky: Do we need extra line feed here?

master/docs/relnotes/0.9.0b8.rst
```diff
@@ -0,0 +1,556 @@
...
+* Optimized UI REST requests for child builds and change page.
+* Fix :class:`DockerLatentWorker` use of `volume` parameter, they now propely manage `src:dest` syntax.
+* Fix :class:`DockerLatentWorker` to properly create properties so that docker parameters can be renderable.
+* Lock down autobahn version for python 2.6 (note that autobahn and twisted are no longer supporting 2.6, and thus do not receive security fixes anymore).
+* fix docs and example to always use port 8020 for the web ui.
+
+Deprecations, Removals, and Non-Compatible Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Deprecated ``workdir`` was removed, ``builddir`` property should be used instead.
```
rutsky: Deprecated ``workdir`` **property** was removed.

master/docs/relnotes/0.9.0b8.rst
```diff
@@ -0,0 +1,556 @@
...
+------
+
+Package buildbot-slave is being renamed buildbot-worker.
+As the work is not completly finished, neither `buildbot-slave==0.9.0b8` or `buildbot-worker==0.9.0b8` have been released.
+
+You can safely use any version of buildbot-slave with `buildbot==0.9.0b8`, either `buildbot-slave==0.8.12` or `buildbot-slave==0.9.0b7`.
+
+
+Deprecations, Removals, and Non-Compatible Changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rutsky: Do we store empty chapters in release notes?

master/docs/relnotes/0.9.0b8.rst
```diff
@@ -0,0 +1,556 @@
...
+``buildslaveid`` key in messages were replaced with ``workerid``.
+
+Details
+-------
+
+For a more detailed description of the changes made in this version, see the git log itself:
+
+.. code-block:: bash
+
+   git log v0.9.0b7..master
```
rutsky: Isn't there should be v0.9.0b7..v0.9.0b8?

rutsky:
I like idea not to release unfinished `buildbot-slave`/`buildbot-worker` packages.

I like removing of sub-TOCs from beta releases.

Here is how rel.notes looks like currently: https://circle-artifacts.com/gh/buildbot/buildbot/323/artifacts/0/home/ubuntu/buildbot/master/docs/_build/html/relnotes/0.9.0b8.html

Maybe we remove empty chapters from final release notes, like `Changes for Developers`, `Deprecations, Removals, and Non-Compatible Changes`?
@tardyp or maybe you wanted to make `Changes for Developers` and `Deprecations, Removals, and Non-Compatible Changes` higher level captions, so that content above will be included in it?

rutsky:
@tardyp can you add `Twisted` into list of preinstalled packages for `setup.py test` running, i.e [here](https://github.com/buildbot/buildbot/blob/master/.travis.yml#L143):

```diff
- - "[ $TESTS != setuppy_test ] || pip install autobahn"
+ - "[ $TESTS != setuppy_test ] || pip install autobahn Twisted"
```

This should fix failing Travis build.

tardyp:
@rutsky I am not sure I understand why we should first instal Twisted and autobahn.
Shouldn't it installed via normal dependencies of '-e master[tls,test]' ?

rutsky:
@tardyp 
> @rutsky I am not sure I understand why we should first instal Twisted and autobahn.
> Shouldn't it installed via normal dependencies of '-e master[tls,test]' ?

In `python setup.py test` invocation the goal is *not* to install anything --- setuptools should automatically download and handle all dependencies without their installation (as they are properly listed in `setup.py`).
So in ideal world this should work and properly run (almost) all Buildbot tests:

```
$ # here we are in *clean* virtualenv
$ git clone https://github.com/buildbot/buildbot.git
$ cd buildbot
$ python setup.py test
...
```

The problem is that due to setuptools bug it can't handle autoinstallation of some packages that depends on Twisted.
I thought that it is only `autobahn`, but looks like `txrequests` fails too.

I added `- "[ $TESTS != setuppy_test ] || pip install autobahn"` to workaround this issue, in `setuppy_test` Travis mode `-e master[tls,test]` is not installed intentionally.
So I suggest to add Twisted too in this workaround.

tardyp:
@rutsky I dont understand the part were txrequests fails so we need to force install of Twisted.

tardyp:
Thanks. This is clear now.

Le lun. 4 avr. 2016 16:04, Vladimir Rutsky <notifications@github.com> a
écrit :

tardyp:
well, its clear, but the build is still failing, even with your suggested fix.

rutsky:
Looks like it's not Twisted, that should be installed, but packages that uses specific codepath that calls Twisted code during installation.

@tardyp can you try this?

```
- "[ $TESTS != setuppy_test ] || pip install autobahn txrequests"
```

tardyp:
ok, I though I could do this before tonight, but it looks like this will be difficult. I am traveling the rest of the week, so I guess the release will be next weekend..

tardyp:
rebased. 

I am disabling the setuppy_test.

I am really not sure we want to support setup.py test if this does not work out of the box, and we have to make such workarounds.
For me there is no point of setup.py test if there is a prerequisite pip install before.

rutsky:
Even with not working without installing some deps, `setup.py test` is still useful when deps are already installed: it enables test-runner-independent Setuptools' way of running tests.

Lets enable `setup.py test` when Setuptools' bug will be fixed.

tardyp:
@rutsky sounds like a plan!

PR #2092, opened 2016-04-05
Fix plugins

sa2ajj:
a different approach for handling transiitons


tardyp:
can you describe a little bit your approach?
 I am not sure I followed exactly the debate

sa2ajj:
It's not a debate: what I want to achieve in this PR (once it's finished) is to provide a generic mechanism to deprecate plugins/namespaces.  On the other hand, I'd like to keep one important properties of plugins: on demand load only.

The test cases are not there yet.  The idea is to provide generic test cases for generic functionality and keep specific ones for the slave-to-worker transition.

sa2ajj:
I will close this for now.

PR #2093, opened 2016-04-05
Send JSON to GitHub instead of form encoded

Chippiewill:
GitHub expects a JSON request body

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -79,7 +79,7 @@ class GithubStatusPush(http.HttpStatusPushBase):
             payload['context'] = context
 
         return self.session.post('/'.join(
-            [self.baseURL, 'repos', repo_user, repo_name, 'statuses', sha]), payload)
+            [self.baseURL, 'repos', repo_user, repo_name, 'statuses', sha]), json=payload)
 
     @defer.inlineCallbacks
     def send(self, build):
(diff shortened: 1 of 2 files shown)


Chippiewill:
Just spent a little while figuring out why the status notifications weren't making it to GitHub. I would also note that:

````python
            try:
                yield self.createStatus(
                    repo_user=repoOwner.encode('utf-8'),
                    repo_name=repoName.encode('utf-8'),
                    sha=sha.encode('utf-8'),
                    state=state.encode('utf-8'),
                    target_url=build['url'].encode('utf-8'),
                    context=context.encode('utf-8'),
                    description=description.encode('utf-8')
                )
                if self.verbose:
                    log.msg(
                        'Status "{state}" sent for '
                        '{repoOwner}/{repoName} at {sha}.'.format(
                            state=state, repoOwner=repoOwner, repoName=repoName, sha=sha))
            except Exception as e:
                log.err(
                    e,
                    'Fail to send status "{state}" for '
                    '{repoOwner}/{repoName} at {sha}'.format(
                        state=state, repoOwner=repoOwner, repoName=repoName, sha=sha))
````

is slightly flawed in that no exception is thrown by requests if it encounters an error. The status code of the yielded response object needs to be checked for something appropriate (e.g. 2xx)

sa2ajj:
@Chippiewill thank you

PR #2097, opened 2016-04-10
Teach HttpStatusPush and StashStatusPush reporters correct response syntax

rayalan:
A couple small bugs I encountered while copying this code to make a buildset monitor. 

HTTPStep has the correct implementation already, and I don't see it being used elsewhere. No unit tests to update as far as I can tell; the existing TestHttpStatusPush and TestStackStatusPush classes seem focused on mocking sure the right calls are made and I don't see a quick way to extend them to mock out responses.


sa2ajj:
LGTM

tardyp:
looks good. Looks like we have a python2.6 regression because of an updated dep... again.

PR #2098, opened 2016-04-10
fix 2.6 tests by fixing txaio version

tardyp:


rutsky:
One more nail in the coffin of Python 2.6 :)

PR #2101, opened 2016-04-10
Use the threadpool from the provided reactor, rather than the global one.

tomprince:
I want to introduce some tests that don't interact with a real reactor (so that they can be synchronous). This is one of the places that uses a hard-coded global reactor.


master/buildbot/test/integration/test_master.py
```diff
@@ -54,6 +54,8 @@ def do_test_master(self):
         # like test code to call!)
         mock_reactor = mock.Mock(spec=reactor)
         mock_reactor.callWhenRunning = reactor.callWhenRunning
+        mock_reactor.getThreadPool = reactor.getThreadPool
+        mock_reactor.callFromThread = reactor.callFromThread
```
tomprince: Ideally, all of these would be replaced with a verified fake, instead of a mock.

tardyp:
looks good!

PR #2102, opened 2016-04-11
Remove ``AbstractLatentWorker.__init__``.

tomprince:
``AbstractLatentWorker`` is a ``BuildbotService``, which has an
``__init__`` that delegates to ``checkConfig``. ``AbstractLatentWorker``
should just have an ``checkConfig`` and not override
``BuildbotService.__init__``.

### DIFF
--- a/master/buildbot/test/unit/test_worker_base.py
+++ b/master/buildbot/test/unit/test_worker_base.py
@@ -517,7 +517,6 @@ class TestWorkerTransition(unittest.TestCase):
         self.assertIdentical(deprecated, AbstractWorker)
 
     def test_AbstractLatentBuildSlave_deprecated_worker(self):
-        from buildbot.worker import AbstractLatentWorker
 
         import buildbot.buildslave as bs
 
@@ -526,7 +525,7 @@ class TestWorkerTransition(unittest.TestCase):
                 message_pattern="AbstractLatentBuildSlave was deprecated"):
             deprecated = bs.AbstractLatentBuildSlave
 
-        self.assertIdentical(deprecated, AbstractLatentWorker)
+        self.assertIdentical(deprecated, base.AbstractLatentWorker)
 
     def test_BuildSlave_deprecated_worker(self):
         from buildbot.worker import Worker
--- a/master/buildbot/test/unit/test_worker_base.py
+++ b/master/buildbot/test/unit/test_worker_base.py
@@ -471,6 +471,37 @@ class TestAbstractWorker(unittest.TestCase):
         self.assertEqual(worker.worker_status.getGraceful(), True)
 
 
+class TestAbstractLatentWorker(unittest.SynchronousTestCase):
+
+    def setUp(self):
+        self.master = fakemaster.make_master(wantDb=True, wantData=True,
+                                             testcase=self)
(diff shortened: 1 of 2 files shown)


master/buildbot/worker/base.py
```diff
@@ -77,6 +79,8 @@ def checkConfig(self, name, password, max_builds=None,
         @type locks: dictionary
         """
         self.name = name = ascii2unicode(name)
+        self.build_wait_timeout = build_wait_timeout
```
tardyp: you rather need to update this in reconfigureService(), or the reconfiguration will not work if build_wait_timeout is changed (better than nothing, though)

tardyp:
also, the subclasses are also implementing __init__ (just checked docker)

master/buildbot/worker/base.py
tomprince: Thanks for catching this. It appears that this previously wan't being reconfigured before.

tardyp: indeed, there are still a bunch of services that needs to be migrating to the reconfigure api.
schedulers and workers are one of those. 
This means buildbot reload won't apply changes for those.

I need to make the list, and see if we want to release with our without the fix.
not complicated, but requires time to make sure this is properly tested.

rutsky:
Windows build fails because of errors during `lz4` installation, looks like this bug: https://github.com/darkdragn/lz4tools/issues/6

master/buildbot/test/unit/test_worker_base.py
```diff
@@ -471,6 +471,37 @@ def test_shutdownRequested(self):
...
+        self.clock = task.Clock()
+        self.patch(reactor, 'callLater', self.clock.callLater)
+        self.patch(reactor, 'seconds', self.clock.seconds)
+
+    def do_test_reconfigService(self, old, new, existingRegistration=True):
+        old.parent = self.master
+        if existingRegistration:
+            old.registration = bworkermanager.FakeWorkerRegistration(old)
+        old.missing_timer = mock.Mock(name='missing_timer')
+        self.successResultOf(old.startService())
```
tardyp: I wasn't aware of this method.

This fails with twisted 12:

  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_worker_base.py", line 500, in test_reconfigService
    self.do_test_reconfigService(old, new)
  File "/home/travis/build/buildbot/buildbot/master/buildbot/test/unit/test_worker_base.py", line 492, in do_test_reconfigService
    self.successResultOf(old.startService())
exceptions.AttributeError: 'TestAbstractLatentWorker' object has no attribute 'successResultOf'

Looks like was introduced in 13.
I am not against dropping such old version of twisted, but we should probably ask the ml first.

tomprince:
> also, the subclasses are also implementing init (just checked docker)

That should be changed, but I don't think that is a prerequisite for this fix? One trick with updating the actual latent workers is that if the credentials change, we won't be able to shutdown any currently running instances.

master/buildbot/test/unit/test_worker_base.py
tomprince: We have now dropped twisted < 14.0.

tardyp:
👍

PR #2105, opened 2016-04-11
configure logging for buildbot-worker runner

rutsky:
As alternative we can revert changes from 4d733e6 commit by @jvlomax
(in buildbot-worker copy).

By default Twisted log is not written anywhere so we don't see any
errors from buildbot-worker invocations, e.g.

```
$ buildbot-worker create-worker w
$ echo $?
1
$
```

### DIFF
--- a/worker/buildbot_worker/scripts/runner.py
+++ b/worker/buildbot_worker/scripts/runner.py
@@ -16,6 +16,8 @@
 # N.B.: don't import anything that might pull in a reactor yet. Some of our
 # subcommands want to load modules that need the gtk reactor.
 
+from __future__ import print_function
+
 import os
 import re
 import sys
@@ -24,10 +26,6 @@ import textwrap
 from twisted.python import log
 from twisted.python import reflect
 from twisted.python import usage
-from twisted.python.log import ILogObserver
-from twisted.python.log import textFromEventDict
-
-from zope.interface import implementer
 
 
 # the create/start/stop commands should all be run as the same user,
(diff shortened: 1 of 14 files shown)


tardyp:
LGTM

tardyp:
twisted.logger does not exist in twisted 12..

worker/buildbot_worker/scripts/runner.py
```diff
@@ -20,9 +20,13 @@
 import re
 import sys
 
+from twisted.logger import ILogObserver
+from twisted.logger import formatEvent
+from twisted.logger import globalLogBeginner
 from twisted.python import log
 from twisted.python import reflect
 from twisted.python import usage
+from zope.interface import implementer
```
sa2ajj: should we put an empty line before this one?

rutsky: @sa2ajj done

rutsky:
Rebased on master, I believe Twisted 12 is not a problem anymore.

No tests yet.

sa2ajj:
Let's see how it fares...  (I'll remove "needs work" for now though)

rutsky:
I resurrected `StdoutAssertionsMixin` and added lines for testing that error message ended in `stdout`.

I changed used logging API — now it works on Twisted 14.0 and 16.1 in my env.

rutsky:
This bug is not as easy as I expected.

Logger is currently managed in the following places:

1. When `buildbot-worker` is invoked and doesn't do anything interesting (e.g. `buildbot-worker --help`) logger is not initialized and managed (but used to output messages).

2. When `buildbot-worker` parts are used from tests, logger is initialized by test runner.

3. When worker is started by `buildbot-worker start ...` `buildbot-worker` invokes `twistd --logfile=twistd.log` log is initialized and managed by `twistd`.

Since there is no single point where logging is initialized and managed it's hard to prevent following issue: if logger is initialized by `twistd` or by test runner, it redirects all output to `stdout` to logfile, which leads to message duplication in log at best.

Also multiple initialization of logging leads to Twisted warning `Warning: primary log target selected twice`.

I think best is to revert outputting messages that are expected to be seen on command line to stdout.

rutsky:
I effectively reverted 4d733e6: most of `buildbot-worker` messages are now printed to stdout.

Printed messages to stdout appears in Twisted log, but not vice versa.

Having messages on stdout is critical for interactive usage of buildbot-worker
command line utility.

@jvlomax I appreciate if you comment on reason you done 4d733e6 and what do you think about my current changes.

djmitche:
I suspect the `log.msg` changes were made as part of Python-3-ification, and the resulting hiding of important messages wasn't noticed at the time.  So I think this revert is the right course.

rutsky:
@tardyp is changes still LGTY?

PR #2106, opened 2016-04-11
fix typo `BuilderFactory()` in example

rutsky:
Actually I don't see any implementation of `GitHubStatus` in Nine.

@tardyp what is with `GitHubStatus`? Is it removed in favor of `GithubStatusPush`?
We need to update docs if it is.

P.S. Same typo in eight branch. @sa2ajj is it you who maintains eight branch?


tardyp:
Indeed, I might have unified the naming for github reporter plugin

PR #2111, opened 2016-04-11
REVERTME: hack to reload the window if the websocket is disconnected

tardyp:
We should fix properly in dataModule using :bug:`3462`, but after initial nine release

fix in dataModule is much harder, as when reconnection is detected, we should
reload all watched collections, take care of sending the proper events, etc


tardyp:
may I merge this?

ryansydnor:
@tardyp - looks decent from my perspective. I think it'll be a good UX improvement for people who drop in/out of connection.

PR #2112, opened 2016-04-11
fix validate.sh to use command instead of which

sa2ajj:
`which` ignores certain directories, wihch results in misleading error messages (in my case `pylint`, `autopep8` are not found).


tardyp:
Wasn't aware of command -v.

If it works better for you 👍

PR #2115, opened 2016-04-12
Fix Appveyor build

rutsky:
* Disable LZ4 on Windows, since it fails to build on Windows: https://github.com/steeve/python-lz4/issues/27

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -433,12 +433,17 @@ else:
         'txrequests',
         'future',
         'pyjade',
-        'lz4',
         'boto',
         'moto',
         'txgithub',
         'mock',
     ]
+    if sys.platform != 'win32':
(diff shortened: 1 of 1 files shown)


rutsky:
Huh, disabling `lz4` on Windows was enough.

tardyp:
❤️

tardyp:
@rutsky can you configure appveyor so that it outputs a Build status rather than sending a message?

rutsky:
@tardyp I don't have "buildbot" in AppVeyor configuration at all. I think it's @djmitche who configured comments (and disabled them now, according to his messages in IRC).

@djmitche I think we need to do steps similar to these: http://help.appveyor.com/discussions/problems/3155-cant-see-any-projects-under-organization

PR #2118, opened 2016-04-12
Set vim filetype, not syntax.

tomprince:
This allows things like completion to take effect as well.


tardyp:
:+1: who needs completion ?

PR #2119, opened 2016-04-13
Fix issue builds and buildrequest display

ssavrim:
When urls of builds are buildbot ones, the url was displayed twice
(as raw and as buildrequestsummary)
This patch allow to display raw url for external urls only
and as buildrequestsummary for buildbot urls only.

Fix issue to display correctly the list of build requests on the builder page


www/base/src/app/common/directives/buildsummary/buildsummary.tpl.jade
```diff
@@ -49,7 +49,7 @@
                 | {{log.name}}
             | &nbsp;({{log.num_lines}} line{{log.num_lines > 1?'s':''}})
       ul
-        li(ng-if='!(isBuildRequestURL(url.url) || isBuildURL(url.url))', ng-repeat="url in step.urls")
```
bastien31: why not executing isBuildRequestURL from buildsummary on that line?

bastien31:
:+1:

tardyp:
thanks for the fix!

PR #2121, opened 2016-04-14
Cleanups.

tomprince:
These are a few unrelated cleanups I discovered while digging into other stuff. I didn't think it was worth splitting into separate PRs, but it is probably easier to review the individual commits.


tardyp:
looks good. I am not sure why appveyor fails. I can't rebuild it :-/

PR #2122, opened 2016-04-14
Allow gerrit reporter callbacks to be deferred

ryansydnor:
This addresses http://trac.buildbot.net/ticket/3522

### DIFF
--- a/master/buildbot/reporters/gerrit.py
+++ b/master/buildbot/reporters/gerrit.py
@@ -280,12 +280,14 @@ class GerritStatusPush(service.BuildbotService):
             builder = yield self.master.data.get(("builders", build['builderid']))
             build['builder'] = builder
             if self.isBuildReported(build):
-                message = self.startCB(builder['name'], build, self.startArg)
+                message = yield self.startCB(builder['name'], build, self.startArg)
                 self.sendCodeReviews(build, message)
 
+    @defer.inlineCallbacks
     def buildFinished(self, builderName, build, result):
         """Do the SSH gerrit verify command to the server."""
-        result = _handleLegacyResult(self.reviewCB(builderName, build, result, self.master, self.reviewArg))
+        result = yield self.reviewCB(builderName, build, result, self.master, self.reviewArg)
+        result = _handleLegacyResult(result)
         self.sendCodeReviews(build, result)
 
     @defer.inlineCallbacks
@@ -313,6 +315,7 @@ class GerritStatusPush(service.BuildbotService):
         buildset = res['buildset']
         self.sendBuildSetSummary(buildset, builds)
 
+    @defer.inlineCallbacks
     def sendBuildSetSummary(self, buildset, builds):
         builds = filter(self.isBuildReported, builds)
         if builds and self.summaryCB:
@@ -334,8 +337,12 @@ class GerritStatusPush(service.BuildbotService):
                         }
             buildInfoList = sorted([getBuildInfo(build) for build in builds], key=lambda bi: bi['name'])
 
-            result = _handleLegacyResult(self.summaryCB(buildInfoList, Results[buildset['results']],
-                                                        self.master, self.summaryArg))
+            result = yield self.summaryCB(buildInfoList,
+                                          Results[buildset['results']],
+                                          self.master,
+                                          self.summaryArg)
+
+            result = _handleLegacyResult(result)
             self.sendCodeReviews(builds[0], result)
 
     def sendCodeReviews(self, build, result):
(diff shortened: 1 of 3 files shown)


master/buildbot/reporters/gerrit.py
```diff
@@ -334,8 +335,12 @@ def getBuildInfo(build):
                         }
             buildInfoList = sorted([getBuildInfo(build) for build in builds], key=lambda bi: bi['name'])
 
-            result = _handleLegacyResult(self.summaryCB(buildInfoList, Results[buildset['results']],
-                                                        self.master, self.summaryArg))
+            result = yield self.summaryCB(buildInfoList,
```
tardyp: I think we should do this for all callbacks

ryansydnor: @tardyp - agree! Do you want to do this as part of this PR? Or use this as the example for future PRs?

tardyp: I think we should keep coherency, and do everything at once (as it is not huge change)

ryansydnor: Okay. I'll work on that tonight for the gerrit reporter.

ryansydnor: @tardyp - done!

master/buildbot/test/unit/test_reporter_gerrit.py
```diff
@@ -70,6 +85,29 @@ def testSummaryCB(buildInfoList, results, status, arg):
...
+            failure = True
+
+    if failure:
+        verified = -1
+    elif success:
+        verified = 1
+    else:
+        verified = 0
+
+    defer.returnValue(makeReviewResult(str(buildInfoList),
```
tardyp: You should rather do:

    result = yield makeReviewResult(...)
    self.expect...(results[..], foo)

master/buildbot/test/unit/test_reporter_gerrit.py
```diff
@@ -44,11 +44,26 @@ def testReviewCB(builderName, build, result, status, arg):
...
+
+
 def testStartCB(builderName, build, arg):
     return makeReviewResult(str({'name': builderName}),
                             (GERRIT_LABEL_REVIEWED, 0))
 
 
+@defer.inlineCallbacks
+def testStartCBDeferred(builderName, build, arg):
+    yield
```
tardyp: This does not make sense to yield nothing in a inlineCallback.

Please look at other tests to see how we usually do.

master/buildbot/test/unit/test_reporter_gerrit.py
ryansydnor: Done.

master/buildbot/test/unit/test_reporter_gerrit.py
ryansydnor: @tardyp - thanks for bearing with me while I learn the twisted framework. 

I thought yielding nothing was "good enough" for a test - it showed we were capable of calling a deferred if necessary. Your proposal makes much more sense, though! Fixed.

master/buildbot/test/unit/test_reporter_gerrit.py
```diff
@@ -70,6 +85,29 @@ def testSummaryCB(buildInfoList, results, status, arg):
...
+    if failure:
+        verified = -1
+    elif success:
+        verified = 1
+    else:
+        verified = 0
+
+    result = yield makeReviewResult(str(buildInfoList),
+                                    (GERRIT_LABEL_VERIFIED, verified))
+    defer.returnValue(result)
```
tardyp: don't you want to at least make sure the result variable is coherent with what you expect?

You can use one of those:
https://twistedmatrix.com/documents/14.0.2/api/twisted.trial._synctest._Assertions.html

ryansydnor: That already happens on line 211.

Basically, I'm testing that the result is exactly the same as calling the summary CB in a non-deferred way.

tardyp: Ah I thoughts those callbacks where tests.

There is a convention that says that methods starting with test are tests.
Please rename those function, like testedSummaryCBDeferred

ryansydnor: Good point. Fixed.

tardyp:
👍

ryansydnor:
@tardyp - waiting on anyone else to review?

tardyp:
I let some time for other reviewers to catch up

sa2ajj:
LGTM

PR #2123, opened 2016-04-14
minimal patch to not write pickle in the master's basedir

tardyp:
There are still relatively unknown scale use of the status objects in the core.

I prefer to delay this cleanup task until after buildbot 0.9.0 is released.

just nop the saveYourSelf functions, and remove the status unit tests.


tardyp:
http://trac.buildbot.net/ticket/3523

tardyp:
May I merge this?

djmitche:
So, this will still load the items, but not save them?  That sounds OK to me.

PR #2125, opened 2016-04-15
worker.openstack: Support novaclient 3.x

cmouse:
Adds client_version support, which is required for python-novaclient 3.x.

### DIFF
--- a/master/buildbot/test/fake/openstack.py
+++ b/master/buildbot/test/fake/openstack.py
@@ -23,10 +23,10 @@ ERROR = 'ERROR'
 UNKNOWN = 'UNKNOWN'
 
 
-# Parts used from novaclient.v1_1.
+# Parts used from novaclient
 class Client():
 
-    def __init__(self, username, password, tenant_name, auth_url):
+    def __init__(self, version, username, password, tenant_name, auth_url):
         self.images = Images()
         self.servers = Servers()
 
--- a/master/buildbot/test/unit/test_worker_openstack.py
+++ b/master/buildbot/test/unit/test_worker_openstack.py
@@ -91,7 +91,7 @@ class TestOpenStackWorker(unittest.TestCase):
 
         bs = openstack.OpenStackLatentWorker('bot', 'pass', flavor=1,
                                              image=image_callable, **self.os_auth)
-        os_client = novaclient.Client('user', 'pass', 'tenant', 'auth')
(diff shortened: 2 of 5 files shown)


tardyp:
Can you update the release notes about it?

cmouse:
AppVeyor failure due to older python-novaclient. Should there be some support for older versions?

tardyp:
Is newer client compatible with older openstack infra?

I am not very confident with openstack to know if its okay to drop support for old version.

Question we have to ask us is will this add regression for people, e.g. this worked for them before, and now it does not work anymore.

If the answer is no, then you can safely update the setup.py requirements to upgrade the required version of novaclient

cmouse:
it's more about the client library not being compatible. the 3.2.0/3.4.0 supports v1.1 infrastructure (at least claims to).

tardyp:
yes, but client library is just a pip install -U away, right?
It does not need anything else fancy installed by the distro?

cmouse:
it would seem so, yes.

tardyp:
Please have a look at the test:

test_worker_openstack.py

It implement a fake novaclient api, which has the 1.1 api.
So for this PR, you need to update the fake, so that it matches the 3.x novaclient api.

So remaining work is:

* [ ] update the fake to match 3.x api
* [ ] update the doc to explain that novaclient>=3.0.0 is required for this mode
* [ ] update the release note explaining that we now require updated version.

cmouse:
Can you provide more information how to relnotes?

tardyp:
CONTRIBUTING.rst links to http://trac.buildbot.net/wiki/SubmittingPatches
which contains all information about this

tardyp:
please make sure flake 8 pass

* [ ] make flake8

seankelly:
I am pretty sure this should work fine with pre-3.0 novaclient too. The `novaclient.client.Client` function exists since 2.6.1; that was released in June 2012 and its API hasn't changed since.

cmouse:
I'll check the 2.6 API and if it looks same, I can update docs and remove mention about 3.x

cmouse:
It seems so, so I will update docs to reflect this.

tardyp:
👍 The coverage miss is actually normal.

You can add # pragma no cover if you want your five checks

https://github.com/coagulant/coveralls-python#tips-for-coveragerc-config

cmouse:
I think it's better it fails so it gets fixed some day.

PR #2126, opened 2016-04-15
Update buildsummary directive to display reason of the build.

ssavrim:
Some users need to have the reason of the build directly in the buildsummary
view.
This patch get the 'reason' property of the build then display it in the
buildsummary view.
Use bootstrap grid to display panel containing: build/build#, reason ...
Refer to screenshot
display_reason

Change-Id: I4502eefff63b58b7a6d88ec4aa51ee4bba504d82

### DIFF
--- a/www/base/src/app/common/directives/buildsummary/buildsummary.tpl.jade
+++ b/www/base/src/app/common/directives/buildsummary/buildsummary.tpl.jade
@@ -1,7 +1,7 @@
 .panel.panel-default(ng-class="results2class(buildsummary.build)")
   .panel-heading.no-select
     .row
-      .col-sm-5
+      .col-sm-6
         .btn.btn-xs.btn-default(ng-click="buildsummary.toggleFullDisplay()", title="Expand all step logs")
             i.fa.fa-chevron-circle-right.rotate(ng-class="{'fa-rotate-90':buildsummary.fulldisplay}")
         .btn.btn-xs.btn-default(ng-click="buildsummary.toggleDetails()", title="Show steps according to their importance")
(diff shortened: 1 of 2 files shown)


www/base/src/app/common/directives/buildsummary/buildsummary.tpl.jade
```diff
@@ -1,32 +1,36 @@
...
-      .label(ng-class="results2class(buildsummary.build)")
-        | {{results2text(buildsummary.build)}}
-      span(ng-if="buildsummary.parentbuild")
-          | &nbsp;
-          a.label(ng-class="results2class(buildsummary.parentbuild)",
-          ui-sref="build({builder:buildsummary.parentbuilder.builderid, build:buildsummary.parentbuild.number})")
-                | {{buildsummary.parentrelationship}}: 
-                | {{buildsummary.parentbuilder.name}}/{{buildsummary.parentbuild.number}}
+    .row
+      .col-sm-5
```
tardyp: I dont think splitting the UI like this is very efficient.
I would prefer sticking with

- stuff in the left
- stuff in the right

So that if the reason is long it can take space from the status string, and the other way around.

tardyp:
Please rebase without mergecommit and update the screenshot.

www/base/src/app/common/directives/buildsummary/buildsummary.tpl.jade
```diff
@@ -1,32 +1,34 @@
...
-      .label(ng-class="results2class(buildsummary.build)")
-        | {{results2text(buildsummary.build)}}
-      span(ng-if="buildsummary.parentbuild")
-          | &nbsp;
-          a.label(ng-class="results2class(buildsummary.parentbuild)",
-          ui-sref="build({builder:buildsummary.parentbuilder.builderid, build:buildsummary.parentbuild.number})")
-                | {{buildsummary.parentrelationship}}:
-                | {{buildsummary.parentbuilder.name}}/{{buildsummary.parentbuild.number}}
+    .row
+      .col-sm-8
```
tardyp: you left col-sm-8 and col-sm8, which restrict the total size of each

ssavrim:
Here the new screenshot.
display_reason2

ssavrim:
Here is the new screenshot taking remarks into account.
display_reason2

www/base/src/app/common/directives/buildsummary/buildsummary.tpl.jade
ssavrim: By doing this the panel is more responsive.

ssavrim: By removing row and col text are not aligned anymore:
without_row_and_col

I propose to split the columns with same size (col-sm-6) as folllow:
with_row_and_col

tardyp:
I agree col-sm-8 looks better, indeed

tardyp:
keep the PR open to get more input from other

nam4dev:
:+1:

PR #2127, opened 2016-04-17
Bump angularjs dependencies

tardyp:
As we are very near the release, we won't be able to upgrade that much our JS dependencies, as people will start to create code against it.

I upgraded guanlecoja-ui to latest version of its dependencies, and here is the needed code in bb in order to have it working under those new deps

https://github.com/buildbot/guanlecoja-ui#changelog

* 1.6.0: Massive upgrade of dependencies:
  - jquery 2.1 -> 2.2.
  - Angular 1.4 -> 1.5.
  - Lodash -> 4.11.
  - Drop underscore.string as lodash got most important string utils.
  - angular-ui-bootstrap 0.11 -> 1.3.2: Most important change is that ui-bootstrap directives are now prefixed with "uib-" e.g: in jade: ".dropdown-toggle" -> ".dropdown-toggle(uib-dropdown-toggle)" This makes the markup more complicated, but the old 0.11 documentation is not even available anymore, so its best to upgrade.

I still need to test that heavily beyond unit tests as we are far from 100% coverage.

### DIFF
--- a/master/buildbot/data/buildsets.py
+++ b/master/buildbot/data/buildsets.py
@@ -86,6 +86,11 @@ class BuildsetsEndpoint(Db2DataMixin, base.Endpoint):
 
         @d.addCallback
         def db2data(buildsets):
+            # buildset db2data is pretty heavy. for the sake of console_view prefilter the data
+            # should fix it properly in http://trac.buildbot.net/ticket/3537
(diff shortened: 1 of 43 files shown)


tardyp:
did verify that this works for me (had to fix a bunch of things though)

PR #2128, opened 2016-04-18
fix documentation for collapseBuildRequest callback

tardyp:
While doing the documentation, I figured out it wasn't possible to write the same logic as before without a master reference.
So I added it in the API (right about time)

Note that the examples are indeed a bit more complicated :-/


delanne:
:+1:

master/docs/manual/cfg-builders.rst
```diff
@@ -136,17 +136,23 @@ Requests are only candidated for a merge if both requests have exactly the same
...
+    Requests will be collapsed if their sourcestamp are compatible (see below for definition of compatible).
 
-A configuration value of ``False`` indicates that requests should never be merged.
+``False``
+    Requests will never be collapsed.
 
-The configuration value can also be a callable, specifying a custom merging function.
-See :ref:`Collapse-Request-Functions` for details.
+``callable(builder, req1, req2)``
+    Requests will be collapsed if the callable returns true.
```
sa2ajj: [nit] `True`

PR #2129, opened 2016-04-18
fix schedulers page unwrapped elements

tardyp:
fix for http://trac.buildbot.net/ticket/3411

There is not much things to display in that page, but that can still help to know which master has which scheduler


tardyp:
unless there is objection, I will merge this trivial fix tomorrow.

PR #2130, opened 2016-04-18
Start using a non-global reactor.

tomprince:
It is much easier to test code that doesn't use the global reactor.

Change `BuildMaster` to take an optional reactor as a first step.

### DIFF
--- a/master/buildbot/test/util/integration.py
+++ b/master/buildbot/test/util/integration.py
@@ -81,6 +81,14 @@ class RunMasterBase(dirs.DirsMixin, unittest.TestCase):
                         configFunc=configFunc,
                         proto=proto,
                         workerclass=workerclass))
+
+        # mock reactor.stop (which trial *really* doesn't
+        # like test code to call!)
+        mock_reactor = mock.Mock(spec=reactor)
+        mock_reactor.callWhenRunning = reactor.callWhenRunning
+        mock_reactor.getThreadPool = reactor.getThreadPool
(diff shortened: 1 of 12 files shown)


master/buildbot/test/util/integration.py
```diff
@@ -81,6 +81,14 @@ def setupConfig(self, configFunc):
                         configFunc=configFunc,
                         proto=proto,
                         workerclass=workerclass))
+
+        # mock reactor.stop (which trial *really* doesn't
```
tardyp: shall we factorize those lines?

tardyp:
I think we should have a buildbot.unit.fake.reactor.Reactor class. good change apart from that!

tomprince:
> I think we should have a buildbot.unit.fake.reactor.Reactor class.

I've started on one.

master/buildbot/test/util/integration.py
tomprince: I'm inclined to wait until there is a fake reactor to do that.

tardyp: so this is for a followup PR?

tomprince: I plan to introduce a better fake reactor, and I think that any factorization of this code should use that fake reactor. I can't commit to adapting all the uses of this to use a fake reactor in timely fashion, though.

djmitche:
+1 from me -- good idea

PR #2131, opened 2016-04-18
Fix `trial -j`.

tomprince:
Both tests used to depend on mutation done in the second one. Change the tests so they don't globally modify the environment, and so both tests make the necessary change.

### DIFF
--- a/master/buildbot/test/unit/test_scripts_cleanupdb.py
+++ b/master/buildbot/test/unit/test_scripts_cleanupdb.py
@@ -43,6 +43,19 @@ def mkconfig(**kwargs):
     return config
 
 
+def patch_environ(case, key, value):
+    """
+    Add an environment variable for the duration of a test.
+    """
+    old_environ = os.environ.copy()
+
+    def cleanup():
+        os.environ.clear()
+        os.environ.update(old_environ)
+    os.environ[key] = value
+    case.addCleanup(cleanup)
+
+
 class TestCleanupDb(misc.StdoutAssertionsMixin, dirs.DirsMixin,
                     db.RealDatabaseMixin, unittest.TestCase):
 
@@ -86,6 +99,12 @@ class TestCleanupDb(misc.StdoutAssertionsMixin, dirs.DirsMixin,
 
     @defer.inlineCallbacks
     def test_cleanup_bad_config2(self):
+        # test may use mysql or pg if configured in env
+        if "BUILDBOT_TEST_DB_URL" not in os.environ:
+
+            patch_environ(self, "BUILDBOT_TEST_DB_URL", "sqlite:///" + os.path.join(self.origcwd,
+                                                                             "basedir", "state.sqlite"))
+
         self.createMasterCfg(extraconfig="++++ # syntaxerror")
         res = yield cleanupdb._cleanupDatabase(mkconfig(basedir='basedir'))
         self.assertEqual(res, 1)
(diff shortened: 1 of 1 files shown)


master/buildbot/test/unit/test_scripts_cleanupdb.py
```diff
@@ -43,6 +43,19 @@ def mkconfig(**kwargs):
     return config
 
 
+def patch_environ(case, key, value):
```
sa2ajj: this function is generic, but its both usages are very specific: the same environment variable name, the same value. does it really have to be that generic?

master/buildbot/test/unit/test_scripts_cleanupdb.py
```diff
@@ -43,6 +43,19 @@ def mkconfig(**kwargs):
...
+    """
+    Add an environment variable for the duration of a test.
+    """
+    old_environ = os.environ.copy()
+
+    def cleanup():
+        os.environ.clear()
+        os.environ.update(old_environ)
```
sa2ajj: what is the benefit of these two lines over simple
```python
os.environ = old_environ
```
?
`old_environ` is already a copy of the original `os.environ`.

master/buildbot/test/unit/test_scripts_cleanupdb.py
tomprince: I don't think that being more specific doesn't make the code any clearer, so the code might as well be this generic. Also. there are a number of other places in the test suite that appear to modify the environment that could benefit from this.

master/buildbot/test/unit/test_scripts_cleanupdb.py
tomprince: `os.environ` is weird, and not a real dictionary. Modifying it actually modifies the running processes environment (which gets inherited by subprocesses, for example). Doing `os.environ = old_environ` would replace it with a regular dictionary that wouldn't have that behavior.

master/buildbot/test/unit/test_scripts_cleanupdb.py
```diff
@@ -43,6 +43,19 @@ def mkconfig(**kwargs):
...
+    os.environ[key] = value
+    case.addCleanup(cleanup)
+
+
 class TestCleanupDb(misc.StdoutAssertionsMixin, dirs.DirsMixin,
                     db.RealDatabaseMixin, unittest.TestCase):
```
tardyp: We already have that much Mixin. why not having a PatchEnvironMixin ?

master/buildbot/test/unit/test_scripts_cleanupdb.py
```diff
@@ -86,6 +99,12 @@ def test_cleanup_bad_config(self):
 
     @defer.inlineCallbacks
     def test_cleanup_bad_config2(self):
+        # test may use mysql or pg if configured in env
+        if "BUILDBOT_TEST_DB_URL" not in os.environ:
+
+            patch_environ(self, "BUILDBOT_TEST_DB_URL", "sqlite:///" + os.path.join(self.origcwd,
+                                                                             "basedir", "state.sqlite"))
+
```
tardyp: again, we could use some factorization

tardyp:
good catch though!

master/buildbot/test/unit/test_scripts_cleanupdb.py
tomprince: I prefer to avoid inheritance if I can help it. Making this into a mixin doesn't provide any benefit over having a free function, so I just went with a free function.

master/buildbot/test/unit/test_scripts_cleanupdb.py
tomprince: I'm not sure if there is really any value in factoring these two lines out.

I think a better solution would be to actually stop mutating the environment, and instead just pass some parameters around. But `test_cleanup` calls `setUpRealDatabase` which also inspects the environment, and the tests appear to depend on this specific value, rather than the one `setUpRealDatabase`.

I'm not directly interested in cleaning up these tests completely, but I think that this change is an improvement on the current state.

PR #2132, opened 2016-04-18
don't remove SLAVEPASS from environment by default in buildbot-worker

rutsky:
These lines were introduced in 75f9ad7fc1dc by @pieterlexis.
I don't think these changes are required anymore.

Worker configuration created with `buildbot-worker create ...` doesn't
use environment variables to set password.

Worker configuration created with
master/contrib/docker/worker/Dockerfile optionally  uses environment
variable to specify password, but also has it's own code that clears
environment.

So I don't see any supported configuration that requires clearing of the
environment variables.

If *user* created configuration that uses environment variable to set
password, *he* must clear environment variables.

@pieterlexis please comment if you think I'm missing something.

### DIFF
--- a/worker/buildbot_worker/scripts/create_worker.py
+++ b/worker/buildbot_worker/scripts/create_worker.py
@@ -52,8 +52,6 @@ buildmaster_host = %(host)r
 port = %(port)d
 workername = %(name)r
 passwd = %(passwd)r
-if "SLAVEPASS" in os.environ:
-    del os.environ['SLAVEPASS']
 keepalive = %(keepalive)d
 usepty = %(usepty)d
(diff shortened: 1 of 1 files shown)


tardyp:
I think I remember it was because the pass appeared in some log.
I think it is better to erase at the same place as it is actually used.

rutsky:
@tardyp I think so too. Environment variables are outputted in each build; and I know only one place where `SLAVEPASS` is used (and renamed): https://github.com/buildbot/buildbot/blob/master/master/contrib/docker/worker/buildbot.tac#L21

tardyp:
indeed, git grep confirms it.

PR #2133, opened 2016-04-19
add section about buildbot-worker to slave/worker transition docs

rutsky:


rutsky:
This is ready to be merged.

PR #2137, opened 2016-04-19
install master[tls] extras on AppVeyor to fix service_identity module warning

rutsky:


rutsky:
Strange, warning still present:

```
UserWarning: You do not have a working installation of the service_identity module: 'No module named service_identity'.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module and a recent enough pyOpenSSL to support it, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected.
```

tardyp:
looks like you really need pip install service_identity

rutsky:
@tardyp we install `master[tls]` which installs `Twisted[tls] >= 14.0.0` which in my env installs `service_identity`.

rutsky:
But installing `master/[tls,test]` doesn't install `service_identity`...

rutsky:
It looks like this issue: https://github.com/pypa/pip/issues/3516

tardyp:
I think then we should make [tls] dupe what is in twisted setup.py, and do not rely on extra deps propagation

rutsky:
@tardyp where do suggest to do duplication?

We can split installation of `master[tls,docs]` by `master[tls]` and `master[docs]`, I think it should work this way.

tardyp:
in master/setup.py
[tls] should not be twisted[tls], but rather the list of deps that twisted[tls] specifies in its own setup.py

rutsky:
Then I'll add [currently used](https://github.com/twisted/twisted/blob/trunk/twisted/python/dist.py#L72):

```
    tls=['pyopenssl >= 0.13',
         'service_identity',
         'idna >= 0.6'],
```

with comment that this should be removed when pip will fix deps issue.

rutsky:
@tardyp done

PR #2139, opened 2016-04-19
Use sphinxcontrib-spelling to check docs spelling

rutsky:
This PR:

1. Adds `spelling` target to doc's Makefile that runs `sphinxcontrib-spelling` to check spelling.

2. Adds all words that currently considered as misspelled to `spelling_wordlist.txt` dictionary.

3. Adds configuration to run spellcheck on Travis.

4. Fixes `Buildbot` project name capital letter (or add appropriate quotations/semantic formatting like `:file:`, `:py:mod:`).

During fixing `Buildbot` I noticed that `sphinxcontrib-spelling` not always points to proper place of error in the source file and misses some words. This is not good, but better than nothing.
I noticed only one false positive: it doesn't handle words in URLs if URL is not explicitly marked as URL. E.g. `sphinxcontrib-spelling` considers error here:

```
visit http://nine.buildbot.net/ site
```

which may be workarounded by adding explicit formatting to URL:

```
visit `<http://nine.buildbot.net/>`__ site
```

I suggest to merge this PR with ignored most of typos and fix those typos/formatting in separate pull requests.

I propose not to only fix typos, but also semantic formatting (`:py:mod:`, `:file:`), case sensitivity (`URL` instead of `url`) and proper project names (`AppVeyor` instead of `Appveyor`).

Also when someone will fix typos I suggest to do `git grep some-typo` to fix this typo not only in docs, but in source code/scripts.


ssavrim:
:+1:

.travis.yml
```diff
@@ -3,6 +3,12 @@
 
 addons:
   postgresql: "9.4"
+  apt_packages:
+    - enchant
+    - aspell
+    - aspell-en
+    - ispell
+    - iamerican
```
tardyp: I would prefer that we do not enable sphinx-spelling if those packages are not found.

The basic pip install ./master[docs] should always work on any system.

We have ansible scripts to build the doc, which we would need to upgrade

rutsky: @tardyp good point, thanks!

`pyenchant` can be installed without supporting `libenchant` (at least current version of `pyenchant`), so `pip install ./master[docs]` should work on any machine.

But docs building will fail with `ImportError: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI.`

I'll add something like this to `conf.py`:

```python
try:
    import enchant
except ImportError as ex:
    print("Enchant import failed: {0}\nSpell checking disabled.".format(ex), file=sys.stderr)
else:
    extensions.append('sphinxcontrib.spelling')

**Comment by rutsky** [comment]
@tardyp I added conditional switching of `sphinxcontrib.spelling`, now docs building works without Enchant library: https://gist.github.com/rutsky/75a9873bbf3150f7e94f6790cdb691ab

File: master/docs/conf.py [hunk 2]
```diff
@@ -275,3 +277,16 @@
...
+    import enchant as _  # noqa
+except ImportError as ex:
+    print("enchant module import failed:\n"
+          "{0}\n"
+          "Spell checking disabled.".format(ex),
+          file=sys.stderr)
+
+else:
+    extensions.append('sphinxcontrib.spelling')
+    spelling_show_suggestions = True
```
rutsky: @tardyp here is conditional switching of `sphinxcontrib.spelling`

tardyp:
pylint is unhappy.

You need to use your imports with a one line statement like

    import myunusedimport
    [myunusedimport]

rutsky:
pylint should be happy now, lets wait for CI to finish.

PR #2141, opened 2016-04-19
[WIP] Handle build_wait_timeout >= 0 consistently

aelsabbahy:
WIP, looking for feedback on approach.

Fixes [#3528](http://trac.buildbot.net/ticket/3528)

### DIFF
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -776,7 +776,7 @@ class AbstractLatentWorker(AbstractWorker):
         self.building.remove(sb.builder_name)
         if not self.building:
             if self.build_wait_timeout == 0:
-                d = self.insubstantiate()
+                d = self.insubstantiate(disconnect=True)
                 # try starting builds for this worker after insubstantiating;
                 # this will cause the worker to re-substantiate immediately if
                 # there are pending build requests.
@@ -799,7 +799,7 @@ class AbstractLatentWorker(AbstractWorker):
             self.build_wait_timeout, self._soft_disconnect)
 
     @defer.inlineCallbacks
-    def insubstantiate(self, fast=False):
+    def insubstantiate(self, fast=False, disconnect=False):
         self.insubstantiating = True
         self._clearBuildWaitTimer()
         d = self.stop_instance(fast)
@@ -810,6 +810,8 @@ class AbstractLatentWorker(AbstractWorker):
         self.substantiated = False
         self.building.clear()  # just to be sure
         yield d
+        if disconnect:
+            yield AbstractWorker.disconnect(self)
         self.insubstantiating = False
         self.botmaster.maybeStartBuildsForWorker(self.name)
 


tomprince:
I don't think this a correct fix. I think the intent of `build_wait_timeout == 0` is that the worker will only run a single build. I don't think this change preserves that behavior.

aelsabbahy:
Hmm, that's definitely not what I want then. Since it's very important for us that we only have single use workers.

Would a `d.addCallback(self._soft_disconnect)` right after the `d = self.insubstantiate()` preserve the behavior while properly disconnecting?

tomprince:
> Would a d.addCallback(self._soft_disconnect) right after the d = self.insubstantiate() preserve the behavior while properly disconnecting?

I'm not sure. The code here is somewhat subtle and I suspect has a number of undocumented dependencies and interactions. I don't think I would be comfortable with any changes that don't explain why the change is correct.

It certainly isn't clear to me that this is any better (or worse) than the existing code.

aelsabbahy:
K, I'm taking a slightly different approach which might be a little more clear and direct, which is basically to ensure that a worker is disconnected after calling d.insubstantiate().

For some reason, our EC2 workers aren't disconnecting properly with the code that's in 20cac3f735b526de13d007a59bf47f58e2dffa17

The issue we're running into today is basically:
* Build queued
* Worker starts
* Worker terminated
* Worker shows up as status: 1 connection, even though the EC2 instance is terminated

master/buildbot/worker/base.py
```diff
@@ -781,6 +781,8 @@ def buildFinished(self, sb):
         if not self.building:
             if self.build_wait_timeout == 0:
                 d = self.insubstantiate()
+                d.addCallback(lambda _:
+                              AbstractWorker.disconnect(self))
```
tardyp: I would believe insubstantiate is supposed to do the diconnection.

aelsabbahy: Is that something insubstantiate should always do, or should it be a kwarg defaulting to False?

I didn't put it in insubstantiate, because I wasn't sure on the intended usage of the method and didn't want to break something else unintentionally.

tomprince: It isn't clear that anybody actually knows what any of this is supposed to do.

aelsabbahy: This change makes sense to me to be added to insubstantiate, just not 100% sure if it should be kwarg flagged or not.

Also, I noticed insubstantiate() is calling `self.botmaster.maybeStartBuildsForWorker(self.name)` at the end, and we turn around and call that method again here. If we move the disconnect logic to insubstantiate can we drop the call to `self.botmaster.maybeStartBuildsForWorker(self.workername)` from this method?

tomprince: > This change makes sense to me to be added to insubstantiate, [...]

It might. But I'd be inclined to reject any change to this code that is "I think this makes sense ..." instead of "This makes sense because a) ... b) ... c) ...". I suspect the code is in the state that it is largely because of changes like the former.

aelsabbahy:
For what it's worth, I've updated this pull request with what we're currently using in our CI environment. Without this change build_wait_timeout=0 is causing instances to terminate but not disconnect or ever relaunch so our server just hangs.

tardyp:
@aelsabbahy How can the instance terminate with the tcp connection being closed as well?

tardyp:
@aelsabbahy ping

aelsabbahy:
Sorry for the late response, I think this PR can be closed out all together.

I believe some of the changes @tomprince made conflicted with the changes here. There were some initial discussions and @tomprince  had some thoughts on how to handle this use-case properly/thoroughly, but I'm not 100% sure if any action was taken since then.

PR #2143, opened 2016-04-19
include raml files onto Buildbot distribution

rutsky:
@tardyp I haven't included `indent.py` in Buildbot distribution. Do we need it?


rutsky:
@tardyp what is the purpose of [indent.py](https://github.com/buildbot/buildbot/blob/master/master/buildbot/spec/indent.py)?

tardyp:
it is to make sure that the raml files are indented by 4.

some tools are using 2 spaces indents, and then its very painful to fix manually

PR #2144, opened 2016-04-19
Make sample.cfg more declarative

jml:
This patch changes the sample.cfg to just define a dictionary, rather than building one up piecemeal over several instructions.

It's partly a matter of personal taste, but I think this style makes the structure more clear. In particular, it makes it very obvious that none of the keys bare any relationship to the others. It also shifts it closer to the 'data' end of the spectrum than the 'code' end of the spectrum, which makes it simpler.

I've taken care not to reflow lines or make any other changes.

### DIFF
--- a/master/buildbot/scripts/sample.cfg
+++ b/master/buildbot/scripts/sample.cfg
@@ -6,99 +6,100 @@ from buildbot.plugins import *
 # This is a sample buildmaster config file. It must be installed as
 # 'master.cfg' in your buildmaster's base directory.
 
-# This is the dictionary that the buildmaster pays attention to. We also use
-# a shorter alias to save typing.
-c = BuildmasterConfig = {}
+# This is the dictionary that the buildmaster pays attention to.
+BuildmasterConfig = {
+    ####### WORKERS
 
-####### WORKERS
+    # The 'workers' list defines the set of recognized workers. Each element is
+    # a Worker object, specifying a unique worker name and password.  The same
+    # worker name and password must be configured on the worker.
+    'workers': [worker.Worker("example-worker", "pass")],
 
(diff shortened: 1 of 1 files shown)


djmitche:
The `c['..']` is how we recommend config files be structured, and how the config entries are referred to in the docs.  Also, in many cases the config is generated programmatically -- it really *isn't* data in any reasonably sophisticated configuration.  See https://github.com/buildbot/metabbotcfg for example.  Fitting that into a dictionary literal can end up being pretty ugly.

So I'm inclined to say "no" here, but I'll certainly let others chime in.

tomprince:
Looking at https://github.com/buildbot/metabbotcfg/blob/master/master.cfg it could be turned into
```python
BuildmasterConfig = {
    'slaves': slaves,
    'schedulers': schedulers,
    'builders': builders,
    'status': status,
    'slavePortnum': 9989,
    'projectName': "Buildbot",
    'projectURL': "http://buildbot.net/",
    'buildbotURL': "http://buildbot.buildbot.net/",
    'changeHorizon': 200,
    'buildHorizon': 100,
    'logHorizon': 100,
    'eventHorizon': 10,
    'buildCacheSize': 300,
}
```
without any difficulty.

I think this would be a good change, but would require changing all the documentation to match this style.

jml:
I'm not going to update all the docs, I'm afraid. Thanks for taking the time to consider the PR though!

tardyp:
I note that this is the second similar PR in a few months.
https://github.com/buildbot/buildbot/pull/2051

While we agree for very simple config, this proposed form looks more pythonic, the current form makes it much easier to document snippets that people can just copy paste inside their config.

We encourage people to embrace the api, and use python as they like in order to implement their own config.

PR #2148, opened 2016-04-20
Split config loading so tests can just load a dictionary.

tomprince:


master/buildbot/config.py
```diff
@@ -277,6 +305,7 @@ def loadConfig(cls, basedir, filename):
         if errors:
             raise errors
 
+        print type(config)
```
tardyp: debug print?

master/buildbot/master.py
```diff
@@ -228,7 +238,7 @@ def startService(self):
                 # run the master.cfg in thread, so that it can use blocking code
                 self.config = yield threads.deferToThreadPool(
```
tardyp: I think we should put the deferToThread in FileLoader.

I would *love* to be able to write asynchronous config_loader

tardyp:
I like those changes a lot!

This opens up a lot of cool stuff that I cannot measure the full scale yet.

I would like at some point to be able to say in my buildbot.tac:

configFunction="buildbot_travis.configurator"

And then configurator can fetch stuff asynchronously in the db.
For now, the way I did it is the a big hack. There is a chicken and the egg problem on the config of the db_url.

master/buildbot/master.py
tomprince: That seems like a reasonable change, but larger than I want to do at this point. The current change makes it easy to test code without needing to generate python code and evaluate it, just to pass an existing dictionary around.

In particular, doing that requires a fair amount of refactoring tests to support deferreds.

tomprince:
How about
```python
from buildbot_travis import TravisLoader
...
....(..., config_loader=TravisLoader(...))
```
instead?

tardyp:
```
from buildbot_travis import TravisLoader
...
....(..., config_loader=TravisLoader(...))
```
Yes, this can work. I can look later the details on how we can make this framework allow defereable config loader

tardyp:
👍

PR #2152, opened 2016-04-21
add --wait option for `buildbot stop`

rutsky:


tardyp:
👍

Stibbons:
we actually made it the default behavior. We could have a "--no-wait" as an option

rutsky:
I like you idea @Stibbons, thanks!

Any objections on making `--wait` default behaviour and adding `--no-wait` instead?

rutsky:
@tardyp, @Stibbons I pushed `--no-wait` version, please take a look.

Stibbons:
I like :)

tardyp:
thx

PR #2155, opened 2016-04-21
Allow slaves to substantiate in parallel

morrone:
BuildRequestDistributor's _maybeStartBuildsOnBuilder() method is
decorated with @defer.inlineCallbacks.  This means that when it
uses "yield bldr.maybeStartBuild", the entire slave substantiation
process is sequentialized at that point.

slave substantiation can take considerable time, especially with
something like EC2 latent slaves.  Even without anything substantial
happening in the cloud init (user-data) script, it can easily take
a couple of minutes.  If any substantial configuration of the
OS takes place ins the user-data script, maybeStartBuild() can
easily take many minutes.

To enable parallel slave substantiation, we need to stop waiting on
bldr.maybeStartBuild() inline.  Instead we move the error
handling code into a callback function that can be added to the
Deferred that is returned by bldr.maybeStartBuild().

At least for for the EC2LatentBuildSlave class, everything else down
the chain is properly event based and/or threaded.  This small
code change allows slaves to substantiate in parallel.

Signed-off-by: Christopher J. Morrone <morrone2@llnl.gov>

### DIFF
--- a/master/buildbot/process/buildrequestdistributor.py
+++ b/master/buildbot/process/buildrequestdistributor.py
@@ -505,12 +505,15 @@ class BuildRequestDistributor(service.AsyncMultiService):
                 bc = self.createBuildChooser(bldr, self.master)
                 continue
 
-            buildStarted = yield bldr.maybeStartBuild(worker, breqs)
-            if not buildStarted:
-                yield self.master.data.updates.unclaimBuildRequests(brids)
-                # try starting builds again.  If we still have a working worker,
-                # then this may re-claim the same buildrequests
-                self.botmaster.maybeStartBuildsForBuilder(self.name)
+            d = bldr.maybeStartBuild(worker, breqs)
+            @defer.inlineCallbacks
+            def checkBuildStart(buildStarted, brids, builderName):
+                if not buildStarted:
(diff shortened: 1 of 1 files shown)


tomprince:
Something like this makes sense, but the issue with this is that the buildrequest will be claimed without a corresponding build while the worker starts up.

I am finishing up a branch that will implement a solution to this that doesn't suffer from that drawback.

morrone:
OK, thanks.  I'm new to the code; can you explain why that is a problem?  It seems like the builds are associated during worker startup just as they would be previously, only now multiple will be happening in parallel.

I'll look forward to seeing you branch.

tomprince:
The code is [here](https://github.com/buildbot/buildbot/commit/0cb89529b4143c6e7d795bd2469c7deb3698dbd6). I plan to submit it (and the dependencies) in several branches.

aelsabbahy:
@tomprince Glad you're doing this, this should be a huge improvement for EC2 latent workers.

If I want to test your code, do you suggest I test the commit you linked, or is there a particular branch I should pull down?

tomprince:
@aelsabbahy https://github.com/buildbot/buildbot/compare/master...tomprince:latent-with-a-Build is the branch.

PR #2157, opened 2016-04-22
fix TestCase.assertRegexpMatches() Python 2.6 backport

rutsky:
Nobody used assertRegexpMatches() in Buildbot by the way,
maybe because of this error :)

Reference: https://docs.python.org/2.6/library/re.html#re.search

### DIFF
--- a/master/buildbot/monkeypatches/testcase_assert.py
+++ b/master/buildbot/monkeypatches/testcase_assert.py
@@ -52,7 +52,7 @@ def _assertRegexpMatches(self, text, regexp, msg=None):
     introduced in python 2.7. The goal for this function is to behave exactly
     as assertRegexpMatches() in standard library.
     """
-    if not re.search(text, regexp):
+    if not re.search(regexp, text):
         if msg is not None:
             self.fail(msg)
         else:


rutsky:
This is not the first time I see this test failing:

```
[FAIL] 
  File "C:\Python27\lib\site-packages\twisted\internet\defer.py", line 1128, in _inlineCallbacks
    result = g.send(result)
  File "c:\projects\buildbot\master\buildbot\test\unit\test_www_hooks_poller.py", line 76, in test_no_args
    self.assertEqual(self.otherpoller.called, True)
  File "C:\Python27\lib\site-packages\twisted\trial\_synctest.py", line 437, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "C:\Python27\lib\unittest\case.py", line 513, in assertEqual
  File "C:\Python27\lib\unittest\case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: False != True
 
buildbot.test.unit.test_www_hooks_poller.TestPollingChangeHook.test_no_args
```

Is this test flaky or just buggy?

tardyp:
@rutsky it looks like this is no flaky. 100% of the PR have this issue since a few days

PR #2159, opened 2016-04-22
Fix the MsBuild steps

SpecLad:
These steps use a list as a command, despite needing shell processing. This broke after Windows argument escaping was made more strict in 693a2cb. Make them use a single string instead.

I also added quoting for the various user-supplied parameters, so that e.g. project paths with spaces in them are processed correctly.

Fixes <http://trac.buildbot.net/ticket/2878>.

### DIFF
--- a/master/buildbot/steps/vstudio.py
+++ b/master/buildbot/steps/vstudio.py
@@ -388,7 +388,7 @@ VS2013 = VC12
 
 class MsBuild4(VisualStudio):
     platform = None
-    vcenv_bat = "\"${VS110COMNTOOLS}..\\..\\VC\\vcvarsall.bat\""
+    vcenv_bat = r"${VS110COMNTOOLS}..\..\VC\vcvarsall.bat"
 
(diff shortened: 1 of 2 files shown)


tardyp:
thanks for the fix!

I would expect that this is also a bug in nine. Please submit first in nine, as we are not sure if we are even releasing a new version of eight

sa2ajj:
@SpecLad if you'd like this change to be maintained further, please submit your fix to `master` (aka `nine`) first.

tardyp:
Since it was merged in nine, I remove nine-first, and will let @sa2ajj  merge it.

sa2ajj:
I believe the only missing thing is the release notes.

SpecLad:
Should I include a release note for this?

sa2ajj:
@SpecLad I just reviewed the change again.  The behaviour is what people *would* expect.  So the "release notes" comment would "now it does what we said it would" :)

PR #2160, opened 2016-04-23
log is master startup fails + handling in `buildbot start`

rutsky:
Now `buildbot start` handles cases when master startup failed and explicitly say that something failed.

By the way, how about I remove `_master_initialized` from BuildMaster?
It's not used anywhere and previously has incorrect value if master startup failed.

### DIFF
--- a/master/buildbot/scripts/start.py
+++ b/master/buildbot/scripts/start.py
@@ -18,6 +18,7 @@ import os
 import sys
 
 from buildbot.scripts import base
+from buildbot.scripts.logwatcher import BuildmasterStartupError
 from buildbot.scripts.logwatcher import BuildmasterTimeoutError
 from buildbot.scripts.logwatcher import LogWatcher
 from buildbot.scripts.logwatcher import ReconfigError
@@ -55,6 +56,10 @@ line that says 'BuildMaster is running' to verify correct startup.
 The buildmaster appears to have encountered an error in the master.cfg config
 file during startup. Please inspect and fix master.cfg, then restart the
 buildmaster.
+""")
+        elif why.check(BuildmasterStartupError):
+            print("""
+The buildmaster startup failed. Please see 'twistd.log' for possible reason.
 """)
         else:
             print("""
--- a/master/buildbot/master.py
+++ b/master/buildbot/master.py
@@ -231,6 +231,7 @@ class BuildMaster(service.ReconfigurableServiceMixin, service.MasterService,
         self.reactor.callWhenRunning(d.callback, None)
         yield d
 
+        startup_succeed = False
         try:
             # load the configuration file, treating errors as fatal
             try:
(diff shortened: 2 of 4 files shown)


master/buildbot/scripts/start.py
```diff
@@ -56,6 +57,10 @@ def _failure(self, why):
 file during startup. Please inspect and fix master.cfg, then restart the
 buildmaster.
 """)
+        elif why.check(BuildmasterStartupError):
+            print("""
+The buildmaster startup failed. Please see 'twistd.log' for possible reason.
```
tardyp: please use textwrap.dedent

master/buildbot/scripts/start.py
```diff
@@ -56,6 +57,10 @@ def _failure(self, why):
...
+""")
         else:
             print("""
 Unable to confirm that the buildmaster started correctly. You may need to
```
tardyp: here also

master/buildbot/scripts/start.py
rutsky: @tardyp how about I'll use `textwrap.dedent` in a following pull request? It's worth to use it not only in my changes, but in other parts of the code, which I don't want to mix with in this PR that fixes master startup failure handling.

tardyp: Well we can first do that on the code that we rework, as we do for inlineCallbak

rutsky: I played with `textwrap.dedent`/`textwrap.fill` and now I think that currently wrapped text is the best approach.

Using just `textwrap.dedent` we have problem of formatting multiline text --- we need to fit in 80 characters limit of source code. So instead of 

```python
       elif why.check(BuildmasterStartupError):
            print("""
The buildmaster startup failed. Please see 'twistd.log' for possible reason.
""")
```

I need to write:

```python
        elif why.check(BuildmasterStartupError):
            print(textwrap.dedent("""
                The buildmaster startup failed. Please see 'twistd.log' for
                possible reason.
                """))
```

which does unwanted split of text that outputted to the console.

Ok, here we can use `textwrap.fill` --- it removes unnecessary spaces and wraps at 80 column (customizable), so it doesn't matter of how we format message in the source code:

```python
        elif why.check(BuildmasterStartupError):
            print(textwrap.fill(
                """The buildmaster startup failed. Please see 'twistd.log' for
                possible reason.
                """, 80))
```

but this is not an option [when](https://github.com/buildbot/buildbot/blob/master/worker/buildbot_worker/scripts/start.py#L48) we actually want to preserve formatting:

```python
        if why.check(WorkerTimeoutError):
            print("""
The worker took more than 10 seconds to start and/or connect to the
buildmaster, so we were unable to confirm that it started and connected
correctly. Please 'tail twistd.log' and look for a line that says 'message
from master: attached' to verify correct startup. If you see a bunch of
messages like 'will retry in 6 seconds', your worker might not have the
correct hostname or portnumber for the buildmaster, or the buildmaster might
not be running. If you see messages like
   'Failure: twisted.cred.error.UnauthorizedLogin'
then your worker might be using the wrong botname or password. Please
correct these problems and then restart the worker.
""")
```

We can't do auto-filling here with `textwrap.join` (we will loose formatting) and I don't want to effectively limit outputted message width due to source formatting.

Another option would be to move messages to module-level constants, which doesn't improves readability, IMO.

@tardyp what do you think?

rutsky: @tardyp take a look at #2172

tardyp:
ok to continue the work in #2172

PR #2163, opened 2016-04-24
Tweak build change retrieval to not error for 'None' change

rayalan:
getChangeFromSSid() can return None, but the change retrieval for a build did
not work properly in that scenario.

---

I took a look at adding a unit test case, but I can't figure out how to recreate this problem in a controlled environment (specifically, the environment of `test_db_changes` / `test_getChangesForBuild`). All my attempts to recreate lead to database integrity issues or don't reproduce.

If someone has pointers on unit tests, I can take another look at adding them.

### DIFF
--- a/master/buildbot/db/changes.py
+++ b/master/buildbot/db/changes.py
@@ -195,7 +195,9 @@ class ChangesConnectorComponent(base.DBConnectorComponent):
 
         # For each codebase, append changes until we match the parent
         for cb, change in iteritems(fromChanges):
-            if change and change['changeid'] != toChanges.get(cb, {}).get('changeid'):
+            # Careful; toChanges[cb] may be None from getChangeFromSSid
+            toCbChange = toChanges.get(cb) or {}
+            if change and change['changeid'] != toCbChange.get('changeid'):
                 changes.append(change)
                 while ((toChanges.get(cb, {}).get('changeid') not in change['parent_changeids']) and
                        change['parent_changeids']):


master/buildbot/db/changes.py
```diff
@@ -195,7 +195,8 @@ def getChangesForBuild(self, buildid):
 
         # For each codebase, append changes until we match the parent
         for cb, change in iteritems(fromChanges):
-            if change and change['changeid'] != toChanges.get(cb, {}).get('changeid'):
+            # Careful; toChanges[cb] may be None from getChangeFromSSid
+            if change and change['changeid'] != (toChanges.get(cb) or {}).get('changeid'):
```
tardyp: get(cb, {})

cmouse: isn't that the original line?

rayalan: Yes, `get(cb, {})` is the original line; it doesn't work because of this:

```
toChanges[cb] = None
badValue = toChanges.get(cb, {})  # badValue = None 
badValue.get('changeId')  # e.g. None.get(), which throws an exception
```

I'm not sure what @tardyp's comment was meant to suggest, but the code as is doesn't reliably work unless changes are made elsewhere to guarantee a dictionary rather than None is returned.

tardyp: Indeed.

    .get(cb) or {}

is not strictly equivalent to

    .get(cb, {})

As it might return None if [cb] is defined to None.
I am not very fond of this way to write it.

Can we calculate toChangeId before the loop, so that this is easier to read?

rayalan:
@tardyp I added a toCbChange calculation for each iteration of the loop; do you like this iteration better?

master/buildbot/db/changes.py
```diff
@@ -195,7 +195,8 @@ def getChangesForBuild(self, buildid):
 
         # For each codebase, append changes until we match the parent
         for cb, change in iteritems(fromChanges):
-            if change and change['changeid'] != toChanges.get(cb, {}).get('changeid'):
+            toCbChange = toChanges.get(cb) or {}  # Careful; toChanges[cb] may be Nnoe from getChangeFromSSid
```
tardyp: There is a typo in the comment + the line is > 110 char. 
Please put the comment on the previous line

looks good apart from that.

tardyp:
thanks

PR #2165, opened 2016-04-26
reporters.gitlab: Use correct state

cmouse:
Fix mistake in code that uses incorrect state for gitlab.


cmouse:
Cannot understand the failure :(

PR #2168, opened 2016-04-27
Publish build started events

ryansydnor:
Lots of reporters (http, gerrit, etc) are listening to the ("build", None, "started") tuple, which is no longer being published. This brings that functionality back.


tardyp:
They should listen for "new" event as per the documentation that I have not yet written :)

ryansydnor:
@tardyp what if I need build properties?

PR #2169, opened 2016-04-27
Fix the MsBuild steps

SpecLad:
These steps use a list as a command, despite needing shell processing. This broke after Windows argument escaping was made more strict in 693a2cb. Make them use a single string instead.

Fixes <http://trac.buildbot.net/ticket/2878>.

### DIFF
--- a/master/buildbot/steps/vstudio.py
+++ b/master/buildbot/steps/vstudio.py
@@ -415,7 +415,7 @@ VS2013 = VC12
 
 class MsBuild4(VisualStudio):
     platform = None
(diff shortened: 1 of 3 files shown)


SpecLad:
This #2159, ported to nine.

tardyp:
looks good!

Tsubashi:
This looks good. We ran across the same issue, but if I recall correctly we solved it by hardcoding the path, and I wasn't about to submit that as a pull request. ^.^

tardyp:
thanks for the windows collaboration!

PR #2170, opened 2016-04-27
Delay sending 'new' build message until properties are populated

ryansydnor:


djmitche:
I'd like to see a call to `master.data.updates.setBuildProperties` in here first -- otherwise the build doesn't actually have its properties yet.

djmitche:
That looks better :)

@tardyp, what do you think?

tardyp:
looks good!

PR #2172, opened 2016-04-28
rewrap outputted to the console text

rutsky:
This PR based on #2160, lets deal with that one first.
<hr>

We want:

1. to have strings properly formatted in the source code;

2. to have outputted to the terminal strings properly wrapped;

3. to preserve formatting of some parts of outputted to the terminal strings.

Currently strings are not well formatted in the source code:

```python
        if why.check(WorkerTimeoutError):
            print("""
The worker took more than 10 seconds to start and/or connect to the
buildmaster, so we were unable to confirm that it started and connected
correctly. Please 'tail twistd.log' and look for a line that says 'message
from master: attached' to verify correct startup. If you see a bunch of
messages like 'will retry in 6 seconds', your worker might not have the
correct hostname or portnumber for the buildmaster, or the buildmaster might
not be running. If you see messages like
   'Failure: twisted.cred.error.UnauthorizedLogin'
then your worker might be using the wrong botname or password. Please
correct these problems and then restart the worker.
""")
```

If we use `textwrap.dedent` here strings will not be properly wrapped,
instead of being wrapped by 80 column, they will be wrapped by 62 column in the console:

```python
        if why.check(WorkerTimeoutError):
            print(textwrap.dedent("""\
                The worker took more than 10 seconds to start and/or connect
                to the buildmaster, so we were unable to confirm that it
                started and connected correctly.
                Please 'tail twistd.log' and look for a line that says
                'message from master: attached' to verify correct startup.
                If you see a bunch of messages like 'will retry in 6 seconds',
                your worker might not have the correct hostname or portnumber
                for the buildmaster, or the buildmaster might not be running.
                If you see messages like
                   'Failure: twisted.cred.error.UnauthorizedLogin'
                then your worker might be using the wrong botname or password.
                Please correct these problems and then restart the worker.
                """))
```

If we use `textwrap.fill` to rewrap paragraphs we will get properly filled
by width strings, but will loose formatting (e.g. of
`  'Failure: twisted.cred.error.UnauthorizedLogin'` in the example above).

`rewrap` function combines `textwrap.dedent` with rewrapping with
`textwrap.fill` paragraphs that doesn't contain raw code blocks
(denoted by spaces at the begin of a line).

### DIFF
--- a/master/buildbot/scripts/reconfig.py
+++ b/master/buildbot/scripts/reconfig.py
@@ -20,6 +20,7 @@ from buildbot.scripts.logwatcher import BuildmasterTimeoutError
 from buildbot.scripts.logwatcher import LogWatcher
 from buildbot.scripts.logwatcher import ReconfigError
 from buildbot.util import in_reactor
+from buildbot.util import rewrap
 
 from twisted.internet import reactor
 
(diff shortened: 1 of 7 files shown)


tardyp:
mhh interresting problem and solution.
👍

PR #2174, opened 2016-04-28
Stop wedging a worker when it fails to substantiate.

tomprince:
I discovered this writing the integration tests here.

### DIFF
--- a/master/buildbot/test/fake/reactor.py
+++ b/master/buildbot/test/fake/reactor.py
@@ -1,0 +1,110 @@
+# Copyright Buildbot Team Members
+# Portions copyright 2015-2016 ClusterHQ Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+from zope.interface import implementer
+
+from twisted.internet.base import _ThreePhaseEvent
+from twisted.internet.interfaces import IReactorCore
+from twisted.internet.interfaces import IReactorThreads
+from twisted.internet.task import Clock
+from twisted.python.failure import Failure
+
+# The code here is based on the implementations in
+# https://twistedmatrix.com/trac/ticket/8295
+# https://twistedmatrix.com/trac/ticket/8296
+
+
+@implementer(IReactorCore)
+class CoreReactor(object):
+    """
+    Partial implementation of ``IReactorCore``.
+    """
+    def __init__(self):
+        self._triggers = {}
+
+    def addSystemEventTrigger(self, phase, eventType, f, *args, **kw):
+        event = self._triggers.setdefault(eventType, _ThreePhaseEvent())
+        return eventType, event.addTrigger(phase, f, *args, **kw)
+
+    def removeSystemEventTrigger(self, triggerID):
+        eventType, handle = triggerID
+        event = self._triggers.setdefault(eventType, _ThreePhaseEvent())
+        event.removeTrigger(handle)
+
+    def fireSystemEvent(self, eventType):
+        event = self._triggers.get(eventType)
+        if event is not None:
+            event.fireEvent()
+
+    def callWhenRunning(self, f, *args, **kwargs):
+        f(*args, **kwargs)
+
+
+class NonThreadPool(object):
+    """
+    A stand-in for ``twisted.python.threadpool.ThreadPool`` so that the
+    majority of the test suite does not need to use multithreading.
+
+    This implementation takes the function call which is meant to run in a
+    thread pool and runs it synchronously in the calling thread.
+
+    :ivar int calls: The number of calls which have been dispatched to this
+        object.
+    """
+    calls = 0
+
+    def __init__(self, **kwargs):
+        pass
+
+    def callInThreadWithCallback(self, onResult, func, *args, **kw):
+        self.calls += 1
+        try:
+            result = func(*args, **kw)
+        except:
+            onResult(False, Failure())
+        else:
+            onResult(True, result)
+
+    def start(self):
+        pass
+
+    def stop(self):
+        pass
+
+
+@implementer(IReactorThreads)
+class NonReactor(object):
+    """
+    A partial implementation of ``IReactorThreads`` which fits into
+    the execution model defined by ``NonThreadPool``.
+    """
+    def callFromThread(self, f, *args, **kwargs):
+        f(*args, **kwargs)
+
+    def getThreadPool(self):
+        return NonThreadPool()
+
+
+class TestReactor(NonReactor, CoreReactor, Clock):
+    def __init__(self):
+        Clock.__init__(self)
+        CoreReactor.__init__(self)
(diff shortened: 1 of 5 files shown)


master/buildbot/test/fake/reactor.py
```diff
@@ -0,0 +1,110 @@
+# Copyright Buildbot Team Members
+# Portions copyright 2015-2016 ClusterHQ Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
```
tomprince: This is the twisted licenses, which is where this code is copied from.

master/buildbot/test/integration/test_latent.py
```diff
@@ -0,0 +1,224 @@
...
+
+from twisted.internet.defer import Deferred
+from twisted.python import threadpool
+from twisted.python.failure import Failure
+from twisted.trial.unittest import SynchronousTestCase
+
+
+class LatentController(object):
+    """
+    A controller for ``ControllableLatentWorker``.
```
tomprince: https://glyph.twistedmatrix.com/2015/05/separate-your-fakes-and-your-inspectors.html

master/buildbot/test/fake/reactor.py
```diff
@@ -0,0 +1,110 @@
...
+        f(*args, **kwargs)
+
+    def getThreadPool(self):
+        return NonThreadPool()
+
+
+class TestReactor(NonReactor, CoreReactor, Clock):
+    def __init__(self):
+        Clock.__init__(self)
+        CoreReactor.__init__(self)
```
tardyp: its really all what is needed to make a fake synchronous reactor?

Can we synchronously read from files? or from network with this? Or this is just about threadpools?

master/buildbot/test/integration/test_latent.py
tardyp: I think this makes sense to add this as a in-code comment.

That blog post is great

master/buildbot/test/integration/test_latent.py
```diff
@@ -0,0 +1,224 @@
...
+        config_dict = {
+            'builders': [
+                BuilderConfig(name="testy",
+                              workernames=["local"],
+                              factory=BuildFactory(),
+                              ),
+            ],
+            'workers': [controller.worker],
+            'protocols': {'null': {}},
+            'multiMaster': True,
```
tardyp: I assume you use this to disable additional coherency tests (no scheduler for testy).
 I think this would make sense to have a 'disableCoherencyChecks': True, and make 'multiMaster' an alias for that.\

master/buildbot/test/fake/reactor.py
tomprince: > its really all what is needed to make a fake synchronous reactor?

This only implements `IReactorThreads`, `IReactorTime` and (part of) `IReactorCore`, so only those bits are faked.

Twisted comes with `twisted.test.proto_helpers.MemoryReactor` that provides the reactor portion of network (basically `IReactorTCP` et al), but that doesn't actually connect or listen to anything. On the other hand, one can just call methods on the factory/protocol that `MemoryReactor` records to be able to simulate talking over the network.

tomprince: Files are read synchronously in any case.

master/buildbot/test/integration/test_latent.py
```diff
@@ -0,0 +1,224 @@
...
+                     'project': ''},
+                ],
+            )
+        )
+
+        unclaimed_build_requests = []
+        self.successResultOf(master.mq.startConsuming(
+            lambda key, request: unclaimed_build_requests.append(request),
+            ('buildrequests', None, 'unclaimed')))
+
```
tardyp: until there I think the code is 100% the same as previous test

master/buildbot/test/integration/test_latent.py
tomprince: I'll leave a comment about why this is here, but leave off changing the config. I'm slightly wary of including test-specific configuration in the config file.

master/buildbot/test/integration/test_latent.py
```diff
@@ -0,0 +1,224 @@
...
+        # The worker fails to substantiate.
+        controller.start_instance(Failure(TestException("substantiation failed")))
+        # Flush the errors logged by the failure.
+        self.flushLoggedErrors(TestException)
+
+        self.assertEqual(controller.started, True)
```
tardyp: I dont get how you verify that the worker can be started here. I would have expected another controller.start_instance(True).

master/buildbot/test/util/integration.py
```diff
@@ -50,7 +49,41 @@ def loadConfig(self):
...
+    yield master.startService()
+
+    defer.returnValue(master)
+
+
+def getBuilderIdByName(master, builder_name):
+    """
+    Get the ``builder_id`` of the given builder.
+    """
+    d = master.data.get(('builders',))
```
tardyp: there is master.db.builders.findBuilderId(builder_name), which does exactly that, backed by a db query.

murder187ss:
know cool sorry just got working

master/buildbot/test/integration/test_latent.py
tomprince: The issue that this catches (and you can verify that it is an issue if you run the test without the corresponding code changes), is that worker doesn't appear as latent but instead as disconnected after failing to substantiate. And so nothing will try to start an instance ever again.

I've added a comment explaining why I test this here.

master/buildbot/test/integration/test_latent.py
```diff
@@ -226,4 +226,7 @@ def test_worker_accepts_builds_after_failure(self):
         # Flush the errors logged by the failure.
         self.flushLoggedErrors(TestException)
 
+        # If the worker started again after the failure, then the retry logic will have
+        # already kicked in to start a new build on this (the only) worker. We check that
+        # a new instance was requested, which indicates that the worker accepted the build.
```
tardyp: makes more sense, thx

tardyp:
pretty cool stuff.

I may leave it open today for more people to learn from it :)

master/buildbot/test/util/integration.py
tomprince: I'm not sure I understand how the `db` and `data` parts are supposed to be layered. This code is essentially how the schedulers get that information currently.

tardyp: > This code is essentially how the schedulers get that information currently.
Really? arg, that is not scalable.

I agree that there should be a data api binding findBuilderId.. 

actually, there is one.

master.data.updates.findBuilderId(name)

tardyp:
Note that I quickly converted the simple buildbot.test.integration.test_integration_template test, and it now takes 1s instead of 4.5s, with profile being 90% sqlite:

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      125    0.373    0.003    0.373    0.003 {method 'commit' of 'sqlite3.Connection' objects}
       74    0.308    0.004    0.308    0.004 {method 'close' of 'sqlite3.Connection' objects}
      240    0.181    0.001    0.181    0.001 {method 'execute' of 'sqlite3.Cursor' objects}

tardyp:
And.. on the same test, replacing master.db by an instance of fakedb.FakeDBConnector, makes that integration test run in 0.046s

tardyp:
👍 for data.findBuilderId()

I went a little bit further on the optimisation, and came up with this:
https://github.com/tardyp/buildbot/commit/e543bf433f757b4bbe99239ebaa5c6dbc0dfe861

using sqlite in memory allows to run a build in 200ms (instead of 1s with file)

so with those two optimisations we came from 4.5 s to 200ms which is x22

tomprince:
I think that is a good follow-up.

tardyp:
looks like there is a syntax error (see travis)
```
[ERROR]
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/twisted/trial/runner.py", line 498, in loadPackage
    module = modinfo.load()
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/twisted/python/modules.py", line 383, in load
    return self.pathEntry.pythonPath.moduleLoader(self.name)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/twisted/python/reflect.py", line 303, in namedAny
    topLevelPackage = _importAndCheckStack(trialname)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/twisted/python/reflect.py", line 242, in _importAndCheckStack
    return __import__(importName)
exceptions.SyntaxError: invalid syntax (test_latent.py, line 129)
buildbot.test.integration.test_latent
```

tomprince:
Looks like it is only a syntax error on py2.6. :frowning:

tardyp:
indeed, I missed the use of the set comprehension..

master/buildbot/test/fake/reactor.py
```diff
@@ -79,7 +79,10 @@ def callInThreadWithCallback(self, onResult, func, *args, **kw):
         self.calls += 1
         try:
             result = func(*args, **kw)
-        except:
+        except:  # pylint: disable=bare-except
+            # We catch *everything* here, since normally this code would be
+            # running in a thread, where there is nothing that will catch
+            # error.
```
tardyp: Shouldn't we treat keyboard interrupt special case here, to be able to properly cancel a trial -u?

http://effbot.org/zone/stupid-exceptions-keyboardinterrupt.htm

PR #2176, opened 2016-04-29
rename Windows helper script to `buildbot-worker.bat`

rutsky:
Now same `buildbot-worker` command can be used both on Windows and POSIX
systems.


djmitche:
lgtm

bdbaddog:
Has anyone used console_scripts in setuptools? Seems like that would be better than making .bat etc file..
http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

tardyp:
I would agree with @bdbaddog

tomprince:
They work fairly well. The only trouble is if you are running them from a 'pip install -e' install, since the version number is encoded the generated script (but that probably isn't enough of a reason to not use them).

rutsky:
If we can use `console_scripts` we should use it --- this is exactly feature that we want.
I opened ticket for this: http://trac.buildbot.net/ticket/3544
The only thing that worries me is that it might not work with Python 2.6 on Windows and inside virtualenv.

Lets continue discussion of switching from `scripts` to `console_scripts` in the ticket.

PR #2177, opened 2016-04-29
run isort only if there are changed Python files

rutsky:
otherwise isort is being run without argument which leads to recursive fixing of all `*.py` files.

Also adds reporting if isort changed files.

BTW, `setup.py` is not included in list of Python files which looks strange to me. 

/cc: @tardyp


tardyp:
👍

PR #2179, opened 2016-04-29
Add Hipchat reporter

ryansydnor:
Looking for feedback on approach. If things look OK, I'll write
tests / docs.

### DIFF
--- a/master/buildbot/reporters/hipchat.py
+++ b/master/buildbot/reporters/hipchat.py
@@ -1,0 +1,95 @@
+from twisted.internet import defer
+from twisted.python import log
+
+from buildbot import config
+from buildbot.process.results import statusToString
+from buildbot.reporters import utils
+from buildbot.reporters.http import HttpStatusPushBase
+
+
+class HipChatStatusPush(HttpStatusPushBase):
+    name = "HipChatStatusPush"
+
+    def checkConfig(self, auth_token, endpoint="https://api.hipchat.com",
+                    builder_room_map=None, builder_user_map=None,
+                    event_messages=None, **kwargs):
+        if not isinstance(auth_token, basestring):
+            config.error('auth_token must be a string')
+        if not isinstance(endpoint, basestring):
+            config.error('endpoint must be a string')
+        if builder_room_map and not isinstance(builder_room_map, dict):
+            config.error('builder_room_map must be a dict')
+        if builder_user_map and not isinstance(builder_user_map, dict):
+            config.error('builder_user_map must be a dict')
+
+    @defer.inlineCallbacks
+    def reconfigService(self, auth_token, endpoint="https://api.hipchat.com",
+                        builder_room_map=None, builder_user_map=None,
+                        event_messages=None, **kwargs):
+        yield HttpStatusPushBase.reconfigService(self, **kwargs)
+        self.auth_token = auth_token
+        self.endpoint = endpoint
+        self.builder_room_map = builder_room_map
+        self.builder_user_map = builder_user_map
+        self.user_notify = '%sv2/user/%s/message?auth_token=%s'
+        self.room_notify = '%sv2/room/%s/notification?auth_token=%s'
+
+    @defer.inlineCallbacks
+    def buildStarted(self, key, build):
+        yield self.send(build, key[2])
+
+    @defer.inlineCallbacks
+    def buildFinished(self, key, build):
+        yield self.send(build, key[2])
+
+    @defer.inlineCallbacks
+    def getBuildDetailsAndSendMessage(self, build, key):
+        yield utils.getDetailsForBuild(self.master, build, **self.neededDetails)
+        postData = yield self.getRecipientList(build, key)
+        postData['message'] = yield self.getMessage(build, key)
+        extra_params = yield self.getExtraParams(build, key)
+        postData.update(extra_params)
+        defer.returnValue(postData)
+
+    def getRecipientList(self, build, event_name):
+        result = {}
+        builder_name = build['builder']['name']
+        if self.builder_user_map and builder_name in self.builder_user_map:
+            result['id_or_email'] = self.builder_user_map[builder_name]
+        if self.builder_room_map and builder_name in self.builder_room_map:
+            result['room_id_or_name'] = self.builder_room_map[builder_name]
+        return result
+
+    def getMessage(self, build, event_name):
+        event_messages = {
+            'new': 'Buildbot started build %s here: %s' % (build['builder']['name'], build['url']),
+            'finished': 'Buildbot finished build %s with result %s here: %s'
+                        % (build['builder']['name'], statusToString(build['results']), build['url'])
+        }
+        return event_messages.get(event_name, '')
+
+    # use this as an extension point to inject extra parameters into your postData
+    def getExtraParams(self, build, event_name):
+        return {}
+
+    @defer.inlineCallbacks
+    def send(self, build, key):
+        postData = yield self.getBuildDetailsAndSendMessage(build, key)
+        if not postData or 'message' not in postData or not postData['message']:
+            return
+
+        if not self.endpoint.endswith('/'):
+            self.endpoint += '/'
+
+        urls = []
+        if 'id_or_email' in postData:
+            urls.append(self.user_notify % (self.endpoint, postData.pop('id_or_email'), self.auth_token))
+        if 'room_id_or_name' in postData:
+            urls.append(self.room_notify % (self.endpoint, postData.pop('room_id_or_name'), self.auth_token))
+
+        if urls:
+            for url in urls:
+                response = yield self.session.post(url, postData)
+                if response.status_code != 200:
+                    log.msg("%s: unable to upload status: %s" %
+                            (response.status_code, response.content))
--- a/master/buildbot/reporters/http.py
+++ b/master/buildbot/reporters/http.py
@@ -57,15 +57,15 @@ class HttpStatusPushBase(service.BuildbotService):
     def startService(self):
         self.session = self.sessionFactory()
         yield service.BuildbotService.startService(self)
-        startConsuming = self.master.mq.startConsuming
 
+        startConsuming = self.master.mq.startConsuming
         self._buildCompleteConsumer = yield startConsuming(
             self.buildFinished,
             ('builds', None, 'finished'))
 
         self._buildStartedConsumer = yield startConsuming(
             self.buildStarted,
-            ('builds', None, 'started'))
+            ('builds', None, 'new'))
 
     def stopService(self):
         self._buildCompleteConsumer.stopConsuming()
@@ -99,9 +99,6 @@ class HttpStatusPush(HttpStatusPushBase):
 
     def checkConfig(self, serverUrl, user, password, **kwargs):
         HttpStatusPushBase.checkConfig(self, **kwargs)
-        if txrequests is None:
-            config.error("Please install txrequests and requests to use %s (pip install txrequest)" %
-                         (self.__class__,))
 
     def reconfigService(self, serverUrl, user, password, **kwargs):
         HttpStatusPushBase.reconfigService(self, **kwargs)
--- a/master/buildbot/test/unit/test_reporter_http.py
+++ b/master/buildbot/test/unit/test_reporter_http.py
@@ -76,7 +76,7 @@ class TestHttpStatusPush(unittest.TestCase, ReporterTestMixin):
         yield self.createReporter()
         build = yield self.setupBuildResults(SUCCESS)
         build['complete'] = False
-        self.sp.buildStarted(("build", 20, "started"), build)
+        self.sp.buildStarted(("build", 20, "new"), build)
         build['complete'] = True
         self.sp.buildFinished(("build", 20, "finished"), build)
         # we make sure proper calls to txrequests have been made
--- a/master/docs/manual/cfg-reporters.rst
+++ b/master/docs/manual/cfg-reporters.rst
@@ -880,7 +880,7 @@ GitLabStatusPush
     from buildbot.plugins import reporters
 
     gl = reporters.GitLabStatusPush('private-token', context='continuous-integration/buildbot', baseUrl='https://git.yourcompany.com')
-    c['services'].append(ss)
+    c['services'].append(gl)
 
 :class:`GitLabStatusPush` publishes build status using `GitLab Commit Status API <http://doc.gitlab.com/ce/api/commits.html#commit-status>`_.
 The build status is published to a specific commit SHA in GitLab.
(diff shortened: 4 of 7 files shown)


ryansydnor:
@tardyp would love your feedback!

tardyp:
Hi, I would prefer that you base your work on HttpStatusPushBase
https://github.com/buildbot/buildbot/blob/master/master/buildbot/reporters/http.py#L34

tardyp:
the  principle is to have all reporters who push to http derive from the same class to share code and unit tests principles.

I dont think it is a good idea to have urlibs2 based reporters, and other based on txrequests.

ryansydnor:
@tardyp - agree with the principle. Took at stab at refactoring HTTPStatusPushBase to be more flexible.

After working on a few reporters - there's a ton of overlap between them. Standardizing on the interface is probably worthwhile (and then we can use a base class where you only have to override "send").

cmouse:
it would be good to use one http request library across board

ryansydnor:
@cmouse I think that's a great idea - but probably beyond the scope of this PR. We could open a ticket in trac for the improvement?

@tardyp - if you think this approach is reasonable, I'll finish the swing and write up some docs.

master/buildbot/reporters/http.py
```diff
@@ -41,11 +41,28 @@ def checkConfig(self, *args, **kwargs):
...
+    def reconfigService(self, builders=None,
+                        startCB=None, startArg=None,
+                        reviewCB=None, reviewArg=None,
+                        summaryCB=None, summaryArg=None,
+                        **kwargs):
         yield service.BuildbotService.reconfigService(self)
         self.builders = builders
+        self.startCB = startCB
+        self.startArg = startArg
+        self.reviewCB = reviewCB
```
tardyp: I don't think we should add more callbacks in the base class. 

* reviewCB does not make sense for most reporters. It is a gerrit thing

* I think it is much more flexible to have people subclass the reporters if they really want to change the details of the messages.

* perhaps we can put an hint for the event kind in send, rather than putting the callback

ryansydnor: Preface - I want to get to the best possible design here. So I'm asking questions to work through it together :)

1. I'm not holding reviewCB as a gerrit thing - I'm viewing it as a builder configuration thing. If you create "pipelines" (or collections of related builders), then a reviewCB makes sense to understand how the entire collection behaved. Does that make sense?

2. This PR introduces the concept of callbacks as opt-in, which I believe will reduce long term code duplication. Different HTTP services may have different interfaces for what type of post data they expect (hipchat vs. stash is a great example). So, I believe we should enable that flexibility in the base class. Where does your understanding of "how this should be" differ?

3. I'm not following the recommendation - can you elaborate?

tardyp: When you say review (non-gerrit) people think code review. buildbot is not doing code review, it is doing pre-commit. Some say pre-submit checks. Others CI vote.

Gerrit users often confuse ci vote as code review (I plaid guilty myself, as I first introduced reviewCB name in gerrit status)

I know you are working with gerrit, so this is why I suspect you are gerrit biased :)


I clearly don't like callbacks, because they are not as flexible as overridable methods. The main advantage of methods, is that they are passed a 'self', which gives access to master, master.data, so that your callbacks can fetch more data as they like.

Also, we shall factorize what is necessary for the current use cases as we know, not try to factorize for future maybe usecases, because that too often leads to over engineering.
If there is a need for refactor in the longer term, we will refactor at that time.

With your current design, GitlabStatusPush will get a new reviewCB argument, which is completely ignored, and not useful

master/buildbot/reporters/hipchat.py
```diff
@@ -0,0 +1,32 @@
+from twisted.internet import defer
+from twisted.python import log
+from buildbot.reporters.http import HttpStatusPushBase
```
sa2ajj: [nit] missing empty line

master/buildbot/reporters/http.py
```diff
@@ -41,11 +41,28 @@ def checkConfig(self, *args, **kwargs):
                          (self.__class__,))
         if not isinstance(kwargs.get('builders'), (type(None), list)):
             config.error("builders must be a list or None")
+        # check that our callbacks are functions
+        if kwargs.get('startCB') and not hasattr(kwargs['startCB'], '__call__'):
```
sa2ajj: why `callable` can't be used here?

master/buildbot/reporters/http.py
```diff
@@ -41,11 +41,28 @@ def checkConfig(self, *args, **kwargs):
...
+            config.error("start callback must be callable")
+        if kwargs.get('reviewCB') and not hasattr(kwargs['reviewCB'], '__call__'):
```
sa2ajj: ditto

master/buildbot/reporters/http.py
```diff
@@ -41,11 +41,28 @@ def checkConfig(self, *args, **kwargs):
...
+            config.error("review callback must be callable")
+        if kwargs.get('summaryCB') and not hasattr(kwargs['summaryCB'], '__call__'):
```
sa2ajj: ditto

master/buildbot/test/unit/test_reporter_http.py
```diff
@@ -77,7 +77,7 @@ def test_basic(self):
         yield self.createReporter()
         build = yield self.setupBuildResults(SUCCESS)
         build['complete'] = False
-        self.sp.buildStarted(("build", 20, "started"), build)
+        self.sp.buildStarted(("build", 20, "new"), build)
```
sa2ajj: whille I appreciate keeping the style, I'd be glad to see `"` changed to '''

sa2ajj: this probably should be added somewhere more explicitly (this "preference" follows PEP-8)

master/buildbot/reporters/http.py
ryansydnor: Ok. I'll remove the callbacks from the base class and move them into the specific reporter. I'll move the corresponding unit tests as well. 

Thanks for the feedback. The biggest piece that resonated with me was the piece about current use cases - design for what we know to avoid over engineering.

master/buildbot/test/unit/test_reporter_http.py
ryansydnor: updated this file to match the spec. maybe throw it into the lint check?

ryansydnor:
Hey @tardyp and @sa2ajj - thanks for your feedback. Made the changes we talked about and added docs. Let me know what you think!

master/buildbot/reporters/hipchat.py
```diff
@@ -0,0 +1,92 @@
...
+
+class HipChatStatusPush(HttpStatusPushBase):
+    name = "HipChatStatusPush"
+
+    def checkConfig(self, *args, **kwargs):
+        if not isinstance(kwargs.get('auth_token'), basestring):
+            config.error('auth token must be a string')
+        if not isinstance(kwargs.get('endpoint'), basestring):
+            config.error('hipchat endpoint must be a string')
+        if kwargs.get('startCB') and not callable(kwargs['startCB']):
```
tardyp: When I said, I dont like callbacks, I meant: Please don't use callbacks.

master/buildbot/reporters/hipchat.py
```diff
@@ -0,0 +1,92 @@
...
+            config.error("start callback must be callable")
+        if kwargs.get('reviewCB') and not callable(kwargs['reviewCB']):
```
tardyp: please don't use reviewCB as this is too much related to your own process bias

master/buildbot/reporters/hipchat.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @defer.inlineCallbacks
+    def getMoreInfoAndSend(self, build, callback, callbackArg):
+        if callback:
+            yield utils.getDetailsForBuild(self.master, build, **self.neededDetails)
+            postData = yield callback(build, callbackArg)
+            yield self.send(postData)
+
+    @defer.inlineCallbacks
+    def send(self, params):
```
tardyp: params is not a good name for a variable.

This is changing the send() API from the other http reporters, which is not coherent.
You can change it for all the http reporter (as buildbot nine is not yet released, this is a good time), or not change it at all

master/buildbot/reporters/hipchat.py
```diff
@@ -0,0 +1,92 @@
...
+from twisted.internet import defer
+from buildbot import config
+from buildbot.reporters import utils
+from buildbot.reporters.http import HttpStatusPushBase
+
+
+class HipChatStatusPush(HttpStatusPushBase):
+    name = "HipChatStatusPush"
+
+    def checkConfig(self, *args, **kwargs):
```
tardyp: checkConfig and reconfigService must have the same signature

master/docs/manual/cfg-reporters.rst
```diff
@@ -899,3 +899,83 @@ It uses private token auth, and the token owner is required to have at least rep
...
+                    it should return a dictionary:
+
+                    .. code-block:: python
+
+                        {
+                            'message': 'Buildbot began building %s here: %s' % (build['builder']['name'], build['url']),
+                            'user_id': '44'
+                        }
+
+    :param reviewCB: (optional) Called each time a build is finished. Can be a deferred.
```
tardyp: please replace by self.getBuildFinishedMessage()

master/docs/manual/cfg-reporters.rst
```diff
@@ -899,3 +899,83 @@ It uses private token auth, and the token owner is required to have at least rep
...
+                    It should return a dictionary:
+
+                    .. code-block:: python
+
+                        {
+                            'message': 'Buildbot finished building %s here: %s' % (build['builder']['name'], build['url']),
+                            'room_id': '25'
+                        }
+
+    :param summaryCB: (optional) Called each time a buildset finishes. Can be a deferred.
```
tardyp: please replace by getBuildsetCompleteMessage()

tardyp:
I'm a little bit disapointed that I see most of my comments ignored.
Please tell me if there is something that is not clear

> When you say review (non-gerrit) people think code review. buildbot is not doing code review, it is doing pre-commit. Some say pre-submit checks. Others CI vote.

> Gerrit users often confuse ci vote as code review (I plaid guilty myself, as I first introduced reviewCB name in gerrit status)

> I know you are working with gerrit, so this is why I suspect you are gerrit biased :)

> I clearly don't like callbacks, because they are not as flexible as overridable methods. The main advantage of methods, is that they are passed a 'self', which gives access to master, master.data, so that your callbacks can fetch more data as they like.

> Also, we shall factorize what is necessary for the current use cases as we know, not try to factorize for future maybe usecases, because that too often leads to over engineering.
> If there is a need for refactor in the longer term, we will refactor at that time.

> With your current design, GitlabStatusPush will get a new reviewCB argument, which is completely ignored, and not useful

ryansydnor:
I misunderstood. I thought you were hesitant to include callbacks in the base class - not hesitant to include callbacks completely.

It's also very possible I'm heavily influencing the design with my current use case, which is part of the reason I'm trying to open source the implementation to begin with (learning). Thanks for working through it with me.

Let's level set on the use case:

A build starts/finishes. I want to report that to Hipchat. How do I know who to report it to? I look at build metadata and combine it with some other piece of metadata I have about Hipchat to determine the contents of the message and where it should be sent.

My hesitancy to remove callbacks was that in order to achieve the above all consumers of this reporter will have to subclass the reporter. I do not see a way to provide sensible "default" behavior (mainly due to the message "destination").

Now, that might be OK - but it is dissimilar from the other reporters (which I was clearly pattern matching against).

tardyp:
> A build starts/finishes. I want to report that to Hipchat. How do I know who to report it to? I look at build metadata and combine it with some other piece of metadata I have about Hipchat to determine the contents of the message and where it should be sent.

So I can see several problems here:

1) what is the text message you want to send. This looks easy to implement a sensible default per event (build start, build finish, buildset finish)

2) which room/recipients to send the message
    there might be a sensible default, which let user define static recipient list, or a dictionnary mapping buildername to recipient_list.
   then if the default is not enough, the user may want to override getRecipientList(event_name, build)

ryansydnor:
@tardyp - tried to incorporate your advice much more heavily this time around. Let me know what you think.

I also decided to remove buildset listeners - it doesn't seem like it fits well with the normal workflow and I can just subclass for my use case.

tardyp:
there are some lints and docs errors. please watch travis results

master/buildbot/reporters/github.py
```diff
@@ -81,7 +81,8 @@ def createStatus(self,
             [self.baseURL, 'repos', repo_user, repo_name, 'statuses', sha]), json=payload)
 
     @defer.inlineCallbacks
-    def send(self, build):
+    def send(self, postData):
```
tardyp: please don't change the semantic. its a build that is sent in send.

having variable argument mean in a method is not a good thing. We can make two. one for build (sendForBuild), and one for buildset (sendForBuildSet)

master/buildbot/test/unit/test_reporter_http.py
tardyp: I am a little bit reluctant to take coding style rule that is not enforced by an automated tool (or better fixed atomatically by autopep8)

master/buildbot/reporters/hipchat.py
```diff
@@ -0,0 +1,93 @@
...
+    def getBuildDetailsAndSendMessage(self, build, key):
+        yield utils.getDetailsForBuild(self.master, build, **self.neededDetails)
+        postData = yield self.getRecipientList(build, key)
+        postData['message'] = yield self.getMessage(build, key)
+        extra_params = yield self.getExtraParams(build, key)
+        postData.update(extra_params)
+        yield self.send(postData)
+
+    def getRecipientList(self, build, event_name):
+        result = {}
```
tardyp: I think this is good. but we should document it. and also document that it can returned a deferred

tardyp:
Thank you for the hard work on this PR. I think it is much better now (still some smaller cleanups needed)

master/buildbot/reporters/github.py
ryansydnor: Hm. I think I was holding "send" as a method that purely handled http requests for the data it was given (how I tried to hold the hipchat reporter).

It seems that the interface (in all cases) actually just expects a build and then does some filtering/logic to end up with the final json payload.

I can follow that pattern in my work as well and remove the need to modify this interface.

master/buildbot/reporters/hipchat.py
ryansydnor: Ok. Added docs for it.

master/buildbot/test/unit/test_reporter_http.py
ryansydnor: Ok. Holding off until it's part of the automated checks.

ryansydnor:
@tardyp - cleaned everything up :)

master/docs/manual/cfg-reporters.rst
```diff
@@ -899,3 +899,84 @@ It uses private token auth, and the token owner is required to have at least rep
...
+
+    {
+        "message": "Buildbot started/finished build MyBuilderName (with result success) here: http://mybuildbot.com/#/builders/23",
+        "id_or_email": "12"
+    }
+
+
+These are the minimum required parameters to send a message to Hipchat. If you require more don't hesitate to subclass
+and modify one or more of the following functions:
+:py:func:`getRecipientList` :py:func:`getMessage`, :py:func:`getExtraParams`
```
tardyp: Please describe each of those APIs with complete signature, and what they are supposed to return, etc.

ryansydnor: Good idea. Documented via example so people can get up and running more quickly.

master/docs/manual/cfg-reporters.rst
```diff
@@ -899,3 +899,96 @@ It uses private token auth, and the token owner is required to have at least rep
...
+The default json object contains the minimal required parameters to send a message to Hipchat.
+
+.. code-block:: json
+
+    {
+        "message": "Buildbot started/finished build MyBuilderName (with result success) here: http://mybuildbot.com/#/builders/23",
+        "id_or_email": "12"
+    }
+
+
```
tardyp: example is good, but there is still a need for a formal api description

ryansydnor: Added.

tardyp:
another one that needs rebase

PR #2184, opened 2016-04-30
remove worker-side usePTY support 

rutsky:
Previously it was deprecated.

`ShellCommand`'s `usePTY` should be used in master configuration instead.


tardyp:
👍

PR #2185, opened 2016-04-30
Fix glob() and runGlob() calls to remote 

rayalan:
There is a disconnect between the way the master calls glob in both the CommandMixin and CompositeMixin classes and the way the worker expects the command, which is defined as:

```
class GlobPath(base.Command):

    header = "glob"

    # args['path'] is relative to Builder directory, and is required.
    requiredArgs = ['path']

    def start(self):
        pathname = os.path.join(self.builder.basedir, self.args['path'])
```

This updates both components to pass the expected `path` arg instead of the `glob` for the glob command. Since the worker and master protocol tests are isolated, though, there isn't any additional verification that this change is correct. However, it did make glob work correctly on my local setup using CommandMixin.

I also updated the documentation on a couple minor points to be consistent with code.

There are a couple bigger questions raised here such as:
  - Should CommandMixin use `runGlob()` instead of `glob()` as the documentation claimed?
  - Should we pass `glob` via the remote protocol instead of `path`?

Both of these seemed like bigger questions with likely painful transitions, so I decided to fix the code as is and not worry about the rest.

### DIFF
--- a/master/buildbot/test/integration/test_commandmixin.py
+++ b/master/buildbot/test/integration/test_commandmixin.py
@@ -1,0 +1,96 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+from twisted.internet import defer
+
+from buildbot.process import results
+from buildbot.process.buildstep import BuildStep
+from buildbot.process.buildstep import CommandMixin
+from buildbot.test.util.integration import RunMasterBase
+
+
+# This integration test creates a master and worker environment,
+# and makes sure the command mixin is working.
+class CommandMixinMaster(RunMasterBase):
+
+    @defer.inlineCallbacks
+    def test_commandmixin(self):
+        yield self.setupConfig(masterConfig())
+
+        change = dict(branch="master",
+                      files=["foo.c"],
+                      author="me@foo.com",
+                      comments="good stuff",
+                      revision="HEAD",
+                      project="none"
+                      )
+        build = yield self.doForceBuild(wantSteps=True, useChange=change,
+                                        wantLogs=True)
+        self.assertEqual(build['buildid'], 1)
+        self.assertEqual(build['results'], results.SUCCESS)
+
+
+class TestCommandMixinStep(BuildStep, CommandMixin):
+
+    def __init__(self, *args, **kwargs):
+        BuildStep.__init__(self, *args, **kwargs)
+
+    @defer.inlineCallbacks
+    def run(self):
+        contents = yield self.runGlob('*')
+        if contents != []:
+            defer.returnValue(results.FAILURE)
+
+        hasPath = yield self.pathExists('composite_mixin_test')
+        if hasPath:
+            defer.returnValue(results.FAILURE)
+
+        yield self.runMkdir('composite_mixin_test')
+
+        hasPath = yield self.pathExists('composite_mixin_test')
+        if not hasPath:
+            defer.returnValue(results.FAILURE)
+
+        contents = yield self.runGlob('*')
+        if not contents[0].endswith('composite_mixin_test'):
+            defer.returnValue(results.FAILURE)
+
+        yield self.runRmdir('composite_mixin_test')
+
+        hasPath = yield self.pathExists('composite_mixin_test')
+        if hasPath:
+            defer.returnValue(results.FAILURE)
+
+        defer.returnValue(results.SUCCESS)
+
+
+# master configuration
+def masterConfig():
+    c = {}
+    from buildbot.config import BuilderConfig
+    from buildbot.process.factory import BuildFactory
+    from buildbot.plugins import schedulers
+
+    c['schedulers'] = [
+        schedulers.AnyBranchScheduler(
+            name="sched",
+            builderNames=["testy"])]
+
+    f = BuildFactory()
+    f.addStep(TestCommandMixinStep())
+    c['builders'] = [
+        BuilderConfig(name="testy",
+                      workernames=["local1"],
+                      factory=f)]
+    return c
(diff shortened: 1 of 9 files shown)


tardyp:
> Since the worker and master protocol tests are isolated, though, there isn't any additional verification that this change is correct. However, it did make glob work correctly on my local setup using CommandMixin.

This is something that we are changing, by introducing integration tests.
You can look at master/buildbot/test/integration/test_transfer.py for an example of such a test.

tardyp:
> Should CommandMixin use runGlob() instead of glob() as the documentation claimed?
I think runGlob is better, as it suggests better that this is remote (as runCommand)

> Should we pass glob via the remote protocol instead of path?
I think glob makes more sense, but if we released the buildbot-slave package with 'path', we must garantee the backward compatibility. That said, there is no bacward compatibility as it looks like this stuff never actually worked.

master/docs/developer/cls-buildsteps.rst
```diff
@@ -584,7 +584,7 @@ This class can only be used in new-style steps.
         Determine if the given path exists on the worker (in any form - file, directory, or otherwise).
         This uses the ``stat`` command.
 
-    .. py:method:: runGlob(glob)
```
tardyp: I would rather say


    runGlob(test)

tardyp:
with worker rework, it is harder to git blame.
there is one bug related: http://trac.buildbot.net/ticket/3328 (which I can't see right now, because it looks like trac is down)

master/docs/developer/cls-buildsteps.rst
rayalan: Do you prefer `runGlob(test)` to `runGlob(pattern)` or `runGlob(path)`? The other sections of help here use:

```
runXXX(param1, param2, ...)
   :param param1 some docs
   :param param2 some docs
```
so runGlob(path) might be more consistent. (Or runGlob(glob) if we end up changing the worker protocol).

rayalan:
I poked at the blame before committing long enough to convince myself that the two mixins themselves never worked. I didn't trace down whether there's another path (no pun intended) where glob did work. 

The 0.8.12 tag shows that it uses the `path` argument, so if the 0.9 series masters are expected to be backward compatible with the 0.8 series workers, backward compatibility is an issue. That is, right now, it can be fixed without forcing a worker upgrade. If we change the protocol, it will force both master and workers to be upgraded.

I'm happy to change the protocol if that's preferred.

rayalan:
I'll take a look at the integration tests and see about adding one.

tardyp:
ah. so you say this is a bug fix that didn't go through the nine branch?

master/docs/developer/cls-buildsteps.rst
tardyp: I don't know why I wrote runGlob(test). I mean runGlob(path), as the following documents the param 'path'

tardyp:
here is the original PR, which has the bug
https://github.com/buildbot/buildbot/pull/970/files

tardyp:
I would really appreciate you come with an integration test to prove that this work

rayalan:
I'm not sure what you mean by "didn't go through the nine branch".

In summary:
- [ ] Glob has never worked, so there are two easy paths:
  1. Leave the parameter at path, which means 0.9.0+ masters will be backward compatible with 0.8.x workers.
   2. Change the parameter to glob, which means 0.9.0+ masters will work with 0.9.0 workers. 
- [ ] Change the command to runGlob(path) since there are no backward compatibility issues (since master glob has never worked).
- [ ] Update the documentation to match the above choices.
- [ ] Add an integration test.

Does that seem right? If so, the only question is whether to change the protocol parameter.

tardyp:
I was confused by your sentence:
> The 0.8.12 tag shows that it uses the path argument

which suggested to me that the *master* used path in the eight branch. This is not true. only slave is using it.

So I agree with all your plan, and would suggest to go for the option which has (restores) compatibility with eight slaves.

thanks!

rayalan:
Ready for the next round of feedback. I added integration tests for the CommandMixin and CompositeStepMixin, except for the deprcated getFileContentFromWorker() method.

master/buildbot/test/integration/test_commandmixin.py
```diff
@@ -0,0 +1,96 @@
...
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+from buildbot.process import results
+from buildbot.process.buildstep import BuildStep
+from buildbot.process.buildstep import CommandMixin
+from buildbot.test.util.integration import RunMasterBase
+
+from twisted.internet import defer
```
sa2ajj: [nit] I believe `twistd` should be imported before `buildbot`

master/buildbot/test/integration/test_compositestepmixin.py
```diff
@@ -0,0 +1,97 @@
...
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+from buildbot.process import results
+from buildbot.process.buildstep import BuildStep
+from buildbot.steps.worker import CompositeStepMixin
+from buildbot.test.util.integration import RunMasterBase
+
+from twisted.internet import defer
```
sa2ajj: ditto

sa2ajj:
LGTM, otherwise (sphinx complains about spelling though)

master/docs/relnotes/index.rst
```diff
@@ -36,6 +36,10 @@ Fixes
 
 * OpenStackLatentWorker uses the novaclient API correctly now.
 
+* runGlob() uses the correct remote protocol for both :py:class:`~buildbot.process.buildstep.CommandMixin` and :py:class:`~buildbot.steps.worker.ComposititeStepMixin`.
```
rutsky: Use verbatim formatting for code blocks or code identifiers:

```
``runGlob()``
```

here and below.

rutsky:
@sa2ajj I think merge-me label should be applied only if there is no work needed, and IMO spelling issues (actually missing code/identifier formatting) should be fixed.

master/buildbot/test/integration/test_commandmixin.py
rayalan: Are you sure about this? All the integration tests have twisted import lines after buildbot, and when I swap them, I get tracebacks and test failures related to 'no such file or directory'.

rayalan:
I think it is all set.
1. Fixed documentation decorators.
2. Didn't change import ordering (per above discussion about other tests and it causing failing integration tests)
3. Tests are failing now, but they seem to have gone crazy over a use-PTY setting in test_transfer, which I'm going to blissfully assume isn't me for a day or so, and then circle back if it is my fault.

tardyp:
looks good. re-running the tests

tardyp:
please rebase, there is conflicts with master branch

rayalan:
Rebased (conflict was release notes).

rutsky:
@rayalan please rebase once more.

tardyp:
ping

PR #2186, opened 2016-05-01
Update remote error handling to relay remote errors

rayalan:
Previously, when a remote's remote (e.g. master) failed during an operation
(such as uploading a file), the remote received a CopiedFailure. The remote
would not not properly reconstruct the failure to pass onto master. This led
the remote raising additional exceptions rather than properly terminating the
current command.

With this approach, the traceback is not passed (although a copy exists in the
original remote's logs, but the error is properly passed through the remote back
to master.

---

I ran into this problem working on a file uploader when I passed in an invalid `masterdest` argument that caused os.unlink to fail. 

I suspect there's a unit test that needs to be written, but I'm not sure how/where to put it, especially since the failure scenario seems to cause the build step in question to run indefinitely since the master never gets notification of the command completing by failing. Thoughts on the best approach to testing?

### DIFF
--- a/slave/buildslave/base.py
+++ b/slave/buildslave/base.py
@@ -209,8 +209,15 @@ class SlaveBuilderBase(service.Service):
             log.err(failure)
             # failure, if present, is a failure.Failure. To send it across
             # the wire, we must turn it into a pb.CopyableFailure.
-            failure = pb.CopyableFailure(failure)
-            failure.unsafeTracebacks = True
+            if isinstance(failure, pb.CopiedFailure):
+                # copiedFailures from other remotes don't construct normally
+                # into CopyableFailure. This won't fully send the traceback,
+                # but is better than failing and hanging the command.
+                failure = pb.CopyableFailure(failure.value, failure.type,
+                                             failure.getTracebackObject())
+            else:
+                failure = pb.CopyableFailure(failure)
+                failure.unsafeTracebacks = True
         else:
             # failure is None
             log.msg("SlaveBuilder.commandComplete", self.command)
--- a/worker/buildbot_worker/base.py
+++ b/worker/buildbot_worker/base.py
@@ -209,8 +209,15 @@ class WorkerForBuilderBase(service.Service):
             log.err(failure)
             # failure, if present, is a failure.Failure. To send it across
             # the wire, we must turn it into a pb.CopyableFailure.
-            failure = pb.CopyableFailure(failure)
-            failure.unsafeTracebacks = True
+            if isinstance(failure, pb.CopiedFailure):
+                # copiedFailures from other remotes don't construct normally
+                # into CopyableFailure. This won't fully send the traceback,
+                # but is better than failing and hanging the command.
+                failure = pb.CopyableFailure(failure.value, failure.type,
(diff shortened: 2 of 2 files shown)


tomprince:
There should probably be an issue raised upstream about this.

tardyp:
I agree with tomprince we must first make sure the rootcause is really this, and that copiedfailure is not copiable. I would guess that is a pretty simple unit test that can then be transmitted to twisted folks.
Once we have that, we can work with them on the best workaround (or better, monkey patch a fix)

rayalan:
Does this look a reasonable approximation of the way buildbot uses the pb interface? 

pbsimple.py:
```
# Copyright (c) Twisted Matrix Laboratories.
# See LICENSE for details.
from twisted.spread import pb
from twisted.internet import reactor

class Echoer(pb.Root):
    def remote_echo(self, st):
        print 'echoing:', st
        return st

    def remote_except(self, st):
        1 / 0

if __name__ == '__main__':
    reactor.listenTCP(8789, pb.PBServerFactory(Echoer()))
    reactor.run()
```

pbsimpleclient.py:
```
# Copyright (c) Twisted Matrix Laboratories.
# See LICENSE for details.
from twisted.spread import pb
from twisted.internet import reactor
from twisted.python import util

factory = pb.PBClientFactory()
reactor.connectTCP("localhost", 8789, factory)

def echo(object_):
    d = factory.getRootObject()
    d.addCallback(lambda object: object.callRemote("echo", object_))
    d.addCallback(lambda echo: 'server echoed: '+ echo)
    d.addErrback(lambda reason: 'error: ' + str(reason.value) + " :" + str(reason.__class__) + "/" + str(type(reason)))
    d.addCallback(util.println)

d = factory.getRootObject()

echo("hello network")

d = factory.getRootObject()
d.addCallback(lambda object: object.callRemote("except", "trigger exception"))
d.addErrback(lambda reason: echo(reason))
d.addCallback(util.println)

reactor.run()
```
To run, one one console: `python pbsimple.py`
On another console: `python pbsimpleclient.py`

The noteworthy output is when the server echoes this:
```
echoing: Unpersistable('Unpersistable data: instance of class twisted.spread.pb.CopiedFailure deemed insecure')
```
which is where it has received a copy of the copy of the original exception. Instead of receiving 2nd CopyiedFailure instance, it has a Unpersistable instance. 

---

Note that this test deals with the primary issue; there's a secondary issue where the `CopyableFailure(copiedFailureInstance)` doesn't work (which is how I discovered this problem sending a copy of the copy).

rayalan:
Also, it looks like buildbot has dealt with these sorts of issues a long time ago (https://twistedmatrix.com/trac/ticket/3174, https://twistedmatrix.com/trac/ticket/4520), but my Twisted understanding isn't good enough to put the issues in context.

tardyp:
Did you send a bug to twisted? what are their advice?

rayalan:
I'm still working on whether my test example case is the right pb approximation for what buildbot is doing; I'll look into sending the bug to twisted.

tardyp:
I'm afraid I have to put this stalled. I would love to see it make progress though

tardyp:
closing as per stalled policy

PR #2189, opened 2016-05-03
Fix project info line

crawfxrd:
Correct "Infos" to "Info", use <strike>tag interpolation</strike> `&nbsp;` to preserve space after colon.

### DIFF
--- a/www/base/src/app/about/about.tpl.jade
+++ b/www/base/src/app/about/about.tpl.jade
@@ -5,7 +5,7 @@
         .row
           .col-sm-4
             ul
-              li Project Infos:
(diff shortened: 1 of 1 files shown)


www/base/src/app/about/about.tpl.jade
```diff
@@ -5,8 +5,7 @@
         .row
           .col-sm-4
             ul
-              li Project Infos:
```
tardyp: I usually prefer &nbsp; because it is less implicit

crawfxrd:
Changed to use `&nbsp;` to match convention.

PR #2190, opened 2016-05-03
Fix consistency in scheduler_changes table usage #3532

gracinet:
The ids referred to in scheduler_changes should be those from the
schedulers table, instead of the objectid (used for object_states)

For now, this PR is work-in-progress, as it doesn't take @djmitche remark that scheduler id could be take from `_getServiceId`, which itself could implement caching.

That being said, untangling OBJECTID and SCHEDULERID in the tests has been somewhat easier than I expected.

### DIFF
--- a/master/buildbot/schedulers/basic.py
+++ b/master/buildbot/schedulers/basic.py
@@ -102,8 +102,8 @@ class BaseBasicScheduler(base.BaseScheduler):
         # changes, so get rid of any hanging around from previous
         # configurations
         else:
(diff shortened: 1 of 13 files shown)


djmitche:
The tests look good :)

gracinet:
Should be ready for review

gracinet:
Looks like the coverage decrease is about `slave/buildslave/null.py` (untouched by this PR) and this looks like changes made in master since my branch started improved coverage on this side.
IOW, that looks to be a false positive.

tardyp:
looks good! thanks!

PR #2192, opened 2016-05-03
Switched some properties tests to inline callbacks

gracinet:
This is something I had in mind after a previous PR related to properties. It didn't sound right at the time to do this as part of that PR


gracinet:
looks like the broken tests in CI are unrelated (I see usePTY). I don't see a relaunch, should I simply rebase ?

tardyp:
👍

PR #2191, opened 2016-05-03
fix usePTY usage in local worker tests

rutsky:

PR #2193, opened 2016-05-03
Don't return deferreds twice when substantiating.

tomprince:
Adding callbacks to a deferred mutates the deferred, and change the result that will be passed to new callbacks. Thus, returning the same deferred to multiple callers doesn't behave sanely. Change the place we do in the latent worker to instead return a fresh deferred every time. (Note: https://github.com/buildbot/buildbot/compare/master...tomprince:latent-with-a-Build appears to be necessary to expose the issue here, since we happen to have only ever called this serially).

### DIFF
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -28,6 +28,7 @@ from buildbot.process import metrics
 from buildbot.process.properties import Properties
 from buildbot.reporters.mail import MailNotifier
 from buildbot.status.worker import WorkerStatus
+from buildbot.util import Notifier
 from buildbot.util import ascii2unicode
 from buildbot.util import service
 from buildbot.util.eventual import eventually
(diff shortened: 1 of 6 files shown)


tardyp:
looks good!

tomprince:
The build failure is #2191.

master/buildbot/worker/base.py
```diff
@@ -678,20 +679,19 @@ def substantiate(self, sb, build):
...
             if self.parent and not self.missing_timer:
                 # start timer.  if timer times out, fail deferred
                 self.missing_timer = self.master.reactor.callLater(
                     self.missing_timeout,
                     self._substantiation_failed, defer.TimeoutError())
-            self.substantiation_deferred = defer.Deferred()
             self.substantiation_build = build
             if self.conn is None:
                 d = self._substantiate(build)  # start up instance
                 d.addErrback(log.err, "while substantiating")
```
tardyp: @tomprince 

if substanciate errbacks, the notifier will never be called. Is this on purpose?

PR #2197, opened 2016-05-06
Use structured constants for state in workerforbuilder. 

tomprince:
Builds on #2196.

This makes it slightly easier debug what is going with state, since the constants have names, rather than just being small integers. Nothing outside this module uses or inspects the state (other than the no-op if removed in #2196).

### DIFF
--- a/master/buildbot/process/workerforbuilder.py
+++ b/master/buildbot/process/workerforbuilder.py
@@ -14,16 +14,19 @@
 # Copyright Buildbot Team Members
 from twisted.internet import defer
 from twisted.python import log
+from twisted.python.constants import NamedConstant
+from twisted.python.constants import Names
 
 from buildbot.worker_transition import WorkerAPICompatMixin
 
-(ATTACHING,  # worker attached, still checking hostinfo/etc
- IDLE,  # idle, available for use
- PINGING,  # build about to start, making sure it is still alive
- BUILDING,  # build is running
- LATENT,  # latent worker is not substantiated; similar to idle
- SUBSTANTIATING,
- ) = range(6)
+
+class States(Names):
+    ATTACHING = NamedConstant()  # worker attached, still checking hostinfo/etc
+    IDLE = NamedConstant()  # idle, available for use
+    PINGING = NamedConstant()  # build about to start, making sure it is still alive
+    BUILDING = NamedConstant()  # build is running
+    LATENT = NamedConstant()  # latent worker is not substantiated; similar to idle
+    SUBSTANTIATING = NamedConstant()
 
 
 class AbstractWorkerForBuilder(WorkerAPICompatMixin, object):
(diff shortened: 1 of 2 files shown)


master/buildbot/process/workerforbuilder.py
```diff
@@ -16,14 +16,16 @@
...
 
-(ATTACHING,  # worker attached, still checking hostinfo/etc
- IDLE,  # idle, available for use
- PINGING,  # build about to start, making sure it is still alive
- BUILDING,  # build is running
- LATENT,  # latent worker is not substantiated; similar to idle
- SUBSTANTIATING,
- ) = range(6)
+
+class States(Names):
```
tardyp: While the source code shows interesting use of metaclass programming, NamedConstant and Names could use a bit more documentation :)
for others reviewers

```
>>> States.ATTACHING
<States=ATTACHING>
```

tardyp:
looks good. perhaps the last commit should be discussed separately.

tardyp:
If I understand correctly, those state are purely for debugging, right? I can't see any conditions that would depend from state

master/buildbot/process/workerforbuilder.py
tomprince: There is a bunch of documentation at http://twistedmatrix.com/documents/current/core/howto/constants.html

tomprince:
No, `isBusy` inspects the state.

tomprince:
But nothing cares about the values of the constants, they are just checked for identity.

master/buildbot/process/workerforbuilder.py
```diff
@@ -14,16 +14,19 @@
...
+    ATTACHING = NamedConstant()  # worker attached, still checking hostinfo/etc
+    IDLE = NamedConstant()  # idle, available for use
+    PINGING = NamedConstant()  # build about to start, making sure it is still alive
+    BUILDING = NamedConstant()  # build is running
+    LATENT = NamedConstant()  # latent worker is not substantiated; similar to idle
+    SUBSTANTIATING = NamedConstant()
```
Stibbons: what is the advantage compared to Enum ?

tardyp: for what I can see Enum is python 3.4 only https://docs.python.org/3/library/enum.html
so we would have to depend on enum34

Stibbons: yes, ``pip install enum36``

tomprince: What are the advantages of `Enum` compare to this? One nice thing about this over enum is that we aren't forced to invent some meaningless values for the constants.

As an aside, `twisted.python.constants` predates the enum pep (which isn't a reason to use it, but is sufficient reason for it to exist; in addition to doing things that `enum` doesn't appear to do).

tomprince: Mostly though, the reason I used `twisted.python.constants` is that is what I'm familiar with. But being in the standard library doesn't necessarily make a choice better (see the pointer to requests [here](https://docs.python.org/2/library/urllib2.html) for example).

PR #2200, opened 2016-05-06
Drop py2.6 support from travis.

tomprince:
This is for testing travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ env:
   # - TWISTED=latest SQLALCHEMY=latest TESTS=setuppy_test
 
   # Configuration to run tests without installed buildbot-slave.
-  - TWISTED=latest SQLALCHEMY=latest TESTS=trial_worker
+  - TWISTED=latest SQLALCHEMY=latest TESTS=trial_worker_only
 
 cache:
   directories:
@@ -76,9 +76,9 @@ cache:
 matrix:
   include:
     - python: "2.6"
-      env: TWISTED=14.0.2 TESTS=worker-slave
+      env: TWISTED=14.0.2 TESTS=trial_worker
     - python: "2.6"
-      env: TWISTED=15.4.0 TESTS=worker-slave
+      env: TWISTED=15.4.0 TESTS=trial_worker
     # python 3 tests
 #    - python: "3.4"
 #      env: TWISTED=latest SQLALCHEMY=latest TESTS=py3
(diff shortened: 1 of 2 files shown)


tardyp:
not sure exactly why you want to hold this. 👍 please merge when you are ready

tomprince:
At least
- documentation
- updating metabbot to also drop the configuration.

.travis.yml
```diff
@@ -54,42 +73,14 @@ cache:
...
-      env: TWISTED=latest SQLALCHEMY=latest TESTS=coverage
-    - python: "2.7"
-      env: TWISTED=latest SQLALCHEMY=latest TESTS=lint
-    - python: "2.7"
-      env: TWISTED=latest SQLALCHEMY=latest TESTS=js
-    - python: "2.7"
-      env: TWISTED=latest SQLALCHEMY=latest TESTS=docs
-
+    - python: "2.6"
+      env: TWISTED=14.0.2 TESTS=worker-slave
```
rutsky: I suggest to use just `worker` here for indentifier.

rutsky: On the second though, I think better would be `workers`. And instead of `trial_worker` use `only_worker`.

rutsky: In accordance with bug 2340 I would like to reduce usage of "slave" as much, as possible.

tomprince:
https://github.com/buildbot/metabbotcfg/pull/40 should be merged and deployed before this lands.

.travis.yml
```diff
@@ -54,44 +75,14 @@ cache:
...
-      env: TWISTED=15.5.0 SQLALCHEMY=latest TESTS=trial
-
-    # add extra tests in separate jobs
-    - python: "2.7"
-      env: TWISTED=latest SQLALCHEMY=latest TESTS=js
-
+    - python: "2.6"
+      env: TWISTED=14.0.2 TESTS=trial_worker
+    - python: "2.6"
+      env: TWISTED=15.4.0 TESTS=trial_worker
```
rutsky: So effectively `trial_worker` is run of `trial buildslave.test buildbot_worker.test` on Python 2.6?

tomprince: Yes.

rutsky: This non-obvious to me from `trial_worker` name, but explains why:

1. `TWISTED=latest` is not specified for it (latest Twisted doesn't support Python 2.6).
2. `master` is not installed in this configuration (Python-level master-worker communication, i.e. local worker, works only with buildbot-worker on Python 2.7, which is tested in `TESTS=trial` configuration).

Maybe we should add comment regarding this?

rutsky:
With the exception of non-obvious `trial_worker` configuration behavior, changes LGTM.

PR #2166, opened 2016-04-27
log.py: Get deferred errors on addStd*/addContent

yetanotherion:
The previous version chose not to return deferreds
on some situations. When errors happened on
those lost deferreds, buildbot:
- reported "Unhandled error in Deferred"
- did not signal the error to the caller.

For this reason, this commit makes
- PlainLog.addContent
- StreamLog.addStdout/addStderr/addHeader

return a deferred instead.

Change-Id: I2dfae20b47e50e41d32b7c0366564580ad839767
Signed-off-by: Ion Alberdi <ion.alberdi@intel.com>

PR #2203, opened 2016-05-09
Fix issue in buildstep class

ssavrim:
Fix following issue raised during step execution:

File "/home/lab/dev_buildbot/src/buildbot-nine/master/buildbot/process/buildstep.py", line 170, in _catchup
    self.step._start_unhandled_deferreds.append(d)
  exceptions.AttributeError: 'NoneType' object has no attribute 'append'

Change-Id: I1abd303f3fa89f6a8d0409b79e4edfeb70144d29

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -167,7 +167,9 @@ class SyncLogFileWrapper(logobserver.LogObserver):
         def next(x):
             self._catchup()
             return x
-        self.step._start_unhandled_deferreds.append(d)
+
+        if self.step._start_unhandled_deferreds is not None:
+            self.step._start_unhandled_deferreds.append(d)
 
     def _delay(self, op):
         self.delayedOperations.append(op)
@@ -739,7 +741,8 @@ class BuildStep(results.ResultComputingConfigMixin,
         # create a logfile instance that acts like old-style status logfiles
         # begin to create a new-style logfile
         loog_d = self.addLog_newStyle(name, type, logEncoding)
-        self._start_unhandled_deferreds.append(loog_d)
+        if self._start_unhandled_deferreds is not None:
+            self._start_unhandled_deferreds.append(loog_d)
         # and wrap the deferred that will eventually fire with that logfile
         # into a write-only logfile instance
         wrapper = SyncLogFileWrapper(self, name, loog_d)


master/buildbot/process/buildstep.py
```diff
@@ -167,7 +167,9 @@ def _catchup(self):
         def next(x):
             self._catchup()
             return x
-        self.step._start_unhandled_deferreds.append(d)
+
+        if self.step._start_unhandled_deferreds is not None:
```
tardyp: _start_unhandled_deferreds shall never be none if this code is called.

if it is none, then there is another problem in the step (related to the new style VS old style)

http://docs.buildbot.net/latest/manual/new-style-steps.html

tardyp:
looks like this is a regression from recently.

Please see analysis from @gracinet 
http://trac.buildbot.net/ticket/3546

gracinet:
@ssavrim in what kind of step do you see this ? If it's not a `MasterShellCommand`, it's worth knowing

ssavrim:
My step inherits from a buildbot.steps.shell.ShellCommand. So it is not a MasterShellCommand.
I will try to test reverting the following patch:
[d17efcab985846551684379bd22d5c1648ddb88e] log.py: change acquire/release by lock.run

bastien31:
By reverting https://github.com/buildbot/buildbot/pull/2166 locally, issue is no more present.

bastien31:
By reverting https://github.com/buildbot/buildbot/pull/2166/commits/ae945f70faa8e0845f115b903aaa3c03f8223dd8, issue is no more present.

master/buildbot/process/buildstep.py
yetanotherion: Note that the maybeUnhandled decorator has the same "if..."
https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/buildstep.py#L111

tardyp:
I believe this is fixed with https://github.com/buildbot/buildbot/pull/2210 (using easy reproducible testcase from @gracinet )

please let me know if this works for you

tardyp:
I don't exactly understand why ae945f7 show the problem. I think this might be because of the inlinecallbacks

tardyp:
closing in favor of #2210

ssavrim:
Thanks for the fix

PR #2204, opened 2016-05-09
Remove PINGING state in workerforbuilder.

tomprince:
Nothing actually inspects this particular state, except checking if this
workerforbuilder is available. Ping is called in two places:

- When a build starts, the builder pings the worker. During the ping,
  the BRD activity lock is held, so nothing cares what state the builder
  is in.
- When `BuilderControl.ping` is called, it pings all attached workers.
  This could in theory have overlapped with trying to schedule a build,
  causing a build on this builder (or perhaps even another builder in
  the presence of `max_builds` on a worker) to not be started. I think
  allowing a build to start here is entirely harmless, and in might in
  fact fix a potential bug. If no workers were available due
  to this type of ping, then nothing would try to schedule the build
  once the pings were completed. In any case, the old webstatus had a
  button for triggering this ping, but there is no code left that does
  it.


tardyp:
👍

PR #2208, opened 2016-05-09
Revert "Replace print with log.msg()"

rutsky:
This reverts commit 4d733e62d5570fdb9e603ebfeb333899bdb1f710.

Use of log was removed since it doesn't work for command line utility.
Fixes missing *all* output from `buildslave` command.

`print` function is used instead of `print` statement (in changed places, there is still use of `print` statement in other places).
Change to new-style exceptions handling left intact.
`fn.__name__` is still used instead of `fn.func_name` (works in Pythin 2.6, 2.7 and 3).

### DIFF
--- a/master/contrib/coverage2text.py
+++ b/master/contrib/coverage2text.py
@@ -2,10 +2,11 @@
 
 import sys
 
+from twisted.python import usage
+
 from coverage import coverage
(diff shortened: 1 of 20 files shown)


djmitche:
:+1:

PR #2212, opened 2016-05-10
Start latent workers as part of the build, rather than before the build.

tomprince:
- This makes the claimed buildrequest visible as a build .
- This makes it so that starting workers isn't serialized as part of the BRD's activity loop.

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -273,14 +273,7 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
         try:
             self.setupBuild()  # create .steps
         except Exception:
-            # the build hasn't started yet, so log the exception as a point
-            # event instead of flunking the build.
-            # TODO: associate this failure with the build instead.
-            # this involves doing
-            # self.build_status.buildStarted() from within the exception
-            # handler
-            log.msg("Build.setupBuild failed")
-            log.err(Failure())
+            log.err(Failure(), "Build.setupBuild failed")
             self.buildFinished(['Build.setupBuild', 'failed'], EXCEPTION)
             return
 
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -214,8 +214,6 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
             props.setProperty("project", source.project, "Build")
 
     def setupWorkerForBuilder(self, workerforbuilder):
-        self.workerforbuilder = workerforbuilder
-
         self.path_module = workerforbuilder.worker.path_module
 
         # navigate our way back to the L{buildbot.worker.Worker}
@@ -237,12 +235,10 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
         """This method sets up the build, then starts it by invoking the
         first Step. It returns a Deferred which will fire when the build
         finishes. This Deferred is guaranteed to never errback."""
-        worker = workerforbuilder.worker
+        self.workerforbuilder = workerforbuilder
+        self.conn = None
 
-        # we are taking responsibility for watching the connection to the
-        # remote. This responsibility was held by the Builder until our
-        # startBuild was called, and will not return to them until we fire
-        # the Deferred returned by this method.
+        worker = workerforbuilder.worker
 
         log.msg("%s.startBuild" % self)
 
@@ -261,14 +257,10 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
                                                                       str(self.buildid),
                                                                       "stop"))
         self.setupOwnProperties()
-        self.setupWorkerForBuilder(workerforbuilder)
 
         # then narrow WorkerLocks down to the right worker
-        self.locks = [(l.getLock(self.workerforbuilder.worker), a)
+        self.locks = [(l.getLock(workerforbuilder.worker), a)
                       for l, a in self.locks]
-        self.conn = workerforbuilder.worker.conn
-        self.subs = self.conn.notifyOnDisconnect(self.lostRemote)
-
         metrics.MetricCountEvent.log('active_builds', 1)
 
         # make sure properties are available to people listening on 'new'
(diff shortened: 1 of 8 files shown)


master/buildbot/process/build.py
```diff
@@ -291,8 +283,59 @@ def startBuild(self, build_status, workerforbuilder):
...
+            self.buildFinished(["worker", "not", "available"], RETRY)
+            return
+
+        # If prepare returns True then it is ready and we start a build
+        # If it returns false then we don't start a new build.
+        if not ready:
+            log.msg("worker %s can't build %s after all; retrying the "
+                    "build" % (self, workerforbuilder))
+            self.stopped = True
+            self.buildFinished(["worker", "not", "available"], RETRY)
```
tardyp: shall we yield this one?

master/buildbot/process/build.py
```diff
@@ -291,8 +283,59 @@ def startBuild(self, build_status, workerforbuilder):
...
             return
+
         # flush properties in the beginning of the build
         yield self._flushProperties(None)
+
+        try:
+            ready = yield workerforbuilder.prepare(self)
+        except Exception:
+            log.err(Failure(), 'while preparing workerforbuilder:')
+            self.buildFinished(["worker", "not", "available"], RETRY)
```
tardyp: and that one

tardyp: note the status text is actually ignored. that text will never be displayed (perhaps, we should fix that in self.buildFinished)

master/buildbot/process/workerforbuilder.py
```diff
@@ -21,11 +21,12 @@
...
-    IDLE = NamedConstant()  # idle, available for use
-    BUILDING = NamedConstant()  # build is running
-    LATENT = NamedConstant()  # latent worker is not substantiated; similar to idle
-    SUBSTANTIATING = NamedConstant()
+    # The worker isn't attached, or is in the process of attaching.
+    DETACHED = NamedConstant()
+    # The worker is available to build: either attached, or a latent worker.
+    AVAILABLE = NamedConstant()
+    # The worker is building.
+    BUILDING = NamedConstant()
```
tardyp: good

tardyp:
I think that would make sense to set the status of the build to a meaningful text during the process of substanciation, in order to help users to understand what is going on with their build, which have a slave but not step yet.

            yield self.master.data.updates.setBuildStateString(self.buildid,
                                                               u'substanciating slave')

ryansydnor:
@tardyp - great idea. The current UX has people in the dark about what's happening behind the scenes. Letting them know Buildbot is spinning up a latent worker in the background could be quite valuable.

tomprince:
Updated with a (so far) fixed build.

ryansydnor:
LGTM!

master/buildbot/process/build.py
tardyp: please yield!

master/buildbot/process/build.py
```diff
@@ -281,20 +273,73 @@ def startBuild(self, build_status, workerforbuilder):
...
+        # fallen off the map (due to a NAT timeout or something), this
+        # will fail in a couple of minutes, depending upon the TCP
+        # timeout.
+        #
+        # TODO: This can unnecessarily suspend the starting of a build, in
+        # situations where the worker is live but is pushing lots of data to
+        # us in a build.
+        yield self.master.data.updates.setBuildStateString(self.buildid,
+                                                           u'pinging worker')
+        log.msg("starting build %s.. pinging the worker %s"
```
tardyp: I dont think we shall log as common thing as starting a build

master/buildbot/process/build.py
```diff
@@ -281,20 +273,73 @@ def startBuild(self, build_status, workerforbuilder):
...
+                % (self, workerforbuilder))
+        try:
+            ping_success = yield workerforbuilder.ping()
+        except Exception:
+            log.err(Failure(), 'while pinging worker before build:')
+            ping_success = False
+
+        if not ping_success:
+            log.msg("worker ping failed; retrying the build")
+            self.buildFinished(["worker", "not", "pinged"], RETRY)
```
tardyp: yield

master/buildbot/process/build.py
```diff
@@ -281,20 +273,73 @@ def startBuild(self, build_status, workerforbuilder):
...
+        self.conn = workerforbuilder.worker.conn
+        self.setupWorkerForBuilder(workerforbuilder)
+        self.subs = self.conn.notifyOnDisconnect(self.lostRemote)
+
+        # tell the remote that it's starting a build, too
+        try:
+            yield self.conn.remoteStartBuild(self.builder.name)
+        except Exception:
+            log.err(Failure(), 'while calling remote startBuild:')
+            self.buildFinished(["worker", "not", "building"], RETRY)
```
tardyp: yield

master/buildbot/process/build.py
tomprince: I don't disagree, but this code is just moved, here. I'd rather address logging separately.

master/buildbot/process/build.py
tomprince: No. The only thing waiting for `buildStarted` is a last-ditch exception logger, and `buildFinished` already has one of those. In addition, `buildFinished` is designed to be called in response to events (where there is nothing waiting on the result), instead of as part of a sequence of action that need to be done serially. If I was writing `buildFinshed` without `inlineCallbacks` (i.e. with raw deferreds), I wouldn't have it return a deferred, so the fact that it does is merely an implementation detail.

tardyp: Well I am a little embarrassed with such reasoning of "it is not needed with current implementation", and usually prefer to just always wait for all the asynchronous stuff that we call.
It is so hard to debug when there are race conditions because of unwaited deferred.

Its more like a coding style rule. You shalt always wait for the deferred, and don't try to be smart at it.
Another additional remark on that topic is that I am not sure if SynchronousTestCase do help catching unwaited deferred.

cmouse:
Related to this, when substantiating OpenStack latent slave, it can sometimes take quite long (for us, a fresh image takes 3 minutes to substantiate on first run) during which no status is given in buildbot that anything is happening. You can only see that things are moving by looking at logs. If this pr can fix this behaviour it would be greatly appreciated. Maybe put step -1: substantiating?

master/buildbot/process/build.py
tomprince: `yield`ing here is *incorrect*.

> Well I am a little embarrassed with such reasoning of "it is not needed with current implementation"

It has nothing to do with the details of the current implementation, but rather the intent.
`buildFinished` doesn't semantically have a return value.

> You shalt always wait for the deferred, and don't try to be smart at it.

Blindly doing that is part of why latent workers start serially, rather than in parallel.

tardyp:
let move forward thanks for the detailled explainations!

PR #2039, opened 2016-03-09
Docker worker network config

jlstewart379:
This PR allows the networking config to be set when launching a latent docker worker.

PR #2213, opened 2016-05-11
Change default value for networking_config

jlstewart379:
The current default value for networking config cannot be read by docker and causes the DockerLatentWorker to error. If the `networking_config` option is not set, it should default to None and if specified, should take a dictionary of values. The original implementation was supposed to pass network_config to the docker client's `start()` method in order to set the `network_mode`, but I inadvertently added it to the client's `create_container` call in PR https://github.com/buildbot/buildbot/pull/2039. If a DockerLatentWorker is used, the current workaround is to set networking_config to None in the constructor, defeating the spirit of being optional. Since the host_config parameter sets the network mode, no need to bother with the docker client's `start()` method.

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -59,7 +59,7 @@ class DockerLatentWorker(AbstractLatentWorker):
 
     def __init__(self, name, password, docker_host, image=None, command=None,
                  volumes=None, dockerfile=None, version=None, tls=None, followStartupLogs=False,
-                 masterFQDN=None, hostconfig=None, networking_config='bridge', **kwargs):
+                 masterFQDN=None, hostconfig=None, networking_config=None, **kwargs):
 
         if not client:
             config.error("The python module 'docker-py>=1.4' is needed to use a"
--- a/master/docs/manual/cfg-workers-docker.rst
+++ b/master/docs/manual/cfg-workers-docker.rst
@@ -221,7 +221,8 @@ In addition to the arguments available for any :ref:`Latent-Workers`, :class:`Do
     Extra host configuration parameters passed as a dictionary used to create HostConfig object. See `docker-py's HostConfig documentation <http://docker-py.readthedocs.org/en/latest/hostconfig/>`_ for all the supported options.
 
 ``networking_config``
-  Set the network configuration for the docker container. It can be one the following: 'bridge', 'host', container:<NAME or ID>, none. The default is bridge. This option is equivalent to using the ``--net=`` command line parameter in docker. 
+  (optional)
+  Extra network configuration settings passed as a dictionary used to create NetworkConfig object. See this docker-py `test <https://github.com/docker/docker-py/blob/81edb398ebf7ce5c7ef14aa0739de5329589aabe/tests/unit/container_test.py#L988>`_ for usage example.
 
 Setting up Volumes
 ..................


tardyp:
I am not sure if you are not changing the default.

If a user does not specify network config, will it be bridge by default?

jlstewart379:
@tardyp Currently it bridge will be passed in if unspecified and if bridge is passed in as it currently does, the docker client's create_container() passes the string 'bridge' to the docker code and there it tries to make a networkConfig object out of it, which throws a 500.  I actually meant to pass the networking_config parameter through to the docker_client's start method as in `docker_client.start(instance, network_mode=self.networking_config)` in the DockerLatentWorker. Even that probably was not necessary since pull https://github.com/buildbot/buildbot/pull/2036 added a host_config parameter, which is now the preferred way to configure network mode.  I think I became a little mixed up going between networking_config and network_mode and create_container() and start() as I was working on the patch.

Now, if we specify None as the default for networking_config then it will eventually be dealt with [here](https://github.com/docker/docker/blob/75e63c942ec8d3852f182b9cd9c221411fc74730/daemon/create.go#L122) and the config for that network_mode will not be changed from the default. Currently it looks like networking_config can only be used to set aliases for containers, making the containers available at different endpoints.  Does this make sense?

tardyp:
It is still not clear to me that with you change the following default configuration

DockerLatentBuildSlave('name', 'passwd')

will create a docker slave with access to internet, as one would normally expect.
I would prefer that ppl wanting to set network_config will have to set network_mode to null themselves

jlstewart379:
For safety, I would error on the side of taking the parameter out all together for the DockerLatentWorker. Leaving it the way it is now means that DockerLatentWorker('name', 'password') will error with 

`APIError: 500 Server Error: Internal Server Error ("json: cannot unmarshal string into Go value of type network.NetworkingConfig")`

My original intent was to pass networking_config into docker to define the network_mode, but I introduced the bug by putting it in the wrong spot, confusing network_mode in start_container with networking_config in create_container. 

To reproduce, just use docker-py to simulate the [DockerLatentWorker call to docker-py](https://github.com/buildbot/buildbot/blob/master/master/buildbot/worker/docker.py#L189)

```python
import docker

cli = Client(base_url='unix://var/run/docker.sock')
cli.start_container('name', networking_config='bridge')
```
Which produces:

```shell
APIError                                  Traceback (most recent call last)
<ipython-input-10-a4918d44bf43> in <module>()
----> 1 cli.create_container('name', networking_config='bridge')

/usr/local/lib/python2.7/dist-packages/docker/api/container.pyc in create_container(self, image, command, hostname, user, detach, stdin_open, tty, mem_limit, ports, environment, dns, volumes, volumes_from, network_disabled, name, entrypoint, cpu_shares, working_dir, domainname, memswap_limit, cpuset, host_config, mac_address, labels, volume_driver, stop_signal, networking_config)
    118             volume_driver, stop_signal, networking_config,
    119         )
--> 120         return self.create_container_from_config(config, name)
    121
    122     def create_container_config(self, *args, **kwargs):

/usr/local/lib/python2.7/dist-packages/docker/api/container.pyc in create_container_from_config(self, config, name)
    129         }
    130         res = self._post_json(u, data=config, params=params)
--> 131         return self._result(res, True)
    132
    133     def create_host_config(self, *args, **kwargs):

/usr/local/lib/python2.7/dist-packages/docker/client.pyc in _result(self, response, json, binary)
    156     def _result(self, response, json=False, binary=False):
    157         assert not (json and binary)
--> 158         self._raise_for_status(response)
    159
    160         if json:

/usr/local/lib/python2.7/dist-packages/docker/client.pyc in _raise_for_status(self, response, explanation)
    152             if e.response.status_code == 404:
    153                 raise errors.NotFound(e, response, explanation=explanation)
--> 154             raise errors.APIError(e, response, explanation=explanation)
    155
    156     def _result(self, response, json=False, binary=False):

APIError: 500 Server Error: Internal Server Error ("json: cannot unmarshal string into Go value of type network.NetworkingConfig")
```

tardyp:
I understand the current implementation does not work with default value unfortunately

I just want to have a clear statement that the new default values are sane.
I think we don't want to have a default latent worker which does not have access to the network.

if there is a need for more time, we can revert #2036. I wish we can have more comment from @anish who could have introduced the bug.

anish:
@tardyp Not my bug (I hope) host_config introduced in #2036 is the preferred way to do this. When I did this implementation (which was just a few days before @jlstewart379 did his change), there was no way to pass host_config at all, which resulted in my dockerlatentworkers always coming up with no network access. I left the default as is (no defaults specified), and just added the ability to specify any host config options, network is just one of them.

tardyp:
also reported here: http://trac.buildbot.net/ticket/3549

Guys, I still do not have a clear answer on how to resolve this.

- Most users will not want to dig into the docker manual and expect sane working default.
- Most users just want to have transparent network access from the worker, which is what I believe was how buildbot behave before #2036.

Right now my understanding of the best approach is to just revert #2036.

anish:
@tardyp can you confirm reverting #2036 is going to fix this. The whole reason I did #2036 is because the containers came up with no network access, definitely no transparent network access pre 2036 in any of my testing.

I am still curious why #2036 would make a difference here. 'network_mode' is an example for the unit test, the default does not touch network mode at all : https://github.com/buildbot/buildbot/pull/2036/files#diff-3cb0ce94f8cea8a51d45342f0a693dfcR161

Please see how docker-py uses host_config vs networking_config, they are two very different things. An empty host_config does not override anything as by itself it is not a parameter of any sort for docker.

anish:
@tardyp see here how network_mode is being used when not found in host_config https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L690

which is the case pre and post #2036 , since in either case there is no network_mode specified by default

tardyp:
@anish. I am using beta9 in prod with docker latent workers without issues. 

docker-py==1.7.2

so now I am confused.. I dont want to brake my prod by trying to upgrade.

anish:
I don't have access to machines till friday, but I can test after that, or I can see if I can finish the master dockerfile. But really, if you look at what changed in #2029 and #2036 you'll see that none of them include any changes to the network configuration by default.

jlstewart379:
@anish @tardyp My PR, https://github.com/buildbot/buildbot/pull/2039, is the problem. `Networking_config` does not and should not specify `network_mode`. The `host_config` stuff from https://github.com/buildbot/buildbot/pull/2036 has worked for me, allowing me to set the network_mode by setting `host_config={"network_mode": "host"}.` `networking_config` does something different. It looks like its useful for things like setting an IP address for a specified network. There is some documentation on it [here](https://github.com/docker/docker/blob/55bc0a344085338826bca00ea75ffb9e4b5e0c79/docs/reference/api/docker_remote_api_v1.22.md). I'd revert https://github.com/buildbot/buildbot/pull/2039 and leave https://github.com/buildbot/buildbot/pull/2036.

anish:
@jlstewart379 that was my understanding as well, fine by me

rutsky:
@tardyp is this PR still needs reply?

tardyp:
I created http://trac.buildbot.net/ticket/3571 in order to look at the state of docker latent.

I am not sure of the current state of the latent workers with this PR, and the confusion around it.

tardyp:
in favor of #2296

PR #2214, opened 2016-05-11
buildstep: finish unfinished logs

yetanotherion:
The previous version did not call finish on
unfinished (log.finished == False) logs. Such
situations happened among others with logs not attached to
a RemoteCommand, like for example stdio on
a MasterShellCommand.

Not calling finish had at least, two side effects:
- the 'complete' of related log lines attribute was not set to 1
  in the database,
- master.data.updates.compressLog was not called on the final
  state of the log.

For these reasons, this commit calls 'finish' on unfinised
logs before ending the step. The result of the step is set
to EXCEPTION if an error occured during that operation.

Change-Id: I30ebfc51ffac12cab22f4864494b3a334a202dac

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -560,10 +560,26 @@ class BuildStep(results.ResultComputingConfigMixin,
 
         yield self.master.data.updates.finishStep(self.stepid, self.results,
                                                   hidden)
-
+        # finish unfinished logs
+        all_finished = yield self.finishUnfinishedLogs()
+        if not all_finished:
+            self.results = EXCEPTION
         self.releaseLocks()
 
         defer.returnValue(self.results)
+
+    @defer.inlineCallbacks
+    def finishUnfinishedLogs(self):
+        ok = True
+        not_finished_logs = [v for (k, v) in iteritems(self.logs)
+                             if not v.finished]
+        finish_logs = yield defer.DeferredList([v.finish() for v in not_finished_logs],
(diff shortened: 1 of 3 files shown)


tardyp:
👍

tardyp:
looks like travis is unhappy

yetanotherion:
@tardyp travis is now happy

aelsabbahy:
I have a subclass of ShellCommand that records log/html files. After this latest PR my builds fail with an exception status and the following errors are in twisted.log: `2016-05-17 18:22:13+0000 [-] when trying to finish a log` and `assert not self.finished`

It seems my subclass is causing `finish()` to be called twice with this change, how should my code be changed to be compatible with this PR?

```
    @defer.inlineCallbacks
    def handle_log_files(self):
        for logname, logpath in self.log_files.iteritems():
            content = yield self.getFileContentFromWorker(logpath)
            if not content:
                continue
            if logpath.lower().endswith('.html'):
                self.addHTMLLog(logname, content)
            else:
                self.addCompleteLog(logname, content)

    @defer.inlineCallbacks
    def run(self):
        parent_results = yield super(SubClassShellCommand, self).run()
        yield self.handle_log_files()
        defer.returnValue(parent_results)
```

tardyp:
@aelsabbahy you can start by just yielding calls to log api
```
               continue
            if logpath.lower().endswith('.html'):
                yield self.addHTMLLog(logname, content)
            else:
                yield self.addCompleteLog(logname, content)

```

PR #2215, opened 2016-05-11
.travis.yml: put lint tests first

yetanotherion:
More than 20 travis jobs are currently
executed to validate a pull request. Travis
however doesn't start the 20 builds in parallel.

The link tests are therefore executed after
previous ones are finished.
To avoid situations where all the unit tests are ok,
but one needs to wait 20 minutes to be aware
of a lint issue (incorrect order of import), which is quite likely
to happen, this commit puts the lint test first.

Change-Id: Id57d6d9fec4742cc7ea768d888e058835ebb2113

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,9 +54,13 @@ cache:
 
 matrix:
   include:
-    # lint first as they're more likely to find issues
+    # lint, docs and coverage first as they're more likely to find issues
     - python: "2.7"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=lint
+    - python: "2.7"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=docs
+    - python: "2.7"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=coverage
 
     # Test different versions of SQLAlchemy
     - python: "2.7"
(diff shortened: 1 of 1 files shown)


ryansydnor:
I'd probably move coverage and docs up there as well to catch things like spelling errors and poorly formatted rst more quickly.

tardyp:
yes, I have been wanting to do that for long

tardyp:
agree, please also put doc first.

yetanotherion:
@tardyp @ryansydnor I put docs and coverage up

yetanotherion:
@ryansydnor @djmitche @tardyp the patch does not seem to work
https://travis-ci.org/buildbot/buildbot/builds/129648472
the lint test is in 17th position. Do you think it's a way for travis to make people subscribe to a service ?

ryansydnor:
It looks like the newer PRs (perhaps those that are rebased on top of your change) are behaving as expected. (for example, https://travis-ci.org/buildbot/buildbot/builds/129687285)

I realized that this change could make the builds even more efficient (from both a time and resource perspective) if we fail fast. Fortunately travis gives us that capability! I've opened another PR here: https://github.com/buildbot/buildbot/pull/2226

PR #2217, opened 2016-05-11
switch setup.py[bundle] to buildbot_worker instead of slave

tardyp:
The release for buildot 0.9.0b9 included buildbot_slave in the [bundle]


rutsky:
LGTM.

PR #2220, opened 2016-05-12
fix pb.NoSuchMethod checking

rutsky:
My testing shows, that in case of remote `pb.NoSuchMethod` error, caller site received pb.RemoteError with information about remote exception, and NOT `pb.NoSuchMethod` exception.

`remoteGetWorkerInfo()` tested with WIP e2e tests in context of transition to new worker protocol.

`remoteShutdown()` is not tested.


tardyp:
:+1:

PR #2221, opened 2016-05-12
reraise error if remoteGetWorkerInfo() fails

rutsky:
otherwise we'll see in log "rejecting duplicate worker" error


master/buildbot/worker/manager.py
```diff
@@ -131,7 +131,7 @@ def newConnection(self, conn, workerName):
         except Exception as e:
             log.msg("Failed to communicate with worker '%s'\n"
                     "%s" % (workerName, e))
-            defer.returnValue(False)
```
tardyp: raise

works better

rutsky:
@tardyp thanks, I used just `raise` instead where you pointed out, and in a place above. Haven't tested, lets wait for CI to finish.

PR #2136, opened 2016-04-19
POC e2e test

rutsky:
Based on PR #2134, which is based on #2105. Relevant changes: https://github.com/buildbot/buildbot/commit/df8900fb239afec7d924dff008a5d020e6749130

POC of e2e tests. Creates master, creates worker, starts master, starts worker, shutdowns worker, shutdowns master.

---

Current progress:

* [x] Setup/teardown tests
* [x] Run tests in daemon and non-daemon mode (Buildbot/Twisted doesn't handle daemonization Windows)
* [x] Tests with deprecated Buildslave
* [x] Tests with pure Python 3 worker
* [ ] Tests with pure Python 2 worker
* [ ] Tests with single-file worker
* [ ] Fix tests on Windows

PR #2222, opened 2016-05-12
rename getSlaveInfo to getWorkerInfo in new master-worker protocol

rutsky:
Based on PR #2220.

This is (expected) backward incompatible change in master-worker protocol. New worker now will work only with new master (which is expected too).
Few more similar (and incompatible) change are upcoming for [bug 2340](http://trac.buildbot.net/ticket/2340).

Tested in WIP e2e tests in PR #2136


jaredgrubb:
LGTM!

PR #2223, opened 2016-05-12
require boto3 in EC2LatentWorker tests

rutsky:
Tests calls EC2LatentWorker, and EC2LatentWorker requires boto3 now.

### DIFF
--- a/master/buildbot/test/unit/test_worker_ec2.py
+++ b/master/buildbot/test/unit/test_worker_ec2.py
@@ -24,8 +24,9 @@ from buildbot.worker_transition import DeprecatedWorkerNameWarning
 try:
     from moto import mock_ec2
     assert mock_ec2
+    # Tests are using boto, but EC2LatentWorker required boto3.
     import boto
-    assert boto
+    import boto3  # noqa pylint: disable=unused-import
 except ImportError:
     boto = None
     ec2 = None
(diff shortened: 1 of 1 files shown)


master/buildbot/test/unit/test_worker_ec2.py
```diff
@@ -38,7 +39,7 @@
 # redefine the mock_ec2 decorator to skip the test if boto or moto
 # isn't installed
 def skip_ec2(f):
-    f.skip = "boto or moto is not installed"
+    f.skip = "boto, or boto3, or moto is not installed"
```
tardyp: boto3 or moto

aelsabbahy:
The test suite should be upgraded to boto3 for completeness, but not following why this import is needed while tests are still using boto2.

rutsky:
@tardyp, @aelsabbahy tests require boto2 (named "boto") and moto. But code which is tested (i.e. EC2LatentWorker) require boto3 now.
So all three are currently needed to run existing tests.
And we need to query all three libraries in import in order to have quick and correct message about why tests are skipped (instead of some strange import errors in all EC2 tests when boto3 is not installed).

tardyp:
I did not understand the tests where needed boto2 while the code needed boto3. This does not look sane. I wouldn't have merged the code if I knew that. Please let me know when we can fix that.

aelsabbahy:
Kept the tests boto2 because I was concerned about boto3 breaking things and wanted to validate against the old mocks/tests first to ensure the boto3 changes were correct. Basically, didn't want to get in a situation were the tests are passing but the code is doing the wrong thing.

I'm on vacation for two weeks, after that I can migrate the tests to boto3.

aelsabbahy:
I started work on it here: #2247

tardyp:
as I merged #2247, I believe this one is obsolete

PR #2224, opened 2016-05-12
builder: _defaultCollapseRequest is a staticmethod

yetanotherion:
The collapse related unit tests in test_process_buildrequest
had two issues:
1. the deferred in the test_* method calling self.do_request_collapse,
   were not returned
2. the mock of getCollapseRequestsFn in *_collapse_default did not
   permit to catch the following bug (
   Exceptions.TypeError: unbound method _defaultCollapseRequestFn()
must be called with Builder instance as first argument
(got FakeMaster instance instead))

This commit fixes the bug (_defaultCollapseRequest should be put
as a staticmethod) and reproduces/validates it modifying related
unit tests.

Change-Id: Iec10a70e5f2d8b96c1f4a6bdb184be6638b88818

### DIFF
--- a/master/buildbot/process/builder.py
+++ b/master/buildbot/process/builder.py
@@ -534,7 +534,7 @@ class Builder(util_service.ReconfigurableServiceMixin,
         if collapseRequests_fn is False:
             collapseRequests_fn = None
         elif collapseRequests_fn is True:
-            collapseRequests_fn = Builder._defaultCollapseRequestFn
+            collapseRequests_fn = self._defaultCollapseRequestFn
 
         return collapseRequests_fn
 
--- a/master/buildbot/process/builder.py
+++ b/master/buildbot/process/builder.py
@@ -538,7 +538,8 @@ class Builder(util_service.ReconfigurableServiceMixin,
 
         return collapseRequests_fn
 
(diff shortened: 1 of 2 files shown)


yetanotherion:
@tardyp @djmitche I'm afraid another release cannot be done without this fix
(all builders that don't redefine collapseRequest are impacted by this bug).

tardyp:
You mean beta 9 had this bug? It this is since beta 9?

Le ven. 13 mai 2016 10:43, yetanotherion <notifications@github.com> a
écrit :

yetanotherion:
It may be since 
7d6ed38410e14af18665787b6c9951d579b694e5

master/buildbot/process/builder.py
```diff
@@ -538,7 +538,8 @@ def getCollapseRequestsFn(self):
 
         return collapseRequests_fn
 
-    def _defaultCollapseRequestFn(self, master, builder, brdict1, brdict2):
+    @staticmethod
```
tardyp: we can also say s/Builder._defaultCollapseRequestFn/self._defaultCollapseRequestFn/ line 537

yetanotherion:
@tardyp your suggestion is implemented in the last commit

PR #2225, opened 2016-05-12
trigger: initialize brids

yetanotherion:
When an exception occurs brids is not initalized
and the step fails on the line
self.brids.extend(itervalues(brids))

Change-Id: Ifd7dab2fb0ee27e773ffb3d629ffd42c139c9867


sa2ajj:
LGTM

PR #2226, opened 2016-05-12
Fail fast in the travis build

ryansydnor:
https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ cache:
     - www/nestedexample/libs
 
 matrix:
+  fast_finish: true
   include:
     # Tests of buildbot-worker and buildbot-slave on python 2.6
     - python: "2.6"


rutsky:
As I understood description in docs this will finish (abort) Travus build as soon, as at least one job will fail.

I don't like this change: it gives us faster red icon on github if something fails at the cost that not all tests will be run.
So if, for example, you have linting issue some of trial tests or docs test will not be run at all, until you fix linting issue.

I would prefer to see *all* failing tests as a result of Travis build.
If you want to get faster build failure result you can always open Travis page with builds and see what builds already failed.

ryansydnor:
@rutsky - I have the same understanding. Travis will abort when something fails.

I was initially under the impression that a project has a fixed number of workers allocated (5 concurrent builds at a time: https://twitter.com/travisci/status/651856122559774722). If that were the case, this would give the Buildbot project more overall throughput.

However, it seems like the builders are shared across all of travis-ci.org (https://blog.travis-ci.com/2014-07-18-per-repository-concurrency-setting/). If we implement this change we'll be better community members (return workers to the pool more efficiently) but minimize our own testing effectiveness. 

If the latter is true, it probably makes sense to abandon this and look for other ways to speed up the build. Curious to hear what @tardyp thinks.

tardyp:
I think we still want to have maximum amount of failure as possible.

it is not clear in doc if travis will report failure immediately, but continue to see if there arn't any other issues

ryansydnor:
Should we try and revert if it aborts the build rather than simply reporting a failure?

rutsky:
@ryansydnor I suggest to make test in this PR (which can be removed later via force push) --- you can add lint error and lets see if Travis will run other jobs.

rutsky:
@tardyp instead testing behavior of this change in this PR we will wait and see how this change behaves in actual builds and probably revert these changes?

PR #2227, opened 2016-05-12
Upgrade pip in travis.

tomprince:
Newer pip will cache wheels for installed packages, which should speed up the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,9 +89,9 @@ matrix:
 install:
   - pip install -U pip
   # codecov is the interface to codecov.io; see after_success
-  - "echo 'travis_fold:start:codecov'; pip install codecov; echo 'travis_fold:end:codecov'"
-  - "echo 'travis_fold:start:tw'; [ $TWISTED = latest ] || pip install Twisted==$TWISTED; echo 'travis_fold:end:tw'"
-  - "echo 'travis_fold:start:sa'; [ $SQLALCHEMY = latest ] || pip install sqlalchemy==$SQLALCHEMY; echo 'travis_fold:end:sa'"
+  - pip install codecov
+  - "[ $TWISTED = latest ] || pip install Twisted==$TWISTED"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -87,10 +87,11 @@ matrix:
 
 # Dependencies installation commands
 install:
+  - pip install -U pip
   # codecov is the interface to codecov.io; see after_success
-  - "echo 'travis_fold:start:codecov'; pip install codecov; echo 'travis_fold:end:codecov'"
-  - "echo 'travis_fold:start:tw'; [ $TWISTED = latest ] || pip install Twisted==$TWISTED; echo 'travis_fold:end:tw'"
-  - "echo 'travis_fold:start:sa'; [ $SQLALCHEMY = latest ] || pip install sqlalchemy==$SQLALCHEMY; echo 'travis_fold:end:sa'"
+  - pip install codecov
```
tomprince: Travis automatically folds individual install steps, so folding again inside the step is a waste (and just makes seeing the contents of the fold more annoying).

tomprince:
Note that because of how cache inheritance works, this will probably only show a benefit after a build of this on master.

rutsky:
Nice side effect: after removing folding we see that there were actually an error: https://travis-ci.org/buildbot/buildbot/jobs/129764461#L334

We need to add `SQLALCHEMY=latest` (or any other version) to `TESTS=trial_worker` specification.

tomprince:
Updated to (hopefully) fix the build.

.travis.yml
```diff
@@ -77,20 +77,22 @@ matrix:
   fast_finish: true
   include:
     # Tests of buildbot-worker and buildbot-slave on python 2.6
+    # Specify SQLALCHEMY=latest to avoid errors installling.
```
rutsky: There is a typo: "insta**lll**ing"

PR #2228, opened 2016-05-12
output list of installed python packages at the end of install step

rutsky:
Output will be collapsed for the same reason as in PR #2227.
Previously if some of testing steps failed `pip list` command is not issued at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,8 +147,6 @@ install:
   # https://github.com/pypa/setuptools/issues/500
   - "[ $TESTS != setuppy_test ] || pip install autobahn Twisted"
 
-  - "pip list"
-
 before_script:
   # create real MySQL database for tests
   - mysql -e 'create database bbtest;'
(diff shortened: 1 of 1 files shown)


tomprince:
Are you sure that is the case? The [documentation](https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step) suggests that all the steps will be run (and [this buld](.https://travis-ci.org/buildbot/buildbot/jobs/128800563) seems to confirm that). I guess it doesn't matter much (since the step appears to only take 0.25s), but I'd rather see the test results earlier. Another option would be putting the step in `after_script`.

rutsky:
@tomprince looks like you right: if some steps are failing in `script` other steps still run.
I mistaken this case with the case when build fails before entering `script` section, i.e. in `install` or `before_install` --- in this case `script` steps are not run at all and build marked as broken (red exclamation mark).

I agree that it's better to put it in `after_script`.

PR #2229, opened 2016-05-12
Only build main-line branches.

tomprince:
There is no reason to build branches like [this](https://travis-ci.org/buildbot/buildbot/builds/129810705), since they also get built as a PRs ([here](https://travis-ci.org/buildbot/buildbot/builds/129810761)).


rutsky:
LGTM

PR #2230, opened 2016-05-16
[WIP] Parallel tests.

tomprince:
Travis workers have ~2 cores.


tomprince:
The improvements seem to be mostly noise.

tardyp:
that was also my understanding -j has overhead that seems to be compensated by using more than 2 cores.

PR #2236, opened 2016-05-29
simplify remoteGetWorkerInfo logic

rutsky:
retrieve "slave_commands" and "version" only when old buildslave is
detected.


djmitche:
So, for new workers the getWorkerInfo call includes all of this?

+1!

rutsky:
@djmitche yes, new buildbot-worker (and latest versions of buildslave) get all info in single `getWorkerInfo()` (`getSlaveInfo()`) call.
I haven't noticed that when previously changed this function, so simplifying logic now.

PR #2237, opened 2016-05-29
rename "slave_commands" key in worker info to "worker_commands"

rutsky:
This is expected backward incompatible change in master-worker protocol.
Users of beta versions of buildbot-worker will need to update worker at the same time as master.

### DIFF
--- a/master/docs/relnotes/index.rst
+++ b/master/docs/relnotes/index.rst
@@ -72,6 +72,8 @@ Deprecations, Removals, and Non-Compatible Changes
 
 * Master/worker protocol has been changed:
 
+  * ``slave_commands`` key in worker information was renamed to ``worker_commands``.
+
   * ``getSlaveInfo`` remote method was renamed to ``getWorkerInfo``.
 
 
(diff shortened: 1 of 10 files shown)


master/docs/relnotes/index.rst
```diff
@@ -72,6 +72,8 @@ Deprecations, Removals, and Non-Compatible Changes
 
 * Master/worker protocol has been changed:
 
+  * ``slave_commands`` key in worker information was renamed to ``worker_commands``.
```
tardyp: dont you want to put that separate?

rutsky: @tardyp what do you mean?

tardyp: usually, you create separate PR for relnotes, especially when there is a PR burst like this.

rutsky: Ah, yes. Let's merge relnotes changes as they already distributed in PRs, I'll resolve/rebase/update them when conflicts will arise.

PR #2231, opened 2016-05-17
test_process_buildstep: unit test to validate run catches exceptions

yetanotherion:
A unit test to ensure the exceptions raised by the run
method are correctly handled.

PR #2240, opened 2016-05-30
db: remove unwanted specialization

yetanotherion:
The previous version had messages related to
MySQL database only, in the code to handle
all connection pools: db/pool.py.

This commit puts that logic in the enginestrategy
instead.

MySQL database connection issues were handled
before checkouting connections from the pool.
However, similar issues were observed when executing queries,
as per the test scenarios related to a builder implementing
a ETL use case. The refactoring implemented above
permits to implement related retries when executing
queries too.

Signed-off-by: Ion Alberdi <ion.alberdi@intel.com>

### DIFF
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -45,6 +45,85 @@ class ReconnectingListener(object):
         self.retried = False
 
 
+class Strategy(object):
+    def set_up(self, u, engine):
+        pass
+
+    def should_retry(self, operational_error):
+        try:
+            text = operational_error.args[0]
+            return 'Lost connection' in text or 'database is locked' in text
+        except Exception:
+            return False
+
+
+class SqlLiteStrategy(Strategy):
+
+    def set_up(self, u, engine):
+        """Special setup for sqlite engines"""
+        def connect_listener_enable_fk(connection, record):
+            # fk must be enabled for all connections
+            if not getattr(engine, "fk_disabled", False):
+                return  # http://trac.buildbot.net/ticket/3490#ticket
+                # connection.execute('pragma foreign_keys=ON')
+
+        sa.event.listen(engine.pool, 'connect', connect_listener_enable_fk)
+        # try to enable WAL logging
+        if u.database:
+            def connect_listener(connection, record):
+                connection.execute("pragma checkpoint_fullfsync = off")
+
+            sa.event.listen(engine.pool, 'connect', connect_listener)
+
+            log.msg("setting database journal mode to 'wal'")
+            try:
+                engine.execute("pragma journal_mode = wal")
+            except Exception:
+                log.msg("failed to set journal mode - database may fail")
+
+
+class MySQLStrategy(Strategy):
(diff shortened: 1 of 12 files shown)


tardyp:
👍 on the principle
Looks like the tests are broken by this.

master/buildbot/db/enginestrategy.py
```diff
@@ -209,6 +238,13 @@ def check_sqlalchemy_version(self):
                 raise RuntimeError("SQLAlchemy version %s is not supported by "
                                    "SQLAlchemy-Migrate version %d.%d.%d" % (version, mvt[0], mvt[1], mvt[2]))
 
+    def get_proxy(self, drivername):
```
djmitche: I like this model, but in what sense is this a "proxy"?

master/buildbot/db/pool.py
```diff
@@ -170,37 +170,30 @@ def __thd(self, with_engine, callable, args, kwargs):
...
-                        # sleep (remember, we're in a thread..)
-                        time.sleep(backoff)
-                        backoff *= self.BACKOFF_MULT
-
-                        # and re-try
-                        continue
-                    else:
+                    elapsed = time.time() - start
+                    if elapsed > self.MAX_OPERATIONALERROR_TIME:
+                        log.err(e, 'Raising due to 1day delay on DB query retries')
```
djmitche: I'm assuming `self.MAX_OPERATIONALERROR_TIME` is one day here, but if that constant is changed then this message would be incorrect.  Maybe just reference the constant?

djmitche:
Conceptual :+1: from me too :)

master/buildbot/db/enginestrategy.py
yetanotherion: Indeed, after re-reading
https://en.wikipedia.org/wiki/Proxy_pattern
the method itself might be the proxy, but not the returned objects.
What about 
- renaming 'get_proxy' to 'get_strategy', and 
- removing 'Proxy' from the name of the returned objects' classes (ex: MySQLStrategyProxy to MySQLStrategy) ?

yetanotherion:
@djmitche @tardyp comments addressed & test fixed.
I have doubts regarding codecov results as in a previous PR that contained unit tests only,
codecov reported a decrease of coverage (https://github.com/buildbot/buildbot/pull/2231)

tardyp:
42% diff coverage? looks like a big reduction. Is that a false positive?

yetanotherion:
@tardyp I tend to think it's a false positive. Indeed there are 
- integration tests on sqlite, and 
- migration tests on mysql, 
(that made me add flushErrors methods on unit tests).

For this reason, the related class-es should not be in green:
https://codecov.io/gh/buildbot/buildbot/compare/f43636fa240090ed5dc8b6ffc003f47d49a96562...7844fcd1a0d2e529c5e244be150c446628691595

yetanotherion:
@tardyp @djmitche how do we go forward on this ?

tardyp:
codecov only is generated on the sqlite paths this is why we have bad coverage for that PR

yetanotherion:
ok, thx @tardyp

PR #2181, opened 2016-04-29
Upgrade ec2.py from boto2 -> boto3

aelsabbahy:
Tests remain as boto2 to ensure validity of the code against the older set
of mocks/expectations.

Add `session` parameter to ec2.py which allows the user to provide their
own boto3.Session object, this allows for more flexibility (ex. assumeRole)

PR #2164, opened 2016-04-25
Add more tests for ec2.py

aelsabbahy:
Increase test coverage in preparation for boto2 -> boto3 upgrade

PR #2247, opened 2016-06-02
Upgrade test_worker_ec2 from boto2 -> boto3

aelsabbahy:
This is the third and last change for the boto3 upgrade, it was broken up into three parts:
1) Increase boto2 test coverage to ensure boto3 changes are correct (#2164) 
2) Upgrade ec2.py implementation from boto2 to boto3 (#2181)
3) Upgrade the test suite to boto3 (this pr)

### DIFF
--- a/master/buildbot/test/unit/test_worker_ec2.py
+++ b/master/buildbot/test/unit/test_worker_ec2.py
@@ -24,23 +24,23 @@ from buildbot.worker_transition import DeprecatedWorkerNameWarning
 try:
     from moto import mock_ec2
     assert mock_ec2
-    import boto
-    assert boto
+    import boto3
+    assert boto3
 except ImportError:
-    boto = None
+    boto3 = None
     ec2 = None
 
 
-if boto is not None:
+if boto3 is not None:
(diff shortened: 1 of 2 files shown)


tardyp:
@djmitche I cannot respawn the appveyor build. So we cannot merge this patch

master/buildbot/test/unit/test_worker_ec2.py
```diff
@@ -40,7 +40,7 @@
 def skip_ec2(f):
     f.skip = "boto or moto is not installed"
```
rutsky: Should be `"boto3 or moto is not installed"`

aelsabbahy: Fixed

rutsky:
@tardyp regarding restarting of builds: I saw in other projects that they close/reopen PR to restart PR status checks. Though this is a quite awful hack, this may help with AppVeyor (AFAIK on Travis you can restart any failed build without a problem).

tardyp:
ah yes. tomprince did actually use that trick last time but I forgot about it

PR #2252, opened 2016-06-05
remove notion of removed source commands from worker

rutsky:
New worker will work only with new master, and new master doesn't use
worker source commands (and they don't work anyway).


tardyp:
ok for me 👍

PR #2254, opened 2016-06-06
use buildbot-worker instead buildbot-slave in epydoc API reference generation script

rutsky:
I even managed to build reference docs with patched Epydoc:
common/generate_buildbot_api_documentation.sh has good starting point to
get list of required patches, unfortunately these patches are no longer
available by those URLs.

Looks like Epydoc should be dropped in favor of other tools, like
pydoctor or pdoc.

### DIFF
--- a/apidocs/Makefile
+++ b/apidocs/Makefile
@@ -3,13 +3,13 @@
 all: reference.tgz
 
 cleanpyc:
-	cd .. && find master slave -name '*.pyc' -exec rm \{} \; || exit 1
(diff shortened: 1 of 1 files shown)


rutsky:
Looking at http://trac.buildbot.net/ticket/2087 and http://trac.buildbot.net/ticket/1839, maybe we should just remove this whole Epydoc stuff?

It doesn't work out of the box and looks like we are not building/publishing API reference anyway...

rutsky:
PR that removed apidocs (in case we decide to just remove it): #2255

rutsky:
Closing, since we decided to completely remove apidocs.

PR #2258, opened 2016-06-06
P4 password

seankelly:
Obfuscate the P4 password in the logs if the worker supports it.


rutsky:
Please add unit test for old worker that doesn't support obfuscation.
You emulate old worker by adding `worker_version` to `self.setupStep()`:

```python
self.setupStep(..., worker_version={'*': "2.15"})
```

rutsky:
LGTM.

tardyp:
thanks!

PR #2262, opened 2016-06-07
get rid of `public_html` directory

rutsky:
It is not used and not supported for some time.

Use of third-party web server suggested.
`public_html` directory in examples were replaces with
`/usr/share/nginx/www/` (so examples assume use of nginx).

Fixes http://trac.buildbot.net/ticket/3547

### DIFF
--- a/master/buildbot/scripts/create_master.py
+++ b/master/buildbot/scripts/create_master.py
@@ -77,18 +77,6 @@ def makeSampleConfig(config):
     os.chmod(target, 0o600)
 
 
-def makePublicHtml(config):
-    webdir = os.path.join(config['basedir'], "public_html")
-    if os.path.exists(webdir):
-        if not config['quiet']:
-            print("public_html/ already exists: not replacing")
-        return
-    else:
-        os.mkdir(webdir)
-    if not config['quiet']:
-        print("populating public_html/")
-
-
 @defer.inlineCallbacks
 def createDB(config, _noMonkey=False):
     # apply the db monkeypatches (and others - no harm)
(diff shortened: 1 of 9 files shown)


rutsky:
On the second thought, maybe add support for serving static files from master's basedir?

It's not required for current web ui to function, but this feature implementation looks pretty simple (or not?) and may significantly help users transitioning from 0.8 to 0.9.

tardyp:
I think it makes few sense to put the files directly in the master's directory.
The reasonable buildbot admin shall put those files in a separate backuped directory/partition.

I like the default /usr/share/nginx/www as this is the default no-worry nginx solution for a quick setup.

rutsky:
@tardyp so do you for or against removing of `public_html`?
I added arguments against removing to the ticket: http://trac.buildbot.net/ticket/3547
Please, share your thoughts too.

tardyp:
I aknowledge the trac's arguments, still I am for removing of public_html. Keeping public_html serving is buildbot encourages users to use inefficient solution to serve build artifacts. 

nginx is very easy to setup, and people can still point it to their master's public_html if they really want.

rutsky:
@tardyp I think having non-efficient but easy to use solution is better than nothing - it may sufficient for small installations.
Anyway, `public_html` is not working right now and this can be treated as one of incompatibilities of 0.9 + missing feature that can be implemented later (for example as plugin).

Lets discuss this at the meeting today.

rutsky:
We decided to remove `public_html` and reimplement it's functionality later (see http://trac.buildbot.net/ticket/3560).
So lets go with this PR.

PR #2263, opened 2016-06-07
Quickfixed poller change_hook broken by site API change

gracinet:
This PR is a work in progress:

- I didn't look at unit or integration tests yet, but I suppose they don't exist at all, otherwise the breaking change would have not been unnoticed
- I don't know if change hooks are even supposed to be able to grab the master object (thee code structure suggests they aren't but this one really needs it). If they are, then a more streamlined way of doing it would be preferable

I can confirm it solves http://trac.buildbot.net/ticket/3559

### DIFF
--- a/master/buildbot/www/hooks/poller.py
+++ b/master/buildbot/www/hooks/poller.py
@@ -18,9 +18,9 @@
 
 from buildbot.changes.base import PollingChangeSource
 
-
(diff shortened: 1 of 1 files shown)


tardyp:
looks like the tests need to be updated

tardyp:
closed in favor of #229

PR #2264, opened 2016-06-07
Allow independant EC2 price_multiplier or max_spot_price usage

### LINKED ISSUE #2898: convert all services to BuildbotService
This ticket is a migrated Trac ticket [3084](http://trac.buildbot.net/ticket/3084)

People contributed to the original ticket: @unknown_contributor, @djmitche, `benoit.allard@...`, @tardyp
Ticket created on: `Dec 01 2014`
Ticket last modified on: `Dec 05 2014`

---

We have a bunch of services in buildbot each of them are deriving from different base
* twisted.application.service.Service
* twisted.application.service.[[MultiService]]
* buildbot.util.service.[[AsyncService]]
* buildbot.util.service.[[AsyncMultiService]]
* buildbot.util.service.[[ReconfigurableServiceMixin]]
* buildbot.util.service.[[ClusteredService]]

Goal of this meta task is to convert all of our services to [[BuildbotService]] or [[ClusteredBuildbotService]].
This will make the code easier to understand and maintain, all our services will be
- async
- reconfigurable
- (multi) potentially a parent

The most tricky thing to do there is to move from reconfigService(new_config) which takes [[ConfigObject]] in argument, and thus configured service must know where it is configured, to a reconfigService(*args, **kwargs), which takes the constructor argument the service was configured.

Some pattern have been written in [[BuildbotService]], to handle the problem for user defined custom services, but this may not fit all the internal usecases (in buildbot everything is configured services)

Please don't resolve this bug directly, but create subtasks

---

_Comment from_: @Ben
_Date_: `Dec 01 2014`

Would this be a step toward remote reconfig of services ?

---
_Comment from_: @tardyp
_Date_: `Dec 01 2014`

>Would this be a step toward remote reconfig of services ?
This is not really the goal, but unifying the way of reconfiguring will indeed make thing less impossible.

The most complicated stuff for your request would be to send the python code for custom steps, and their dependancies.

If one does not use them, or synchronise them by other means, we could eventually think of serializing the config object, and send that through the wire.

---
_Comment from_: @Ben
_Date_: `Dec 01 2014`

I know this is not the goal, I was just wondering ...

I'm still having in mind my architecture where one single 'arbiter' / 'supervisor' / 'you-name-it' would have the config, and would care about dispatching it to masters. In this case, no need to serialise / duplicate the config, just compile it once, and dispatch it to the right instances. config are code, but once compiled, they're pretty much static, which make that whole 'throwing over the wire' easier, or am I missing something ?

---
_Comment from_: @unknown_contributor
_Date_: `Dec 01 2014`

In [changeset:"d567123cb53617d82cd22c40cfde6a4f4d3badd9"]:
```[[CommitTicketReference]] repository="" revision="d567123cb53617d82cd22c40cfde6a4f4d3badd9"
Merge pull request #1394 from tardyp/services

rename reconfigService() to reconfig[[ServiceWithBuildbotConfig]]

See ticket:3084
```

---
_Comment from_: @dustin
_Date_: `Dec 04 2014`

Ben, now that I think about it we could probably use some custom import hook magic to achieve that one-config thing you describe.  And I think it could be done in the configuration file itself, not in Buildbot.  We'd have to use some synchronous mechanism (not Twisted) to fetch Python code, but that's not hard - python-requests could probably do it.  Anyway, can you open up a bug?

---
_Comment from_: @Ben
_Date_: `Dec 05 2014`

Dustin, good idea, see #3093.

morrone:
Change the EC2LatentBuildSlave to allow the use of just one
of either max_spot_price or price_multiplier.  In other words,
either value can be set to None so that only the other value
is used.  When spot instances are used, at least one of those two
parameters must be set.

If only max_spot_price is set, that is the value that is used
as the bid price.  There is not need to look up the historical
bid prices in that case.  It seems likely that in the real world
most people will want to set the maximum price they are willing
to pay.  Artificially bidding lower than one is willing to pay
based on historical information is probably not a very common use
case.

Also, change the logic to use as the bid price the minumum of either
the multiple of the historical average or the max_spot_price.  The
previous behavior was to raise an exception if the multiple was in
excess of the max_spot_price.  That doesn't make much sense.  The
historical average tells us little about the price _now_, so it probably
never makes sense to abort early without at least making an attempt
to substantiate with the max_spot_price.

Fixes #2898 problems 1 and 3.

### DIFF
--- a/master/buildbot/test/unit/test_worker_ec2.py
+++ b/master/buildbot/test/unit/test_worker_ec2.py
@@ -501,6 +501,28 @@ class TestEC2LatentWorker(unittest.TestCase):
 
         self.assertRaises(ValueError, create_worker)
 
+    @mock_ec2
+    def test_fail_multiplier_and_max_are_none(self):
+        '''
+        price_multiplier and max_spot_price may not be None at the same time.
+        '''
+        c, r = self.botoSetup()
+        amis = list(r.images.all())
+
+        def create_worker():
+            ec2.EC2LatentWorker('bot1', 'sekrit', 'm1.large',
+                                identifier='publickey',
+                                secret_identifier='privatekey',
+                                keypair_name="latent_buildbot_worker",
+                                security_name='latent_buildbot_worker',
+                                ami=amis[0].id,
+                                region='us-west-1',
+                                spot_instance=True,
+                                price_multiplier=None,
+                                max_spot_price=None
+                                )
+        self.assertRaises(ValueError, create_worker)
+
 
 class TestEC2LatentWorkerDefaultKeyairSecurityGroup(unittest.TestCase):
     ec2_connection = None
--- a/master/buildbot/test/unit/test_worker_ec2.py
+++ b/master/buildbot/test/unit/test_worker_ec2.py
@@ -435,42 +435,10 @@ class TestEC2LatentWorker(unittest.TestCase):
             Filters=[{'Name': 'instance-state-name', 'Values': ['running']}])
         instances = list(instances)
         self.assertTrue(bs.spot_instance)
-        self.assertEqual(bs.retry, 1)
         self.assertEqual(bs.product_description, product_description)
         self.assertEqual(len(instances), 1)
         self.assertEqual(instances[0].id, instance_id)
         self.assertEqual(instances[0].tags, [])
-
-    @mock_ec2
-    def test_start_spot_instance_retry(self):
-        c, r = self.botoSetup('latent_buildbot_slave')
-        amis = list(r.images.all())
-        product_description = 'Linux/Unix'
-        retry = 3
-        with assertProducesWarnings(
-                DeprecatedWorkerNameWarning,
-                messages_patterns=[
-                    r"Use of default value of 'keypair_name' of "
-                    r"EC2LatentWorker constructor is deprecated",
-                    r"Use of default value of 'security_name' of "
-                    r"EC2LatentWorker constructor is deprecated"
-                ]):
-            bs = ec2.EC2LatentWorker('bot1', 'sekrit', 'm1.large',
-                                     identifier='publickey',
-                                     secret_identifier='privatekey',
-                                     ami=amis[0].id, retry=retry,
-                                     spot_instance=True, max_spot_price=1.5,
-                                     product_description=product_description
-                                     )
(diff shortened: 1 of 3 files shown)


morrone:
Does anyone have any suggestions to address the code coverage failures?  I tried adding some EC2 spot instance unit tests relevant to my changes, but the test system doesn't support spot instance testing.

rutsky:
What if you add test similar to `test_start_instance`, but with `spot_instance=True` flag in worker constructor?
If it fails we can try to look at the failing code and try to properly mock it.

morrone:
In one of my iterations of this patch, I had three more tests that all used spot_instance=True.  Unfortunately, the test system doesn't support spot instances right now.  The tests failed on retrieving the historical spot data.

In fact, if my reading is correct the existing spot instance tests, test_start_vpc_spot_instance(), test_start_spot_instance(), and test_start_spot_instance_retry() don't actually test spot instances.  They are calling EC2LatentWorker()'s internal method _start_instance() which starts a regular instance, not a spot instance.  Some time in the past, I think maybe that class method did handle both cases, but now it only starts non-spot instances.   Now _request_spot_instance() is the method that starts spot instances.

I tried writing my tests to use _request_spot_instance(), and it failed looking up the historical spot pricing data.  My suspicion is that resolving that testing problem is going to take a lot more effort than I have the time to apply to this problem.

morrone:
I just added pull request #2271 to illustrate the testing problem on the existing spot instance tests.  If that can be resolved, then I can add more tests to this pull request.

davebx:
> @morrone So unless I am missing some use case, return_price_adjustment seems to be based on a faulty understanding of AWS bid prices.

Your interpretation is probably correct, if my (admittedly imperfect) memory is serving me correctly. I wrote this code several years ago, but our use case for spot instances evaporated about 6 months after the code was accepted.

morrone:
@davebx, great, thanks for the feedback.  I updated my pull request with a commit to take the retries back out.

davebx:
@morrone Looks good, I'm +1 on this conceptually, I'll let the people with commit access decide if it's programmatically solid.

tardyp:
looks good. It looks that the codecov fails because the tests are not run inside travis, probably because they are skipped cause of dependencoes not installed in travis.
Could you look at that issue?

PR #2267, opened 2016-06-08
Providing Apache2 sample conf

gracinet:
I'm actually not sure about SSE: I think the exclusion in Nginx conf is specific to the way Nginx works, but I'm not sure. My Apache2 setups seem to work well. 

If some can tell me effortlessly how it'd look if SSE was broken,  that'd spare me installing a broken Nginx config just for the sake of comparison ? (I'd include relevant testing info in the PR itself)


tardyp:
on nginx, you need to disable buffering for sse to work in real time. I would expect similar for apache.

PR #2270, opened 2016-06-09
enginestrategy: retry on MySQL/InnoDB deadlock

yetanotherion:
Retry when observing the following Error

sqlalchemy.exc.OperationalError: (OperationalError)
(1213, 'Deadlock found when trying to get lock;
        try restarting transaction')

### DIFF
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -84,11 +84,18 @@ class SqlLiteStrategy(Strategy):
 
 class MySQLStrategy(Strategy):
     disconnect_error_codes = (2006, 2013, 2014, 2045, 2055)
+    deadlock_error_codes = (1213,)
+
+    def in_error_codes(self, args, error_codes):
+        if args:
+            return args[0] in error_codes
+        return False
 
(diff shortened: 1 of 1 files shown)


rutsky:
@yetanotherion cay you give more details when this error occurs?

yetanotherion:
It's an ETL (extract transform load) use case, on an old-style step.
It don't have more information to reproduce. Note however, that: 
- it seems coherent with the already existing (retry on deadlock error) strategy
https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/enginestrategy.py#L55, and that
- the innodb manual seems to support that strategy
http://dev.mysql.com/doc/refman/5.7/en/innodb-error-handling.html
"Both deadlocks and lock wait timeouts are normal on busy servers and it is necessary for applications to  be aware that they may happen and handle them by retrying"

yetanotherion:
I don't undertsand where the travis failure is, some tests became flaky ?

yetanotherion:
Don't know who did it, but thanks for the rebuild :)

tardyp:
looks good!

PR #2276, opened 2016-06-16
Fix recovery upon master unclean kill or crash (:bug:`3564`)

tardyp:


tardyp:

PR #2277, opened 2016-06-17
fix autoreload on reconnection

tardyp:
Angularjs 1.5 added this regression


tardyp:

PR #2278, opened 2016-06-17
fix sqlite maxconn to 1

tardyp:
PR #2270, shows lots of issues related to sqlite multiprocessing.

I think this not really related to this PR, but rather related to sqlite not working well with threads.

After investigating a couple of hours, I wasn't really able to fix the issue.

I added a bunch of connection close(), but those are not really sufficient (this removes some operational errors, but not all)

putting max_conn to 1 to sqlite as default to secure 0.9.0 release, and filling this bug
http://trac.buildbot.net/ticket/3565


tardyp:

PR #2280, opened 2016-06-21
Update flake8 to version 2.6.

tomprince:
This introduces a few new errors:

- `W503` line break before binary operator

   I don't think this is a sensible choice and the codebase doesn't follow this convention.
- `E731` do not assign a lambda expression, use a def

   This is used often in tests for functions that return canned values. I think turning them into `def`s obscures that.
- `E402` module level import not at top of file

  I've fixed most of the occurrences of this, as they are fairly trivial.

### DIFF
--- a/master/buildbot/steps/slave.py
+++ b/master/buildbot/steps/slave.py
@@ -17,12 +17,25 @@
 # It should never be imported by Buildbot.
 
 
-# pylint: disable=wildcard-import
-# pylint: disable=unused-wildcard-import
-from buildbot.steps.worker import *  # noqa
+from buildbot.steps.worker import CompositeStepMixin
+from buildbot.steps.worker import CopyDirectory
+from buildbot.steps.worker import FileExists
+from buildbot.steps.worker import MakeDirectory
+from buildbot.steps.worker import RemoveDirectory
+from buildbot.steps.worker import SetPropertiesFromEnv
+from buildbot.steps.worker import WorkerBuildStep
 # pylint: disable=unused-import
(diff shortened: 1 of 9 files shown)


master/setup.py
```diff
@@ -464,7 +464,6 @@ def define_plugin_entries(groups):
     setup_args['extras_require'] = {
         'test': [
             'setuptools_trial',
-            'pep8',
```
tomprince: `pep8` was renamed `pycodestyle`, but `flake8` depends on it already, so we don't need to depend on it here.

tardyp:
http://trac.buildbot.net/ticket/3566#comment:1

tardyp:
There is still one issue:

buildbot/steps/slave.py:32:43: F405 WorkerBuildStep may be undefined, or defined from star imports: buildbot.steps.worker
deprecatedWorkerModuleAttribute(locals(), WorkerBuildStep)
                                          ^
1       F405 WorkerBuildStep may be undefined, or defined from star imports: buildbot.steps.worker

master/buildbot/steps/slave.py
```diff
@@ -19,6 +19,7 @@
 
 # pylint: disable=wildcard-import
 # pylint: disable=unused-wildcard-import
+from buildbot.steps.worker import WorkerBuildStep
```
tardyp: I think you should add the noqa on that error line

switching the import will lead to import order error

tomprince: The trouble is that isort is changing the order of comments. I fixed it by just getting rid of the `*` import.

PR #2281, opened 2016-06-22
Added user toggle to schedulers.

gtmacdonald:
Hello!

I have added the ability for the user to toggle a scheduler.

In the schedulers UI there's now a new column with checkboxes to enable or disable a scheduler. Periodic schedulers will print a log message if a build is being skipped. Force schedulers will have their build button disabled and display a red warning banner reminding the user the scheduler is disabled.

To do this I've added an enabled column to the schedulers table in the database. I've added unit tests by mirroring what was done for the step.hidden attribute which was similarly added. I couldn't see where to add docs. Let me know if you need me to add some.

force_sched

scheduler

Please let me know if you need anything. :) Thx!

### DIFF
--- a/master/buildbot/data/forceschedulers.py
+++ b/master/buildbot/data/forceschedulers.py
@@ -32,7 +32,8 @@ def forceScheduler2Data(sched):
                button_name=text_type(sched.buttonName),
                label=text_type(sched.label),
                builder_names=[text_type(name)
-                              for name in sched.builderNames])
+                              for name in sched.builderNames],
+               enabled=sched.enabled)
     ret["all_fields"] = [field.getSpec() for field in sched.all_fields]
     return ret
 
@@ -105,5 +106,6 @@ class ForceScheduler(base.ResourceType):
         button_name = types.String()
         label = types.String()
         builder_names = types.List(of=types.Identifier(50))
+        enabled = types.Boolean()
         all_fields = types.List(of=types.JsonObject())
     entityType = EntityType(name)
--- a/master/buildbot/data/schedulers.py
+++ b/master/buildbot/data/schedulers.py
@@ -35,6 +35,7 @@ class Db2DataMixin(object):
         data = {
             'schedulerid': dbdict['id'],
             'name': dbdict['name'],
+            'enabled': dbdict['enabled'],
             'master': master,
         }
         defer.returnValue(data)
@@ -57,6 +58,14 @@ class SchedulerEndpoint(Db2DataMixin, base.Endpoint):
                 return
         defer.returnValue((yield self.db2data(dbdict))
                           if dbdict else None)
+
+    @defer.inlineCallbacks
+    def control(self, action, args, kwargs):
+        if action == 'enable':
+            schedulerid = kwargs['schedulerid']
+            v = args['enabled']
+            yield self.master.data.updates.schedulerEnable(schedulerid, v)
+        defer.returnValue(None)
 
 
 class SchedulersEndpoint(Db2DataMixin, base.Endpoint):
@@ -84,12 +93,28 @@ class Scheduler(base.ResourceType):
     plural = "schedulers"
     endpoints = [SchedulerEndpoint, SchedulersEndpoint]
     keyFields = ['schedulerid']
+    eventPathPatterns = """
+        /schedulers/:schedulerid
+    """
 
     class EntityType(types.Entity):
         schedulerid = types.Integer()
         name = types.String()
+        enabled = types.Boolean()
         master = types.NoneOk(masters.Master.entityType)
     entityType = EntityType(name)
+
+    @defer.inlineCallbacks
+    def generateEvent(self, schedulerid, event):
+        scheduler = yield self.master.data.get(('schedulers', str(schedulerid)))
+        self.produceEvent(scheduler, event)
+
+    @base.updateMethod
+    @defer.inlineCallbacks
+    def schedulerEnable(self, schedulerid, v):
+        yield self.master.db.schedulers.enable(schedulerid, v)
+        yield self.generateEvent(schedulerid, 'updated')
+        defer.returnValue(None)
 
     @base.updateMethod
     def findSchedulerId(self, name):
--- a/master/buildbot/db/schedulers.py
+++ b/master/buildbot/db/schedulers.py
@@ -33,6 +33,13 @@ class SchedulerAlreadyClaimedError(Exception):
 class SchedulersConnectorComponent(base.DBConnectorComponent):
     # Documentation is in developer/db.rst
 
+    def enable(self, schedulerid, v):
+        def thd(conn):
+            tbl = self.db.model.schedulers
+            q = tbl.update(whereclause=(tbl.c.id == schedulerid))
+            conn.execute(q, enabled=int(v))
+        return self.db.pool.do(thd)
+
     def classifyChanges(self, schedulerid, classifications):
         def thd(conn):
             tbl = self.db.model.scheduler_changes
@@ -184,11 +191,11 @@ class SchedulersConnectorComponent(base.DBConnectorComponent):
                 elif active is not None:
                     wc = (sch_mst_tbl.c.masterid == NULL)
 
-            q = sa.select([sch_tbl.c.id, sch_tbl.c.name,
+            q = sa.select([sch_tbl.c.id, sch_tbl.c.name, sch_tbl.c.enabled,
                            sch_mst_tbl.c.masterid],
                           from_obj=join, whereclause=wc)
 
-            return [dict(id=row.id, name=row.name,
+            return [dict(id=row.id, name=row.name, enabled=bool(row.enabled),
                          masterid=row.masterid)
                     for row in conn.execute(q).fetchall()]
         return self.db.pool.do(thd)
--- a/master/buildbot/schedulers/base.py
+++ b/master/buildbot/schedulers/base.py
@@ -90,23 +90,61 @@ class BaseScheduler(ClusteredBuildbotService, StateMixin):
 
         # internal variables
         self._change_consumer = None
+        self._enable_consumer = None
         self._change_consumption_lock = defer.DeferredLock()
+
+        self.enabled = True
 
     def reconfigService(self, *args, **kwargs):
         raise NotImplementedError()
 
     # activity handling
-
+    @defer.inlineCallbacks
     def activate(self):
-        return defer.succeed(None)
+        if not self.enabled:
+            yield defer.returnValue(None)
+            return
 
+        # even if we aren't called via _activityPoll(), at this point we
+        # need to ensure the service id is set correctly
+        if self.serviceid is None:
+            self.serviceid = yield self._getServiceId()
+            assert self.serviceid is not None
+
+        schedulerData = yield self._getScheduler(self.serviceid)
+
+        if schedulerData:
+            self.enabled = schedulerData['enabled']
+
+        if not self._enable_consumer:
+            yield self.startConsumingEnableEvents()
+
+    def _enabledCallback(self, key, msg):
+        if msg['enabled']:
+            self.enabled = True
+            d = self.activate()
+        else:
+            d = self.deactivate()
+
+            def fn(x):
+                self.enabled = False
+
+            d.addCallback(fn)
+        return d
+
+    @defer.inlineCallbacks
     def deactivate(self):
-        return defer.maybeDeferred(self._stopConsumingChanges)
+        if not self.enabled:
+            return
+        yield defer.maybeDeferred(self._stopConsumingChanges)
 
     # service handling
 
     def _getServiceId(self):
         return self.master.data.updates.findSchedulerId(self.name)
+
+    def _getScheduler(self, sid):
+        return self.master.db.schedulers.getScheduler(sid)
 
     def _claimService(self):
         return self.master.data.updates.trySetSchedulerMaster(self.serviceid,
@@ -138,6 +176,13 @@ class BaseScheduler(ClusteredBuildbotService, StateMixin):
             ('changes', None, 'new'))
 
     @defer.inlineCallbacks
+    def startConsumingEnableEvents(self):
+        assert not self._enable_consumer
+        self._enable_consumer = yield self.master.mq.startConsuming(
+            self._enabledCallback,
+            ('schedulers', str(self.serviceid), 'updated'))
+
+    @defer.inlineCallbacks
     def _changeCallback(self, key, msg, fileIsImportant, change_filter,
                         onlyImportant):
 
@@ -157,6 +202,7 @@ class BaseScheduler(ClusteredBuildbotService, StateMixin):
                     'not processed by scheduler %(name)s',
                     codebase=change.codebase, name=self.name)
             return
+
         if fileIsImportant:
             try:
                 important = fileIsImportant(change)
--- a/master/buildbot/schedulers/timed.py
+++ b/master/buildbot/schedulers/timed.py
@@ -83,6 +83,10 @@ class Timed(base.BaseScheduler, AbsoluteSourceStampsMixin):
     def activate(self):
         yield base.BaseScheduler.activate(self)
 
+        if not self.enabled:
+            yield defer.returnValue(None)
+            return
+
         # no need to lock this
         # nothing else can run before the service is started
         self.actuateOk = True
@@ -103,6 +107,10 @@ class Timed(base.BaseScheduler, AbsoluteSourceStampsMixin):
     @defer.inlineCallbacks
     def deactivate(self):
         yield base.BaseScheduler.deactivate(self)
+
+        if not self.enabled:
+            yield defer.returnValue(None)
+            return
 
         # shut down any pending actuation, and ensure that we wait for any
         # current actuation to complete by acquiring the lock.  This ensures
@@ -135,6 +143,12 @@ class Timed(base.BaseScheduler, AbsoluteSourceStampsMixin):
 
     @defer.inlineCallbacks
     def startBuild(self):
+        if not self.enabled:
+            log.msg(format='ignoring build from %(name)s because scheduler '
+                           'has been disabled by the user',
+                    name=self.name)
+            return
+
         # use the collected changes to start a build
         scheds = self.master.db.schedulers
         classifications = yield scheds.getChangeClassifications(self.objectid)
@@ -355,6 +369,10 @@ class NightlyTriggerable(NightlyBase):
     @defer.inlineCallbacks
     def activate(self):
         yield NightlyBase.activate(self)
+
+        if not self.enabled:
+            return
+
         lastTrigger = yield self.getState('lastTrigger', None)
         self._lastTrigger = None
         if lastTrigger:
@@ -412,6 +430,12 @@ class NightlyTriggerable(NightlyBase):
 
     @defer.inlineCallbacks
     def startBuild(self):
+        if not self.enabled:
+            log.msg(format='ignoring build from %(name)s because scheduler '
+                           'has been disabled by the user',
+                    name=self.name)
+            return
+
         if self._lastTrigger is None:
             return
 
--- a/master/buildbot/schedulers/trysched.py
+++ b/master/buildbot/schedulers/trysched.py
@@ -105,6 +105,9 @@ class Try_Jobdir(TryBase):
     def activate(self):
         yield TryBase.activate(self)
 
+        if not self.enabled:
+            return
+
         # set the watcher's basedir now that we have a master
         jobdir = os.path.join(self.master.basedir, self.jobdir)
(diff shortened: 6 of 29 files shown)


master/buildbot/data/forceschedulers.py
```diff
@@ -27,7 +27,8 @@ def forceScheduler2Data(sched):
                name=unicode(sched.name),
                button_name=unicode(sched.buttonName),
                label=unicode(sched.label),
-               builder_names=map(unicode, sched.builderNames))
+               builder_names=map(unicode, sched.builderNames),
```
sa2ajj: while you are at it, do you think you could convert this `map` to a list comprehension?  (that'd be a bit more readable)

master/buildbot/schedulers/base.py
```diff
@@ -86,12 +86,29 @@ def __init__(self, name, builderNames, properties=None,
...
+    # activity handling
+    @defer.inlineCallbacks
     def activate(self):
-        return defer.succeed(None)
+        # even if we aren't called via _activityPoll(), at this point we
+        # need to ensure the service id is set correctly
+        if self.serviceid is None:
+            self.serviceid = yield self._getServiceId()
+            assert self.serviceid is not None
+
```
sa2ajj: there should only be one empty line for separating logical blocks.

master/buildbot/schedulers/base.py
```diff
@@ -150,6 +177,13 @@ def _changeCallback(self, key, msg, fileIsImportant, change_filter,
                     'not processed by scheduler %(name)s',
                     codebase=change.codebase, name=self.name)
             return
+
+        if not self.enabled:
+            log.msg(format='ignoring change from %(name)s because ' \
```
sa2ajj: no need for `\` at the end of the line

master/buildbot/schedulers/timed.py
```diff
@@ -130,6 +130,12 @@ def gotChange(self, change, important):
 
     @defer.inlineCallbacks
     def startBuild(self):
+        if not self.enabled:
+            log.msg(format='ignoring build from %(name)s because scheduler' \
```
sa2ajj: ditto

master/buildbot/schedulers/timed.py
```diff
@@ -407,6 +413,12 @@ def trigger(self, waited_for, sourcestamps=None, set_props=None,
 
     @defer.inlineCallbacks
     def startBuild(self):
+        if not self.enabled:
+            log.msg(format='ignoring build from %(name)s because scheduler ' \
+                           'has been disabled by the user',
```
sa2ajj: ditto

sa2ajj:
LGTM otherwise

(I wish GitHub would allow just one review "item" with all messages inside)

gtmacdonald:
Great! I'll make those changes and resubmit.

gtmacdonald:
Aw so close. Looks like postgres wants a tiny int instead of a boolean for some reason.

gtmacdonald:
Coverage doesn't look so good but I think it's being pulled down by the worker code, which isn't related AFAIK.

.travis.yml
```diff
@@ -126,7 +126,7 @@ install:
 
   # Note pylint version is pinned because newer versions can't import zope.interface - http://www.logilab.org/92792
   # astroid > 1.4 is incompatible with pylint 1.1.0
-  - "[ $TESTS != lint ] || pip install pylint==1.1.0 pep8==1.5.7 flake8  astroid==1.3.8"
+  - "[ $TESTS != lint ] || pip install pylint==1.1.0 pep8==1.5.7 flake8==2.5.5  astroid==1.3.8"
```
tardyp: this will clash with https://github.com/buildbot/buildbot/pull/2280

tardyp:
I like the idea of dynamic enable/disable of stuff.

I think this would make sense also for workers. To disable them instead of disconnecting (will allow to implement autorestart on the worker side)


I am a bit annoyed to merge this big change before rc1 though.
Would you care if I merge it for 0.9.1?

gtmacdonald:
Hi Pierre! I'm glad it'll be useful for other things. I wasn't sure if it fit within the bigger picture or not. I don't mind waiting. :) Thx!

tardyp:
rc1 has been released for long, and I forgot this one.
Please rebase, and I will merge it.

gtmacdonald:
Okie dokie.

gtmacdonald:
Finished rebase.

master/buildbot/data/schedulers.py
```diff
@@ -55,6 +56,14 @@ def get(self, resultSpec, kwargs):
         defer.returnValue((yield self.db2data(dbdict))
                           if dbdict else None)
 
+    @defer.inlineCallbacks
+    def control(self, action, args, kwargs):
+        if action == 'setenabled':
+            schedulerid = kwargs['schedulerid']
+            enabledValue = args['enabledValue']
+
+            yield self.master.data.updates.schedulerSetEnabled(schedulerid, enabledValue)
```
tardyp: this requires a unit test

master/buildbot/db/schedulers.py
```diff
@@ -28,6 +28,13 @@ class SchedulerAlreadyClaimedError(Exception):
 class SchedulersConnectorComponent(base.DBConnectorComponent):
     # Documentation is in developer/db.rst
 
+    def setEnabled(self, schedulerid, enabled_value):
+        def thd(conn):
+            tbl = self.db.model.schedulers
+            q = tbl.update(whereclause=(tbl.c.id == schedulerid))
+            conn.execute(q, enabled=int(enabled_value))
+        return self.db.pool.do(thd)
```
tardyp: unit test please

master/buildbot/data/schedulers.py
```diff
@@ -81,13 +90,29 @@ class Scheduler(base.ResourceType):
...
+    def generateEvent(self, schedulerid, event):
+        scheduler = yield self.master.data.get(('schedulers', str(schedulerid)))
+        self.produceEvent(scheduler, event)
+
+    @base.updateMethod
+    @defer.inlineCallbacks
+    def schedulerSetEnabled(self, schedulerid, enabled_value):
+        yield self.master.db.schedulers.setEnabled(
+            schedulerid=schedulerid, enabled_value=enabled_value)
+        yield self.generateEvent(schedulerid, 'updated')
```
tardyp: unit tests

tardyp:
please add some unit tests for the added python code, and also a line in the release note is required for such change

tardyp:
@gtmacdonald do you intend to finish this PR?

gtmacdonald:
I do, we'll need this feature. I've just been swamped with school and work. And it hasn't been a priority because we're only partially converted over to buildbot. Sorry for the delay.

gtmacdonald:
@tardyp Hi Pierre, Sorry for the delay. I've rebased with unit tests and release notes. Please let me know if there's anything else to do.

gtmacdonald:
@tardyp I've fixed the extra space lint issue. It doesn't seem to be rebuilding for some reason. There's also something about an unclean reactor and workers that I'm not sure is relevant.

master/buildbot/test/unit/test_db_schedulers.py
```diff
@@ -44,11 +44,53 @@ class Tests(interfaces.InterfaceTests):
...
+            name='schname',
+            enabled=True,
+            masterid=13))
+
     def test_signature_classifyChanges(self):
         @self.assertArgSpecMatches(self.db.schedulers.classifyChanges)
         def classifyChanges(self, schedulerid, classifications):
             pass
 
+    # def setEnabled(self, schedulerid, enabled_value):
```
tardyp: I think this is unintendended left overs

master/buildbot/data/schedulers.py
```diff
@@ -81,13 +90,30 @@ class Scheduler(base.ResourceType):
...
     entityType = EntityType(name)
 
+    @defer.inlineCallbacks
+    def generateEvent(self, schedulerid, event):
+        scheduler = yield self.master.data.get(('schedulers', str(schedulerid)))
+        self.produceEvent(scheduler, event)
+
+    @base.updateMethod
+    @defer.inlineCallbacks
+    def schedulerSetEnabled(self, schedulerid, enabled_value):
```
tardyp: I would rather call this API 

    def schedulerEnable(self, schedulerid, enable=True)

master/buildbot/data/schedulers.py
```diff
@@ -55,6 +56,14 @@ def get(self, resultSpec, kwargs):
         defer.returnValue((yield self.db2data(dbdict))
                           if dbdict else None)
 
+    @defer.inlineCallbacks
+    def control(self, action, args, kwargs):
+        if action == 'setenabled':
```
tardyp: I am not very fond of the setenabled name.

I think this would be better to change 'setenabled' -> 'enable' and 'enabledValue' to 'enable'

gtmacdonald:
I finally got around to finishing this. Please merge. Thx!

master/buildbot/schedulers/base.py
```diff
@@ -87,12 +87,29 @@ def __init__(self, name, builderNames, properties=None,
...
+        if self.serviceid is None:
+            self.serviceid = yield self._getServiceId()
+            assert self.serviceid is not None
+
+        schedulerData = yield self._getScheduler(self.serviceid)
+
+        if schedulerData:
+            self.enabled = schedulerData['enabled']
+
+        yield self.startConsumingEnabledChanges()
```
tardyp: changes is a thing in builbot its about changes to SCM. here we are watching events.

master/buildbot/schedulers/base.py
```diff
@@ -151,6 +178,13 @@ def _changeCallback(self, key, msg, fileIsImportant, change_filter,
                     'not processed by scheduler %(name)s',
                     codebase=change.codebase, name=self.name)
             return
+
+        if not self.enabled:
```
tardyp: Shouldn't we rather stop the scheduler?
When it is restarted again it will consume changes that were left rather than just forget them

www/base/src/app/schedulers/schedulers.controller.coffee
```diff
@@ -2,3 +2,23 @@ class schedulers extends Controller
...
+        # Assumes an order.
+        $scope.schedulers.onNew = (s) ->
+            $scope.enabledValues.push(s.enabled)
+
+        $scope.change = ->
+            i = 0
+            while i<$scope.enabledValues.length
+                oldValue = $scope.enabledValues[i]
+                scheduler = $scope.schedulers[i]
+                newValue = scheduler.enabled
```
tardyp: looks too complicated.
why dont you pass the scheduler id in the change callback?

www/base/src/app/schedulers/schedulers.tpl.jade
```diff
@@ -3,16 +3,22 @@
       uib-tab(heading="All schedulers")
         table.table.table-hover.table-striped.table-condensed
           tr
+              td Enabled
               td Scheduler Name
               td Master
           tr(ng-repeat='scheduler in schedulers | orderBy: ["name"]')
+              td
+                input(type="checkbox", ng-model="scheduler.enabled", ng-change="change()")
```
tardyp: ng-change="change(scheduler)"

so that you get the schedulerid and the new value

master/buildbot/schedulers/base.py
gtmacdonald: I'm not sure I understand. I'm following the same logic to ignore changes as the isFileImportant callback. So it will ignore the changes when the scheduler isn't on. Are you suggesting it SHOULD remember? If so then should I disable the change consumer on the enable toggle?

gtmacdonald:
@tardyp I've finished the change requests except for your question about stopping the scheduler. I'll do some more testing to see if the change requests are backing up.

gtmacdonald:
@tardyp I think I understand now what you're asking for about stopping the scheduler. The current implementation suits my purposes with the timed scheduler. Please let me know if you'd like me to change it.

tardyp:
@gtmacdonald I think indeed, it makes much more sense to just temporarily stop the scheduler so that it can keep up later.

For me another use case for the scheduler disabling is if you have some issue with the infra and do not want more spurious failed builds while repairing it (that might also be your usecase, I am not sure)
When you enable it back you would expect to have buildbot keep up with the changes that were recorded in between. For TimedScheduler, this will be equivalent as your current implementation

Usually people would just stop their master, but then this also stop the change source, and you will miss builds.

gtmacdonald:
That makes sense. In our case we sometimes need to shut off a long running build to switch to a quicker incremental one; otherwise, for legacy reasons, they'll end up interfering.

So will I need to add a queue for the missed changes in the scheduler? Or is there a method call to stop the scheduler but keep it registered for changes?

tardyp:
You dont need to do anything, the scheduler is already keeping the id of the last change it processed. At restart it will start from there. I think you can just call deactivate, and in activate make sure the scheduler is not disabled.

gtmacdonald:
I think it's good 2 go. Please let me know if there's anything else.

tardyp:
this needs some rebase :-}

www/base/src/app/schedulers/schedulers.controller.coffee
```diff
@@ -2,3 +2,10 @@ class schedulers extends Controller
     constructor: ($log, $scope, $location, dataService) ->
         data = dataService.open().closeOnDestroy($scope)
         $scope.schedulers = data.getSchedulers()
+
+        $scope.change = (s) ->
+        	newValue = s.enabled
+        	param = enabled: newValue
+        	res = dataService.control 'schedulers', s.schedulerid, 'enable', param
+        	res.then (-> $log.debug("succeeded")), (data) => $log.error(data.error)
```
tardyp: no need for logging

gtmacdonald:
how's it look now? BTW, do I need to comment here to alert you that it's ready? Or do you get an automated message?

gtmacdonald:
This should be g2g now. :)

master/buildbot/test/unit/test_schedulers_base.py
```diff
@@ -158,9 +158,10 @@ def gotChange(got_change, got_important):
...
 
         def test(_):
-            # check that it registered a callback
-            self.assertEqual(len(self.mq.qrefs), 1)
-            qref = self.mq.qrefs[0]
+            # check that it registered callbacks
+            self.assertEqual(len(self.mq.qrefs), 2)
+
+            qref = self.mq.qrefs[1]
             self.assertEqual(qref.filter, ('changes', None, 'new'))
```
tardyp: There is a need to test that the scheduler correctly react to the disable events

master/buildbot/schedulers/trysched.py
```diff
@@ -117,6 +120,9 @@ def activate(self):
 
     @defer.inlineCallbacks
     def deactivate(self):
+        if not self.enabled:
```
tardyp: base scheduler calls this method when a disable message is received, so the disable message will not actually disable.

Actually, this might work because self.enabled is set after calling the method.
..but the calling does not wait for results, and inlineCallback does not guarantee that the decorated method is called synchronously.

A unit test should test that to make sure it works for every version of twisted/python.

master/buildbot/schedulers/base.py
```diff
@@ -90,15 +90,38 @@ def __init__(self, name, builderNames, properties=None,
...
+
+        if not self._enable_consumer:
+            yield self.startConsumingEnableEvents()
+
+    def _enabledCallback(self, key, msg):
+        if msg['enabled']:
+            self.enabled = True
+            self.activate()
+        else:
+            self.deactivate()
```
tardyp: should wait for result of deactivate before actually disable

master/buildbot/schedulers/trysched.py
gtmacdonald: I think waiting for the deactivate as you mention below should fix this. I just don't want the enabled flag to be out of sync with reality.

master/buildbot/test/unit/test_schedulers_base.py
gtmacdonald: So if I defer the enabled state like we're planning, I'm not sure how to wait for the enabled state to change in the test. This is what I was planning on doing to defer the enabled state:

    def _enabledCallback(self, key, msg):
        if msg['enabled']:
            self.enabled = True
            self.activate()
        else:
            d = self.deactivate()
            def fn():
                self.enabled = False
            d.addCallback(fn)

Any suggestions? I'm not quite comfortable with this reactor pattern yet.

tardyp: You can call the callback directly from the tests instead of using the mq api to send the message. you need to make your test inlineCallbacks and then wait for the results of the callback in the test and make the appropriate assertion after that.
There are plenty of examples in the current tests

gtmacdonald: gotcha, thanks!

master/buildbot/schedulers/trysched.py
gtmacdonald: How should I go about testing each version of twisted/python? Is there an example of this somewhere?

tardyp: We have a CI so that your unit tests are run against all version of twisted/python.

So you just have to write the unit tests, and then they will run everywhere when you push your branch

gtmacdonald: oh that's great!

gtmacdonald:
So I think I've added the last unit test. One of the checks isn't passing but it seems unrelated.

tardyp:
@gtmacdonald please install codecov.io chrome extension, or use the website to checkout the coverage. There are still too many corner case which are not covered by the unit tests.

gtmacdonald:
@tardyp Hi Pierre, When I try to see the coverage using the chrome extension I get an error. It says something like, "there was a problem with the coverage, sorry." I tried running it on my own branch but I can't figure out how to generate the .coverage files.

gtmacdonald:
Doing a fresh rebase and push to see if I can get the code coverage tool to work.

gtmacdonald:
I like that codecov tool.

Should be g2g now. :)

tardyp:
One of the longest to land PR. congrats! thanks for keeping up!

gtmacdonald:
lol, thanks!

tardyp:
@gtmacdonald I got a report from use for a security problem with this PR.

Any anonymous user can disable a scheduler with this.
We need to add an endpointmatcher for this action, and make sure the doc is updated to explain how to require administrator login to disable schedulers

gtmacdonald:
Ok, I'll start looking into it.

tardyp:
@gtmacdonald I did implement this https://github.com/buildbot/buildbot/pull/2822

I would like you to test it with your environment, and make sure it works beyond the unit tests

gtmacdonald:
Sure thing. Thanks for implementing it. :)

gtmacdonald:
This looks like it works well with one root user and the following config

c['www'] = dict(port=8010,                                                                             
                plugins=dict(waterfall_view={}, console_view={}))                                      
                                                                                                       
authz = util.Authz(                                                                                    
    allowRules=[util.AnyEndpointMatcher(role="admins")],                                               
    roleMatchers=[util.RolesFromUsername(roles=['admins'], usernames=['root'])]                        
)                                                                                                      
                                                                                                       
auth=util.UserPasswordAuth({'root': 'pass'})                                                           
                                                                                                       
c['www']['auth'] = auth                                                                                
c['www']['authz'] = authz 

If i switch AnyEndpointMatcher to EnabledSchedulerEndpointMatcher. It still works but when I log out and switch back to anonymous, the UI is still there. The check boxes are ineffectual until I log back in as root though. Aside from this UI glitch it seems to be working.

gtmacdonald:
I'll try with a regular non-admin user shortly.

gtmacdonald:
Ok, if I add another user instead of just root everything works as expected. The UI clears immediately after I logout.

gtmacdonald:
so this fixed it:

auth=util.UserPasswordAuth({'root': 'pass', 'greg': 'pass'})

gtmacdonald:
It'd be nice if there was authentication that plugged into Microsoft's user management stuff.

gtmacdonald:
Idk, do you think we can ship it as is or should we fix this bug. It's kind of small. Might confuse build admins though.

tardyp:
If you can come up with a fix on the UI which displays an error message if
user is not allowed to change the scheduler (and put the state of the
checkbox back), that'd be nice.


Le lun. 6 mars 2017 à 22:25, Greg MacDonald <notifications@github.com> a
écrit :

gtmacdonald:
I think the proper fix is for it to remove those UI elements after it logs out of root. Like it does when there's more than just the root user. So the logic is there it's just not happening on an edge case for some reason. I'll "root" around for it tomorrow. :D ha ha i'm so funny.

PR #2283, opened 2016-06-23
Upgrade from eight is not supported

tardyp:
As per discussion in the weekly meeting, we will not support upgrading from eight to nine.

This PR removes all the migration scripts before the release of the first beta of buildbot nine.


master/buildbot/db/model.py
```diff
@@ -32,6 +32,15 @@
     ControlledSchema = None
 
 
+class EightUpgradeError(Exception):
+    def __init__(self):
+        message = """You are trying to upgrade a buildbot 0.8.x master to buildbot 0.9.x
+        This is not supported. Please start from a clean database
+        http://docs.buildbot.net/latest/manual/installation/nine-upgrade.htm"""
```
rutsky: Should be with `html` at the end: http://docs.buildbot.net/latest/manual/installation/nine-upgrade.html

PR #2286, opened 2016-06-24
put the JS test earlier to optimize overall time

tardyp:
with current way, travis tests finish with a single job slowly building the JS
better start the JS job earlier to make sure we use parallelism as much as possible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,9 @@ env:
   - TWISTED=latest SQLALCHEMY=latest TESTS=docs
   - TWISTED=latest SQLALCHEMY=latest TESTS=coverage
 
+  # add js tests in separate job. Start it early because it is quite long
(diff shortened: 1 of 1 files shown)


tomprince:
This looks reasonable.

.travis.yml
```diff
@@ -146,7 +146,9 @@ before_script:
 
 # Tests running commands
 script:
-  - "[ $TESTS != js ] || make frontend_install_tests"
+  # make frontend_install_tests takes 17 min, so we only do it post submit
+  - "[ $TESTS = js -a ${TRAVIS_PULL_REQUEST} != false ] && make frontend || false"
```
tomprince: This should use the `[ ... ] || ...` style of the other commands. Otherwise, the command just always fails.

PR #2287, opened 2016-06-24
Workerinfo

tardyp:


master/buildbot/worker/protocols/pb.py
```diff
@@ -191,47 +191,58 @@ def remotePrint(self, message):
...
+        defer.returnValue(info)
 
+    @defer.inlineCallbacks
+    def remoteGetWorkerInfo(self):
+        # we need to first try the legacy command in order
+        # to avoid legacy workers to output a stack trace in the logs
+        info = yield self.remoteGetWorkerInfoLegacy()
+        if info is not None:
+            # instead of logging, we mark the version as deprecated in the workerinfo
+            info["version"] = "deprecated:" + info.get("version", "unknown version")
```
rutsky: Wouldn't change of version for older workers break something?
I would expect that some users may actually rely on the fact that slave version is the same as version of installed `buildbot-slave` package.

And what is the benefit of having changed version string?
If we want to warn users in the UI I would expect to have added check and custom renderer on the UI side.

rutsky:
Just for the record, this PR fixes http://trac.buildbot.net/ticket/3567

tomprince:
I don't think we should pessimize the up-to-date case just to avoid an exception with an out-of-date slave. Note that we already have cases where we log similar exceptions on the slave (for example, the old and new ways of shutting-down slaves first tries the new one, and if there is an error, tries the older one.

We could, in the upcoming 0.8.x release, explicitly handle the call to log a message about use with a newer master, but otherwise I don't think we should worry about this.

master/buildbot/worker/protocols/pb.py
tardyp: indeed, probably it is a better idea to have info['deprecated']  = True

tardyp:
Well, as this is not the datapath. I would think it is not big problem to have one more round trip for detection purpose. we can leave that detection when we remove support for deprecated say 0.9.0.

new 0.9.0 will probably have lots of exceptions logs to deal with. I would prefer to avoid those spurious one.
I wasn't ware of the same situation for other stuff (I actually saw the startup exception in my docker experiments).
I would have expected that the master remembers the available commands and use that in order to run the proper command

tardyp:
We could also make a configuration hint, or better, if the  user use the class Buildslave() in the master.cfg, then we try old method first.

tomprince:
> I wasn't ware of the same situation for other stuff (I actually saw the startup exception in my docker experiments).

I'm not entirely surprised. This was a change introduced in 0.8.3.

--

A simpler solution could just to be to *also* log an appropriate message on the slave, rather than try to avoid logging a traceback.

tardyp:
Le dim. 26 juin 2016 à 23:32, Tom Prince <notifications@github.com> a
écrit :

> I wasn't ware of the same situation for other stuff (I actually saw the
> startup exception in my docker experiments).
>
> I'm not entirely surprised. This was a change introduced in 0.8.3.
>
> A simpler solution could just to be to *also* log an appropriate message
> on the slave, rather than try to avoid logging a traceback.
>
The slave code is released already. We cannot change it. We can only touch
the new worker

tomprince:
> The slave code is released already. We cannot change it.

I found a way around that. See #2294.

PR #2288, opened 2016-06-24
documentation fixes for cls-forcescheduler

tardyp:


rutsky:
This PR should fix http://trac.buildbot.net/ticket/3477

@tardyp there is a warning at the beginning if the file with reference to 3477 that should be removed (since I believe you fix that issue with this PR).

Also there is not-updated part about Jinja templates in this file:

```
    .. py:attribute:: type

           A list of types that the parameter conforms to.
           These are used by the jinja template to create appropriate html form widget.
           The available values are visible in ``master/buildbot/status/web/template/forms.html`` in the ``force_build_one_scheduler`` macro.
```

Not sure about other parts of this file.

P.S. It should be noted that Github preview of out RST files with `.. py:module::`/`.. py:class::` hides most parts of the rst file.

tardyp:
@rutsky thanks for the second read. I indeed missed those :-}

hopefully this is better now.

master/docs/developer/cls-forcesched.rst
```diff
@@ -71,9 +67,15 @@ This section contains information to help users or developers who are interested
 
     .. py:attribute:: type
 
-           A list of types that the parameter conforms to.
-           These are used by the jinja template to create appropriate html form widget.
-           The available values are visible in ``master/buildbot/status/web/template/forms.html`` in the ``force_build_one_scheduler`` macro.
+           A string identifying the type that the parameter conforms to.
+           It is used by the angular application to find which angular directive to use for showing the form widget.
+           The available values are visible in ``www/base/src/app/common/directives/forcefields/forcefields.directive.coffee``.
```
rutsky: I think link on the source file will be better:

```
The available values are visible in :src:`www/base/src/app/common/directives/forcefields/forcefields.directive.coffee`.
```

Previously I removed link because `forms.html` was not present in the source tree and Sphinx' linkcheck warned about it.

rutsky:
@tardyp besides linking source file via `:src:`, this LGTM, thanks!

tardyp:
I wasn't aware of :src: . will do.

PR #2289, opened 2016-06-24
Remove buildbot-slave.

tomprince:
There have been no substantive changes to the slave in 0.9.x, and
buildbot-worker provides equivalent functionality, so get rid of the
duplication.


tardyp:
I am not sure of why it has been kept, but I am sure @rutsky will have more insight

circleci.yml is to be fixed

tomprince:
> I am not sure of why it has been kept [...]

My assumption was so that it was still available during the transition.

rutsky:
While `buildbot-slave` usage will be discouraged after stable release of `buildbot-worker` I assumed that `buildbot-slave` will live for some time.

1. We need to have working `buildbot-slave` to test that changes in master/worker protocol are backward compatible with old worker (some of these test in WIP branch #2136 ). If we will have "final" release of `buildbot-slave` we could use that version from PyPI for testing, instead of using version from git.

2. `buildbot-worker` has backward-incompatible changes (some deprecated stuff was dropped, protocol changes, etc), so some users may not like upgrading their `buildbot-slave` to `buildbot-worker`s, but instead would like to have some bugfixes or security fixes for `buildbot-slave` instead. I doubt that there would be any fixes of `buildbot-slave`, and upgrading is quite straightforward, but still this argument exists.

Even if we would like to remove `buildbot-slave` from master branch (it may be confusing to have `worker` and `slave` in single repo for newcomers), I think we need to have it somewhere in maintainable state (in separate branch or separate repo).

tomprince:
> We need to have working buildbot-slave to test that changes in master/worker protocol are backward compatible with old worker (some of these test in WIP branch #2136 ). If we will have "final" release of buildbot-slave we could use that version from PyPI for testing, instead of using version from git.

We could just use the 0.8.x version from PyPI for testing. That seems like it would be a more realistic test of backwards compatability anyway.

> so some users may not like upgrading their buildbot-slave to buildbot-workers, but instead would like to have some bugfixes or security fixes for buildbot-slave instead. I doubt that there would be any fixes of buildbot-slave, and upgrading is quite straightforward, but still this argument exists.

I don't think it is worth the maintenance effort to keep this code around. 

> I think we need to have it somewhere in maintainable state (in separate branch or separate repo).

We can always make releases of the slave from the `eight` branch if there are security or other issues. The only people that couldn't reasonably update to `buildbot-worker` are those that are using a 0.8.x master anyway.

tardyp:
I agree with @tomprince 

buildbot-slave package has his home branch in eight.

.travis.yml
```diff
@@ -115,8 +108,7 @@ install:
   # for python 3 we dont install everything yet..
   - |
       [ $TESTS != py3 ] || \
-      pip install -e slave \
```
tardyp: so here, I'd guess we would need pip install buildbot-slave in order to not skip the compat tests

.travis.yml
```diff
@@ -95,18 +92,14 @@ install:
       [ $TESTS != trial -a $TESTS != coverage -a $TESTS != lint -a $TESTS != js ] || \
       pip install -e pkg \
                   -e master[tls,test] \
```
tardyp: or.. we add buildbot-slave in the test dependencies

.travis.yml
tomprince: As far as I can tell, there aren't currently any compatibility tests. We can deal with this if/when we add some.

rutsky: There is no such tests in master branch indeed. I have compatibility tests in #2136 , but I can add installation of `buidlbot-slave` in Travis there, when I will finish it.

rutsky:
I explained known to me issues regarding removal of `slave` from master and agree that this can be done.

PR #2290, opened 2016-06-24
fix change hook integration

tardyp:
This overrides #2263 
We keep the same api as before. Maybe other people have dependencies on request.site.master being valid.

This is untested. @gracinet could you test it on your env?
This would deserve an integration test, but that would be a bit complicated

### DIFF
--- a/master/buildbot/www/service.py
+++ b/master/buildbot/www/service.py
@@ -216,6 +216,8 @@ class WWWService(service.ReconfigurableServiceMixin, service.AsyncMultiService):
 
         self.site.sessionFactory = BuildbotSession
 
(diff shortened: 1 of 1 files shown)


gracinet:
@tardyp, well this is totally in line with the suggestion I had to "provide a proper API to get a handle on the master"
:+1: 

I'll give it a try, but that's about the last thing I'll be able to do before going on leave tomorrow until mi-July.

gracinet:
Yes, it works (had to cherry-pick it, because of too much delta on the production setup to directly use your branch)

tardyp:
I really appreciate you took the time to test it before going vacations!

have fun and take care!

PR #2292, opened 2016-06-24
Warn about ignored buildHorizon and logHorizion.

tomprince:
These aren't currently used, but they could be supported in the database. Either we should get rid of them, or open a ticket to add back support (and probably document that they aren't currently working).

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -823,6 +823,9 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
             s.checkConfig(self.status)
 
     def check_horizons(self):
+        if self.logHorizon is not None or self.buildHorizon is not None:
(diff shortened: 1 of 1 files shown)


tardyp:
Indeed, for me this is an important feature for 0.9.1

I would rather keep it like this, with a compatibility warning

tardyp:
http://trac.buildbot.net/ticket/3572#ticket

tomprince:
I've update the patch to warn that the configuration is currently ignored, pointing at the ticket.

tardyp:
> I've update the patch to warn that the configuration is currently ignored, pointing at the ticket.

I could swear I saw in the deleted lines of your first version a similar warning.

tomprince:
The original version said they were ignored due to being deprecated. The new one due to a missing feature.

PR #2293, opened 2016-06-24
Remove the trace of status for generating and reading status pickles.

tomprince:


tomprince:
The missing coverage is in deprecated status code. It could probably be deleted too, but I want to do things in stages.

tardyp:
looks good!

PR #2294, opened 2016-06-26
Log a message on the worker when buildbot-slave is detected.

tomprince:
Rather than try to get rid of the exception, just print a log message explaining it.

Supersedes #2287.

```
2016-06-26T15:27:15-0600 [Broker,client] Peer will receive following PB traceback:
2016-06-26T15:27:15-0600 [Broker,client] Unhandled Error
	  File ".../lib/python2.7/site-packages/twisted/spread/banana.py", line 167, in gotItem
	    self.callExpressionReceived(item)
	  File ".../lib/python2.7/site-packages/twisted/spread/banana.py", line 130, in callExpressionReceived
	    self.expressionReceived(obj)
	  File ".../lib/python2.7/site-packages/twisted/spread/pb.py", line 563, in expressionReceived
	    method(*sexp[1:])
	  File ".../lib/python2.7/site-packages/twisted/spread/pb.py", line 875, in proto_message
	    self._recvMessage(self.localObjectForID, requestID, objectID, message, answerRequired, netArgs, netKw)
	  File ".../lib/python2.7/site-packages/twisted/spread/pb.py", line 889, in _recvMessage
	    netResult = object.remoteMessageReceived(self, message, netArgs, netKw)
	  File ".../lib/python2.7/site-packages/twisted/spread/flavors.py", line 112, in remoteMessageReceived
	    raise NoSuchMethod("No such method: remote_%s" % (message,))
	twisted.spread.flavors.NoSuchMethod: No such method: remote_getWorkerInfo
	
2016-06-26T15:27:15-0600 [Broker,client] message from master: buildbot-slave detected, failing to deprecated buildslave API. (Ignoring missing getWorkerInfo method.)
```

### DIFF
--- a/master/buildbot/test/unit/test_worker_protocols_pb.py
+++ b/master/buildbot/test/unit/test_worker_protocols_pb.py
@@ -157,9 +157,12 @@ class TestConnection(unittest.TestCase):
         self.assertEqual(info, r)
         calls = [
             mock.call('getWorkerInfo'),
+            mock.call('print',
+                      message='buildbot-slave detected, failing back to deprecated buildslave API. '
+                              '(Ignoring missing getWorkerInfo method.)'),
             mock.call('getSlaveInfo'),
             mock.call('getCommands'),
-            mock.call('getVersion')
+            mock.call('getVersion'),
         ]
         self.mind.callRemote.assert_has_calls(calls)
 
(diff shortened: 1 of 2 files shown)


rutsky:
Having stack traces in logs during normal operation is still ugly IMO, but I like this approach better since it doesn't introduce fallbacks in new `buildbot-worker`, so it is better for the future versions.

tomprince:
Well, I would argue that using `buildbot-slave` isn't *normal* operation. I agree about it being ugly, though.

tardyp:
ok, lets go for this solution. We'll see how much we get support from that.

PR #2299, opened 2016-07-01
support non critical triggered schedulers

Riziero:
This is not done yet. Some tests are failing. Just wanted to know if my approach is acceptable.

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -50,9 +50,19 @@ class Trigger(BuildStep):
                  copy_properties=None, parent_relationship="Triggered from", **kwargs):
         if schedulerNames is None:
             schedulerNames = []
+        self.schedulerNames = []
         if not schedulerNames:
             config.error(
                 "You must specify a scheduler to trigger")
+                
+        self.schedulerNames_to_critical_dict = {}
+                
+        types = [type(sched) for sched in schedulerNames]
+        if len(set(types)) != 1:
+            config.error(
+                "You can either specify all scheduler names or all tuples of the form (schedulername, True/False)")
+        sched_type = type(schedulerNames[0]) or str
+        
         if (sourceStamp or sourceStamps) and (updateSourceStamp is not None):
             config.error(
                 "You can't specify both sourceStamps and updateSourceStamp")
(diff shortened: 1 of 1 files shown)


tardyp:
Can you explain the details of you needs, and your design?

tardyp:
I think the best is to start with the documentation. This helps to design a proper api change that is clear and usable by everybody.

For what I undestand your usecase looks interresting to have

Riziero:
@tardyp 

Right now when you have a trigger step kicking a number od schedullers, any fail of the schedulers will cause the Trigger step to be marked as failed. In my use case, while I still want to wait for all the kicked schedulers to finish, I want to be able to mark a scheduler as "non critical". When such a scheduler fails, the kicking trigget step will still be green.

With my code you can still do

> factory.addStep(steps.Trigger(schedulerNames=['triggerable_2_fail', 'triggerable_1_green'], waitForFinish=False))

but you can also so:
factory.addStep(steps.Trigger(schedulerNames=[('triggerable_1_green', True), ('triggerable_2_fail', False)], waitForFinish=True))

which reads: if triggerable_2_fail fails the trigger step is "still green"

tardyp:
Here what worries me the most is that your new api is not compatible with the old API.
This is very important for us that our API moves always in a backward compatible way.

Also, for me implicit meaning for a tuple looks not very easy to understand. You change the meaning of schedulerNames to be no more a list of scheduler names, and be a list of schedulername + whether to take the result in account for the global result.

I think I would prefer to have an additional unimportantSchedulerNames argument that worstStatus will take in account

Riziero:
Cool, that was my concern too. I'll fix it up and upload again! Thanks

tardyp:
You can reset your branch to buildbot/master no need to revert the last commit.

Riziero:
would it be fine to change the return value of getSchedulersAndProperties()?

it's now returning a list. I would have it return the same list plus "getSchedulersAndProperties". This way users can also choose getSchedulersAndProperties in a synamic scheduler

tardyp:
I think it would make sense, but we need to make it backward compatible

tardyp:
I'm closing since it does not have code anymore. Please reopen when you find the time to implement the agreed solution

PR #2302, opened 2016-07-02
fix: GerritStatusPush startCB callback is never called

rkashapov:
I found out that startCB callback is never called because startConsuming called with the wrong routing key.


tardyp:
good catch!

PR #2306, opened 2016-07-06
ec2tests: make the tests work in my environment

tardyp:
my environment:
- has proxies configured. but moto does not mock the proxy code properly
- does not have aws creds configured, which does not make create_key_pair happy (it needs to create a fake request to aws, with signed URL, so it needs at least some creds)


tardyp:
Lets merge this and see if metabuildbot is happier

PR #2307, opened 2016-07-06
Get rid of workaround for https://github.com/pypa/setuptools/issues/261

tomprince:


tardyp:
This one is probably worth to take for release

PR #2313, opened 2016-07-10
Make rpm dist a renderable property

anish:
Also changes default dist from el5 to el6


master/buildbot/steps/package/rpm/rpmbuild.py
```diff
@@ -88,6 +94,9 @@ def __init__(self,
             'stdio', logobserver.LineConsumerLogObserver(self.logConsumer))
 
     def start(self):
+
+        self.rpmbuild = ('%s --define "dist %s"' % (self.rpmbuild, self.dist))
```
djmitche: It would be nicer to create a new local variable for this, rather than modifying in place.  Step instances aren't re-used, but absent that knowledge this looks like every run of the step would add an additional `--define dist` argument.

anish: @djmitche That was easy enough to fix, but this pattern repeats itself in the code already. Would it make sense to covert it to a dict instead and then generate command based on that ?

djmitche:
Good point, yes, I think that would be great if you don't mind.

anish:
Already done :)

djmitche:
Hah, I didn't even look at the changes -- yes, looks nice!

PR #2300, opened 2016-07-01
0.9.0rc1 release notes

tardyp:
Creation of a 0.9.0 release note containing concatanation of all beta release notes

After we merge that, I plan to create a 0.9.0 branch to hold the 0.9.0 code freeze. master will become 0.9.1 development

PR #2315, opened 2016-07-11
Non critical triggered schedulers

Riziero:
capture

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -47,9 +47,12 @@ class Trigger(BuildStep):
     def __init__(self, schedulerNames=None, sourceStamp=None, sourceStamps=None,
                  updateSourceStamp=None, alwaysUseLatest=False,
                  waitForFinish=False, set_properties=None,
-                 copy_properties=None, parent_relationship="Triggered from", **kwargs):
+                 copy_properties=None, parent_relationship="Triggered from",
+                 unimportantSchedulerNames=None, **kwargs):
         if schedulerNames is None:
             schedulerNames = []
+        if unimportantSchedulerNames is None:
+            unimportantSchedulerNames = []
         if not schedulerNames:
             config.error(
                 "You must specify a scheduler to trigger")
@@ -63,7 +66,13 @@ class Trigger(BuildStep):
             config.error(
                 "You can't specify both alwaysUseLatest and updateSourceStamp"
             )
+        if not set(schedulerNames).issuperset(set(unimportantSchedulerNames)):
+            config.error(
+                "unimportantSchedulerNames must be a subset of schedulerNames"
+            )
+
         self.schedulerNames = schedulerNames
+        self.unimportantSchedulerNames = unimportantSchedulerNames
         self.sourceStamps = sourceStamps or []
         if sourceStamp:
             self.sourceStamps.append(sourceStamp)
@@ -129,7 +138,11 @@ class Trigger(BuildStep):
     # This customization enpoint allows users to dynamically select which
     # scheduler and properties to trigger
     def getSchedulersAndProperties(self):
-        return [(sched, self.set_properties) for sched in self.schedulerNames]
+        return [{
+            'sched_name': sched,
+            'props_to_set': self.set_properties,
+            'unimportant': sched in self.unimportantSchedulerNames}
+            for sched in self.schedulerNames]
 
     def prepareSourcestampListForTrigger(self):
         if self.sourceStamps:
@@ -167,14 +180,20 @@ class Trigger(BuildStep):
         return all_got_revisions
 
     @defer.inlineCallbacks
-    def worstStatus(self, overall_results, rclist):
+    def worstStatus(self, overall_results, rclist, unimportant_brids):
         for was_cb, results in rclist:
             if isinstance(results, tuple):
-                results, _ = results
+                results, brids_dict = results
 
             if not was_cb:
                 yield self.addLogWithFailure(results)
                 results = EXCEPTION
+
+            # brids_dict.values() rapresents the list of brids kicked by a certain scheduler.
+            # We want to ignore the result of ANY brid that was kicked off
+            # by an UNimportant scheduler.
+            if set(unimportant_brids).issuperset(set(brids_dict.values())):
+                continue
             overall_results = worst_status(overall_results, results)
         defer.returnValue(overall_results)
 
@@ -199,12 +218,30 @@ class Trigger(BuildStep):
     def run(self):
         schedulers_and_props = yield self.getSchedulersAndProperties()
 
+        schedulers_and_props_list = []
+
+        # To be back compatible we need to differ between old and new style
+        # schedulers_and_props can either consist of 2 elements tuple or dictionary
+        for element in schedulers_and_props:
+            if isinstance(element, dict):
+                schedulers_and_props_list = schedulers_and_props
+                break
+            else:
+                # Old-style back compatibility: Convert tuple to dict and make it important
+                d = {
+                    'sched_name': element[0],
+                    'props_to_set': element[1],
+                    'unimportant': False
+                }
+                schedulers_and_props_list.append(d)
+
         # post process the schedulernames, and raw properties
         # we do this out of the loop, as this can result in errors
         schedulers_and_props = [(
-            self.getSchedulerByName(sch),
-            self.createTriggerProperties(props_to_set))
-            for sch, props_to_set in schedulers_and_props]
+            self.getSchedulerByName(entry_dict['sched_name']),
+            self.createTriggerProperties(entry_dict['props_to_set']),
+            entry_dict['unimportant'])
+            for entry_dict in schedulers_and_props_list]
 
         ss_for_trigger = self.prepareSourcestampListForTrigger()
 
@@ -213,7 +250,9 @@ class Trigger(BuildStep):
         results = SUCCESS
         self.running = True
 
-        for sch, props_to_set in schedulers_and_props:
+        unimportant_brids = []
+
+        for sch, props_to_set, unimportant in schedulers_and_props:
             idsDeferred, resultsDeferred = sch.trigger(
                 waited_for=self.waitForFinish, sourcestamps=ss_for_trigger,
                 set_props=props_to_set,
@@ -228,7 +267,8 @@ class Trigger(BuildStep):
             except Exception as e:
                 yield self.addLogWithException(e)
                 results = EXCEPTION
-
+            if unimportant:
+                unimportant_brids.extend(itervalues(brids))
             self.brids.extend(itervalues(brids))
             for brid in brids.values():
                 # put the url to the brids, so that we can have the status from
@@ -247,7 +287,7 @@ class Trigger(BuildStep):
             if self.ended:
                 defer.returnValue(CANCELLED)
             yield self.addBuildUrls(rclist)
-            results = yield self.worstStatus(results, rclist)
+            results = yield self.worstStatus(results, rclist, unimportant_brids)
         else:
             # do something to handle errors
             for d in dl:
--- a/master/buildbot/test/unit/test_steps_trigger.py
+++ b/master/buildbot/test/unit/test_steps_trigger.py
@@ -112,10 +112,12 @@ class TestTrigger(steps.BuildStepMixin, unittest.TestCase):
 
         self.scheduler_a = a = FakeTriggerable(name='a')
         self.scheduler_b = b = FakeTriggerable(name='b')
-        m.scheduler_manager.namedServices = dict(a=a, b=b)
+        self.scheduler_c = c = FakeTriggerable(name='c')
+        m.scheduler_manager.namedServices = dict(a=a, b=b, c=c)
 
         a.brids = {77: 11}
         b.brids = {78: 22}
+        c.brids = {79: 33, 80: 44}
 
         make_fake_br = lambda brid, builderid: fakedb.BuildRequest(
             id=brid, buildsetid=BRID_TO_BSID(brid), builderid=builderid)
@@ -127,14 +129,21 @@ class TestTrigger(steps.BuildStepMixin, unittest.TestCase):
         m.db.insertTestData([
             fakedb.Builder(id=77, name='A'),
             fakedb.Builder(id=78, name='B'),
+            fakedb.Builder(id=79, name='C1'),
+            fakedb.Builder(id=80, name='C2'),
             fakedb.Master(id=9),
             fakedb.Buildset(id=2022),
             fakedb.Buildset(id=2011),
+            fakedb.Buildset(id=2033),
             fakedb.Worker(id=13, name="some:worker"),
             make_fake_br(11, 77),
             make_fake_br(22, 78),
+            fakedb.BuildRequest(id=33, buildsetid=2033, builderid=79),
+            fakedb.BuildRequest(id=44, buildsetid=2033, builderid=80),
             make_fake_build(11),
             make_fake_build(22),
+            make_fake_build(33),
+            make_fake_build(44),
         ])
 
         def getAllSourceStamps():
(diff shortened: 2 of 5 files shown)


Riziero:
in the shot you can see that an unimportant scheduler wont cause a fail of the triggering step.

Riziero:
anybody know what's wrong with the docs? Why's it failing?

master/buildbot/steps/trigger.py
```diff
@@ -198,13 +210,21 @@ def addBuildUrls(self, rclist):
     @defer.inlineCallbacks
     def run(self):
         schedulers_and_props = yield self.getSchedulersAndProperties()
+        
+        # To be back compatible we need to differ between old and new style
+        # schedulers_and_props can either consist of 2 elements or 3 elements tuples
+        
+        # If the user getSchedulersAndProperties funcion does not specify the importance (red len(tup)==2), we make all schedulers important
+        schedulers_and_props = [tup if len(tup)==3 else (tup[0], tup[1], False) for tup in schedulers_and_props]
```
tardyp: I think it would be better to change to a dict or better, a namedtuple so that futureproof is even easier.
https://docs.python.org/3/library/collections.html#collections.namedtuple

If we need to add another argument we would need if len == 4, which will be even worse.
I usually dont like too much if-in-the-future arguments, but in this case this choice wont be undoable as this is API.

Another option is to add self.getSchedulersAndPropertiesWithImportance, which would call the older version, but this is not very clean either.

Riziero: so you are suggesting something like:

if tuple:
  old_style
else:
 #dict
?

sa2ajj:
re failing docs: you use a new name that is not part of the dictionary. you probably want to check formatting of your docs update.

master/buildbot/steps/trigger.py
tardyp: I would rather say if dict: break else: convert_old_style

master/buildbot/steps/trigger.py
```diff
@@ -199,12 +225,29 @@ def addBuildUrls(self, rclist):
     def run(self):
         schedulers_and_props = yield self.getSchedulersAndProperties()
 
+        schedulers_and_props_list = []
+        
+        # To be back compatible we need to differ between old and new style
+        # schedulers_and_props can either consist of 2 elements tuple or dictionary
+        for element in schedulers_and_props:
+            if type(element) == dict:
+                schedulers_and_props_list.append(element)
```
tardyp: I would do short path:
schedulers_and_props_list = schedulers_and_props
break

Riziero:
guys I've got no idea how to fix the docs. Could you give a lead? Thanks :)

tardyp:
You can use "make docschecks" to run locally those tests

Then you can look at the other doc PR to see how to change the spelling file
https://github.com/buildbot/buildbot/pull/2300/files

Riziero:
Getting this:

Running Sphinx v1.4.3
Exception occurred:
  File "conf.py", line 35, in <module>
    raise RuntimeError("sphinxcontrib.blockdiag is not installed. "
RuntimeError: sphinxcontrib.blockdiag is not installed. Please install documentation dependencies with `pip install buildbot[docs]`
The full traceback has been saved in /tmp/sphinx-err-ecxuiN.log, if you want to report the issue to the developers.

pip install buildbot[docs] says I alrady have everything I need.

tardyp:
@Riziero you have to upgrade pip in order to not have the recursive dep bug

pip install -U pip

Riziero:
@tardyp Now getting:

loading pickled environment... not yet created

Sphinx error:
Builder name spelling not registered
make[1]: *** [spelling] Error 1
make[1]: Leaving directory `/home/vagrant/work/master/docs'
make: *** [docschecks] Error 2

Sorry to be a bummer

tardyp:
https://bitbucket.org/dhellmann/sphinxcontrib-spelling/issues/6/builder-name-spelling-not-registered

Is it possible there's a missing C library needed by PyEnchant? If there's a wheel-based package we wouldn't see a failure at installation time but would at runtime when the import fails. Getting the traceback from sphinx would be useful. If that's hard, you could add something like "python -c 'import sphinxcontrib.spelling'" as a command in tox.ini before running sphinx properly.

Riziero:
@tardyp  sudo apt-get install libenchant1c2a did the trick

master/buildbot/steps/trigger.py
```diff
@@ -44,12 +44,14 @@ class Trigger(BuildStep):
 
     flunkOnFailure = True
 
-    def __init__(self, schedulerNames=None, sourceStamp=None, sourceStamps=None,
+    def __init__(self, schedulerNames=None, unimportantSchedulerNames=None, sourceStamp=None, sourceStamps=None,
```
sa2ajj: my suggestion would be to add this new parameter before `**kwargs`: while all over the place we promote explicit parameter naming, some users might "rely" on a particular order of parameters.

master/buildbot/steps/trigger.py
```diff
@@ -63,7 +65,17 @@ def __init__(self, schedulerNames=None, sourceStamp=None, sourceStamps=None,
             config.error(
                 "You can't specify both alwaysUseLatest and updateSourceStamp"
             )
+        if len(set(schedulerNames) & set(unimportantSchedulerNames)) != len(unimportantSchedulerNames):
```
sa2ajj: there's an easier way to check this:
```python
if not set(schedulerNames).issuperset(set(unimportantSchedulerNames)):
 ...
```

(I'm using `superset` here to make it easier to read the code)

master/buildbot/steps/trigger.py
```diff
@@ -129,7 +141,17 @@ def getSchedulerByName(self, name):
     # This customization enpoint allows users to dynamically select which
     # scheduler and properties to trigger
     def getSchedulersAndProperties(self):
-        return [(sched, self.set_properties) for sched in self.schedulerNames]
+
+            schedulers_and_props = []
+            for sched in self.schedulerNames:
+                d = {
+                        'sched_name': sched,
```
sa2ajj: please de-dent these three lines

master/buildbot/steps/trigger.py
```diff
@@ -129,7 +141,17 @@ def getSchedulerByName(self, name):
     # This customization enpoint allows users to dynamically select which
     # scheduler and properties to trigger
     def getSchedulersAndProperties(self):
-        return [(sched, self.set_properties) for sched in self.schedulerNames]
+
+            schedulers_and_props = []
+            for sched in self.schedulerNames:
+                d = {
```
sa2ajj: is there a reason to introduce this temporary variable?  its introduction does not seem to help anything.

master/buildbot/steps/trigger.py
```diff
@@ -63,7 +65,17 @@ def __init__(self, schedulerNames=None, sourceStamp=None, sourceStamps=None,
...
+            config.error(
+                "unimportantSchedulerNames must be a subset of schedulerNames"
+            )
+        if len(schedulerNames) <= len(unimportantSchedulerNames):
+            config.error(
+                    "SchedulerNames must at least contain an important scheduler"
+            )
+
         self.schedulerNames = schedulerNames
+        self.unimportantSchedulerNames = unimportantSchedulerNames
```
sa2ajj: should we store a set instead? (it's already used above and then it'd be handy in the code below)

master/buildbot/steps/trigger.py
```diff
@@ -167,14 +189,18 @@ def getAllGotRevisions(self):
...
             if isinstance(results, tuple):
-                results, _ = results
+                results, brids_dict = results
 
             if not was_cb:
                 yield self.addLogWithFailure(results)
                 results = EXCEPTION
+
+            # continue if this brid has to be ignored
+            if len(set(unimportant_brids) & set(brids_dict.values())) != 0:
```
sa2ajj: do you ignore the whole set if *any* of them were unimportant?

master/buildbot/steps/trigger.py
```diff
@@ -199,12 +225,30 @@ def addBuildUrls(self, rclist):
     def run(self):
         schedulers_and_props = yield self.getSchedulersAndProperties()
 
+        schedulers_and_props_list = []
+
+        # To be back compatible we need to differ between old and new style
+        # schedulers_and_props can either consist of 2 elements tuple or dictionary
+        for element in schedulers_and_props:
+            if type(element) == dict:
```
sa2ajj: `isinstance(element, dict)`

master/buildbot/steps/trigger.py
```diff
@@ -199,12 +225,30 @@ def addBuildUrls(self, rclist):
...
+
         # post process the schedulernames, and raw properties
         # we do this out of the loop, as this can result in errors
         schedulers_and_props = [(
-            self.getSchedulerByName(sch),
-            self.createTriggerProperties(props_to_set))
-            for sch, props_to_set in schedulers_and_props]
+            self.getSchedulerByName(entry_dict['sched_name']),
+            self.createTriggerProperties(entry_dict['props_to_set']),
+            entry_dict['unimportant'])
```
sa2ajj: wrong indentation

master/buildbot/steps/trigger.py
tardyp: +1 good point

master/buildbot/steps/trigger.py
tardyp: It is indeed a complexity of the difference between buildsets and buildrequests.

In this particular case schedulers are marked unimportants. There is one bsid by scheduler, and then theorically one brid by builder (in the case the scheduler triggers several builders).
For the usecase we are working on this is highly unlikely, but we should take care of this corner case though.

So I think this line is correct (it should use superset though), but it might deserve some explaination in comment

master/buildbot/test/unit/test_steps_trigger.py
```diff
@@ -542,3 +550,23 @@ def test_waitForFinish_interrupt(self):
...
+        self.expectTriggeredWith(a=(False, [], {}), b=(False, [], {}))
+        return self.runStep()
+
+    def test_unimportantsShedulerNames(self):
+        self.setupStep(trigger.Trigger(schedulerNames=['a', 'b'], unimportantSchedulerNames=['b']))
+        self.scheduler_a.result = SUCCESS
+        self.scheduler_b.result = FAILURE
+        self.expectOutcome(result=SUCCESS, state_string='triggered a, b')
+        self.expectTriggeredWith(a=(False, [], {}), b=(False, [], {}))
+        return self.runStep()
```
tardyp: We should also test the case where one of the scheduler runs several builders

master/docs/manual/cfg-buildsteps.rst
```diff
@@ -2443,6 +2443,14 @@ Hyperlinks are added to the build detail web pages for each triggered build.
 
         It is possible, but not advisable, to create a cycle where a build continually triggers itself, because the schedulers are specified by name.
 
+``unimportantSchedulerNames``
+    When ``waitForFinish`` is ``True``, all schedulers in this list will not cause the trigger step to fail. unimportantSchedulerNames must be a subset of schedulerNames
+    If ``waitForFinish`` is ``False``, unimportantSchedulerNames will simply be ignored.
+    
+.. note::
+
+        ``schedulerNames`` must have at least one scheduler not present in ``unimportantSchedulerNames``
```
tardyp: I am not sure it is a requirement.

If all schedulers are unimportants, then the result of the trigger step is always success, provided that the schedulers exist

tardyp:
thanks @sa2ajj for the great review and @Riziero for the great improvements. A little bit effort remaining and we can land this!

master/buildbot/steps/trigger.py
Riziero: what I had in mind: if a scheduler is marked as unimportant, then all the builders this scheduler might kick do not affect the result os the step.

master/docs/manual/cfg-buildsteps.rst
Riziero: @tardyp If all schedulers are unimportants. This can not happen. There must at least be an important schduler.

master/buildbot/steps/trigger.py
Riziero: @sa2ajj can you elaborate on this? it's how it was before in terms of indentation.

sa2ajj: Sorry, my mistake: I misread it as a sequence of method calls, while it's a list comprehension.

master/buildbot/test/unit/test_steps_trigger.py
Riziero: @tardyp setupStep how would you about this? SetupStep only setups 2 scheds with 1 build/buildrequest each. Should I override it?

master/docs/manual/cfg-buildsteps.rst
Riziero: Bump @tardyp

tardyp: Why do you say there must be at least one important scheduler.

I can see usecase where I want to trigger lots of parallel tests just for information, and not take the result in account for the global step status

Riziero: you're right. After switching to super set SchedulerNames and unimportantSchedulerNames can actually match.

master/buildbot/test/unit/test_steps_trigger.py
Riziero: bump ;)

master/docs/manual/cfg-buildsteps.rst
Riziero: that also means that the following code (and its test) must go

 if len(schedulerNames) <= len(unimportantSchedulerNames):
         config.error(
                 "SchedulerNames must at least contain an important scheduler"
         )

master/buildbot/test/unit/test_steps_trigger.py
Riziero: @tardyp how would you go about this? SetupStep only setups 2 scheds with 1 build/buildrequest each. Should I override it?

master/docs/manual/cfg-buildsteps.rst
tardyp: indeed

master/buildbot/test/unit/test_steps_trigger.py
tardyp: You can just create another scheduler 'c' in SetupStep which only one unit test will use

Riziero:
can we land this or some more stuff to polish? Thanks!
@tardyp

tardyp:
Please clean the git history. Git rebate --help

Riziero:
what do you mean by clening up the history? Squashing all commits into one? Pardon my git knwoledge is very rudimental.

tardyp:
Yes please

Riziero:
@tardyp been trying for 2 hours with no luck - I ended up messing up my branch with 2 more stupid commits, one of them being just empty.

tardyp:
Sorry, I was afk here is the manual
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

This is important for github contributor to know this very basic git tricks. I would not do you a favor to do it for you.

Riziero:
Hey @tardyp 

I am really trying my best. All I am doing from cmd after switching to my WIP branch is

git rebase -i -k e925c49e4d752ce82b8b1b4defe552dc8f856e8 with this hash being last commit not made by me. Basically if I understand things rightly I am telling git to collapse all the following commits into a signle one. 

image

After I save and quit, I get an error about which I found http://git.661346.n2.nabble.com/Confusing-error-message-in-rebase-when-commit-becomes-empty-td7612948.html

tardyp:
You can hang on irc for live help.

basically you should do this in case of empty commits:
git rebase --skip

tardyp:
good work! thanks for keeping up!

Riziero:
@tardyp I was expecting this. When I forked nineRC1 release notes hadnt yet been made. As per guidelines I added my rel note line to index.rst which has changed a lot. hence the conflict.

How do I go about this?
Thanks

tardyp:
looks good! thanks!

PR #2316, opened 2016-07-12
Fix the document for GerritStatusPush callback

loganchien:
This commit removes the redundant `msg=` from gerritSummaryCB because it
will result in SyntaxError.  This error is not in master branch.  It
seems to be introduced while backporting the patch to **eight** branch.


sa2ajj:
Thank you

PR #2317, opened 2016-07-13
LatentWorker do not insubstantiate at shutdown

tardyp:
we already stop the workers at stopService, which is called by the Application's hierarchy stop

Doing the complicated SystemEventTrigger is just uneeded duplication

### DIFF
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -17,6 +17,7 @@ import time
 from email.message import Message
 from email.utils import formatdate
 
+from future.utils import itervalues
 from twisted.internet import defer
 from twisted.python import log
 from twisted.python.reflect import namedModule
@@ -35,7 +36,6 @@ from buildbot.util import ascii2unicode
 from buildbot.util import service
 from buildbot.util.eventual import eventually
 from buildbot.worker_transition import deprecatedWorkerClassProperty
-from future.utils import itervalues
 
 
 class AbstractWorker(service.BuildbotService, object):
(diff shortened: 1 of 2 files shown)


master/buildbot/worker/base.py
```diff
@@ -826,7 +818,9 @@ def _soft_disconnect(self, fast=False):
             self.missing_timer.cancel()
             self.missing_timer = None
 
-        if self._substantiation_notifier:
+        # if master is stopping, we will never achieve consistent state, as workermanager
+        # wont accept new connection
+        if self._substantiation_notifier and self.master.running:
```
tardyp: This is a fix for

http://trac.buildbot.net/ticket/3581

tardyp:
Note that those fixes are not unit tested in this PR. They are tested in the tests for hyper.
I still need to write the tests specific for the core

master/buildbot/worker/base.py
```diff
@@ -17,7 +17,6 @@
 from email.message import Message
 from email.utils import formatdate
 
-from future.utils import itervalues
```
sa2ajj: this import order change seems to be unnecessary

tardyp: its made automatically by isort.

For some reason isort does not run on travis as it should, so we start taking technical debt about it.

That said. I would expect that future is part of the first. So maybe it is the config of my editor which cannot find the isort configuration

tardyp: ahah I did not see the build failure. So isort travis definitly work and my editor is misconfigured :-p

PR #2321, opened 2016-07-16
few Python 3 related fixes

sa2ajj:
* do not use 'print' use 'twisted.python.log'
* use 'except ... as ...' instead of 'except ..., ...'

### DIFF
--- a/master/buildbot/www/oauth2.py
+++ b/master/buildbot/www/oauth2.py
@@ -15,9 +15,11 @@
 from posixpath import join
 
 import requests
+
 from future.moves.urllib.parse import parse_qs
 from future.moves.urllib.parse import urlencode
 from future.utils import iteritems
+
 from twisted.internet import defer
 from twisted.internet import threads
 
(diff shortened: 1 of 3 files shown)


master/buildbot/www/oauth2.py
```diff
@@ -51,7 +54,7 @@ def renderLogin(self, request):
                 details.update(infos)
             request.getSession().user_info = details
             state = request.args.get("state", [""])[0]
-            print repr(state)
+            log.msg(repr(state))
```
tardyp: I think that one is a left-over debug print.

master/buildbot/www/oauth2.py
```diff
@@ -15,11 +15,14 @@
 from posixpath import join
 
 import requests
+
```
tardyp: isort should group those 2 paragraphs back

sa2ajj: yup: build failure :smile:

master/buildbot/www/oauth2.py
sa2ajj: it might be, however since it's in the code, i'd like to make it Python 3 compatible.

tardyp: or just delete the line

sa2ajj:
Yeah, deleting those lines is probably a better idea.

tardyp:
thanks

PR #2324, opened 2016-07-25
Fix the cancel and stop story

tardyp:
Cancel of a buildrequest was not working in the buildrequest page.

(pushing to release branch for that reason http://trac.buildbot.net/ticket/3582)

When there are tons of buildrequests in the queue this is very painful to stop them all manually

So we also add a feature that was in buildbot eight cancel whole queue. Unlike in eight the whole queue means buildrequests *and* running builds for the sake of simplicity

We can still change the code to add a separate button to avoid killing running builds.

### DIFF
--- a/master/docs/relnotes/index.rst
+++ b/master/docs/relnotes/index.rst
@@ -25,8 +25,12 @@ Master
 Features
 ~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


djmitche:
Looks good to me!

PR #2326, opened 2016-07-28
Enhance buildrequest summary so that it look more like a buildsummary

tardyp:
When there are a lot of triggered builds, and no slave available
there are few information to the user on the buildrequest

This commit also enhances the display of the buildsumary by:

- use flexbox to automatically fill the space wether reason or status is the larger

- stop using &nbsp; for trailing space which will not wrap properly

- fix triggerable so that it properly set the reason of the buildset

image

previous version:
image


tardyp:
fixed for better alignment

image

ssavrim:
:+1:

tardyp:
I will merge that one as this is more straight forward. Let me know if there are issues I can address them in a later PR

PR #2327, opened 2016-07-28
LatentWorker better handling of substanciation failure cases

tardyp:
2 cases are handled here:

- substanciation timeout
- cannot create the build directories

integration test those cases, and refactor to simplify the code

PR #2329, opened 2016-07-28
Master shutdown

tardyp:
Based on #2327 and #2328 

    master shutdown: stop all builds at master shutdown

    Previously when master shutdown, all the builds where marked RETRY
    because the workers were forcefully disconnected.

    This has bad impact like not putting correct reason why the build has been stopped.
    User sees slave disconnected, while the reason is master shutdown.


tardyp:
need unit tests. pushing it early to get feedback from @tomprince who did rework the latent code recently.

tardyp:
@tomprince would you be able to give some feedback?

tardyp:
please let me know if there are comments. merging this to move forward on metabuildbot hyper

PR #2330, opened 2016-07-31
Dockerfiles are broken, fix and upgrade ubuntu container to 16.04

anish:
Fixes issues like
```
docker: Error response from daemon: Container command '/usr/src/contrib/docker/master/start_buildbot.sh' not found or does not exist..
```


tardyp:
Please note that there is another dockerfile which is using alpine and which is much more lightweight.

What is the advantage for you to use ubuntu version?

anish:
None,  I was just using it to test changes

PR #2332, opened 2016-07-31
Increased size of builder identifier from 20 to 50 

faulteh:
Fixes #3430

The DB schema has these fields of type TEXT which doesn't have an upper size limit. The builder identifier was only 20 while steps' and workers' names in the data module had size 50.

Going to a URL such as 
http://10.100.200.25:8010/api/v2/builders?name=runtestswithareallylongnamethatshouldnotbreak
no longer produces an error with this minor patch.


tardyp:
looks good to me

PR #2336, opened 2016-08-07
Error class stores status in strings

tardyp:
but setResponseCode wants an int.
tw16 started to enforce Error's status  to be string, while before we always sent an int.

http://trac.buildbot.net/ticket/3580


djmitche:
looks good to me

PR #2337, opened 2016-08-07
forcesched: fix owner parameter when no authentication is used

tardyp:
- if 'anonymous' user is send in owner parameter, then we use the username parameter from forcesched
- fix username validation so that it uses the same regex as the mail reporter
- fix forcesched to correctly display the validation error message

fixes http://trac.buildbot.net/ticket/3587

### DIFF
--- a/master/buildbot/schedulers/forcesched.py
+++ b/master/buildbot/schedulers/forcesched.py
@@ -12,7 +12,6 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-import email.utils as email_utils
 import re
 import traceback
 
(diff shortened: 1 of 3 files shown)


master/buildbot/schedulers/forcesched.py
```diff
@@ -264,8 +264,8 @@ def parse_from_arg(self, s):
         if not s and not self.required:
             return s
         if self.need_email:
-            e = email_utils.parseaddr(s)
-            if e[0] == '' or e[1] == '':
+            res = VALID_EMAIL_ADDR.search(s)
```
sa2ajj: you don't use `res` after this line, do we need an extra variable?

sa2ajj:
I'm still OK with the change. If nobody objects it, I'll merge it tomorrow.

master/buildbot/schedulers/forcesched.py
tardyp: I use res the line after

sa2ajj: my reading skills are not getting better :(  sorry

PR #2339, opened 2016-08-08
fix for sphinx > 1.4.3

tardyp:
sphinx now wants 5 columns in its index description...
here you go.. I am not sure what is the last item, but None works (and is what sphinx is doing in the warning code)

https://github.com/sphinx-doc/sphinx/issues/2673

### DIFF
--- a/master/docs/bbdocs/ext.py
+++ b/master/docs/bbdocs/ext.py
@@ -71,7 +71,7 @@ class BBRefTargetDirective(Directive):
             else:
                 indextype = 'single'
                 indexentry = tpl % (fullname,)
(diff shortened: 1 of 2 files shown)


tardyp:
@tatsuhiro-t is this the right fix?

sa2ajj:
Thanks looking into it.

sa2ajj:
I think we should *require* 1.4.4 to avoid maintenance hassle...

tardyp:
I agree it is annoying to have regression because of dependency update, but I think it is worse to have pinning for every deps. This makes distro job even harder and also prevents us to know there is regression

PR #2344, opened 2016-08-18
worker: fix invalid DockerLatent volumes name

faudebert:
Volumes from DockerLatentWorker were named against the 'bind:volume'
string instead of the 'volume' part only.

According to docker-py documentation[1], `volumes` parameter from
create_container() must only contain volumes name (ie. '/mnt/vol1') and
not the complete definition (ie. '/home/user1:/mnt/vol1').

This lead to unwanted entry in the container filesystem (ie.
'/home/user1:/mnt/vol1/' folder was created inside the container).

[1] https://docker-py.readthedocs.io/en/latest/volumes/

### DIFF
--- a/master/buildbot/test/unit/test_worker_docker.py
+++ b/master/buildbot/test/unit/test_worker_docker.py
@@ -86,14 +86,14 @@ class TestDockerLatentWorker(unittest.TestCase):
         bs = self.ConcreteWorker('bot', 'pass', 'tcp://1234:2375', 'worker', ['bin/bash'],
                                  volumes=[Interpolate('/data:/buildslave/%(kw:builder)s/build', builder=Property('builder'))])
         id, name = yield bs.start_instance(self.build)
-        self.assertEqual(bs.volumes, ['/data:/buildslave/docker_worker/build'])
+        self.assertEqual(bs.volumes, ['/buildslave/docker_worker/build'])
 
     @defer.inlineCallbacks
     def test_volume_no_suffix(self):
         bs = self.ConcreteWorker(
             'bot', 'pass', 'tcp://1234:2375', 'worker', ['bin/bash'], volumes=['/src/webapp:/opt/webapp'])
         yield bs.start_instance(self.build)
-        self.assertEqual(bs.volumes, ['/src/webapp:/opt/webapp'])
+        self.assertEqual(bs.volumes, ['/opt/webapp'])
         self.assertEqual(
             bs.binds, {'/src/webapp': {'bind': '/opt/webapp', 'ro': False}})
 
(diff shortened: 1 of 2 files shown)


anish:
@faudebert that curious, I'm using binds on my latent worker and do not see this behavior. Can you share a config to reproduce ?

sa2ajj:
LGTM

faudebert:
@anish It does not prevent the path to be mounted inside the container but this creates an unwanted volume.

Here is a minimal configuration based on PyFlakes illustrating the issue:

```python
from buildbot.plugins import *
from buildbot.worker.docker import DockerLatentWorker

hostname = 'some_hostname_here'
volume_options = '/tmp/host:/mnt/host:ro'
bind, volume, opts = volume_options.split(":", 2)

c = BuildmasterConfig = {}

c['workers'] = []
c['workers'].append(DockerLatentWorker(
    'docker-worker',
    'docker-password',
    docker_host="tcp://%s:2375" % hostname,
    followStartupLogs=True,
    hostconfig={'network_mode': 'host'},
    masterFQDN=hostname,
    image='buildbot/buildbot-worker:v0.9.0rc1',
    volumes=["%s:%s:%s" % (bind, volume, opts)]
))

c['schedulers'] = []
c['schedulers'].append(schedulers.ForceScheduler(
    name="force",
    builderNames=["runtests"]
))

factory = util.BuildFactory()
factory.addStep(steps.ShellCommand(command=["ls", "-ld", '--', volume_options, volume]))

c['builders'] = []
c['builders'].append(
    util.BuilderConfig(name="runtests",
                       workernames=["docker-worker"],
                       factory=factory)
)

c['status'] = []
c['title'] = "Pyflakes"
c['titleURL'] = "https://launchpad.net/pyflakes"
c['buildbotURL'] = "http://%s:8010/" % hostname
c['www'] = {'port': 8010, 'plugins': {'waterfall_view': {}, 'console_view': {}}}
c['db'] = {'db_url': "sqlite:///state.sqlite"}
```
The build will output something like this:

```shell
ls -ld -- /tmp/host:/mnt/host:ro /mnt/host
[...]
drwxr-xr-x 2 1000 1000 4096 Aug 19 07:58 /mnt/host
drwxr-xr-x 2 root root   40 Aug 19 09:15 /tmp/host:/mnt/host:ro
```

Here is a way to reproduce it using few `docker-py` calls:

```python
from docker import Client

cli = Client(base_url='unix://var/run/docker.sock')

volume_options = '/tmp/host:/mnt/host'
bind, volume = volume_options.split(":", 1)

container = cli.create_container(
    'busybox', 'sleep 60',
    volumes=[volume_options],
    host_config=cli.create_host_config(binds={
        bind: {
            'bind': volume,
            'mode': 'rw',
        }
    })
)

container_id = container.get('Id')
cli.start(container_id)

exec_id = cli.exec_create(
    container_id,
    cmd=['ls', '-ld', volume, volume_options]
)

exec_stdout = cli.exec_start(exec_id, tty=True)
print(exec_stdout)
```

It's unclear to me what's really going on inside docker. It seems to me we have an host mount without an associated volume. AFAIK, using docker CLI, it's not possible to do so.

I got the same behavior with those versions:
 - `docker`: 1.11.2 and 1.12.0
 - `docker-py`: 1.8.1 and 1.9.0

tardyp:
I am not sure if we did not merge this too early..

@anish could you give some feedback and tell if your volume use is not broken by this PR?

tardyp:
With docker often breaking apis I fear this could brake older version of docker.

That would be fine by me but we should at least document if volumes are only supported with docker >= 1.11.2

PR #2346, opened 2016-08-20
Fix BitbucketPullrequestPoller change detection

sam-bristow:
Currently the BitbucketPullrequestPoller treats all pull requests as
if they have been updated, rather than filtering them properly. This is
because the existing check compares the short hash of the incoming
changeset with the full hash of the current revision.

This patch changes the behaviour to compart the short hashes of the
incoming and current revisions.

The Bitbucket API returns short hashes for a number of queries. The
test data has been updated to reflect this.

fixes #3597

Signed-off-by: Sam Bristow <sam.bristow@gmail.com>

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -102,7 +102,8 @@ class BitbucketPullrequestPoller(base.PollingChangeSource):
             if not self.branch or branch in self.branch:
                 current = yield self._getCurrentRev(nr)
 
-                if not current or current != revision:
+                # compare _short_ hashes to check if the PR has been updated
+                if not current or current[0:12] != revision[0:12]:
                     # parse pull request api page (required for the filter)
                     page = yield client.getPage(str(pr['links']['self']['href']))
                     pr_json = json.loads(page, encoding=self.encoding)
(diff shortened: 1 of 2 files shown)


tardyp:
Hi Sam,
thanks for contributing this fix!

We need to make sure that the code still works for older version of bitbucket, which still use the long sha1 (if I understand correctly)

Can you add some tests for the older version?

sam-bristow:
I'm not quite sure what you mean by older versions of Bitbucket? Bitbucket Cloud, which is what this change-source  works with, is a web service and Atlassian only exposes a single "version" at bitbucket.org.

Can you point me towards an example of an older version which we need to work with?

sam-bristow:
I've amended the original PR to be more explicit about comparing short hashes. This should be more robust in case the Bitbucket API starts returning full hashes for PR changes.

zapster:
@sam-bristow thanks for fixing this. Regarding different versions. There is also [Bitbucket Server](https://www.atlassian.com/software/bitbucket/server) although I'm not sure if it is API compatible (but I would hope so).

sam-bristow:
Bitbucket Cloud and Bitbucket server (nee Stash) are completely separate code-bases, and don't share an API. It's all very confusing. This change-source will only work with the Cloud version.

tardyp:
Indeed, I was thinking stash. A shame they dont share api

PR #2351, opened 2016-08-24
Allow JSONStringDownload to work on renderables

dhouck:
Currently the `json.dumps` call is in the `__init__` method, where it does not have access to property values or other renderable content. I move it to the `start` method and add the object to dump to the `renderables` list.

I do not convert it or any of the other steps in the file to new-style build steps.

### DIFF
--- a/master/buildbot/steps/transfer.py
+++ b/master/buildbot/steps/transfer.py
@@ -562,6 +562,8 @@ class JSONStringDownload(StringDownload, WorkerAPICompatMixin):
 
     name = "json_download"
 
+    renderables = ['o']
+
     def __init__(self, o, workerdest=None,
                  slavedest=None,  # deprecated, use `workerdest` instead
                  **buildstep_kwargs):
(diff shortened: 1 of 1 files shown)


master/buildbot/steps/transfer.py
```diff
@@ -562,6 +562,8 @@ class JSONStringDownload(StringDownload, WorkerAPICompatMixin):
 
     name = "json_download"
 
+    renderables = ['o']
+
     def __init__(self, o, workerdest=None,
```
Stibbons: it is a pretty bad thing to have parameter name with only 1 letter...

master/buildbot/steps/transfer.py
```diff
@@ -577,11 +579,15 @@ def __init__(self, o, workerdest=None,
...
         if 's' in buildstep_kwargs:
             del buildstep_kwargs['s']
-        s = json.dumps(o)
         StringDownload.__init__(
-            self, s=s, workerdest=workerdest, **buildstep_kwargs)
+            self, s=None, workerdest=workerdest, **buildstep_kwargs)
+
+    def start(self):
+        self.s = json.dumps(self.o)
+        return self.super_class.start(self)
```
tardyp: why don't you use super?

            super(JSONStringDownload, self).start()

master/buildbot/steps/transfer.py
tardyp: agreed. This is also the case for StringDownload :-/

tardyp:
also, unit tests needs to be updated

dhouck:
This is what I get for moving code around without testing that the move didn’t break things; I’ll upload the fix after I test it this time to make sure it works right.

master/buildbot/steps/transfer.py
dhouck: Because the other code in the file didn’t use `super` and I didn’t know why. I will change to `super` if there isn’t a reason for not doing it.

master/buildbot/steps/transfer.py
dhouck: Yeah, I was trying to make the minimal set of changes that would work; the parameter was already named `o` so I didn’t change it.

dhouck:
I’ll be able to get the tests put on this over the weekend, but probably not earlier.

tardyp:
The week end is over? did you get too much beer? :)

tardyp:
I am closing this stalled pull request. Please reopen if you get some time to finish it!

PR #2353, opened 2016-08-25
Use print() function everywhere as specified in PEP3105.

rodrigc:
This works on Python 2.6 and higher.  This is mandatory on Python 3.

### DIFF
--- a/common/porttostable.py
+++ b/common/porttostable.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import os
 from subprocess import CalledProcessError
 from subprocess import check_output
@@ -28,7 +29,7 @@ for pr in to_port['items']:
                 continue
             if 'fatal: bad object' in e.output:
                 continue
-            print "cannot automatically cherry-pick", pr['number'], c['sha'], title,e.output
+            print("cannot automatically cherry-pick", pr['number'], c['sha'], title,e.output)
         else:
             summary += "\n#{number}: {title}".format(number=pr['number'], title=title , **c)
-print summary
+print(summary)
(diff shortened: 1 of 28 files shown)


tardyp:
LGTM.

what is the command you used to do that?
Couldn't we had an automated test to make sure print are not reintroduced by mistake?

like we do for isort:

  - "[ $TESTS != lint ] || isort --check -df `git ls-files |grep '.py$'`"

rodrigc:
I used this command in the top-level directory:

```
2to3 -f print .
```

I am not sure how to implement a check to prevent re-introducing this.
Hopefully we can get to the point where we can add a Python 3 build to CI, since
that is the easiest way to verify it.  It's too early for a Python 3 CI build at this point.

tardyp:
what you can do is to run the command, and then look at git diff to see if some file were modified.

This is what I did with isort until I realized there were a check option

rodrigc:
I'm not too handy with those kinds of checks and am scared I may break something,
for example if there is a legitimate use of "print " somewhere like in a string.
I'll leave that for a subsequent PR.

tardyp:
You did break isort actually.
please run make isort on root dir to fix it.

tardyp:
this needs rebase now I merged the other PR.. :-/

rodrigc:
@tardyp I rebased this

djmitche:
:+1: to the idea of checking these things to prevent backsliding.  At one point we had replaced all print with print(), I think, but clearly we lost some!

tardyp:
Indeed. And I plaid mostly culprit of that prints

Le sam. 27 août 2016 15:34, Dustin J. Mitchell <notifications@github.com> a
écrit :

Stibbons:
you can check "from __future__ import print_function" appears in every python files

PR #2356, opened 2016-08-25
Replace xrange() with range() for Python 3 compatibility

rodrigc:
xrange() is gone in Python 3.
In Python 3, range() returns a view, not a list,
so in a few places where we really need a list, we
need to wrap it with list().

See:
http://python3porting.com/differences.html#range-and-xrange

### DIFF
--- a/master/buildbot/data/base.py
+++ b/master/buildbot/data/base.py
@@ -49,7 +49,7 @@ class ResourceType(object):
 
     def getEndpoints(self):
         endpoints = self.endpoints[:]
-        for i in xrange(len(endpoints)):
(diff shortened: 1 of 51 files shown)


tardyp:
but xrange is an iterator in py2 so we need to make sure the code where xrange is used does not use a list in place (and use huge memory for nothing)

I think future as something to deal with that. (we are using future as a compat library)

rodrigc:
@tardyp I reworked this.

tardyp:
it needs rebase after merge of the other branch :-}

rodrigc:
OK, I rebased.

PR #2359, opened 2016-08-26
Replace zope.interface.implements() with @zope.interface.implementer

rodrigc:
On Python 3, zope.interface.implements() fails with:

builtins.TypeError: Class advice impossible in Python3.  Use the @implementer class decorator instead.


rutsky:
LGTM

PR #2360, opened 2016-08-26
Use byte strings when writing data.

rodrigc:
Without this fix, on Python 3, we get errors such as:

  File "/Users/crodrigues/b1/worker/buildbot_worker/commands/transfer.py", line 369, in _writeData
    self.fp.write(data)
builtins.TypeError: a bytes-like object is required, not 'str'


rutsky:
LGTM, thanks!
Should work, since in Python 2.6 and 2.7 byte literals are equivalent to `str`.

PR #2361, opened 2016-08-26
Replace basestring with future.utils.string_types.

rodrigc:
basestring is gone on Python 3.
On Python 2, future.utils.string_types expands to basestring.
On Python 3, future.utils.string_types expands to str.

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -21,6 +21,7 @@ from types import MethodType
 
 from future.utils import iteritems
 from future.utils import itervalues
+from future.utils import string_types
 from future.utils import text_type
 from twisted.python import failure
 from twisted.python import log
@@ -338,7 +339,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
 
         def copy_str_param(name, alt_key=None):
             copy_param(name, alt_key=alt_key,
-                       check_type=basestring, check_type_name='a string')
+                       check_type=string_types, check_type_name='a string')
 
         copy_str_param('title', alt_key='projectName')
         copy_str_param('titleURL', alt_key='projectURL')
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -16,7 +16,9 @@ import re
 
 from future.utils import iteritems
 from future.utils import itervalues
+from future.utils import string_types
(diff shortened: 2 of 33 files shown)


rutsky:
LGTM.

rutsky:
@rodrigc when you will finish Python 3 porting please don't forget to add release notes for your changes to https://github.com/buildbot/buildbot/blob/master/master/docs/relnotes/index.rst

master/buildbot/process/buildstep.py
```diff
@@ -1229,7 +1231,7 @@ def regex_log_evaluator(cmd, _, regexes):
         # we won't be changing "worst" unless possible_status is worse than it,
         # so we don't even need to check the log if that's the case
         if worst_status(worst, possible_status) == possible_status:
-            if isinstance(err, (basestring)):
+            if isinstance(err, (string_types)):
```
Stibbons: why there is parenthesis ? This does not add anything

master/buildbot/config.py
```diff
@@ -21,6 +21,7 @@
 
 from future.utils import iteritems
 from future.utils import itervalues
+from future.utils import string_types
```
Stibbons: This [do](http://python-future.org/compatible_idioms.html#basestring)c recommends to use:
```
from past.builtins import basestring
```

which one is better?

master/buildbot/process/buildstep.py
rodrigc: You are correct, the parenthesis is not needed here.  It was there before my change,
but it wasn't needed before.  I can take it out.  That is not a problem.

master/buildbot/config.py
rodrigc: future.utils.string_types does the job quite nicely here, and I am confident that its usage
here is correct.  The past module is to be avoided and used sparingly.  The past module is meant for people who have huge codebases and don't want to make lots of changes.

The strategy that I am using is to bring the buildbot codebase forward while still preserving behavior and not breaking things on Python 2.6/2.7.

Stibbons: make sense, thx

tardyp:
needs rebase

rodrigc:
I rebased

sa2ajj:
@rodrigc it seems that this PR needs another rebase.

PR #2362, opened 2016-08-26
On Python 3, the __builtin__ module is gone, replaced by builtins.

rodrigc:


master/buildbot/test/unit/test_config.py
```diff
@@ -12,7 +12,10 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-import __builtin__
+try:
```
tardyp: looks like future already implement a buildtins module for py2

http://python-future.org/imports.html#imports-of-builtins

rodrigc: I can't use the future package's implementation of the builtins module for py2,
because this test is doing funky stuff of patching the open() method to override it in a test.  I tried to use it and the test broke.

tardyp: can you add a comment about that?

rodrigc:
OK, I added a comment.

PR #2371, opened 2016-08-27
Use future.utils.iterkeys() and future.utils.itervalues() for Python 3.

rodrigc:
dict.keys() / dict.values() returns list on Python 2, and
a generator on Python 3.  We need a list.

Without this fix, we get this error on Python 3:

```
 File "/Users/crodrigues/uio/buildbot/master/buildbot/worker_transition.py", line 202, in deprecatedWorkerModuleAttribute
    attribute_name = scope.keys()[scope.values().index(attribute)]
AttributeError: 'dict_values' object has no attribute 'index'
```

### DIFF
--- a/master/buildbot/worker_transition.py
+++ b/master/buildbot/worker_transition.py
@@ -199,7 +199,9 @@ def deprecatedWorkerModuleAttribute(scope, attribute, compat_name=None,
         "scope must be module, i.e. locals()"
 
     if new_name is None:
-        attribute_name = scope.keys()[scope.values().index(attribute)]
+        scope_keys = list(scope.keys())
+        scope_values = list(scope.values())
+        attribute_name = scope_keys[scope_values.index(attribute)]
     else:
         attribute_name = new_name
 
@@ -235,7 +237,9 @@ def deprecatedWorkerClassProperty(scope, prop, compat_name=None,
     """
 
     if new_name is None:
-        attribute_name = scope.keys()[scope.values().index(prop)]
+        scope_keys = list(scope.keys())
+        scope_values = list(scope.values())
+        attribute_name = scope_keys[scope_values.index(prop)]
     else:
         attribute_name = new_name
 


master/buildbot/worker_transition.py
```diff
@@ -199,7 +201,9 @@ def deprecatedWorkerModuleAttribute(scope, attribute, compat_name=None,
         "scope must be module, i.e. locals()"
 
     if new_name is None:
-        attribute_name = scope.keys()[scope.values().index(attribute)]
+        scope_keys = list(iterkeys(scope))
```
tardyp: if we need a list, there is no need to use the future util.

We can just do list(scope.keys()) which will work well on py2 and py3

even if it might be less optimal, I think it is more readable

rodrigc: @tardyp do you want me to get rid of itervalues() also, and use list(scope.values()) ?

master/buildbot/worker_transition.py
```diff
@@ -199,7 +200,9 @@ def deprecatedWorkerModuleAttribute(scope, attribute, compat_name=None,
         "scope must be module, i.e. locals()"
 
     if new_name is None:
-        attribute_name = scope.keys()[scope.values().index(attribute)]
+        scope_keys = list(scope.keys())
+        scope_values = list(itervalues(scope))
```
tardyp: same apply for itervalues, of course.

rodrigc:
@tardyp OK, I made the change

vbraun:
Any chance of getting this into 0.9? Right now, 0.9.0rc4 is totally unusable on py3.

rodrigc:
@vbraun I doubt that this is worth getting into 0.9.  buildbot needs a lot more than just this small patch to get it even functional on Python 3.  I've submitted multiple patches.

These ones were accepted and merged:
https://github.com/buildbot/buildbot/pulls?q=is%3Apr+author%3Arodrigc+is%3Aclosed

and these ones are still pending:
https://github.com/buildbot/buildbot/pulls/rodrigc

PR #2373, opened 2016-08-28
Fix invocation of trial on Windows with Twisted 16.4.0

rodrigc:
In Twisted 16.4.0, trial was converted to a Python console
script.  It must be invoked as "trial" or "python -m twisted.trial".
%PYTHON%\\Scripts\trial.py is gone in Twisted 16.4.0.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,7 @@ install:
 build: false
 
 test_script:
-  - "%PYTHON%\\python.exe %PYTHON%\\Scripts\\trial.py --reporter=text --rterrors buildbot.test buildbot_worker.test"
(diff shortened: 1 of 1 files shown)


Stibbons:
someone knows why "python -m trial" is better than just trial, just for my information?

rodrigc:
"python -m twisted.trial" is equivalent, it is not better or worse.
In Twisted 16.4.0, **pip install twisted** on Windows will install a binary **trial.exe**
in the PATH.  I needed to quickly submit something that worked, because with the new Twisted,
the appveyor build suddenly started breaking, and I am submitting a lot of patches.

Stibbons:
ok thx :)

PR #2380, opened 2016-08-29
Use inspect.signature() on Python 3, since inspect.getargspec() is deprecated

rodrigc:


tardyp:
gotta make a coveragepy3 builder so that py3 coverage is also reported

PR #2382, opened 2016-08-29
Remove class variables interfering with __slots__ in ListResult

rodrigc:
In Python 3, this was triggering this error:

builtins.ValueError: 'offset' in __slots__ conflicts with class variable

In Python 2, if you defined slots and after that defined a property
with the same name the slots descriptor was replaced silently. In
Python 3 this isn't allowed anymore.

See:
https://github.com/paramiko/paramiko/issues/103#issuecomment-15127799
http://bugs.python.org/issue12766

### DIFF
--- a/master/buildbot/data/base.py
+++ b/master/buildbot/data/base.py
@@ -134,21 +134,18 @@ class ListResult(UserList):
 
     __slots__ = ['offset', 'total', 'limit']
 
-    # if set, this is the index in the overall results of the first element of
-    # this list
-    offset = None
-
-    # if set, this is the total number of results
-    total = None
-
-    # if set, this is the limit, either from the user or the implementation
-    limit = None
-
     def __init__(self, values,
                  offset=None, total=None, limit=None):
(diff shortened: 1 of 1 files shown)


tardyp:
I am not sure of the usefulness of _slots_ so 👍 for me

rodrigc:
If used properly, __slots__ can reduce memory usage if you create a lot of the same object:
http://book.pythontips.com/en/latest/__slots__magic.html

Based on that web page, I'm thinking I should change the patch to keep the __slots__, but eliminate the class-level variables offset, total, limit.

tardyp:
I dont think there are so many of those objects in a typical buildbot run, but that's better anyway :+1:

rodrigc:
I've submitted a new patch which keeps the __slots__ in ListResult

djmitche:
That was the intent here -- saving space and time -- but that's not backed up by any actual measurements, so either removing `__slots__` or keeping it is fine.

PR #2383, opened 2016-08-29
Fix test_contrib_buildbot_cvs_mail on Python 3

rodrigc:
This mainly addresses unicode issues.

I highly recommend that anyone doing Python 2 -> Python 3
porting work which involves unicode should watch Ned Batchelder's presentation:

http://nedbatchelder.com/text/unipain.html

Specifically, he talks about the **Unicode sandwich**, where
as much as possible, within the confines of your application, you should
try to maintain your data as unicode,
but when you _send the data somewhere else_ such as *writing to a file* or
*sending to a socket*, you should encode the data to bytes, and treat it as bytes.

### DIFF
--- a/master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
+++ b/master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
@@ -13,16 +13,20 @@
 #
 # Copyright Buildbot Team Members
 from __future__ import unicode_literals
+
 import os
 import re
 import sys
 
+from future.utils import text_type
 from twisted.internet import defer
 from twisted.internet import protocol
 from twisted.internet import reactor
 from twisted.internet import utils
 from twisted.python import log
 from twisted.trial import unittest
+
+from buildbot.test.util.misc import encodeExecutableAndArgs
 
 test = '''
 Update of /cvsroot/test
@@ -87,9 +91,11 @@ golden_1_12_regex = [
 class _SubprocessProtocol(protocol.ProcessProtocol):
 
     def __init__(self, input, deferred):
+        if isinstance(input, text_type):
+            input = input.encode('utf-8')
         self.input = input
         self.deferred = deferred
-        self.output = ''
+        self.output = b''
 
     def outReceived(self, s):
         self.output += s
@@ -108,6 +114,8 @@ def getProcessOutputAndValueWithInput(executable, args, input):
     "similar to getProcessOutputAndValue, but also allows injection of input on stdin"
     d = defer.Deferred()
     p = _SubprocessProtocol(input, d)
+
+    (executable, args) = encodeExecutableAndArgs(executable, args)
     reactor.spawnProcess(p, executable, (executable,) + tuple(args))
     return d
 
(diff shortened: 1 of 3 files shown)


master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
```diff
@@ -12,10 +12,14 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
+from __future__ import unicode_literals
```
sa2ajj: empty line after this one?

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
```diff
@@ -12,10 +12,14 @@
...
 import os
 import re
 import sys
 
+from future.utils import text_type
+
+from buildbot.test.util.misc import _encodeExecutableAndArgs
```
sa2ajj: `buildbot` is more local that `twisted`: change the order?

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
```diff
@@ -107,6 +113,8 @@ def getProcessOutputAndValueWithInput(executable, args, input):
     "similar to getProcessOutputAndValue, but also allows injection of input on stdin"
     d = defer.Deferred()
     p = _SubprocessProtocol(input, d)
+
+    (executable, args) = _encodeExecutableAndArgs(executable, args)
```
sa2ajj: do we need `()` on the left side?

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
```diff
@@ -118,9 +126,11 @@ class TestBuildbotCvsMail(unittest.TestCase):
         skip = ("'%s' does not exist (normal unless run from git)"
                 % buildbot_cvs_mail_path)
 
-    def assertOutputOk(self, xxx_todo_changeme4, regexList):
+    def assertOutputOk(self, result, regexList):
```
sa2ajj: I'm not sure what do we need `result` for: why not just replace it with `(output, code)`?

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
sa2ajj: in addition, i feel really uneasy about importing a private function

Stibbons: so this method is no more private and thus should be renamed without the '_' :)

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
Stibbons: equivalent, no added value

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
tardyp: isort should enforce that.

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
rodrigc: You can't use (output, code) because tuple parameter unpacking is a syntax error in Python 3: https://www.python.org/dev/peps/pep-3113/

tardyp:
isort needs to be fixed

rodrigc:
@tardyp isort is fixed

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
rodrigc: I renamed the method

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
sa2ajj: oh, good to know

/me needs to spend some time on actually looking @ Python 3

master/buildbot/test/unit/test_contrib_buildbot_cvs_mail.py
sa2ajj: i know it's equivalent, for *my* eyes it's a bit lighter...  let's consider that as [nit]

PR #2384, opened 2016-08-29
Change map() to list comprehension

rodrigc:
On Python 2, map() returns a list, and on Python 3 it returns a generator.
We really need a list here.

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -731,7 +731,7 @@ class BuildStep(results.ResultComputingConfigMixin,
         sv = self.build.getWorkerCommandVersion(command, None)
         if sv is None:
             return True
-        if list(map(int, sv.split("."))) < list(map(int, minversion.split("."))):
+        if [int(s) for s in sv.split(".")] < [int(m) for m in minversion.split(".")]:
             return True
         return False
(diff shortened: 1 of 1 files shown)


master/buildbot/process/buildstep.py
```diff
@@ -731,7 +731,7 @@ def workerVersionIsOlderThan(self, command, minversion):
         sv = self.build.getWorkerCommandVersion(command, None)
         if sv is None:
             return True
-        if map(int, sv.split(".")) < map(int, minversion.split(".")):
+        if list(map(int, sv.split("."))) < list(map(int, minversion.split("."))):
```
Stibbons: Isn't it more readable using list comprehension?
```
if [int(s) for s in sv.split(".")] < [int(m) for m in minversion.split(".")]:
```

rodrigc: List comprehension would work fine here.  I don't feel strongly one way or the other.

tardyp: list comprehension is usually much faster (especially on pypy)

rodrigc: OK, I took @Stibbons code and changed it to list comprehension

PR #2386, opened 2016-09-01
master/setup.py: publish buildbot.buildslave

tardyp:
it was unpublished during the worker transition, but is necessary for the backward compatibility

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -152,6 +152,7 @@ setup_args = {
 
     'packages': [
         "buildbot",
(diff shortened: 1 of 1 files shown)


ryansydnor:
Looks like this will take care of http://trac.buildbot.net/ticket/3594. Thanks!

djmitche:
Looks good to me (although I'd prefer reformatting be separated into a different commit, in general)

tardyp:
oups. those reformating are not on purpose

PR #2388, opened 2016-09-03
Remove types.StringTypes which is gone in Python 3

rodrigc:


master/buildbot/steps/source/p4.py
```diff
@@ -231,12 +232,18 @@ def _buildVCCommand(self, doCommand):
         command.extend(doCommand)
 
         def encodeArg(arg):
-            if isinstance(arg, string_types):
+            if isinstance(arg, binary_type):
```
tardyp: I think we should have an utility for this kind of thing in util/__init__.py near ascii2unicode we good have enforce_utf8_string()

rodrigc: I don't see how this function can be moved into a general utility.
The logic in here is pretty specific to this module, such as for a tuple,
encoding only the second tuple argument

tardyp: I was more thinking about the first part which is the reverse for ascii2unicode

I was thinking about writing this as
```
if isinstance(arg, tuple):
     return (arg[0], encodeUTF8(arg[1]), arg[2])
else:
     return encodeUTF8(arg)
```

master/buildbot/plugins/db.py
```diff
@@ -14,10 +14,9 @@
 # Copyright Buildbot Team Members
 #
 # pylint: disable=C0111
-from types import StringTypes
-
 from future.utils import iteritems
 from future.utils import itervalues
+from future.utils import string_types
 from pkg_resources import iter_entry_points
```
sa2ajj: i think an empty line *before* and *after* this line is warranted.

master/buildbot/steps/source/p4.py
```diff
@@ -14,14 +14,14 @@
 # Copyright Buildbot Team Members
 # Portions Copyright 2013 Bad Dog Consulting
 import re
-from types import StringType
 
 from future.utils import string_types
 from twisted.internet import defer
```
sa2ajj: An empty line before this one.

master/buildbot/steps/source/p4.py
```diff
@@ -14,14 +14,14 @@
...
 from twisted.python import log
 
 from buildbot import config
 from buildbot import interfaces
+from buildbot import util
```
sa2ajj: `util` goes at the end of the import list.

rodrigc: I hace updated the diff.
isort disagrees with you on this particular point about where util goes in the import list.

PR #2390, opened 2016-09-04
Port buildbot_worker.test.unit.test_scripts_runner.TestOptions.test_version to Python 3

rodrigc:


worker/buildbot_worker/test/util/misc.py
```diff
@@ -186,7 +188,16 @@ class StdoutAssertionsMixin(object):
...
+        #
+        # sys.stdout is implemented differently
+        # in Python 2 and Python 3, so we need to
+        # override it differently.
+        # In Python 2, sys.stdout is a byte stream.
+        # In Python 3, sys.stdout is a text stream.
+        if PY3:
+            self.stdout = StringIO()
+        else:
+            self.stdout = BytesIO()
```
tardyp: it is weird, as the previous implementation implemented stdout as stringio.

rodrigc: It looks weird, but the old StringIO and cStringIO modules cannot even be imported
in Python 3.  io.StringIO deals strictly with Unicode on Python 2, Python3,
and io.BytesIO deals strictly with bytes on Python 2, Python 3.

sys.stdout / sys.stderr are implemented very differently on Python 3 as I mentioned in the comments,
so in order to keep this code working on both Python 2 and Python 3,
using BytesIO on Python 2, and StringIO on Python 3 is necessary.
For this particular case of overriding sys.stdout, it cannot be avoided.

worker/buildbot_worker/test/util/misc.py
```diff
@@ -23,14 +23,16 @@
 except ImportError:
     # Python 3
     import builtins
-import cStringIO
 import errno
 import os
 import re
 import shutil
 import sys
+from io import BytesIO
```
sa2ajj: is this order OK?

worker/buildbot_worker/test/util/misc.py
```diff
@@ -23,14 +23,16 @@
...
+from io import StringIO
 
 import mock
```
sa2ajj: I'd expect this import to be after the use of `future`

worker/buildbot_worker/test/util/misc.py
```diff
@@ -186,7 +188,16 @@ class StdoutAssertionsMixin(object):
     """
 
     def setUpStdoutAssertions(self):
-        self.stdout = cStringIO.StringIO()
+        #
+        # sys.stdout is implemented differently
+        # in Python 2 and Python 3, so we need to
```
sa2ajj: why this kind of of formatting?  no need to wrap the line here.

worker/buildbot_worker/test/util/misc.py
```diff
@@ -186,7 +188,16 @@ class StdoutAssertionsMixin(object):
...
+        # override it differently.
+        # In Python 2, sys.stdout is a byte stream.
+        # In Python 3, sys.stdout is a text stream.
+        if PY3:
```
sa2ajj: I'd expect this kind of thing to be handled by `future`, does it not handle it well?

rodrigc: **future** doesn't handle this well, because the differences in sys.stdout between Python 2 and Python 3 are quite different and rather tricky.

worker/buildbot_worker/test/util/misc.py
rodrigc: No reason.  Does it really matter?

worker/buildbot_worker/test/util/misc.py
rodrigc: I think this order is OK.  isort accepts it.

worker/buildbot_worker/test/util/misc.py
rodrigc: Why?  isort accepts this.

worker/buildbot_worker/test/util/misc.py
sa2ajj: I'd say it's NOT OK: imports within a "block" must be sorted by the module name.

worker/buildbot_worker/test/util/misc.py
sa2ajj: Well, `mock` offers the functionality that is:

* not related to future (=> an empty line)
* more specific to the content of this file (=> move after imports that are more generic)

worker/buildbot_worker/test/util/misc.py
```diff
@@ -23,14 +23,16 @@
...
 import re
 import shutil
 import sys
+from io import BytesIO
+from io import StringIO
 
 import mock
+from future.utils import PY3
 from future.utils import string_types
 from twisted.python import log
```
sa2ajj: On related note, there should be an empty line before this one.

worker/buildbot_worker/test/util/misc.py
sa2ajj: It does not and it does: a split line for no reason is more difficult to read.

worker/buildbot_worker/test/util/misc.py
sa2ajj: Right. OK.

worker/buildbot_worker/test/util/misc.py
tardyp: I'd say its ok. the from are placed after the imports within the stdlib import group

PR #2391, opened 2016-09-04
Replace Exception.message with str(Exception)

rodrigc:
BaseException.message is deprecated in Python 2.6 and gone in Python 3.

See:
https://docs.python.org/2/whatsnew/2.6.html#deprecations-and-removals


tardyp:
needs to update tests

rodrigc:
@tardyp I updated the patch, and now all the tests pass

tardyp:
thansk for comin back!

PR #2395, opened 2016-09-07
make github hook record even non-distinct commits

martingkelly:
In the github hook, we currently ignore commits that we have seen before. This
causes a bug in the following situation:

- Builds are scheduled for master only
- Commit X is pushed to a dev branch
- Commit X gets recorded as a change, associated with that dev branch. It does
  not trigger a build because the branch is not master.
- Commit X is pushed to master without a merge (it's already cleanly rebased on
  master).
- Commit X is ignored, thus not triggering a build, even though the commit has
  never been built before and it's a new addition to master.

Fix this bug by recording a change every time a commit enters a scheduled
branch, even if we have seen the commit before. Update the tests accordingly.

### DIFF
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -173,11 +173,6 @@ class GitHubEventHandler(object):
             return changes
 
         for commit in payload['commits']:
-            if not commit.get('distinct', True):
-                log.msg('Commit `%s` is a non-distinct commit, ignoring...' %
-                        (commit['id'],))
-                continue
-
             files = []
             for kind in ('added', 'modified', 'removed'):
                 files.extend(commit.get(kind, []))
@@ -196,7 +191,8 @@ class GitHubEventHandler(object):
                 'branch': branch,
                 'revlink': commit['url'],
                 'repository': repo_url,
-                'project': project
+                'project': project,
+                'properties': {'github_distinct': commit.get('distinct', True)}
             }
 
             if callable(self._codebase):
--- a/master/buildbot/test/unit/test_www_hooks_github.py
+++ b/master/buildbot/test/unit/test_www_hooks_github.py
@@ -388,17 +388,45 @@ class TestChangeHookConfiguredWithGitChange(unittest.TestCase):
     def test_git_with_change_json(self):
         self._check_git_with_change(gitJsonPayload)
 
+    # Test that, even with commits not marked as distinct, the changes get
+    # recorded each time we receive the payload. This is important because
+    # without it, commits can get pushed to a non-scheduled branch, get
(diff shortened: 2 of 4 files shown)


tardyp:
Hi @martingkelly 
I also did encounter this issue when force pushing existing dev branch to the master branch.
I think we should still keep the information of distinct commit inside the change event so that it can be filtered via change filter.

We also need a release note as this is a change of behaviour.
The note should explain how to filter distinct commit by using change_filters

martingkelly:
@tardyp, thanks for the feedback; I'm happy to make the changes suggested. Since I'm new to the codebase, could you verify for me exactly where I'd need to make changes? Based on a quick look through, it seems like I'd need to add a *distinct* member to both *changes.py* and *filter.py*, and update the *asDict* and *asText* methods in the *Change* class. Is that correct, and am I missing anything?

I'm also happy to add a release note, but I don't see where in the codebase I would do that. Could you point me in the right direction?

tardyp:
@martingkelly no need to change the core of buildbot, you can use the change property feature.

just add a 'properties': {'github_distinct': commit.get('distinct', True)}

inside the change definition, and then you document that one can use a changefilter like

ChangeFilter(filter_fn=lambda c: c.getProperty('github_distinct'))

in order to keep the old behaviour (I did not test it, would be a good thing to test that one)

http://docs.buildbot.net/latest/manual/cfg-schedulers.html#change-filters

martingkelly:
OK, I've updated the branch with the *github_distinct* property. I tested that it works as expected, and the given *ChangeFilter* works too.

I also made the documentation changes here:
https://github.com/martingkelly/bbdocs/tree/fix-github-branch-push

Once the code changes are merged, I can submit a pull request for that branch, with a reference to the relevant code commit.

tardyp:
@martingkelly bbdocs is for storing the old docs.

There is a folder master/docs in the main repo where you can update the doc
and also a relnotes/index.rst where to update the release notes.

master/buildbot/www/hooks/github.py
```diff
@@ -196,7 +191,10 @@ def _process_change(self, payload, user, repo, repo_url, project):
                 'branch': branch,
                 'revlink': commit['url'],
                 'repository': repo_url,
-                'project': project
+                'project': project,
+                'properties': {
+                                'github_distinct': commit.get('distinct', True)
```
tardyp: I dont think pep8 will be happy. we usually indent only once for this kind of code (instead of aligning with the {

martingkelly: Good point; I will fix that up.

martingkelly:
OK, my latest fixes the PEP8 issue and adds the documentation in the proper place. The rst format is much easier for documentation editing than the raw HTML/CSS!

PR #2400, opened 2016-09-14
Add .bbtravis file for use with buildbot_travis

tardyp:
buildbot travis looks at .bbtravis.yml first and after that .travis.yml

Our buildbot_travis setup does not support yet all that travis is supporting (e.g mysql, pg)
Also bbtravis has some more features as original travis (like doStepIf)
That's why we maintain the conf in a separate file

The goal is to eventually get rid of all travis for our internal CI needs (and eventually circle and appveyor).

Implementing this  shown to some bugs with how we manage BUILBOT_TEST_DB_URL, which are fixed in this PR as well.

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -1,0 +1,188 @@
+# BBTravis CI configuration file
+
+services:
+  - mysql
+  - postgresql
+
+language: python
+
+# Available Python versions:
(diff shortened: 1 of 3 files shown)


tardyp:
I have enabled PR testing with nine.buildbot.net starting with that PR. It is not mandatory yet.
Lets see how it is running

PR #2402, opened 2016-09-14
github: CI should use refs/pull/xx/merge refs to make sure the PR is …

tardyp:
…tested against latest code.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -159,6 +159,9 @@ class Git(Source):
     def startVC(self, branch, revision, patch):
         self.branch = branch or 'HEAD'
         self.revision = revision
+        # ignore the revision if the branch ends with /merge
+        if self.branch.endswith("/merge"):
+            self.revision = None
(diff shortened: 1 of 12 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -159,6 +159,9 @@ def __init__(self, repourl=None, branch='HEAD', mode='incremental', method=None,
     def startVC(self, branch, revision, patch):
         self.branch = branch or 'HEAD'
         self.revision = revision
+        # ignore the revision if the branch ends with /merge
+        if self.branch.endswith("/merge"):
```
sa2ajj: I do not really like it: this seems like GitHub's thingie, while the change is for the common git step.

tardyp: indeed, we should probably create a GitHub step for that matter (like we have a Gerrit step, which does similar thing)

tardyp: @sa2ajj I think I addressed your request. I need your approval to merge as you expressly required changes

PR #2403, opened 2016-09-14
Stop using loop variable outside the loop

seankelly:
The builder variable is from the now-exited loop over self.builders.


master/buildbot/process/botmaster.py
```diff
@@ -110,7 +110,7 @@ def cleanShutdown(self, quickMode=False, stopReactor=True, _reactor=reactor):
                 n += len(builder.building)
             if n > 0:
                 log.msg(
-                    "Not shutting down, builder %s has %i builds running" % (builder, n))
```
tardyp: perhaps we should say

```
if builder.building:
     log.msg(...)
```
in the loop. I guess the info of what builder is building is important

PR #2405, opened 2016-09-15
Call existing disconnect method if there should be no wait

seankelly:
When build_wait_timeout is greater than 0, a timer is set to call _soft_disconnect. That method has some checking and then does a proper disconnect and insubstantiate. Furthermore, within the insubstantiate
method, it already checks if there are builds pending to cause the worker to substantiate again immediately. When build_wait_timeout is 0, the timer doesn't need to be set so call what it would otherwise call. This will ensure the two cases stay in sync.

Fixes trac #3610.

### DIFF
--- a/master/buildbot/worker/latent.py
+++ b/master/buildbot/worker/latent.py
@@ -195,12 +195,7 @@ class AbstractLatentWorker(AbstractWorker):
 
         if not self.building:
             if self.build_wait_timeout == 0:
-                d = self.insubstantiate()
-                # try starting builds for this worker after insubstantiating;
-                # this will cause the worker to re-substantiate immediately if
-                # there are pending build requests.
-                d.addCallback(lambda _:
-                              self.botmaster.maybeStartBuildsForWorker(self.workername))
+                self._soft_disconnect()
             else:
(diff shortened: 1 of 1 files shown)


tardyp:
We need a proper integration test for any changes in that code. We have been accepting too much hacks on that code. I can look at it

seankelly:
I rebased to fix the merge conflict but haven't added tests.

master/buildbot/worker/latent.py
```diff
@@ -195,12 +195,7 @@ def buildFinished(self, sb):
 
         if not self.building:
             if self.build_wait_timeout == 0:
-                d = self.insubstantiate()
-                # try starting builds for this worker after insubstantiating;
-                # this will cause the worker to re-substantiate immediately if
-                # there are pending build requests.
-                d.addCallback(lambda _:
-                              self.botmaster.maybeStartBuildsForWorker(self.workername))
+                self._soft_disconnect()
```
tardyp: I think _soft_disconnect does too many things, this is why I prefer using insubstantiate directly

tardyp:
close in favor of #2420

PR #2408, opened 2016-09-22
Add logEnviron flag to MasterShellCommand

dinatale2:
Added the logEnviron flag to the MasterShellCommand
step. The master may have variables in it's environment
(such as AWS credentials) which we do not want to leak
into a log.

Signed-off-by: Giuseppe Di Natale <dinatale2@llnl.gov>


sa2ajj:
Thank you.

PR #2409, opened 2016-09-22
Fix the tag handling for the "builders" page

starcruiseromega:
This patch fixes an issue where navigating to /builders?tag=something gives "exceptions.AttributeError: 'Builder' object
has no attribute 'matchesAnyTag'". This was because the self.botmaster.builders is a list of Builder objects that don't
have that particular method. The Builder object has a member that is a BuilderStatus object that does have that method
though, and it looks like that is intended to hold information like tags, so calling matchesAnyTag on
bldr.builder_status instead of directly on bldr solves the problem.

### DIFF
--- a/master/buildbot/status/master.py
+++ b/master/buildbot/status/master.py
@@ -241,7 +241,7 @@ class Status(config.ReconfigurableServiceMixin, service.MultiService):
         # respect addition order
         for name in self.botmaster.builderNames:
             bldr = self.getBuilder(name)
-            if bldr.matchesAnyTag(tags):
+            if bldr.builder_status.matchesAnyTag(tags):
                 l.append(name)
         return util.naturalSort(l)
 


starcruiseromega:
I'm not sure what to make of the checks that failed. The AppVeyor one looks like it just didn't run, but the Travis one looks like it ran into an error while installing Twisted. I'm not sure that has anything to do with my change though.

I tested this change by hot patching our internal buildbot. I had some problems trying to set up the environment for common/validate.sh but if that's required let me know and I'll try again.

tardyp:
Hi @starcruiseromega the buildbot eight unit tests have been broken on travis and appveyor.
As buildbot eight belongs to its community. if you can get some time to figure out what is wrong, I think this would help a lot. If not, this is not a problem, eight travis votes are not mandatory.

starcruiseromega:
Thanks @tardyp. I'm not sure I'll have any time to look at the unit tests this week, but I'll keep it in mind in case I do.

seankelly:
This is strange. A previous commit, 869803df, claims to address the same issue. Checking the code, `getBuilder` returns the `builder_status` attribute already.

@starcruiseromega Does your Buildbot instance have 869803df in it? It seems like your commit is cherry-picked from 0.8.12 or earlier, but a different fix was applied to the unreleased eight branch.

starcruiseromega:
Yes, the branch that I based this commit on does have that change. I did try to see whether someone had already fixed the issue between 0.8.12 and the top of tree on eight, but I guess I didn't look hard enough! So this patch isn't necessary then. Thanks!

PR #2410, opened 2016-09-23
more AbstractLatentWorker to its own file

tardyp:
This is easier to work with when it has its own file


seankelly:
LGTM.

sa2ajj:
The coverage result seems to suggest not to.

tardyp:
@sa2ajj my next PR will be about increasing the coverage of that file actually.
I wanted to do this prep first as this is harmless refactoring.

PR #2414, opened 2016-09-27
Implement exponential backoff for latent worker start

tardyp:
when there is an issue on a latent worker backend
buildbot will create tons of "retry" builds.

Here, we will just make sure we wait a bit before retrying

Adding some debugging utils to test_latent, as this shit is damn hard to debug.

### DIFF
--- a/master/buildbot/test/util/misc.py
+++ b/master/buildbot/test/util/misc.py
@@ -12,10 +12,15 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
+from __future__ import print_function
+
 import cStringIO
 import os
 import sys
+
 from future.utils import text_type
+
+import buildbot
 
 
 class PatcherMixin(object):
@@ -71,3 +76,28 @@ def encodeExecutableAndArgs(executable, args, encoding="utf-8"):
         argsBytes.append(arg)
 
     return (executable, argsBytes)
+
+
+def enable_trace(case, trace_exclusions=None, f=sys.stdout):
+    """This function can be called to enable tracing of the execution
+    """
+    if trace_exclusions is None:
+        trace_exclusions = ["twisted", "worker_transition.py", "util/tu",
+                            "log.py", "/mq/", "/db/", "buildbot/data/", "fake/reactor.py"]
+
+    bbbase = os.path.dirname(buildbot.__file__)
+    state = {'indent': 0}
+
+    def tracefunc(frame, event, arg):
+        if frame.f_code.co_filename.startswith(bbbase):
+            if not any(te in frame.f_code.co_filename for te in trace_exclusions):
+                if event == "call":
+                    state['indent'] += 2
+                    print("-" * state['indent'], frame.f_code.co_filename.replace(bbbase, ""),
+                          frame.f_code.co_name, frame.f_code.co_varnames, file=f)
+                if event == "return":
+                    state['indent'] -= 2
+        return tracefunc
+
+    sys.settrace(tracefunc)
+    case.addCleanup(sys.settrace, lambda _a, _b, _c: None)
--- a/master/buildbot/worker/latent.py
+++ b/master/buildbot/worker/latent.py
@@ -23,6 +23,7 @@ from zope.interface import implementer
 
 from buildbot.interfaces import ILatentWorker
 from buildbot.interfaces import LatentWorkerFailedToSubstantiate
+from buildbot.interfaces import LatentWorkerSubstantiatiationCancelled
 from buildbot.util import Notifier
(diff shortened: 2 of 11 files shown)


master/buildbot/worker/latent.py
```diff
@@ -223,7 +246,10 @@ def insubstantiate(self, fast=False):
         self._clearBuildWaitTimer()
         d = self.stop_instance(fast)
         self.substantiated = False
-        yield d
+        try:
+            yield d
+        except Exception as e:
+            log.err(e, "while insubstanciating")
```
seankelly: Typo: insubstantiating.

master/buildbot/worker/latent.py
```diff
@@ -101,27 +104,47 @@ def substantiate(self, sb, build):
...
-        d.addCallbacks(start_instance_result, clean_up)
+        d.addCallback(start_instance_result)
+        d.addErrback(clean_up)
         return d
 
+    def startSubstantiationRetryTimer(self):
+        self.retry_timer = self.master.reactor.callLater(self.retry_timeout, self.retrySubstantiation)
+        # next time we will wait twice
+        # timer reset on substanciation success
+        if self.retry_timeout < self.retry_max_timeout:
```
seankelly: This would allow the timeout to be greater than the max timeout. If it's ever in the range of (retry_max_timeout/2, retry_max_timeout), when doubled it will then be greater than retry_max_timeout. I think it should be doubled and then clamped to the maximum.

master/buildbot/worker/latent.py
```diff
@@ -101,27 +104,47 @@ def substantiate(self, sb, build):
...
             # swallow the failure as it is given to notified
             return None
-        d.addCallbacks(start_instance_result, clean_up)
+        d.addCallback(start_instance_result)
+        d.addErrback(clean_up)
         return d
 
+    def startSubstantiationRetryTimer(self):
+        self.retry_timer = self.master.reactor.callLater(self.retry_timeout, self.retrySubstantiation)
+        # next time we will wait twice
```
seankelly: Should clarify this to: "Next we will wait up to twice as long."

master/buildbot/test/util/misc.py
```diff
@@ -71,3 +76,29 @@ def encodeExecutableAndArgs(executable, args, encoding="utf-8"):
         argsBytes.append(arg)
 
     return (executable, argsBytes)
+
+
+def enable_trace(case, trace_exclusions=None, f=sys.stdout):
+    """This function can be called to enable tracing of the execution
+    TODO put this in an utility module
```
seankelly: Still need the TODO?

master/buildbot/worker/latent.py
```diff
@@ -249,7 +250,11 @@ def insubstantiate(self, fast=False):
         try:
             yield d
         except Exception as e:
-            log.err(e, "while insubstanciating")
+            # The case of failure for insubstanciation is bad as we have a left-over costing ressource
+            # There is not much thing to do here generically, so we must put the problem of stop_instance
+            # reliability to the backend driver
+            log.err(e, "while insubstantiating")
```
rutsky: Maybe `while substantiating`?
Also in comment above + `ressource` -> `resource`.

seankelly: It's in the `insubstantiate` method though.

tardyp: yeah. I think we should change the naming. as a non-native this does not make much sense, and I keep making typo.

@seankelly I dont know if as native it does mean something for you.

tardyp: maybe spawn / kill would be more idiomatic than substantiate/insubstantiate

seankelly: Yes, it does make sense, assuming the process for creating the thing is named "substantiate". It's also an awkward word and sort of requires the destroy process to use a word that doesn't exist.

I would not mind a renaming. OpenStack uses boot and delete, Docker has create/run and kill, while libvirt has create/define and delete. I like create and delete.

master/buildbot/worker/latent.py
```diff
@@ -101,27 +105,49 @@ def substantiate(self, sb, build):
...
             # swallow the failure as it is given to notified
             return None
-        d.addCallbacks(start_instance_result, clean_up)
+        d.addCallback(start_instance_result)
+        d.addErrback(clean_up)
         return d
 
+    def startSubstantiationRetryTimer(self):
+        self.retry_timer = self.master.reactor.callLater(self.retry_timeout, self.retrySubstantiation)
+        log.msg("{}: Failed to substanciate! Will retry in {}s".format(self.name, self.retry_timeout))
```
seankelly: Typo: substantiate.

PR #2415, opened 2016-09-28
remove scripts from the dist

tardyp:
scripts are not portable and deprecated.
We rather use the console_scripts entrypoint which is nice a portable

fixes http://trac.buildbot.net/ticket/3611


worker/buildbot_worker/scripts/windows_service.py
```diff
@@ -46,14 +46,14 @@
...
-#   % python buildbot_service.py restart
+#   % buildbot_worker_windows_service restart
 #
 # Troubleshooting:
 # * Check the Windows event log for any errors.
 # * Check the "twistd.log" file in your buildbot directories - once each
 #   bot has been started it just writes to this log as normal.
 # * Try executing:
-#   % python buildbot_service.py debug
+#   % buildbot_windows_service debug
```
rutsky: Maybe  `buildbot_worker_windows_service`?
BTW, is `debug` is still handled by worker?..

tardyp: indeed.
No idea if this is working actually. this has to be tested by windows community.

PR #2420, opened 2016-09-29
force latent worker to restart if build_wait_timeout is 0

tardyp:
one commit based on #2414

build_wait_timeout = 0 kind of worked in real reactor, but sometimes some build start, and then are stopped right after

This is because of a deferred race condition which is not happening on the fake reactor

baseWorker.buildFinished will start the process of searching new build
but latentWorker.buildFinished will insubstantiate in parrallel
depending on how fast the db will return the unclaimed buildrequest, a new build can start before the insubstantiation.

by calling baseWorker.buildFinished after doing the insubstantiate, we make sure that this worker will not be selected

This commit also contains test for this case, and fix for LocalSlave to correctly handle its lifecycle.

### DIFF
--- a/master/buildbot/worker/latent.py
+++ b/master/buildbot/worker/latent.py
@@ -196,21 +196,20 @@ class AbstractLatentWorker(AbstractWorker):
         return AbstractWorker.canStartBuild(self)
 
     def buildStarted(self, sb):
+        assert sb.isBusy()
         self._clearBuildWaitTimer()
 
     def buildFinished(self, sb):
-        AbstractWorker.buildFinished(self, sb)
-
+        assert not sb.isBusy()
         if not self.building:
             if self.build_wait_timeout == 0:
-                d = self.insubstantiate()
-                # try starting builds for this worker after insubstantiating;
-                # this will cause the worker to re-substantiate immediately if
-                # there are pending build requests.
-                d.addCallback(lambda _:
-                              self.botmaster.maybeStartBuildsForWorker(self.workername))
+                self.insubstantiate()
+                # insubstantiate will automatically retry to create build for this worker
             else:
                 self._setBuildWaitTimer()
+
+        # AbstractWorker.buildFinished() will try to start the next build for that worker
+        AbstractWorker.buildFinished(self, sb)
 
     def _clearBuildWaitTimer(self):
         if self.build_wait_timer is not None:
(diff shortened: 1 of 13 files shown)


master/buildbot/worker/latent.py
```diff
@@ -183,27 +209,26 @@ def _substantiation_failed(self, failure):
...
+        assert not sb.isBusy()
         if not self.building:
             if self.build_wait_timeout == 0:
-                d = self.insubstantiate()
-                # try starting builds for this worker after insubstantiating;
-                # this will cause the worker to re-substantiate immediately if
-                # there are pending build requests.
-                d.addCallback(lambda _:
-                              self.botmaster.maybeStartBuildsForWorker(self.workername))
+                self.insubstantiate()
```
seankelly: Calling this method means the instance is stopped but does the master update the state of the worker to DETACHED? I'm not sure this will fix the issue of a worker disconnecting and immediately being available to be created again.

sa2ajj:
LGTM

master/buildbot/worker/latent.py
```diff
@@ -183,27 +209,26 @@ def _substantiation_failed(self, failure):
...
+                # insubstantiate will automatically retry to create build for this worker
             else:
                 self._setBuildWaitTimer()
 
+        # AbstractWorker.buildFinished() will try to start the next build for that worker
+        AbstractWorker.buildFinished(self, sb)
```
seankelly: This seems redundant. The `buildFinished` method calls `self.botmaster.maybeStartBuildsForWorker()`, but the `insubstantiate` method calls the same method at the end.

master/buildbot/worker/latent.py
seankelly: I should clarify that the worker needs to immediately be available and at the moment it isn't.

tardyp: The state DETACHED is for the workerforbuilder. Its weird that the state is managed here, as actually all the wfd for the same worker are supposed to be in the same state.

So actually insubstanciate will not move the wbd to detached state. The  worker will not take new build during insubstanciation because canStartBuild will return false whel self.insubstanciating = True

The wfb state is DETACHED during substantiation, though. I can't remember exactly the reason why I had to do that.

This is part of the cleanups that needs to be done once this thing is fully tested at integration level.

I think there are still lots of simplification of the state.

notably I think we shouldn't have a state for wfb. They are several boolean state which makes sense, and combine into a canStartBuild()

the new canStartBuild for a latent wfb would be:

if self.running -> false
if self.worker.deleting -> false
if self.worker.creating -> true
else -> true

the new canStartBuild for a normal wfb would be:

if self.running -> false
if not self.attached -> false
else -> true

what do you think @seankelly , @tomprince ?

master/buildbot/worker/latent.py
tardyp: Agreed, but for now buildFinished is only about starting another build for that builder, but it could do other thing.

seankelly:
I'm going to remove the merge me label because I don't want this PR accidentally merged before 2414 is.

tardyp:
merging so that we can move forward on metabuildbot. Let me know if you have anymore issue with the review

PR #2427, opened 2016-10-03
Replace execfile() with exec()

rodrigc:
execfile() is gone in Python 3.
See:
http://python3porting.com/differences.html#execfile


tardyp:
thanks!

PR #2429, opened 2016-10-03
Eliminate StringIO.StringIO and cStringIO.StringIO which are not available in Python 3

rodrigc:


master/buildbot/www/hooks/github.py
```diff
@@ -73,7 +74,13 @@ def _get_payload(self, request):
             if hash_type != 'sha1':
                 raise ValueError('Unknown hash type: %s' % (hash_type,))
 
-            mac = hmac.new(self._secret, msg=content, digestmod=sha1)
+            _secret = self._secret
+            if isinstance(_secret, text_type):
```
tardyp: we really need an utility function for this kind of code.
One simple line has to be replaced by 6 lines here. we need something lighter.

rodrigc: What kind of utility function do you want?

master/buildbot/db/migrate_utils.py
```diff
@@ -15,8 +15,8 @@
 
 
 import sqlalchemy as sa
-
 from future.utils import text_type
```
sa2ajj: I'd put `future.utils` *before* `sqlalchemy` and definitely would not put them in the same group.

master/buildbot/plugins/db.py
```diff
@@ -14,10 +14,10 @@
 # Copyright Buildbot Team Members
 #
 # pylint: disable=C0111
-from types import StringTypes
 
 from future.utils import iteritems
 from future.utils import itervalues
+from future.utils import string_types
 from pkg_resources import iter_entry_points
```
sa2ajj: Please add an empty line before and after this one.  (Since you are on it)

master/buildbot/process/build.py
```diff
@@ -17,6 +17,7 @@
 
 from functools import reduce
 
+from future.utils import string_types
```
sa2ajj: `future.utils` is not part of `twisted`: please add an empty line afterwards.

master/buildbot/process/build.py
```diff
@@ -224,9 +225,13 @@ def setupWorkerForBuilder(self, workerforbuilder):
         worker_properties = workerforbuilder.worker.properties
         self.getProperties().updateFromProperties(worker_properties)
         if workerforbuilder.worker.worker_basedir:
-            builddir = self.path_module.join(
-                workerforbuilder.worker.worker_basedir,
-                self.builder.config.workerbuilddir)
+            worker_basedir = workerforbuilder.worker.worker_basedir
+            if not isinstance(worker_basedir, string_types):
+                worker_basedir = worker_basedir.decode("utf-8")
+            workerbuilddir = self.builder.config.workerbuilddir
```
sa2ajj: please add an empty line before this one

master/buildbot/process/build.py
```diff
@@ -224,9 +225,13 @@ def setupWorkerForBuilder(self, workerforbuilder):
...
+            if not isinstance(workerbuilddir, string_types):
+                workerbuilddir = workerbuilddir.decode("utf-8")
+            builddir = self.path_module.join(worker_basedir, workerbuilddir)
```
sa2ajj: please add an empty line before this one

master/buildbot/process/builder.py
```diff
@@ -16,7 +16,6 @@
 import weakref
 
 from future.utils import string_types
-
```
sa2ajj: why did you remove this empty line?

master/buildbot/process/buildrequestdistributor.py
```diff
@@ -14,8 +14,8 @@
 # Copyright Buildbot Team Members
 import random
 from datetime import datetime
-from dateutil.tz import tzutc
 
+from dateutil.tz import tzutc
```
sa2ajj: `dateutil.tz` belongs to the previous group of imports.

master/buildbot/process/remotetransfer.py
```diff
@@ -16,14 +16,11 @@
 import os
 import tarfile
 import tempfile
+from io import BytesIO
```
sa2ajj: I wonder if import checker complains about this order...

master/buildbot/reporters/mail.py
```diff
@@ -43,6 +43,11 @@
 from buildbot.reporters.message import MessageFormatter as DefaultMessageFormatter
 from buildbot.util import service
 
+if PY3:
```
sa2ajj: Do we really need to repeat this sequence every time?

master/buildbot/steps/subunit.py
```diff
@@ -12,14 +12,20 @@
...
 from unittest import TestResult
 
+from future.utils import PY3
+
 from buildbot.process import logobserver
 from buildbot.process.results import FAILURE
 from buildbot.process.results import SUCCESS
 from buildbot.steps.shell import ShellCommand
 
+if PY3:
```
sa2ajj: ditto

master/buildbot/test/fake/web.py
```diff
@@ -12,12 +12,16 @@
...
 #
 # Copyright Buildbot Team Members
-from StringIO import StringIO
-
+from future.utils import PY3
 from mock import Mock
 from twisted.internet import defer
 from twisted.web import server
 
+if PY3:
```
sa2ajj: ditto

master/buildbot/test/fake/web.py
```diff
@@ -12,12 +12,16 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-from StringIO import StringIO
-
+from future.utils import PY3
 from mock import Mock
 from twisted.internet import defer
```
sa2ajj: Please add an empty line before this one.

master/buildbot/test/integration/test_custom_buildstep.py
```diff
@@ -12,9 +12,8 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-from StringIO import StringIO
-
 import mock
+from future.utils import PY3
```
sa2ajj: Please add an empty line before this one.

My suggestion is to move `import mock` at the very end...

master/buildbot/test/unit/test_changes_manager.py
```diff
@@ -13,7 +13,6 @@
 #
 # Copyright Buildbot Team Members
 import mock
-
```
sa2ajj: why??

master/buildbot/test/unit/test_changes_manager.py
```diff
@@ -13,7 +13,6 @@
...
 from future.builtins import range
 from twisted.trial import unittest
```
sa2ajj: empty line before this one.

master/buildbot/test/unit/test_db_changes.py
```diff
@@ -13,7 +13,6 @@
 #
 # Copyright Buildbot Team Members
 import sqlalchemy as sa
-
```
sa2ajj: why?

master/buildbot/test/unit/test_db_changes.py
```diff
@@ -13,7 +13,6 @@
...
 from future.builtins import range
 from future.utils import iteritems
 from twisted.internet import defer
```
sa2ajj: empty line before this one.

master/buildbot/test/unit/test_db_workers.py
```diff
@@ -13,7 +13,6 @@
 #
 # Copyright Buildbot Team Members
 import mock
-
```
sa2ajj: why?

master/buildbot/reporters/mail.py
tardyp: http://python-future.org/compatible_idioms.html#stringio

The recommendation of future author is to use

    from io import StringIO

master/buildbot/test/unit/test_db_changes.py
tardyp: probably because of isort said so.

master/buildbot/process/build.py
tardyp: we need to stick to what is automated with isort.
I would be incline to put future as part of the stdlib group. So we would need to add it to the config in the force stdlib section

sa2ajj: This sounds like a proper approach.

master/buildbot/reporters/mail.py
sa2ajj:
```
from io import StringIO
```
looks like the way for me...

So there's no need to repeat those conditions on Python version.

master/buildbot/test/unit/test_db_changes.py
sa2ajj: then we really need to look at `isort`'s configuration.

sa2ajj:
I don't work to waste the time either: if it's `isort`'s recommendation, then let's fix its configuration.

master/buildbot/reporters/mail.py
rodrigc: StringIO doesn't work the same on Python 2 and Python 3, since in Python 2, strings are bytes, and in Python 3 strings are unicode.
If I follow your advice, the code will break on Python 2, and clients which use the code will need to change to use.

That's why I used the technique I did.


While it is useful to look at the advice from the future author,
you also need to understand the code and do the right thing, and not blindly
follow advice.

rodrigc:
I updated the patch and used the latest isort configuration.

master/buildbot/reporters/mail.py
rodrigc: One other aspect of the code is that StringIO is used to interact with sys.stdin.
On Python 2, sys.stdin deals with bytes, and on Python 3, sys.stdin deals with unicode.

So the conditional statements I implemented are needed for the code to work properly on Python 2 and Python 3.

master/buildbot/www/hooks/github.py
sa2ajj: probably something like this:
```
def as_utf8_bytes(value):
    if isinstance(value, text_types):
        value = value.encode('utf-8')
    return value
```

one might add a check that otherwise it's of correct type.

master/buildbot/reporters/mail.py
tardyp: If String means different thing in both python, I would suggest not to use that name.

from buildbot.util import BytesIO
from buildbot.util import UnicodeIO

This is for me the most unambiguous way of dealing with the problem.
Sorry to be a bit tough but I think those bytes vs string vs unicode are very error prone when a bunch of python2 coders want to write py3 compatible code.

master/buildbot/www/hooks/github.py
tardyp: There is already ascii2unicode here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/__init__.py#L178

for coherency, I would call it

    def unicode2utf8(value):
        if isinstance(value, text_types):
            value = value.encode('utf-8')
        return value

and put it just after ascii2unicode

master/buildbot/reporters/mail.py
rodrigc: @tardyp I requested clarification of your feedback, and you did not provide it.
Please provide a clear example of what you want, so I can move forward with this patch.

This is a major blocker in terms of moving the buildbot code forward with Python 3.

Thanks.

tardyp: Sorry, I was sure I did provide feedback on this.

The thing is that I am not sure what is the solution to this without actually spending lot of time trying the different options and eventually doing the work myself.

Doing import StringIO as BytesIO does not sound right, that I know for sure.

The current io.StringIO and io.BytesIO are not behaving the same in py2 and py3, and I think this is the core of the problem.

I dont think there is a nice solution to this problem.

So my input would be:
- stick with the future suggestion:
   only do from io import StringIO and from io import BytesIO.
- use BytesIO when we want bytes as *output*
- use StringIO when we want unicode as *output*
- pass the StringIO input through buildbot.util.ascii2unicode
- pass the BytesIO input through buildbot.util.as_utf8_bytes (to be created as @sa2ajj suggested)

tardyp: We can talk about it live in the buildbot meeting at on irc at 4h30PM UTC

tardyp: @edschofield Someone suggested on IRC to pull you in this discussion as a python3 transition expert. Would you like to advise us on this topic?

sa2ajj: just to make it a bit more specific (it did not get to our [meetings minutes](https://supybot.buildbot.net/meetings/buildbot/2016/buildbot.2016-11-15-16.29.html):

```
> 19:11 <@djmitche> https://github.com/edschofield maybe?
```

edschofield:
The approach @tardyp described looks like the right one to me:

from io import BytesIO    # for byte-strings
from io import StringIO    # for Unicode text

These do behave consistently on Py2 and Py3. This is one area of compatibility that the Python core revs did get right with the transition. (Even better would have been to call it TextIO and call the new data type "text".)

Where the interface between the code and the "real world" differs between Py2 and Py3 (eg sys.stdin), the code would still require "if PY2:" or "if PY3:" clauses. If this is happening repeatedly, I would wrap this into a utility function: get_sys_stdin_as_utf8_bytes() ... and call this everywhere instead of referring to sys.stdin directly.

master/buildbot/reporters/mail.py
rodrigc: @tardyp sorry I am not able to attend the buildbot meeting.

The places where I used StringIO all interact with sys.stdin and sys.stdio.
Most of these instances are located in tests.
These are implemented differently on Python 2 and Python 3, where on Python 2
they are bytes, and and Python 3, there are unicode.

As @edschofield has [pointed out](https://github.com/buildbot/buildbot/pull/2429#issuecomment-260717355)
So, some logic around "if PY3" *is* required to get these tests working on Python 2 and 3,
since sys.stdin and sys.stdout are different on Python 2 and Python 3.

What I have done does work on Python 2 and Python 3.  If you don't like that approach
what do you want to do instead?

rodrigc:
Moving to other PR's

tardyp:
thanks. The split helped a lot I think.

PR #2430, opened 2016-10-03
Replace urllib2 with future.moves.urllib.

rodrigc:
urllib2 is gone in Python 3.

### DIFF
--- a/master/buildbot/buildbot_net_usage_data.py
+++ b/master/buildbot/buildbot_net_usage_data.py
@@ -15,19 +15,19 @@
 
 """
 This files implement buildbotNetUsageData options
-It uses urllib2 instead of requests in order to avoid requiring another dependency for statistics feature.
-urllib2 supports http_proxy already urllib2 is blocking and thus everything is done from a thread.
+It uses urllib instead of requests in order to avoid requiring another dependency for statistics feature.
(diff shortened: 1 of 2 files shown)


master/buildbot/buildbot_net_usage_data.py
```diff
@@ -15,15 +15,16 @@
 
 """
 This files implement buildbotNetUsageData options
-It uses urllib2 instead of requests in order to avoid requiring another dependency for statistics feature.
-urllib2 supports http_proxy already urllib2 is blocking and thus everything is done from a thread.
+It uses urllib instead of requests in order to avoid requiring another dependency for statistics feature.
+urllib supports http_proxy already urllib is blocking and thus everything is done from a thread.
```
tardyp: did you test this?

I am effraid by this from future doc:
http://python-future.org/imports.html


The newly exposed urllib submodules are backports of those from Py3.x. This means, for example, that urllib.parse.unquote() now exists and takes an optional encoding argument on Py2.x as it does on Py3.x.

Limitation: Note that the http-based backports do not currently support HTTPS (as of 2015-09-11) because the SSL support changed considerably in Python 3.x. If you need HTTPS support, please use this idiom for now:

from future.moves.urllib.request import urlopen

rodrigc:
Replaced with https://github.com/tardyp/buildbot/pull/12

PR #2431, opened 2016-10-03
REVERTME: force sdist for pycparser

tardyp:
revert when this is fixed https://github.com/eliben/pycparser/issues/148


tomprince:
The linked issues is closed as fixed.

PR #2433, opened 2016-10-03
Fix required access for GitLabStatusPush doc

Pluggi:
Buildbot's doc says that GitLab requires reporter access to a repository in order to report build statuses, but this is wrong as you can see here : https://docs.gitlab.com/ee/integration/jenkins.html#configure-gitlab-users

> Developer permission is required for reporting build status.


cmouse:
"This user will need to be a global Admin or added as a member to each Group/Project"

cmouse:
It has changed in subsequent gitlab release. Reporter was enough. Is not anymore.

tardyp:
thanks!

PR #2434, opened 2016-10-03
Add a few more filters to MailNotifier

martingkelly:
This patchset contains a combination of minor changes (typo fixes, PEP8 cleanup) and the addition of two new filters for MailNotifier: schedulers and branches. These new filters let you send mail only if a given build was generated by a given set of schedulers, or a given set of branches. An example of when this is useful is when you want to send mail for builds triggered by the "checkin" scheduler, or the official nightly, but not for unofficial builds triggered by developers.

### DIFF
--- a/master/buildbot/reporters/mail.py
+++ b/master/buildbot/reporters/mail.py
@@ -111,8 +111,7 @@ class MailNotifier(service.BuildbotService):
                     messageFormatter=None, extraHeaders=None,
                     addPatch=True, useTls=False,
                     smtpUser=None, smtpPassword=None, smtpPort=25,
-                    name=None
-                    ):
+                    name=None):
         if ESMTPSenderFactory is None:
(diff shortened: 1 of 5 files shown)


master/buildbot/reporters/mail.py
```diff
@@ -269,9 +273,15 @@ def matchesAnyTag(self, tags):
     def isMailNeeded(self, build):
         # here is where we actually do something.
         builder = build['builder']
+        scheduler = build['properties']['scheduler'][0]
```
tardyp: I dont think that we are garanteed to have scheduler and branch properties.
please use 

```
build['properties'].get('scheduler',[None])[0]
```
and same for branch

martingkelly: Thanks, fixed. I also added release notes for both changes.

PR #2438, opened 2016-10-05
update dockerfiles according to best practices

tardyp:


master/Dockerfile.ubuntu
```diff
@@ -10,19 +12,29 @@ MAINTAINER  Buildbot maintainers
 # that everything is rebuilt
 ENV         security_updates_as_of 2016-07-30
 
+# This will make apt-get install without question
+ENV         DEBIAN_FRONTEND noninteractive
```
rutsky: I believe this is against Docker best practices, see https://github.com/docker/docker/issues/4032 and https://github.com/docker/docker/pull/4035

tardyp: Thanks, I wasn't aware of that.

but... I disagree. The only argument is that when you do docker exec -it bash, then you wont have interactive debian stuff anymore.

Those containers are not meant for interactive (who does?), and I think it is better to put that environment rather than force setting it inside each RUN command

There are some packages that wont install if you dont put this variable as they insist to ask question. This is just lost time to try and debug those.

the discussion is from 2014. they may have change their mind since, and they dont mention it in https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

rutsky: AFAIK in ideal world `DEBIAN_FRONTEND=noninteractive` shouldn't be specified at all: apt/dpkg and their configuration in base image should be smart enough to handle this case properly. But, AFAIK too, the process of fixing this is quite slow.

Though `docker exec -ti bash` + dpkg/apt is rarely needed with BB images it might be used, for example, for debugging.
And instead of introducing potentially new issue by workarounding original issue, I would prefer to limit scope of the workaround to image build process.

I agree with @thaJeztah conclusion https://github.com/docker/docker/issues/4032#issuecomment-163689851:

> @drastik best practice is to not use ENV, because that will make the env-var persist in the final image.

> In most cases, omitting "DEBIAN_FRONTEND" is harmless and only produces some warnings during build (which can be ignored). If really needed, RUN DEBIAN_FRONTEND=noninteractive apt-get install -y -q package should work

In the same bug report @markriggins suggested feasible alternative to specifying `DEBIAN_FRONTEND` as env variable https://github.com/docker/docker/issues/4032#issuecomment-192327844: 

> The new ARG directive in the Dockerfile can set this only during the build

> ARG DEBIAN_FRONTEND=noninteractive

I suggest to stick with `ARG DEBIAN_FRONTEND=noninteractive` solution (though personally I prefer to specify `DEBIAN_FRONTEND=noninteractive` only otherwise it doesn't work at all, and mostly it works).

tardyp: I like ARG if it works. I'll try that tomorrow

tardyp:
@rutsky I did a big amount of changes on that PR. I would appreciate that you give it a look if you can

- it now downloads its config from http server
- I updated the smoke test to be able to test the dockerfile locally (and eventually on the CI)

worker/Dockerfile
```diff
@@ -1,43 +1,50 @@
...
 RUN         apt-get update && \
             apt-get -y upgrade && \
-            apt-get -y install -q \
-            build-essential \
-            git \
-            subversion \
-            python-dev libffi-dev libssl-dev python-pip python-virtualenv curl && \
-            apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
+            apt-get -y install -q  -o APT::Get::Install-Recommends="false"  -o APT::Get::Install-Suggests="false" \
```
rutsky: @tardyp are you sure that `APT::Get::Install-Recommends="false"` is a proper key?

```
# docker run -ti --rm ubuntu:16.04 bash -c 'apt-config dump | grep Recommends'
APT::Install-Recommends "1";
# docker run -ti --rm ubuntu:16.04 bash -c 'apt-config dump | grep Suggests'
APT::Install-Suggests "0";
APT::AutoRemove::SuggestsImportant "false";
```

Looks like should be `APT::Install-Recommends`/`APT::Install-Suggests`. Also `APT::Install-Suggests=false` is not required, since this is a default for a base image.

And instead of `APT::Install-Recommends` it's better to use explicit `--no-install-recommends` `apt-get`option.

PR #2443, opened 2016-10-11
change isort configuration

tardyp:
move future library on top of the imports
put twisted+zope+autobahn appart from the other 3rd party

We discussed on the stringio PR that we would put the future at the same place as the stdlib.
I wasn't able to do that, isort will always put the __future__ import at the same place as the future imports.
I think actually it is better to put future at the same place as __future__ (__future__ is required to be always first by python anyway)

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,6 +1,9 @@
 [settings]
 line_length=110
-known_third_party=twisted,migrate,sqlalchemy,ldap3,klein,future,requests,MySQLdb,zope,coverage,autobahn,pkg_resources,jinja2,mock,dateutil,sphinx,setuptools
+known_future_library=future
+known_twisted=twisted,zope,autobahn,klein
+known_mock=mock
+known_third_party=migrate,sqlalchemy,ldap3,txrequests,requests,MySQLdb,coverage,jinja2,dateutil,sphinx,setuptools
 known_first_party=buildbot,buildbot_worker
 force_single_line=1
-
+sections=FUTURE,FUTURE_LIBRARY,STDLIB,THIRDPARTY,MOCK,TWISTED,FIRSTPARTY,LOCALFOLDER
--- a/master/buildbot/changes/changes.py
+++ b/master/buildbot/changes/changes.py
@@ -12,10 +12,11 @@
 # Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
 # Copyright Buildbot Team Members
-import time
-
 from future.utils import iteritems
 from future.utils import text_type
+
+import time
+
 from twisted.internet import defer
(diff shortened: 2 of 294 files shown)


.isort.cfg
```diff
@@ -1,6 +1,9 @@
 [settings]
 line_length=110
-known_third_party=twisted,migrate,sqlalchemy,ldap3,klein,future,requests,MySQLdb,zope,coverage,autobahn,pkg_resources,jinja2,mock,dateutil,sphinx,setuptools
+known_future_library=future
+known_twisted=twisted,zope,autobahn,klein
+known_third_party=migrate,sqlalchemy,ldap3,txrequests,requests,MySQLdb,coverage,jinja2,mock,dateutil,sphinx,setuptools
 known_first_party=buildbot,buildbot_worker
 force_single_line=1
+sections=FUTURE,FUTURE_LIBRARY,STDLIB,TWISTED,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
```
sa2ajj: I'd change the order of `TWISTED` and `THIRDPARTY`

tardyp: ok

.isort.cfg
```diff
@@ -1,6 +1,9 @@
 [settings]
 line_length=110
-known_third_party=twisted,migrate,sqlalchemy,ldap3,klein,future,requests,MySQLdb,zope,coverage,autobahn,pkg_resources,jinja2,mock,dateutil,sphinx,setuptools
+known_future_library=future
+known_twisted=twisted,zope,autobahn,klein
+known_third_party=migrate,sqlalchemy,ldap3,txrequests,requests,MySQLdb,coverage,jinja2,mock,dateutil,sphinx,setuptools
```
sa2ajj: I'd put `mock` into a separate group and place it *after* `_thirdparty`.

After that I am *happy*! :smile:

.isort.cfg
```diff
@@ -1,6 +1,9 @@
 [settings]
 line_length=110
-known_third_party=twisted,migrate,sqlalchemy,ldap3,klein,future,requests,MySQLdb,zope,coverage,autobahn,pkg_resources,jinja2,mock,dateutil,sphinx,setuptools
+known_future_library=future
+known_twisted=twisted,zope,autobahn,klein
```
sa2ajj: [nit] zope, autobah, klein, twisted

sa2ajj: aha, thanks for the reminder.

then order of 3rd party and twisted + mock after 3rd party are the only things standing between me and +2.

tardyp: done

sa2ajj:
@tardyp thanks again

PR #2444, opened 2016-10-11
Singleton service

tardyp:
This class implements a generic Service that needs to be instantiated as a singleton according to its parameters.
It is a common use case to need this for accessing remote services.
Having a singleton to access a remote allows to limit the number of simultaneous access to the same services.
Thus, several completely independent buildbot services can use that singleton to access the service, and automatically synchronize themselves to not overwhelm it.

PR #2446, opened 2016-10-12
Http client service

tardyp:
based on #2444

### DIFF
--- a/master/docs/developer/utils.rst
+++ b/master/docs/developer/utils.rst
@@ -1040,3 +1040,101 @@ For example, a particular daily scheduler could be configured on multiple master
         c['services'] = [
             MyService(arg1=1)
         ]
+
+:py:mod:`buildbot.util.httpclientservice`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. py:class:: HTTPClientService
+
+    This class implements a SharedService for doing http client access.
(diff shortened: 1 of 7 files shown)


master/docs/developer/utils.rst
```diff
@@ -1040,3 +1040,110 @@ For example, a particular daily scheduler could be configured on multiple master
...
+        :param params: optional dictionary that will be encoded in the query part of the url (e.g. ``?param1=foo``)
+        :returns: implementation of :`IHTTPResponse` via deferred
+
+        Performs a HTTP ``PUT``
+
+        .. note::
+
+            json and data cannot be used at the same time.
+
+.. py:class:: IHTTPResponse
```
sa2ajj: what is this class?  I do not see any references in the code...

tardyp: Its only for documentation.
As the real response is returned by treq, there is no mean to make treq's Response implement this interface.

sa2ajj: I think it's worth mentioning where this interface is coming from.

tardyp: It is already mentioned:

        The api is preferably a subset of the `treq`_ API, so that `treq`_ can be implemented as passthrough.

I'll try to rephrase to make it clearer

tardyp: I added IHTTPResponse to the code, and added some documentation for the fakes. Let me know if this is clear enough.

PR #2449, opened 2016-10-13
put more stuff in the buildbot_travis

tardyp:
goal is to remove travis in the end

probably the first try will fail


Stibbons:
do you plan to have the deploy as well ? That's very cool in travis :p

PR #2450, opened 2016-10-17
Sort tags in natural order

### LINKED ISSUE #3609: Worker port change requires master restart
Hi,

I just noticed that if I have a master waiting for connections at `16124` (for example):
```
c['protocols'] = {'pb': {'port': 16124}}
```
Then start a master and a worker connecting to the master at `9989`, as expected the worker won't be able to find the master since the master is listening to `16124`. However, I changed the above to:
```
```

and then did a:
```
$ buildbot reconfig
```

It says everything went well but the worker still didn't connect. I had to issue a 
```
$ buildbot restart
```

I can't find this mentioned anywhere. Either it's a bug or we should document this.

This is with buildbot 0.9.11 on python 3.6.2.

ryandesign:
Fixes #3609

### DIFF
--- a/master/buildbot/status/web/console.py
+++ b/master/buildbot/status/web/console.py
@@ -316,7 +316,7 @@ class ConsoleStatusResource(HtmlResource):
         for tag in builderList:
             count += len(builderList[tag])
 
-        tags = sorted(builderList.keys())
+        tags = util.naturalSort(builderList.keys())
 
         cs = []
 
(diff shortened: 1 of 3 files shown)


JustAMan:
Code looks easy enough, so if everything works the code is good.
I don't see added "import util" in two of the files, but maybe the imports were already there.

I'd look into test failures if I were you, though. I'm not sure if those are expected to pass as I had not  contributed for quite a while, but still worth looking at.

ryandesign:
Yes, it works; we are using it in production at https://build.macports.org

I don't know anything about your test suite, sorry.

sa2ajj:
thank you, @ryandesign

PR #2452, opened 2016-10-19
Fix the documentation about the HttpStatusPush

benallard:
user and password are actually mandatory parameters.

As shortly said in IRC, I think it would make more sense to have an 'auth' parameter there, that would allow us to offer other types of authentication tham simply BasicAuth, and that would allow us to set the default to None.

### DIFF
--- a/master/docs/manual/cfg-reporters.rst
+++ b/master/docs/manual/cfg-reporters.rst
@@ -766,7 +766,7 @@ HttpStatusPush
 ::
 
     from buildbot.plugins import reporters
-    sp = reporters.HttpStatusPush(serverUrl="http://example.com/submit")
+    sp = reporters.HttpStatusPush(serverUrl="http://example.com/submit", user="jdoe", password="S3cre1")
     c['services'].append(sp)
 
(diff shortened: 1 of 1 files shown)


tardyp:
I think we should rather fix the code to not mandate it.

benallard:
I agree with you that it would make more sense. Unfortunately, requests is not documenting the behavior of an ``auth`` tuple with ``None`` values. I fear that behavior might change unnoticed ... Which would really weight in favor of using an ``auth`` parameter directly ...

tardyp:
For me HttpStatusPush is more meant as a very simple example of a class that is meant t be copy pasted and customized as per the user needs

See the code, its 18loc:
https://github.com/tardyp/buildbot/blob/master/master/buildbot/reporters/http.py#L97
The fix to make auth optional is pretty obvious.

So for me the real API is HttpStatusPushBase

I am working on this code right now so that reporters can optionally use treq instead of txrequests
see #2446 . I will fix this bug as part of this task.

Also this bug was reported yesterday http://trac.buildbot.net/ticket/3625

benallard:
Alright, I'm closing this PR then.

PR #2453, opened 2016-10-19
Small improvements to the HttpStatusPush Reporter

benallard:
* Add an auth parameter
 * Deprecate the user/password one
 * Add a format_fn parameter


master/buildbot/reporters/http.py
```diff
@@ -97,17 +97,28 @@ def send(self, build):
...
         HttpStatusPushBase.checkConfig(self, **kwargs)
 
-    def reconfigService(self, serverUrl, user, password, **kwargs):
+    def reconfigService(self, serverUrl, user=None, password=None, auth=None, format_fn=None, **kwargs):
         HttpStatusPushBase.reconfigService(self, **kwargs)
         self.serverUrl = serverUrl
-        self.auth = (user, password)
+        self.auth = auth
+        if user is not None:
+            self.auth = (user, password)
```
sa2ajj: a slightly cleaner version would be:
```
if user is not None:
    ...
else:
    self.auth = auth
```

master/buildbot/reporters/http.py
```diff
@@ -97,17 +97,28 @@ def send(self, build):
...
+        self.format_fn = format_fn
+        if format_fn is None:
```
sa2ajj: same here.

master/buildbot/reporters/http.py
benallard: One can argue about the "cleaness" of both. The version I wrote made it clear that ``user is not None`` is a special case, and the rest the normal case. In your version, ``user is not None`` is the very first check, and the rest is a fallback, which convey a slightly different meaning. Additionally, from python zen: _Flat is better than nested._.

I'll fix it anyway.

benallard:
... and rebased.

tardyp:
:+1:

tardyp:
I'll let you @sa2ajj merge if you are happy with the changes

PR #2454, opened 2016-10-19
Fixes for Hyper disconnect from the other end

tardyp:
Trying to fix issues we had last week on metabuildbot_travis, I had to fix several corner case issues on the core.

Each issue is fixed in a separate commit, while the integration tests I used to show them is in the last commit

### DIFF
--- a/master/buildbot/test/util/integration.py
+++ b/master/buildbot/test/util/integration.py
@@ -184,7 +184,6 @@ class RunMasterBase(unittest.TestCase):
 
         # wait until we receive the build finished event
         buildrequest = yield d
-        print(buildrequest)
         builds = yield self.master.data.get(
             ('builds',),
             filters=[resultspec.Filter('buildrequestid', 'eq', [buildrequest['buildrequestid']])])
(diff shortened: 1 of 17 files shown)


master/buildbot/worker/latent.py
```diff
@@ -210,7 +209,7 @@ def buildFinished(self, sb):
             if self.build_wait_timeout == 0:
                 # we insubstantiate asynchronously to trigger more bugs with
                 # the fake reactor
-                self.master.reactor.callLater(0, self.insubstantiate)
+                self.master.reactor.callLater(0, self._soft_disconnect)
```
seankelly: :smile:

master/buildbot/test/util/integration.py
```diff
@@ -182,7 +183,13 @@ def finishedCallback(_, data):
             yield self.master.data.updates.addChange(**useChange)
 
         # wait until we receive the build finished event
-        build = yield d
+        buildrequest = yield d
+        print(buildrequest)
```
seankelly: Should this print be here?

seankelly:
Other than the errant `print`, things look good internally. When workers are disconnected, they are immediately available for the next build.

PR #2455, opened 2016-10-19
base-ui: setup windows title according to the page

tardyp:
Having always buildbot is annoying when you try to bookmark pages
and look in the browser history.


sa2ajj:
I am not very proficient in `coffee`, so another pair of eyes would be welcome.

LGTM

PR #2456, opened 2016-10-21
Fix the 5min. tutorial: messageFormatter is now a class.

benallard:


tardyp:
reporter.rst also needs update, but this is a good start

PR #2458, opened 2016-10-21
worker Dockerfile bugfix

capitalaslash:
before `apt-get install` always `apt-get update`, otherwise no packages will be found.

### DIFF
--- a/master/contrib/docker/pythonnode_worker/Dockerfile
+++ b/master/contrib/docker/pythonnode_worker/Dockerfile
@@ -12,7 +12,7 @@ ARG         DEBIAN_FRONTEND=noninteractive
 user root
 # Install required npm packages
 RUN  curl -sL https://deb.nodesource.com/setup_6.x | bash - && \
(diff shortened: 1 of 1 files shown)


master/contrib/docker/pythonnode_worker/Dockerfile
```diff
@@ -12,7 +12,7 @@ ARG         DEBIAN_FRONTEND=noninteractive
 user root
 # Install required npm packages
 RUN  curl -sL https://deb.nodesource.com/setup_6.x | bash - && \
-        apt-get install -y -o APT::Install-Recommends=false -o APT::Install-Suggests=false \
```
tardyp: setup+6.x os doing an apt-get update already.

capitalaslash: sorry, I removed that line and did not bother to check the script.
anyway, as this should be an example for general purpose workers, it seems not so clear (at least to me). maybe mention it in the comment above?

tardyp:
I agree that an example shouldn't have implicit optimization like this

capitalaslash:
thanks!

PR #2463, opened 2016-10-24
Implement missing worker email in the email reporter

tardyp:
This is done via creating a custom event message
which the mail reporter can listen to
note that an irc reporter could also listen to the message and alert accordingly

### DIFF
--- a/master/buildbot/data/workers.py
+++ b/master/buildbot/data/workers.py
@@ -137,6 +137,14 @@ class Worker(base.ResourceType):
         self.produceEvent(bs, 'disconnected')
 
     @base.updateMethod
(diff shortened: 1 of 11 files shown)


tardyp:
@benallard see this fix for the worker missing notification stuff.

Its WIP. I'd like your stuff to land before starting to refactor the code

benallard:
I'd rather have you make another instance of MessageFormatter than make it being able to cope with two different body contents ... But that's probably what you already meant with 'WIP' ...

tardyp:
@benallard please have a look at the updated code

sa2ajj:
:+1:

benallard:
looks good to me.

PR #2465, opened 2016-10-26
Return source branch for Bitbucket PR

sam-bristow:
Instead of returning the source branch for a pull-request, the previous
code was returning the destination branch. This would mean the change
that was emitted would have the wrong branch details and the git
checkout would fail.

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -146,7 +146,7 @@ class BitbucketPullrequestPoller(base.PollingChangeSource):
                         comments=u'pull-request #%d: %s\n%s' % (
                             nr, title, prlink),
                         when_timestamp=datetime2epoch(updated),
-                        branch=self.branch,
+                        branch=branch,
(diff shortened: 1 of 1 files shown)


sam-bristow:
I'll update the tests and update the PR when they are all passing again.

master/buildbot/changes/bitbucket.py
```diff
@@ -146,7 +146,7 @@ def _processChanges(self, page):
                         comments=u'pull-request #%d: %s\n%s' % (
                             nr, title, prlink),
                         when_timestamp=datetime2epoch(updated),
-                        branch=self.branch,
```
tardyp: You may want to make the destination branch available somewhere so that a ChangeFilter can filter and route changes to the right builder

tardyp:
@sam-bristow do you intend to finish this?

sam-bristow:
Sorry, I've got pulled off onto other tasks.

I will create a new pull request when I find some more time to do it properly.

PR #2466, opened 2016-10-26
DebLintian: Put the log lines, each on their own line.

benallard:


benallard:
I'll make a more complete PR.

PR #2469, opened 2016-10-27
fixed availabale typo

valeriat:


benallard:
:+1: I wonder how the misspelling made it to the wordlist !

tardyp:
When we started with the spelling plugin, we put all the words in the whitelist as this was too much of work to fix everything at once.
Goal was to try to not put more spelling mistakes in the doc.

But there is still work to do to review the spelling_list.txt and remove bad words from it.
any help is very much appreciated!

PR #2471, opened 2016-10-27
Fix url to the sqlalchemy-migrate repository

benallard:


rutsky:
To fix the build make link anonymous: use two underscores at the end:
```
`SQLAlchemy-Migrate <https://github.com/openstack/sqlalchemy-migrate>`__.
```

PR #2474, opened 2016-10-27
Render OpenStackLatentWorker block devices

seankelly:
Render the values in the `block_devices` parameter. Also support automatically calculating the `volume_size`.

### DIFF
--- a/master/buildbot/test/unit/test_worker_openstack.py
+++ b/master/buildbot/test/unit/test_worker_openstack.py
@@ -81,20 +81,24 @@ class TestOpenStackWorker(unittest.TestCase):
                           openstack.OpenStackLatentWorker, 'bot', 'pass',
                           flavor=1, **self.os_auth)
 
+    @defer.inlineCallbacks
     def test_getImage_string(self):
         bs = openstack.OpenStackLatentWorker(
             'bot', 'pass', **self.bs_image_args)
-        self.assertEqual('image-uuid', bs._getImage(None, bs.image))
+        image_uuid = yield bs._getImage(self.build)
+        self.assertEqual('image-uuid', image_uuid)
 
+    @defer.inlineCallbacks
     def test_getImage_callable(self):
         def image_callable(images):
             return images[0]
 
         bs = openstack.OpenStackLatentWorker('bot', 'pass', flavor=1,
                                              image=image_callable, **self.os_auth)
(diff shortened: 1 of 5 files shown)


master/buildbot/test/unit/test_worker_openstack.py
```diff
@@ -40,6 +42,8 @@ class TestOpenStackWorker(unittest.TestCase):
     def setUp(self):
         self.patch(openstack, "nce", novaclient)
         self.patch(openstack, "client", novaclient)
+        self.build = Properties(
```
sa2ajj: [nit] this could be a single line

seankelly: I must have had more things planned/done to make it longer than 80 characters but then removed them. I thought it still was but just checked and it's well under. Oops, I should have checked earlier and I would have changed it. Oh well.

tardyp:
damn. I forgot to hit the send button for those comments. I'd appreciate a new PR for fixups.

master/buildbot/worker/openstack.py
```diff
@@ -111,30 +112,81 @@ def _parseBlockDevice(self, block_device):
...
+        # Allow None here. It will be rendered later.
+        client_block_device['volume_size'] = block_device.get('volume_size')
         return client_block_device
 
-    @staticmethod
-    def _getImage(os_client, image):
+    @defer.inlineCallbacks
+    def _renderBlockDevice(self, block_device, build):
+        """Render all of the block device's values."""
+        rendered_block_device = yield build.render(block_device.copy())
```
seankelly: Will rendering it do the copy? It does need to be different because it gets modified on line 133.

master/buildbot/worker/openstack.py
```diff
@@ -111,30 +112,81 @@ def _parseBlockDevice(self, block_device):
...
+        if rendered_block_device['volume_size'] is None:
+            source_type = rendered_block_device['source_type']
+            source_uuid = rendered_block_device['uuid']
+            volume_size = self._determineVolumeSize(source_type, source_uuid)
+            if volume_size is None:
```
seankelly: Will do.

master/buildbot/worker/openstack.py
tardyp: yes it does. You can convince yourself here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/properties.py#L796

PR #2464, opened 2016-10-25
Factor out the code that checks for a PB client connecting to an HTTP server hanging.

tomprince:
`buildbot_worker.pbutil` contains some code to detect if the worker connects to the HTTP port of the master. Factor that code into a wrapper that doesn't use inheritance, in preparation for supporting connecting via endpoints. (This is part of spitting up and improving #2387).

PR #2475, opened 2016-10-29
Add a backport of `SynchronousTestCasa`

tomprince:
This actually just makes `TestCase` pretend to be `SynchronousTestCase` currently, since the difference is mostly just that `SynchronousTestCase` does less. (There is also some stuff that was add to `SynchronousTestCase` that makes synchronous testing easier, but on versions with that code `TestCase` inherits from it).

This incidentally bumps the minimum version of twisted required for the worker to 10.2.0, which is 5y 12mo old, and is the old version that we tested against that supports deferred cancellation, which was used in #2464. (It was introduced in 10.1.0, bu that that wasn't in our test matrix, and 10.2.0 is plenty old enough).

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -42,6 +42,12 @@ env:
   # Configuration to run `python setup.py test` to check this test runner.
   # - TWISTED=latest SQLALCHEMY=latest TESTS=setuppy_test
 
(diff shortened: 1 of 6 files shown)


tardyp:
thanks for the quick fix!

PR #2314, opened 2016-07-11
Proposal to implement Buildbot credentials management

protatremy:
Preliminary documentation explain planned work for supporting buildbot credential management

PR #2477, opened 2016-10-30
secrets

sa2ajj:
A return to the topic raised in #2314.  Please note that this is in no way complete: it's only a skeleton to show the basic idea.

Design goals:
- **not** to implement any kind of sophisticated secret storage
- make it possible to implement interfaces to secret storages out there
- provide incremental path to convert to the new API

---

Let me try to write down first what I have in mind...

The `master.cfg` will have something like this

``` python
c['secrets'] = [
    ...,
    secrets.SecretInAFile('basic'),
    ...
]
```

In the places where we want to use one of the secrets provided by it, we'd refer to that particular provider as

``` python
c['workers'] = [
    ...,
    worker.Worker('user', secrets.Secret('basic'), ...),
    ...
]
```

and in the place where we actually need that secret, we'd use

``` python
    def something(user, secret, ...):
        actual_secret = self.master.secrets.get(secret, user)
        ...
        actual_secret.value
```

whatever passed to `.get` after `secret` is provider specific as well as the actual value of the secret.

### DIFF
--- a/master/buildbot/secrets/manager.py
+++ b/master/buildbot/secrets/manager.py
@@ -1,0 +1,70 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+"""
+manage providers and handle secrets
+"""
+from future.utils import itervalues
+from future.utils import text_type
+
+from buildbot.secrets.secret import ActualSecret
+from buildbot.secrets.secret import Secret
+from buildbot.util.service import BuildbotService
+
+
+class SecretManager(BuildbotService):
+    """
+    Secret manager
+    """
+    name = 'secrets'
+
+    config_attr = 'providers'
+
+    def getConfigDict(self):
+        return {
+            'name': self.name,
+            'providers': [
+                p.getConfigDict() for p in itervalues(self.namedServices)
+            ]
+        }
+
+    def get_provider(self, provider_name):
+        """
+        get provider by its name
+        """
+        provider = self.namedServices.get(provider_name, None)
+
+        assert provider, 'unknown secret provider: {}'.format(provider_name)
+
+    def get(self, secret, *args, **kwargs):
+        """
+        get a secret from the provider defined in the secret using args and
+        kwargs
+        """
+        allowed_types = (text_type, Secret)
+        assert isinstance(secret, allowed_types), \
+            'secret must be an instance of {!r}'.format(allowed_types)
+
+        if isinstance(secret, text_type):
+            # backward compatibility, produce a warning?
+            value, props = secret, None
+            source_name = 'Plain text secret'
+        else:
+            provider = self.get_provider(secret.provider_name)
+            value, props = provider.get(*args, **kwargs)
+            source_name = repr(provider)
+
+        return ActualSecret(
+            '{}: args={!r}, kwargs={!r}'.format(source_name, args, kwargs),
+            value, props=props)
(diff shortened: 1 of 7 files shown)


master/buildbot/secrets/manager.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+
+    @classmethod
+    def get_manager(cls, master):
+        """
+        get secret manager provided the given master
+        """
+
+    @classmethod
+    def get(cls, master, secret, *args, **kwargs):
```
tardyp: I think this would make sense to make the VaultManager a first class service, like data and mq, so that you can access it via

    self.master.secrets.get(secret_name)

master/buildbot/secrets/secret.py
```diff
@@ -0,0 +1,69 @@
...
+
+class ActualSecret(object):
+    """
+    ...
+    """
+    def __init__(self, source, value, props=None):
+        if props is None:
+            props = dict()
+
+        self._source = source
```
tardyp: don't you want to reference a secret by a known name or id?

sa2ajj: not really, i think your main comment applies here:

> I think it is best to reason on how to use it rather than how we implement it.
> With this implementation, I have difficulties to get how you would use the SecretInAFile

n054:
This still has a do not merge label so I'm waiting to run this.

sa2ajj:
@n054 there's nothing to run here yet: I'm just in a process of fleshing out the high level design to see if it can get anywhere...

sa2ajj:
updated head comment

master/buildbot/secrets/manager.py
```diff
@@ -0,0 +1,70 @@
...
+    def get(self, secret, *args, **kwargs):
+        """
+        get a secret from the provider defined in the secret using args and
+        kwargs
+        """
+        allowed_types = (text_type, Secret)
+        assert isinstance(secret, allowed_types), \
+            'secret must be an instance of {!r}'.format(allowed_types)
+
+        if isinstance(secret, text_type):
```
tardyp: I think we should use the zope adapters like we do for renderables. The concept is quite similar to renderable, so it make sense to have a similar implementation. Here is how you would implement the adapter for a string.


    from twisted.python.components import registerAdapter
    from zope.interface import Interface
    from zope.interface import implementer


    class ISecret(Interface):

        """An Reference to a secret"""

        def getSecret(master):
            """Return a deferred that fires with the secret value and the secret source
            @param master: the master object
            """


    @implementer(ISecret)
    class _DefaultSecret(object):

        """
        Default ISecret adaptor (for people who dont use a secretprovider), return the actual string
        """

        def __init__(self, value):
            self.value = value

        def getSecret(self, master):
            return self.value, None

    registerAdapter(_DefaultSecret, str, ISecret)


    class SecretManager(...):
        [...]
        def get_secret(master, secret):
            return ISecret(secret).getSecret(master)

master/buildbot/secrets/secret.py
```diff
@@ -0,0 +1,71 @@
...
+        return self._source
+
+    @property
+    def value(self):
+        """
+        secret value
+        """
+        return self._value
+
+    @property
```
tardyp: What are the purpose of those properties? I can't find a usecase for it.

tardyp:
I would like to avoid having to define all the passwords in the secret config.
I think it would be easier to only define providers.
Each providers would be asked to look for a password value in its DB (the SecretInFiles would stat for 'filename.pass', and return None if it does not exist to handover to the next provider)

    c['secretsProviders'] = [
        secrets.SecretInFiles(fileExt='.pass'),
        secrets.SecretInYaml(filename="secrets.yaml"),
        secrets.SecretInVault(server="tcp:localhost:2343"),
    ]

tardyp:
Also, this design only take in account the password usecase, and there is a need to also take in account the ssh key needs:

ssh keys needs to be uploaded to the worker at the begining of a step in a specific place, and removed at the end of the step.

this is the same for example for:

- ssh keys in ~/.ssh/id_rsa
- hyper keys in ~/.hyper/config.json
- pypirc in ~/.pypirc
- wgetrc in ~/.wgetrc

This is a little bit more complex, so it does not need to be implemented in the beginning, but we need to have a path toward it. (maybe that would be a usecase for your properties dictionary)

protatremy:
Hi Sa2ajj, I'm interested in your work.
I made some  for Vault on my side, so I propose to work on a Vault provider to complete this topic.

sa2ajj:
@protatremy do you think yoiu could describe your use case? (honestly, I'm currently shooting in the dark(

protatremy:
@sa2ajj my use case is to store credentials, encrypted by Vault and stored in the Consul back-end. A way to not store credentials in files but in a secured solution.
The provider that I purpose to write will get the credentials as your provider does.
I will add a docker File to install Vault and Consul and a tool (command tool) to insert new credentials.

tardyp:
superseed by https://github.com/buildbot/buildbot/pull/2660

PR #2481, opened 2016-11-02
add qtile theme files to sdist

termim:
Building documentation from sdist fails as qtile theme is missing in the tarball


tardyp:
looks good! thanks!

PR #2483, opened 2016-11-03
TxRequestResponseWrapper has a 'code' property

benallard:
This should fix :bug:`3642`


tardyp:
CI fail because of https://github.com/sass/node-sass/issues/1781
waiting for a fix

PR #2484, opened 2016-11-03
Replace localhost with 127.0.0.1

benallard:
It seems buildbot (mine at least) is not listening on the ipv6 variant of localhost, which is just losing every second request made by nginx.

### DIFF
--- a/master/docs/manual/cfg-www.rst
+++ b/master/docs/manual/cfg-www.rst
@@ -419,19 +419,19 @@ Here is an nginx configuration that is known to work (nginx 1.6.2):
 
             # you could use / if you use domain based proxy instead of path based proxy
             location /buildbot/ {
-                proxy_pass http://localhost:5000/;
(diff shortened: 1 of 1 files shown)


tardyp:
CI fail because of https://github.com/sass/node-sass/issues/1781
waiting for a fix

benallard:
From https://twistedmatrix.com/documents/16.1.1/api/twisted.internet.interfaces.IReactorTCP.listenTCP.html (which I believe is what we use):

> To bind to all IPv4 and IPv6 addresses, you must call this method twice.

Which I believe is what would be better ...

tardyp:
We are using https://github.com/twisted/twisted/blob/twisted-16.5.0/src/twisted/application/strports.py#L41
which in turn uses https://github.com/twisted/twisted/blob/twisted-16.5.0/src/twisted/internet/endpoints.py#L462
but calls listenTCP only once. That requirement to call listen twice looks very weird and racy to me.

PR #2490, opened 2016-11-07
bugfix: cmake step now inherits from new style BuildStep

noc0lour:
Otherwise using env in cmake leads to `unkown keyword` error.
Inherting from BuildStep instead from LoggingBuildStep and modifying __init__ process according to ShellSequence seems to solve the issue for me.

### DIFF
--- a/master/buildbot/steps/cmake.py
+++ b/master/buildbot/steps/cmake.py
@@ -17,11 +17,11 @@ from future.utils import iteritems
 from twisted.internet import defer
 
 from buildbot import config
-from buildbot.process.buildstep import LoggingBuildStep
+from buildbot.process.buildstep import BuildStep
 from buildbot.process.buildstep import ShellMixin
 
 
-class CMake(ShellMixin, LoggingBuildStep):
+class CMake(ShellMixin, BuildStep):
     DEFAULT_CMAKE = 'cmake'
 
     name = 'cmake'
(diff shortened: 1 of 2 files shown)


tardyp:
This is not very clear to me what is being resolved.

Please provide a unit test which tests the case that you are trying to fix.
The test should fail with the old code, and success with the new code.

noc0lour:
I've added a unittest to check if environment in CMake works.
Running trial buildbot.test with HEAD of master produces following error:
`buildbot.config.ConfigErrors: <class 'buildbot.steps.cmake.CMake'>.__init__ got unexpected keyword argument(s) ['env']`

Running unittests with this branch cmake_steps work fine.

sa2ajj:
@noc0lour could you please describe a bit more what scenario you are fixing?  I've been using the step in production w/ no problem, so I'd really like to understand if it's something I just did not hit.

noc0lour:
IIf I try to initialize CMake with a env= keyword the step isn't constructed properly and causes the error above. Like in the unittest try to construct a step: `steps.cmake.CMake(env={'a':'b;})` or run the unittest on master. This is especially important if you want to use libraries/binaries which are not systemwide installed and you want to supply a custom PATH/LIBRARY_PATH environment variable. Also documentation states users can specify `env=` in the CMake step.

sa2ajj:
aha, I see: I do not use this feature so I did not hit it.  Lemme review the change again...

noc0lour:
I don't know if the changed inheritance has other implication. I just took an example in the ShellSequence step and at least in this use case it works out. Maybe another fix is appropriate.

tardyp:
I think this is good now. @sa2ajj I'll let you merge if you agree

sa2ajj:
thank you, @noc0lour

PR #2492, opened 2016-11-09
Install windows service scripts only on windows.

termim:
Windows service scripts are not needed on non Windows platforms.


tardyp:
We can'tdo that because this will mean the wheels wont work for windows.
We dont want to publish a wheel specific for windows

PR #2494, opened 2016-11-09
Use api.bitbucket.org for endpoint

k4ml:
Using bitbucket.org/api/ endpoint has consequences that subsequent url in the results will use bitbucket.org/!api/ which intended for bitbucket own use. If someone try to add authentication support, bitbucket.org/!api/ will return 401 status code even the credentials is valid.

Reference - https://bitbucket.org/atlassian/atlassian-connect-express/pull-requests/109/use-apibitbucketorg-instead-of-api/diff

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -83,7 +83,7 @@ class BitbucketPullrequestPoller(base.PollingChangeSource):
         self.lastPoll = time.time()
         log.msg("BitbucketPullrequestPoller: polling "
                 "Bitbucket repository %s/%s, branch: %s" % (self.owner, self.slug, self.branch))
-        url = "https://bitbucket.org/api/2.0/repositories/%s/%s/pullrequests" % (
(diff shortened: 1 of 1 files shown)


tardyp:
tests needs to be updated

tardyp:
There is also a bitbucket change hook. I believe change hooks are a much better approach.

It is not obvious why the tests would fail, and this poller code is not of very good quality.
If the code is broken I would be inclined to remove this poller from buildbot.

noc0lour:
@tardyp: Somewhat related: We have a working Poller for Github PRs based on this Poller and was planning to wrap things up to get it upstream. To maintain support for Pollers would be great because this provides support for people who cannot expose a public endpoint for Github/BB. But if you plan to drop PR-poller support I don't need to make things pretty ;)

tardyp:
@noc0lour I think this indeed make sense to keep a poller for firewalled bots.

Problem is this poller does not use the HttpClientService (which wasn't there when it was first written), and thus the tests are not very good.

So for new code, I would request that it uses HttpClientService.

tardyp:
closing as per stalled policy. please feel free to re-open if you get time to finish it.

PR #2495, opened 2016-11-10
Support SSL connections in mail notifier

perher:
Currently the mail notifier report plugin only supports encryption using STARTTLS. Since my provider doesn't have any support for this I never managed to get the plugin working.
With this patch I got it working by adding optional support for connecting over SSL.

### DIFF
--- a/master/docs/manual/cfg-reporters.rst
+++ b/master/docs/manual/cfg-reporters.rst
@@ -256,6 +256,10 @@ MailNotifier arguments
     When this argument is ``True`` (default is ``False``) ``MailNotifier`` sends emails using TLS and authenticates with the ``relayhost``.
     When using TLS the arguments ``smtpUser`` and ``smtpPassword`` must also be specified.
 
+``useSsl``
+    (boolean).
+    When this argument is ``True`` (default is ``False``) ``MailNotifier`` sends emails using SSL and authenticates with the ``relayhost``.
+
 ``smtpUser``
     (string).
     The user name to use when authenticating with the ``relayhost``.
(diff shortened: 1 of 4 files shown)


master/docs/manual/cfg-reporters.rst
```diff
@@ -253,6 +253,10 @@ MailNotifier arguments
     When this argument is ``True`` (default is ``False``) ``MailNotifier`` sends emails using TLS and authenticates with the ``relayhost``.
     When using TLS the arguments ``smtpUser`` and ``smtpPassword`` must also be specified.
 
+``useSsl``
+    (boolean).
+    When this argument is ``True`` (default is ``False``) ``MailNotifier`` sends emails using SSL and authenticates with the ``relayhost``.
```
tardyp: We should explain better what is the difference between TLS mode, and SSL mode

seankelly: I think the `useSsl` parameter should be renamed. Perhaps `connectSsl` to match that `connectSSL` is used or `connectTls` because SSL is completely deprecated now.

The `useTls` option controls `requireTransportSecurity` which makes Twisted require that STARTTLS is used or that the connection is over SSL/TLS in the first place. That option in MailNotifier should also be renamed but too late.

perher:
I renamed the new variable to avoid confusion between SSL and TLS, I think SMTPS is the correct term for this. I also made an attempt at writing some unit test for it.

perher:
Thanks for review, I've now also updated useTls documentation

sa2ajj:
@perher could you please update the release notes?

perher:
@sa2ajj is this ok?

perher:
do you mean to add release notes for the unit tests I've already added?

sa2ajj:
@perher sorry, my reading skills fail these days: i only looked at the release notes and failed to see the tests.

n054:
Looks like Travis is pending only. I will push commits and see if Travis picks them up and tests them. @n054

tardyp:
@n054 Looks like I forgot about this one. I fixed the relnotes to use the new towncrier system. I'll wait for CI and merge it.

PR #2499, opened 2016-11-13
Introduction text for the authorization framework

tardyp:
blockdiag looks like this:

image

### DIFF
--- a/master/docs/manual/cfg-www.rst
+++ b/master/docs/manual/cfg-www.rst
@@ -477,7 +477,7 @@ Authorization rules
 ~~~~~~~~~~~~~~~~~~~
 
 The authorization framework in Buildbot is very generic and flexible.
-Drawback is that it is not very obvious for new comers.
+Drawback is that it is not very obvious for newcomers.
 The 'simple' example will however allow you to easily start by implementing an admins-have-all-rights setup.
 
 Please carefully read the following documentation to understand how to setup authorization in Buildbot.
@@ -502,7 +502,7 @@ It allows or denies access to the REST APIs according to rules.
 
   It is similar but different to the usual notion of group:
 
-  - a user can have several roles, and a role can be given to several users.
+  - A user can have several roles, and a role can be given to several users.
   - Role is an application specific notion, while group is more organization specific notion.
   - Groups are given by the auth plugin, e.g ``ldap``, ``github``, and are not always in the precise control of the buildbot admins.
   - Roles can be dynamically assigned, according to the context.
@@ -532,7 +532,7 @@ Endpoint matchers are responsible for creating rules to match REST endpoints, an
 Endpoint matchers are processed in the order they are configured.
 The first rule matching an endpoint will prevent further rules from being checked.
 To continue checking other rules when the result is `deny`, set `defaultDeny=False`.
-If not endpoint matcher matches, then the access is granted.
+If no endpoint matcher matches, then the access is granted.
 
 One can implement the default deny policy by putting an :py:class:`AnyEndpointMatcher` with nonexistent role in the end of the list.
 Please note that this will deny all REST apis, and most of the UI do not implement proper access denied message in case of such error.
--- a/master/docs/manual/cfg-www.rst
+++ b/master/docs/manual/cfg-www.rst
@@ -499,18 +499,28 @@ It allows or denies access to the REST APIs according to rules.
     }
 
 - Roles is a label that you give to a user.
-  It is similar to the notion of group, but with more flexibility.
-  For an example of how it is different to group, there is the ``owner`` role, which can be given to a user for a build that he is at the origin.
 
-- Endpoint matchers associates role requirements to REST API endpoints.
+  It is similar but different to the usual notion of group:
 
-- Role matchers associates authenticated users to roles.
+  - a user can have several roles, and a role can be given to several users.
+  - Role is an application specific notion, while group is more organization specific notion.
+  - Groups are given by the auth plugin, e.g ``ldap``, ``github``, and are not always in the precise control of the buildbot admins.
+  - Roles can be dynamically assigned, according to the context.
+    For example, there is the ``owner`` role, which can be given to a user for a build that he is at the origin, so that he can stop or rebuild only builds of his own.
+
+- Endpoint matchers associate role requirements to REST API endpoints.
+  The default policy is allow in case no matcher matches (see below why)
+
+- Role matchers associate authenticated users to roles.
+
+Authz Configuration
++++++++++++++++++++
 
 .. py:class:: buildbot.www.authz.Authz(allowRules=[], roleMatcher=[], stringsMatcher=util.fnmatchStrMatcher)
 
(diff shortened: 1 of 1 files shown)


master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
 Authorization rules
 ~~~~~~~~~~~~~~~~~~~
 
+The authorization framework in Buildbot is very generic and flexible.
+Drawback is that it is not very obvious for new comers.
```
Lekensteyn: "newcomers"

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+
+      User  [shape = actor];
+      AuthenticatedUser  [shape = actor];
+      RoleMatcher [shape = diamond];
+      EndpointMatcher [shape = diamond];
+    }
+
+- Roles is a label that you give to a user.
+  It is similar to the notion of group, but with more flexibility.
+  For an example of how it is different to group, there is the ``owner`` role, which can be given to a user for a build that he is at the origin.
```
Lekensteyn: I would expect a group to be a collection of users. Users can be a member of multiple groups.
I would expect a role to be similar to this notion of groups. What is the difference? Groups are fixed while roles can dynamically assigned?

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+
+- Endpoint matchers associates role requirements to REST API endpoints.
```
Lekensteyn: "associate"

Maybe clarify the default policy when nothing matches? (reject by default, but there seems to be a special case when the list is empty (allow all). I see it is mentioned below, but for a global picture I think it is an important detail.

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+
+- Role matchers associates authenticated users to roles.
```
Lekensteyn: "associate" (or "A role matcher associates ...")

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+
+.. py:class:: buildbot.www.authz.Authz(allowRules=[], roleMatcher=[], stringsMatcher=util.fnmatchStrMatcher)
+
+    :param allowRules: List of EndpointMatchers processed in order for each endpoint grant request.
+    :param roleMatcher: List of RoleMatchers
+    :param stringsMatcher: Selects algorithm used to make string comparaison.
```
Lekensteyn: "to make [a] string comparison" -> "to match a role with a username"? (if that makes sense?)

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+       can be :py:class:`util.fnmatchStrMatcher` or :py:class:`util.reStrMatcher` from ``from buildbot.plugins import util``
+
+    :py:class:`Authz` needs to be configured in ``c['www']['authz']``
+
 Endpoint matchers
 +++++++++++++++++
 
 Endpoint matchers are responsible for creating rules to match REST endpoints, and requiring roles for them.
-The following sequence is implemented by each EndpointMatcher class
+Endpoint matchers are process in the order they are configured.
```
Lekensteyn: "are processed in"

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,71 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+The first that matches stop the sequence unless ``defaultDeny=False`` is passed.
```
Lekensteyn: What about:

    The first rule matching an endpoint will prevent further rules from being checked. To continue checking other rules when the result is `deny`, set `defaultDeny=False`.

master/docs/manual/cfg-www.rst
sa2ajj: and, as spellchecking suggested, "comparison"

master/docs/manual/cfg-www.rst
tardyp: there is no reject by default. It is allow by default.

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,86 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+
 Endpoint matchers
 +++++++++++++++++
 
 Endpoint matchers are responsible for creating rules to match REST endpoints, and requiring roles for them.
-The following sequence is implemented by each EndpointMatcher class
+Endpoint matchers are processed in the order they are configured.
+The first rule matching an endpoint will prevent further rules from being checked.
+To continue checking other rules when the result is `deny`, set `defaultDeny=False`.
+If not endpoint matcher matches, then the access is granted.
```
Lekensteyn: "If no"

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,86 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+      AuthenticatedUser  [shape = actor];
+      RoleMatcher [shape = diamond];
+      EndpointMatcher [shape = diamond];
+    }
+
+- Roles is a label that you give to a user.
+
+  It is similar but different to the usual notion of group:
+
+  - a user can have several roles, and a role can be given to several users.
```
Lekensteyn: Nit: "A user"

master/docs/manual/cfg-www.rst
```diff
@@ -476,17 +476,86 @@ Here is a configuration that is known to work (Apache 2.4.10 / Debian 8), direct
...
+- Look if the users has the required role.
+
+Several endpoints matchers are currently implemented.
+If you need a very complex setup, you may need to implement your own endpoint matchers.
+In this case, you can look at the source code for detailed examples on how to write endpoint matchers.
+
+.. py:class:: buildbot.www.authz.endpointmatchers.EndpointMatcherBase(role, defaultDeny=True)
+
+    :param role: The role which grants access to this endpoint.
+        List of roles is not supported, but a ``fnmatch`` expression can be provided to match several roles.
```
Lekensteyn: Isn't fnmatch only available when `stringsMatcher=fnmatchStrMatcher` (default)?

tardyp: indeed. I did not want to make it more complicated than it needs

sa2ajj:
After @Lekensteyn's issues are address, this is good to go.

tardyp:
@Lekensteyn remarks should be addressed now

PR #2502, opened 2016-11-15
gerritVerifyStatusPush: gerrit reporter using gerrit verify-status plugin

tardyp:
Gerrit 2.13 has a new verify-status plugin which allows several CI status to be pushed for the same commit (like for github)

https://gerrit.googlesource.com/plugins/verify-status/

### DIFF
--- a/master/buildbot/reporters/gerrit_verify_status.py
+++ b/master/buildbot/reporters/gerrit_verify_status.py
@@ -48,10 +48,17 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
     DEFAULT_RESULT = -1
 
     @defer.inlineCallbacks
-    def reconfigService(self, baseURL, auth,
-                        startDescription=None, endDescription=None,
-                        verification_name=None, abstain=False, category=None, reporter=None,
-                        verbose=False, **kwargs):
+    def reconfigService(self,
+                        baseURL,
+                        auth,
+                        startDescription=None,
+                        endDescription=None,
+                        verification_name=None,
+                        abstain=False,
+                        category=None,
+                        reporter=None,
+                        verbose=False,
+                        **kwargs):
         yield http.HttpStatusPushBase.reconfigService(self, **kwargs)
 
         if baseURL.endswith('/'):
@@ -60,18 +67,27 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
         self._http = yield httpclientservice.HTTPClientService.getService(
             self.master, baseURL, auth=auth)
 
-        self.verification_name = verification_name or Interpolate('%(prop:buildername)s')
-        self.reporter = reporter or "buildbot"
-        self.abstain = abstain
-        self.category = category
-        self.startDescription = startDescription or 'Build started.'
-        self.endDescription = endDescription or 'Build done.'
-        self.verbose = verbose
+        self._verification_name = verification_name or Interpolate(
+            '%(prop:buildername)s')
+        self._reporter = reporter or "buildbot"
+        self._abstain = abstain
+        self._category = category
+        self._startDescription = startDescription or 'Build started.'
+        self._endDescription = endDescription or 'Build done.'
+        self._verbose = verbose
 
     def createStatus(self,
-                     change_id, revision_id, name, value, abstain=None,
-                     rerun=None, comment=None, url=None, reporter=None,
-                     category=None, duration=None):
+                     change_id,
+                     revision_id,
+                     name,
+                     value,
+                     abstain=None,
+                     rerun=None,
+                     comment=None,
+                     url=None,
+                     reporter=None,
+                     category=None,
+                     duration=None):
         """
         Abstract the POST REST api documented here:
         https://gerrit.googlesource.com/plugins/verify-status/+/master/src/main/resources/Documentation/rest-api-changes.md
@@ -92,9 +108,7 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
 
         :return: A deferred with the result from Gerrit.
         """
-        payload = {
-            'name': name,
-            'value': value}
+        payload = {'name': name, 'value': value}
 
         if abstain is not None:
             payload['abstain'] = abstain
@@ -117,13 +131,18 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
         if duration is not None:
             payload['duration'] = duration
 
-        if self.verbose:
-            log.info(
+        if self._verbose:
+            log.debug(
                 'Sending Gerrit status for {change_id}/{revision_id}: data={data}',
-                change_id=change_id, revision_id=revision_id, data=payload)
+                change_id=change_id,
+                revision_id=revision_id,
+                data=payload)
 
-        return self._http.post('/'.join([
-            '/a/changes', str(change_id), 'revisions', str(revision_id), 'verify-status~verifications']),
+        return self._http.post(
+            '/'.join([
+                '/a/changes', str(change_id), 'revisions', str(revision_id),
+                'verify-status~verifications'
+            ]),
             json=payload)
 
     def formatDuration(self, duration):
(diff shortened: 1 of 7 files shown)


master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+            payload['reporter'] = reporter
+
+        if category is not None:
+            payload['category'] = category
+
+        if duration is not None:
+            payload['duration'] = duration
+
+        if self.verbose:
+            log.info(
```
sa2ajj: maybe change this to `.debug` and remove the condition?

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+        self._http = yield httpclientservice.HTTPClientService.getService(
+            self.master, baseURL, auth=auth)
+
+        self.verification_name = verification_name or Interpolate('%(prop:buildername)s')
+        self.reporter = reporter or "buildbot"
+        self.abstain = abstain
+        self.category = category
+        self.startDescription = startDescription or 'Build started.'
+        self.endDescription = endDescription or 'Build done.'
+        self.verbose = verbose
```
sa2ajj: do we really need this?  with the fine grained control over the logging this seems to be unnecessary [any more]?

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+                        verbose=False, **kwargs):
+        yield http.HttpStatusPushBase.reconfigService(self, **kwargs)
+
+        if baseURL.endswith('/'):
+            baseURL = baseURL[:-1]
+
+        self._http = yield httpclientservice.HTTPClientService.getService(
+            self.master, baseURL, auth=auth)
+
+        self.verification_name = verification_name or Interpolate('%(prop:buildername)s')
```
sa2ajj: which of these data members we really want to expose as the part of the API?

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+        :param url: The url link to more info about this job
+        :reporter: The user that verified this job
+        :category: A category for this job
+        "duration": The time it took to run this job
+
+        :return: A deferred with the result from Gerrit.
+        """
+        payload = {
+            'name': name,
+            'value': value}
```
sa2ajj: [nit] i'd put `}` on a separate line

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def formatDuration(self, duration):
+        """Format the duration.
+
+        This method could be overriden if really needed, as the duration format in gerrit
+        is an arbitrary string.
+        :param duration: duration in timedelta
+        """
+        days = duration.days
+        hours, remainder = divmod(duration.seconds, 3600)
```
sa2ajj: [side note] i'm somewhat surprise `timedelta` does not provide these considering what its constructor looks like

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+        if self.verbose:
+            log.info(
+                'Sending Gerrit status for {change_id}/{revision_id}: data={data}',
+                change_id=change_id, revision_id=revision_id, data=payload)
+
+        return self._http.post('/'.join([
+            'changes', str(change_id), 'revisions', str(revision_id), 'verify-status~verifications']),
+            json=payload)
+
+    def formatDuration(self, duration):
```
sa2ajj: is it a part of the API?  (i think it's OK to override "private" methods as well)

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+        if days:
+            return '{} day{} {}h {}m {}s'.format(
+                days, "s" if days > 1 else "",
+                hours, minutes, seconds)
+        elif hours:
+            return '{}h {}m {}s'.format(hours, minutes, seconds)
+        else:
+            return '{}m {}s'.format(minutes, seconds)
+
+    def getGerritChanges(self, props):
```
sa2ajj: `@staticmethod`?

on a related note, is it a part of the API?

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+        return []
+
+    @defer.inlineCallbacks
+    def send(self, build):
+        props = Properties.fromDict(build['properties'])
+        if build['complete']:
+            value = self.RESULTS_TABLE.get(build['results'], self.DEFAULT_RESULT)
+            comment = yield props.render(self.endDescription)
+            duration = self.formatDuration(build['complete_at'] - build['started_at'])
+
```
sa2ajj: why do we need this empty line?

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+            value = 0
+            comment = yield props.render(self.startDescription)
+            duration = 'pending'
+
+        print build['results'], value
```
sa2ajj: i think logging should be used here (if it's not a debug print, in the latter case -- remove)

master/buildbot/test/fake/httpclientservice.py
```diff
@@ -57,6 +57,7 @@ class HTTPClientService(service.SharedService):
 
     getName from the fake and getName from the real must return the same values.
     """
+    quiet = False
```
sa2ajj: I think this and related changes should be a separate commit: otherwise it's a bit difficult to see what was introduced for the new functionality and what was otherwise required *but* unrelated

master/buildbot/test/fake/httpclientservice.py
```diff
@@ -108,8 +109,9 @@ def assertNoOutstanding(self):
                               "expected more http requests:\n {!r}".format(self._expected))
 
     def _doRequest(self, method, ep, params=None, data=None, json=None):
-        log.debug("{method} {ep} {params!r} <- {data!r}",
-                  method=method, ep=ep, params=params, data=data or json)
+        if not self.quiet:
```
sa2ajj: why do we need this condition since below we use `.debug`, then it's up to the global logging configuration...

tardyp: its for testing the .verbose method. You can test expectWasQuiet, because the fake is verbose.

master/buildbot/test/fake/httpclientservice.py
tardyp: its related, needed to create the test_not_verbose test

master/buildbot/reporters/gerrit_verify_status.py
tardyp: oups. debug print

master/buildbot/reporters/gerrit_verify_status.py
tardyp: Well, it all http reporters have the same.

I think the logging system is more for advanced users, and that being able to easily debug by adding verbose to the reporter makes more sense for beginner.
It is not so much of a burden to maintain.

master/buildbot/reporters/gerrit_verify_status.py
tardyp: apparently not, according to http://stackoverflow.com/questions/538666/python-format-timedelta-to-string, where I borrowed the code.

master/buildbot/reporters/gerrit_verify_status.py
tardyp: It is meant to be overriden if really needed (as per the docstring)

PR #2503, opened 2016-11-16
fix the dockerfile

tardyp:
no install recommend is not such a good idea, as recommends pulls important packages like setuptools.


tardyp:
I added a fix for the pip instability issues we have @sa2ajj please merge if you are okay with it.

PR #2504, opened 2016-11-16
GerritEventLogPoller

tardyp:
Support for the events-log gerrit plugin

This is a first PR.

The second PR should be a GerritEventLogChangeSource, which should combine ssh and eventlog REST API to be sure not to miss events in case of ssh disconnection.
Multiple inheritence should make that pretty trivial.

A third PR will be the GerritKafkaChangeSource which will connect to a kafka server


sa2ajj:
:+1: 

I'm looking forward to seeing the next patch :smile:

PR #2505, opened 2016-11-16
httpclientservice: do not use urljoin

tardyp:
urljoin is stripping the last path if it does not with /
we rather join manually, and enforce baseurl does not end with / and ep start with /

### DIFF
--- a/master/buildbot/test/fake/httpclientservice.py
+++ b/master/buildbot/test/fake/httpclientservice.py
@@ -60,8 +60,8 @@ class HTTPClientService(service.SharedService):
     quiet = False
 
     def __init__(self, base_url, auth=None, headers=None):
-        service.SharedService.__init__(self)
         assert not base_url.endswith("/"), "baseurl should not end with /"
+        service.SharedService.__init__(self)
         self._base_url = base_url
         self._auth = auth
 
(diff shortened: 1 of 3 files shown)


tardyp:
I think this is good to go. This is serious issue.  Shall we make a 0.9.0p1 for that?

seankelly:
Maybe add in #2507 too.

master/buildbot/test/fake/httpclientservice.py
```diff
@@ -61,6 +61,7 @@ class HTTPClientService(service.SharedService):
 
     def __init__(self, base_url, auth=None, headers=None):
         service.SharedService.__init__(self)
+        assert not base_url.endswith("/"), "baseurl should not end with /"
```
sa2ajj: I'd rather put it *before* `SharedService.__init__` call: why do extra if things are bad already?

master/buildbot/util/httpclientservice.py
```diff
@@ -90,6 +89,7 @@ class HTTPClientService(service.SharedService):
 
     def __init__(self, base_url, auth=None, headers=None):
         service.SharedService.__init__(self)
+        assert not base_url.endswith("/"), "baseurl should not end with /: " + base_url
```
sa2ajj: ditto

tardyp:
@sa2ajj please approve the change, so that we can merge it for 0.9.2

PR #2506, opened 2016-11-17
switch releasenote generation to towncrier

tardyp:
I got tired of rebasing the PR because the releasenote conflicts.

As per @tomprince suggestion, I propose we move to towncrier to manage release notes.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,13 @@
+## Remove this paragraph
+Please have a look at our developer documentation before submitting your Pull Request.
+
+http://trac.buildbot.net/wiki/Development
+And especially:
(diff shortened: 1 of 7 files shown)


tardyp:
@sa2ajj can we merge this so that we can require people to rewrite their relnotes as newsfiles? (mostly valid for my PRs)

sa2ajj:
This is a *bold* move :smile:

sa2ajj:
One comment *after* I merged it: towncrier does not seem to support sections (at least, this is how I read the code); I believe, it is important to have changes under different categories.

On a related note [cutting releases page](http://trac.buildbot.net/wiki/CuttingReleases) needs to be updated.

tomprince:
It does support sections, although not sub-sections, by default. It pick the section by the extension of the news fragment. It looks like it would be possible to have sub-sections of some sort, if we used a custom template.

PR #2507, opened 2016-11-17
fixes GitHub status 403 error

hcwiley:
https://developer.github.com/v3/#user-agent-required

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -52,7 +52,7 @@ class GitHubStatusPush(http.HttpStatusPushBase):
         self._http = yield httpclientservice.HTTPClientService.getService(
             self.master, baseURL, headers={
                 'Authorization': 'token ' + token,
-                'User-Agent': 'buildbot'
+                'User-Agent': 'Buildbot'
             })
         self.verbose = verbose
 
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -50,7 +50,10 @@ class GitHubStatusPush(http.HttpStatusPushBase):
             baseURL = baseURL[:-1]
 
         self._http = yield httpclientservice.HTTPClientService.getService(
-            self.master, baseURL, headers={'Authorization': 'token ' + token})
+            self.master, baseURL, headers={
+                'Authorization': 'token ' + token,
+                'User-Agent': 'buildbot'
+            })
         self.verbose = verbose
 
     def createStatus(self,
--- a/master/buildbot/test/unit/test_reporter_github.py
+++ b/master/buildbot/test/unit/test_reporter_github.py
@@ -41,7 +41,10 @@ class TestGitHubStatusPush(unittest.TestCase, ReporterTestMixin):
         yield self.master.startService()
         self._http = yield fakehttpclientservice.HTTPClientService.getFakeService(
             self.master, self,
-            HOSTED_BASE_URL, headers={'Authorization': 'token XXYYZZ'})
+            HOSTED_BASE_URL, headers={
+                'Authorization': 'token XXYYZZ',
+                'User-Agent': 'buildbot'
+            })
         self.sp = sp = GitHubStatusPush('XXYYZZ')
         sp.sessionFactory = Mock(return_value=Mock())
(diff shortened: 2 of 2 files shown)


tardyp:
Hi thanks for the fix. Is this using treq or txrequests?

hcwiley:
hey @tardyp, its using treq

tardyp:
I think it might make sense to put this as a default header in httpClientService for all reporters even for txrequests backend.

hcwiley:
i'll defer to your judgement here. Its required for all github api calls now. Not sure if there's other api that's require it as well, if there are others then yes I think it makes sense to put the header at a lower level.

master/buildbot/reporters/github.py
```diff
@@ -50,7 +50,10 @@ def reconfigService(self, token,
             baseURL = baseURL[:-1]
 
         self._http = yield httpclientservice.HTTPClientService.getService(
-            self.master, baseURL, headers={'Authorization': 'token ' + token})
+            self.master, baseURL, headers={
+                'Authorization': 'token ' + token,
+                'User-Agent': 'buildbot'
```
sa2ajj: Two comments here:

* it's `Buildbot`
* personally, I'd prefer to include the version

hcwiley: ok changed to "Buildbot". Not that any of this matters because the User-agent isn't being used for anything other than preventing the `403` response from github. In the github example its: `User-Agent: Awesome-Octocat-App`

If we are in the mood for *really* fixing things... https://github.com/Scandy-co/buildbot/blob/master/master/buildbot/reporters/github.py#L132 this has been returning a `403` error, but there's no error handling being done. so looking at the logs, it looks like it worked. going to github... there's no status updates

I made a temporary GitHubStatusPush to diagnose this:

```python
                res = yield self.createStatus(
                    repo_user=repoOwner.encode('utf-8'),
                    repo_name=repoName.encode('utf-8'),
                    sha=sha.encode('utf-8'),
                    state=state.encode('utf-8'),
                    target_url=build['url'].encode('utf-8'),
                    context=context.encode('utf-8'),
                    description=description.encode('utf-8')
                )
                res = yield res.json()
                if self.verbose:
            		  log.msg(
            		        'res: "{res}"'
            		        .format(
            		            res=res))
```

tardyp:
Indeed, looking at the code, it looks the error handling code assumes http.post raise exception in case of error which is not the case. we can fix it by printing an error in case the res.code is different from 200, where res is the result from sendStatus.

hcwiley:
@tardyp 👍 , that's what I was thinking. again, not knowing the Buildbot philosophy on error states and handling I didn't want to start slinging out my own stuff. but basically yes, if ```res.code >= 300``` 🔥

tardyp:
Hi @hcwiley Can you take some time to update the tests for this PR? We would love to get this fix merged for 0.9.2 that is due for this week.

hcwiley:
Hey @tardyp I thought I did here: https://github.com/Scandy-co/buildbot/commit/20214c27fccdb1c2851eebf0d331d35aa0151c90

What test is still failing because of this change?

master/buildbot/test/unit/test_reporter_github.py
```diff
@@ -41,7 +41,10 @@ def setUp(self):
         yield self.master.startService()
         self._http = yield fakehttpclientservice.HTTPClientService.getFakeService(
             self.master, self,
-            HOSTED_BASE_URL, headers={'Authorization': 'token XXYYZZ'})
+            HOSTED_BASE_URL, headers={
+                'Authorization': 'token XXYYZZ',
+                'User-Agent': 'buildbot'
```
tardyp: the user-agent is 'Buildbot' and not 'buildbot' hence the tests are failing.

hcwiley: Ahhh right, forgot I was asked to change that. Thanks

tardyp:
I did the change myself. Its pending CI

Le mar. 6 déc. 2016 à 14:57, H. Cole Wiley <notifications@github.com> a
écrit :

hcwiley:
:+1:

PR #2511, opened 2016-11-21
build towncrier in precommit

tardyp:
Looks like I messed up the commit, and forgot to include the makefile part


tardyp:
Problem is that towncrier generates a relnote that does not pass the spell checks.

sa2ajj:
(Side note: I could not possibly add `in progress` label at the time it says I did)

sa2ajj:
(it's probably via waffle integration, removed)

master/docs/Makefile
```diff
@@ -48,38 +48,41 @@ help:
 clean:
 	-rm -rf $(BUILDDIR)/*
 
-html: conf.py
+towncrier:
+	cd ../../; yes n | towncrier
```
sa2ajj: this look hackish: no way to provide `yes` to `towncrier` otherwise?

PR #2514, opened 2016-11-23
DockerLatentWorker: fix threading issues

dominiquehunziker:
Previously, the function `_thd_start_instance` was modifying the member variables `binds` and `volumes`. Because the function is running in a separate thread it is possible that multiple threads are trying to modify the variables at the same time. Switching to thread-local variables fixes this issue.

### DIFF
--- a/master/buildbot/newsfragments/thread_safety_of_dockerlatentworker.bugfix
+++ b/master/buildbot/newsfragments/thread_safety_of_dockerlatentworker.bugfix
@@ -1,0 +1,1 @@
+fix threading issue in :py:class:`~buildbot.plugins.worker.DockerLatentWorker`
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -15,7 +15,12 @@
 
 
 class Client(object):
+    latest = None
+
     def __init__(self, base_url):
+        Client.latest = self
+        self.call_args_create_container = []
+        self.call_args_create_host_config = []
         self._images = [{'RepoTags': ['busybox:latest', 'worker:latest']}]
         self._containers = {}
 
@@ -49,15 +54,16 @@ class Client(object):
         return self._containers.values()
 
     def create_host_config(self, *args, **kwargs):
-        pass
+        self.call_args_create_host_config.append(kwargs)
 
     def create_container(self, image, *args, **kwargs):
+        self.call_args_create_container.append(kwargs)
         name = kwargs.get('name', None)
         if 'buggy' in image:
-            raise Exception("we could not create this container")
+            raise Exception('we could not create this container')
         for c in self._containers.values():
             if c['name'] == name:
-                raise Exception("cannot create with same name")
+                raise Exception('cannot create with same name')
         ret = {
             'Id':
             '8a61192da2b3bb2d922875585e29b74ec0dc4e0117fcbf84c962204e97564cd7',
@@ -67,7 +73,7 @@ class Client(object):
             'started': False,
             'image': image,
             'Id': ret['Id'],
-            "name": name
+            'name': name
         }
         return ret
 
--- a/master/buildbot/test/unit/test_worker_docker.py
+++ b/master/buildbot/test/unit/test_worker_docker.py
@@ -98,7 +98,6 @@ class TestDockerLatentWorker(unittest.SynchronousTestCase):
         self.assertEqual(bs.command, ['/bin/sh'])
         self.assertEqual(bs.dockerfile, "FROM ubuntu")
         self.assertEqual(bs.volumes, [])
-        self.assertEqual(bs.binds, {})
         self.assertEqual(bs.client_args, {
                          'base_url': 'unix:///var/run/docker.sock', 'version': '1.9', 'tls': True})
         self.assertEqual(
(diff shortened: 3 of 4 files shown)


master/buildbot/newsfragments/thread_safety_of_dockerlatentworker.bugfix
```diff
@@ -0,0 +1 @@
+The function _thd_start_instance in :py:class:`~buildbot.plugins.worker.DockerLatentWorker` uses now thread-local variables to hold the binds and volumes arguments. Previously member variables were used which is not thread-safe.
```
tardyp: This is too much details.

Just use: fix threading issue in DockerLatentWorker

The details belongs to the commit messages

master/buildbot/worker/docker.py
```diff
@@ -148,7 +147,8 @@ def reconfigService(self, name, password, docker_host, image=None, command=None,
             self.client_args['tls'] = tls
 
     def parse_volumes(self, volumes):
-        self.volumes = []
+        volume_list = []
```
tardyp: Please rename this function _thd_parse_volumes() to tell that this method must take care of thread safety

dominiquehunziker:
I addressed the comments and I added a fix for the unittests. Please let me know if the unittest fix is ok.

I can merge all changes in a single commit if this is desired.

master/buildbot/test/unit/test_worker_docker.py
```diff
@@ -29,6 +29,12 @@
 from buildbot.worker import docker as dockerworker
 
 
+_thd_parse_volumes__orig = dockerworker.DockerLatentWorker._thd_parse_volumes
+def _thd_parse_volumes(self, volumes):
+    self._test_volumes, self._test_binds = retval = _thd_parse_volumes__orig(self, volumes)
```
tardyp: This patching looks a bit ugly to me. I think it would be better to assert that create_container has been called properly.

You can modify the fake docker in order to record any call to create container, and test this instead. (this probably should have been tested like this from the start)

dominiquehunziker: I'm not sure how this would work.

The create_container call is a member of the docker.Client, but the Client object gets instantiated in _thd_start_container and is only available/valid for the duration of the function. Does this mean I have to store the created objects in a global/class variable in order to access/assert it in the test function later on or is there a better way?

dominiquehunziker:
I'm not sure we were talking about the same issue. Therefore, some quick changes such that we can discuss the code directly...

master/buildbot/test/fake/docker.py
```diff
@@ -15,8 +15,12 @@
 
 
 class Client(object):
+    latest = None
 
     def __init__(self, base_url):
+        Client.latest = self
+        self.call_args_create_container = []
```
sa2ajj: my usual question for this kind of data members: is `self.call_args_create_container` a part of the API? in other words, do we expect outside world to use these data members?  if not, then make them private (= prepend their names w/ `_`)

master/buildbot/test/fake/docker.py
```diff
@@ -15,8 +15,12 @@
...
+        self.call_args_create_host_config = []
```
sa2ajj: ditto

master/buildbot/test/fake/docker.py
```diff
@@ -41,8 +45,9 @@ def build(self, fileobj, tag):
             self._images.append({'RepoTags': [tag + ':latest']})
 
     def create_host_config(self, *args, **kwargs):
-        pass
+        self.call_args_create_host_config.append(kwargs)
 
-    def create_container(self, image, *args, **kwargs):
+    def create_container(self, *args, **kwargs):
```
sa2ajj: this is a change of method signature (`image` is no more welcome): is it OK?  is there a possibility that somebody uses this parameter for something outside of the Buildbot code base?

master/buildbot/test/fake/docker.py
dominiquehunziker: that depends. It's not part of the original docker.Client API, but it is part of the public fake.Client API (it is used in the unit tests to access the arguments used when calling create_container & create_host_config). So, are we talking about the API of docker.Client or fake.Client?

master/buildbot/test/fake/docker.py
dominiquehunziker: image is still a valid argument. `*args, **kw` catches all possible arguments; therefore, image is still a valid argument. It's just not used in the create_container implementation so there is no need to explicitly catch the argument. I can revert the change if this is a concern.

dominiquehunziker: I guess it was explicitly mentioned because image is the only positional argument in the real docker.Client API and all other arguments to create_container are keyworded arguments. So, it would be feasible now to call create_container like this `create_container(1, image=some_image)`, whereas before this was not possible. I will revert the change once we clarified the other issues.

tardyp: looks good to me

tardyp:
I'll let you @sa2ajj confirm and merge this

dominiquehunziker:
I rebased to the current HEAD and squashed everything into one commit. Note that there were some conflicts during the rebase.

PR #2521, opened 2016-11-29
fix CID 1382001: Attribute/item access or function call before check for None or undefined

sa2ajj:
I am not 100% sure this is the right fix (hence a cryptic commit message)

(`Do not merge` to give me a chance to fix the commit message)

---

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/ws.py
+++ b/master/buildbot/www/ws.py
@@ -79,7 +79,7 @@ class WsProtocol(WebSocketServerProtocol):
             return
 
         # if it's already subscribed, don't leak a subscription
-        if path in self.qrefs:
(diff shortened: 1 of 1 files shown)


seankelly:
Seems like it would be better to reset `qrefs` to `{}`. On line 93, `self.qrefs` is checked to see if it's None but on line 96 it's assumed to be a dict anyway.

The attribute is only set to `None` when the connection is lost. Does any of this code run after?

tardyp:
I think the real fix is to set it to {} on connectionLost

tardyp:
@sa2ajj I used the new edit PR feature from GH in order to propose an improvement to your fix (there is another null pointer exception after in the function)

PR #2523, opened 2016-12-01
Don't recommend using addCallback as a function decorator

wallrj:
``addCallback`` returns a ``Deferred``.

A decorator should return a callable.

```
In [10]: d = succeed(object())

In [11]: @d.addCallback
def foo(result):
    pass
   ....: 

In [12]: foo
Out[12]: <Deferred at 0x7f6d5ca75170 current result: None>
```

### DIFF
--- a/master/docs/developer/style.rst
+++ b/master/docs/developer/style.rst
@@ -96,6 +96,7 @@ code for a particular function or method within the same indented block, using
 nested functions::
 
     def getRevInfo(revname):
+        # for example only! See above a better implementation with inlineCallbacks
         results = {}
         d = defer.succeed(None)
         def rev_parse(_): # note use of '_' to quietly indicate an ignored parameter
@@ -121,6 +122,15 @@ Deferred.  As a shortcut, ``d.addCallback`` works as a decorator::
     @d.addCallback
     def rev_parse(_): # note use of '_' to quietly indicate an ignored parameter
         return utils.getProcessOutput(git, [ 'rev-parse', revname ])
+
+.. note::
+
+    ``d.addCallback`` is not really a decorator as it does not return a modified function.
+    As a result in previous code, ``rev_parse`` value is actually the Deferred.
+    In general the :class:`inlineCallbacks` method is preferred inside new code as it keeps the code easier to read.
+    As a general rule of thumb, when you need more than 2 callbacks in the same method, its time to switch it to  :class:`inlineCallbacks`.
+    This would be for example the case for previous :py:func:`getRevInfo`.
+    See this `discussion <https://github.com/buildbot/buildbot/pull/2523>`_ with twisted experts for more information.
 
 Be careful with local variables. For example, if ``parse_rev_parse``, above,
 merely assigned ``rev = res.strip()``, then that variable would be local to
@@ -175,6 +185,19 @@ it is easy to represent a loop, or even nested loops, in this style without
 losing any readability.
 
 Note that code using ``deferredGenerator`` is no longer acceptable in Buildbot.
+
+The previous :py:func:`getRevInfo` example implementation should rather be written as:
+
+.. code-block::
+
+    @defer.inlineCallbacks
+    def getRevInfo(revname):
+        results = {}
+        res = yield utils.getProcessOutput(git, [ 'rev-parse', revname ])
+        results['rev'] = res.strip()
+        res = yield utils.getProcessOutput(git, [ 'log', '-1', '--format=%s%n%b', results['rev'] ])
+        results['comments'] = res.strip()
+        defer.returnValue(results)
 
 Locking
 .......
--- a/master/docs/developer/style.rst
+++ b/master/docs/developer/style.rst
@@ -82,7 +82,10 @@ In all cases, please prefer maintainability and readability over performance.
 Nested Callbacks
 ................
 
-First, an admonition: do not create extra class methods that represent the continuations of the first::
+First, an admonition: do not create extra class methods that represent the continuations of the first:
+
+
+.. code-block:: python
 
     def myMethod(self):
         d = ...
@@ -93,7 +96,9 @@ First, an admonition: do not create extra class methods that represent the conti
 Invariably, this extra method gets separated from its parent as the code
 evolves, and the result is completely unreadable. Instead, include all of the
 code for a particular function or method within the same indented block, using
-nested functions::
+nested functions:
+
+.. code-block:: python
 
     def getRevInfo(revname):
         # for example only! See above a better implementation with inlineCallbacks
@@ -116,7 +121,9 @@ nested functions::
 
 it is usually best to make the first operation occur within a callback, as the
 deferred machinery will then handle any exceptions as a failure in the outer
-Deferred.  As a shortcut, ``d.addCallback`` works as a decorator::
+Deferred.  As a shortcut, ``d.addCallback`` works as a decorator:
+
+.. code-block:: python
 
     d = defer.succeed(None)
     @d.addCallback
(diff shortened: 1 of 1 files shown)


tardyp:
Well, I think this is a useful style for writing inline anonymous(ish) callabacks, so I'll be inclined to keep this style.
Those functions are not supposed to be reused outside of the addCallbacks, so I'd say it's okay that addCallbacks does not return the callback.

tardyp:
@wallrj I don't want to forcefully close this PR as you may have more arguments against this practice.
Here is an email on python-idea suggesting that this is an existing practice (I couldn't find another)
https://mail.python.org/pipermail/python-ideas/2007-December/001279.html

Please let me know why you think this is a bad idea.
Apart from its not really a decorator, if this is the only argument, then we can change this PR with a warning about that.

wallrj:
Hey @tardyp 

I've tried using this style before on https://github.com/ClusterHQ/flocker or https://github.com/twisted/twisted/ and it was rejected by @exarkun for the same reasons, I think (but I can't find the reference. )
Perhaps I was doing something subtly different.

Then a colleague of mine, @moshez used this style recently on a private repo and I grumbled about it.

I found a couple of examples, introduced by @glyph in the Twisted code:
 * https://github.com/twisted/twisted/pull/548
 * https://github.com/twisted/twisted/commit/e331b4e5541a2a567f4dd2a54b2d46e675711432

This buildbot documentation is the only place I can find it discussed / documented so I decided to start a discussion :-)

I guess the main objection is that `addCallback`, `addErrback` and `addBoth` weren't intended to be used this way and `addCallbacks` can't be used this way....which makes it a bit of a hack.

tardyp:
Since we are adding other people, adding @djmitche who write the recommendation in the first place (the git blame says it came from our wiki)

I agree its a bit of a hack, but it imho adds readability compared to

    def postProcess(res):
          pass
    d.addCallback(postProcess)

Which is written backward imho.

exarkun:
It's true the trick does make the order more readable for some folks.  However, the resulting code is still rather dense and can be difficult to read _overall_.  A recommendation I would make is that functions should focus on doing one thing well.  A function that starts an async operation and then is packed full of nested functions for operating on the result may be just as poorly factored as a synchronous function that does ten unrelated operations.

Instead of defining `postProcess` inline, consider whether it can be factored out.  Then the code in question is just:

```
    d.addCallback(parseCommandIntoTestResult)
```

which I think we should be able to agree is more readable than either of the other options presented here.

glyph:
Personally I like this style very much.  For a while I worried that this was a too-clever impulse (I do have those sometimes) and an abuse of syntax; however, recent teaching experiences have confirmed my suspicions that explaining the `@` idiom once is much easier than repeatedly having to point out the subtle ordering issue presented by callbacks appearing before the method invocation that kicks them off.  It's a "weird once" vs. "weird every time" sort of difference.

Combined with another (unfortunately somewhat obscure) idiom, it can avoid redundant temporary naming as well as putting things in the right order.

For example, instead of:

```python
d = foo()
def bar(result):
    return doStuff(result)
return d.addCallback(bar)
```

you can do

```python
passthru = lambda x:x # (or 'from twisted.internet.defer import passthru')
@passthru(foo())
def d(result):
    return doStuff(result)
return d
```

Notice that `bar` disappears here.

_However_, I'd agree with @exarkun in that if you're inclined to have an intense discussion about the merits of this style, chances are that the real issue is that you've got too many callbacks, nested too deeply, and you should consider breaking things up into more top-level functions instead.  Buildbot certainly has that problem regardless of its stylistic representation :-).

tardyp:
I think the problem of nested function, and callback hell is orthogonal to this debate. 

With or without @d.addCallback, we all agree that more than 2 inline functions is to be avoid in python code (which btw is far from being the case in the example given https://github.com/twisted/twisted/commit/e331b4e5541a2a567f4dd2a54b2d46e675711432)

Buildbot used to suffer from callback hell, and I think we removed most of that technical debt from the core (we still have a lot of the source steps which is quite hard to follow)

We replaced lots of the callback hell code with inlineCallbacks logic.
This helps a lot readability, but in exchange I fear we traded performance. I may want to reintroduce some callbacks in the critical paths.

glyph:
> This helps a lot readability, but in exchange I fear we traded performance. I may want to reintroduce some callbacks in the critical paths.

If you're concerned about performance, the thing to introduce is benchmarks, not callbacks :-).  PyPy might be the right option, or shifting back to callbacks, or perhaps the lower-overhead async/await trampoline in 3.5 instead of the yield-based trampoline (i.e. switch to `ensureDeferred` instead of `inlineCallbacks`).

tardyp:
@glyph we do have live profiling:
https://pypi.python.org/pypi/buildbot-profiler/

I was surprised to see inlineCallback trampoline high in some of those profile, but I did not investigate further enough yet. It wasn't obvious which code path(s) those traces were about

tardyp:
@wallrj I updated your branch with a (opinionated I concede) summary of the discussion

master/docs/developer/style.rst
```diff
@@ -122,6 +123,15 @@ Deferred.  As a shortcut, ``d.addCallback`` works as a decorator::
...
         return utils.getProcessOutput(git, [ 'rev-parse', revname ])
 
+.. note::
+
+    ``d.addCallback`` is not really a decorator as it does not return a modified function.
+    As a result in previous code, ``rev_parse`` value is actually the Deferred.
+    In general the :class:`inlineCallbacks` method is preferred inside new code as it keeps the code easier to read.
+    As a general rule of thumb, when you need more than 2 callbacks in the same method, its time to switch it to  :class:`inlineCallbacks`.
+    This would be for example the case for previous :py:func:`getRevInfo`.
+    See this `discussion <https://github.com/buildbot/buildbot/pull/2523>`_ with twisted experts for more information.
```
exarkun: Twisted is a proper noun (not an adjective:wink:).

tardyp: ahah! right. You made my day. You guys are really twisted!

master/docs/developer/style.rst
```diff
@@ -122,6 +123,15 @@ Deferred.  As a shortcut, ``d.addCallback`` works as a decorator::
     def rev_parse(_): # note use of '_' to quietly indicate an ignored parameter
         return utils.getProcessOutput(git, [ 'rev-parse', revname ])
 
+.. note::
+
+    ``d.addCallback`` is not really a decorator as it does not return a modified function.
+    As a result in previous code, ``rev_parse`` value is actually the Deferred.
+    In general the :class:`inlineCallbacks` method is preferred inside new code as it keeps the code easier to read.
```
wallrj: ❓ shouldn't this be something like 
```
twisted:func:`inlineCallbacks` 
```

wallrj: Otherwise how will it know where to link to when rendering?

master/docs/developer/style.rst
```diff
@@ -122,6 +123,15 @@ Deferred.  As a shortcut, ``d.addCallback`` works as a decorator::
...
+    As a general rule of thumb, when you need more than 2 callbacks in the same method, its time to switch it to  :class:`inlineCallbacks`.
```
wallrj: ❓ Same here `:class:...` > `...:func:...`

master/docs/developer/style.rst
```diff
@@ -122,6 +123,15 @@ Deferred.  As a shortcut, ``d.addCallback`` works as a decorator::
...
+    This would be for example the case for previous :py:func:`getRevInfo`.
```
wallrj: ❓ And as i understand it, this would attempt to link to the Python docs....should probably just be ``getRevInfo``

master/docs/developer/style.rst
```diff
@@ -176,6 +186,19 @@ losing any readability.
 
 Note that code using ``deferredGenerator`` is no longer acceptable in Buildbot.
 
+The previous :py:func:`getRevInfo` example implementation should rather be written as:
```
wallrj: ❓ Same here.

master/docs/developer/style.rst
tardyp: It does not this is just used for styling. We don't have back reference helpers to twisted docs

wallrj:
👍

PR #2524, opened 2016-12-03
RFC: Authentication for read only access to Web UI

bendikro:
This is most definitely not mergeable as is, and is mainly a request for comments.

## Authentication for read only access to Web UI
I have tried to figure out how to enforce authentication for accessing the Web UI. The last trace I found of a discussion on this is in [these emails](http://comments.gmane.org/gmane.comp.python.buildbot.devel/11478) which indicate this is not supported yet.

This PR adds a call to _assertUserAllowed()_ in `www/config.IndexResource.renderIndex()`
which enforces authentication based on the allowRules. However, this changes the behavior of the current code such that AnyEndpointMatcher now matches any access to the Web UI.

According to the docs, the endpoint matchers are only responsible for the REST endpoints, so
I presume a proper solution would add a separate set of allowRules for viewing the different parts of the Web UI?

## oauth2.GitHubAuth
`https://api.github.com/users/{username}/orgs` only returns the organizations for the user that are publicly visible. Is that intended?
This has been changed to `https://api.github.com/user/orgs' which returns all organziations for the user (requires scope=user)

## oauth2.BitbucketAuth
Implemented

## Contributor Checklist:

## Feedback

Please give feedback on how I can improve this, or even better, make the required changes yourself if you desire.

### DIFF
--- a/master/buildbot/www/config.py
+++ b/master/buildbot/www/config.py
@@ -24,6 +24,7 @@ from twisted.web.error import Error
 from buildbot.interfaces import IConfigured
 from buildbot.util import json
 from buildbot.www import resource
+from buildbot.www.authz import Forbidden
 
 
 class IndexResource(resource.Resource):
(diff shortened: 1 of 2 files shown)


master/buildbot/www/config.py
```diff
@@ -135,12 +182,6 @@ def renderIndex(self, request):
         if 'change_hook_dialects' in config:
             del config['change_hook_dialects']
 
-        def toJson(obj):
-            obj = IConfigured(obj).getConfigDict()
-            if isinstance(obj, dict):
-                return obj
-            return repr(obj) + " not yet IConfigured"
-
         tpl = self.jinja.get_template('index.html')
```
tardyp: This is how the template is found from the main www module

tardyp:
I believe this has be splitted into several PR alread

PR #2531, opened 2016-12-08
trigger: naming of triggered build result depending on triggered build not on overall status

noc0lour:
Currently the exposed URLs in `step['urls']` have a `name` key which has a prepended result string which is derived from the current status of the triggerbuild I think.  This makes trigger step results i.e. in reporters useless because if one build inside the trigger fails all builds are prepended with `failure: `. To make life easier for mail report generation I changed the corresponding code to use the results of the triggered builds to generatio the results string.

I am aware that tests with `waitForFinish` in TriggerSteps are failing now and that's the reason why  I submit this PR now. I'm not sure how to `fix` the tests to resemble the same result. To pass checks for generated URLs the corresponding builds for each trigger would have to be marked with `SUCCESS/FAILURE` in the db to receive URLs with `success/failure` in them instead of `not finished`.

For me this PR is now working and user are happy to receive meaningful emails telling them which of the triggered builds failed.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -216,7 +216,7 @@ class Trigger(BuildStep):
                     for build in builds:
                         num = build['number']
                         url = self.master.status.getURLForBuild(builderid, num)
-                        yield self.addURL("%s: %s #%d" % (statusToString(results),
+                        yield self.addURL("%s: %s #%d" % (statusToString(build["results"]),
                                                           builderDict["name"], num), url)
 
     @defer.inlineCallbacks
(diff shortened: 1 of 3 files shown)


master/buildbot/steps/trigger.py
```diff
@@ -216,7 +216,7 @@ def addBuildUrls(self, rclist):
                     for build in builds:
                         num = build['number']
                         url = self.master.status.getURLForBuild(builderid, num)
-                        yield self.addURL("%s: %s #%d" % (statusToString(results),
+                        yield self.addURL("%s: %s #%d" % (statusToString(build["results"]),
```
tardyp: Indeed, I think this is how it should behave

noc0lour:
The thing is. I'm not sure how to fix the tests ;) Should I just write a helper function which changes the test results of the builds in the database which belong to a certain trigger scheduler?

noc0lour:
Awesome! thanks for finishing up :)

PR #2532, opened 2016-12-09
fix for older version of sqlite

tardyp:
Fix for buildbot eight CI regression


tardyp:
I guess I'll have to fix everything at once

PR #2534, opened 2016-12-12
[RFC] Auth check on REST API

bendikro:
As per [this comment](https://github.com/buildbot/buildbot/pull/2524#pullrequestreview-11279351) adding authentication check on the REST API is the most important step to improve security.

Would this be close to a solution? This does change the behavior of the current `AnyEndpointMatcher`, making it even more strict that before.

### DIFF
--- a/master/buildbot/newsfragments/www_rest_auth.feature
+++ b/master/buildbot/newsfragments/www_rest_auth.feature
@@ -1,1 +1,1 @@
-Implement support for authentication in REST API in :py:class:`~buildbot.www.rest.V2RootResource`.
+Buildbot can know be configured to deny read access to REST api resources based on authorization rules.
--- a/master/buildbot/newsfragments/www_rest_auth.feature
+++ b/master/buildbot/newsfragments/www_rest_auth.feature
@@ -1,1 +1,1 @@
-Buildbot can know be configured to deny read access to REST api resources based on authorization rules.
+Buildbot can now be configured to deny read access to REST api resources based on authorization rules.
--- a/master/buildbot/newsfragments/www_rest_auth.feature
+++ b/master/buildbot/newsfragments/www_rest_auth.feature
@@ -1,0 +1,1 @@
+Implement support for authentication in REST API in :py:class:`~buildbot.www.rest.V2RootResource`.
--- a/master/buildbot/www/rest.py
+++ b/master/buildbot/www/rest.py
@@ -337,6 +337,8 @@ class V2RootResource(resource.Resource):
             request.write(json.dumps(dict(error=msg)))
 
         with self.handleErrors(writeError):
+            yield self.master.www.assertUserAllowed(request, tuple(request.postpath), request.method, {})
+
             ep, kwargs = self.getEndpoint(request)
 
             rspec = self.decodeResultSpec(request, ep)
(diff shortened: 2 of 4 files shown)


master/buildbot/www/authz/endpointmatchers.py
```diff
@@ -174,17 +174,19 @@ def match_BuildEndpoint_rebuild(self, epobject, epdict, options):
...
+    def __init__(self, **kwargs):
         EndpointMatcherBase.__init__(self, **kwargs)
-        self.branch = branch
-        self.project = project
-        self.builder = builder
+
+    def match(self, ep, action="get", options=None):
+        if not "builds" in ep:
+            return None
+        return defer.succeed(Match(self.master))
```
tardyp: this will not restrict all build elements. Only those that you can access via "builds/xxx"
but you can also access logs via "/logs/:logid", which will contain probably the data you want to restrict. The logids are easy to guess as they are incremental

master/buildbot/www/rest.py
```diff
@@ -337,6 +337,8 @@ def writeError(msg, errcode=404, jsonrpccode=None):
             request.write(json.dumps(dict(error=msg)))
 
         with self.handleErrors(writeError):
+            yield self.master.www.assertUserAllowed(request, tuple(request.postpath), request.method, {})
```
tardyp: looks good

master/buildbot/www/authz/endpointmatchers.py
bendikro: Yes, indeed. I'll remove this change as it needs more work.

master/buildbot/www/rest.py
bendikro: Any comment on this change causing new behavior with existing configs using `AnyEndpointMatcher`?

bendikro: Ok, updated the tests to work, and added a test `test_authz_forbidden`

master/buildbot/newsfragments/www_rest_auth.feature
```diff
@@ -1 +1 @@
-Implement support for authentication in REST API in :py:class:`~buildbot.www.rest.V2RootResource`.
\ No newline at end of file
+Buildbot can know be configured to deny read access to REST api resources based on authorization rules.
```
tardyp: probably I did the spell mistake, but its *now*

bendikro: Ah, indeed. Fixed.

bendikro:
@tardyp, you really should avoid merging fixup commits. Those are to easily show changes in the PR which get lost when continually squashing.

Easily squash these with:
`git rebase --interactive --autosquash`
or tell me to do it before a force push.

tardyp:
@bendikro We don't merge the PR via command line, this would be too much manual overhead. The branches are CI protected, and this would delay the merge by another 15min.

I prefer that the branch are forced push or that the fixup commit have a real commit message, describing the actual fixup. 

Please use this process for next submissions.

bendikro:
Ok, if you don't mind continuously amending and force pushing I'll start doing that.

Frodox:
Does it affect www/web-ui?
How can I enable it in config..? :)

bendikro:
It affects all REST calls, so if you have a `util.AnyEndpointMatcher(role="<some role>", defaultDeny=True)` at the end, all calls require authentication.

PR #2540, opened 2016-12-13
Gitlab reporter was missing in the setup.

brkindar:
The gitlabstatuspush reporter (mentioned in the doc http://docs.buildbot.net/latest/full.html#gitlabstatuspush) was not in the package. I assume that is the way to add it correctly.
## Contributor Checklist:

No more errors were reported by trial buildbot.tests although I'm not sure this is the right way to fix the bug.

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -305,6 +305,7 @@ setup_args = {
             ('buildbot.reporters.gerrit_verify_status', ['GerritVerifyStatusPush']),
             ('buildbot.reporters.http', ['HttpStatusPush']),
             ('buildbot.reporters.github', ['GitHubStatusPush']),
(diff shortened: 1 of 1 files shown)


master/setup.py
```diff
@@ -305,6 +305,7 @@ def define_plugin_entries(groups):
             ('buildbot.reporters.gerrit_verify_status', ['GerritVerifyStatusPush']),
             ('buildbot.reporters.http', ['HttpStatusPush']),
             ('buildbot.reporters.github', ['GitHubStatusPush']),
+            ('buildbot.reporters.gitlab.py', ['GitLabStatusPush']),
```
seankelly: This uses includes the `.py` at the end. Will this still work?

tardyp: arg. it wont!

PR #2543, opened 2016-12-14
Handle any exception in checkconfig

tardyp:
It can be very frustrating to not being able to start buildbot,
because there is a random error in the buildbot tac,
and if no message is displayed

- change in base to display error for buildbot start/upgrade, etc
- also change in checkconfig
- keep the quiet behaviour

### DIFF
--- a/master/buildbot/scripts/base.py
+++ b/master/buildbot/scripts/base.py
@@ -113,20 +113,25 @@ def isBuildmasterDir(dir):
     return True
 
 
-def getConfigFromTac(basedir):
+def getConfigFromTac(basedir, quiet=False):
     tacFile = os.path.join(basedir, 'buildbot.tac')
     if os.path.exists(tacFile):
         # don't mess with the global namespace, but set __file__ for
         # relocatable buildmasters
         tacGlobals = {'__file__': tacFile}
-        exec(open(tacFile).read(), tacGlobals)
+        try:
+            exec(open(tacFile).read(), tacGlobals)
+        except Exception as e:
+            if not quiet:
+                print(e)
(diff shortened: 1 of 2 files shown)


master/buildbot/scripts/base.py
```diff
@@ -113,20 +113,25 @@ def print_error(error_message):
...
     if os.path.exists(tacFile):
         # don't mess with the global namespace, but set __file__ for
         # relocatable buildmasters
         tacGlobals = {'__file__': tacFile}
-        exec(open(tacFile).read(), tacGlobals)
+        try:
+            exec(open(tacFile).read(), tacGlobals)
+        except Exception as e:
+            if not quiet:
+                print(e)
```
seankelly: The exception is printed in the checkconfig method and now also here. Could this result in an exception being printed twice?

master/buildbot/scripts/checkconfig.py
```diff
@@ -46,8 +46,8 @@ def checkconfig(config):
     if os.path.isdir(configFile):
         basedir = configFile
         try:
-            configFile = getConfigFileFromTac(basedir)
-        except (SyntaxError, ImportError) as e:
+            configFile = getConfigFileFromTac(basedir, quiet=quiet)
+        except (Exception) as e:
```
rutsky: Brackets around `(Exception)` are not needed anymore, since there is only one exception.

master/buildbot/scripts/base.py
rutsky: Why not something like `traceback.print_exc()` instead of `print(e)`?

PR #2544, opened 2016-12-14
fix sendchange --help's description for --encoding

Frodox:
remove extra one 'default', since it is added automatically

```
...
-e, --encoding=       Encoding of other parameters (default utf8) [default:
		              utf8]
```


tardyp:
thanks!

PR #2545, opened 2016-12-14
Fix duplicate worker usecase

tardyp:
## Contributor Checklist:

A worker with the same name would make the other worker also disconnect.

Fix the flakyness of the test case that tested this case.
(needed to force disconnection of the pb for both workers)


master/buildbot/test/integration/test_worker_comm.py
```diff
@@ -254,7 +254,8 @@ def logged_in(persp):
 
             return workerworker
 
-        self.connector = reactor.connectTCP("127.0.0.1", self.port, factory)
+        self.connectors.append(
+            reactor.connectTCP("127.0.0.1", self.port, factory))
```
djmitche: good call :)

PR #2546, opened 2016-12-16
Order the waterfall view by builder name.

chrwen-omicron:
I think the waterfall view would be more readable if the builders are sorted by name like the builders view.


master/buildbot/newsfragments/waterfall_view.feature
```diff
@@ -0,0 +1 @@
+-Sort the builders in the waterfall view by name instead of ID.
```
tardyp: please remove the heading '-'

chrwen-omicron: uups, my bad

PR #2549, opened 2016-12-21
fix bb.reporter.gitlab setup from #2540

Frodox:
Fix setup, so it works as expected now

## Contributor Checklist:


tardyp:
thanks!

PR #2553, opened 2016-12-30
buildPushesWithNoCommits also rebuilds known branch after update

Ricordel:
The buildPushesWithNoCommits in gitpoller.py determines if a push
on a new branch or update of an already known branch with already
known commits should trigger a build.

Until this, it worked fine for new branches, but would not trigger
a build if, for instance, you fast-forwarded your master branch onto
a topic branch for which a build had already happened. This commit
fixes this behaviour.


master/buildbot/newsfragments/fix-buildPushesWithNoCommit-for-merge.bugfix
```diff
@@ -0,0 +1 @@
+:py:class:`~buildbot.changes.GitPoller`: buildPushesWithNoCommits now rebuilds for a known branch that was updated to an existing commit.
```
tardyp: please double backquote (`) ``buildPushesWithNoCommits`` so that the spellchecker is happy

PR #2557, opened 2017-01-04
rename simple cases of "slave" to "worker" in contrib scripts

rutsky:


master/contrib/github_buildbot.py
```diff
@@ -5,7 +5,7 @@
 github_buildbot.py will determine the repository information from the JSON
 HTTP POST it receives from github.com and build the appropriate repository.
 If your github repository is private, you must add a ssh key to the github
-repository for the user who initiated the build on the buildslave.
+repository for the user who initiated the build on the Buildbot worker.
```
ewongbb: I think in this context, Buildbot isn't needed.

rutsky: @ewongbb done!

PR #2562, opened 2017-01-05
master/setup.py: Add missed testing templates

dol-sen:
This patch applied to buildbot-0.9.2 along with unpacking the missed files/dirs corrects the test failure I experienced.  This patch adds the installation of those missed files.  With this I can now run "trial buildbot" on the installed code which now completes successfully.  The tests run via the ebuild during the install process using "setup.py test" also pass.

The missing templates tarball I created for patching the source tarball can be found here:
http://dev.gentoo.org/~dolsen/distfiles/buildbot-test_templates.tar.xz

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -206,6 +206,8 @@ setup_args = {
         ]),
         include("buildbot/spec", "*.raml"),
         include("buildbot/spec/types", "*.raml"),
+        include("buildbot/test/unit/test_templates_dir", "*.html"),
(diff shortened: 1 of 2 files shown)


dol-sen:
This fixes the totally missed tests for worker.

The source tarball I used to unpack over the original tarball is located:
http://dev.gentoo.org/~dolsen/distfiles/buildbot_worker-9-tests.tar.xz

Ran 228 tests in 3.527s

PASSED (skips=1, successes=227)

PR #2563, opened 2017-01-05
Fix flake8 nits.

ewongbb:
This is the rebased version of flakechecks (which I kinda screwed up with merges).


rutsky:
@ewongbb how these changes will be enforced in the future?

tardyp:
looks good. @rutsky as soon as we have the newer flake8 to pass, we will upgrade the dep in [test]

PR #2566, opened 2017-01-06
Hg meta sub repo update

alerickson0:
Greetings,

This is the PR for my updates to the hgpoller.py module and other modules to support Mercurial meta repository sub repository functionality. Basically, the updates I have made allow the hgpoller to notice changes in any sub repositories.

Also, I had problems trying to update the unit tests to verify my updates. I kept getting errors when I tried to fix the tests. At the request of Pierre, I am submitting this PR so the Buildbot developers can look at my updates to the tests and to the source code.

Regards,

Alvaro

## Contributor Checklist:


tardyp:
Waw, this is a lot of work.

I am pretty uncomfortable with it, though.

- we try to avoid doing db changes especially if this is only for the support of one workflow (hg)

- We try to avoid recommending poller. It might makes sense when using a firewalled bot, but it is not scalable. Http hooks are much better design. So it makes me unconfortable to have sub-repo feature implemented first for a poller.

- We already have the codebase feature which has been designed especially for the hg subrepo.
http://docs.buildbot.net/latest/manual/concepts.html#multiple-codebase-builds
I think codebase is not completely finished design, and I'd like to avoid doing the same mistake of integrating another not fully designed concept addressing similar thing.
I think this should first be discussed (pep style https://www.python.org/dev/peps/) with a strong design document explaining why codebase does not fit the hg community needs, why it cannot be improved, and why including subrepo support in db is better

- The PR is far too big for review and testing. There are so much tests to fix it is difficult to start.
   When the design discussion is done, this should be split into several PR:
   - database changes, with unit tests and developer doc
   - data api changes, with unit tests and developer doc
   - core changes to let the core understand sub repo changes


I know this is asking a lot, and I am sad to welcome a new visibly talented developer in this community with so much requirements, but I don't want to make the mistake of accepting changes to the core without strong design.

alerickson0:
Hi Pierre,

Thanks for the feedback, also for the suggestions on what else I can use. I
will look into using the:

   - Http hooks or
   - The codebase feature for multiple-codebase-builds

And if these do not work for me, I will work on a design document
explaining why codebase does not fit the hg community needs, why it cannot
be improved, and why including sub repo support in the db is better.

Regards,

Alvaro

alerickson0:
I forgot to add a problem statement that better explain why I made these updates:

I am working on a project that has a Mercurial meta repository with
sub repositories. We want to do a standard build by pulling from all the meta repo's sub repos. Also, we have 3 problems to solve:

1. Meta-repository standard build - We want to build the meta-repo with the sub-repos as described, and want to analyze the sub-repo logs for blame-list generation.  This is like a standard build, but with sub-repo log inspection.

2. Tip build - We want to build from all of the sub-repo using the latest commit on the branch configured in Buildbot.  This is like a multiple codebase build?

3. Specific revision/changeset build - We want the developers to be able to provide a specific
set of branches to use for a tip build, in support of a feature branch development workflow.

tardyp:
@alerickson0 thanks for stating your requirements.

For 1), clearly this is not a very easy task to do generically in the core. The source stamp is already generating one revision for each codebase, but there is not really the data model necessary to handle links between the past revisions so that you can create a reliable blame log. That information is really in the SCM, so imho this is where to get it, not in the CI.
This is something I have implemented for git at work. We do it with a simple script that uses git in order to know the difference between two builds. I don't know hg, but I would expect it do generate logs easily between several builds, using the sourcestamps.

For 2) indeed, that is what the codebase is made for. Without changing the hg step, you need to create as much hg steps as you have codebases. I think this is easy change to make hg step more codebase aware, and select which subrepo the change is for. Then you will need to take care of the multi-project dependencies (if a commit in project 1 needs a commit in project 2, how do you make sure you pull the 2 projects at the same time?). This is something that I have been working on and have solution for (not yet upstream), but for Gerrit. Again I have no idea how hg works.

3) This is something that I have implemented for Gerrit and is much easier to do in a custom script imho. I created a force scheduler with a multiline text entry where developers are supposed to paste the links to their commits, and then I have a step which parses, and cherry-pick all those commits.

alerickson0:
@tardyp, thank you for your reply, I have some more questions:
1) For the blamelist generation, what is the best way to incorporate an external script into Buildbot? And how can an external script obtain access to data in Buildbot?
2) A Mercurial extension could be written, but how could we get Buildbot to recognize/call the extension?
3) When you recommend a custom script for my question 3., what is the best way to incorporate it with/in Buildbot?

tardyp:
The way to integrate scripts and buildbot is usually via Interpolate, in order to pass properties to step. But probably what you will need for you complex requirements is a custom step using the ShellMixin:

I would expect this documention to answer all your three questions. Let me know if it doesn't
http://docs.buildbot.net/latest/manual/customization.html#running-commands

For blamelistI wouldn't try and create changes from a step, I think the best way is to create a HtmlLog (using addHtmlLog() ) which will contain your diff log report, and probably update the property "owners" which is used by the blame list generator to find blame emails that weren't found by changes.

alerickson0:
@tardyp for your custom script using a Force Scheduler, how did you checkout/update the source code? Did you checkout using a standard Buildbot source checkout class? Or use a Python module?

alerickson0:
@tardyp thank you again for answering my questions, I have another question about your answer to my question 2) above:
Which scheduler do you recommend for the multicodebase? A Single Branch scheduler for each sub repo to watch for changes? Also, hg is similar to Git. You make a commit to your local repo. Then, push the change to the remote repo it is tied to. Now, I know the other sub repos to pull from. So, after a scheduler senses the change, how should I pull from the other sub repos?

tardyp:
Hi @alerickson0 The answer is really depending on your actual code.
Maybe you want to have a builder common to all your repository, and as many scheduler as you have repositories.
Maybe you want many builders, but the same steps.
Maybe you want many builders, with specific steps to optimize the testsuite for the commit you are testing.
As per how to fetch changes, again there are many options. 
You can have several hg steps, one per codebase
You can change the official Hg step to support multiple codebase and sub repo (this would be for me the best solution, but probably needs a little bit more investment in the begining)
You can also have a custom shell step or python script fetching the patch.

tardyp:
@alerickson0 So what should we do to make progress on this one?

alerickson0:
Hi Pierre,

So, I have worked on implementing some of the alternate solutions you gave
me like:

   - Creating a force scheduler with a multiline text entry where
   developers are supposed to paste the links to their commits, and then
   having a step which parses, and cherry-pick all those commits.
   - Having a hg step, one per codebase

All have worked for me. Now, I have not had a chance to try implementing
the http hooks to tell Buildbot if there any changes in the repositories.

Regards,

Alvaro

tardyp:
thanks @alerickson0 for the update. Closing this for now

PR #2570, opened 2017-01-08
update windows_service.py to handle new workers

rutsky:
Also update "slave"/"worker" naming.

## Contributor Checklist:


rutsky:
Oh, by the way, I haven't tested this change - I don't have proper environment to do that.

PR #2583, opened 2017-01-12
Setuptools is required

tomprince:
The flake8 dependency requires at least setuptools 8.0 to parse, so require that.

## Contributor Checklist:


master/setup.py
```diff
@@ -415,113 +416,98 @@ def define_plugin_entries(groups):
...
-    autobahn_ver = ">= 0.10.2, < 0.13.0"
-    txaio_ver = "== 2.2.2"
-
-else:
-    twisted_ver = ">= 14.0.1"
-    autobahn_ver = ">= 0.16.0"
-    txaio_ver = ">= 2.2.2"
+twisted_ver = ">= 14.0.1"
+autobahn_ver = ">= 0.16.0"
+txaio_ver = ">= 2.2.2"
```
tomprince: This isn't related, but was next to the code I was changing anyway.

tomprince:
The test failures look unrelated.

master/setup.py
```diff
@@ -415,113 +416,98 @@ def define_plugin_entries(groups):
...
-        # boto3 and moto required for running EC2 tests
-        'boto3',
-        'moto',
-        # txgithub required to run buildbot.status.github module tests
-        'txgithub',
+setup_args['tests_require'] = test_deps
+
+setup_args['extras_require'] = {
+    'test': [
+        'setuptools_trial',
```
tardyp: cool. I wasn't aware of it. @rutsky did try already to support setup.py test, but it didn't work

master/buildbot/newsfragments/setuptools-is-required.bugfix
```diff
@@ -0,0 +1 @@
+setuptools is now marked as required to install buildbot (non-setuptools install was not tested and broken).
```
tardyp: @tomprince I don't see the use of logging .misc newsfragments (git log is fine for that). But this fix is actually imo useful to heads up to users

master/setup.py
rutsky: Actually `setup.py test` works, but it can not handle side-by-side installation of `autobahn` due to `setuptools` bug: https://github.com/pypa/setuptools/issues/500
So if current Python environment have `autobahn` installed, `setup.py test` should handle side-by-side installation of other deps and properly run tests.

master/buildbot/newsfragments/setuptools-is-required.bugfix
tomprince: The point is you can add checks for newsfragments to CI, and you don't need an exception for things that don't need to be in the changelog. And they show up in the diff, so it is easier to remember about them in review and consider whether they do need a newsfragement.

tardyp:
ugh.. that would be 3 merge commits in a row. Could you rebase that properly?

PR #2584, opened 2017-01-12
Add absolute_import and print_function future imports.

tomprince:
I tripped over not having `absolute_import` in #2582, so add them everywhere (and insure they stay added).

Builds on #2582 since they touch the same imports.

## Contributor Checklist:


tardyp:
Looks like there are some old school prints remaining..

PR #2585, opened 2017-01-13
changes: add GitHubPullrequestPoller as changesource

noc0lour:
This is still WIP (tests are missing) and I'd like to get some comments if this Poller is written the right way.
I tried to use the (new) reconfigAPI to be able to use HTTPClient, in order to to that I had to make                   PollingChangeSource compatible with that (Basically handing more arguments and kwargs down to the BuildbotService class).

I'd be happy to get some comments & reviews on the changes. 

## Contributor Checklist:


master/buildbot/changes/github.py
```diff
@@ -0,0 +1,262 @@
...
+            branches = ['master']
+
+        self.token = token
+        self.owner = owner
+        self.repo = repo
+        self.branches = branches
+        self.project = project
+        self.encoding = encoding
+
+        if hasattr(pullrequest_filter, '__call__'):
```
tardyp: There is a 'callable' function in python std which does the same function

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,262 @@
...
+        self.project = ascii2unicode(project)
+
+    def describe(self):
+        return "GitHubPullrequestPoller watching the "\
+            "GitHub repository %s/%s, branch: %s" % (
+                self.owner, self.repo, self.branch)
+
+    def _getPulls(self):
+        self.lastPoll = time.time()
+        log.msg("GitHubPullrequestPoller: polling "
```
tardyp: I dont think it is a good idea to have a log for each and every polling.

Please use buildbot.util.logger for new code

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,262 @@
...
+                "GitHub repository %s/%s, branches: %s" %
+                (self.owner, self.repo, self.branches))
+        d = self._http.get(
```
tardyp: please use inlineCallbacks instead of addCallbacks

http://docs.buildbot.net/latest/developer/style.html#inlinecallbacks

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,262 @@
...
+            @email.addCallback
+            def return_email(result):
+                return result["email"]
+
+            return email
+
+        return d
+
+    def _getFiles(self, prnumber):
+        log.msg("GitHubPullrequestPoller: fetching changed files"
```
tardyp: no need to log that.

master/buildbot/test/unit/test_changes_github.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def setUp(self):
+        pass
+
+
+    def tearDown(self):
+        pass
+
+
+    def test_PR(self):
```
tardyp: ah. We will need more than that :)

You can look at the gerritpoller tests for examples

master/buildbot/changes/base.py
```diff
@@ -52,10 +52,13 @@ def _unclaimService(self):
 
 class PollingChangeSource(ChangeSource):
 
-    def __init__(self, name=None, pollInterval=60 * 10, pollAtLaunch=False):
-        ChangeSource.__init__(self, name=name)
-        self.pollInterval = pollInterval
-        self.pollAtLaunch = pollAtLaunch
+    def __init__(self, *args, **kwargs):
+        # Provide backwards compatibility
```
tardyp: I am not sure we really need this.

Anyway we eventually need to switch the change source to checkConfig/reconfigService

noc0lour: From what I know/tested I need the changes so args/kwargs are propagated to reconfigService correctly but we still have compatibility with other currently implemented Polling Sources.

master/buildbot/test/unit/test_changes_github.py
noc0lour: Yeah, I know and I'm planning on writing the tests ;) Basically just boilerplate right now

noc0lour:
Have some wip. Adding the unit tests will take at least a week or so, so don't throw this PR away ;)

noc0lour:
Rebased and reworked to pull in changes to `changes/base.py` and got first unit tests working. I'll update this PR by friday :), finally.

noc0lour:
@tardyp added tests and stuff is now based on `ReconfigPollingChangeSource` and uses `@defer.inlineCallbacks` decorators.

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,260 @@
...
+        self.lastPoll = time.time()
+        log.debug("GitHubPullrequestPoller: polling "
+                  "GitHub repository %s/%s, branches: %s" %
+                  (self.owner, self.repo, self.branches))
+        result = yield self._http.get(
+            '/'.join(['/repos', self.owner, self.repo, 'pulls']),
+            timeout=self.pollInterval)
+        my_json = yield result.json()
+        defer.returnValue(my_json)
+        return
```
tardyp: no need return when defer.returnValue

noc0lour:
Heh, left my branch on work machine, then I'll push monday.

noc0lour:
I guess I need 100% coverage?

master/buildbot/test/fake/httpclientservice.py
```diff
@@ -113,7 +113,7 @@ def assertNoOutstanding(self):
         self.case.assertEqual(0, len(self._expected),
                               "expected more http requests:\n {!r}".format(self._expected))
 
-    def _doRequest(self, method, ep, params=None, data=None, json=None):
+    def _doRequest(self, method, ep, params=None, data=None, json=None, timeout=None):
```
tardyp: if you are adding a requests based option here, you need to make sure it will also work with treq too (with unit test and integration tests proving that it works).
I think this belongs to another improvement PR

master/buildbot/test/unit/test_changes_github.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+        self._http.expect(
+            method='get',
+            ep='/repos/defunkt/defunkt/pulls',
+            content_json=json.loads(gitJsonPayloadPullRequest))
+        yield self.startChangeSource()
+        yield self.changesource.poll()
+        self.assertEqual(len(self.master.data.updates.changesAdded), 0)
+
+    def test_failFiles(self):
```
tardyp: please implement those tests or delete them..

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def _processChangesFailure(self, f):
+        log.err('GitHubPullrequestPoller: json api poll failed')
+        log.err(f)
+        return None
+
+    @defer.inlineCallbacks
+    def poll(self):
+        result = yield self._getPulls()
+        self._processChanges(result)
```
tardyp: probably there is a yield missing

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+
+                results = yield dl
+                failures = [r[1] for r in results if not r[0]]
+                if failures:
+                    for failure in failures:
+                        log.err(failure, "while processing changes for "
+                                "Pullrequest {} revision {}".format(prnumber,
+                                                                    revision))
+                        # Fail on the first error!
+                        failures[0].raiseException()
```
tardyp: probably we need a test on that too.
What are we doing in that case? are we retrying the same PR in the next poll?

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+        self.owner = owner
+        self.repo = repo
+        self.branches = branches
+        self.project = project
+        self.encoding = encoding
+        self.pollInterval = pollInterval
+
+        if callable(pullrequest_filter):
+            self.pullrequest_filter = pullrequest_filter
+        elif pullrequest_filter:
```
tardyp: that elif is not really needed
```
else:
    self.pullrequest_filter = (lambda _: pullrequest_filter)
```
Will work either case

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+        for pr in github_result:
+            # Track PRs for specified branches
+            base_branch = pr['base']['ref']
+            prnumber = pr['number']
+            revision = pr['head']['sha']
+
+            # Check to see if the branch is set or matches
+            if base_branch not in self.branches:
+                return
+            if self.pullrequest_filter and not self.pullrequest_filter(pr):
```
tardyp: ```if self.pullrequest_filter is not None and not self.pullrequest_filter(pr):```

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @defer.inlineCallbacks
+    def _getPulls(self):
+        self.lastPoll = time.time()
+        log.debug("GitHubPullrequestPoller: polling "
+                  "GitHub repository %s/%s, branches: %s" %
+                  (self.owner, self.repo, self.branches))
+        result = yield self._http.get(
+            '/'.join(['/repos', self.owner, self.repo, 'pulls']),
+            timeout=self.pollInterval)
```
tardyp: where does this timeout comes from? I think we can drop all those timeouts.
certainly not use the same value as pollInterval. maybe another configurable constant.

In my experience, increasing timeouts is putting dust under the carpet and will bite much harder later.

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+        self.project = ascii2unicode(project)
+
+    def describe(self):
+        return "GitHubPullrequestPoller watching the "\
+            "GitHub repository %s/%s" % (
+                self.owner, self.repo)
+
+    @defer.inlineCallbacks
+    def _getPulls(self):
+        self.lastPoll = time.time()
```
tardyp: this variable does not seems like used anywhere

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+        self.pollInterval = pollInterval
+
+        if callable(pullrequest_filter):
+            self.pullrequest_filter = pullrequest_filter
+        elif pullrequest_filter:
+            self.pullrequest_filter = (lambda _: pullrequest_filter)
+        else:
+            self.pullrequest_filter = None
+
+        self.lastChange = time.time()
```
tardyp: those two variables look like unused

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+                return
+            current = yield self._getCurrentRev(prnumber)
+            if not current or current[0:12] != revision[0:12]:
+                # Access title, repo, html link, and comments
+                branch = pr['head']['ref']
+                title = pr['title']
+                repo = pr['head']['repo']['name']
+                revlink = pr['html_url']
+                comments = pr['body']
+                if self.useTimestamps:
```
tardyp: why would you set useTimestamps to false?

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+                    branches=None,
+                    pollInterval=10 * 60,
+                    useTimestamps=True,
+                    category=None,
+                    baseURL=None,
+                    project='',
+                    pullrequest_filter=True,
+                    token=None,
+                    encoding='utf-8',
+                    pollAtLaunch=False,
```
tardyp: pollAtLaunch and pollInterval can be left in the kwargs (simplify code)

master/buildbot/changes/github.py
```diff
@@ -0,0 +1,259 @@
...
+                    **kwargs):
+        base.ReconfigurablePollingChangeSource.checkConfig(
+            self,
+            name=self.name,
+            pollAtLaunch=pollAtLaunch,
+            pollInterval=pollInterval)
```
tardyp: you forgot to pass kwargs (so that we can add feature to PollingChangeSource later and don't need to modify subclasses_

noc0lour:
No worries ;) I'll update the docs last when everything looks good enough.

tardyp:
Its good to have doc draft early to understand what you want to create

master/buildbot/test/fake/httpclientservice.py
tardyp: The goal of the fake is so that we can avoid to test the httpclientservuce users against treq and txrequests.

All the API that is on the fake is guaranteed to work on both implementation. It is a voluntarily subset of txrequests. If we start to add timeout, we start to break this guarantee

master/docs/relnotes/index.rst
```diff
@@ -10,6 +10,167 @@ Release Notes
 
 .. towncrier release notes start
 
+Buildbot ``0.9.4-111-g93f4b41`` ( ``2017-02-16`` )
```
tardyp: towncrier will modify this file when doing "make docs". Please do not include it in your commits

noc0lour: Yeah, overlooked that.

noc0lour:
Ha, maybe now it's good to go?

davidtgoldblatt:
I played around with this as I was setting up something similar. I think this sets the 'repository' build property incorrectly: for consistency with, e.g., the git poller, it should be 'git://github.com/\<owner name\>/\<repo name\>.git instead of just \<repo name\>. I think you want repo to be pr['head']['repo']['git_url'] instead of pr['head']['repo']['name']. (This matters if you want to be able to clone the repo by getting its name from a property in the builder).

noc0lour:
Good catch, probably it should be either user configurable or setting it correctly.

noc0lour:
@davidtgoldblatt Maybe even give the user a selection to choose one of those when creating the Poller? 
```
        "git_url": "git://github.com/gnuradio/gnuradio.git",
        "ssh_url": "git@github.com:gnuradio/gnuradio.git",
        "clone_url": "https://github.com/gnuradio/gnuradio.git",
        "svn_url": "https://github.com/gnuradio/gnuradio",
```

davidtgoldblatt:
I think in general the user won't know the urls in advance; a pull request can come from any repository.

noc0lour:
Right, I meant the $protocol_url entries. Since it depends on the user if he wants to use https, ssh or git to clone the repo.

noc0lour:
But basically you're right. I'll fix this up. in my installation it doesn't matter since all PRs are coming from the same repo but that's certainly not the case for most people.

davidtgoldblatt:
Oh, I see. Makes sense to me, but I also can't claim to know what I'm talking about here.

noc0lour:
You already pointed out the main flaw. Thanks, I'll fix it later and resubmit :)

noc0lour:
@davidtgoldblatt Since you want to use this changesource can you review this as well? I added selection for the repo link type as well. Thanks again!

davidtgoldblatt:
Thanks for doing all this, it really makes life much simpler for people like me.

The things I ran into trying this out:
- Maybe 'repository_type' over 'repository_link' for the parameter name? It's not really a link.
- I don't think there's a way to watch for pull requests to all branches. (E.g. in my project we have "dev", "master", and an "rc-\<something\>" branch, where the name of the last one changes over time).
- What about emitting the destination repository (or even the whole pull request? that's a little bit spewy though) as a build property? I'm trying to hook up a GitHubStatusPush on the other end of the pipeline, and it tries to send the status to the source of the PR, rather than the destination (which is where it would get displayed). This is arguably a bug in GitHubStatusPush, but it can't get the information it needs to fix it from the information currently exposed to it.

noc0lour:
1. repository_type will be better, I agree

2. Just don't set a `branches` argument and write a function to filter based on what you want in `pullrequest_filter`. 

3. I don't get your last point. Can you clarify?

I'd love to have a API backend for different Hosters (e.g. GitHub, GitLab, Bitbucket) and be consistent across Push, pull, webhook. then we can change thing in one place and they will be applied to all different services for the API.

davidtgoldblatt:
Re: `branches`: if branches is set to None, it gets changed to ['master'] [link](https://github.com/buildbot/buildbot/pull/2585/commits/e56bd69bd333224fd26aa5764a5a7f6a363d964a#diff-5ea4f760b8f332fe61cf7ba8d7d9d560R100), and anything not in the branches list is filtered regardless of pullrequest_filter [link](https://github.com/buildbot/buildbot/pull/2585/commits/e56bd69bd333224fd26aa5764a5a7f6a363d964a#diff-5ea4f760b8f332fe61cf7ba8d7d9d560R183). I think the semantics you described seem like a good idea.

Actually, looking at those sections more closely, I think we want `continue` rather than `return` as the bailout option, if the first pull request doesn't match the filter but a later one does.

Re: the last point: right now, this emits a change where the 'repository' property is the source of the pull request (as it should). There's no way to tell what the destination of the pull request is though, given only that change. GitHubStatusPush tries to push CI statuses to the repository indicated by the build, which is gleaned from the change that generated it (I believe; perhaps someone who knows the buildbot internals better will correct me). But for pushing a CI status to github, this is the wrong semantics -- statuses live with the repo that's the target of the pull request, not the source.

I think being platform independent makes sense; perhaps it's better to only include "target repository", and add more properties as they're needed.

noc0lour:
@davidtgoldblatt Yeah, continue instead of return is the right choice. And I shouldn't default to master if branches is not set either. If it is `None` I'll just process PRs for all branches.

Yeah, for the GithubAPI matching we need to figure out where to push the status and maybe unify our API calls to GithHub  API across webhook, push and pull

noc0lour:
@tardyp Now, I'm wondering if we should set branch to the "magic" ref link of GitHub as well, or make in configurable? So people are able to use the GitHubSource here as well?

tardyp:
@noc0lour 
> Now, I'm wondering if we should set branch to the "magic" ref link of GitHub as well, or make in configurable? So people are able to use the GitHubSource here as well?

I did not check this particular detail when reviewing your poller, but indeed, I would suggest to populate the magic branch correctly in the change.

https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/hooks/github.py#L126

master/buildbot/changes/github.py
```diff
@@ -97,16 +97,13 @@ def reconfigService(self,
...
-            branches = ['master']
-
         self.token = token
         self.owner = owner
         self.repo = repo
         self.branches = branches
         self.project = project
         self.pollInterval = pollInterval
-        self.repository_link = link_urls[repository_link]
+        self.repository_type = link_urls[repository_type]
```
tardyp: The hook does not have this feature.
It is possible that it is not compatible with the reporter which could only know about http repository type (to be verified)

noc0lour:
For now I'd like to add a switch so it is at least compatible with the GitHub Source.

noc0lour:
@tardyp I put in an option to enable GitHub magic links in case someone is using GitHub sources downstream.

PR #2589, opened 2017-01-14
Use twisted.python.compat.NativeStringIO when overriding stdout or stderr

rodrigc:
sys.stdout is bytes on Py2, and unicode on Py3,
so we need to be careful when overriding it.

### DIFF
--- a/master/buildbot/test/unit/test_scripts_base.py
+++ b/master/buildbot/test/unit/test_scripts_base.py
@@ -16,13 +16,13 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
-import cStringIO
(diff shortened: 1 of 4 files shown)


tardyp:
We've been trying to support very old version of twisted for the sake of supporting old distros testing.
We can still backport NativeStringIO for the worker package

rodrigc:
OK, I changed the patch to use the back to using the backport version inside
buildbot_worker/test/util/misc.py, so it will work on really old Twisted versions.

tardyp:
I dont see worker patch anymore. this is only for master but that is fine for one PR

PR #2596, opened 2017-01-15
Make sure that we pass bytes to hashlib.update().

rodrigc:
This eliminates errors on Python 3:
  "Unicode-objects must be encoded before hashing"


rodrigc:
Refactoring depends on https://github.com/buildbot/buildbot/pull/2600

PR #2597, opened 2017-01-15
Remove googlecode

rodrigc:
This fixes the following test on Python 3:

```
trial buildbot.test.unit.test_www_hooks_googlecode
```


tardyp:
google code is no more. we can remove that code

rodrigc:
OK, changing this PR to remove googlecode.

tardyp:
looks like there are some bb:chsrc:GoogleCodeAtomPoller refs in the doc to remove

PR #2599, opened 2017-01-15
remove reflection from worker commands registry

rutsky:
Worker is tending to be more "dumb" with more logic moved to master.
This way it's easier to maintain hordes of workers --- no need to
upgrade each one for every minor fix or feature (since all features are in
master).

Previously having lazily-loaded commands was useful --- not all workers
had Python modules to work with different flavours of VCS and tricky
commands.

Currently worker have only minimal basic commands which are available
on all worker installation (they mostly use standard Python modules),
so there is no point in lazy-loading.

Also removal of lazy loading makes modules dependencies explicit which
should help IDE and tools such as pyinstaller.

### DIFF
--- a/worker/buildbot_worker/commands/registry.py
+++ b/worker/buildbot_worker/commands/registry.py
@@ -16,26 +16,27 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
-from twisted.python import reflect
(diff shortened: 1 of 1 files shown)


tardyp:
I let this here in case anybody wants to step in

PR #2604, opened 2017-01-15
Update pylint

rutsky:
## Contributor Checklist:


pselle:
This looks good to me! I like some of the updates (ex. simplify condition) that I'm guessing came from pylint.

PR #2607, opened 2017-01-16
enable master tests for python 3

tardyp:
but ignore the results..

so that we can have a status for master tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,13 +60,14 @@ matrix:
   include:
     # Tests of buildbot-worker on python 2.6
     # Specify SQLALCHEMY=latest to avoid errors installing.
+    - python: "3.5"
+      env: TWISTED=trunk TESTS=trial SQLALCHEMY=latest
+
     - python: "2.6"
       env: TWISTED=14.0.2 TESTS=trial_worker SQLALCHEMY=latest
     # python 3 tests
     - python: "3.5"
       env: TWISTED=latest TESTS=trial_worker SQLALCHEMY=latest
-    - python: "3.5"
-      env: TWISTED=latest TESTS=trial SQLALCHEMY=latest
 
   allow_failures:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -62,11 +62,15 @@ matrix:
     # Specify SQLALCHEMY=latest to avoid errors installing.
     - python: "2.6"
       env: TWISTED=14.0.2 TESTS=trial_worker SQLALCHEMY=latest
-    - python: "2.6"
```
seankelly: Why the deletion of this block?

tardyp: its tested by bbtravis anyway. As travis is slower, I think we need to remove as much as possible stuff from travis

seankelly: Makes sense. Any way to have bbtravis also test Python 3?

tardyp: I will, but it takes more time, as I need to create a py3 capable docker image.

rodrigc:
Apparently, [Mysql-python 1.2.5](https://pypi.python.org/pypi/MySQL-python/1.2.5) does not work on Python 3

According to the [Python 3 Wall of SuperPowers](https://python3wos.appspot.com/), if you look at the **py3k** link next to Mysql-python, it points to [mysqlclient](https://pypi.python.org/pypi/mysqlclient) which which is a PY3-compatible fork of Mysql-python

tardyp:
@rodrigc I've got a bunch of Deferred object is unsubscriptable. Do I need to test against twisted trunk?

rodrigc:
Twisted trunk has the same problem.  I posted this question: http://twistedmatrix.com/pipermail/twisted-python/2017-January/031091.html

rodrigc:
@tardyp I have a fix for some of those Deferred object not subscriptable: https://github.com/buildbot/buildbot/pull/2661

PR #2608, opened 2017-01-16
Enable pylint spellcheck

rutsky:
Starting from pylint 1.4.0 pylint has spellchecker: https://docs.pylint.org/en/1.6.0/features.html#spelling-checker
Requirements for it are the same as for doc's spellchecker (`pyenchant`).

This PR enables it and puts all "misspelled" words into ignore list.
Later this list should be reviewed and typos should be fixed.

This list is different from ignore list of documentation, since it will contain much more exclusion which are not suitable for docs (like Python's modules names, variable names, etc).


rutsky:
Now `master[test]` depends on `pyenchant` (since it depends on `pylint` I think it's OK).

On my machine adding spellchecking doesn't increased pylint-ing time. In fact it even started a bit faster: 5 min 13 sec vs 5 min 29 sec.

rutsky:
Typo is not detected... Need to debug...

rutsky:
Examples of misspell detected in our CI:
https://nine.buildbot.net/#/builders/2/builds/88675

tardyp:
I am concerned that for the docs, the spelling_list.txt was not cleaned up after a long while, and some very obvious common typo did leak into that which gives some false sense of safety.

rutsky:
@tardyp at least current spell checking in docs prevents some new errors.
I haven't got time to cleanup word list in docs, this is a good easily-doable task -- I opened ticked for this: http://trac.buildbot.net/ticket/3663

common/pylintrc
```diff
@@ -175,6 +175,26 @@ evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / stateme
...
+# installed, pylint will fail.
+# If this will be left empty pylint will ignore all spelling errors.
+spelling-dict=en_US
+
+# List of comma separated words that should not be checked.
+spelling-ignore-words=
+
+# A path to a file that contains private dictionary; one word per line.
+# Path relative to current working directory.
+spelling-private-dict-file=
```
rutsky: It's relative to current working directory (as I specified in the comment), but pylint is being run from different directories (`master/`, `worker/`).

tardyp: why can't we use ../common/code_spelling_ignore_words.txt then?

rutsky: Good point, since they are the same we can rely on that. Let me fix that.

To run pylint manually from other directory it will be enough to provide `--spelling-private-dict-file` with proper path on command line.

tardyp:
restarting the tests to make sure it still against the previous PRs.

PR #2609, opened 2017-01-16
Replace MySQL-python with mysqlclient.

rodrigc:
MySQL-python does not work on Python 3.
On the Python Wall of Superpowers ( https://python3wos.appspot.com/ ),
mysqlclient is listed as the Python 3 alternative to MySQL-python.

### DIFF
--- a/master/buildbot/monkeypatches/__init__.py
+++ b/master/buildbot/monkeypatches/__init__.py
@@ -63,10 +63,10 @@ def patch_mysqlclient_warnings():
         from _mysql_exceptions import Warning
         # MySQLdb.compat is only present in mysqlclient
         import MySQLdb.compat  # noqa pylint: disable=unused-import
-    except ImportError as e:
+    except ImportError:
         return
     # workaround for https://twistedmatrix.com/trac/ticket/9005
-    # libmysqlclient is easier to patch than twisted
+    # mysqlclient is easier to patch than twisted
     # we swap _mysql_exceptions.Warning arguments so that the code is in second place
 
     def patched_init(self, *args):
--- a/master/buildbot/monkeypatches/__init__.py
+++ b/master/buildbot/monkeypatches/__init__.py
@@ -58,6 +58,26 @@ def patch_testcase_assert_raises_regexp():
 
 
 @onlyOnce
+def patch_mysqlclient_warnings():
+    try:
+        from _mysql_exceptions import Warning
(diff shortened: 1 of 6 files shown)


tardyp:
I can see in http://docs.sqlalchemy.org/en/latest/dialects/mysql.html#py3k-support that sqlalchemy confirms it supported. So the problem should be easy. Do you need me to take a look?

rutsky:
@rodrigc docs contains info about running Buildbot test suite using MySQL database from Docker (quite easy to setup on Linux): https://docs.buildbot.net/latest/developer/database.html#run-databases-in-docker

You can use these docs for testing locally, plus please update examples there to use `mysqlclient`.

tardyp:
updated the PR with some fixes for mysqlclient

master/buildbot/monkeypatches/__init__.py
```diff
@@ -59,7 +59,12 @@ def patch_testcase_assert_raises_regexp():
 
 @onlyOnce
 def patch_mysqlclient_warnings():
-    from _mysql_exceptions import Warning
+    try:
+        from _mysql_exceptions import Warning
+        import mysqlclient
+        [mysqlclient]  # noqa
```
rutsky: Please try following:

```
import mysqlclient  # noqa pylint: disable=unused-import
```

instead of

```
import mysqlclient
[mysqlclient]  # noqa
```

master/buildbot/monkeypatches/__init__.py
```diff
@@ -58,6 +58,26 @@ def patch_testcase_assert_raises_regexp():
...
+        # MySQLdb.compat is only present in mysqlclient
+        import MySQLdb.compat  # noqa pylint: disable=unused-import
+    except ImportError:
+        return
+    # workaround for https://twistedmatrix.com/trac/ticket/9005
+    # mysqlclient is easier to patch than twisted
+    # we swap _mysql_exceptions.Warning arguments so that the code is in second place
+
+    def patched_init(self, *args):
+        if isinstance(args[0], long):
```
rodrigc: @tardyp I just noticed this change that you did.  This causes a problem on Python 3, because
there is no `long` type there.  Can you fix this?

As you are introducing new patches, can you run the tests in a Python3 virtualenv, just to double check
that new Python 3 regressions are being introduced as we move forward?

Thanks.

PR #2613, opened 2017-01-17
Replace first argument to exec() with a string.

rodrigc:
exec() in Python 3 no longer accepts an open file descriptor.

Python 2 exec statement accepts an open file descriptor:
https://docs.python.org/2/reference/simple_stmts.html#exec

but Python 3 exec does not:
https://docs.python.org/3/library/functions.html#exec

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -29,6 +29,7 @@ from types import MethodType
 
 from twisted.python import failure
 from twisted.python import log
+from twisted.python.compat import execfile
 from zope.interface import implementer
 
 from buildbot import interfaces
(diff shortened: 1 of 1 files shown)


master/buildbot/config.py
```diff
@@ -121,7 +121,7 @@ def loadConfigDict(basedir, configFileName):
     sys.path.append(basedir)
     try:
         try:
-            exec(f, localDict)
```
tardyp: Problem of this is that the stacktrace will not generate proper stack traces, and will not tell the problem is in master.cfg.

This suggests to use compile(code, filename, 'exec') if PY3

http://stackoverflow.com/questions/436198/what-is-an-alternative-to-execfile-in-python-3

rodrigc: OK I replaced with twisted.python.compat.execfile()

PR #2615, opened 2017-01-17
WIP: use buildbotURLAliases to cover different hostnames.

ewongbb:
a possible wip to allow different aliases for the host.

### DIFF
--- a/master/buildbot/status/master.py
+++ b/master/buildbot/status/master.py
@@ -131,7 +131,8 @@ class Status(service.ReconfigurableServiceMixin, service.AsyncMultiService):
 
     def getBuildbotURLAliases(self):
         if self.master.config.buildbotURLAliases:
-            return self.master.config.buildbotURLAliases
+            return self.master.config.buildbotURLAliases.append(
+              self.master.config.buildbotURL)
         return [self.master.config.buildbotURL]
 
     def getStatus(self):
--- a/master/buildbot/status/master.py
+++ b/master/buildbot/status/master.py
@@ -130,10 +130,10 @@ class Status(service.ReconfigurableServiceMixin, service.AsyncMultiService):
         return self.master.config.buildbotURL
 
     def getBuildbotURLAliases(self):
-        if self.master.config.buildbotURLAliases:
-            return self.master.config.buildbotURLAliases.append(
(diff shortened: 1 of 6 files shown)


www/base/src/app/home/home.tpl.jade
```diff
@@ -1,5 +1,5 @@
 .container
-  .row(ng-if="config.buildbotURL != baseurl")
+  .row(ng-if="config.buildbotURL != baseurl and config.buildbotURLAliases.indexof(baseurl) === -1")
```
tardyp: I think we can have the buildbotURLAliases include the buildbotURL so that this expression is simplified

master/buildbot/status/master.py
```diff
@@ -129,6 +129,12 @@ def getTitleURL(self):
     def getBuildbotURL(self):
         return self.master.config.buildbotURL
 
+    def getBuildbotURLAliases(self):
```
tardyp: Please add unit tests for this function.

list.append() does not return the list

tardyp (changes requested):
I think it would be useful to have the Aliases defaults to  "http://localhost:{}".format(self.www.get('port', 8010))

also, the default buildbotURL should be :8010 and not :8080 like I see  there

master/buildbot/status/master.py
ewongbb: so I need to change https://github.com/buildbot/buildbot/blob/master/master/buildbot/config.py#L194 to :8010?

tardyp: Yes, I think that was a mistake

tardyp:
@ewongbb looks like we forgot this PR. There are some lint checks to fix, and very few comments to address, and then we can merge

ewongbb:
@tardyp re: 8080->8010..  with all those mentions of 8080 in the tests, am I doing something wrong?

master/buildbot/status/master.py
andrewdicosmo: Addressed in 1abe424c. Tenant discovery now requires the pending recipient tenant to match its broadcast tenant. The new collision regression verifies no tenant is enqueued from another tenant's recipient row.

PR #2624, opened 2017-01-17
Replace filter() with list comprehensions in a few places for Python 3

rodrigc:
Replace filter() with list comprehensions in a few places,
do deal with different filter implementation in Python 3.
Code is slightly more readable as well.


tardyp:
:heart:

PR #2627, opened 2017-01-18
Use timeout from configuration file when deleting directory.

dirkvdb:
The svn step does not take the timeout value from the configuration file into account when deleting directories. This causes timeouts on our setup when deleting big directories over network shares.

This patch makess the svn step behave like e.g. the git step.


tardyp:
Looks good. You will need to update the unit tests

dirkvdb:
I updated the svn step test

tardyp:
thanks!

PR #2631, opened 2017-01-18
improvements for buildbotNetUsageData

tardyp:
:bb:cfg:`buildbotNetUsageData` now uses ``requests`` if available and will default to HTTP if a bogus SSL implementation is found.
It will also correctly send information about the platform type.

### DIFF
--- a/master/buildbot/buildbot_net_usage_data.py
+++ b/master/buildbot/buildbot_net_usage_data.py
@@ -23,7 +23,6 @@ from __future__ import absolute_import
 from __future__ import print_function
 from future.moves import urllib
 from future.moves.urllib import request as urllib_request
-from future.moves.urllib.request import urlopen as urllib_open
 
 import hashlib
 import inspect
@@ -93,8 +92,7 @@ def basicData(master):
     # to get as much as possible an unique id
     # we hash it to not leak private information about the installation such as hostnames and domain names
     installid = hashlib.sha1(
-        master.name  # master name contains hostname + master basepath
-        +
+        master.name +  # master name contains hostname + master basepath
         socket.getfqdn()  # we add the fqdn to account for people
                           # call their buildbot host 'buildbot' and install it in /var/lib/buildbot
     ).hexdigest()
@@ -184,7 +182,7 @@ def _sendBuildbotNetUsageData(data):
         log.msg("buildbotNetUsageData: Could not send using https, "
                 "please `pip install 'requests[security]'` for proper SSL implementation`")
         data['buggySSL'] = True
-        res = _sendWithUrlib2(PHONE_HOME_URL.replace("https://", "http://"), data)
+        res = _sendWithUrlib(PHONE_HOME_URL.replace("https://", "http://"), data)
 
     log.msg("buildbotNetUsageData: buildbot.net said:", res)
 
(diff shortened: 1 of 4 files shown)


master/buildbot/test/unit/test_buildbot_net_usage_data.py
```diff
@@ -118,5 +122,12 @@ def close(self):
...
         self.assertEqual(len(open_url), 1)
         self.assertEqual(open_url[0].request.args,
-                        ('https://events.buildbot.net/events/phone_home',
+                         ('https://events.buildbot.net/events/phone_home',
                          '{"foo": "bar"}', {'Content-Length': 14, 'Content-Type': 'application/json'}))
+
+    def test_urllib2_real(self):
+        if "TEST_BUILDBOTNET_USAGEDATA" not in os.environ:
+            raise SkipTest(
+                "hyper integration tests only run when environment variable TEST_HYPER is set")
```
seankelly: "hyper integration tests"?

tardyp: oups bad copy paste. thanks for watching!

rodrigc:
@tardyp since you are moving forward with this code, can you use future.moves.urllib, to make your changes Python 3 friendly?  I submitted a patch against your tree: https://github.com/tardyp/buildbot/pull/12

Specifically this changeset:
https://github.com/tardyp/buildbot/pull/12/commits/ed27ffd3b44f2f43a1fd4cba06d668ee1ef72cbd

I tested it with Python 2.7 and HTTPS by doing:

```
export TEST_BUILDBOTNET_USAGEDATA=true
trial buildbot.test.unit.test_buildbot_net_usage_data
```

```
buildbot.test.unit.test_buildbot_net_usage_data
  Tests
    test_basic ...                                                         [OK]
/Users/crodrigues/buildbot2/master/buildbot/config.py:90: ConfigWarning: [0.9.0 and later] `buildbotNetUsageData` is not configured and defaults to basic
This parameter helps the buildbot development team to understand the installation base
No personal information is collected.
Only installation software version info and plugin usage is sent
You can `opt-out` by setting this variable to None
Or `opt-in` for more information by setting it to "full"

    test_custom ...                                                        [OK]
    test_full ...                                                          [OK]
    test_real ...                                                          [OK]
    test_urllib ...                                                        [OK]

Ran 5 tests in 0.728s

PASSED (successes=5)
```

tomprince:
This could just use `twisted.web.client.Agent`, instead of using an external dependency.

tardyp:
@tomprince The problem of twisted agent (beyond its complicated API) is that it does not support http proxy, which would prevent any usage data from corporates. Also it would anyway require additional dependencies for proper SSL support.

@rodrigc thanks for the reminder. I cherry-pick your py3 fix!

master/buildbot/buildbot_net_usage_data.py
```diff
@@ -143,14 +147,45 @@ def computeUsageData(master):
...
+    res = _sendWithRequests(PHONE_HOME_URL, data)
+    # then we try with stdlib, which not always work with https
+    if res is None:
+        res = _sendWithUrlib(PHONE_HOME_URL, data)
+    # at last stage
+    if res is None:
+        log.msg("buildbotNetUsageData: Could not send using https, "
+                "please `pip install 'requests[security]'` for proper SSL implementation`")
+        data['buggySSL'] = True
+        res = _sendWithUrlib2(PHONE_HOME_URL.replace("https://", "http://"), data)
```
rodrigc: This should be _sendWithUrlilb()

Can you fix this and then get this merged in?  This will fix a bunch of tests on Python 3.

PR #2641, opened 2017-01-21
Add __le__() and __eq__() methods in classes with __cmp__().

rodrigc:
In Python 3, __cmp__() and cmp() are gone, in favor of __le__()
and __eq__().

See:
http://python3porting.com/problems.html#unorderable-types-cmp-and-cmp

### DIFF
--- a/master/buildbot/changes/changes.py
+++ b/master/buildbot/changes/changes.py
@@ -143,7 +143,15 @@ class Change:
             self.codebase)
 
     def __cmp__(self, other):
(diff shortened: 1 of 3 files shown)


rodrigc:
Thanks for merging my fixes.  I tried running `trial buildbot.test` under Python 3.6 and got:

```
FAILED (skips=158, failures=148, errors=618, successes=3952)
```

This is compared to Python 2.7:

```
PASSED (skips=232, successes=4614)
```

So roughly 86% of tests are passing under Python 3.6.

PR #2645, opened 2017-01-21
remove pyflakes from requirements - use flake8 instead

rutsky:


rutsky:
@tardyp can you remind why `flake8` version is locked on `2.6.0`?

tardyp:
probably latest version of flake8 gives more errors we need to fix

PR #2648, opened 2017-01-21
Fix buildbot.test.unit.test_steps_shell on Python 3

rodrigc:
This fixes on Python 3:

```
trial buildbot.test.unit.test_steps_shell
```

### DIFF
--- a/master/buildbot/steps/shell.py
+++ b/master/buildbot/steps/shell.py
@@ -134,8 +134,9 @@ class ShellCommand(buildstep.LoggingBuildStep):
         # check validity of arguments being passed to RemoteShellCommand
         invalid_args = []
         if PY3:
-            valid_rsc_args = inspect.signature(
+            signature = inspect.signature(
                 remotecommand.RemoteShellCommand.__init__)
+            valid_rsc_args = signature.parameters.keys()
         else:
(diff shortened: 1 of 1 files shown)


master/buildbot/test/unit/test_steps_shell.py
```diff
@@ -421,7 +430,11 @@ def test_run_property_no_strip(self):
         self.expectOutcome(result=SUCCESS,
                            state_string="property 'res' set")
         self.expectProperty("res", "\n\nabcdef\n")
-        self.expectLogfile('property changes', r"res: u'\n\nabcdef\n'")
+        if PY3:
```
tardyp: I think it would be wiser to have same behaviour on py2 and py3, and probably the one that makes more sense is

    res = r"res: '\n\nabcdef\n'"

I am not sure exactly what are the implication though. can you look at that?

rodrigc: OK, I have removed this particular part of the patch, and will work on this in a separate PR.

The rest of this patch should be good to go in.

rodrigc:
I've removed the part of the patch which was holding up this PR and will address it in another PR.
This PR should be OK.

PR #2653, opened 2017-01-22
Add key functions to sorted() to fix Python 3 errors.

rodrigc:


master/buildbot/test/unit/test_db_steps.py
```diff
@@ -262,9 +262,15 @@ def test_addURL_race(self):
...
 
         stepdict = yield self.db.steps.getStep(stepid=72)
-        # order is not guaranteed though
-        self.assertEqual(sorted(stepdict['urls']), [{'name': u'foo', 'url': u'bar'},
-                                                    {'name': u'foo2', 'url': u'bar2'}])
+
+        def urlKey(url):
+            return url['name']
+
+        # order is not garanteed though
```
rodrigc: @rutsky I fixed that

master/buildbot/test/fake/fakedb.py
```diff
@@ -2322,7 +2322,11 @@ def getMaster(self, masterid):
         return defer.succeed(None)
 
     def getMasters(self):
-        return defer.succeed(sorted(self.masters.values()))
+        def masterKey(master):
+            return master['id']
+
+        return defer.succeed(sorted(self.masters.values(),
+                                    key=masterKey))
```
tardyp: lambda is probably more concise and readable

master/buildbot/test/integration/test_upgrade.py
```diff
@@ -135,7 +135,7 @@ def comp(engine):
                     dict(name=idx.name,
                          unique=idx.unique and 1 or 0,
                          column_names=sorted([c.name for c in idx.columns]))
-                    for idx in tbl.indexes])
+                    for idx in tbl.indexes], key=lambda x: x['name'])
```
tardyp: I think we indeed should prefer the lambda x: x['thing'] form. We should probably use that style for all dictionary sort

master/buildbot/test/unit/test_data_builders.py
```diff
@@ -185,37 +185,41 @@ def test_updateBuilderList(self):
 
         # add another
         yield self.rtype.updateBuilderList(13, [u'somebuilder', u'another'])
-        self.assertEqual(sorted((yield self.master.db.builders.getBuilders())),
+
+        def builderKey(builder):
+            return builder['id']
+
+        self.assertEqual(sorted((yield self.master.db.builders.getBuilders()), key=builderKey),
```
tardyp: For those big function this make sense to make it reusable

master/buildbot/test/fake/fakedb.py
rodrigc: I disagree with you here.  I find lambdas hard to read.  In this case
the masterKey function is only two lines, and is right above this line.

rodrigc: Even though I disagree with you, I have changed this instance to use a lambda,
because I don't want this PR to be blocked on this.

tardyp: I agree this is the kind of styling issue that leads to disagrement. I think we both have reasonable arguments, I am often reluctant going to fight for it.

But we need to be consistent at least.

PR #2655, opened 2017-01-23
Fix broken docs links

rutsky:
Fixed `make docschecks` invocation.
Looks like currently `linkcheck` is not being run nor on Travis, nor on MetaBuildbot.


tardyp:
I disabled the check because it is too noisy and dependent of websites being up.

rutsky:
@tardyp this is understandable. At least each developer can do `make docschecks` in the sources to run these checks. Maybe we can automate running this check before making new release?

PR #2656, opened 2017-01-23
Minor renamings of slave-like variables to worker-like ("s", "sl", "sb")

rutsky:


tardyp:
@sa2ajj would argue that one letter variable names are bad in the first place. He certainly is right, and I am certainly culprit of this bad habit

rutsky:
@tardyp well, for loops like `defer.returnValue([self.db2data(w) for w in workers_dicts])` single-letter variables looks fine to me. Inside *tests* `self.wrk` probably ok too (though personally I would use `self.worker`). For other cases I agree that single-letter variables are not good.

PR #2658, opened 2017-01-23
Replace data.sort(cmp= with data.sort(key= for Python 3

rodrigc:
list.sort(cmp= is not available on Python 3

### DIFF
--- a/master/buildbot/data/resultspec.py
+++ b/master/buildbot/data/resultspec.py
@@ -336,15 +336,15 @@ class ResultSpec(object):
                 # Do a multi-level sort by sorting from
                 # the least significant key to the most
                 # significant key.  If we are given the keys
-                # ('ln', 'fn', 'age'):
+                # ('lastName', 'firstName', 'age'):
                 #   - first sort by 'age'
-                #   - then sort by 'fn'
-                #   - then sort by 'ln'
+                #   - then sort by 'firstName'
+                #   - then sort by 'lastName'
                 for k in reversed(self.order):
                     doReverse = False
                     if k[0] == '-':
-                        # If we get a key '-ln',
-                        # it means sort by 'ln' in reverse.
+                        # If we get a key '-lastName',
+                        # it means sort by 'lastName' in reverse.
                         k = k[1:]
                         doReverse = True
 
--- a/master/buildbot/data/resultspec.py
+++ b/master/buildbot/data/resultspec.py
@@ -348,8 +348,8 @@ class ResultSpec(object):
                         k = k[1:]
                         doReverse = True
 
-                    def keyFunc(a):
-                        val = a[k]
+                    def keyFunc(a, sortBy=k):
+                        val = a[sortBy]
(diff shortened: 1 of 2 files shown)


master/buildbot/data/resultspec.py
```diff
@@ -109,6 +109,14 @@ def nonecmp(a, b):
     return cmp(a, b)
 
 
+class MinType(object):
```
tardyp: This is a clever trick, but needs at least a comment on why it does this like that.

master/buildbot/data/resultspec.py
```diff
@@ -325,18 +333,28 @@ def includeFields(d):
...
-                         for k in self.order]
-
-                def cmpFunc(a, b):
-                    for f in order:
-                        c = f(a, b)
-                        if c:
-                            return c
-                    return 0
-                data.sort(cmp=cmpFunc)
+                # Do a multi-level sort by sorting from
```
tardyp: we should probably put this in a separate function, with this comments put as a docstring

master/buildbot/test/unit/test_data_resultspec.py
```diff
@@ -309,6 +309,11 @@ def test_popField_not_present(self):
 
 class NoneCmp(unittest.TestCase):
 
+    try:
+        cmpFunc = cmp
+    except NameError:
+        skip = "cmp() builtin not available on Python 3"
+
     def test_nonecmp(self):
```
tardyp: I'd believe nonecmp is not used anymore as we are switching the implementation to use key=, so you can remove the test and implementation

master/buildbot/data/resultspec.py
```diff
@@ -325,18 +333,28 @@ def includeFields(d):
...
+                        k = k[1:]
+                        doReverse = True
+
+                    def keyFunc(a, sortBy=k):
+                        val = a[sortBy]
+                        if val is None:
+                            val = MinType()
+                        return val
+
+                    data.sort(key=keyFunc, reverse=doReverse)
```
tardyp: we are doing N sorts instead of one? sounds like a big performance regression. This sorting is in the heart of our data api, so we need to make it optimized (at least bigO wise, no need for micro optimization).


http://stackoverflow.com/questions/5212870/sorting-a-python-list-by-two-criteria

second answer looks like the one we need:

    sorted(list, key=lambda x: (x[0], -x[1]))

We need to build a function that will turn 

    ('lastName', '-firstName', 'age')

to

    lambda x: (NoneAndReverseComparator(x['lastname'], reverse=False), (NoneAndReverseComparator(x['firstName'], reverse=True), NoneAndReverseComparator(x['age'], reverse=False))

with NoneAndReverseComparator to implement __eq__ and __lt__ taking in account None to be the minimum, and optionally reverse the output of __lt__.

rodrigc:
Moved to https://github.com/buildbot/buildbot/pull/2680

PR #2659, opened 2017-01-23
Remove RUN_FLAKY_TESTS requirements for a few tests.

rodrigc:
Unicode fixes for Python 3 have fixed these tests on Python 2.

Fixes: #3339


tardyp:
@rodrigc #3339 is not about u'xx' vs 'xx' it is about
  
    log:stdio (0)\n

vs

    log:stdio (29)

but there have been changes since on the log system, and maybe this is more reliable (notably there is no more a timing when compressing logs where the log line count comes to zero)

PR #2661, opened 2017-01-24
Replace Python 3 generator expressions with lists

rodrigc:
@exarkun explained this here: http://twistedmatrix.com/pipermail/twisted-python/2017-January/031104.html

```
On Python 2, [(yield self.db2data(br)) for br in buildrequests] is a list comprehension.  It will have len(buildrequests) elements and each will be the value sent back in to the generator via the yield expression.

On Python 3, the same expression is a list of one element which is a generator expression.

This form is much clearer, I think, and behaves as intended on both versions of Python:

    results = []
    for br in buildrequests:
        results.append((yield self.db2data(br)))
    defer.returnValue(results)
```

Thanks to @exarkun for the explanation.

### DIFF
--- a/master/buildbot/data/buildrequests.py
+++ b/master/buildbot/data/buildrequests.py
@@ -143,8 +143,10 @@ class BuildRequestsEndpoint(Db2DataMixin, base.Endpoint):
             claimed=claimed,
             bsid=bsid,
             resultSpec=resultSpec)
-        defer.returnValue(
-            [(yield self.db2data(br)) for br in buildrequests])
+        results = []
+        for br in buildrequests:
+            results.append((yield self.db2data(br)))
+        defer.returnValue(results)
 
 
 class BuildRequest(base.ResourceType):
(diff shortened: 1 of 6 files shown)


master/buildbot/data/buildrequests.py
```diff
@@ -143,8 +143,10 @@ def get(self, resultSpec, kwargs):
             claimed=claimed,
             bsid=bsid,
             resultSpec=resultSpec)
-        defer.returnValue(
-            [(yield self.db2data(br)) for br in buildrequests])
+        results = []
+        for br in buildrequests:
+            results.append((yield self.db2data(br)))
+        defer.returnValue(results)
```
seankelly: I'm not well versed in Twisted, but could this be replaced with a `DeferredList`?

rodrigc:
@seankelly I gave it a quick go with DeferredList, and it didn't seem to work out.  The upper layers calling this stuff didn't work well with it.

master/buildbot/data/buildrequests.py
seankelly: I got it to work with this:

        steps_dict = [self.db2data(dbdict) for dbdict in steps] 
        list_results = yield defer.DeferredList(steps_dict)
        results = [res[1] for res in list_results]
        defer.returnValue(results)

Not quite what I had in mind. It does allow the `self.db2data` calls to run concurrently but isn't that pretty.

tardyp:
I agree that these code path needs DeferredList. but we can do it later

PR #2666, opened 2017-01-24
Fix bytes2NativeString() in master, due to future.types.newbytes

rodrigc:


master/buildbot/util/__init__.py
```diff
@@ -252,7 +252,7 @@ def bytes2NativeString(x, encoding='utf-8'):
     @param encoding: an optional codec, default: 'utf-8'
     @return: a string of type C{str}
     """
-    if isinstance(x, bytes) and str != bytes:
+    if isinstance(x, bytes) and type("") != type(b""):
```
tardyp: why dont you just use following?

    if isinstance(x, bytes) and PY3:

rodrigc: I wanted to leave the future.utils constructs out of this function as much as possible,
and to future proof things, for example whenever Python 4 comes out.

tardyp: well, at least we can add a comment saying that this means PY>=3

rodrigc: OK I added a comment.

tardyp: I meant a comment in the code. I dont see it. did you upload?

PR #2667, opened 2017-01-24
fix the py3status

tardyp:
I merged too early :-}

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,10 +69,6 @@ matrix:
     - python: "3.5"
       env: TWISTED=latest TESTS=trial_worker SQLALCHEMY=latest
 
-  allow_failures:
-    - python: 3.5
-      TESTS: py3status
-
 # Dependencies installation commands
 install:
   - pip install -U pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,7 +131,8 @@ script:
 
   - |
     if [ $TESTS = py3status ]; then
-        trial  --reporter=summary buildbot.test buildbot_worker.test| tail -n 1 | python -c "
+       trial --reporter=summary buildbot.test | tee py3status
+       tail -n 1 py3status | python -c "
     import sys, requests, os
     (timeTaken, tests, _, errors, failures, skips) = sys.stdin.read().split()
     requests.post('https://events.buildbot.net/events/py3kstatus', json={'py3kstatus': {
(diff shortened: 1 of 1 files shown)


seankelly:
Would also be nice if the output got written to stdout/stderr.

.travis.yml
```diff
@@ -130,14 +130,13 @@ script:
   - "[ $TESTS != js -o ${TRAVIS_PULL_REQUEST} != false ] || make frontend_install_tests"
 
   - |
-    TESTS=py3status
     if [ $TESTS = py3status ]; then
         trial  --reporter=summary buildbot.test buildbot_worker.test| tail -n 1 | python -c "
```
seankelly: Maybe:

    trial --reporter=summary buildbot.test buildbot_worker.test | tee py3status
    tail -n 1 py3status | python -c "
    ...

rodrigc:
In .travis.yml, there are a bunch of lines like:

```
 - "[ $TESTS != lint ] || make pylint"
 - "[ $TESTS != lint ] || make flake8"
 - "[ $TESTS != lint ] || isort --check -df `git ls-files |grep '.py$'`"
```

Why isn't the logic like:

```
 - "[ $TESTS = lint ] && make pylint"
 - "[ $TESTS = lint ] && make flake8"
 - "[ $TESTS = lint ] && isort --check -df `git ls-files |grep '.py$'`"
```

That would make the logic a lot more readable.

tardyp:
@rodrigc I agree, I dont know exactly the reason. I guess this is from a best practice in travis doc, but I have no idea why.
in bbtravis, we are using a better DSL imho.

tardyp:
@rodrigc I this that is because if you use the && form, then the shell will not detect the return value in case of error

rodrigc:
What do you care what the error is if you have:

```
[ $TESTS == foo ] && do_foo
```

$TESTS is just a string that you are comparing.

tardyp:
@rodrigc 
if you do 

    - [ $TESTS == foo ] && do_foo

and $TEST is e.g. equal to 'trial', then  [ $TESTS == foo ]  is false, so the return value will not be zero, so this line of shell will be considered failure, and the build will be marked fail :-(

rutsky:
The problem is that Travis support different envs, but doesn't support different scripts for different envs to run.
So different scripts are emulated in such ugly way.

rutsky:
Not "the" problem of `&& vs ||`, but the reason why logic is "define N cools envs, and run single ugly script with N conditions to switch between them".

tardyp:
image

at least this works..

rodrigc:
Are you sure this is working as you want?
If I input the .travis.yml to: https://lint.travis-ci.org, I get:

```
in matrix.allow_failures section: unexpected key TESTS, dropping
```

PR #2670, opened 2017-01-24
Replace '||' with 'if []; then' to make logic more readable

rodrigc:


rodrigc:
@rutsky you raised a very valid point here: https://github.com/buildbot/buildbot/pull/2667#issuecomment-274953165
However, this can be worked around by replacing

```
- [ $TESTS != foo ] || do_foo
```

with

```
- if [ $TESTS == foo ]; then do_foo ; fi
```

seankelly:
Another option is the obscure

    [ $TEST = foo ] && {
        things
    } || :

.travis.yml
```diff
@@ -71,51 +71,53 @@ matrix:
...
 # Dependencies installation commands
 install:
   - pip install -U pip
   # codecov is the interface to codecov.io; see after_success
   - pip install codecov
-  - "[ $TWISTED != trunk ] || pip install git+https://github.com/twisted/twisted"
-  - "[ $TWISTED = latest ] || [ $TWISTED = trunk ] || pip install Twisted==$TWISTED"
-  - "[ $SQLALCHEMY = latest ] || pip install sqlalchemy==$SQLALCHEMY"
+  - "if [ $TWISTED == trunk ]; then pip install git+https://github.com/twisted/twisted ; fi"
+  - "if [ $TWISTED != latest -a  $TWISTED != trunk ] ; then pip install Twisted==$TWISTED ; fi"
```
seankelly: Use `[ ... ] && [ ... ]` or `[ ... ] || [ ... ]` instead.

seankelly: For consistency with the rest of the lines, remove the extra space between `]` and `;`.

.travis.yml
```diff
@@ -71,51 +71,53 @@ matrix:
 
   allow_failures:
     - python: 3.5
-      TESTS: py3status
+      env: TESTS=py3status
```
rutsky: This is actual fix of something?

.travis.yml
```diff
@@ -71,51 +71,53 @@ matrix:
...
 
 # Dependencies installation commands
 install:
   - pip install -U pip
   # codecov is the interface to codecov.io; see after_success
   - pip install codecov
-  - "[ $TWISTED != trunk ] || pip install git+https://github.com/twisted/twisted"
-  - "[ $TWISTED = latest ] || [ $TWISTED = trunk ] || pip install Twisted==$TWISTED"
-  - "[ $SQLALCHEMY = latest ] || pip install sqlalchemy==$SQLALCHEMY"
+  - "if [ $TWISTED == trunk ]; then pip install git+https://github.com/twisted/twisted ; fi"
```
rutsky: Use single `=` instead of `==`.

.travis.yml
```diff
@@ -126,11 +128,10 @@ before_script:
 # Tests running commands
 script:
   # make frontend_install_tests takes 17 min, so we only do it post submit
-  - "[ $TESTS != js -o ${TRAVIS_PULL_REQUEST} = false ] || make frontend"
-  - "[ $TESTS != js -o ${TRAVIS_PULL_REQUEST} != false ] || make frontend_install_tests"
+  - "if [ $TESTS == js -a ${TRAVIS_PULL_REQUEST} == false ]; then make frontend ; fi"
+  - "if [ $TESTS == js -a ${TRAVIS_PULL_REQUEST} == false ]; then make frontend_install_tests ; fi"
 
   - |
-    TESTS=py3status
```
rutsky: This is fix of something?

.travis.yml
rodrigc: Yes, this is a fix.  If you run the .travis.yml file under https://lint.travis-ci.org/ you get an error.
See:
https://github.com/buildbot/buildbot/pull/2669

.travis.yml
rodrigc: Yes, this is a fix.  You aren't supposed to assign TESTS=pystatus inside the script.
It needs to be done in the matrix section.

.travis.yml
rodrigc: single `=` is POSIX shell compliant, but `==` works here, because under the covers Travis is using bash which allows `==`.

.travis.yml
rodrigc: OK I removed the space.

.travis.yml
rodrigc: I changed to single `=`

rutsky: Yes, `==` works, but looks like it's better to stick with POSIX variant, also it's shorter and already used in this file.

.travis.yml
```diff
@@ -71,51 +71,53 @@ matrix:
...
 # Dependencies installation commands
 install:
   - pip install -U pip
   # codecov is the interface to codecov.io; see after_success
   - pip install codecov
-  - "[ $TWISTED != trunk ] || pip install git+https://github.com/twisted/twisted"
-  - "[ $TWISTED = latest ] || [ $TWISTED = trunk ] || pip install Twisted==$TWISTED"
-  - "[ $SQLALCHEMY = latest ] || pip install sqlalchemy==$SQLALCHEMY"
+  - "if [ $TWISTED = trunk ]; then pip install git+https://github.com/twisted/twisted ; fi"
+  - "if [ $TWISTED != latest -a  $TWISTED != trunk ]; then pip install Twisted=$TWISTED ; fi"
```
rutsky: In pip install only `==` allowed, not `=`.

rodrigc: OK, I made the pip install `==`.

PR #2672, opened 2017-01-25
Change dict.keys()[0] to list(dict.keys())[0]

rodrigc:


master/buildbot/test/unit/test_stats_service.py
```diff
@@ -313,7 +313,7 @@ def test_build_start_time_capturing(self):
         yield self.end_build_call_consumers()
 
         self.assertEqual(
-            'start-time', self.fake_storage_service.stored_data[0][0].keys()[0])
```
tardyp: I think, for first item of the key generator, its best to just use

    future.util.iterkeys(self.fake_storage_service.stored_data[0][0]).next()

instead of transforming it to a list

rodrigc:
```
iterkeys(self.fake_storage_service.stored_data[0][0]).next()
```

fails with:

```
    'start-time', iterkeys(self.fake_storage_service.stored_data[0][0]).next())
builtins.AttributeError: 'dict_keys' object has no attribute 'next'
```


```
next(iterkeys(self.fake_storage_service.stored_data[0][0])))
```

fails with:

```
  File "/Users/crodrigues/buildbot2/master/buildbot/test/unit/test_stats_service.py", line 317, in test_build_start_time_capturing
    'start-time', next(iterkeys(self.fake_storage_service.stored_data[0][0])))
builtins.TypeError: 'dict_keys' object is not an iterator
```

rodrigc: This works though:

```
next(iter(iterkeys(self.fake_storage_service.stored_data[0][0]))))
```

Seriously do you want that?  That's pretty ugly, and this is only in test code where
the lists are not large.

tardyp: right

PR #2677, opened 2017-01-25
Use urlparse from future module for Python 3

rodrigc:


master/docs/manual/cfg-workers-docker.rst
```diff
@@ -285,9 +285,10 @@ In addition to the arguments available for any :ref:`Latent-Workers`, :class:`Hy
 
 .. code-block:: python
 
-    import requests, urlparse
+    from future.moves.urllib.parse import urlparse
```
rutsky: I suggest not to use `future.moves` in the examples, rather use Python 2 or Python 3 style, or both and one leave commented:

```python
# On Python 2:
from urlparse import urlparse
# On Python 3:
# from urllib.parse import urlparse

import requests

r = requests.get("http://localhost:4040/api/tunnels/command_line").json()
masterFQDN = urlparse(r['public_url']).netloc
```

rodrigc: OK, I made the change

rodrigc:
@tardyp invited me to be a member of the GitHub buildbot group.  Does that mean I can merge stuff after it is approved, or should I hold off and let him do it?

rutsky:
In general it's good to have committer, reviewer and merger three different people (AFAIR discussion in the IRC). We have [document](http://trac.buildbot.net/wiki/GitHubWorkflow) describing GitHub workflow, but I'm unsure is it actual (at least it was written before GitHub landed "review" feature).

If you not in a hurry IMO it's better to wait for others to merge (@tardyp does merges quite often).

rodrigc:
Sounds good to me, I'll wait.
@tardyp has been doing most if it, but If you can help review some of my Py3 fixes: https://github.com/buildbot/buildbot/pulls/rodrigc

That would be helpful.

tardyp:
Those days, we are missing people to afford the three men review process.
I'll would be glad to be able to go back to that, if more people could commit to 1 or 2 reviews pass per day.

With the high rate of PR for py3 tasks we need to be quick to merge, or the burden of rebase will be to hard for @rodrigc.

As for my own PRs I rarely get any review ( :-( ), so I usually merge them after 3/4 days to not have too much backlog.

The main reason I gave you commit access is so that you can remove the need reply/need test labels on your PRs, to tell me I need to look at it again.
I spend a lot of time going back on PR that didn't yet get answer, and also I often do not see that other PR concerns have been addressed.

Also, I would appreciate @rodrigc if you could stay on IRC. I use https://www.irccloud.com/ to be able to always be online, I can give you invites if you need.

master/docs/manual/cfg-workers-docker.rst
```diff
@@ -285,9 +285,15 @@ In addition to the arguments available for any :ref:`Latent-Workers`, :class:`Hy
 
 .. code-block:: python
 
-    import requests, urlparse
+    try:
```
tardyp: I think as buildbot requires future, it is safe to just document with

    from future.moves.urllib.parse import urlparse

or the sake of simplicity

rodrigc: That's what I had before @rutsky asked for the change.
OK, I've put it back to what was there before.

PR #2680, opened 2017-01-25
Replace data.sort(cmp= with data.sort(key= for Python 3

rodrigc:


master/buildbot/data/resultspec.py
```diff
@@ -98,17 +98,41 @@ class Filter(FieldBase):
...
-        return 1
-    return cmp(a, b)
+class NoneComparator(object):
+    """
+    Object which can be substituted for 'None'
+    when doing comparisons in sorted().
+    '> None' and '< None' are not supported
+    in Python 3.
+    """
+    def __lt__(self, other):
```
tardyp: We will need a few unit tests for those two new classes (similar to the noncmp unit tests you removed)

master/buildbot/data/resultspec.py
```diff
@@ -324,20 +348,39 @@ def includeFields(d):
...
-
-                def cmpFunc(a, b):
-                    for f in order:
-                        c = f(a, b)
-                        if c:
-                            return c
-                    return 0
-                data.sort(cmp=cmpFunc)
+                def keyFunc(elem, order=self.order,
+                            noneComparator=NoneComparator()):
```
tardyp: I am not sure why you pass NoneComparator in argument. I think it is a relique from a past design, we should simplify it.

master/buildbot/data/resultspec.py
```diff
@@ -324,20 +348,39 @@ def includeFields(d):
...
+                            # it means sort by 'lastName' in reverse.
+                            k = k[1:]
+                            doReverse = True
+                        val = elem[k]
+                        if val is None:
+                            val = noneComparator
+                        elif doReverse:
+                            val = ReverseComparator(val)
+                        compareKey.append(val)
+                    return compareKey
```
tardyp: algorithm looks good. cool!

rodrigc:
OK, I added tests, and changed the implementation slightly based on what I found with the tests.

PR #2682, opened 2017-01-27
Use context managers to plug leaks of open file handles.

rodrigc:
Fixes this warning on Python 3:
    builtins.ResourceWarning: unclosed file

### DIFF
--- a/worker/buildbot_worker/test/util/misc.py
+++ b/worker/buildbot_worker/test/util/misc.py
@@ -112,13 +112,16 @@ class FileIOMixin(object):
         @param file_contents: contents that will be returned by file object's
                               read() method
         """
-        # create mocked file object that returns 'file_contents' on read()
-        # and tracks any write() calls
-        self.fileobj = mock.Mock()
-        self.fileobj.read = mock.Mock(return_value=file_contents)
-        self.fileobj.write = mock.Mock()
+        # Use mock.mock_open() to create a substitute for
+        # open().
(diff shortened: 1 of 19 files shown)


worker/buildbot_worker/test/util/misc.py
```diff
@@ -112,13 +112,16 @@ def setUpOpen(self, file_contents="dummy-contents"):
...
                               read() method
         """
-        # create mocked file object that returns 'file_contents' on read()
-        # and tracks any write() calls
-        self.fileobj = mock.Mock()
-        self.fileobj.read = mock.Mock(return_value=file_contents)
-        self.fileobj.write = mock.Mock()
+        # Use mock.mock_open() to create a substitute for
+        # open().
+        fakeOpen = mock.mock_open(read_data=file_contents)
```
tardyp: good catch!

tardyp:
@rutsky you can look here https://nine.buildbot.net/#/builders/2/builds/95465
the kind of false positive that are quite annoying.

rodrigc:
I have changed the comment so pylint doesn't fail

rutsky:
@tardyp @rodrigc looks like the only way to "fix" spelling in `# patch open() to always return self.fileobj` is to add `fileobj` to list of ignored words, I don't thing pylint is smart enough to distinguish code samples from comments/docstrings.
IMO it's OK to add such exceptions.

PR #2693, opened 2017-01-30
Skip some tests until we can fix twisted.mail upstream on Python 3

rodrigc:
This ticket is tracking twisted.mail being ported to Python 3: https://twistedmatrix.com/trac/ticket/8770

### DIFF
--- a/master/buildbot/test/integration/test_mailnotifier.py
+++ b/master/buildbot/test/integration/test_mailnotifier.py
@@ -20,6 +20,7 @@ import base64
 
 from twisted.internet import defer
 
+from buildbot.reporters.mail import ESMTPSenderFactory
 from buildbot.reporters.mail import MailNotifier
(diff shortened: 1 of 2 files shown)


master/buildbot/test/integration/test_mailnotifier.py
```diff
@@ -19,6 +19,10 @@
 import base64
 
 from twisted.internet import defer
+try:
+    from twisted.mail.smtp import ESMTPSenderFactory
+except ImportError:
+    ESMTPSenderFactory = None
 
 from buildbot.reporters.mail import MailNotifier
```
tardyp: There is a isort issue. Note that same trick is in mail.py, so you can also do:

     from buildbot.reporters.mail import ESMTPSenderFactory

which will be None if twisted mail does an import error

rodrigc:
OK, I made the change.  Got some isort-related whitespace changes in there too.

PR #2703, opened 2017-01-31
use Jwt tokens for web sessions

tardyp:
Annoying problem with buildbot is that the session are lost when
the master is reboot.
With twisted web sessions are stored in memory, and the API being fully synchronous,
there is no sane way to store it in the db.

JWT is a technic which stores the session state within a cookie,
and sign it with secret key backed hmac.

We store the session secret in the db, and this is the only thing that we
need to store in the db.

Previously when you logged into one master, your cookie was invalid for the other master, now, it is valid for every master

With this implementation we can safely implement multimaster web UI
without the need for IPHash load balancing.

## Contributor Checklist:


tardyp:
@rutsky ok for merge?

rutsky:
@tardyp I made only cursory review of your PR, so can't say confidently "LGTM".
I don't have enough time to properly review, so lets merge it.

rutsky:
And, btw, JWT tokens are very good idea, IMO, good job!

tardyp:
I know its big. I'm working on another one even bigger :-/
its hard to split into easier tor review ones.

PR #2704, opened 2017-02-01
test_checkconfig: always create different config dir

tardyp:
fix for python3

It looks like python3 added some optimization for the import method.

both tests are importing files from _trialtemp/configdir , but between the two tests, the list available modules actually change.

In the second test however, if you run from within the same python runtime, then python won't look again at the directory, and thus wont see it has changed since last time.
So it will not find the new module. If we change 'configdir' with mkstemp, then it will work.

also remove a warning from https://twistedmatrix.com/trac/ticket/9023#ticket

### DIFF
--- a/master/buildbot/test/__init__.py
+++ b/master/buildbot/test/__init__.py
@@ -119,3 +119,6 @@ warnings.filterwarnings('ignore', "Not importing directory.*docker': missing __i
 
 # FIXME: needs to be sorted out (#3666)
 warnings.filterwarnings('ignore', "Invalid utf8 character string")
+
+# twisted.compat.execfile is using 'U' https://twistedmatrix.com/trac/ticket/9023
+warnings.filterwarnings('ignore', "'U' mode is deprecated", DeprecationWarning)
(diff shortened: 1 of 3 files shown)


master/buildbot/test/__init__.py
```diff
@@ -119,3 +119,6 @@
 
 # FIXME: needs to be sorted out (#3666)
 warnings.filterwarnings('ignore', "Invalid utf8 character string")
+
+# twisted.compat.execfile is using 'U' https://twistedmatrix.com/trac/ticket/9023
```
rodrigc: This is triggering lint errors.  Not sure what you want to do here, but I say to waste less time,
leave it out, and for now only test Py3 against Twisted from trunk.

tardyp: I think it is important to keep the comment. the spellcheck linter must *not* encourage us to write less comments! @rutsky

rutsky: @rodrigc next time please add exceptions for those words to https://github.com/buildbot/buildbot/blob/master/common/code_spelling_ignore_words.txt

PR #2705, opened 2017-02-01
fix vacuum for python 3.6

tardyp:
Changed in version 3.6: sqlite3 used to implicitly commit an open transaction before DDL statements. This is no longer the case.

I removed all the actions of the cleanupdb command, and sqlite3 still think I am in a transaction.
looking at Connection.in_transaction, it returns False.

So I am not sure if it is a bug or something outside of my understanding.

### DIFF
--- a/master/buildbot/db/enginestrategy.py
+++ b/master/buildbot/db/enginestrategy.py
@@ -76,6 +76,9 @@ class SqlLiteStrategy(Strategy):
         # try to enable WAL logging
         if u.database:
             def connect_listener(connection, record):
+                # default for PY2, not anymore for PY3, and cause issue with VACUUM
+                # (it thinks it is in a transaction while it is not)
+                connection.isolation_level = None
                 connection.execute("pragma checkpoint_fullfsync = off")
 
             sa.event.listen(engine.pool, 'connect', connect_listener)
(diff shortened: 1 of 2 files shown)


master/buildbot/db/enginestrategy.py
```diff
@@ -76,6 +76,9 @@ def connect_listener_enable_fk(connection, record):
         # try to enable WAL logging
         if u.database:
             def connect_listener(connection, record):
+                # default for PY2, not anymore for PY3, and cause issue with VACUUM
```
rodrigc: Change this comment to:
```
# Sets auto-commit mode, which was the default for PY2, but not anymore fore PY3.
# This causes and issue with VACUUM (which thinks it is in a transaction while it is not).
```

Also make sure to check with pylint, and add any words that pylint doesn't like to the magic list. :)

tardyp: looks like I was lucky this time. VACUUM is an english word. :)

tardyp:
looks like disabling isolation_level will actually make the test fail. So we need another solution.

seankelly:
I've always gotten that error with 2.7 when running with multiple processes (`--jobs=2`). I've not had any idea how to fix it so I ignored the few errors.

rodrigc:
@seankelly I'm look for a hero with the magic fix to the problem!

tardyp:
@rodrigc I think until we got more hint from pysqlite guys, we can target py3.5

tardyp:
superseeded by https://github.com/buildbot/buildbot/pull/2738

PR #2708, opened 2017-02-01
worker: fix cancellation error in missing timer if it is already called

noc0lour:
Just a small fix to catch the error and to allow workers from attaching if missing_timeout is specified.
Otherwise I get:
```
2017-01-30 18:50:08-0800 [-] worker Ubuntu_14_04_64_0 cannot attach
        Failure: twisted.internet.error.AlreadyCalled: Tried to cancel an already-called event.
```

## Contributor Checklist:

### DIFF
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -302,7 +302,8 @@ class AbstractWorker(service.BuildbotService, object):
 
     def stopMissingTimer(self):
         if self.missing_timer:
-            self.missing_timer.cancel()
+            if self.missing_timer.active():
+                self.missing_timer.cancel()
             self.missing_timer = None
 
     def isConnected(self):
@@ -314,7 +315,7 @@ class AbstractWorker(service.BuildbotService, object):
         if not self.parent:
             return
         last_connection = time.ctime(time.time() - self.missing_timeout)
-        yield self.master.data.updates.workerMissing(
+        self.master.data.updates.workerMissing(
             workerid=self.workerid,
             masterid=self.master.masterid,
             last_connection=last_connection,
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -55,6 +55,7 @@ class AbstractWorker(service.BuildbotService, object):
     quarantine_timer = None
     quarantine_timeout = quarantine_initial_timeout = 10
     quarantine_max_timeout = 60 * 60
+    start_missing_on_startup = True
 
     def checkConfig(self, name, password, max_builds=None,
                     notify_on_missing=None,
@@ -220,12 +221,15 @@ class AbstractWorker(service.BuildbotService, object):
     @defer.inlineCallbacks
     def startService(self):
         self.updateLocks()
-        self.startMissingTimer()
         self.workerid = yield self.master.data.updates.findWorkerId(
             self.name)
 
         yield self._getWorkerInfo()
         yield service.BuildbotService.startService(self)
+
+        # startMissingTimer wants the service to be running to really start
+        if self.start_missing_on_startup:
+            self.startMissingTimer()
 
     @defer.inlineCallbacks
     def reconfigService(self, name, password, max_builds=None,
@@ -295,7 +299,7 @@ class AbstractWorker(service.BuildbotService, object):
         yield service.BuildbotService.stopService(self)
 
     def startMissingTimer(self):
-        if self.missing_timeout and self.parent:
+        if self.missing_timeout and self.parent and self.running:
             self.stopMissingTimer()  # in case it's already running
             self.missing_timer = self.master.reactor.callLater(self.missing_timeout,
                                                                self._missing_timer_fired)
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -295,7 +295,7 @@ class AbstractWorker(service.BuildbotService, object):
         yield service.BuildbotService.stopService(self)
 
     def startMissingTimer(self):
-        if self.notify_on_missing and self.missing_timeout and self.parent:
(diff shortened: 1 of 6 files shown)


noc0lour:
I have to catch another Exception in AbstractWorker then worker can attach even if the missing timer was already cancelled.

master/buildbot/worker/base.py
```diff
@@ -296,7 +297,10 @@ def stopService(self):
 
     def startMissingTimer(self):
         if self.notify_on_missing and self.missing_timeout and self.parent:
-            self.stopMissingTimer()  # in case it's already running
+            try:
+                self.stopMissingTimer()  # in case it's already running
+            except error.AlreadyCalled:
```
tardyp: I think here, we are pushing the dust under the carpet.
I would prefer that there is more investigation. Can you reproduce with latest code?

I would prefer if the fix is in stopMissingTimer.

    def stopMissingTimer(self):
        if self.missing_timer:
            if self.missing_timer.active():
                self.missing_timer.cancel()
            self.missing_timer = None

noc0lour: Yeah, I thought about that as well, but that would hide if stopMissingTimer fails unexpected. Currently failing in worker start should be expected since we have no way of querying in which state the `callLater` instance is. 

I'm running on top of yesterdays master, so should be new enough?

tardyp: > Currently failing in worker start should be expected since we have no way of querying in which state the callLater instance is.

What do you mean? there is self.missing_timer.active(), and self.missing_timer is not None.
missing_timer is always reset to None when cancelled, and first thing after the callback is called.

There might be a small case when the timer has triggered, but its deferred is in the queue, but I really don't think this can happen. 
I dont think we should log something as there is nothing the user can do about that. This would lead to more confusion that anything,
So this is why I still prefer my suggestion.

Also, I would like to have a unit test to cover the situation.

noc0lour: Okay, I'll try the suggestion with `is not None` or `missing_timer.active()` to avoid the `try..except` construct, whatever works.

noc0lour:
I tracked down the problem: If a worker tries to attach after a start of the master when the missing_timeout is reached already `self.missing_timer` is not `None` but `self.missing_timer` is also not `active`.
A check `if self.missing_timer.active()` after the current check will solve the problem.
Added a unit test which just cancels() the timer which reflects current behaviour.

tardyp:
@noc0lour how much is your missing_timeout?

The time between the missing_timeout is reached already and self.missing_timer is not None is very short, as missing_timer is reset during the callback. So for me there is still something unclear

tardyp:
Here, you can see
https://github.com/twisted/twisted/blob/twisted-16.5.0/src/twisted/internet/base.py#L824
called is set to 1 just before calling the function
and the function is here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/worker/base.py#L312
The first thing is to reset the missing_timer to None.

The only place where missing_timer is cancelled is here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/worker/base.py#L305

https://github.com/buildbot/buildbot/search?utf8=%E2%9C%93&q=missing_timer.cancel&type=Code
There is another place in latent.py where missing_timer is cancelled without being deleted,

https://github.com/buildbot/buildbot/blob/959950c31fd70006d9b688a345aac8dcb31b990c/master/buildbot/worker/latent.py#L142

I think this might be the problem. Are you using latents?

     if self.missing_timer is not None:
            self.missing_timer.cancel()

should instead be

    self.stopMissingTimer()

noc0lour:
I'm just using normal workers, and I have `missing_timeout` set to just `60` to quickly detect missing workers. The only time this is an issue is after I start the master and no worker attaches for the first `missing_timeout` period. Then `missing_timer` is not `None` and also not `active`.  If I remove a worker during runtime and then reattach it it isn't a problem.

tardyp:
@noc0lour I just did a test trying to reproduce your problem, setting missing_timeout to 1, and I wont reproduce your problem.

It does not make sense to me by looking at the code, and I cannot reproduce as well.
So for me there is another weirdness in your configuration that we need to identify.

I really would like to solve this problem rather than just hiding it.

tardyp:
here are the two quick tests that I've done.
One with the integration tests, one with the e2e smoke tests

https://github.com/tardyp/buildbot/commit/52f5f684e3009844ca6445255cb2f82f0c936b57

both are implementing 1 second missing_timeout, and 2second delay to first worker connection.

noc0lour:
I have a example log: 
https://gist.github.com/noc0lour/b752e9d03415c13ac96407901e3c8731

In this log you see that in the first minute after the master starts RemoteWorker can attach. And after one minute I  get `AlreadyCalled` errors. I can reliably reproduce this if I try to start a worker one minute after my master started.

```
            name,
            passwd,
            properties=self.__properties,
            max_builds=self.__max_builds,
            notify_on_missing="andrej.rode@ettus.com",
            missing_timeout=60)
```
These are the arguments for my workers which have the class `worker.Worker`.
With my patch to check for `active()` as well everything is fine and workers can freely attach and detach, without it everything is hosed after waiting one minute.

tardyp:
@noc0lour I think I would like to see your full master.cfg if you can share it (privately). I'm willing to merge this patch as its pretty harmless (provided you fix the lint issue)

But I really would like to understand the issue. For now, this just doesn't make sense.

tardyp:
@noc0lour I finally found the root cause of this. 

_missing_timer_fired was not called because it was a generator, and not decorated by inlineCallbacks.

I fixed your branch to add some more factorization to this code, preventing this bug class, and added a unit test to improve code coverage.

noc0lour:
@tardyp awesome find :) Shall I continue and finish up?

PR #2710, opened 2017-02-02
Get rid of __cmp__() and implement all 6 rich comparison operators.

rodrigc:
This makes things consistent between Python 2 and Python 3,
where `__cmp__()` is gone.

I tried using @functools.total_ordering, but for the ComparableMixin and fakedb, it
didn't work due to the "interesting" logic in those.  To be safe, I just implemented
all 6 comparison operators, and it worked out.


tardyp:
looks like we need some more unit tests

PR #2713, opened 2017-02-02
checkPidFile: erase old pidfile if the process is not live anymore

tardyp:
Like for buildbot start, which uses twistd directly,

buildbot upgrade-master will now erase an old pidfile if the process is not live anymore instead of just failing.

This will make the ansible setup of metabuildbot more stable.

### DIFF
--- a/master/buildbot/scripts/base.py
+++ b/master/buildbot/scripts/base.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 from future.builtins import range
 
 import copy
+import errno
 import os
 import stat
 import sys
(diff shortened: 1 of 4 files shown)


master/buildbot/scripts/base.py
```diff
@@ -42,6 +43,31 @@ def captureErrors(errors, msg):
...
+        except OSError as why:
+            if why.errno == errno.ESRCH:
+                # The pid doesn't exist.
+                print('Removing stale pidfile %s' % pidfile)
+                os.remove(pidfile)
+            else:
+                raise OSError("Can't check status of PID %s from pidfile %s: %s" %
+                              (pid, pidfile, why[1]))
+        else:
+            raise AssertionError("'%s' exists - is this master still running?")
```
rutsky: Why `AssertionError`, not `RuntimeError`?

tardyp: I thought about it for a while, and reasoned we are actually asserting there is not pidfile here.

I was looking for a BusyError or something, but I couldn't find it. So after looking again, looks like RuntimeError is for the error where you couldn't find anything more appropriate, so fair enough

master/buildbot/test/unit/test_scripts_base.py
```diff
@@ -320,11 +321,61 @@ def test_checkBasedir_no_dir(self):
...
+        with open(os.path.join('test', 'twistd.pid'), 'w') as f:
+            f.write(str(os.getpid()))
         rv = base.checkBasedir(mkconfig())
         self.assertFalse(rv)
         self.assertInStdout('still running')
 
+    @skipUnlessPlatformIs('posix')
+    def test_checkBasedir_bad_pidfile(self):
+        self.activeBasedir()
+        # write our own pid in the file
```
rutsky: `write our own pid in the file` - should be `write invalid string as pid in the file` or something like that

master/buildbot/test/unit/test_scripts_base.py
```diff
@@ -320,11 +321,61 @@ def test_checkBasedir_no_dir(self):
...
+        self.assertInStdout('Removing stale pidfile test')
+        self.assertFalse(os.path.exists(pidfile))
+
+    @skipUnlessPlatformIs('posix')
+    def test_checkBasedir_pidfile_kill_error(self):
+        """
+        if ping-killing the PID file does not work, we should error out.
+        """
+        self.activeBasedir()
+        # write our own pid in the file
```
rutsky: copy/pasted comment too

PR #2715, opened 2017-02-02
Enable flake8 on py3

rodrigc:


master/buildbot/monkeypatches/__init__.py
```diff
@@ -16,7 +16,7 @@
 import sys
 
 from twisted.python import util
-
+from builtins import int
```
seankelly: Need one more blank line here.

tardyp:
fixing this PR, as this is my regression

PR #2716, opened 2017-02-02
REST API: Allow selecting fields and ordering descending at the same time

mikedld:
REST API allows to specify fields to return (via "field=" query param(s)) and sorting order (via "order=" query param(s)); it also requires that order fields are present in fields list. When order is descending field name in "order=" param is to be prepended with "-", and this expression is an invalid field name which cannot be found in fields list (note that prepending "-" to field name in "field=" param is prohibited).

Strip leading "-" from order fields when checking whether all order fields are also present in fields list. This would allow specifying fields to return and sort in descending order at the same time.

### DIFF
--- a/master/buildbot/www/rest.py
+++ b/master/buildbot/www/rest.py
@@ -305,7 +305,7 @@ class V2RootResource(resource.Resource):
         # if ordering or filtering is on a field that's not in fields, bail out
         if fields:
             fieldsSet = set(fields)
-            if order and set(order) - fieldsSet:
+            if order and set([o.lstrip('-') for o in order]) - fieldsSet:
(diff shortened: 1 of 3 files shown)


master/buildbot/www/rest.py
```diff
@@ -305,7 +305,7 @@ def checkFields(fields, negOk=False):
         # if ordering or filtering is on a field that's not in fields, bail out
         if fields:
             fieldsSet = set(fields)
-            if order and set(order) - fieldsSet:
+            if order and set([o.lstrip('-') for o in order]) - fieldsSet:
```
tardyp: good catch!

PR #2717, opened 2017-02-02
Add --category argument to github_buildbot.py

clopez:
This is useful when there are several build categories and you want to run this script to send the change notification to a given category.


tardyp:
glad this stuff still works, as it is not documented not unit tested :-/
anyway, that is contrib...

PR #2719, opened 2017-02-03
Bytes vs. unicode string fixes for Python 3

rodrigc:


tardyp:
looks like on windows,
 "echo 'http://localhost:8080/#builders/1/builds/1'" not in
u'hecho http://localhost:8080/#builders/1/builds/1\nh 

The quote are stripped, no sure why.

tardyp:
Loos like windows is still unhappy about it

rodrigc:
The quotes are stripped on win32 because shell_quote() has different logic on win32: https://github.com/buildbot/buildbot/blob/master/worker/buildbot_worker/runprocess.py#L70

I have updated the patch to take this into account.

PR #2725, opened 2017-02-04
speed up ec2 tests

tardyp:
with the start instance polling sleep, we ended up 44s of tests
now we are in a more reasonable 10s


master/buildbot/test/unit/test_worker_ec2.py
```diff
@@ -53,7 +53,6 @@ def skip_ec2(f):
 
 class TestEC2LatentWorker(unittest.TestCase):
     ec2_connection = None
-
```
seankelly: flake8 disagrees with this change.

seankelly:
Looks like this shaves ~5 minutes off the total CPU time on Travis. Looks good except for the flake8 issue.

PR #2730, opened 2017-02-05
On Python 3, run all tests not specified in .py3.notworking.txt

rodrigc:


tardyp:
I though it would be easier to do the while list, but if you prefer the blacklist, fine by me.
Just remove the whitelisting code and list, then.

rodrigc:
Please clarify your last comment:

```
Just remove the whitelisting code and list, then.
```

I have removed the whitelisting code.  I have also updated the notworking list,
and have got a green build on the py3.5 build.

Do you give approval for this, so that I can merge it?

tardyp:
good work, thanks

PR #2735, opened 2017-02-05
Simplify test a little bit so that it passes on Python 2 and Python 3.

rodrigc:
The format of the exception stacktrace is quite different between
Python 2 and Python 3, and the twisted.python.failure.Failure class
doesn't try to present the exact same stacktrace on Python 2 and 3.
The correct exception is being thrown, so we can still check for that
on both Python versions.


rodrigc:
@seankelly thanks for the review

PR #2738, opened 2017-02-06
With sqlite3, turn on auto-commit mode before doing a VACUUM.

rodrigc:
This fixes a test on Python 3.

### DIFF
--- a/master/buildbot/scripts/cleanupdb.py
+++ b/master/buildbot/scripts/cleanupdb.py
@@ -59,8 +59,18 @@ def doCleanupDatabase(config, master_cfg):
         # sqlite vacuum function rebuild the whole database to claim
         # free disk space back
         def thd(engine):
-            r = engine.execute("vacuum;")
-            r.close()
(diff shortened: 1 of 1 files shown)


master/buildbot/scripts/cleanupdb.py
```diff
@@ -59,8 +59,18 @@ def doCleanupDatabase(config, master_cfg):
...
+            # open transaction before DDL statements.
+            # It is necessary to set the isolation_level to none
+            # for auto-commit mode before doing a VACUUM.
+            # See: https://bugs.python.org/issue28518
+
+            # Get the underlying sqlite connection from SQLAlchemy.
+            sqlite_conn = engine.connection.connection
+            # Set isolation_level to 'auto-commit mode'
+            sqlite_conn.isolation_level = None
+            sqlite_conn.execute("vacuum;").close()
```
tardyp: cool! I did try that, but what I actually try is 

    engine.connection.isolation_level = None

looks like I was missing a level of connection abstraction.

PR #2729, opened 2017-02-05
Add github reporter that can comment on Pull Requests

anish:

PR #2739, opened 2017-02-06
Previous fix for github issues was incomplete

anish:
Previous fix in #2724 only partially fixed the issue

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -110,10 +110,11 @@ class GitHubStatusPush(http.HttpStatusPushBase):
 
         context = yield props.render(self.context)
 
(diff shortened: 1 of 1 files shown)


rodrigc:
Can you add a testcase that exercises the code paths in your new patch?

anish:
@rodrigc Plan to do just that, unfortunately the source stamps are hard coded into the test case helpers, looking for a simpler way.

anish:
Unit tests added as part of larger changes under #2729

anish:
@tardyp Slight modification to #2724 , test coverage coming in #2729

tardyp:
lets merge this one first as it is obviously better than previous

PR #2744, opened 2017-02-06
Add --filter-push-branch option to github_buildbot

clopez:
This allows to ignore any commits not going to a specified branch name on the script github_builbot itself.

For example, to trigger only builds for commits on the master branch you can use --filter-push-branch=master with this script.

This is useful when for any reason is preferred to filter the changes for a branch on this script rather than doing that on the buildbot configuration.

For example, the /console output doesn't seem to honour the `change_filter` defined for a build worker and it instead shows all the changes received (even if those have not triggered a build because they didn't matched the `change_filter`.). By placing the filter instead on github_buildbot this problem can be avoided.

### DIFF
--- a/master/contrib/github_buildbot.py
+++ b/master/contrib/github_buildbot.py
@@ -183,10 +183,16 @@ class GitHubBuildBot(resource.Resource):
         changes = None
         refname = payload['ref']
 
+        if self.filter_push_branch:
+            if refname != "refs/heads/%s" % self.filter_push_branch:
+                logging.info("Ignoring refname '%s': Not a push to branch '%s'"
+                             % (refname, self.filter_push_branch))
+                return changes
+
         m = re.match(r"^refs/(heads|tags)/(.+)$", refname)
         if not m:
             logging.info(
-                "Ignoring refname `%s': Not a branch or a tag", refname)
+                "Ignoring refname '%s': Not a branch or a tag", refname)
             return changes
 
         refname = m.group(2)
@@ -373,6 +379,11 @@ def setup_options():
                       help="Category for the build change",
                       default=None, dest="category")
 
+    parser.add_option("--filter-push-branch",
+                      help="Only trigger builds for pushes to a given "
+                      "branch name.",
+                      default=None, dest="filter_push_branch")
+
     (options, _) = parser.parse_args()
 
     if options.auth is not None and ":" not in options.auth:
@@ -404,6 +415,7 @@ def run_hook(options):
     github_bot.auth = options.auth
     github_bot.head_commit = options.head_commit
     github_bot.category = options.category
+    github_bot.filter_push_branch = options.filter_push_branch
 
     site = server.Site(github_bot)
     reactor.listenTCP(options.port, site)


tardyp:
I am really not convinced we should keep this server in the buildbot source code.
I would prefer to spend time enhancing feature of the hook

If your bot cannot have its  web port open to the internet, fine, just use a small nginx configured to only reverse proxy to the hook endpoints.

before merging I'd like to discuss that in the weekly meeting.
adding an item to the agenda.

https://titanpad.com/buildbot-agenda

clopez:
I have already a proxy-reverse for this. Actually I run this script on the same machine than the buildmaster.

But I choose to use this script instead, because it allows me more control over the ChangeSources with complex configurations.

I currently run a buildmaster that supports triggering new builds from different repositories (not only different Git repositories / forks, but also from SVN repositories). 

That way, I can have several scripts like this, each one monitoring an upstream repository (svn or git). Then I can also filter unwanted commits on the script itself, and finally I can simply send to the buildmaster the changesource message for the commits I want to process. 

I'm not sure I would be able to have this configuration in buildbot itself without using this scripts.

opalmer:
@tardyp I for one would love to see feature parity between the hook and this script. I can also understand the need to run the contrib hook too though if you have some more advanced filtering you need to do. If it were me I would:

* Bring the hook up to feature parity with contribute to the greatest extent possible. I have not audited the code myself so I'm honestly not sure exactly what this would involve.
* Add some additional methods on `GitHubEventHandler` so that filtering and more customized handling could be performed prior to it being submitted to buildbot. This should make it easier to extend the event handler without having to completely replace whole methods.
* Remove most of the code in the contrib script and subclass `GitHubEventHandler` instead or just completely remove it (maybe update the docs on how to extend a hook?)

tardyp:
@clopez @opalmer if I understand correctly, you are two independent users of this feature, right?

Thanks for detailling your usecases anyway.

In buildbot the change filtering is supposed to happen at the scheduler level, and we have the concept of the change filter, which is nice and generic.
http://docs.buildbot.net/latest/manual/cfg-schedulers.html#change-filters

If I understand correctly the requirement you have is to apply the change filters at the level of the change source, so that it does not flood the database with unwanted changes.
This is something that for me makes a lot of sense, and has been requested a lot of time by other users. I dont think this would be complex to implement. just a new parameter to the change source or hooks, so that it accepts a list of change filters.

For the comment about triggering builds from different repositories like GIT/SVN, etc, I am not sure I understand this exactlty, as this is actually the core features of Buildbot. We have gitpoller, svnpoller, and all the change sources that are needed to implement this.

I doing it in another process is a requirement, we now have the multimaster feature which you can setup to have a buildbot master dedicated to change sources.
 
Please let me know if I missed something

Frodox:
@clopez what is your use-case? Why/how do you send changes from svg/git repo directly to `githab_buildbot` ? Why not to use web-hooks for this feature? 

> I currently run a buildmaster that supports triggering new builds from different repositories (not only different Git repositories / forks, but also from SVN repositories).
>
> That way, I can have several scripts like this, each one monitoring an upstream repository (svn or git). Then I can also filter unwanted commits on the script itself, and finally I can simply send to the buildmaster the changesource message for the commits I want to process.

I'm agree, that having multiple `change_source` for many repos (100-500-1k-5k) not optimal way. But personally I use [change www hooks](http://docs.buildbot.net/current/manual/cfg-wwwhooks.html), so I get all changes from all repos to my schedulers. Every scheduler has neede `change_filter` to triger only needed changes in needed branches and so on. It looks transparent and simple enough, doesn't it?

I can't understand use-case, why do you want to use extra 'script' to monitor upstream..:) 
And in any case, script (gitlab, bitbucket one and so on) can add some property and scheduler would filter by that and by branch, if you need.

tardyp:
So we had the meeting today on IRC. Everybody agreed that it would be better to host the contrib directory in a separate project http://github.com/buildbot/buildbot-contrib/. 

We want to focus the core buildbot project to be of quality with the best code coverage and documentation, while letting the contributors share their methods and work even if it is not yet fully tested (in the buildbot-contrib repository, or in their own repository they maintain).

I think github_buildbot was a good tool when hooks were not implemented, but now they are working good, and should be recommended method for everybody.
If you are behind a firewall, you can use nginx to forward the hooks. I tested this week such method and have good results with it. I will document it soon.

We will rediscuss and conclude on this on next weekly meeting, everybody is invited to take part and bring his point, please dont hesitate to join.

https://titanpad.com/buildbot-agenda

tardyp:
@clopez 

I created https://github.com/buildbot/buildbot-contrib which is the new home for the contrib scripts.
I will make a PR later this week to remove the directory, and edit the documentation to point to this.

I've added you to the list of writers for this repository.

clopez:
@tardyp thanks for the write access.

Not sure however what is the policy here. 
It is ok if I commit directly to master? or is better to open a PR and wait for a review?

For the moment, I have re-opened this PR for the new repository at https://github.com/buildbot/buildbot-contrib/pull/1

tardyp:
I believe PR is best to track the changes. Your merge rights allow you to push the merge button for merge requests. I would propose to less people 3-4 days to review the patch, and then merge if its not already

PR #2753, opened 2017-02-07
Fix test_steps_http.py which is periodically failing on Python 3.

rodrigc:
Sometimes the output message contains "HEAD, GET", and sometimes
it contains "GET, HEAD".
The errors status is always 405.

### DIFF
--- a/master/buildbot/test/unit/test_steps_http.py
+++ b/master/buildbot/test/unit/test_steps_http.py
@@ -99,15 +99,7 @@ class TestHTTPStep(steps.BuildStepMixin, unittest.TestCase):
 
     def test_POST(self):
         url = self.getURL("POST")
-        content = "\n<html>\n" \
-                  "  <head><title>405 - Method Not Allowed</title></head>\n" \
-                  "  <body>\n    <h1>Method Not Allowed</h1>\n    <p>Your browser " \
-                  "approached me (at /POST) with the method \"POST\".  I only allow the " \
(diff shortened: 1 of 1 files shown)


master/buildbot/test/unit/test_steps_http.py
```diff
@@ -99,15 +99,19 @@ def test_404(self):
 
     def test_POST(self):
         url = self.getURL("POST")
-        content = "\n<html>\n" \
-                  "  <head><title>405 - Method Not Allowed</title></head>\n" \
-                  "  <body>\n    <h1>Method Not Allowed</h1>\n    <p>Your browser " \
-                  "approached me (at /POST) with the method \"POST\".  I only allow the " \
-                  "methods HEAD, GET here.</p>\n  </body>\n</html>\n"
+        #
```
tardyp: I already disable that test in one of my PR I believe we should fix the thing properly. probably hardcoding the whole log is not a good idea, but looking at some of its content looks better

rodrigc: I looked at the code and it is quite intertwined, so you need to look at all of the output or none.  There is no easy way to do a partial compare of the output. 

In the interest of moving forward, I thought that just checking for a 405 status was a reasonable compromise.

tardyp: in this, case, please just drop the comment, but please don't leave commented out dead code.

rodrigc:
OK, I removed the commented out lines.

PR #2755, opened 2017-02-07
worker: close file handle in FileUpload

noc0lour:

### DIFF
--- a/master/buildbot/newsfragments/fileupload_close.bugfix
+++ b/master/buildbot/newsfragments/fileupload_close.bugfix
@@ -1,0 +1,1 @@
+Fix issue with buildbot_worker not closing file handles when using the transfer steps
--- a/worker/buildbot_worker/commands/transfer.py
+++ b/worker/buildbot_worker/commands/transfer.py
@@ -76,6 +76,7 @@ class WorkerFileUploadCommand(TransferCommand):
         self.keepstamp = args.get('keepstamp', False)
         self.stderr = None
         self.rc = 0
+        self.fp = None
 
     def start(self):
         if self.debug:
@@ -107,6 +108,8 @@ class WorkerFileUploadCommand(TransferCommand):
         self._reactor.callLater(0, self._loop, d)
 
         def _close_ok(res):
+            if self.fp:
+                self.fp.close()
             self.fp = None
             d1 = self.writer.callRemote("close")
 
@@ -118,6 +121,8 @@ class WorkerFileUploadCommand(TransferCommand):
 
         def _close_err(f):
             self.rc = 1
+            if self.fp:
+                self.fp.close()
             self.fp = None
             # call remote's close(), but keep the existing failure
             d1 = self.writer.callRemote("close")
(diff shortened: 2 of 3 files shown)


worker/buildbot_worker/commands/transfer.py
```diff
@@ -107,7 +107,8 @@ def start(self):
         self._reactor.callLater(0, self._loop, d)
 
         def _close_ok(res):
-            self.fp = None
+            if self.fp:
+                self.fp.close()
```
rodrigc: Maybe you should set self.fp = None
after closing it?  Maybe not 100% necessary, but way not?

rodrigc: Also on line 98, in the `except` block where self.fp is set to `None` do we need to close self.fp?
Maybe at the beginning of the function it would be good to initialize self.fp to None before
calling open(), just to make our intention clear.

rodrigc: Can you also comment out these two lines from the .py3.notworking.txt file:

```
buildbot.test.integration.test_transfer.TransferStepsMasterNull.test_transfer
buildbot.test.integration.test_transfer.TransferStepsMasterPb.test_transfer
```

and add that to your review?

noc0lour: I don't think it is necessary to set `self.fp` to `None`. Probably we should change the checks to see if it is `True` instead of checking for `not None`. I don't know if there is a way we could change syntax to the safer `with open(..) as self.fp` form. (Open for every write loop?) I guess this is not as easy  without changing a lot more.

master/buildbot/newsfragments/fileupload_close.bugfix
```diff
@@ -0,0 +1 @@
+Close file handles in the end before closing the connection.
```
tardyp: This is a bit vague sentence for a release note.

```
Fix issue with buildbot_worker not closing file handles when using the transfer steps
```

worker/buildbot_worker/commands/transfer.py
tardyp: I think it is always safe to break the object links and help gc by forcefully put stuff to None when you dont use it anymore.

noc0lour:
So, basically I add self.fp = None after every close?

tardyp:
Yes

noc0lour:
Then this should be good to go now.

rodrigc:
@noc0lour thanks for fixing this!
In https://github.com/buildbot/buildbot/pull/2682 , I plugged most of the file descriptor leaks
reported by Python 3.  For the one you fixed, I looked all over the code, and couldn't figure out where the file was being opened.  I didn't think to look in the buildbot_worker!!  Great you fixed this!

Were all these file descriptor leaks in the code legitimate problems that were overlooked before?  Were they causing resource problems as the buildbot process ran for a long time?

noc0lour:
I don't think they were serious problems as they are read handles and probably garbage collection kills them in the end. As long as you don't exceede the systems file descriptor limit everything is fine. But no problem ;) if a problem annoys me I'm going to fix it (failing tests annoy me ;) )

rodrigc:
@noc0lour You have really good skills in finding and fixing difficult issues!
Can you help me with this one: https://lists.buildbot.net/pipermail/devel/2017-February/012328.html

PR #2757, opened 2017-02-08
Update master Dockerfile to fix build problems

Tranquility:
Otherwise the build fails:

https://hub.docker.com/r/buildbot/buildbot-master/builds/


tardyp:
thanks for this! just in time for the release!

PR #2762, opened 2017-02-10
Add instance id to build properties

dinatale2:
## Contributor Checklist:

Note sure if this is a good idea or not. But, having the instance id is handy when attempting to run various ec2 related tools. Any thoughts?


seankelly:
If we do this for EC2, then I think we should do this for the rest of the latent workers.

tardyp:
looks good, but please watch out the contributor checklist.
- tests
- docs
- release notes

PR #2763, opened 2017-02-10
Add a GitTagPoller that polls for new tags in a git repo

dinatale2:
Create a new change source called the GitTagPoller
which polls a git repository for new tags. Many projects
tag new builds and may wish to build new releases that get
tagged.

Signed-off-by: Giuseppe Di Natale <dinatale2@llnl.gov>

## Contributor Checklist:

Another WIP pull request. Again, not sure how useful this would be. I find it useful to build tagged releases. Thoughts are welcome. If it would be useful, I'll add tests, etc.

### DIFF
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -56,7 +56,7 @@ class GitPoller(base.PollingChangeSource, StateMixin):
                  category=None, project=None,
                  pollinterval=-2, fetch_refspec=None,
                  encoding='utf-8', name=None, pollAtLaunch=False,
-                 buildPushesWithNoCommits=False):
+                 buildPushesWithNoCommits=False, only_tags=False):
 
         # for backward compatibility; the parameter used to be spelled with 'i'
         if pollinterval != -2:
@@ -72,12 +72,17 @@ class GitPoller(base.PollingChangeSource, StateMixin):
         if project is None:
             project = ''
 
+        if only_tags and (branch or branches):
+            config.error("GitPoller: can't specify only_tags and branch/branches")
         if branch and branches:
             config.error("GitPoller: can't specify both branch and branches")
         elif branch:
             branches = [branch]
         elif not branches:
-            branches = ['master']
+            if only_tags:
+                branches = lambda ref: ref.startswith('refs/tags/')
+            else:
+                branches = ['master']
 
         self.repourl = repourl
         self.branches = branches
--- a/master/buildbot/newsfragments/gitpoller_onlytags.feature
+++ b/master/buildbot/newsfragments/gitpoller_onlytags.feature
@@ -1,0 +1,1 @@
+:py:class: `~buildbot.changes.GitPoller` now supports polling tags in a git repository.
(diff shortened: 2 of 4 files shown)


noc0lour:
I wonder if it would make even more sense to extend the current GitPoller implementation with a `only_tags` option? This should have the same fucntionality?

seankelly:
[GitPoller supports it right now](https://irclogs.jackgrigg.com/irc.freenode.net/buildbot/2017-01-17#i_4011022).

    GitPoller(branches=lambda ref: ref.startswith('ref/tags/'))

noc0lour:
I guess then we only have to extend the docs?

tardyp:
I agree we should extend git_poller.
We could have a only_tags option to GitPoller that would translate to branches=lambda ref: ref.startswith('refs/tags/')

tardyp:
@dinatale2 please argue why we should have separate the two classes or close

dinatale2:
I don't think there is a real need for a separate class. If you'd like I can introduce the only_tags option which will set branches accordingly. Another option is to derive a GitTagPoller from GitPoller and do some sanity checks on the arguments in its `__init__` function. @tardyp Thoughts or preferences?

tardyp:
@dinatale2 I think only_tags is more simple

seankelly:
The documentation needs to be updated with the new `only_tags` parameter.

master/buildbot/changes/gitpoller.py
```diff
@@ -72,12 +72,17 @@ def __init__(self, repourl, branches=None, branch=None,
...
+        if only_tags and (branch or branches):
+            config.error("GitPoller: can't specify only_tags and branch/branches")
         if branch and branches:
             config.error("GitPoller: can't specify both branch and branches")
         elif branch:
             branches = [branch]
         elif not branches:
-            branches = ['master']
+            if only_tags:
+                branches = lambda ref: ref.startswith('refs/tags/')
```
tardyp: looks good, but this line is uncovered

dinatale2:
@tardyp I think this is all set.

thmo:
Will I be able to point both, a "standard" `GitPoller` and a `GitPoller` with `only_tags` at the same repository?

Or would I have to set up a `GitPoller` that monitors branches and tags at the same time? Say, using
```
branches=lambda ref: ref.startswith('refs/tags/') or ref.startswith("refs/heads/")
```

dinatale2:
@thmo Based on my understanding of the `GitPoller`, you should be able to do both. My change simply sets up a lambda function for `branches`. The core functionality of the `GitPoller` hasn't changed.

tardyp:
thanks!

thmo:
Two remarks:

1. Maybe my question was not very well formulated. The docs say that only one `GitPoller` should point to a repository at the same time. That means that you can monitor a repository from one master with or without `tags_only`, but not both. There might be a use case, still I think the docs should be made clear wrt that.

2. Maybe more important: Annotated tags will show up in `git ls-remote` with `^{}` appended. I think you have to filter those, otherwise the fetch operations done later by the `GitPoller` will fail with "... failed with exit code 128: fatal: Invalid refspec: ..."

dragomirtodorov:
Hi,

about Annotated tags the quick fix is to use --refs in git ls-remote
```
@@ -136,7 +136,7 @@ class GitPoller(base.PollingChangeSource, StateMixin):
         return str

     def _getBranches(self):
-        d = self._dovccmd('ls-remote', ['--refs', self.repourl])
+        d = self._dovccmd('ls-remote', [self.repourl])

         @d.addCallback
         def parseRemote(rows):

```

PR #2764, opened 2017-02-10
Fix test_www_sse.py on Python 3

rodrigc:


tardyp:
restart to use the new trial script

rodrigc:
Moving to https://github.com/buildbot/buildbot/pull/2811

PR #2769, opened 2017-02-10
Fix test_www_service.py now works on Python 3

rodrigc:


rodrigc:
Moved to https://github.com/buildbot/buildbot/pull/2764

PR #2776, opened 2017-02-10
[DONOTMERGE]Modified makefile for release tarball shpinx docs generation

dol-sen:
This PR is strictly for reference.  This is how I patched the makefile to be able to generate sphinx doc (html) from the 0.9.4 release tarball after adding the missing docs/ files.

I'll update any future releases to any changes you make to support generation from the release tarball.


tardyp:
I did merge a fix which skips towncrier if it is not present

PR #2785, opened 2017-02-13
Fix test_worker_transition on Python 3

rodrigc:


master/buildbot/test/unit/test_worker_transition.py
```diff
@@ -333,8 +333,13 @@ class C(WorkerAPICompatMixin):
...
 
         c = C()
-
-        self.assertRaisesRegexp(
+        if getattr(self, "assertRaisesRegex", None):
+            # Python 3
+            assertFunc = self.assertRaisesRegex
+        else:
+            # Python 2
+            assertFunc = self.assertRaisesRegexp
```
tardyp: We use this in a bunch of tests. We bettr patch TestCase so that it has assertRaisesRegex everywhere

rodrigc:
OK, I added a monkeypatch

PR #2787, opened 2017-02-13
db: fix issue with db being closed by wrong thread

noc0lour:
Fixed an issue reported by @rodrigc on users@bb.net `threads and SQLite problem`.
Problem was that `self.engine.dispose()` was called by the parent thread and not the thread pool.

### DIFF
--- a/master/buildbot/db/pool.py
+++ b/master/buildbot/db/pool.py
@@ -32,6 +32,7 @@ from buildbot.db.changesources import ChangeSourceAlreadyClaimedError
 from buildbot.db.schedulers import SchedulerAlreadyClaimedError
 from buildbot.process import metrics
 
+
 # set this to True for *very* verbose query debugging output; this can
 # be monkey-patched from master.cfg, too:
 #     from buildbot.db import pool
@@ -145,8 +146,8 @@ class DBThreadPool(object):
 
     def _stop(self):
         self._stop_evt = None
+        threads.deferToThreadPool(self.reactor, self._pool, self.engine.dispose)
         self._pool.stop()
-        self.engine.dispose()
         self.running = False
 
     def shutdown(self):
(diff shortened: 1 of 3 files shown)


master/buildbot/db/pool.py
```diff
@@ -145,8 +146,8 @@ def _start(self):
 
     def _stop(self):
         self._stop_evt = None
+        threads.deferToThreadPool(self.reactor, self._pool, self.engine.dispose)
```
tardyp: does this really works?
I've been scratching my head for hours with this. looking at the code, I saw it was actually harmless, as this is in the stop flow.

tardyp: The problem is that this will close the engine in the thread pool, but if there is more than one thread, then you will actually close the other engines in the wrong threads

master/buildbot/test/unit/test_db_pool.py
```diff
@@ -37,7 +37,6 @@ class Basic(unittest.TestCase):
     def setUp(self):
         self.engine = sa.create_engine('sqlite://')
         self.engine.should_retry = lambda _: False
-        self.engine.optimal_thread_pool_size = 1
```
tardyp: So I would not remove this line

tardyp:
@noc0lour I did fix, and merge via github UI. If you agree with my changes, please rebase and squash.
Please add a comment, that we need to have only one thread in order not to have this issue.

noc0lour:
@tardyp thanks, I was curious if we can remove the limitation an number of threads. But apparently not. I'll squash and rebase.

noc0lour:
I this only an issue with tests or are we seeing this kind of issues in production as well?

tardyp:
yeah, sqlite don't like to be run in more than one threads. unrelated to this close issue, there are database is locked instabilities that are triggered when we use more threads.

the commit has been done in 67a8efb403bac4569e71b2ae38d814e73dedf58b
and is marked revertme and re-enable in this ticket http://trac.buildbot.net/ticket/3565

The following PR https://github.com/buildbot/buildbot/pull/2270 has actually enable warning about db locks, and we started seeing a lot of them.

rodrigc:
Yes!! Yes!! Yes!!
Thanks @noc0lour and @tardyp for fixing this.  This was one of the hard problems I encountered during Python 3 porting.

tardyp:
@rodrigc thank *you* for your hard work and commitment!

PR #2791, opened 2017-02-14
Secret returned by files provider. The secret is stored in files.

protatremy:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/secrets/providers/file.py
+++ b/master/buildbot/secrets/providers/file.py
@@ -34,9 +34,11 @@ class SecretInAFile(SecretProviderBase):
     def checkFileIsReadOnly(self, dirname, secretfile):
         filepath = os.path.join(dirname, secretfile)
         obs_stat = stat.S_IMODE(os.stat(filepath).st_mode)
-        if obs_stat not in [stat.S_IRWXU, stat.S_IRUSR]:
-            config.error("the file %s is not user readable only" %
-                         (secretfile))
+        if (obs_stat & 0o77) != 0:
+            config.error("Permissions %s on file %s are too open."
+                         " It is required that your secret files are NOT"
+                         " accessible by others!" % (oct(obs_stat),
+                                                     secretfile))
 
     def checkSecretDirectoryIsAvailableAndReadable(self, dirname, suffixes):
         if not os.access(dirname, os.F_OK):
@@ -55,10 +57,8 @@ class SecretInAFile(SecretProviderBase):
                     with open(os.path.join(dirname, secretfile)) as source:
                         secretvalue = source.read()
                     if suffix:
-                        secrets.update(
-                            {secretfile.rstrip(suffix): secretvalue})
-                    else:
-                        secrets.update({secretfile: secretvalue})
+                        secretfile = secretfile[:-len(suffix)]
+                    secrets[secretfile] = secretvalue
         return secrets
 
     def checkConfig(self, dirname, suffixes=None):
--- a/master/buildbot/secrets/providers/file.py
+++ b/master/buildbot/secrets/providers/file.py
@@ -21,7 +21,7 @@ from __future__ import print_function
 import os
 
 from buildbot import config
-from buildbot.secrets.provider.base import SecretProviderBase
+from buildbot.secrets.providers.base import SecretProviderBase
 
 
 class SecretInAFile(SecretProviderBase):
--- a/master/buildbot/secrets/providers/file.py
+++ b/master/buildbot/secrets/providers/file.py
@@ -30,20 +30,50 @@ class SecretInAFile(SecretProviderBase):
     """
     name = "SecretInAFile"
 
-    def checkConfig(self, name, dirname):
-        if dirname is None:
-            config.error("directory name could not be empty")
+    def checkFileIsReadOnly(self, dirname, secretfile):
+        filepath = os.path.join(dirname, secretfile)
+        if not os.access(filepath, os.R_OK) or os.access(filepath, os.W_OK):
+            config.error("the file %s is not read-only for user" %
+                         (secretfile))
 
-    def reconfigService(self, name, dirname):
+    def checkSecretDirectoryIsAvailableAndReadable(self, dirname, suffix=None):
+        if not os.access(dirname, os.F_OK):
+            config.error("directory %s does not exists" % dirname)
+        for secretfile in os.listdir(dirname):
+            if suffix and secretfile.endswith(suffix):
+                self.checkFileIsReadOnly(dirname, secretfile)
+            elif not suffix:
+                self.checkFileIsReadOnly(dirname, secretfile)
+
+    def loadSecrets(self, dirname, suffix=None):
+        secrets = {}
+        for secretfile in os.listdir(dirname):
+            secretvalue = None
+            if suffix and secretfile.endswith(suffix):
+                with open(os.path.join(dirname, secretfile)) as source:
+                    secretvalue = source.read()
+            elif not suffix:
+                with open(os.path.join(dirname, secretfile)) as source:
+                    secretvalue = source.read()
+            secrets.update({secretfile: secretvalue})
+        return secrets
+
+    def checkConfig(self, dirname, suffix=None):
         self._dirname = dirname
+        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
+        self.secrets = self.loadSecrets(self._dirname, suffix=suffix)
+
+    def reconfigService(self, dirname, suffix=None):
+        self._dirname = dirname
+        self.secrets = {}
+        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
+        self.secrets = self.loadSecrets(self._dirname, suffix=suffix)
 
     def get(self, entry):
         """
         get the value from the file identified by 'entry'
         """
-        filename = os.path.join(self._dirname, entry)
-        assert os.path.isfile(filename), \
-            'File {} does not exist'.format(filename)
-        with open(filename) as source:
-            secret = source.read().strip()
-        return secret
+        if entry in self.secrets.keys():
+            return self.secrets[entry]
+        else:
+            return None
--- a/master/buildbot/secrets/providers/file.py
+++ b/master/buildbot/secrets/providers/file.py
@@ -19,6 +19,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 
 import os
+import stat
 
 from buildbot import config
 from buildbot.secrets.providers.base import SecretProviderBase
@@ -32,48 +33,50 @@ class SecretInAFile(SecretProviderBase):
 
     def checkFileIsReadOnly(self, dirname, secretfile):
         filepath = os.path.join(dirname, secretfile)
-        if not os.access(filepath, os.R_OK) or os.access(filepath, os.W_OK):
-            config.error("the file %s is not read-only for user" %
+        obs_stat = stat.S_IMODE(os.stat(filepath).st_mode)
+        if obs_stat not in [stat.S_IRWXU, stat.S_IRUSR]:
+            config.error("the file %s is not user readable only" %
                          (secretfile))
 
-    def checkSecretDirectoryIsAvailableAndReadable(self, dirname, suffix=None):
+    def checkSecretDirectoryIsAvailableAndReadable(self, dirname, suffixes):
         if not os.access(dirname, os.F_OK):
             config.error("directory %s does not exists" % dirname)
         for secretfile in os.listdir(dirname):
-            if suffix and secretfile.endswith(suffix):
-                self.checkFileIsReadOnly(dirname, secretfile)
-            elif not suffix:
-                self.checkFileIsReadOnly(dirname, secretfile)
+            for suffix in suffixes:
+                if secretfile.endswith(suffix):
+                    self.checkFileIsReadOnly(dirname, secretfile)
 
-    def loadSecrets(self, dirname, suffix=None):
+    def loadSecrets(self, dirname, suffixes):
         secrets = {}
         for secretfile in os.listdir(dirname):
             secretvalue = None
-            if suffix and secretfile.endswith(suffix):
-                with open(os.path.join(dirname, secretfile)) as source:
-                    secretvalue = source.read()
-            elif not suffix:
-                with open(os.path.join(dirname, secretfile)) as source:
-                    secretvalue = source.read()
-            secrets.update({secretfile: secretvalue})
+            for suffix in suffixes:
+                if secretfile.endswith(suffix):
+                    with open(os.path.join(dirname, secretfile)) as source:
+                        secretvalue = source.read()
+                    if suffix:
+                        secrets.update(
+                            {secretfile.rstrip(suffix): secretvalue})
+                    else:
+                        secrets.update({secretfile: secretvalue})
         return secrets
 
-    def checkConfig(self, dirname, suffix=None):
+    def checkConfig(self, dirname, suffixes=None):
         self._dirname = dirname
-        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
-        self.secrets = self.loadSecrets(self._dirname, suffix=suffix)
+        if suffixes is None:
+            suffixes = [""]
+        self.checkSecretDirectoryIsAvailableAndReadable(dirname,
+                                                        suffixes=suffixes)
 
-    def reconfigService(self, dirname, suffix=None):
+    def reconfigService(self, dirname, suffixes=None):
         self._dirname = dirname
         self.secrets = {}
-        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
-        self.secrets = self.loadSecrets(self._dirname, suffix=suffix)
+        if suffixes is None:
+            suffixes = [""]
+        self.secrets = self.loadSecrets(self._dirname, suffixes=suffixes)
 
     def get(self, entry):
         """
         get the value from the file identified by 'entry'
         """
-        if entry in self.secrets.keys():
-            return self.secrets[entry]
-        else:
-            return None
+        return self.secrets.get(entry)
--- a/master/buildbot/secrets/providers/file.py
+++ b/master/buildbot/secrets/providers/file.py
@@ -34,7 +34,7 @@ class SecretInAFile(SecretProviderBase):
     def checkFileIsReadOnly(self, dirname, secretfile):
         filepath = os.path.join(dirname, secretfile)
         obs_stat = stat.S_IMODE(os.stat(filepath).st_mode)
-        if (obs_stat & 0o77) != 0 and os.name=="posix":
+        if (obs_stat & 0o77) != 0 and os.name == "posix":
             config.error("Permissions %s on file %s are too open."
                          " It is required that your secret files are NOT"
                          " accessible by others!" % (oct(obs_stat),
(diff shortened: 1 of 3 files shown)


master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class SecretInAFile(SecretProviderBase):
+    """
+    secret is stored in a separate file under the given directory name
+    """
+    name = "SecretInAFile"
+
+    def checkConfig(self, name, dirname):
+        if dirname is None:
```
tardyp: maybe check that the directory exists and is a directory?

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,49 @@
...
+            config.error("directory name could not be empty")
```
tardyp: directory name cannot be None

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def reconfigService(self, name, dirname):
+        self._dirname = dirname
+
+    def get(self, entry):
+        """
+        get the value from the file identified by 'entry'
+        """
+        filename = os.path.join(self._dirname, entry)
+        assert os.path.isfile(filename), \
```
tardyp: if it does not exist, that means this provider does not provide this secret, then we should return None

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,49 @@
...
+            'File {} does not exist'.format(filename)
+        with open(filename) as source:
+            secret = source.read().strip()
```
tardyp: I think we should make the strip optional, or remove it altogether. what if some secret need an empty lines in the end of the secret?

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,59 @@
...
+        self.srvfile = SecretInAFile("name", "/path/todirname")
+        yield self.srvfile.startService()
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        yield self.srvfile.stopService()
+
+    def testCheckConfigSecretInAFileService(self):
+        self.assertEqual(self.srvfile.name, "SecretInAFile")
+        self.assertEqual(self.srvfile._dirname, "/path/todirname")
```
tardyp: _dirname is part of the private API, not sure if it is worth to test

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,59 @@
...
+         lambda: self.srvfile.checkConfig("name2", None))
+
+    @defer.inlineCallbacks
+    def testReconfigSecretInAFileService(self):
+        yield self.srvfile.reconfigService("name2", "/path/todirname2")
+        self.assertEqual(self.srvfile.name, "SecretInAFile")
+        self.assertEqual(self.srvfile._dirname, "/path/todirname2")
+
+    def testGetFile(self):
+        (fd, name) = mkstemp()
```
tardyp: use self.mktemp()

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,59 @@
...
+        os.write(fd, 'key value')
+        value = self.srvfile.get(name)
+        self.assertEqual(value, "key value")
+        os.close(fd)
```
tardyp: need to delete the temp file in the end of the test

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @defer.inlineCallbacks
+    def testReconfigSecretInAFileService(self):
+        yield self.srvfile.reconfigService("name2", "/path/todirname2")
+        self.assertEqual(self.srvfile.name, "SecretInAFile")
+        self.assertEqual(self.srvfile._dirname, "/path/todirname2")
+
+    def testGetFile(self):
+        (fd, name) = mkstemp()
+        os.write(fd, 'key value')
```
tardyp: this is on py3 compliant. which self.mktemp, we will use normal python files, so that should be easier (make sure you use e.g b'key value'  to create byte strings)

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,49 @@
...
+from buildbot.secrets.providers.base import SecretProviderBase
+
+
+class SecretInAFile(SecretProviderBase):
+    """
+    secret is stored in a separate file under the given directory name
+    """
+    name = "SecretInAFile"
+
+    def checkConfig(self, name, dirname):
```
tardyp: I think it would make sense to add an optional list of suffixes e.g ('.pass', '.txt', '.cred')

tardyp: need to check that all the files in that directory are only readable by the current user.

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,49 @@
...
+        if dirname is None:
+            config.error("directory name could not be empty")
+
+    def reconfigService(self, name, dirname):
+        self._dirname = dirname
```
tardyp: probably it would be faster to load all the secrets at reconfig, so that the get is only a dictionary lookup

master/buildbot/test/unit/test_secret_in_file.py
protatremy: https://docs.python.org/2/library/tempfile.html#tempfile.mktemp says that mktemp is deprecated

tardyp: http://twistedmatrix.com/documents/current/api/twisted.trial.unittest.SynchronousTestCase.html#mktemp

master/buildbot/secrets/providers/file.py
protatremy: removed

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,79 @@
...
+from buildbot.secrets.providers.base import SecretProviderBase
+
+
+class SecretInAFile(SecretProviderBase):
+    """
+    secret is stored in a separate file under the given directory name
+    """
+    name = "SecretInAFile"
+
+    def checkFileIsReadOnly(self, dirname, secretfile):
```
tardyp: The file should not be read-only for that user, but it should be readable only by this user.
Means the owner should be current user, and the access control should be 700 or something like this (not group readble, not world readable)

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def checkConfig(self, dirname, suffix=None):
+        self._dirname = dirname
+        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
+        self.secrets = self.loadSecrets(self._dirname, suffix=suffix)
+
+    def reconfigService(self, dirname, suffix=None):
+        self._dirname = dirname
+        self.secrets = {}
+        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
```
tardyp: no need to redo the check.

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,79 @@
...
+        secrets = {}
+        for secretfile in os.listdir(dirname):
+            secretvalue = None
+            if suffix and secretfile.endswith(suffix):
+                with open(os.path.join(dirname, secretfile)) as source:
+                    secretvalue = source.read()
+            elif not suffix:
+                with open(os.path.join(dirname, secretfile)) as source:
+                    secretvalue = source.read()
+            secrets.update({secretfile: secretvalue})
```
tardyp: you need to strip the suffix from the secret file.

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,79 @@
...
+            if suffix and secretfile.endswith(suffix):
+                self.checkFileIsReadOnly(dirname, secretfile)
+            elif not suffix:
+                self.checkFileIsReadOnly(dirname, secretfile)
+
+    def loadSecrets(self, dirname, suffix=None):
+        secrets = {}
+        for secretfile in os.listdir(dirname):
+            secretvalue = None
+            if suffix and secretfile.endswith(suffix):
```
tardyp: I had in mind a list of possible suffixes. You can simplify the algorithm.
```
for suffix in suffixes:
     if secretfile.endswith(suffix): # also works with suffix == ''
           secret_key = secretfile[:-len(suffix)]
           with open(os.path.join(dirname, secretfile)) as source:
                  secrets[secret_key] = source.read() 
```

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,79 @@
...
+        self._dirname = dirname
+        self.secrets = {}
+        self.checkSecretDirectoryIsAvailableAndReadable(dirname, suffix=suffix)
+        self.secrets = self.loadSecrets(self._dirname, suffix=suffix)
+
+    def get(self, entry):
+        """
+        get the value from the file identified by 'entry'
+        """
+        if entry in self.secrets.keys():
```
tardyp: equivalent to
```
return self.secrets.get(entry)
```

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,104 @@
...
+from twisted.trial import unittest
+
+from buildbot.secrets.providers.file import SecretInAFile
+from buildbot.test.util.config import ConfigErrorsMixin
+
+
+class TestSecretInFile(ConfigErrorsMixin, unittest.TestCase):
+
+    def createTempDir(self, dirname):
+        tempdir = self.mktemp()
```
tardyp: see https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/util/integration.py#L60
for a better example on how to create a temp dir

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,104 @@
...
+            filetmp.close()
+        return filetmp, file_path
+
+    @defer.inlineCallbacks
+    def setUp(self):
+        self.tmp_dir = self.createTempDir("temp")
+        filetmp, self.filepath = self.createFileTemp(self.tmp_dir,
+                                                     "tempfile.txt",
+                                                     text="key value")
+        os.chmod(self.filepath, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
```
tardyp: should only be stat.S_IRUSR, no stat.S_IRGRP

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,104 @@
...
+        basedir = os.path.dirname(tempdir)
+        tempdir = os.path.join(basedir, "temp")
+        os.mkdir(tempdir)
+        return tempdir
+
+    def createFileTemp(self, tempdir, filename, text=None):
+        file_path = os.path.join(tempdir, filename)
+        filetmp = open(file_path, 'w')
+        with open(file_path, 'w') as filetmp:
+            if text:
```
tardyp: why if text?

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,104 @@
...
+                filetmp.write(text)
+            filetmp.close()
```
tardyp: ``with`` will actually take care of closing the file

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,104 @@
...
+        filetmp, self.filepath = self.createFileTemp(self.tmp_dir,
+                                                     "tempfile.txt",
+                                                     text="key value")
+        os.chmod(self.filepath, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
+        self.srvfile = SecretInAFile(self.tmp_dir)
+        yield self.srvfile.startService()
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        os.remove(self.filepath)
```
tardyp: no real need to remove the files (for debug)

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,104 @@
...
+        yield self.srvfile.stopService()
+
+    def testCheckConfigSecretInAFileService(self):
+        self.assertEqual(self.srvfile.name, "SecretInAFile")
+        self.assertEqual(self.srvfile._dirname, self.tmp_dir)
+
+    def testCheckConfigErrorSecretInAFileService(self):
+        file_path_not_readable, filepath = self.createFileTemp(self.tmp_dir,
+                                                               "tempfile2.txt")
+        os.chmod(filepath, stat.S_IWRITE)
```
tardyp: stat.S_IRGRP should raise error, not stat.s_IWRITE

need to make sure this works on non posix system (aka windows).

master/buildbot/test/unit/test_secret_in_file.py
protatremy: because I create files with or without text for unit tests

master/buildbot/test/unit/test_secret_in_file.py
protatremy: https://docs.python.org/2/library/os.html
Note Although Windows supports chmod(), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

master/buildbot/secrets/providers/file.py
protatremy: replaced secretfile[:-len(suffix)] by strip

tardyp: no, strip will not work.
```"xxx.txt".rstrip('.txt')```

I see not other option as an if for that part

File: master/buildbot/test/unit/test_secret_in_file.py (hunk 18 above)
Comment (tardyp) [inline master/buildbot/test/unit/test_secret_in_file.py]: it does not make sense to create them without text, it make sense to create them with empty text in that case you make text=''

File: master/buildbot/secrets/providers/file.py [hunk 22]
```diff
@@ -32,48 +33,50 @@ class SecretInAFile(SecretProviderBase):
...
-                    secretvalue = source.read()
-            elif not suffix:
-                with open(os.path.join(dirname, secretfile)) as source:
-                    secretvalue = source.read()
-            secrets.update({secretfile: secretvalue})
+            for suffix in suffixes:
+                if secretfile.endswith(suffix):
+                    with open(os.path.join(dirname, secretfile)) as source:
+                        secretvalue = source.read()
+                    if suffix:
```
tardyp: rstrip will not work + what is the point of using update with a single entry dictionary (why not just assign?)
```
if suffix:
    secretfile = secretfile[:-len(suffix)]

secret[secretfile] = secretvalue
```

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -66,8 +63,8 @@ def testCheckConfigSecretInAFileService(self):
     def testCheckConfigErrorSecretInAFileService(self):
         file_path_not_readable, filepath = self.createFileTemp(self.tmp_dir,
                                                                "tempfile2.txt")
-        os.chmod(filepath, stat.S_IWRITE)
-        expctd_msg_error = "the file tempfile2.txt is not read-only for user"
+        os.chmod(filepath, stat.S_IRGRP)
+        expctd_msg_error = "the file tempfile2.txt is not user readable only"
```
tardyp: This message is not clear for me. Please inspire from the message from ssh

```
Permissions 0640 for 'id_rsa' are too open.
It is required that your private key files are NOT accessible by others.
```

master/buildbot/secrets/providers/file.py
```diff
@@ -32,48 +33,50 @@ class SecretInAFile(SecretProviderBase):
 
     def checkFileIsReadOnly(self, dirname, secretfile):
         filepath = os.path.join(dirname, secretfile)
-        if not os.access(filepath, os.R_OK) or os.access(filepath, os.W_OK):
-            config.error("the file %s is not read-only for user" %
+        obs_stat = stat.S_IMODE(os.stat(filepath).st_mode)
+        if obs_stat not in [stat.S_IRWXU, stat.S_IRUSR]:
```
tardyp:
```
if (obs_stat & 0o77) != 0: # if there are group or world permissions set
    config.error("Permissions %o on file %s are too open. It is required that your secret files are NOT accessible by others!" % ( obs_stat, secretfile))

File: master/buildbot/test/unit/test_secret_in_file.py (hunk 18 above)
Comment (protatremy) [inline master/buildbot/test/unit/test_secret_in_file.py]: ok done

File: master/buildbot/test/unit/test_secret_in_file.py [hunk 25]
```diff
@@ -35,11 +35,12 @@ def createTempDir(self, dirname):
         return tempdir.path
 
     def createFileTemp(self, tempdir, filename, text=None):
+        if text is None:
```
tardyp: this initialization pattern is when you initialize to a list of dict, if you initialize to a string(immutable), this is better to initialize in the default method parameter

tardyp:
@protatremy you will also need to fix windows and python3!

master/buildbot/secrets/providers/file.py
Frodox: @tardyp Which 'that' user? Do you mean *user, which runs bb-master* ?

tardyp: @Frodox correct

master/buildbot/secrets/providers/file.py
```diff
@@ -0,0 +1,82 @@
...
+class SecretInAFile(SecretProviderBase):
+    """
+    secret is stored in a separate file under the given directory name
+    """
+    name = "SecretInAFile"
+
+    def checkFileIsReadOnly(self, dirname, secretfile):
+        filepath = os.path.join(dirname, secretfile)
+        obs_stat = stat.S_IMODE(os.stat(filepath).st_mode)
+        if (obs_stat & 0o77) != 0 and os.name == "posix":
```
tardyp: 👍

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,114 @@
...
+                                                     "tempfile.txt",
+                                                     text="key value")
+        os.chmod(self.filepath, 0o700)
+        self.srvfile = SecretInAFile(self.tmp_dir)
+        yield self.srvfile.startService()
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        yield self.srvfile.stopService()
+        # windows requires to close all existing open temp files not closed by
```
tardyp: looks dirty. Why open temp files were not open with "with open" in the first place?

master/buildbot/test/unit/test_secret_in_file.py
```diff
@@ -0,0 +1,114 @@
...
+        for tempfile in os.listdir(self.tmp_dir):
+            with open(os.path.join(self.tmp_dir, tempfile), 'w') as filetmp:
+                filetmp.close()
+
+    def testCheckConfigSecretInAFileService(self):
+        self.assertEqual(self.srvfile.name, "SecretInAFile")
+        self.assertEqual(self.srvfile._dirname, self.tmp_dir)
+
+    def testCheckConfigErrorSecretInAFileService(self):
+        if not os.name == "posix":
```
tardyp:
```
if os.name != "posix":
     raise SkipTest("Permission checks only works on posix systems")
```

PR #2796, opened 2017-02-15
Fix test_steps_source_svn on Python 3

rodrigc:
I tried various combinations to get this test to work on Python 2 and Python 3,
but due to some corner cases where "somestring" is unicode on Python 3,
and bytes on Python 2, and the bytes in this testcase can't be decoded as ASCII,
I couldn't get this to work any other way.

### DIFF
--- a/.py3.notworking.txt
+++ b/.py3.notworking.txt
@@ -13,7 +13,6 @@ buildbot.test.unit.test_schedulers_trysched.Try_Jobdir.test_parseJob_v5
 buildbot.test.unit.test_schedulers_trysched.Try_Jobdir.test_parseJob_v5_invalid_json
 buildbot.test.unit.test_schedulers_trysched.Try_Jobdir.test_parseJob_v5_no_builders
 buildbot.test.unit.test_schedulers_trysched.Try_Jobdir.test_parseJob_v5_no_properties
-buildbot.test.unit.test_steps_source_svn.TestSvnUriCanonicalize.test_quote_funny_chars
 buildbot.test.unit.test_www_config.IndexResource.test_render
 buildbot.test.unit.test_www_rest.V2RootResource_REST.test_api_collection
 buildbot.test.unit.test_www_rest.V2RootResource_REST.test_api_collection_fields
--- a/master/buildbot/compat.py
+++ b/master/buildbot/compat.py
@@ -1,0 +1,30 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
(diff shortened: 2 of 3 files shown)


master/buildbot/test/unit/test_steps_source_svn.py
```diff
@@ -2092,7 +2092,10 @@ def _makeSUCTest(input, exp):
     test_dont_remove_port_other = _makeSUCTest(
         "https://foo.com:2093/bar", "https://foo.com:2093/bar")
     test_quote_funny_chars = _makeSUCTest(
-        "http://foo.com/\x10\xe6%", "http://foo.com/%10%E6%25")
+        "http://foo.com/\x10\xe6%",
+        ("http://foo.com/%10%E6%25",    # Python 2
+         "http://foo.com/%10%C3%A6%25"  # Python 3
```
tardyp: wait a minute.. I think this just means the svnUriCanonicalize does not work as expected with python3.

if for svn, url canonyfication of http://foo.com/\x10\xe6% is supposed to return http://foo.com/%10%E6%25, we cannot make it return http://foo.com/%10%C3%A6%25.

I think what can be assumed here is that the input of svnUriCanonicalize is supposed to be bytes (and not native string). Wouldn't it simplify the problem you had?

rodrigc: Using bytes doesn't work, and results in ascii decode errors on Python 3.

tardyp: normal, it is not ascii. We really need to fix svnUriCanonicalize to work

rodrigc:
@noc0lour I need your help with your superior debugging skills.  Can you help with this one?
If I set up a Python 3 environment with this: https://lists.buildbot.net/pipermail/devel/2017-February/012310.html

and then type:

```
trial buildbot.test.unit.test_steps_source_svn.TestSvnUriCanonicalize.test_quote_funny_chars
```

I get a test failure.  I tracked it down to somewhere inside urllib unquote(), where
on Python 2, things get decoded with latin-1 codec, while on Python 3, things get decoded with
ascii, so things get decoded differently.

noc0lour:
With your branch I don't get a test failure on python3?

rodrigc:
Re-run the test from master branch.  @tardyp doesn't like the fix I did here.  I need another pair of eyes and smart brain (yours!) to look at it from another angle.

noc0lour:
@rodrigc https://github.com/noc0lour/buildbot/tree/fix_svn_urlparse
Not very pythonic though, urlparse behaves differently in Python2/Python3 and you can specify `encoding` in Python3 but not Python2 to fix it. So I just check for Python3 and set urlopen to the correct function. But it looks... dirty. Test passed now.

rodrigc:
@noc0lour can you submit a PR for the patch you came up with?  I don't think there is a clean solution for this, so let's just move forward with this and see where it goes.

tardyp:
I find @noc0lour solution pretty clean.

We may add it in from  buildbot.compat import urlquote

rodrigc:
@noc0lour based on commits done to my branch by @tardyp, I took your solution in a slightly different direction.  Please take a look.

noc0lour:
I think your solution now is fine as well. i think I can close my PR now ;)

rodrigc:
@noc0lour thank you for investigating and leading to the solution!

tardyp:
great cooperation guys thank you!

noc0lour:
No problem, I'm glad I can help out :) Since my virtualenvs on my machines are organized using virtualevnwrapper it takes literally no time to test on py2/py3.

PR #2798, opened 2017-02-15
Fix test_www_auth on Python 3

rodrigc:


master/buildbot/test/unit/test_www_auth.py
```diff
@@ -122,7 +122,7 @@ def test_maybeAutoLogin_no_header(self):
         try:
             yield self.auth.maybeAutoLogin(self.request)
         except Error as e:
-            self.assertEqual(str(e.status), '403')
+            self.assertEqual(e.status, b'403')
```
seankelly: This now fails in 2.7. Is `e.status` an int?

rodrigc:
Turns out that in older versions of Twisted, `e.status` is an `int`,  but in newer versions of Twisted, `e.status` is bytes.
I've changed the patch, so it works now.

PR #2807, opened 2017-02-16
make buildsteps comparable, and representable

tardyp:
This will help unit tests, and debugging

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -747,7 +747,7 @@ class _Renderer(util.ComparableMixin, object):
         return d
 
     def __repr__(self):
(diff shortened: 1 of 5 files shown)


master/buildbot/process/properties.py
```diff
@@ -747,7 +747,7 @@ def getRenderingFor(self, props):
         return d
 
     def __repr__(self):
-        return 'renderer(%r)' % (self.getRenderingFor,)
+        return 'renderer(%r)' % (self.fn,)
```
seankelly: Is this tested?

tardyp: now it is

PR #2810, opened 2017-02-18
source: fix different url parsing in python2/python3

noc0lour:


noc0lour:
based on fix by @rodrigc, but used the simpler solution with a lamda instead of creating a function. But for sure we can go as well with creating a function in `buildbot.compat`

tardyp:
closed in favor of #2796

PR #2811, opened 2017-02-19
Fix test_www_rest on Python 3

rodrigc:
This also fixes many related tests.


tardyp:
wow thats a big one! not many tests are left when its merged
The smoke test are broken on the first run. I did a merge commit. Lets see if this was false positive.

rodrigc:
Can you point me to where the smoke tests are?
I'm actually interested in learning how to set up protractor.js and xvfb to set up CI testing for a web-based application I am working on.

rodrigc:
I couldn't avoid this being a big patch, BTW.  The rest and www stuff has a lot of interrelated pieces.
This is one of thie last major barriers to all the tests passing on Python 3.

tardyp:
@rodrigc the smoke tests are in the smokes directoy
Following file runs the tests.

https://github.com/buildbot/buildbot/blob/master/smokes/run.sh#L1

That run.sh is not very good, because it does not cleanup correctly when the tests fails, but still they are  simple recipe on how to run the tests

The dockerfile here has the dependencies recipe to run then on buildbot
https://github.com/buildbot/metabbotcfg/blob/nine/docker/metaworker/Dockerfile

thats basically following if I remember correctly

        # selenium is a java thing
        default-jre \
        # chromium needs xvfb even with --headless
        xvfb \
        chromium-browser \

master/buildbot/test/fake/fakedb.py
```diff
@@ -1655,7 +1656,7 @@ def setState(self, objectid, name, value):
 
     def atomicCreateState(self, objectid, name, thd_create_callback):
         value = thd_create_callback()
-        self.states[objectid][name] = json.dumps(value)
+        self.states[objectid][name] = json.dumps(bytes2NativeString(value))
```
rodrigc: This is needed because if you do this:
```
json.dumps(b"hello world")
```

you get this:

```
TypeError: Object of type 'bytes' is not JSON serializable
```

tardyp: that is also the case for  

    json.dumps({'a': b'b'})

but bytes2NativeString will not handle that case.

How many callers have that issue?

Normally, the contract for using db.state is that the input is json serializable.

So I would say the bytes2NativeString should be done by the caller.
And we should make sure that the caller is okay to receive native string when the data is deserialized.

master/buildbot/test/integration/test_www.py
```diff
@@ -139,10 +142,10 @@ def apiGet(self, url, expect200=True):
         if expect200 and pg.code != 200:
             self.fail("did not get 200 response for '%s'" % (url,))
 
-        defer.returnValue(json.loads(body))
+        defer.returnValue(json.loads(bytes2NativeString(body)))
 
     def link(self, suffix):
-        return self.url + 'api/v2/' + suffix
+        return unicode2bytes(self.url) + b'api/v2/' + suffix
```
tardyp: in setup(), self.url is already ensured to be bytes, so that unicode2bytes is not needed

master/buildbot/www/service.py
```diff
@@ -132,6 +133,12 @@ def uid(self):
         return jwt.encode(claims, self.site.session_secret, algorithm=SESSION_SECRET_ALGORITHM)
 
 
+class BuildbotLogFile(LogFile):
```
tardyp: Is this really needed?

I think this is rather an issue with twisted's LogFile.

tardyp:
I can see two problematics hunks in this PR.
I would suggest we merge the unproblematic rest of this PR, and discuss the two remaining patches in 2 separates PR

tardyp:
to move faster, I uploaded the suggested change to your branch.

master/buildbot/test/fake/fakedb.py
rodrigc: Well, that's your opinion.  To have a working branch, I made the change as I did.
Without the change I did, things break.
Hunting down all the callers and making sure they are doing things right with native string
is a lot of work, with potential to break more things in other places.

master/buildbot/www/service.py
rodrigc: This is definitely needed.  That's why I added it.
Without it, the Twisted logger opens in text-mode `"w"`, but there are many places
where the code tries to write bytes to the logger, and things break.
This behavior in the Twisted logger cannot be easily changed or patched.

That's why I went with the subclass of the Logger class.

rodrigc:
Well if you are going to push things to my branch, they are not really suggestions, and are they?

I don't mind you giving review feedback, but I don't like what you have done with pushing things
to a branch without testing.  Are you going to fix the lint failures you have introduced
in the branch?
Are you going to fix the test failures that you have introduced?

Or do I have to clean up the stuff you have broken?
Please clarify.

tardyp:
CI is made for testing. I wanted to quickly see which test I needed to re-add to the blacklist

rodrigc:
OK, as long as you are going to fix the branch.

It would not have been too difficult for you to run the tests locally, and at least
push changes which were not so broken.
I've asked you a few times to run a py3 virtualenv locally,
but you seem to be resistant to changing the way you work.

master/buildbot/test/fake/fakedb.py
tardyp: I know this is nearly done, lets keep calm and continue working cleanly.

I needed to change only one place to fix the test_www. see:	aba82ca

tardyp:
I wanted to give a quick shot, and I hadn't time to run locally. running tests locally takes twice as long as the few fixes I have made

master/buildbot/www/service.py
tardyp: The stack trace lead me to https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http.py#L2801

which had all the code to do the decoding. just neded to set self._nativeize

tardyp:
@rodrigc Please merge if you agree with my changes

rodrigc:
@tardyp Looks good!  Thanks for the fixes!

PR #2812, opened 2017-02-20
Fix test_www_config on Python 3

rodrigc:


rodrigc:
@tardyp nine.buildbot.net seems to be "stuck".  Any ideas?

tardyp:
nine.buildbot.net still has upgrade difficulties :-(

PR #2814, opened 2017-02-21
Fix routingKey for Python 3

rodrigc:
routingKey must be bytes


rodrigc:
@noc0lour sir, I have great need of your superior debugging skills! :)

I fixed a bunch of tests on Python 3 that have to do with scheduling,
but there is one test that I can't figure out.

Can you check out my branch and do:

```
trial buildbot.test.integration.test_try_client.Schedulers.test_userpass_wait
```

I'm getting a test failure and can't figure out how to fix it.
I think that buildbot is not sending the *finished* message for this test,
and I can't figure out why.

Could you help me?

Thanks.

tardyp:
I think from a programming perspective, it makes more sense to make the routing keys native strings or unicode.

They are textual identifiers. For example you can't use bytes for dict identifiers.

This will probably make the patch shorter to keep them native strings.

I'm sorry to have to push back on this as I realize this was a lot of work ton convert all those native strings to byte.

noc0lour:
@rodrigc I can take a look tomorrow if this is still an issue.

rodrigc:
@noc0lour wait a bit, I need to re-work this patch to incorporate @tardyp's feedback.

I have some other issues in need of your brainpower. :)

tardyp:
superseeded by #2816

rodrigc:
@noc0lour this issue has been fixed, but I still need your help on another issue. :)

I have a bug which only occurs on Python 3.6 (not on Python 3.5).

If I do:

```
trial buildbot.test.unit.test_www_rest.V2RootResource.test_default_origin
```

I get:

```
[FAIL]
  File "/Users/crodrigues/buildbot2/master/buildbot/test/unit/test_www_rest.py", line 106, in test_default_origin
    [r.pattern for r in self.rsrc.origins], [r'http\:\/\/server\Z(?ms)'])
  File "/Users/crodrigues/twisted8/src/twisted/trial/_synctest.py", line 432, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 821, in assertEqual
  File "/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 1019, in assertListEqual
    self.assertSequenceEqual(list1, list2, msg, seq_type=list)
  File "/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/lib/python3.6/unittest/case.py", line 1001, in assertSequenceEqual
    self.fail(msg)
twisted.trial.unittest.FailTest: Lists differ: ['(?s:http\\:\\/\\/server)\\Z'] != ['http\\:\\/\\/server\\Z(?ms)']

First differing element 0:
'(?s:http\\:\\/\\/server)\\Z'
'http\\:\\/\\/server\\Z(?ms)'

- ['(?s:http\\:\\/\\/server)\\Z']
?   ----                   -

+ ['http\\:\\/\\/server\\Z(?ms)']
?                         +++++


buildbot.test.unit.test_www_rest.V2RootResource.test_default_origin
Ran 1 tests in 0.090s

FAILED (failures=1)
```

Can you help with this?
Note, you need Python 3.6 to reproduce this.

tardyp:
this regexp is created by https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/rest.py#L464-L465

             self.origins.append(re.compile(fnmatch.translate(origin)))

I would say that fnmatch.translate has changed its algorithm bw 3.5 and 3.6 ( I could not find any reference to it in the changelog though)

So we can either do the fix:

- change test to check against the two outputs possible
- change test to make sure the regexp match url that it is supposed to match, and does not match url it is not supposed to match.

rodrigc:
@tardyp, this change in fnmatch was caused by this commit in cpython: https://github.com/python/cpython/commit/bd48d27944453ad83d3ce37b2c867fa0d59a1c15

tardyp:
@rodrigc I am taking care of this test. I will implement second proposal

PR #2816, opened 2017-02-22
Switch routingKey to be a tuple of str (native string) on Python 2 and 3

rodrigc:


tardyp:
Cool! that was easier than expected!

PR #2817, opened 2017-02-23
Interpolate secrets

protatremy:
interpolate should return corresponding secrets values

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -382,9 +382,8 @@ class _Lookup(util.ComparableMixin, object):
 
     @defer.inlineCallbacks
     def getRenderingFor(self, build):
-        value = build.render(self.value)
-        index = build.render(self.index)
-        value, index = yield defer.gatherResults([value, index])
+        value = yield build.render(self.value)
+        index = yield build.render(self.index)
         if index not in value:
             rv = yield build.render(self.default)
         else:
@@ -428,13 +427,14 @@ class _SecretRenderer(object):
 
     @defer.inlineCallbacks
     def getRenderingFor(self, build):
-        servicenames = [service.name for service in build.getBuild().master.services]
-        if "secrets" not in servicenames:
-            defer.returnValue(None)
+        secretsSrv = build.getBuild().master.namedServices.get("secrets")
+        if not secretsSrv:
+            raise KeyError("secrets service not found")
         credsservice = build.getBuild().master.namedServices['secrets']
         secret_detail = yield credsservice.get(self.secret_name)
-        if secret_detail is not None:
-            defer.returnValue(secret_detail.value)
+        if secret_detail is None:
+            raise KeyError("secret value for key %s is None" % self.secret_name)
+        defer.returnValue(secret_detail.value)
 
 
 class _SecretIndexer(object):
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -429,7 +429,7 @@ class _SecretRenderer(object):
     @defer.inlineCallbacks
     def getRenderingFor(self, build):
         servicenames = [service.name for service in build.getBuild().master.services]
-        if not "secrets" in servicenames:
+        if "secrets" not in servicenames:
             defer.returnValue(None)
         credsservice = build.getBuild().master.namedServices['secrets']
         secret_detail = yield credsservice.get(self.secret_name)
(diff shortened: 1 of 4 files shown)


master/buildbot/process/properties.py
```diff
@@ -421,6 +421,27 @@ def getRenderingFor(self, build):
 
 
 @implementer(IRenderable)
+class _SecretRenderer(object):
+
+    def __init__(self, secret_name):
+        self.secret_name = secret_name
+
+    def getRenderingFor(self, build):
+        credsservice = build.getBuild().master.namedServices['secrets']
```
tardyp: what if secret manager is not configured? I think you should add a  .get() and if credsservice is not None

master/buildbot/test/unit/test_interpolate_secrets.py
```diff
@@ -0,0 +1,75 @@
...
+
+from buildbot.process.properties import Interpolate
+from buildbot.secrets.providers.base import SecretProviderBase
+from buildbot.secrets.secret import SecretDetails
+from buildbot.test.fake import fakemaster
+from buildbot.test.fake.fakebuild import FakeBuild
+from buildbot.test.util.config import ConfigErrorsMixin
+
+
+class FakeSecretStorage(SecretProviderBase):
```
tardyp: Didn't I see this fake already? I think you should put it in buildbot.test.fake.secrets

master/buildbot/test/unit/test_interpolate_secrets.py
```diff
@@ -0,0 +1,75 @@
...
+        return FakeService(self.dictionary)
+
+
+class TestInterpolateSecrets(unittest.TestCase, ConfigErrorsMixin):
+
+    def setUp(self):
+        self.master = fakemaster.make_master()
+        self.master.config.secretsManagers = [FakeSecretStorage({"foo": "bar",
+                                                                 "other": "value"})]
+        self.master.namedServices = FakeServiceManager({"foo": "bar",
```
tardyp: Why not use the real service manager? I think that a bit too much fakes here, makes me nervous.
Normally fakes are tested against real implementation to make sure they implement the same interface.

You need to do fakes when using real will make it too costly in term of runtime or test setup complexity.
I think this is not the case here.

master/buildbot/process/properties.py
```diff
@@ -645,8 +678,9 @@ def _parse(self, fmtstring):
                     config.error(
                         "invalid Interpolate default type '%s'" % repl[0])
 
-    def getRenderingFor(self, props):
-        props = props.getProperties()
+    def getRenderingFor(self, build):
+        props = build.getProperties()
+        self.build = build
```
tardyp: self.build=build? why? no!

master/buildbot/process/properties.py
```diff
@@ -421,6 +421,32 @@ def getRenderingFor(self, build):
...
 
 @implementer(IRenderable)
+class _SecretRenderer(object):
+
+    def __init__(self, secret_name):
+        self.secret_name = secret_name
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, build):
+        servicenames = [service.name for service in build.getBuild().master.services]
```
tardyp: does not look right. why get does not work?

try catch is better solution than all the time computing that list of names

master/buildbot/process/properties.py
```diff
@@ -421,6 +420,33 @@ def getRenderingFor(self, build):
...
+class _SecretRenderer(object):
+
+    def __init__(self, secret_name):
+        self.secret_name = secret_name
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, build):
+        secretsSrv = build.getBuild().master.namedServices.get("secrets")
+        if not secretsSrv:
+            raise KeyError("secrets service not found")
```
tardyp: need to configure SecretManager in c['services'] to use 'secrets' in Interpolate

master/buildbot/process/properties.py
```diff
@@ -421,6 +420,33 @@ def getRenderingFor(self, build):
...
+        credsservice = build.getBuild().master.namedServices['secrets']
+        secret_detail = yield credsservice.get(self.secret_name)
+        if secret_detail is None:
+            raise KeyError("secret value for key %s is None" % self.secret_name)
```
tardyp: "secret key %s is not found in any provider"

master/buildbot/process/properties.py
```diff
@@ -382,9 +382,8 @@ def __repr__(self):
 
     @defer.inlineCallbacks
     def getRenderingFor(self, build):
-        value = build.render(self.value)
-        index = build.render(self.index)
-        value, index = yield defer.gatherResults([value, index])
+        value = yield build.render(self.value)
```
tardyp: is it really necessary

PR #2818, opened 2017-02-24
last small py3 fixes for running the e2e smoke tests on py3

tardyp:
@rodrigc congrats! great job!

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -79,6 +79,8 @@ matrix:
       env: TWISTED=trunk SQLALCHEMY=latest TESTS=py3
     - python: "3.5"
       env: TWISTED=trunk SQLALCHEMY=latest TESTS=flake8
+    - python: "3.5"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=smokes
 
 # Dependencies installation commands
 install:
--- a/common/smokedist.sh
+++ b/common/smokedist.sh
@@ -6,7 +6,11 @@ for suffix in whl tar.gz
 do
     VE=sandbox.$suffix
     rm -rf $VE
-    virtualenv $VE
+    if [ -z "$python" ]; then
+        virtualenv $VE
+    else
+        virtualenv --python python$python $VE
+    fi
     . $VE/bin/activate
     pip install -U pip
     pip install  mock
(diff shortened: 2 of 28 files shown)


rodrigc:
Wow this is a big milestone for buildbot!!

rodrigc:
According to https://nine.buildbot.net/#/builders/2/builds/112533
there is a failure in buildbot_pkg.py on Python 3.
Where is this code and where should the fix go?

seankelly:
`buildbot_pkg` is in the [pkg directory](https://github.com/buildbot/buildbot/blob/master/pkg/buildbot_pkg.py).

seankelly:
So I think the [`check_output`](https://github.com/buildbot/buildbot/blob/c6fe4587a0bfa417a2ccbb4f22f729f877d37d83/pkg/buildbot_pkg.py#L42) function needs to return a string instead of bytes. I bet the entire file needs to be checked for the bytes/string split.

rodrigc:
subprocess.check_output () returns bytes on Python 2 and 3.

If other code needs str, you need to do something like what I did here: http://bit.ly/2IOuXd5

rodrigc:
According to http://docs.buildbot.net/current/manual/installation/requirements.html#buildmaster-requirements

Python 2.6 is the minimum Python version required for the master and worker.  Is that still true?  Right now all the CI for the master is on 2.7 and 3.5.  The worker has CI for 2.6, 2.7, and 3.5

seankelly:
It's 2.7 for master. The `setup.py` only lists 2.7 too. The worker supports 2.6+.

seankelly:
It's not `subprocess.check_output`, but a `check_output` function that wraps `Popen.communicate`. Even if it returns bytes on both 2 and 3, does 2 care about using `re` on those bytes? I'm guessing no.

rodrigc:
Well, I added the section in master's `setup.py` in https://github.com/buildbot/buildbot/commit/5d19f32c4ca12739ab0bff6b80cff56c790ced6c , purely based on what versions were being run in the CI system.
If this is official project policy then I would suggest that the docs be updated to reflect this.
I can submit a separate patch for that.

seankelly:
`setup.py` used to mention 2.6 in the code for dependencies but it was removed not that long ago. The docs are simply out of date.

rodrigc:
@tardyp I have a few small fixes for your branch: https://github.com/tardyp/buildbot/pull/13

tardyp:
Eventually that wasn't that much of a small patch, but the smoke tests are now working, and we are able to generate universal binaries.

PR #2819, opened 2017-02-25
Python 2.7 is required on the master.

rodrigc:


rodrigc:
@seankelly: This commit https://github.com/buildbot/buildbot/commit/c6c60600837925e0a887c2a1d3f647add4d2c628
removed Python 2.6 checks from setup.py in the master, but the docs weren't updated at the same time.

tardyp:
gaah. I think there might be a dozen places with requirements in the doc (tutorial, installation, requirements)

PR #2820, opened 2017-02-25
Move lint to bbtravis

rodrigc:
Move lint to bbtravis.
The travis builds are getting longer and longer, so shuffle things around
to try to optimize things.

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -97,9 +97,8 @@ install:
                   mysqlclient \
                   psycopg2 \
                   pg8000 \
-                  codecov
-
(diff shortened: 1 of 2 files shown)


rodrigc:
@tardyp is there a way you can configure certain bbtravis builds so that they are **Required**?

- bb/flake8/2.7/tw:latest/sqla:latest/db:sqlite:// 
- bb/flake8/3.5/tw:latest/sqla:latest/db:sqlite://
- bb/isort/2.7/tw:latest/sqla:latest/db:sqlite://
- bb/lint/2.7/tw:latest/sqla:latest/db:sqlite://

tardyp:
@rodrigc I did require the global "bb" builder, so that every subbuilds are actually required.
This avoids to have to change the github configuration everytime we change the bbtravis.

PR #2821, opened 2017-02-25
py3 is now experimentally supported!

tardyp:


rodrigc:
Wow, wow, wow!!!

seankelly:
If 3.5 works, I bet 3.6 would too.

tardyp:
@seankelly it does not. There are a couple of tests that fail on py3.6

rodrigc:
I think I have fixed most of the problems that were seen on Py3.6.  Let's see what happens here:
https://github.com/buildbot/buildbot/pull/2824

rodrigc:
I think that this should be labelled "experimental" support on Python 3 rather than official.

Right now we need to check Twisted out from git in order for all the buildbot tests to pass
on Python 3.  I don't think this is acceptable for end-users.

rodrigc:
I'm working on finishing off this item in Twisted, so that sending e-mail from buildbot will work in Python 3:
https://github.com/twisted/twisted/pull/509

tardyp:
@rodrigc fo you have in minde a particular way to mark py3 support experimental?

Note that the smoke tests script is re-creating a virtualenv, so we are testing against twisted latest, and not twisted trunk.

But indeed, we should start pushing @hawkowl @glyph for a 17.0.0 twisted release.

rodrigc:
I would say:

1.  Change the title of this pull request to "py3 is now experimentally supported",
     so that people don't get the wrong idea when they look at tickets and search engines.
2.  Hold off on merging this until we get Twisted folks to release a version to pypi
     that addresses all the issues that buildbot has on Python 3.

I have merged the Py3 fixes for twisted.mail upstream in Twisted.
These fixes are now needed for buildbot: https://github.com/buildbot/buildbot/pull/2828

tardyp:
Agreed. Lets wait until twisted 17.

hawkowl:
@tardyp 17.1 is already released :) But a 17.x with these changes will be sometime next month, I imagine.

rodrigc:
I think if we can get this Twisted fix in https://github.com/twisted/twisted/pull/722
that will address all the major Twisted roadblocks for running buildbot on Python 3.

tardyp:
@hawkowl I did not realize that :)  This explains why the smoke test with twisted latest worked, while @rodrigc said 2 month ago that it wont.

I think we can experimentally release with 17.1 with disclaimer that it only works py3 master <-> py3 worker

rodrigc:
@tardyp I would hold off on releasing with 17.1, because e-mail won't work until you get https://github.com/buildbot/buildbot/pull/2828
Lack of e-mail notifications is a big hole to have in a CI system.

tardyp:
I am not sure where how far we are from twisted 17.2.0.
I think it will be good enough if we say its experimental. At least people can start making comparaison (is py3 faster, does it consume less memory, etc)

I will be releasing next week, as per our one monthrelease cadence and we should probably make py3 a big highlight of that release.

PR #2823, opened 2017-02-25
Fix test_trigger on Python 3

rodrigc:


.py3.notworking.txt
```diff
@@ -1 +0,0 @@
-buildbot.test.integration.test_trigger.TriggeringMaster.test_trigger
```
tardyp: oh. I thought that file was empty. thanks for finishing it.

PR #2828, opened 2017-02-27
Fix mail tests on Python 3

rodrigc:
These fixes are needed, since I merged this upstream in Twisted:

https://github.com/twisted/twisted/pull/509


tardyp:
thanks for quick reaction, I started to freak out because of the other PR failures!

PR #2831, opened 2017-02-28
Fix some rst formatting

rutsky:


master/docs/manual/cfg-changesources.rst
```diff
@@ -300,14 +300,14 @@ cd to the CVSROOT directory and edit the file loginfo, adding a line like:
 
 The above example you put the :src:`buildbot_cvs_mail.py <master/contrib/buildbot_cvs_mail.py>` script under /cvsroot/CVSROOT.
 It can be anywhere.
-Run the script with --help to see all the options.
+Run the script with ``--help`` to see all the options.
```
tardyp: I guess --help must be in the spelling_words.txt

rutsky: Good point! Looks like Sphinx spell checker trims characters like `-` from words, so it considers `--help` as `help` --- there is no `--help` in [spelling_wordlist.txt](https://github.com/buildbot/buildbot/blob/master/master/docs/spelling_wordlist.txt)

PR #2835, opened 2017-03-04
Adding Vault provider

protatremy:
Adding the provider to use Vault backend with secrets management.

## Contributor Checklist:


master/buildbot/secrets/providers/vault.py
```diff
@@ -0,0 +1,57 @@
...
+from __future__ import absolute_import
+from __future__ import print_function
+
+import hvac
+
+from buildbot import config
+from buildbot.secrets.providers.base import SecretProviderBase
+
+
+class SecretInVault(SecretProviderBase):
```
tardyp: as Vault is a generic name, SecretInVault is not a very good name.

HashiCorpVaultSecretProvider is better name imho

master/buildbot/secrets/providers/vault.py
```diff
@@ -0,0 +1,57 @@
...
+    name = 'SecretInVault'
+
+    def checkConfig(self, vaultServer=None, vaultToken=None):
+        if not isinstance(vaultServer, str):
+            config.error("vaultServer must be a string while it is %s" % (type(vaultServer,)))
+        if not isinstance(vaultToken, str):
+            config.error("vaultToken must be a string while it is %s" % (type(vaultToken,)))
+        self.vaultServer = vaultServer
+        self.token = vaultToken
+        self.client = hvac.Client(url=vaultServer, token=vaultToken)
```
tardyp: here you should do config.error if hvac could not be imported

master/buildbot/secrets/providers/vault.py
```diff
@@ -0,0 +1,57 @@
...
+    def reconfigService(self, vaultServer=None, vaultToken=None):
+        self.vaultServer = vaultServer
+        self.token = vaultToken
+        self.client = hvac.Client(url=vaultServer, token=vaultToken)
+
+    def get(self, entry):
+        """
+        get the value from vault secret backend
+        """
+        secret = self.client.read('secret/' + entry)
```
tardyp: 'secret/' why is it hardcoded? is it something like a convention with hc vault?

we might want to put it as a class constant

master/buildbot/secrets/providers/vault.py
```diff
@@ -0,0 +1,57 @@
...
+        if 'data' not in secret.keys():
```
tardyp: you can replace the whole thing by:

    secret.get('data', {}).get('value')

master/buildbot/test/unit/test_secret_in_vault.py
```diff
@@ -0,0 +1,86 @@
...
+import mock
+
+from twisted.internet import defer
+from twisted.trial import unittest
+
+from buildbot.secrets.providers.vault import SecretInVault
+from buildbot.test.util.config import ConfigErrorsMixin
+
+
+class FakeVaultClient(object):
```
tardyp: looks like it is not used actually

master/setup.py
```diff
@@ -444,7 +444,8 @@ def define_plugin_entries(groups):
     'txaio ' + txaio_ver,
     'autobahn ' + autobahn_ver,
     'PyJWT',
-    'distro'
+    'distro',
```
tardyp: we can't make hvac an hard dependency to buildbot. We have to keep it an optionnal dependency.

master/buildbot/secrets/providers/vault.py
tardyp: dont create a client in the checkConfig, it will be created in the reconfigService

master/buildbot/secrets/providers/vault.py
tardyp: also self.client.read is using requests as a backend so this is a blocking call.

Looking at the hvac implementation, I think it will be much easier to not use hvac, and rather use the REST api as our main hashicorp vault API.
We are using only on api (read, with a token)
So this is reallly a very basic http call that will be much better handled (and unit tested) with httpclientservice.

master/buildbot/secrets/providers/vault.py
protatremy: hvac is now removed

master/buildbot/secrets/providers/vault.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# Copyright Buildbot Team Members
+"""
+vault based providers
+"""
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import hvac
```
protatremy: removed as requested after

master/buildbot/secrets/providers/vault.py
protatremy: removed

master/buildbot/secrets/providers/vault.py
```diff
@@ -0,0 +1,66 @@
...
+    @defer.inlineCallbacks
+    def get(self, entry):
+        """
+        get the value from vault secret backend
+        """
+        path = self.secretsmount + '/' + entry
+        proj = yield self._http.get('/v1/{0}'.format(path))
+        code = yield proj.code
+        if code != 200:
+            config.error("%s unable to get key %s" % (code, entry))
```
tardyp: config.error do not work at runtime. Please raise KeyError instead. Please make a better message making it clear that the key does not exist in Vault

master/buildbot/test/integration/test_integration_secrets_with_vault.py
```diff
@@ -0,0 +1,72 @@
...
+
+from buildbot.process.properties import Interpolate
+from buildbot.secrets.providers.vault import HashiCorpVaultSecretProvider
+from buildbot.steps.shell import ShellCommand
+from buildbot.test.util.integration import RunMasterBase
+
+
+# This integration test creates a master and worker environment,
+# with one builders and a shellcommand step
+# meant to be a template for integration steps
```
tardyp: Please write recipe here on how to bring up a vault server to run that test
- which docker image?
- what is the docker run command?
- how to provision the token?
- how to provision the needed secret?

master/buildbot/test/integration/test_integration_secrets_with_vault.py
```diff
@@ -0,0 +1,72 @@
...
+                "Vault integration tests only run when environment variable "
+                "BBTEST_VAULTURL is set with url to Vault api server and "
+                "variable BBTEST_VAULTTOKEN is set with a valid token to Vault "
+                "server")
+
+    @defer.inlineCallbacks
+    def test_secret(self):
+        yield self.setupConfig(masterConfig())
+        build = yield self.doForceBuild(wantSteps=True, wantLogs=True)
+        self.assertEqual(build['buildid'], 1)
```
tardyp: please check that the echo is the right one

master/buildbot/test/unit/test_secret_in_vault.py
```diff
@@ -0,0 +1,93 @@
...
+        self._http.expect(method='get', ep='/v1/secret/valueNotFound', params=None,
+         data=None, json=None, code=200,
+         content_json={"data": {"valueNotFound": "value1"}})
+        value = yield self.srvcVault.get("valueNotFound")
+        self.assertEqual(value, None)
+
+    @defer.inlineCallbacks
+    def testGetError(self):
+        self._http.expect(method='get', ep='/v1/secret/valueNotFound', params=None,
+         data=None, json=None, code=201,
```
tardyp: 201 is not an error 404 is.

master/buildbot/test/unit/test_secret_in_vault.py
```diff
@@ -0,0 +1,93 @@
...
+        value = yield self.srvcVault.get("value")
+        self.assertEqual(value, "value1")
+
+    @defer.inlineCallbacks
+    def testGetValueNotFound(self):
+        self._http.expect(method='get', ep='/v1/secret/valueNotFound', params=None,
+         data=None, json=None, code=200,
+         content_json={"data": {"valueNotFound": "value1"}})
+        value = yield self.srvcVault.get("valueNotFound")
+        self.assertEqual(value, None)
```
tardyp: here the expect is doing things weird. It returns 200 code, with data provided for "ValueNotFound".
So I would expect srvcVault.get() should return 'value1'. Not sure how this can work

master/buildbot/test/integration/test_integration_secrets_with_vault.py
protatremy: I would prefer to give a link to the documentation, but any way, I had the recipe in the comments

master/buildbot/test/integration/test_integration_secrets_with_vault.py
```diff
@@ -0,0 +1,93 @@
...
+from buildbot.secrets.providers.vault import HashiCorpVaultSecretProvider
+from buildbot.steps.shell import ShellCommand
+from buildbot.test.util.integration import RunMasterBase
+
+
+# This integration test creates a master and worker environment,
+# with one builders and a shellcommand step
+# To be run, the test use a Vault server instance.
+# To easily run a Vault instance find the dockerfile dedictaed to Vault.
+# command: docker-compose up
```
tardyp: where is the compose file associated with this?

master/buildbot/test/integration/test_integration_secrets_with_vault.py
```diff
@@ -0,0 +1,93 @@
...
+# Note the Vault docker image address and port.
+# connect to the docker instance:
+# command: docker exec -i -t ``docker_vault_image_name`` /bin/sh
```
tardyp: which image name? how to build it. This would be best to have one that is already exists on docker hub

master/buildbot/test/integration/test_integration_secrets_with_vault.py
tardyp: I have find with the documentation. Indeed it makes more sense to have a vault quick setup in the doc, and this test pointing to it + adding the specific stuff to do so that the test is happy.

master/buildbot/test/integration/test_integration_secrets_with_vault.py
```diff
@@ -0,0 +1,87 @@
...
+# By default Vault stores password in the "secret" mount. Add a password:
+# command: vault write secret/key value=word
+# In the Buildbot environment, export BBTEST_VAULTURL and BBTEST_VAULTTOKEN
+# command: export BBTEST_VAULTURL='http://vaultServeraddress:port'
+# command: export BBTEST_VAULTTOKEN='vault token'
+
+class SecretsConfig(RunMasterBase):
+
+    def setUp(self):
+        if "BBTEST_VAULTURL" not in os.environ and "BBTEST_VAULTTOKEN" not in os.environ:
```
tardyp: os.system("docker run --name vault_for_buildbot -d --privileged -p 8200:8200 vault")
    token = subpprocess.check_output("docker logs vault_for_buildbot 2>/dev/null |grep 'Root Token' |awk '{print $3;}'")
    os.system("docker exec vault_for_buildbot /bin/sh -c \"export VAULT_ADDR='http://localhost:8200'; vault write secret/key value=word\"")

tardyp:
Lets make another PR for the doc updates, and then we can release 0.9.5!

PR #2838, opened 2017-03-06
Fix traceback in buildbot/db/logs.py

dol-sen:
This traceback was triggered by a custom LogLineObserver class traceback.
This is likely a corner case.  But none the less, this code change should
prevent further tracebacks and the log system from also failing.

2017-03-06 21:36:39+0000 [-] Got fatal Exception on DB
        Failure: exceptions.TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'

2017-03-06 21:36:39+0000 [-] while compressing log 1456 (ignored)
          File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 434, in errback
            self._startRunCallbacks(fail)
          File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 501, in _startRunCallbacks
            self._runCallbacks()
          File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
          File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 1184, in gotResult
          File "/usr/lib64/python2.7/site-packages/twisted/internet/defer.py", line 1126, in _inlineCallbacks
          File "/usr/lib64/python2.7/site-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
          File "/usr/lib64/python2.7/site-packages/buildbot/db/logs.py", line 334, in compressLog
            saved = yield self.db.pool.do(thdcompressLog)
          File "/usr/lib64/python2.7/site-packages/twisted/python/threadpool.py", line 246, in inContext
            result = inContext.theWork()
          File "/usr/lib64/python2.7/site-packages/twisted/python/threadpool.py", line 262, in <lambda>
          File "/usr/lib64/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
          File "/usr/lib64/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
            return func(*args,**kw)
          File "/usr/lib64/python2.7/site-packages/buildbot/db/pool.py", line 182, in __thd
            rv = callable(arg, *args, **kwargs)
          File "/usr/lib64/python2.7/site-packages/buildbot/db/logs.py", line 332, in thdcompressLog
            return totlength - newsize
        exceptions.TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'

## Contributor Checklist:

* [ *] I have updated the unit tests
* [ *] I have created a file in the master/buildbot/newsfragment directory (and read the README.txt in that directory)
* [*] I have updated the appropriate documentation

NOTE:  I didn't have enough info about what fetchone() returns to know if a specific check needs to be done to detect a Nonetype condition or not.  This was a simple solution.  Please regard this as a starting point.  I can update it to do a specific check if you prefer.

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -328,7 +328,8 @@ class LogsConnectorComponent(base.DBConnectorComponent):
             # calculate how many bytes we saved
             q = sa.select([sa.func.sum(sa.func.length(tbl.c.content))])
             q = q.where(tbl.c.logid == logid)
-            newsize = conn.execute(q).fetchone()[0]
+            # if fetchone()[0] is None, then set it to 0  # pylint:disable=wrong-spelling-in-comment
+            newsize = conn.execute(q).fetchone()[0] or 0
             return totlength - newsize
 
         saved = yield self.db.pool.do(thdcompressLog)


master/buildbot/db/logs.py
```diff
@@ -328,7 +328,8 @@ def thdcompressLog(conn):
             # calculate how many bytes we saved
             q = sa.select([sa.func.sum(sa.func.length(tbl.c.content))])
             q = q.where(tbl.c.logid == logid)
-            newsize = conn.execute(q).fetchone()[0]
+            # if fetchone()[0] is None, then set it to 0
+            newsize = conn.execute(q).fetchone()[0] or 0
```
tardyp: I can't find a reason why fetchone()[0] will return None.

I would prefer to fix the rootcause rather than hiding the problem under the carpet.
Do you have any more context of what is the log that is being compressed? is it empty or anyhow special?

dol-sen: Normally, there won't be a reason for it to return None.

I have fixed the reason the code puked.  This change still makes the code more fault tolerant, but does not hide the real cause.  It just prevents a secondary crash which could be misleading in determining the cause.

In our case for buildbot-8 we had to subclass the LogLineObserver to relay important status and other information about the step since the output was not available until completion of the step.
For buildbot-9 we get stdout near realtime, so they were not needed.  I removed all those (but missed one), except the one that retrieved a url link  to a custom test system status.  Since the RealTimeLogLineObserver class had been simplified, the one use case I missed caused it to traceback.  That traceback triggered the fetchone()[0] to return None, which caused the log system to also crash.

This change just makes the logging code more robust.  It will not affect normal operation.  It will only prevent the log system from crashing due to some other crash like the one which was caused by my missed removal of the RealtimeLogLineObserver use in the build script.

dol-sen:
the test failures are because pylint is stupid :/  complaining about my use of fetchone() in a comment.  That it is misspelled.  When it technically is not misspelled.

dol-sen:
closed in favor of https://github.com/buildbot/buildbot/pull/3020

PR #2839, opened 2017-03-07
Fix GNUAutoconf factory.

wainersm:
The GNUAutoconf factory execute by default a sequence of compile,
test, and distcheck, but one is supposed to disable any of them by
setting None. Currently it doesn't work though.

This change fix the issue so that these steps may be disabled.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/factory.py
+++ b/master/buildbot/process/factory.py
@@ -86,24 +86,29 @@ class BuildFactory(util.ComparableMixin):
 # BuildFactory subclasses for common build tools
 
 
+class _DefaultCommand(object):
+    # Used to indicate a default command to the step.
+    pass
+
+
 class GNUAutoconf(BuildFactory):
 
     def __init__(self, source, configure="./configure",
                  configureEnv=None,
                  configureFlags=None,
                  reconf=None,
-                 compile=None,
-                 test=None,
-                 distcheck=None):
+                 compile=_DefaultCommand,
+                 test=_DefaultCommand,
+                 distcheck=_DefaultCommand):
         if configureEnv is None:
             configureEnv = {}
         if configureFlags is None:
             configureFlags = []
-        if compile is None:
+        if compile is _DefaultCommand:
             compile = ["make", "all"]
-        if test is None:
+        if test is _DefaultCommand:
             test = ["make", "check"]
-        if distcheck is None:
+        if distcheck is _DefaultCommand:
             distcheck = ["make", "distcheck"]
 
         BuildFactory.__init__(self, [source])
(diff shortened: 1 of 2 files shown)


wainersm:
Hold on this review, I am preparing a test case to integrate into the commit.

wainersm:
Updated the commit. Now it contains tests for the fix.

seankelly:
What is this fixing?

master/buildbot/process/factory.py
```diff
@@ -92,19 +92,13 @@ def __init__(self, source, configure="./configure",
                  configureEnv=None,
                  configureFlags=None,
                  reconf=None,
-                 compile=None,
-                 test=None,
-                 distcheck=None):
+                 compile=["make", "all"],
```
seankelly: Not sure about setting the steps with list values because of how Python handles it. Since the docs are already wrong, I think it would be better to use `False` to disable them.

wainersm: I'm not changing the semantics of compile, distcheck and test steps: currently they already receive a list of values. The fix consists in initialize the variables or skip the step if None is set, currently setting to None will keep the step (and use default command).

The manual says None should be set to skip the step. Other factories (e.g. Distutils and Trial) uses same semantic, i.e., set None to disable a step. Also IMHO setting False is more suitable if the variable expects a boolean value.

tardyp: I agree with seankelly.
Its a python felony to initialize method argument with lists.
The correct way is to use a marker object
```python
class Marker(object):
    pass

def my_method(compile=Marker):
     if compile is Marker:
          compile=["make", "all"]
```

tardyp: I agree that this is bad coding style to initialize method argument with mutable objects.
http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

in your case, None has a meaning, so you cannot use it as a default Marker, so usually what you do is:

```
class defaultMarker(object):
    pass

[...]
def __init__(self, compile=defaultMarker):
    if compile is defaultMarker:
        compile = ['make', 'all']

```

wainersm:
@seankelly , @tardyp I've update the commit with changes you suggested. I hope it's ok now. Thanks for the review.

master/buildbot/process/factory.py
```diff
@@ -86,24 +86,29 @@ def addSteps(self, steps):
 # BuildFactory subclasses for common build tools
 
 
+class _DefaultCommand(object):
+        # Used to indicate a default command to the step.
```
seankelly: Make these two lines indented by four spaces instead of eight.

wainersm: done.

PR #2840, opened 2017-03-08
Fix latest novaclient

seankelly:
## Contributor Checklist:

### DIFF
--- a/master/docs/manual/cfg-workers-openstack.rst
+++ b/master/docs/manual/cfg-workers-openstack.rst
@@ -106,12 +106,15 @@ These are the same details set in either environment variables or passed as opti
 ``nova_args``
     (optional)
     A dict that will be appended to the arguments when creating a VM.
-    Buildbot uses the OpenStack Nova version 1.1 API by default (see client_version).
+    Buildbot uses the OpenStack Nova version 2 API by default (see client_version).
 
 ``client_version``
     (optional)
-    Nova client version to use. Defaults to 1.1 (deprecated). Use 2 or 2.minor for
-    version 2 API.
+    A string containing the Nova client version to use.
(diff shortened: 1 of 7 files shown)


seankelly:
Found an issue elsewhere when booting an instance, so don't merge this yet.

master/docs/manual/cfg-workers-openstack.rst
```diff
@@ -106,12 +106,14 @@ These are the same details set in either environment variables or passed as opti
 ``nova_args``
     (optional)
     A dict that will be appended to the arguments when creating a VM.
-    Buildbot uses the OpenStack Nova version 1.1 API by default (see client_version).
+    Buildbot uses the OpenStack Nova version 2 API by default (see client_version).
```
tardyp: Do we still support version 1 after this?

seankelly: At the moment, yes. It depends on the version of novaclient.

For a while it has automatically converted to using v2 of the API instead of v1.1. If you need to use the older API version, you still can by changing `client_version` combined with an older version of novaclient.

tardyp: I would add this answer in the documentation(and release note). Not sure how deprecated is 1.1

seankelly: Starting with novaclient 2.27.0 (released in April 2015), it will automatically use version 2 if you ask to use version 1.1. This message is printed out when creating the client:

    Version 1.1 is deprecated, using alternative version 2 instead.

I noted in the docs that 1.1 is deprecated, so I can add that you need a version older than 2.27.0 to actually use 1.1.

PR #2841, opened 2017-03-08
db/logs.py: Fix exception of compressing empty logs

pmisik:
We have running BuildBot with custom msbuild steps that logs warnings and errors into separate logfiles.
In case build finishes and log contains no line we see exception stack
trace below.
At the moment I don't have depth knowledge of BuildBot. It seems following patch could fix the problem.

[-] while compressing log 10 (ignored)
	  File "c:\bb\bb-master\lib\site-packages\twisted\internet\defer.py", line 500, in errback
	    self._startRunCallbacks(fail)
	  File "c:\bb\bb-master\lib\site-packages\twisted\internet\defer.py", line 567, in _startRunCallbacks
	    self._runCallbacks()
	  File "c:\bb\bb-master\lib\site-packages\twisted\internet\defer.py", line 653, in _runCallbacks
	  File "c:\bb\bb-master\lib\site-packages\twisted\internet\defer.py", line 1357, in gotResult
	  File "c:\bb\bb-master\lib\site-packages\twisted\internet\defer.py", line 1299, in _inlineCallbacks
	  File "c:\bb\bb-master\lib\site-packages\twisted\python\failure.py", line 393, in throwExceptionIntoGenerator
	  File "c:\bb\bb-master\lib\site-packages\buildbot\db\logs.py", line 334, in compressLog
	    saved = yield self.db.pool.do(thdcompressLog)
	  File "c:\bb\bb-master\lib\site-packages\twisted\python\threadpool.py", line 250, in inContext
	    result = inContext.theWork()
	  File "c:\bb\bb-master\lib\site-packages\twisted\python\threadpool.py", line 266, in <lambda>
	  File "c:\bb\bb-master\lib\site-packages\twisted\python\context.py", line 122, in callWithContext
	  File "c:\bb\bb-master\lib\site-packages\twisted\python\context.py", line 85, in callWithContext
	    return func(*args,**kw)
	  File "c:\bb\bb-master\lib\site-packages\buildbot\db\pool.py", line 182, in __thd
	    rv = callable(arg, *args, **kwargs)
	  File "c:\bb\bb-master\lib\site-packages\buildbot\db\logs.py", line 332, in thdcompressLog
	    return totlength - newsize
	exceptions.TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'

## Remove this paragraph

And especially:

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -329,6 +329,8 @@ class LogsConnectorComponent(base.DBConnectorComponent):
             q = sa.select([sa.func.sum(sa.func.length(tbl.c.content))])
             q = q.where(tbl.c.logid == logid)
             newsize = conn.execute(q).fetchone()[0]
+            if newsize is None:
+                newsize = totlength
             return totlength - newsize
 
         saved = yield self.db.pool.do(thdcompressLog)


dol-sen:
This is a duplicate of my pull request from the other day.

https://github.com/buildbot/buildbot/pull/2838

I just used an or which is a little simpler code

seankelly:
Very related: #2838.

master/buildbot/db/logs.py
```diff
@@ -329,6 +329,8 @@ def thdcompressLog(conn):
             q = sa.select([sa.func.sum(sa.func.length(tbl.c.content))])
             q = q.where(tbl.c.logid == logid)
             newsize = conn.execute(q).fetchone()[0]
+            if newsize is None:
+                newsize = totlength
```
dol-sen: On my patch I just used 0

newsize = conn.execute(q).fetchone()[0] or 0

but it could just as easily be:

newsize = conn.execute(q).fetchone()[0] or totlength

I don't know which would be more correct.

dol-sen: I too don't know enough of the possible values that code can generate...

>>> None or 5
5
>>> 0 or 5
5
>>> 0 or -5
-5
>>> None or -5
-5

tardyp:
@pmisik @dol-sen Thanks for the patches.
The thing is a patch is worthless if it is not baked by a strong rootcause analysis. This is why I rejected dol-sen patch. This was very unclear to me what is the reason of the problem.

I looks now more obvious that the problem is with log compression not working with empty logs.
So we need a unit test to back this theory.
Then, if this is the case, I would rather fix it by adding, line 288:

    if totLength == 0:
       return 0

This would skip the last problematic request.

Do you guys want to do that patch and unit test and get the credits, or you need me to do the fix?

pmisik:
@tardyp  
I have created new pull request #3020 including test

PR #3020, opened 2017-03-09
Fix exception in case of compressing empty logs

pmisik:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_db_logs.py
+++ b/master/buildbot/test/unit/test_db_logs.py
@@ -523,6 +523,7 @@ class RealTests(Tests):
         logdict = yield self.db.logs.getLog(201)
         self.assertEqual(logdict, None)
 
+
 class TestFakeDB(unittest.TestCase, Tests):
 
     def setUp(self):
--- a/master/buildbot/test/unit/test_db_logs.py
+++ b/master/buildbot/test/unit/test_db_logs.py
@@ -518,10 +518,27 @@ class RealTests(Tests):
         return self.do_addLogLines_huge_log(NUM_CHUNKS=3000, chunk=u'\N{SNOWMAN}\n' * 50)
 
     @defer.inlineCallbacks
-    def test_compressEmptyLog(self):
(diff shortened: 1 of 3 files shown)


master/buildbot/test/unit/test_db_logs.py
```diff
@@ -517,6 +517,11 @@ def test_addLogLines_huge_log_lots_line(self):
     def test_addLogLines_huge_log_lots_snowmans(self):
         return self.do_addLogLines_huge_log(NUM_CHUNKS=3000, chunk=u'\N{SNOWMAN}\n' * 50)
 
+    @defer.inlineCallbacks
+    def test_compressEmptyLog(self):
+        yield self.db.logs.compressLog(201)
+        logdict = yield self.db.logs.getLog(201)
+        self.assertEqual(logdict, None)
```
seankelly: Need to add one more blank line after the new method so there are two blank lines between the method and class.

master/buildbot/test/unit/test_db_logs.py
```diff
@@ -517,6 +517,12 @@ def test_addLogLines_huge_log_lots_line(self):
     def test_addLogLines_huge_log_lots_snowmans(self):
         return self.do_addLogLines_huge_log(NUM_CHUNKS=3000, chunk=u'\N{SNOWMAN}\n' * 50)
 
+    @defer.inlineCallbacks
+    def test_compressEmptyLog(self):
```
tardyp: This test tests the not existing log use case (which is nice to test too, but not coherent with the test name)

In order to test the EmptyLog use case, you need to insert the following in the db:

```python
        yield self.insertTestData(self.backgroundData + [
            fakedb.Log(id=201, stepid=101, name=u'stdio', slug=u'stdio',
                       complete=1, num_lines=0, type=u's'),
        ])
```

pmisik:
Thank you for review. 
Original test is renamed and added requested one.

tardyp:
great! @pmisik congrats for your first github contribution!

PR #3039, opened 2017-03-13
Process changes from git tag push events

dragomirtodorov:
Hello,
since GitLab/GitHUB has webhooks for git tag push events and now in master branch we have in the GitPoller polling from tags i believe it is reasonable to process this events ( tag push ) and in webhooks.
## Contributor Checklist:


tardyp:
I think this is fairly reasonable. I wonder why this wasn't already there.

As always, we need tests, documentation update and release notes

dragomirtodorov:
should be okay now?

PR #3042, opened 2017-03-15
docs: fix port numbers in docker tutorial

noc0lour:
Either we fix the portnumbers here in the docs or we fix the example config in buildbot/buildbot-docker-example-config. 
Since the config is using `8080` and the tutorial tells to use `8010`

### DIFF
--- a/master/docs/tutorial/docker.rst
+++ b/master/docs/tutorial/docker.rst
@@ -64,12 +64,12 @@ Building and running Buildbot
   docker-compose up
 
 
-You should now be able to go to http://localhost:8010 and see a web page similar to:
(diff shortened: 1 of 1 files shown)


master/docs/tutorial/docker.rst
```diff
@@ -64,12 +64,12 @@ Building and running Buildbot
   docker-compose up
 
 
-You should now be able to go to http://localhost:8010 and see a web page similar to:
```
tardyp: We should rather fix the docker config to use 8010, as it is our real default port

noc0lour:
I guess I can close this PR and reopen in the right repo ;)

PR #3049, opened 2017-03-19
Remove note for Python 3 virtualenv

rodrigc:
I think it is safe to remove this now.


rodrigc:
@rutsky I added a small not to use *-m venv" on Python 3, which works better there.

PR #3054, opened 2017-03-20
fix the usePTY feature

tardyp:
some people have reported the usePTY feature to not work.
This is a unit test and integration test to make sure it actually works

test is only for posix

### DIFF
--- a/master/buildbot/steps/shell.py
+++ b/master/buildbot/steps/shell.py
@@ -248,7 +248,7 @@ class ShellCommand(buildstep.LoggingBuildStep):
 
         # check for the usePTY flag
         if 'usePTY' in kwargs and kwargs['usePTY'] is not None:
-            if self.workerVersionIsOlderThan("svn", "2.7"):
+            if self.workerVersionIsOlderThan("shell", "2.7"):
                 warnings.append(
                     "NOTE: worker does not allow master to override usePTY\n")
                 del kwargs['usePTY']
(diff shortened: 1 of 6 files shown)


master/buildbot/steps/shell.py
```diff
@@ -248,7 +248,7 @@ def buildCommandKwargs(self, warnings):
 
         # check for the usePTY flag
         if 'usePTY' in kwargs and kwargs['usePTY'] is not None:
-            if self.workerVersionIsOlderThan("svn", "2.7"):
+            if self.workerVersionIsOlderThan("shell", "2.7"):
```
tardyp: This shall not change anything as buildbot worker advertise versions for '*', just fix it for coherency

master/buildbot/test/integration/test_usePty.py
```diff
@@ -0,0 +1,71 @@
...
+from __future__ import print_function
+
+from twisted.internet import defer
+
+from buildbot.test.util.decorators import skipUnlessPlatformIs
+from buildbot.test.util.integration import RunMasterBase
+
+
+# This integration test creates a master and worker environment,
+# with one builders and a shellcommand step
```
rutsky: "with one builder~~s~~"

master/buildbot/steps/shell.py
tardyp: actually, reporter says that this change did matter.
because there is no "svn" command anymore in the worker package
I looked for other lines with the same problem, and there is none.

PR #3063, opened 2017-03-20
add top-level LICENSE duplicating {master,worker}/COPYING

djmitche:
Apparently some legal departments care not just which license is in use,
but where in the repository that use is declared.  As a nice
side-effect, this should get GitHub to properly display BB's license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,280 @@
+		    GNU GENERAL PUBLIC LICENSE
+		       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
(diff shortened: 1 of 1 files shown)


tardyp:
www/* is MIT. I'd fear that having this COPYING in the root directory misjudge people that the everything is GPL.

djmitche:
I had forgotten about that.  It's worth noting that there's no COPYING in
that directory, and most of the license headers I can find say GPLv2.

tardyp:
Right, I though I did put that under MIT, but looks like I did not. So that's fine to put COPYING there.

PR #3074, opened 2017-03-25
changes: add additional information to properties in GitHubPullrequestPoller

noc0lour:
I want to checkout if this would have chances to get included into the main tree once I'm done refining it. 
Basically I'd like to extract more information from the PR and add it to the properties which then can be used in the builders.

Yes, no? Or in a different way?

### DIFF
--- a/master/buildbot/changes/github.py
+++ b/master/buildbot/changes/github.py
@@ -105,6 +105,7 @@ class GitHubPullrequestPoller(base.ReconfigurablePollingChangeSource,
         self.branches = branches
         self.project = project
         self.pollInterval = pollInterval
+        self.pollAtLaunch = pollAtLaunch
         self.repository_type = link_urls[repository_type]
         self.magic_link = magic_link
 
@@ -219,7 +220,7 @@ class GitHubPullrequestPoller(base.ReconfigurablePollingChangeSource,
                         failures[0].raiseException()
                 [files, email] = [r[1] for r in results]
 
-                if email is not None or email is not "null":
+                if email is not None and email is not "null":
                     author += " <" + str(email) + ">"
 
                 # emit the change
--- a/master/buildbot/changes/github.py
+++ b/master/buildbot/changes/github.py
@@ -15,8 +15,10 @@
 
 from __future__ import absolute_import
 from __future__ import print_function
+from future.utils import iteritems
 
 from datetime import datetime
+from fnmatch import fnmatch
 
 from twisted.internet import defer
 
@@ -39,8 +41,24 @@ link_urls = {
 }
 
 
+class PullRequestMixin(object):
+    def extractProperties(self, pr_info):
+        def flatten(properties, base, info_dict):
+            for k, v in iteritems(info_dict):
+                name = ".".join([base, k])
+                if isinstance(v, dict):
+                    flatten(properties, name, v)
+                elif any([fnmatch(name, expr)
+                          for expr in self.github_property_whitelist]):
+                    properties[name] = v
+
+        properties = {}
+        flatten(properties, "github", pr_info)
+        return properties
+
+
 class GitHubPullrequestPoller(base.ReconfigurablePollingChangeSource,
-                              StateMixin):
+                              StateMixin, PullRequestMixin):
     compare_attrs = ("owner", "repo", "token", "branches", "pollInterval",
                      "category", "project", "pollAtLaunch", "name")
     db_class_name = 'GitHubPullrequestPoller'
@@ -62,6 +80,7 @@ class GitHubPullrequestPoller(base.ReconfigurablePollingChangeSource,
                     token=None,
                     magic_link=False,
                     repository_type="https",
+                    github_property_whitelist=None,
                     **kwargs):
         if repository_type not in ["https", "svn", "git", "ssh"]:
             config.error(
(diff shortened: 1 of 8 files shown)


noc0lour:
@tardyp this is more a "design" question I guess, if we allow additional special properties to propagate from different source steps.

tardyp:
You can look at the gerrit change source source code, it has similar code.
It does create one property per event attribute (recursively e.g
{
     author: {'email': 'foo@bar'}
}
will create a property 'gerrit_event.author.email'

noc0lour:
A cool, I'll have a look, probably I want to flatten them as well

tardyp:
indeed, this is a mistake that I made for gerrit design, we should have a whitelist configuration in the changesource (maybe fnmatch based).

noc0lour:
I'd like to have the parsing of PRs in the webhook the same as it is in this PR poller. (e.g. add these changes there as well :) ). Should I just copy & paste ?

seankelly:
Yes. The poller and webhook should stay in sync for what features they support.

master/buildbot/changes/github.py
```diff
@@ -42,9 +43,31 @@
...
+        "github.base.ref",
+        "github.base.sha",
+        "github.base.repo.full_name",
+        "github.base.private",
+        "github.base.fork",
+        "github.merged",
+        "github.mergeable",
+        "github.mergeable_state",
+        "github.merged_by",
+    ]
```
tardyp: looks like a lot of properties for the defaults.
This should be configurable

master/buildbot/changes/github.py
```diff
@@ -171,6 +198,24 @@ def _getStateObjectId(self):
             '%s/%s' % (self.owner, self.repo), self.db_class_name)
         defer.returnValue(result)
 
+    def _parseGitHubPR(self, pr_info):
+        def flatten(properties, base, info_dict):
+            for k, v in iteritems(info_dict):
+                name = ".".join([base, k])
+                if not any([
+                        prop.startswith(name)
+                        for prop in self.GH_PROPERTY_WHITELIST
```
tardyp: I would profer is you used fnmatch https://docs.python.org/2/library/fnmatch.html

master/buildbot/www/hooks/github.py
```diff
@@ -108,6 +130,23 @@ def _get_payload(self, request):
 
         return payload
 
+    def _get_PR_information(self, pr_info):
+        def flatten(properties, base, info_dict):
```
tardyp: I dont like to copy the exact same code in two different places. maybe we should have it in a changes/github.py mixin class

master/buildbot/changes/github.py
noc0lour: I looked into fnmatch but it didn't seem to provide a lot of flexibility if I use a preconfigured list. I'll look into making the list of properties configurable.

tardyp:
I was thinking of a list of fnmatch expr

noc0lour:
I think the drawback of having a `fnmatch` expression is the whole tree has to be flattened first before it can be applied. Otherwise it couldn't match subtrees like `github.base`. or we match backwards (e.g. `fnmatch.match(desired_key, key+"*")` but then we don't have any advantages over using `startswith`.
if you are ok with first flatten the whole file and then matching desired entries (I'd prefer `re` over `fnmatch` anytime).

noc0lour:
I have no idea why the coverage suddenly drops in files I did not even touch? And random things fail in the CI tests? @tardyp ? All files I touched have the same or better coverage.

tardyp:
@noc0lour its because py3.5 tests fail, thus the coverage of py3 is not taken in account

master/buildbot/test/unit/test_changes_github.py
```diff
@@ -162,6 +239,11 @@ def test_SimplePR(self):
         self.assertEqual(change['repository'],
                          'https://github.com/defunkt/buildbot.git')
         self.assertEqual(change['files'], ['README.md'])
+        self.assertDictContainsSubset(_GH_PARSED_PROPS, change['properties'])
```
tardyp: this is deprecated in py3.2 (and we dont allow warnings)

https://docs.python.org/3/whatsnew/3.2.html?highlight=assertdictcontainssubset

you probably will need to replace that with a for k,v in _GH_PARSED_PROPS.items()

master/buildbot/test/unit/test_changes_github.py
```diff
@@ -210,6 +297,11 @@ def test_baseURL(self):
...
```
tardyp: ditto

master/buildbot/changes/github.py
```diff
@@ -39,10 +41,26 @@
...
 
 
+class PullRequestMixin(object):
+    def extractProperties(self, pr_info):
+        def flatten(properties, base, info_dict):
+            for k, v in iteritems(info_dict):
+                name = ".".join([base, k])
+                if isinstance(v, dict):
+                    flatten(properties, name, v)
+                elif any([fnmatch(name, expr)
```
tardyp: 👍

noc0lour:
Yay, codecov is happy again :)

noc0lour:
anything left :) ?

PR #3077, opened 2017-03-25
Added a https://pushover.net/ reporter.

macdems:
This patch added a Pushover (https://pushover.net) notifier. It allows to receive instant notification into a mobile device.

## Contributor Checklist:


macdems:
Will resubmit with fixes...

tardyp:
Please don't close pull request when you are working on fixes. You can just push updates to your branch so that we can review the changes and keep history

master/buildbot/reporters/pushover.py
```diff
@@ -0,0 +1,294 @@
...
+from buildbot.process.properties import Properties
+from buildbot.process.results import CANCELLED, EXCEPTION, FAILURE, SUCCESS, WARNINGS
+from buildbot.process.results import Results
+from buildbot.reporters import utils
+from buildbot.reporters.message import MessageFormatter as DefaultMessageFormatter
+from buildbot.reporters.message import MessageFormatterMissingWorker
+from buildbot.util import service
+
+try:
+    from pushover import Client
```
tardyp: pushover has a rest api. Please use this rest api rather than a python wrapper. this is best integrated with twisted.

master/buildbot/reporters/pushover.py
```diff
@@ -0,0 +1,294 @@
...
+            else:
+                mode = (mode,)
+        return mode
+
+    def checkConfig(self, user_key, api_token, mode=("failing", "passing", "warnings"),
+                    tags=None, builders=None, buildSetSummary=False, messageFormatter=None,
+                    subject="Buildbot %(result)s in %(title)s on %(builder)s",
+                    name=None, schedulers=None, branches=None, priorities=None, other_args=None):
+
+        if Client is None:
```
tardyp: please look at other reporters based on the HttpBase for example on how to use httpClientService.Please note that httpClientService has a framwework for unit tests that you should use.
http://docs.buildbot.net/current/developer/utils.html#buildbot-util-httpclientservice

master/buildbot/reporters/pushover.py
```diff
@@ -0,0 +1,294 @@
...
+                                               'title': self.master.config.title,
+                                               'builder': name}
+
+        yield self.sendMessage(**msg)
+
+    def sendMessage(self, **kwargs):
+        twlog.msg("sending pushover notification")
+        client = Client(self.user_key, api_token=self.api_token)
+        kwargs.update(self.other_args)
+        return client.send_message(**kwargs)
```
tardyp: this api call is probably blocking. we dont allow blocking api inside the main thread.

macdems:
You are right.  However, this would require splitting MailNotifier. I may take a look at it, but it will take some time. I would appreciate any suggestions what should stay in MailNotifier notifier and what should go to the new base class (of course I can just identify the common code and move it up, nut maybe there is a better way to make this decision).

By the way, Pushover is only one of many instant notifiers (e.g. Pushbullet, Pusjet, etc.). I guess it would be beneficial to be able to use this framework for the others as well.

tardyp:
I guess identifying common code is a good first step. No need to duplicate the test code as well.
Perhaps, later we can make a better factorization, but now we have a good example of two pretty different low level backends, doing the exact same high level message building.

We already have some of the other reporters which could benefit such factorization (stash, hipchat)
I am not asking you do the that bigger factorization right now, obviously.

tardyp:
I know this was asking a lot. Where you able to look at it? I think this would really be a great improvement to this codebase.

macdems:
This is ok. I'm sorry for the delay, but I have some urgent assignments at work. I should be able to get back to this next week.

master/buildbot/reporters/notifier.py
```diff
@@ -0,0 +1,292 @@
...
+            if 'subject' in buildmsg:
+                subject = buildmsg['subject']
+
+        yield self.sendMessage(body, subject, msgtype, name, results, builds,
+                               list(users), patches, logs)
+
+    def sendMessage(self, body, subject=None, type=None, builderName=None,
+                    results=None, builds=None, users=None, patches=None,
+                    logs=None, worker=None):
+        return
```
tardyp: raise NotImplemented()

macdems:
I am sorry for the delay, but I had some urgent work to do first.

The `PushoverNotifier` class is indeed strongly reduced now and it should be straightforward to adapt it for other notification services.

There are two things that still bother me:

1. How to test if the notifications are correctly sent? Of course the tests can be carefully written according to the service API, but it would be still desirable to check against the real service at least once (I confirm that Pushover works: it has already notified me about several failed builds of my job project — the very reason I did the patch in the first place).

1. Missing worker notification will probably have to be refactored, as a user might want to have mail and Pushover notification configured independently. However, this will probably require changes in user sofig, so I did not touch it now.

tardyp:
@macdems 

1- You can create an integration test for pushover. This test will take the creds as environment variable, and will raise SkipTest if its not configured. you can look at test_hyper.py for an example.

2-a user could setup two reporters, one for normal build, one for workers.

master/buildbot/test/unit/test_reporters_pushover.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @defer.inlineCallbacks
+    def setupBuildResults(self, results, wantPreviousBuild=False):
+        # this testsuite always goes through setupBuildResults so that
+        # the data is sure to be the real data schema known coming from data
+        # api
+
+        self.db = self.master.db
+        self.db.insertTestData([
+            fakedb.Master(id=92),
```
tardyp: Sorry I did not see this test, as it was hidden by default by github.

No need to retest the behaviour for pushover. It should also be factorized in a test for notifier.py

In unit/test_reporter_pushover, you should only test the correctness of send message, so you dont need to create all the fake data in fakedb.

master/buildbot/test/unit/test_reporters_pushover.py
```diff
@@ -0,0 +1,386 @@
...
+        self.assertRaises(config.ConfigErrors,
+                          PushoverNotifier, '1234', 'abcd',
+                          tags=['fast', 'slow'], builders=['a', 'b'])
+
+    def test_init_warns_notifier_mode_all_in_iter(self):
+        self.assertRaisesConfigError(
+            "mode 'all' is not valid in an iterator and must be passed in as a separate string",
+            lambda: PushoverNotifier('1234', 'abcd', mode=['all']))
+
+    @defer.inlineCallbacks
```
tardyp: I think only the above tests are needed. below are just the same, and should be factorized to a test for the baseclass.

you would just subclass and implement sendMessage in a UnderTestNotifier, and then keep the below tests there.

macdems:
From Buildbot documentation: 

> The email sent this way will use a MailNotifier (see MailNotifier) status target, if one is configured.

Recipients are configured inside the worker. But it can only be an email, so one needs an alternative way to configure other notifications. I would suggest adding a list of worker names to a reporter.

Do you have an integration test for a missing worker? I think it is needed to test notifications with two or more methods (e.g. two Pushover reporters).

tardyp:
> Recipients are configured inside the worker. But it can only be an email, so one needs an alternative way to configure other notifications. I would suggest adding a list of worker names to a reporter.

Good point. Actually the reporter can fully ignore the email list.
It receives a message from the mq. 
It can do the filtering that makes the most sense.
There is a lot of legacy in this design. Actually in eight the message sending code is literally hardcoded in worker code!

> Do you have an integration test for a missing worker? I think it is needed to test notifications with two or more methods (e.g. two Pushover reporters).
I dont think we have one. You can call worker missing data api in your integration test:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/worker/base.py#L322

macdems:
I tried to add integration test for missing worker. But if fails. Do I miss something? Could you tell me why the event is not triggered in https://github.com/macdems/buildbot/blob/master/master/buildbot/test/integration/test_notifier.py#L104

macdems:
Ok, I have found the problem (it was actually a bug in the upstream mail notifier). Please ignore the last push. I will try to finish refactoring of the tests and when done I will push again.

macdems:
I have removed spurious tests and added missing integration test for missingWorker (which revealed a buig in the current MailNotifier).

I have decided to keep `test_reporters_mail.py` unchanged, as (1) I believe the tests should to be touched if the behaviour of the tested code is not supposed to change and (2) this test tests the common code. `test_reporters_pushover.py` tests only the new part (Pushover `sendMessage`).

tardyp:
> I have decided to keep test_reporters_mail.py unchanged, as (1) I believe the tests should to be touched if the behaviour of the tested code is not supposed to change
I would agree to that but in this case you don't check that the factorisation has been done correctly, and that pushover has the same properties as mailnotifier.

I fear I have to insist in separating the tests about sending mail, and tests about generating correct sendMessage calls according to build history.
I think those tests are already well separated, this should not be too difficult.

macdems:
Would you prefer to make a fake descendant of `NotifierBase` and test the common code or to make a common ancestor of `TestMailNotifier` and `TestPushoverNotifier` which will test the common code twice for the real classes?

tardyp:
I think it is prefereable to do the  fake descendant of NotifierBase.
eventually we will have dozens of descendants of NotifierBase

macdems:
I hope it should do now...

tardyp:
hi @macdems I would prefer we do this one at a time.
Can you look at pushjet after that one is merged?

macdems:
Sure. I did Pushjet last night to test how easy it would be to add another notifier after the refactor. Took less than half an hour all together (most of which was tests and documentation).

master/buildbot/test/unit/test_reporters_pushover.py
```diff
@@ -0,0 +1,93 @@
...
+from twisted.trial import unittest
+
+from buildbot.process.results import SUCCESS
+from buildbot.reporters.pushover import PushoverNotifier
+from buildbot.test.fake import httpclientservice as fakehttpclientservice
+from buildbot.test.fake import fakemaster
+from buildbot.test.util.config import ConfigErrorsMixin
+from buildbot.util import httpclientservice
+
+py_27 = sys.version_info[0] > 2 or (sys.version_info[0] == 2
```
tardyp: please remove this line. It looks like its not used. For further code, please use this instead:

    from future import PY3

master/buildbot/test/unit/test_reporters_notifier.py
```diff
@@ -0,0 +1,347 @@
...
+from buildbot.process.results import FAILURE
+from buildbot.process.results import SUCCESS
+from buildbot.process.results import WARNINGS
+from buildbot.reporters.notifier import NotifierBase
+from buildbot.test.fake import fakedb
+from buildbot.test.fake import fakemaster
+from buildbot.test.util.config import ConfigErrorsMixin
+from buildbot.test.util.notifier import NotifierTestMixin
+
+py_27 = sys.version_info[0] > 2 or (sys.version_info[0] == 2
```
tardyp:
```
from future import PY3
```

master/buildbot/test/util/notifier.py
```diff
@@ -0,0 +1,77 @@
...
+from __future__ import absolute_import
+from __future__ import print_function
+
+from twisted.internet import defer
+
+from buildbot.reporters import utils
+from buildbot.test.fake import fakedb
+
+
+class NotifierTestMixin(object):
```
tardyp: 👍

master/docs/manual/cfg-reporters.rst
```diff
@@ -418,6 +422,51 @@ The default ``ctx`` for the missing worker email is made of:
...
+However, ``watchedWorkers`` defaults to *None*.
+
+The following additional parameters are accepted by this class:
+
+``user_key``
+    The user key from the Pushover website. It is used to identify the notification recipient.
+
+``api_token``
+    API token for a custom application from the Pushover website.
+
```
tardyp: a link to api_token generation will be useful for people. Or we will get support requests from people putting their password in there

master/buildbot/reporters/notifier.py
```diff
@@ -0,0 +1,304 @@
...
+                    name=None, schedulers=None, branches=None,
+                    watchedWorkers=None, messageFormatterMissingWorker=None):
+
+        for m in self.computeShortcutModes(mode):
+            if m not in self.possible_modes:
+                if m == "all":
+                    config.error(
+                        "mode 'all' is not valid in an iterator and must be passed in as a separate string")
+                else:
+                    config.error(
```
tardyp: Do you have https://codecov.io/ chrome plugin installed?

There are a trivial small tests that are missing here to get 100% diff coverage

master/setup.py
```diff
@@ -317,7 +317,9 @@ def define_plugin_entries(groups):
                 'RemoveDirectory', 'MakeDirectory']),
         ]),
         ('buildbot.reporters', [
+            ('buildbot.reporters.notifier', ['Notifier']),
```
tardyp: Where is this Notifier plugin? I don't see it.

macdems:
Should be `NotifierBase`, but should this abstract class be put there at all?

tardyp:
> Should be NotifierBase, but should this abstract class be put there at all?
No need to put it there. If ppl want custom notifier, they can import in real module path

macdems:
I'm lost. I have just realized a trivial change 1bd1e0d resulted in all these failed tests. Honestly, I have no idea why . The log errors seem totally unrelated.

PR #3079, opened 2017-03-26
steps: add assert renderable step to master steps

noc0lour:
This step allows to assert an user created expression. It supports renderables, callables and constant values. (But constant values don't make sense to me).
Unit tests should cover every code path.

### DIFF
--- a/master/buildbot/newsfragments/process_property.feature
+++ b/master/buildbot/newsfragments/process_property.feature
@@ -1,0 +1,1 @@
+The class :py:class:`~buildbot.process.properties.Property` now allows being used with Python built in comparators. It will return a Renderable which executes the comparison.
--- a/master/buildbot/newsfragments/steps_master_assert.feature
+++ b/master/buildbot/newsfragments/steps_master_assert.feature
@@ -1,0 +1,1 @@
+New build step :py:class:`~buildbot.steps.master.Assert` Tests a renderable or constant if it evaluates to true. It will succeed or fail to step according to the result.
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -713,6 +713,29 @@ class Interpolate(util.ComparableMixin, object):
 
 
 @implementer(IRenderable)
+class _ComparisonRenderer(util.ComparableMixin, object):
+    """
+    An instance of this class renders a comparison given by a comparator
+    function with v1 and v2
+
+    """
+
+    compare_attrs = ('fn',)
+
+    def __init__(self, v1, v2, cstr, comparator):
+        self.v1, self.v2, self.comparator, self.cstr = v1, v2, comparator, cstr
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, props):
+        v1 = yield props.render(self.v1)
+        v2 = yield props.render(self.v2)
+        defer.returnValue(self.comparator(v1, v2))
+
+    def __repr__(self):
+        return '%r %r %r' % (self.v1, self.cstr, self.v2)
+
+
+@implementer(IRenderable)
 class Property(util.ComparableMixin):
 
     """
@@ -735,6 +758,27 @@ class Property(util.ComparableMixin):
 
         # Report on parent frame.
         _on_property_usage(key, stacklevel=1)
+
+    def __eq__(self, other):
+        return _ComparisonRenderer(self, other, "==", lambda v1, v2: v1 == v2)
+
+    def __ne__(self, other):
+        return _ComparisonRenderer(self, other, "!=", lambda v1, v2: v1 != v2)
+
+    def __lt__(self, other):
+        return _ComparisonRenderer(self, other, "<", lambda v1, v2: v1 < v2)
+
+    def __le__(self, other):
+        return _ComparisonRenderer(self, other, "<=", lambda v1, v2: v1 <= v2)
+
+    def __gt__(self, other):
+        return _ComparisonRenderer(self, other, ">", lambda v1, v2: v1 > v2)
+
+    def __ge__(self, other):
+        return _ComparisonRenderer(self, other, ">=", lambda v1, v2: v1 >= v2)
+
+    def __repr__(self):
+        return "Property({0})".format(self.key)
 
     def getRenderingFor(self, props):
         if self.defaultWhenFalse:
--- a/master/buildbot/steps/master.py
+++ b/master/buildbot/steps/master.py
@@ -213,6 +213,23 @@ class SetProperties(BuildStep):
         return defer.succeed(SUCCESS)
 
 
+class Assert(BuildStep):
+    name = 'Assert'
+    description = ['Checking..']
+    descriptionDone = ["checked"]
+    renderables = ['check']
+
+    def __init__(self, check, **kwargs):
+        BuildStep.__init__(self, **kwargs)
+        self.check = check
+        self.descriptionDone = ["checked {}".format(repr(self.check))]
+
+    def run(self):
+        if self.check:
+            return defer.succeed(SUCCESS)
+        return defer.succeed(FAILURE)
+
+
 class LogRenderable(BuildStep):
     name = 'LogRenderable'
     description = ['Logging']
--- a/master/buildbot/test/unit/test_process_properties.py
+++ b/master/buildbot/test/unit/test_process_properties.py
@@ -1232,6 +1232,62 @@ class TestProperty(unittest.TestCase):
                       ["string", "bla", "string", "bla"])
         return d
 
+    @defer.inlineCallbacks
+    def testCompEq(self):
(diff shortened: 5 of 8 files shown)


noc0lour:
looks like the hyper latent worker is not happy with the smoke tests.

noc0lour:
Any comments :) ?

master/buildbot/steps/master.py
```diff
@@ -213,6 +213,26 @@ def run(self):
         return defer.succeed(SUCCESS)
 
 
+class AssertRenderable(BuildStep):
```
tardyp: Would it be better to just call it Assert()

Renderable is kindof implicit in buildbot, everything could be renderable

master/docs/manual/cfg-buildsteps.rst
```diff
@@ -2327,6 +2327,15 @@ LogRenderable
...
 
+.. bb:step:: AssertRenderable
+
+AssertRenderable
+++++++++++++++++
+
+.. py:class:: buildbot.steps.master.AssertRenderable
+
+This build step takes the expression passed in as first argument. If the expression is not callable (already evaluated by the renderer) it will test if this result evaluates to ``True``. If the expression is callable it will pass in ``self.getProperties()`` and test if the result evaluates to ``True``.
+
```
tardyp: I agree that the callable functionality can be handy, but I would stick with only supporting renderable mechanism.

Another thing that would be good, and is coherent with the all-is-renderable principle is to support comparaisons for Property using __eq__ and __ne__


so that
```python
   Assert(util.Property("foo") == "bar")
```
works as expected, and generates a renderable like
```python
   Assert(util.renderer(lambda x: x.getProperty("foo") == "bar")))
```

master/buildbot/test/unit/test_steps_master.py
```diff
@@ -280,3 +280,44 @@ def manipulate(props):
...
+
+    def setUp(self):
+        return self.setUpBuildStep()
+
+    def tearDown(self):
+        return self.tearDownBuildStep()
+
+    def test_lambda_pass(self):
+        self.setupStep(master.AssertRenderable(
+            lambda prop: prop.getProperty("test_prop") == "foo"))
```
tardyp: well, this is not really a renderable, this is a function.
I think it is better to stick with renderable
```python
    self.setupStep(master.Assert(util.renderable(lambda prop:prop.getProperty("test_prop") == "foo"))))
```

master/buildbot/steps/master.py
```diff
@@ -213,6 +213,26 @@ def run(self):
...
+    description = ['Checking..']
+    descriptionDone = ['Checked']
+    renderables = ['fncheck']
+
+    def __init__(self, fncheck, **kwargs):
+        BuildStep.__init__(self, **kwargs)
+        self.fncheck = fncheck
+
+    def run(self):
+        if not callable(self.fncheck):
```
tardyp: I would remove the callable check, and only support boolean via renderable.

noc0lour:
Addressed issues from the review. Added the nice comparator feature to `Property` so people can use any comparator on a `Property`

master/buildbot/process/properties.py
```diff
@@ -720,6 +720,24 @@ def __init__(self, key, default=None, defaultWhenFalse=True):
         # Report on parent frame.
         _on_property_usage(key, stacklevel=1)
 
+    def __eq__(self, other):
+        return renderer(lambda props: props.getProperty(self.key) == other)
```
tardyp: I would render other, so that we can say Property("x") == Property("y")

```python
@implementer(IRenderable)
class _ComparaisonRenderer(util.ComparableMixin, object):

    compare_attrs = ('fn',)

    def __init__(self, v1, v2, cstr, comparator):
        self.v1, self.v2, self.comparator, self.cstr = v1, v2, comparator, cstr

    @defer.inlineCallbacks
    def getRenderingFor(self, props):
        v1 = yield props.render(self.v1)
        v2 = yield props.render(self.v2)
        defer.returnValue(self.comparator(v1, v2))

    def __repr__(self):
        return '%r %r %r' % (self.v1, self.cstr, self.v2)

def __eq__(self, other):
    return _ComparaisonRenderer(self, other, "==", lambda v1, v2: v1 == v2)
```

master/buildbot/test/unit/test_process_properties.py
```diff
@@ -1232,6 +1232,42 @@ def testFlattenListAdd2(self):
                       ["string", "bla", "string", "bla"])
         return d
 
+    @defer.inlineCallbacks
+    def testCompEq(self):
+        self.props.setProperty("do-tests", "string", "scheduler")
+        result = yield self.build.render(Property("do-tests") == "string")
+        self.assertEqual(result, True)
```
tardyp: need a test for
```python
        result = yield self.build.render("string" == Property("do-tests"))
```
Not sure if it will work, but at least it should be tested (maybe put it unittest.expectedFailure)

master/docs/manual/cfg-buildsteps.rst
```diff
@@ -2327,6 +2327,15 @@ LogRenderable
 This build step takes content which can be renderable and logs it in a pretty-printed format.
 It can be useful for debugging properties during a build.
 
+.. bb:step:: Assert
+
+Assert
+++++++++++++++++
```
tardyp: no need for that many ``+++``

master/buildbot/steps/master.py
```diff
@@ -213,6 +213,22 @@ def run(self):
...
+    descriptionDone = ['Checked']
+    renderables = ['check']
+
+    def __init__(self, check, **kwargs):
+        BuildStep.__init__(self, **kwargs)
+        self.check = check
+
+    def run(self):
+        if self.check:
+            return defer.succeed(SUCCESS)
```
tardyp: I would set descriptionDone to  ["checked {}".format(self.check_repr)]

master/buildbot/steps/master.py
```diff
@@ -213,6 +213,22 @@ def run(self):
...
 
+class Assert(BuildStep):
+    name = 'Assert'
+    description = ['Checking..']
+    descriptionDone = ['Checked']
+    renderables = ['check']
+
+    def __init__(self, check, **kwargs):
+        BuildStep.__init__(self, **kwargs)
+        self.check = check
```
tardyp: self.check_repr = repr(check)

noc0lour:
I'm currently pretty involved at work, probably I'll fix the comments next week.

tardyp:
I'm sorry I need to delay this for next release. I hope you can finish it soon. Its pretty awesome!

noc0lour:
No worries, I was pinned down by other things right now (:

master/buildbot/steps/master.py
noc0lour: will this just mangle the `__str__` representation and won't change the value?

master/buildbot/steps/master.py
noc0lour: Is that useful for debugging: `'checked renderer(<function test_renderer at 0x7fd8d139c230>)'` ?

master/buildbot/steps/master.py
tardyp: it is on addition to self.check = check

master/buildbot/steps/master.py
tardyp: maybe the repr for renderer should make sure the address of function is not written

master/buildbot/test/unit/test_process_properties.py
noc0lour: Works, since Python uses reflection with the rhs comparators if the type on the lhs throws NotImplemented.

noc0lour:
Now all possible scenarios work. Comparison renderable with renderable, renderable with value and value with renderable. :tada:

noc0lour:
Derp, need to fix the expected output for Assert renderable step.

master/buildbot/newsfragments/steps_master_assert.feature
```diff
@@ -0,0 +1 @@
+New build step :py:class: `~buildbot.steps.master.Assert` Tests a renderable or constant if it evaluates to true. It will succeed or fail to step according to the result.
```
tardyp: no space between 'class:' and '`~'

master/buildbot/newsfragments/process_property.feature
```diff
@@ -0,0 +1 @@
+The class :py:class: `~buildbot.process.properties.Property` now allows being used with Python built in comparators. It will return a Renderable which executes the comparison.
```
tardyp: no space between 'class:' and '`~'

Frodox:
@noc0lour could you please write newsfragment into Property's doc also, with examples? Since it is very strong and powerful feature :)

PR #3080, opened 2017-03-27
Support docker-py > 1.10.6

anish:
This has been renamed to docker in pypi, and provides a backwards compatible API

### DIFF
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -16,6 +16,8 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+version = "1.10.6"
+
 
 class Client(object):
     latest = None
@@ -24,6 +26,7 @@ class Client(object):
         Client.latest = self
         self.call_args_create_container = []
         self.call_args_create_host_config = []
+        self.called_class_name = None
         self._images = [{'RepoTags': ['busybox:latest', 'worker:latest']}]
         self._containers = {}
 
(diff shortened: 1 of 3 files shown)


master/buildbot/worker/docker.py
```diff
@@ -201,7 +202,10 @@ def _image_exists(self, client, name):
         return False
 
     def _thd_start_instance(self, image, dockerfile, volumes):
-        docker_client = client.Client(**self.client_args)
+        if docker.version[0] == '1':
```
tardyp: You should factorize this in getDockerClient:

master/buildbot/test/fake/docker.py
```diff
@@ -82,3 +84,7 @@ def create_container(self, image, *args, **kwargs):
 
     def remove_container(self, id, **kwargs):
         del self._containers[id]
+
+
+class APIClient(Client):
```
tardyp: I dont think it is really needed

master/buildbot/worker/docker.py
anish: sounds good, will do

master/buildbot/test/fake/docker.py
anish: mostly added as a test requirement, I will double check it if it's still needed after refactorization

anish:
changes made, github doesn't pick up force push

master/buildbot/test/fake/docker.py
anish: yep, still needed for tests

PR #3083, opened 2017-03-27
added github tag and release handlers

manikantanr:
@tardyp Please review these changes.

This adds github tag and release event handlers.

### DIFF
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -177,6 +177,8 @@ class GitHubEventHandler(object):
             branch = payload['ref']
         elif category == 'release':
             branch = payload['release']['tag_name']
+        else:
+            return []
 
         log.debug('Processing {0},{1}'.format(category, branch))
         changes = []
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -27,6 +27,7 @@ from twisted.python import log
 
 from buildbot.util import bytes2NativeString
 from buildbot.util import unicode2bytes
+from buildbot.util.logger import Logger
 
 try:
     import json
@@ -170,20 +171,21 @@ class GitHubEventHandler(object):
             len(changes), number))
         return changes, 'git'
 
-    def get_changes(self, category, payload):
+    def _get_changes_for_misc_event(self, category, payload):
+        log = Logger()
         if category == 'create':
             branch = payload['ref']
         elif category == 'release':
             branch = payload['release']['tag_name']
 
-        log.msg('Processing {0},{1}'.format(category, branch),
-                logLevel=logging.DEBUG)
+        log.debug('Processing {0},{1}'.format(category, branch))
         changes = []
         change = {
             'when_timestamp': dateparse(payload['repository']['created_at']),
             'repository': payload['repository']['html_url'],
             'project': payload['repository']['full_name'],
-            'branch': branch, 'revision': branch,
+            'branch': branch,
+            'revision': branch,
             'category': category,
             'author': payload['sender']['login'],
             'comments': 'Created %s' % branch,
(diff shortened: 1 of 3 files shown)


master/buildbot/www/hooks/github.py
```diff
@@ -170,6 +170,41 @@ def handle_pull_request(self, payload, event):
             len(changes), number))
         return changes, 'git'
 
+    def get_changes(self, category, payload):
```
tardyp: what is the difference with _process_changes?

manikantanr: create event need another set of json , so kept it in separate functions which wont disturb other events.

tardyp: Then this needs a renaming in order to have better understanding of which does what

master/buildbot/www/hooks/github.py
```diff
@@ -170,6 +170,44 @@ def handle_pull_request(self, payload, event):
...
+        if category == 'create':
+            branch = payload['ref']
+        elif category == 'release':
+            branch = payload['release']['tag_name']
+
+        log.msg("Processing %s #%s" % (category, branch))
```
tardyp: please put such log message optional (with a debug flag)

master/buildbot/www/hooks/github.py
```diff
@@ -176,7 +176,8 @@ def get_changes(self, category, payload):
         elif category == 'release':
             branch = payload['release']['tag_name']
 
-        log.msg("Processing %s #%s" % (category, branch))
+        log.msg('Processing {0},{1}'.format(category, branch),
```
tardyp: if you want to implement this with log level, please use new logger api:

https://twistedmatrix.com/documents/current/core/howto/logger.html
```python
from buildbot.util.logger import Logger

log = Logger()
...
log.debug("xxx")
```

master/buildbot/www/hooks/github.py
tardyp: Please call this something like

_get_changes_for_misc_event()

add a comment telling that this is not really a change, as it does not reflect a strict code change, but still is a useful event to start a build on.

This needs more explaination also in the doc (it was really not clear for me at first)

master/docs/manual/cfg-wwwhooks.rst
```diff
@@ -150,7 +150,7 @@ The parameters are:
                 ...))
 
 :guilabel:`Which events would you like to trigger this webhook?`
-    Leave the default -- ``Just the push [tag]  events`` -- other kind of events are not currently supported.
+    Leave the default -- ``Just the push [tag], create, release,  events`` -- other kind of events are not currently supported.
```
tardyp: Please write a paragraph explaining that push and pull_requests events are generating normal change events with list of file changed, while create and release events generate "dummy" changes, which does not reflect code changes, but rather useful event to start a build upon

rmkrjy: @tardyp made few changes based on your suggestions.

manikantanr:
@tardyp  I have updated the docs based on your suggestions , please check it and let me know your views on this PR.

master/docs/manual/cfg-wwwhooks.rst
```diff
@@ -150,10 +150,18 @@ The parameters are:
...
 And then press the ``Add Webhook`` button.
 
+.. note::
+
+   Create and Release events does not reflect a strict code change, but still is a useful event to start a build on.
+
+   Even though tag and release does not associate with real change it contains a change set. This change set has branch and revision properties with tag name as value.
+
+   These values can be utilized in buildsteps using interpolations, util.Interpolate("%(prop:branch)s") and util.Interpolate("%(prop:revision)s").
+
```
noc0lour: in this case you could also just use `util.Property("branch")` and `util.Property("revision")` as example.

master/buildbot/www/hooks/github.py
```diff
@@ -170,6 +171,46 @@ def handle_pull_request(self, payload, event):
...
+        elif category == 'release':
+            branch = payload['release']['tag_name']
+
+        log.debug('Processing {0},{1}'.format(category, branch))
+        changes = []
+        change = {
+            'when_timestamp': dateparse(payload['repository']['created_at']),
+            'repository': payload['repository']['html_url'],
+            'project': payload['repository']['full_name'],
+            'branch': branch,
```
noc0lour: This will fail if category is not `create` or `release`. For the current code that's fine. But maybe you can put in a `else:` case and return an empty list of changes and emit a log message.

manikantanr:
@noc0lour made changes you requested.

noc0lour:
Looks good, the codecov results look somewhat wrong. They don't hit lines which are hit every run ... Other than that I'd say this should be good to go @tardyp.

PR #3087, opened 2017-03-28
Only require 'distro' on linux (as it is linux specific)

grembo:
This is a trivial/obvious fix.

### DIFF
--- a/master/setup.py
+++ b/master/setup.py
@@ -453,7 +453,7 @@ setup_args['install_requires'] = [
     'txaio ' + txaio_ver,
     'autobahn ' + autobahn_ver,
     'PyJWT',
-    'distro'
+    'distro;platform_system==="Linux"'
 ]
 
 # Unit test dependencies.
--- a/master/setup.py
+++ b/master/setup.py
@@ -25,11 +25,13 @@ from __future__ import print_function
 import glob
 import os
 import pkg_resources
+import platform
 import sys
 from distutils.command.install_data import install_data
 from distutils.command.sdist import sdist
 from distutils.version import LooseVersion
 
+from setuptools import version as setuptools_version
 from setuptools import setup
 
 from buildbot import version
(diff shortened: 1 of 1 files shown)


master/setup.py
```diff
@@ -453,7 +453,7 @@ def define_plugin_entries(groups):
     'txaio ' + txaio_ver,
     'autobahn ' + autobahn_ver,
     'PyJWT',
-    'distro'
+    'distro;platform_system==="Linux"'
```
tardyp: I wasn't aware of this construct. Is this supported by all pip/setuptools versions?
Does that work with wheels?

grembo: This was defined in https://www.python.org/dev/peps/pep-0508/. Based on this post https://discourse.numenta.org/t/setup-py-error-invalid-environment-marker/1298/3 it's supported by:

 - setuptools: 20.2.2 +
 - wheel: 0.24.0 +
 - pip: 8.1.2 +

seankelly: That is a much newer version of setuptools than currently required.

grembo: If this is critical, the alternative would be something like

```
import platform
...
if platform.system() == 'Linux':
    setup_args['install_requires'].append('distro')

```
If you prefer this, I can update the pull request.

grembo: updated to use a more conservative approach to accomplish the same

tardyp: Problem with that approach is that it does not work with wheels. As I generate the universal wheels on my mac, it wont include distro anymore in the wheels deps.

I would be more in favor of the combination, which would test setup tools version :-/

grembo: Can you give me an example of what your suggested approach would look like?

tardyp:
```python
import platform
from distutils.version import LooseVersion
import setuptools
...
if LooseVersion(setuptools.version.__version__) >= LooseVersion("20.2.2"):
    setup_args['install_requires'].append('distro;platform_system==="Linux')
elif platform.system() == 'Linux':
    setup_args['install_requires'].append('distro')
```

tardyp:
@grembo Hope you don't mind, in order to move faster I did edit your PR last commit to include the proposed change. Let me know if this works for you (and your BSD port)

grembo:
@tardyp Sorry for the late response, this is ok with me. I'll change/remove the patch to the port on the next release, so it's all good.

PR #3089, opened 2017-03-29
Add fix to docker latent work to support tar.gz files and checksum to tell when to rebuild docker file

dragon512:
This add the ability to pass in tar.gz data which is useful for building an dockerfile on a remote host. dockerfile is built when dockerfile is updated (not every time). This also allows image and dockerfile to be defined and have the image get updated if the content of the dockerfile gets updated.

add custom_context arg to control if dockerfile is tar.gz or a text string
Add checksum logic to know when to build docker file data

**Need help with writing tests as I tested this in "production"**

Will update the docs shortly

## Contributor Checklist:

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -117,9 +117,10 @@ class DockerLatentWorker(DockerBaseWorker):
 
     def checkConfig(self, name, password, docker_host, image=None, command=None,
                     volumes=None, dockerfile=None, version=None, tls=None, followStartupLogs=False,
-                    masterFQDN=None, hostconfig=None, **kwargs):
+                    masterFQDN=None, hostconfig=None, custom_context=False, **kwargs):
 
-        DockerBaseWorker.checkConfig(self, name, password, image, masterFQDN, **kwargs)
+        DockerBaseWorker.checkConfig(
+            self, name, password, image, masterFQDN, **kwargs)
 
         if not client:
             config.error("The python module 'docker-py>=1.4' is needed to use a"
@@ -145,7 +146,7 @@ class DockerLatentWorker(DockerBaseWorker):
     @defer.inlineCallbacks
     def reconfigService(self, name, password, docker_host, image=None, command=None,
                         volumes=None, dockerfile=None, version=None, tls=None, followStartupLogs=False,
-                        masterFQDN=None, hostconfig=None, **kwargs):
+                        masterFQDN=None, hostconfig=None, custom_context=False, **kwargs):
 
         yield DockerBaseWorker.reconfigService(self, name, password, image, masterFQDN, **kwargs)
         self.volumes = volumes or []
@@ -153,6 +154,7 @@ class DockerLatentWorker(DockerBaseWorker):
 
         self.command = command or []
         self.dockerfile = dockerfile
+        self.custom_context = custom_context
         self.hostconfig = hostconfig or {}
         # Prepare the parameters for the Docker Client object.
         self.client_args = {'base_url': docker_host}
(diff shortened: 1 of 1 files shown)


tardyp:
I would prefer not having that much logic in the docker latent worker code.

I would prefer this is separated from the latent worker, and rather have people setup a separate builder to rebuild the worker dockerfile when the dockerfile repo changes.

- It is quite hard to debug when things goes bad in docker build as you dont have the logs accessible easily.
- Having the dockerfile build parametrised will imho make it even more error prone and hard to debug

I think it is better to choose different prebuilt image according to properties, and also to send runtime environments to dynamically customize the image.

dragon512:
I was going to ask how ( if it is possible) to pass data to the build step. I agree it is a pain that this goes the buildmaster log file.

I was looking how to setup logic to have buildbot rebuild the docker image and push it to a remote host correctly when the content changed as a builder. This was more complex as:

1) you want builder dependent on the update image to be stalled until the new image is made
2) this has to be push to the remote host correctly without requiring a docker repository. 
3) the general logic for the builder/source change is more complex as one would like a repo change to see if some files that define the dockerfile info/content have changed. This is need to allow 1) to happen in some form.

As a side note here it would be helpful to allow buildbot-worker create logic to pass some value to say look this value up in Environment. so the password, worker name can be easily defined as buildbot suggest in the documenation

tardyp:
docker repository is not very complex to setup, I think you should rather go that way instead of trying hacks, it will then unlock lots of options.

You should have worker dockerfile in separate repos as your tested source, so that you don't need to bother about dependent scheduler.
if a build triggers while the dockerfile is being rebuilt, so be it, it will run on the older one without the new feature.

>As a side note here it would be helpful to allow buildbot-worker create logic to pass some value to say look this value up in Environment. so the password, worker name can be easily defined as buildbot suggest in the documenation

This is why our example dockerfile for buildbot-worker are not using create worker.
create worker is really just a simple script that generates the buildbot.tac. Its actually simpler from dockerfile perspective to just copy your custom environment driven buildbot.tac where it should be.

dragon512:
The issue here is more about having the source contain information about the build. At least here it is big deal for various people to have a commit fix a few lines of source that requires an updated package to be installed. This commit should build correctly, not build on an old setup. 

If the suggestion is that people should have a source change start a "build the image" then trigger the real build chains. I don't disagree. 

to clarify this. What would be ideal is to have a repo with the source and buildbot info. the build bot info would consist of information about different environments ( as dockerfile or something ) and information about build steps to have run on any given environment, etc. At this abstract level it is easy to see how to have some python code that can read and define a information about workers, number of workers, status updates, build steps. This can be done in a way that would build the "Workers" before building the "code". minus where the text goes for the docker worker when it builds a dockerfile this works nice over trigger as the chain is nice an linear in one file ( not bouncing around different chains). 

However it is unclear how to mange the workers. I don't have examples on how I can start/stop/wait for workers in the build master. Ideally I looking how does one have a buildbot gracefully shutdown another buildbot and start it gain with updated build info. This would be ideal in that changes can be pushed to a branch and the main "build manager" system restart correct and continue building ( details missing on doing this correctly) This was all changes can be "tested" before being promoted using the same system. I don't know how this could be done yet with Buildbot.. but given this and a UI extension.. I agree this would be a better known method than what I suggest in this PR as a fix

tardyp:
> If the suggestion is that people should have a source change start a "build the image" then trigger the real build chains. I don't disagree.

This is how we do for the CI of buildbot https://github.com/buildbot/metabbotcfg/tree/nine/docker/metaworker

Imho, You don't really need to put hard dependencies between docker builds and the CI. just let them leave independently, and maybe sometimes, hit the rebuild button.

dragon512:
thanks for the update

FYI, In our case we want to check in "build info" so it can be updated like source code. In this case we get a case in which the image needs to be rebuilt before the source, the source build will fail. Ya a rebuild button can be pushed, however this is view here as a broken build setup. make a new prototype to have change filters help with this case, and build the images as part of a normal "source change" that will then trigger the source build after the image part has finished updating the image. If the image does not need to be updated the source build just gets trigger and we skip the image build. This will be better I think as well, as it can test and validate requirements we might have on the image before it is used.

Issue #3021, opened 2017-03-10
Inconsistency in foreign key when inserting entry in scheduler_changes table


mthalmei:
The trac ticket http://trac.buildbot.net/ticket/3532 already adresses a problem when inserting entries into scheduler_changes table.

Although this issue is fixed for SingleBranchScheduler it seems to be still unfixed for the Nightlyscheduler we are using and still getting the same errors.

This ticket is a migrated Trac ticket 3660

wstering:
PostgreSQL 9.4 shows the following error, when processing a change for the NightlyScheduler:
```
ERROR:  insert or update on table "scheduler_changes" violates foreign key constraint "scheduler_changes_schedulerid_fkey"
DETAIL:  Key (schedulerid)=(18) is not present in table "schedulers".
STATEMENT:  INSERT INTO scheduler_changes (schedulerid, changeid, important) VALUES (18, 721, 1)
```

Picking up changes works OK for SingleBranchScheduler, as noted above.

PR #3090, opened 2017-03-31
Nightly: use serviceid instead of objectid for classification

tardyp:
fixes https://github.com/buildbot/buildbot/issues/3021

serviceid is pointing to the schedulerid, and is the id required for classification, while objectid is for the state table
this is the same fix as http://trac.buildbot.net/ticket/3532.
I did grep for other use of xxxChangeClassifications to find the same bug class, but it seems nightly is the only remaining

### DIFF
--- a/master/docs/conf.py
+++ b/master/docs/conf.py
@@ -148,7 +148,8 @@ intersphinx_mapping = {
 extlinks = {
     'pull': ('https://github.com/buildbot/buildbot/pull/%s', 'pull request '),
     'issue': ('https://github.com/buildbot/buildbot/issue/%s', 'issue # '),
-    'bug': ('http://trac.buildbot.net/ticket/%s', 'bug #'),  # deprecated. Use issue instead, and point to GH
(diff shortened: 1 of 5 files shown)


master/docs/conf.py
```diff
@@ -146,8 +146,9 @@
 }
 
 extlinks = {
-    'bug': ('http://trac.buildbot.net/ticket/%s', 'bug #'),
     'pull': ('https://github.com/buildbot/buildbot/pull/%s', 'pull request '),
+    'issue': ('https://github.com/buildbot/buildbot/issue/%s', 'issue # '),
+    'bug': ('http://trac.buildbot.net/ticket/%s', 'bug #'),  # deprecated. Use issue instead, and point to GH
```
seankelly: pylint does like using `GH` for GitHub.

seankelly:
The bb/trial/3.6 failures appear to be unrelated. They're all in buildbot_worker.

PR #3094, opened 2017-04-03
Trigger step issue 3088

MikeAlexeev:
Hello,
this is PR for issue 3088 (trigger: combine waitForFinish and not waitForFinish behavior )

There is a draft version. I ran all tests, but haven't added any new.

The main idea is to start next step in build before scheduling child build from trigger.
So build is not blocked (it can execute next step) and build is waiting child build for finish.

There is one change in Build.addStep to prevent race conditions when build adds steps to DB. 
Race condition could be because addStep can be invoked simultaniosly in startNextStep() when step is finished and from trigger step in startNextStep() call.

Also there can be case when last step in parent build finished earlier then children.
Decision about finish of build was taken in Build.startNextStep. Build is finished If there is no next steps.
But it is not work in case that I have described. So I moved this check to Build._stepDone.
I also had to compare number of done steps with total amount of steps.
Build.executedSteps is updated before step is done. That's why I have added doneSteps list and update 
it when step is done for correct compare.

Maybe executedSteps can be updated when step is done? If so, doneSteps can be removed.

Does this solution look ok?
What tests I should add?

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -101,6 +101,7 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
         # overall results, may downgrade after each step
         self.results = SUCCESS
         self.properties = properties.Properties()
+        self.nsteps = 0
 
     def setBuilder(self, builder):
         """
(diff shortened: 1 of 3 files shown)


MikeAlexeev:
Hi, Pierre. 
Thank you for review.
I think build.runningSteps should work.
I will prepare fixes in next two or tree days or earlier if there is enough time.

I haven't thought yet about generalization of this behavior to BaseStep.
Maybe it would be useful

PR #3095, opened 2017-04-03
fixed pattern-based authz check

MikeAlexeev:
Hello.

It seems that I have found couple of bugs in authz.
First, pattern-based roles checking didn't work as described in doc.
There was only simple check if rule's role in user's list of roles.

Second, authz check finished successfully if role had defaultDeny=False.
For example in tests ( buildbot.test.unit.test_www_authz ) all checks returned success.
Because first role to compare was
AnyEndpointMatcher(role="admins", defaultDeny=False)
So if user has not "admins" in his roles this check was successful too.

Please correct me if I'm wrong.

### DIFF
--- a/master/buildbot/test/unit/test_www_authz.py
+++ b/master/buildbot/test/unit/test_www_authz.py
@@ -104,6 +104,13 @@ class Authz(www.WwwTestMixin, unittest.TestCase):
                          buildrequestid=82, number=5),
         ])
 
+    def setAllowRules(self, allow_rules):
+        # we should add links to autz and master instances in each new rule
+        for r in allow_rules:
+            r.setAuthz(self.authz)
+
+        self.authz.allowRules = allow_rules
+
     def assertUserAllowed(self, ep, action, options, user):
         return self.authz.assertUserAllowed(tuple(ep.split("/")), action, options, self.users[user])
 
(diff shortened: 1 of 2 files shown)


master/buildbot/www/authz/authz.py
```diff
@@ -89,10 +89,13 @@ def assertUserAllowed(self, ep, action, options, userDetails):
...
                         roles.update(
                             self.getOwnerRolesFromUser(userDetails, owner))
-                if rule.role not in roles:
-                    if rule.defaultDeny:
-                        raise Forbidden(
-                            "you need to have role '%s'" % (rule.role,))
-                elif not rule.defaultDeny:
-                    defer.returnValue(None)
+                for role in roles:
+                    if self.match(role, rule.role):
```
tardyp: looks good, but we need several unit tests to match these use cases

MikeAlexeev: Ok. I'm going to prepare tests today.

MikeAlexeev: I have question about tests.
In Authz class we have setUp method where self.authz.match (via authz.Authz constructor)
is set to fnmatchStrMatcher.
Is it ok to override self.authz.match to reStrMatcher in test method (after setUp)
 to also test regex matching?

MikeAlexeev: I think that there is a problem with authz testing. Please correct me if I'm wrong.

As Authz.assertUserAllowed method does not return which combination of rule and role succeeded
 the check we can't test that exact one rule succeeded but not some other.
For example if I want to check regex-based role check I need to add some rule with role pattern.
How can I be shure that exactly that rule succeeded or failed in test but not some other in
long list of rules?

I think that there are two solutions:
1) return combination of rule and role that succeeded the authz check
(it seems that return value is not checked anywhere in buildbot).
So we can compare it in test and make tests more reliable.
2) override list of rules and list of roles in test method to make tests more isolated and predictable.

Maybe that's why where is a bug when idefaultDeny=False gives access if access should not be given

tardyp: Testing regexmatcher vs regex matcher is not very relevant for me. We dont need to test all the combinations, we can just test that the regex matcher is working as expected, and then only test with fnmatchstrmatcher (or a fake matcher).

As for testing the appropriate usecase, I agree on solution 2. If we need to have a (or set of) test with a specific rule combination, we shall implement it like this.

MikeAlexeev:
I have added unit tests.

Maybe I misunderstood meaning of defaultDeny parameter.
I think that if check for rule fails and defaultDeny == True action is not allowed for user.
If check for rule fails and defaultDeny is False then check next rule.

But there is a description in the doc:
defaultDeny -- The role matcher algorithm will stop if this value is true, and if the endpoint matched.

Also in the example in the doc we have this comment:
# defaultDeny=False: if user does not have the admin role, we continue parsing rules

It is not obvious description, is it?

tardyp:
I think both what you say and the doc is true, and it is just a rephrase. What the doc is describing is the exact behaviour of the algorithm, what you are describing is more user centric rephrasing. Please feel free to add your rephrase as a clarification in the doc.

master/buildbot/test/unit/test_www_authz.py
```diff
@@ -104,6 +104,13 @@ def setUp(self):
                          buildrequestid=82, number=5),
         ])
 
+    def setAllowRules(self, allow_rules):
+        # we should add links to autz and master instances in each new rule
```
tardyp: authz

master/buildbot/test/unit/test_www_authz.py
```diff
@@ -140,3 +147,72 @@ def test_rebuildBuild(self):
         # not owner cannot rebuild
         yield self.assertUserForbidden("builds/13", "rebuild", {}, "eightuser")
         yield self.assertUserForbidden("buildrequests/82", "rebuild", {}, "eightuser")
+
+    @defer.inlineCallbacks
+    def test_fnmatchPatternRoleCheck(self):
+        # set defaultDeny to True so action is denied if no match
```
tardyp: it is the default anyway..

master/buildbot/test/unit/test_www_authz.py
```diff
@@ -140,3 +147,72 @@ def test_rebuildBuild(self):
...
+        allow_rules = [
+            AnyEndpointMatcher(role="not-exists1", defaultDeny=False),
+            AnyEndpointMatcher(role="not-exists2", defaultDeny=False),
+            AnyEndpointMatcher(role="not-exists3", defaultDeny=True)
+        ]
+
+        self.setAllowRules(allow_rules)
+
+        # check if action is denied and last check was exact against not-exist3
+        with self.assertRaisesRegexp(authz.Forbidden, '.*not-exists3.*'):
```
tardyp: assertRaisesRegex

MikeAlexeev:
I did typo fixes but leaved defaultDeny=True in tests where testing it's behavior to show more clearly 
testing conditions

master/buildbot/test/unit/test_www_authz.py
```diff
@@ -140,3 +147,72 @@ def test_rebuildBuild(self):
...
+        allow_rules = [
+                        AnyEndpointMatcher(role="[a,b]dmin?")
+        ]
+
+        self.setAllowRules(allow_rules)
+
+        yield self.assertUserAllowed("builds/13", "rebuild", {}, "homer")
+
+        # check if action is denied
+        with self.assertRaises(authz.Forbidden):
```
tardyp: this is not supported with tw14, please use assertRaisesRegex (which is monkey patched in buildbot)

MikeAlexeev: sorry for delay. I haven't much time now.
I will probably be able to fix it on monday

MikeAlexeev:
Hello.
I have fixed compatibility errors.
it was easier to use unicode2bytes util function, because b'' literal has other problems in
this case. For example "unsupported operand type(s) for %: 'bytes' and 'bytes'"
or "builtins.TypeError: unsupported operand type(s) for %: 'bytes' and 'str'"

Also I'v found that there were no bugs with defaultDeny. Problems with it appeared when I started to fix
pattern-based authz. So it is my mistake and I have renamed this PR.

MikeAlexeev:
Also I have a question about testing with both py3 and py2.
Maybe it should be better to create one virtualenv for each python version?
virtualenv target in Makefile can be modified for this purpose.
I did so and run trial in both virtualenvs

tardyp:
You can use make isort in order to fix the isort issues.
I think make py3virtualenv is a good idea.

I would prefer to avoid having virtualenv do both as some people may not have py2 or py3 installed.

MikeAlexeev:
Should I add "make py3virtualenv" in separate PR?
Is it ok to name it's virtualenv folder .venvpy3  ?
What if default python version is 3, so make py3virtualenv has no sense in this case.
Is it good to also add "make py2virtualenv"?

For example:
- make virtualenv  creates .venv  and uses default python version
- make py2virtualenv  creates .venvpy2 and explicitly uses python2 
- make py3virtualenv  creates .venvpy3 and explicitly uses python3

tardyp:
thanks!

tardyp:
@MikeAlexeev btw make virtualenv change proposal looks good to me

PR #3099, opened 2017-04-06
authz: fix AnyControlEndpointMatcher

noc0lour:

Fixes serious bug in the AnyControlEndpointMatcher and fixes the unit test. Please don't test against default values in unit tests. Otherwise we wouldn't need them :) real world values we need :)
In this case we always get a uppercase `GET` and match against a lowercase `get` which conveniently was our default value for action.

### DIFF
--- a/master/buildbot/www/authz/endpointmatchers.py
+++ b/master/buildbot/www/authz/endpointmatchers.py
@@ -109,8 +109,8 @@ class AnyControlEndpointMatcher(EndpointMatcherBase):
     def __init__(self, **kwargs):
         EndpointMatcherBase.__init__(self, **kwargs)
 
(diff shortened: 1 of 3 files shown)


master/buildbot/www/authz/endpointmatchers.py
```diff
@@ -109,8 +109,8 @@ class AnyControlEndpointMatcher(EndpointMatcherBase):
     def __init__(self, **kwargs):
         EndpointMatcherBase.__init__(self, **kwargs)
 
-    def match(self, ep, action="get", options=None):
-        if action != "get":
+    def match(self, ep, action="", options=None):
+        if action != "GET":
```
tardyp: lets implement that with action.lower() != "get"

PR #3100, opened 2017-04-07
Waterfall fixes

msylw:
Updates to make waterfall view work in current Chrome. I wasn't able to test this in other browsers, according to MDN should be a change for better.
- Zooming used deprecated attributes of KeyboardEvent, the old one is not working anymore with any current browser, the new one should be pretty well supported
- Fixed scoping issue in handler responsible for fetching additional data while scrolling

## Contributor Checklist:

 - no tests for waterfall view
* [X] I have created a file in the master/buildbot/newsfragment directory (and read the README.txt in that directory
 - no functionality changes


tardyp:
thank you very much!

PR #3103, opened 2017-04-07
virtual builders

tardyp:
see doc for description and purpose.
I need to experiment a bit more with bbtravis,to see how it works.

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -72,7 +72,7 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
                         collects our status
     """
 
-    VIRTUAL_BUILDERNAME_PROP = "virtual_buildername"
+    VIRTUAL_BUILDERNAME_PROP = "virtual_builder_name"
     VIRTUAL_BUILDERDESCRIPTION_PROP = "virtual_builder_description"
     VIRTUAL_BUILDERTAGS_PROP = "virtual_builder_tags"
     workdir = "build"
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -72,6 +72,9 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
                         collects our status
     """
 
+    VIRTUAL_BUILDERNAME_PROP = "virtual_buildername"
+    VIRTUAL_BUILDERDESCRIPTION_PROP = "virtual_builder_description"
+    VIRTUAL_BUILDERTAGS_PROP = "virtual_builder_tags"
     workdir = "build"
     build_status = None
     reason = "changes"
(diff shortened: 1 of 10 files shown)


noc0lour:
I am wondering if we can achieve the same by having a `display`/`status` property or something which can be defined in the builder configuration and can be populated in a trigger step. And the UI can then access/look for a display/status property in the REST API and add it to the description. 

With this way you wouldn't have to add a whole new class

tardyp:
I did think to resolve that problem solely in the UI, but this means you have to build UI specially for bb travis, which I'd like to avoid as much as possible. 
This does not sound right to say that bbtravis is not adapted to console view (or the other way around).

Having virtual builder also help to get fast access to all the builds from the same builder.
This will help for plotting or ETA calculations.

noc0lour:
I just looked at the actual implementation and it looks fairly simple and straightforward :+1:

master/buildbot/process/build.py
```diff
@@ -72,6 +72,9 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
                         collects our status
     """
 
+    VIRTUAL_BUILDERNAME_PROP = "virtual_buildername"
```
seankelly: The other two properties are of the form `virtual_builder_*` but this one isn't. Seems like that should all have the same prefix.

master/buildbot/steps/trigger.py
```diff
@@ -216,9 +216,14 @@ def addBuildUrls(self, rclist):
...
                 for builderid, br in iteritems(brids):
-                    builderDict = yield self.master.data.get(("builders", builderid))
                     builds = yield self.master.db.builds.getBuilds(buildrequestid=br)
+                    builderDict = yield self.master.data.get(("builders", builderid))
                     for build in builds:
+                        # When virtual builders are used, the builderid used for triggering
+                        # is not the same as the one that the build actually got
+                        if build['builderid'] != builderid:
+                            builderid = build['builderid']
+                            builderDict = yield self.master.data.get(("builders", builderid))
```
seankelly: Is it okay if this `builderDict` is used for other builds? My concern is if some builds have the differing `builderid` but others don't.

tardyp: if builderid changes, builderDict is then  recomputed.

tardyp:
@seankelly I hope I have fixed your concerns

seankelly:
Let me start with saying I think this is a good idea. A higher-level way to abstract things like the bbtravis builds is definitely needed.

I thought about it more and I wonder if using virtual builders, instead of say tags, is the right way. On one hand, I tend to only view the builder pages on my Buildbots and having a more useful builder organization would be great. On the other hand, I think having a different view that uses some metadata for categorizing the builds would be better.

Overall I lean toward merging this PR. I think there is the possibility of confusing between the normal builders and virtual builders, but maybe I'm over thinking it.

PR #3105, opened 2017-04-08
pre-install pbr and setuptools_scm

tardyp:
workaround for https://github.com/pypa/setuptools_scm/issues/164


tardyp:
fixed upstream

PR #3107, opened 2017-04-08
fix E741 on test_process_build

tardyp:
recent pep8 does not like l (the letter) as a variable because some fonts make it undistinguishable from 1 (the number).


tardyp:
@seankelly no, its just that I worked on this file as part of #3103 and my editor pointed a bunch of pep8 errors. I then did feel like putting that patch in #3103

PR #3108, opened 2017-04-10
console_view: fix broken UI

### LINKED ISSUE #3062: Console_view lines don't change position on expansion
If I expand a changeset in `console_view` the horizontal separation line between changes won't expand as well and then the UI looks weird.

tardyp:
Fixes #3062

Not sure if it was because of an angularJS upgrade, but console view was actually completely broken.

- changes were not displayed
- column were not aligned

before:
image

after:

image


tardyp:
new version, with lots of simplification.
We are back at a table design, which is much more simple to maintain.

image

www/console_view/src/styles/styles.less
```diff
@@ -1,197 +1,57 @@
...
-            .badge-status {
-              cursor: pointer;
-            }
-          }
-        }
-      }
+        text-align: center;
+        vertical-align: middle;
+        margin-top: 30px;
+        -webkit-transform: translateX(0px) rotate(-15deg);
```
seankelly: Are the vendor prefixed versions still needed? It appears everyone has [supported the standard version for quite some time](https://caniuse.com/#feat=transforms2d).

Stibbons:
+1

PR #3112, opened 2017-04-11
gerritchangesource: add gerritprojects option.

wainersm:
It adds the 'gerritprojects' to GerritChangeSource and
GerritEventLogPoller so that it can be set a  list of
projects which events should be handled. By default
events of any project is handled.

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -65,6 +65,11 @@ def _gerrit_user_to_author(props, username=u"unknown"):
     return username
 
 
+class _AnyGerritProject(object):
+    # Represents any gerrit project.
+    pass
+
+
 class GerritChangeSourceBase(base.ChangeSource):
 
     """This source will maintain a connection to gerrit ssh server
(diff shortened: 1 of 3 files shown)


tardyp:
Like I said in the associated issue, I would prefer that we tackle this problem in a way that would benefit the whole community rather than just the gerrit users.
I dont think it is actually more difficulty to reuse the ChangeFilter mechanism to implement what you need, i.e filtering at source.

tardyp:
Hi @wainersm. I think this is a good feature! When will you have time to work on it?

wainersm:
Hi @tardyp ! I've time to work on it. Indeed yesterday I was looking on how I could implement your suggestion. So soon I will send  another patch, but if I get blocked I let you know.

tardyp:
@wainersm don't hesitate to hang-on on IRC even before you are totally blocked!

wainersm:
@tardyp I was thinking about your suggestion. If I understood correctly, the idea is to filter out using ChangeFilter at scheduler. Won't it end up recording in the master database unneeded code changes?
Taking my patch as an example, the gerrit event isn't even added in the database.

tardyp:
@wainersm no, the idea is to have changesource understand change filter, and apply the filter just before calling addChange data api.
So that we don't spam the database with unneeded changes.

PR #3115, opened 2017-04-13
need more underline under the version number

tardyp:


master/setup.py
```diff
@@ -487,7 +487,7 @@ def define_plugin_entries(groups):
         'setuptools_trial',
         'isort',
         # spellcheck introduced in version 1.4.0
-        'pylint>=1.4.0',
+        'pylint<1.7.0',
```
Stibbons: I think version should be frozen for app. Libraries can have version range, but I tend to see more advantages for end application when dependencies are completely fixed.

PR #3117, opened 2017-04-14
forcescheduler: fix error when builder names is unicode

rjarry:
The `BuilderConfig.name` attribute [is forced to unicode](https://github.com/buildbot/buildbot/blob/master/master/buildbot/config.py#L1016), `ForceScheduler` should accept `unicode` as valid in the `builderNames` list.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/forcesched-unicode-builderNames.bugfix
+++ b/master/buildbot/newsfragments/forcesched-unicode-builderNames.bugfix
@@ -1,0 +1,1 @@
+Passing ``unicode`` ``builderNames`` to :bb:sched:`ForceScheduler` no longer causes an error.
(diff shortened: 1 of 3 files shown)


rutsky:
AFAIK builder names are part of Data API, @tardyp is Unicode properly handled there (and in web UI)?

tardyp:
@rutsky data API internal representation for strings should always be unicode. The REST api is converting those strings to utf-8, and web UI should handle that correctly.
Then, all inputs (e.g master.cfg) shall handle unicode or utf-8 byte strings, for backward compatibility.
Here we only handle byte strings, which is a mistake.
@rjarry thanks for the fix

PR #3121, opened 2017-04-18
Make Gerrit Change Source and Reporter PATH Aware

jammerful:
Without these changes these processes would not look
into PATH for the ssh binary, which is a problem
for operating systems that don't store the
ssh binary in the usual locations.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -19,6 +19,7 @@ from future.utils import iteritems
 
 import datetime
 import json
+import os
 
 from twisted.internet import defer
 from twisted.internet import reactor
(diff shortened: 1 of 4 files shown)


master/buildbot/changes/gerritchangesource.py
```diff
@@ -313,7 +314,8 @@ def startStreamProcess(self):
             args = args + ['-i', self.identity_file]
         self.process = reactor.spawnProcess(
             self.LocalPP(self), "ssh",
-            ["ssh"] + args + ["gerrit", "stream-events"])
+            ["ssh"] + args + ["gerrit", "stream-events"],
+            env={'PATH': os.environ.get('PATH', '')})
```
tardyp: looks like this is a weirdness of twisted. I was sure spawnProcess would inherit parent process environment.
but its not, according to doc, unless you *set* env=None
http://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IReactorProcess.html
I think it would be more consistent with behaviour of the worker if we rather set env=None

There might be more environement variable than PATH that ssh might want to use

seankelly: Agreed. I think that is what people would expect to happen.

jammerful: Great. Thanks for catching that, I will set env=None in my PR.

jammerful:
@tardyp Looks good now?
I did an end to end test on my non standard Linux OS that has ssh installed in an unusual path, and everything worked as planned.

tardyp:
@jammerful perfect. just in time for 0.9.6!

PR #3122, opened 2017-04-18
Waterfall view

tardyp:
image


seankelly:
I'm only barely familiar with the frontend and coffeescript, but the changes seem good.

PR #3133, opened 2017-04-20
add tweaks to allow buildbot to start and stop nicely on win32

dragon512:
This adds some small tweak to allow the same 
```python
buildbot start
buildbot stop
buildbot restart
```
commands to work on windows as it does on posix systems

And especially:

## Contributor Checklist:

### DIFF
--- a/master/buildbot/scripts/start.py
+++ b/master/buildbot/scripts/start.py
@@ -114,8 +114,12 @@ def launch(config):
             "--python=buildbot.tac"]
 
     # ProcessProtocol just ignores all output
-    reactor.spawnProcess(
+    proc = reactor.spawnProcess(
         protocol.ProcessProtocol(), sys.executable, argv, env=os.environ)
+
+    if platformType == "win32":
+        with open("twistd.pid", "w") as pidfile:
+            pidfile.write("{0}".format(proc.pid))
 
 
 def start(config):
--- a/master/buildbot/scripts/stop.py
+++ b/master/buildbot/scripts/stop.py
@@ -22,6 +22,8 @@ import os
 import signal
 import time
 
+from twisted.python.runtime import platformType
+
 from buildbot.scripts import base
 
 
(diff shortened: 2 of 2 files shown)


dragon512:
I should note this does not support the --clean option. As a different way to communicate SIGUSR1 signal is needed as Windows does not support this.

.coveragerc
```diff
@@ -1 +0,0 @@
-common/coveragerc
```
tardyp: why did you remove that link?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "yapf"
```
tardyp: I would prefer we do not commit IDE specific code here. Please add it in gitignore

master/docs/manual/mydashboard.html
```diff
@@ -1 +0,0 @@
-../../../smokes/templates/mydashboard.html
```
tardyp: does windows not support symlinks?
Please do not remove them!
We need those symlinks so that we can test the mydashboard in the smoke tests

.vscode/settings.json
dragon512: Ya.. that was a mistake on my end. Let me correct.

master/docs/manual/mydashboard.html
dragon512: It does.. and i have them enabled for my logon. I am not sure what happened here.

dragon512:
Fix those issue... do you have an concerns with updating the .gitignore file to ignore the .vscode directory?

seankelly:
Editor-specific ignores should be in the person's own .gitignore file. That way it will apply to everything they work on automatically.

dragon512:
Well I don't see a way to test it here... you need to add a windows container or something

Like I stated --clean will fail. I was thinking I could have a temp be created that the buildbot could see instead of the SIGUSER1 to have the clean shut logic work.

PR #3139, opened 2017-04-21
master: replace most str occurences by string_types

rjarry:
When configuring some services in `master.cfg` with:

```python
from __future__ import unicode_literals
from buildbot.plugins.db import steps
b = steps.ShellCommand(
    name='do stuff',
    command=['stuff'],
)
```

there are errors with python 2:

```
BuildStep name must be a string: u'do stuff'
```

`BuildStep` is only an example, there are other constructors that do not accept unicode strings.

Since almost all internal API work on unicode strings, replace most checks that a variable is a string by testing if it is an instance of `string_types` (from `future.utils`).

## Contributor Checklist:

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -477,7 +477,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
         protocols = config_dict.get('protocols', {})
         if isinstance(protocols, dict):
             for proto, options in iteritems(protocols):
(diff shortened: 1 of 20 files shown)


rjarry:
Ok I understand. I'll have a look at maybe enhancing the tests and force convert to unicode.

p12tic:
Removing labels as we no longer need this in master.

PR #3140, opened 2017-04-21
introduce logchunk deletion db API

tardyp:
The largest part of buildbot db is storing the logchunks.
for the first version of the Janitor, we will be deleting only the old logchunks to keep as much metadata as possible

build/steps/logs will be managed in a second step

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -345,6 +345,39 @@ class LogsConnectorComponent(base.DBConnectorComponent):
         saved = yield self.db.pool.do(thdcompressLog)
         defer.returnValue(saved)
 
+    def deleteOldLogChunks(self, older_than_timestamp):
+        def thddeleteOldLogs(conn):
+            model = self.db.model
+            res = conn.execute("SELECT count(*) from `logchunks`")
+            count1 = res.fetchone()[0]
+            res.close()
+            # join the step and log tables
+            from_clause = model.logs.join(model.steps, model.steps.c.id == model.logs.c.stepid)
+
+            # query all logs older than timestamps
+            q = sa.select([model.logs.c.id]).select_from(from_clause).where(
+                model.steps.c.started_at < older_than_timestamp)
+
+            # update their types
+            res = conn.execute(
+                model.logs.update()
+                .where(model.logs.c.id.in_(q))
+                .values(type='d')
+            )
+            res.close()
+
+            # delete their logchunks
+            res = conn.execute(
+                model.logchunks.delete()
+                .where(model.logchunks.c.logid.in_(q))
+            )
+            res.close()
+            res = conn.execute("SELECT count(*) from `logchunks`")
+            count2 = res.fetchone()[0]
+            res.close()
+            return count1 - count2
+        return self.db.pool.do(thddeleteOldLogs)
+
     def _logdictFromRow(self, row):
         rv = dict(row)
         rv['complete'] = bool(rv['complete'])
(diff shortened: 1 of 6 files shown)


master/docs/developer/database.rst
```diff
@@ -550,6 +550,16 @@ logs
         It should only be called for finished logs.
         This method may take some time to complete.
 
+    .. py:method:: deleteOldLogChunks(older_than_timestamp)
+
+        :param integer older_than_timestamp: the logs whose step's ``started_at`` is older than ``older_than_timestamp`` will be deleted.
```
seankelly: Should it be steps or builds? Seems like it could be confusing if the first part of a build has its logs deleted but the second half doesn.t

master/buildbot/db/logs.py
```diff
@@ -345,6 +345,40 @@ def thdcompressLog(conn):
...
+                model.logs.update()
+                .where(model.logs.c.stepid.in_(
+                    sa.select([model.steps.c.id])
+                    .where(model.steps.c.started_at < older_than_timestamp)))
+                .values(type='d')
+            )
+            res.close()
+
+            # query all logs older than timestamps
+            q = sa.select([model.logs.c.id])
```
seankelly: My SQLAlchemy knowledge is lacking, but this partial query seems somewhat redundant with the previous query. The update sets all old logs to have a type of 'd'. This query seems to then set up the parameters to find those logs again. Why not select all logs with a 'd' type?

master/buildbot/db/logs.py
```diff
@@ -345,6 +345,40 @@ def thdcompressLog(conn):
...
+            q = q.select_from(model.logs.join(model.steps, model.steps.c.id == model.logs.c.stepid))
+            q = q.where(model.steps.c.started_at < older_than_timestamp)
+
+            # delete their logchunks
+            res = conn.execute(
+                model.logchunks.delete()
+                .where(model.logchunks.c.logid.in_(q))
```
seankelly: Do you have an idea on whether this will be slow on the first time when clearing a lot of log chunks? I've never pictured `IN` as being particularly fast. This is executed in a thread so I think it won't block the rest of the process, but I wonder about the database. Checking the Postgres docs and looks to be fine there so I would think MySQL and sqlite will also be fine.

master/docs/developer/database.rst
tardyp: I was trying to simplify the sql request. i don't think this corner case is worth the additional join. We can make it clear in the UI why this step log was deleted

master/buildbot/db/logs.py
tardyp: I think this should not change any thing unless we add an index to either column (and we probably should)

master/buildbot/db/logs.py
tardyp: Yeah this is a great question. I would have expected the delete to be doable with a join, but it looks like this is not possible with SQL.

tardyp: I did the test on copy of nine.bb.net (sqlite3) database with 419982 logchunks.
deleting all logchunks
```
DELETE FROM logchunks WHERE logchunks.logid IN (SELECT logs.id
   ...>         FROM logs JOIN steps ON steps.id = logs.stepid
   ...>         WHERE steps.started_at < 1493027316)
   ...> ;
sqlite> vacuum;
```
it takes ~10s on first run, and <1s on next run
db goes from 1.3G to 279M

I think this good enough numbers to avoid doing any more optimization

tardyp: I did the fix of doing the second query only for logs.type='d' see commit comment for 420650897609979724c889f062e4afdc66c916ce for details

PR #3142, opened 2017-04-21
[RFC] janitorConfigurator (replacing logHorizon)

tardyp:
This is highly untested for now, but this is a change of the horizon interface I intend to introduce.

- old interface is number based, which will be quite painful to write efficient sql query.
- I personally prefer the timedelta interface, which I find more adapted to the CI job. I'd like to find some feedback on this if ppl agree or not according to their experience.

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -77,6 +77,8 @@ _errors = None
 
 DEFAULT_DB_URL = 'sqlite:///state.sqlite'
 
+RESERVED_UNDERSCORE_NAMES = ["__Janitor"]
+
 
 def error(error, always_raise=False):
     if _errors is not None and not always_raise:
@@ -690,7 +692,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
 
         for worker in workers:
             if not interfaces.IWorker.providedBy(worker):
-                msg = "{0} must be a list of Worker instances".format(conf_key)
+                msg = "{} must be a list of Worker instances but there is {!r}".format(conf_key, worker)
                 error(msg)
                 return False
 
@@ -925,9 +927,9 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
             s.checkConfig(self.status)
 
     def check_horizons(self):
-        if self.logHorizon is not None or self.buildHorizon is not None:
-            log.msg("NOTE: `buildHorizon` and `logHorizon` are currently ignored "
-                    "(see http://trac.buildbot.net/ticket/3572)")
+        if self.buildHorizon is not None:
+            log.msg("NOTE: `buildHorizon` and `logHorizon` are deprecated and ignored "
+                    "They are replaced by util.JanitorConfigurator")
         if self.logHorizon is not None and self.buildHorizon is not None:
             if self.logHorizon > self.buildHorizon:
                 error("logHorizon must be less than or equal to buildHorizon")
@@ -1009,7 +1011,7 @@ class BuilderConfig(util_config.ConfiguredMixin, WorkerAPICompatMixin):
         if not name or type(name) not in (bytes, text_type):
             error("builder's name is required")
             name = '<unknown>'
-        elif name[0] == '_':
+        elif name[0] == '_' and name not in RESERVED_UNDERSCORE_NAMES:
             error(
                 "builder names must not start with an underscore: '%s'" % name)
         try:
@@ -1039,7 +1041,7 @@ class BuilderConfig(util_config.ConfiguredMixin, WorkerAPICompatMixin):
 
         if workername:
             if not isinstance(workername, str):
-                error("builder '%s': workername must be a string" % (name,))
+                error("builder '%s': workername must be a string but it is %r" % (name, workername))
             workernames = workernames + [workername]
         if not workernames:
             error("builder '%s': at least one workername is required" %
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -201,8 +201,6 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
         self.titleURL = 'http://buildbot.net'
         self.buildbotURL = 'http://localhost:8080/'
         self.changeHorizon = None
-        self.logHorizon = None
-        self.buildHorizon = None
         self.logCompressionLimit = 4 * 1024
         self.logCompressionMethod = 'gz'
         self.logEncoding = 'utf-8'
@@ -262,6 +260,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
         "caches",
         "change_source",
         "codebaseGenerator",
+        "configurators"
         "changeCacheSize",
         "changeHorizon",
         'db',
@@ -319,6 +318,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
 
     @classmethod
     def loadFromDict(cls, config_dict, filename):
+        # warning, all of this is loaded from a thread
         global _errors
         _errors = errors = ConfigErrors()
 
@@ -338,6 +338,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
 
         # and defer the rest to sub-functions, for code clarity
         try:
+            config.run_configurators(filename, config_dict)
             config.load_global(filename, config_dict)
             config.load_validation(filename, config_dict)
             config.load_db(filename, config_dict)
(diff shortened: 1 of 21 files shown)


rjarry:
This requires having a `__janitor` builder displayed among the other builders. This looks more like an internal feature that shouldn't interfere with the normal building process.

Would that be possible to cleanup old stuff when logs are actually saved to db?

What do you think?

tardyp:
@rjarry the motivation behind this is that if anything goes wrong, there is a builder to look at.
if there is an exception, the builder will fail in exception and be visible in the dashboards.
This first version is quite simple, but the idea is to add more complexity overtime with more custom steps.
- delete old builds
- delete old builders
- delete old anything actually

I added '__' in the name in order to make it clear that this is internal, and the worker used is also internal so it should not interfer with anything.
Also, having a builder makes it straightforward to support multimaster.
If implemented as an internal service, you have to make sure that the service does not trigger on several masters at the same time. The concept of buildrequest is exactly doing this job.

tardyp:
I have another PoC of a Janitor service, but I abandonned it because I found it too complicated:
https://github.com/tardyp/buildbot/blob/janitor/master/buildbot/process/janitor.py

rjarry:
ok I understand. well, your draft looks quite simple and that is positive.

would it be possible to display the actual disk space that was cleaned up?

master/docs/manual/cfg-global.rst
```diff
@@ -276,27 +276,39 @@ Data Lifetime
...
-    c['changeHorizon'] = 200
-    c['buildHorizon'] = 100
-    c['logHorizon'] = 40
-    c['buildCacheSize'] = 15
+    from buildbot.plugins import util
+    from datetime import timedelta
+
+    # at the end of your master.cfg
+    # configure a janitor which will delete all logs older than one month, and will run on sundays at noon
+    util.JanitorConfigurator(c,
```
seankelly: The name Janitor implies cleaning things but this is targeted toward cleaning logs. Either the `JanitorConfigurator` will gain a lot of options or the name needs to be a bit more precise. Maybe `LogJanitorConfigurator` or similar?

seankelly: And, this applies to all logs. I imagine people would like some logs to be kept longer than others.

master/docs/manual/cfg-global.rst
```diff
@@ -276,27 +276,39 @@ Data Lifetime
...
 
 Buildbot stores historical information in its database.
 In a large installation, these can quickly consume disk space, yet in many cases developers never consult this historical information.
 
-The :bb:cfg:`changeHorizon` key determines how many changes the master will keep a record of.
-One place these changes are displayed is on the waterfall page.
-This parameter defaults to 0, which means keep all changes indefinitely.
+Parameters for :bb:cfg:`JanitorConfigurator` are:
+
+``c``
```
seankelly: This seems like a clunky requirement. There can only be one `BuildmasterConfig` dict but this is requiring people to pass it to the Janitor.

master/docs/manual/installation/nine-upgrade.rst
```diff
@@ -197,6 +197,18 @@ Build History
...
+-------------
+
+Buildbot Nine data being implemented fully in an sql database, the ``buildHorizon`` feature had to be reworked.
+Instead of being number-of-things based, it is now time based.
+This makes more sense from a user perspective but makes it harder to predict the database average size.
+Please be careful to provision enough disk space for your database.
+
+The old ``c['logHorizon']`` way of configuring is not supported anymore.
+See :bb:cfg:`JanitorConfigurator` to learn how to configure.
+A new ``__Janitor`` builder will be created to help keeping an eye on the cleanup activities.
```
seankelly: "...to help keep* an eye on the cleanup activities."

master/docs/manual/cfg-global.rst
tardyp: Indeed, the goal is to extend it with more features.

At the moment I guess we need the bare minimum to not explode your hdd, and we will see if people ask for more.

master/docs/manual/cfg-global.rst
tardyp: Indeed.. :-}  it is the same for all configurators (like buildbot_travis)

There is no other way in the master.cfg context to have the BuildmasterConfig than to let it be passed by the user.

tardyp: Another option is to add another 'configurators' global params in the global namespace
```python
c['configurators'] = [ JanitorConfigurator(...) ]
```

This would mean the configurators will run on the main reactor thread, which could be useful

seankelly: Adding another global params key seems like it would fit in well with the rest of the Buildbot configuration. Everything else uses that except for configurators.

I can't comment on ramifications of running on the main reactor thread.

tardyp:
@seankelly I think it is good to go now. see ec761b4 for the change about moving it to c['configurators']

Eventually, it is not run in the master thread so the difference is minimal, and UX is imo better.
We can then move buildbot_travis to this new method.

master/buildbot/interfaces.py
```diff
@@ -1095,3 +1095,16 @@ def json():
...
                        "http status code of the request's response (e.g 200)")
+
+
+class IConfigurator(Interface):
+
+    def configure(config_dict):
+        """
+        Alter the buildbot config_dict, as defined in master.cfg
+
+        like the master.cfg, this is run from the a thread, so this can block, but this can't
```
seankelly: "this is run from the a thread..."
Which thread?

master/docs/developer/cls-iconfigurator.rst
```diff
@@ -0,0 +1,16 @@
...
+.. class:: buildbot.interfaces.IConfigurator::
+
+    A configurator is an object which configures several components of Buildbot in a coherent manner.
+    This can be used to implement higher level configuration tools.
+
+    .. method:: configure(config_dict)
+
+       Alter the buildbot ``config_dict``, as defined in master.cfg
+
+       like the master.cfg, this is run from the a thread, so this can block, but this can't
```
seankelly: Same issue here identifying the thread.

master/docs/manual/cfg-configurators.rst
```diff
@@ -0,0 +1,43 @@
+.. bb:cfg:: configurators
+
+Configurators
+-------------
+
+For advanced users or plugins writers, the 'configurators' key is available, and holds a list of :py:class:`buildbot.interfaces.IConfigurator`.
+Configurators will run after the master.cfg has been processed, and will modify the config dictionary.
+Configurator implementers should make sure that they are inter operable with each other, which means carefully modify the config to avoid overriding a setting already made by the user or by another configurator.
```
seankelly: Change "modify" to "modifying"

master/docs/manual/cfg-configurators.rst
```diff
@@ -0,0 +1,43 @@
...
+.. bb:configurator:: JanitorConfigurator
+
+JanitorConfigurator
+~~~~~~~~~~~~~~~~~~~
+
+Buildbot stores historical information in its database.
+In a large installation, these can quickly consume disk space, yet in many cases developers never consult this historical information.
+
+:bb:configurator::`JanitorConfigurator` creates a builder and :bb:sched:`Nightly` scheduler which will regularly remove old information.
+At the moment it only supports cleaning of logs, but it will contain more feature as we implement them.
```
seankelly: Change "feature" to "features".

seankelly:
Failures all appear to be in WAMP.

master/buildbot/config.py
```diff
@@ -924,14 +928,6 @@ def check_status(self):
         for s in self.status:
             s.checkConfig(self.status)
 
-    def check_horizons(self):
-        if self.logHorizon is not None or self.buildHorizon is not None:
-            log.msg("NOTE: `buildHorizon` and `logHorizon` are currently ignored "
-                    "(see http://trac.buildbot.net/ticket/3572)")
```
seankelly: This makes me wonder if there are any other trac references in the code.

master/docs/manual/cfg-configurators.rst
```diff
@@ -0,0 +1,43 @@
+.. bb:cfg:: configurators
+
+Configurators
+-------------
+
+For advanced users or plugins writers, the 'configurators' key is available, and holds a list of :py:class:`buildbot.interfaces.IConfigurator`.
+Configurators will run after the master.cfg has been processed, and will modify the config dictionary.
+Configurator implementers should make sure that they are inter operable with each other, which means carefully modifying the config to avoid overriding a setting already made by the user or by another configurator.
```
seankelly: Interoperable is one word.

PR #3144, opened 2017-04-24
unicode literrals

tardyp:
based on irc discussion with @rjarry

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -15,6 +15,7 @@
 
 from __future__ import absolute_import
 from __future__ import print_function
+from __future__ import unicode_literals
 from future.builtins import range
 from future.utils import itervalues
 
@@ -121,7 +122,8 @@ class LogsConnectorComponent(base.DBConnectorComponent):
             rv = []
             for row in conn.execute(q):
                 # Retrieve associated "reader" and extract the data
-                # Note that row.content is stored as bytes, and our caller expects unicode
+                # Note that row.content is stored as bytes, and our caller
+                # expects unicode
                 data = self.COMPRESSION_BYID[
                     row.compressed]["read"](row.content)
                 content = data.decode('utf-8')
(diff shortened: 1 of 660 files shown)


master/buildbot/db/logs.py
```diff
@@ -195,7 +197,8 @@ def thdAppendLog(self, conn, logid, content):
         # check for trailing newline and strip it for storage -- chunks omit
         # the trailing newline
         assert content[-1] == u'\n'
-        # Note that row.content is stored as bytes, and our caller is sending unicode
+        # Note that row.content is stored as bytes, and our caller is sending
```
seankelly: Same orphan issue here but not as glaring at least because of the longer word.

master/buildbot/db/migrate/versions/047_cascading_deletes.py
```diff
@@ -44,7 +46,8 @@ def upgrade(migrate_engine):
             for c in fk.elements:
                 if c.column in keys:
                     # migrate.xx.ForeignKeyConstraint is changing the model so initializing here
-                    # would break the iteration (Set changed size during iteration)
+                    # would break the iteration (Set changed size during
```
seankelly: Unusual to wrap this line but not the previous line which is longer.

master/buildbot/scripts/runner.py
```diff
@@ -663,7 +664,8 @@ class CleanupDBOptions(base.BasedirMixin, base.SubcommandOptions):
     subcommandFunction = "buildbot.scripts.cleanupdb.cleanupDatabase"
     optFlags = [
         ["quiet", "q", "Do not emit the commands being run"],
-        ["force", "f", "Force log recompression (useful when changing compression algorithm)"],
+        ["force", "f",
+            "Force log recompression (useful when changing compression algorithm)"],
```
seankelly: Wouldn't this need to be lined up with the `[`?

master/buildbot/steps/trigger.py
```diff
@@ -220,7 +224,8 @@ def addBuildUrls(self, rclist):
                     for build in builds:
                         builderid = build['builderid']
                         # When virtual builders are used, the builderid used for triggering
-                        # is not the same as the one that the build actually got
+                        # is not the same as the one that the build actually
```
seankelly: Same strange wrapping here too.

Issue #3151, opened 2017-04-26
buildRequestCollapsing not working in 0.9.6 as expected 


sphet:
I am trying to get build request merging to work in 0.9 - we are using it effectively in 0.8.6. The documentation states: 

> Sourcestamps are compatible if all of the below conditions are met:
> 
> Their codebase, branch, project, and repository attributes match exactly
> Neither source stamp has a patch (e.g., from a try scheduler)
> Either both source stamps are associated with changes, or neither are associated with changes but they have matching revisions.

In my case I have a long-running builder - it takes five minutes to complete - and I submit the following sequence of changes to Perforce:

edit & submit fileA.txt at 10:00AM ; CL461078 --> builder starts
edit & submit fileB.txt at 10:02AM ; CL461079 --> builder is busy, submission is pending
edit & submit fileC.txt at 10:03AM ; CL461080 --> builder is busy, submission is pending

Once CL001 is finished building, I would expect CL002 and CL003 to be collapsed into one build request. This is not happening. I have configued the builder with 'collapseRequests=True'. The debug output shown for the source stamps is:

[{"branch":null,"codebase":"","created_at":1493227979,"patch":null,"project":"","repository":"","revision":"461078","ssid":427}]
[{"branch":null,"codebase":"","created_at":1493228009,"patch":null,"project":"","repository":"","revision":"461079","ssid":428}]
[{"branch":null,"codebase":"","created_at":1493228039,"patch":null,"project":"","repository":"","revision":"461080","ssid":429}]

Looking at the source code I don't really see any code that would handle the 'both source stamps are associated with changes', as certainly in the canBeCollapsed function in buildRequest.pyL254 the collapse is failing because the revisions are different.

Is my reading of the documentation wrong? In 8.6 it seemed this worked - stacked up pending requests (regardless of  revision number) would all be built once the builder is free, not atomically. How would we modify the code to handle this?

Reviewing the source code to 0.8.x I see that in sourcestamp.py the canBeMergedWith reads, in addition to other things:

        if self.changes and other.changes:
            return True
        elif self.changes and not other.changes:
            return False  # we're using changes, they aren't
        elif not self.changes and other.changes:
            return False  # they're using changes, we aren't

I don't see parallel code in the 0.9 branch.

tardyp:
Indeed, problem comes from this very old change from the begining of nine refactor:
[...]

PR #3152, opened 2017-04-26
Added logic to check the changes of the sourcestamps for collapsing build sets

sphet:
I modified the collapseRequest logic to include an comparison the changes in the source stamps. This matches the logic described in the documentation, and the behaviour of 0.8.x branch

This is for issue https://github.com/buildbot/buildbot/issues/3151

Sorry if I did the pull request wrong this is the first time i have used github

### DIFF
--- a/master/buildbot/process/buildrequest.py
+++ b/master/buildbot/process/buildrequest.py
@@ -274,9 +274,6 @@ class BuildRequest(object):
 
         for c, selfSS in iteritems(selfSources):
             otherSS = otherSources[c]
-            if selfSS['revision'] != otherSS['revision']:
-                defer.returnValue(False)
-                return
             if selfSS['repository'] != otherSS['repository']:
                 defer.returnValue(False)
                 return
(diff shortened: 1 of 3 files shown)


master/buildbot/process/buildrequest.py
```diff
@@ -290,6 +287,22 @@ def canBeCollapsed(master, br1, br2):
             if selfSS['patch'] or otherSS['patch']:
                 defer.returnValue(False)
                 return
+            # get changes & compare
+            selfChanges = yield master.data.get(('sourcestamps', selfSS['ssid'], 'changes'))
```
tardyp: I think this logic will not work because the `if selfSS['revision'] != otherSS['revision']:` will happen before. You should move the revision check after the change check

master/buildbot/process/buildrequest.py
```diff
@@ -290,6 +287,22 @@ def canBeCollapsed(master, br1, br2):
...
+            if selfChanges and otherChanges:
+                continue
+            elif selfChanges and not otherChanges:
+                defer.returnValue(False)
+                return
+            elif not selfChanges and otherChanges:
+                defer.returnValue(False)
+                return
+            # else check revisions
+            elif selfSS['repository'] != otherSS['repository']:
```
tardyp: looks like you moved the wrong code :)

sphet:
I will write a unit test if you can point me to the right place to create fake 'changes', as well as how to run the tests locally!

tardyp:
@sphet google for "buildbot unit tests" will point you to this doc: http://docs.buildbot.net/latest/developer/tests.html
It's my job to have this straightforward and easy. Please complain if it is not!


you will need to add tests in this file: https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/test_process_buildrequest.py#L31
I dont think you need FakeBuildRequest object, just use BuildRequest.fromBrDict() in your test to build your test cases.

tardyp:
updated your branch to add unit tests and relnotes

PR #3157, opened 2017-04-27
steps: make SetProperty property renderable

noc0lour:
there is no reason we can't have value a renderable, make setting properties more flexible.

### DIFF
--- a/master/buildbot/newsfragments/set_property.feature
+++ b/master/buildbot/newsfragments/set_property.feature
@@ -1,1 +1,1 @@
-``value`` argument in SetPropery is now renderable.
+``property`` argument in SetPropery is now renderable.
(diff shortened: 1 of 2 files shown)


master/buildbot/newsfragments/set_property.feature
```diff
@@ -0,0 +1 @@
+``value`` argument in SetPropery is now renderable.
```
seankelly: `value` was already renderable. I think you mean `property`.

noc0lour: Yeah, true, I'll fix that.

tardyp:
@noc0lour I've fixed the newsfragment myself

noc0lour:
Thanks. I had that lying around somewhere but forgot to push it, apparently

tardyp:
well, it is just two clicks away using github UI.

PR #3172, opened 2017-05-03
transfer: fix workersrcs argument parsing in MultipleFileUpload

### LINKED ISSUE #3167: Giving a Property as workersrcs to MultipleFileUpload breaks
This issue was introduced by 4173daceb97c6e2583e7be9bc63ed4dcf7ba8f2e.

I give a Property parameter to MultipleFileUploads, the property resolves to a list.

Due to the mentioned commit, The value is now a list of list, and this breaks.

``` pytb
2017-05-02 12:55:50+0200 [-] MultipleFileUpload started, from worker [[u'build/filename']] to master '/home/buildsystem/master/public_html/destination'
2017-05-02 12:55:50+0200 [-] BuildStep.failed; traceback follows
          File "/home/buildsystem/bb-master/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1357, in gotResult
          File "/home/buildsystem/bb-master/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1301, in _inlineCallbacks
            result = g.send(result)
          File "/home/buildsystem/bb-master/local/lib/python2.7/site-packages/buildbot/process/buildstep.py", line 547, in startStep
            self.results = yield self.run()
          File "/home/buildsystem/bb-master/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1447, in unwindGenerator
            return _inlineCallbacks(None, gen, Deferred())
          File "/home/buildsystem/bb-master/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1301, in _inlineCallbacks
            result = g.send(result)
          File "/home/buildsystem/bb-master/local/lib/python2.7/site-packages/buildbot/process/buildstep.py", line 677, in run
            results = yield self._start_deferred
        exceptions.AttributeError: 'list' object has no attribute 'rfind'
```

noc0lour:
## Contributor Checklist:

This should fix #3167.
@benallard Can you add a unit test which represents your usecase/missing corner cases?

### DIFF
--- a/master/docs/manual/cfg-buildsteps.rst
+++ b/master/docs/manual/cfg-buildsteps.rst
@@ -2194,7 +2194,7 @@ Transferring Multiple Files At Once
 
 In addition to the :bb:step:`FileUpload` and :bb:step:`DirectoryUpload` steps there is the :bb:step:`MultipleFileUpload` step for uploading a bunch of files (and directories) in a single :class:`BuildStep`.
 The step supports all arguments that are supported by :bb:step:`FileUpload` and :bb:step:`DirectoryUpload`, but instead of a the single ``workersrc`` parameter it takes a (plural) ``workersrcs`` parameter.
-This parameter should either be a list, or something that can be rendered as a list.
+This parameter should either be a list, something that can be rendered as a list or a string which will be converted to a list.
 Additionally it supports the ``glob`` parameter if this parameter is set to ``True`` all arguments in ``workersrcs`` will be parsed through ``glob`` and the results will be uploaded to ``masterdest``.::
 
     from buildbot.plugins import steps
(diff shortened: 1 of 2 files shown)


tardyp:
As pointed out by @benallard, Please add a documentation snippet explaining that multifile upload accept both list or string for workersrcs

noc0lour:
If we only want to support lists as argument I'm fine with that as well, then we would have to throw an error after checking if it is a list.

tardyp:
I think having a single string makes sense for glob, and the code is simple enough and tested. We just need it to be documented

noc0lour:
Sure, I'll look at it later today ;)

noc0lour:
Added a sentence to the docs (:

master/docs/manual/cfg-buildsteps.rst
```diff
@@ -2204,7 +2204,7 @@ Additionally it supports the ``glob`` parameter if this parameter is set to ``Tr
     f.addStep(steps.MultipleFileUpload(workersrcs=["docs", "test-results.html"],
                                        masterdest="/usr/share/nginx/www/",
                                        url="~buildbot"))
-
+..
```
tardyp: looks like those two dots are not wanted

noc0lour: Yep, now (:

PR #3185, opened 2017-05-11
Obfuscated secrets

protatremy:
Completing @tardyp pull request [WIP] obfuscate secrets #3068

buildbot-worker has this support for obfuscating secrets from the list of commands but:

this works only for shell=False, list kinds ShellCommands
this works only for shell commands
this does not prevent secrets to leak in the logs
With this method, we prevent secrets to leak anywhere in the db

This method does not prevent secrets to leak in the twisted log of the worker
## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_interpolate_secrets.py
+++ b/master/buildbot/test/unit/test_interpolate_secrets.py
@@ -55,3 +55,23 @@ class TestInterpolateSecretsNoService(unittest.TestCase, ConfigErrorsMixin):
         self.assertFailure(self.build.render(command), defer.FirstError)
         self.flushLoggedErrors(defer.FirstError)
         self.flushLoggedErrors(KeyError)
+
+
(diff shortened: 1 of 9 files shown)


master/buildbot/test/unit/test_interpolate_secrets.py
```diff
@@ -55,3 +55,23 @@ def test_secret(self):
...
+        fakeStorageService.reconfigService(secretdict={"foo": "bar",
+                                                       "other": "value"})
+        self.secretsrv = SecretManager()
+        self.secretsrv.services = [fakeStorageService]
+        self.secretsrv.setServiceParent(self.master)
+        self.build = FakeBuildWithMaster(self.master)
+
+    @defer.inlineCallbacks
+    def test_secret(self):
+        command = Interpolate("echo %(secrets:foo)s")
```
tardyp: this should be %(secret:foo)s (no plural secret)

PR #3190, opened 2017-05-15
optional VENV_PY_VERSION arg for virtualenv make target

MikeAlexeev:
Hello.

I added VENV_PY_VERSION option to **make virtualenv**  target.
It can be invoked so:
**make virtualenv VENV_PY_VERSION=python3.4**

There are 2 cases:
1) If no version is provided (simply make virtualenv) current default python version is used
Folder is called .venv 
2) If version is provided than that version is used.
Folder is called like .venvpython3.4

So we can create several venv-s for different python versions to test code in several versions.
Also python version can be set for  **make release**   (I haven't tried this).

Maybe other names are better? For example simply PY_VERSION instead of VENV_PY_VERSION
and .venv_python3.4 instead of .venvpython3.4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ DOCKERBUILD := docker build --build-arg http_proxy=$$http_proxy --build-arg http
 
 PIP?=pip
 
+VENV_NAME:=.venv$(VENV_PY_VERSION)
+VENV_PY_VERSION?=python
+
 # build rst documentation
 docs:
 	$(MAKE) -C master/docs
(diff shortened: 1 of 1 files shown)


MikeAlexeev:
Sorry, it seems that i'v taken old version of Makefile. 
I will fix it soon

Makefile
```diff
@@ -5,6 +5,9 @@ DOCKERBUILD := docker build --build-arg http_proxy=$$http_proxy --build-arg http
 
 PIP?=pip
 
+VENV_NAME:=.venv$(VENV_PY_VERSION)
```
tardyp: problem with this is that the de-facto standard for venv is .venv, which you can configure your editor to automatically find.
So I would prefer to have default python venv use ".venv"

MikeAlexeev: For defaul python (if no version specified) it is .venv as it was

MikeAlexeev: Maybe I didn't understand you correctly.

What should I fix?
If no python version is specified (simply make virtualenv)  then $VENV_PY_VERSION is not
defined when we define VENV_NAME. 
So VENV_NAME is set to .venv because we use ":=" assignment. In this case variables are 
expanded immediately. 
If it was  "=" assignment then variables would expand recursively. So it would be
.venvpython because later we define VENV_PY_VERSION (set it to python if not defined).

I did so because I thought what it is easier than several ifs.
Maybe I should add comment about usage with no python version specified

tardyp: ok. I did not realize that I didn't see the := and though that lazy eval would be used,
 
Yesterday. I did some experiment to replace make virtualenv by pyenv.
Don't you think it could be a good alternative to custom makefile?

MikeAlexeev: I haven't tried pyenv yet. I will try it today

MikeAlexeev: I have tested pyenv a little. 
It is very handy and has plugin for virtualenv management (create, delete, activate).
How shall we install dependencies to virtualenv? Something like requirements.txt ?
So buildbot dev environment setup looks like:
```
pyenv install <some python version>
pyenv virtualenv <some python version> <venv name>
pyenv activate <venv name>
pip install -r requirements.txt

```
Is it right?

It seems that some system packages should be installed. 
For example, pip from python installed (from sources) via pyenv couldn't install twisted because 
that python was built without bz2 support. After I have installed libbz2-dev and reinstalled python
 in pyenv it worked as should.  
It is described here: https://github.com/pyenv/pyenv/wiki/Common-build-problems

Maybe it would be good to use requirements.txt in Makefile too. So both methods can be
used and there will be one packages list for both methods until pyenv tested well in
buildbot development.

tardyp: sorry. I was not talking about pyenv, but rather pipenv

basically with pipenv you do
```
sudo pip install pipenv  # install once
pipenv install --three  # or pipenv install --two
```
pipenv requires a Pipfile created, which describe the list of dependencies needed for prod and development

MikeAlexeev: maybe dev-requirements.txt ?

MikeAlexeev: ok, I will try pipenv tomorrow.

MikeAlexeev: Hello!
Sorry for delay please.
I have tried pipenv. It is very handy but I had two problems.
1) Have you written Pipfile for buildbot development? 
How to specify something like 
pip install -e 'master[tls,test,docs]'  ?
I haven't found example in doc.
Also I tried to install it using pip and then generate Pipefile using pipenv. 
But I haven't found how to do it.

2) Venv folder is the same for every python version.
For example when I switched from py2 to py3, py2 venv was removed.
So each time you wanna switch to another py version fresh venv will be created again with all
dependencies. It takes some time. You cannot simply activate one of venvs created earlier.

Issue #3178, opened 2017-05-08
SecretsInAFile: usability suggestions


Lekensteyn:
The current SecretInAFile provider seems a good start at separating the keys from the main config file. Some observations/suggestions:

 - Strip trailing newlines (or just read the first line from the config file). When the secrets provider is used to read passwords, this seems appropriate and it makes editing passwords with `vim` or `echo sikrit > foo` easier.
 - (Python 3 compatibility note: a str is read, so the secrets framework will/should likely only work with strings rather than arbitrary byte data, making the "read a line" suggestion even more plausible.)
 - Perhaps resolve relative paths to `basedir` by default (and document this behavior). Saves a call to `os.path.join(basedir, dirname)` in master.cfg

For use as password database, maybe it is also an idea to implement a passwd(5)-like file (with lines such as `key:password`).

tardyp:
Hi @Lekensteyn Secret management is very new in Buildbot and indeed it will need some tweaks to make it awesome to use.
I agree with all your suggestions.

Would you like to help and provide those little tweaks?

Lekensteyn:
@tardyp I'll try to give it a go, was checking if you were okay with these changes. Thanks for the feedback.

tardyp:
strip is fine. I would add a strip option to SecretInFile with default to True. Just in case some people actually need the `\n`, and want to set it to False.

I also agree to try and make as small PRs as possible, and thus send one PR per tweak.

Lekensteyn:
Apparently it is not possible to access "basedir" during the checkConfig phase (because the "master" object) is not available (https://irclogs.jackgrigg.com/irc.freenode.net/buildbot/2017-05-16#i_4111140).

So that idea is dropped for now, `os.path.join(basedir, "whatever")` is the current compromise.

PR #3194, opened 2017-05-16
SecretInAFile: strip newlines by default

Lekensteyn:
The secrets provider can be used to read passwords which likely do not
have trailing newlines. Strip them by default, this makes it easier
when editing via vim or using a command like `echo sikrit > foo`.

Add a strip option just in case you want to override this behavior. Note
that for some reason "\r" are converted to newlines anyway, so perhaps
this option can just be removed completely.

Fixes https://github.com/buildbot/buildbot/issues/3178

## Contributor Checklist:

### DIFF
--- a/master/buildbot/secrets/providers/file.py
+++ b/master/buildbot/secrets/providers/file.py
@@ -48,7 +48,7 @@ class SecretInAFile(SecretProviderBase):
                 if secretfile.endswith(suffix):
                     self.checkFileIsReadOnly(dirname, secretfile)
 
-    def loadSecrets(self, dirname, suffixes):
+    def loadSecrets(self, dirname, suffixes, strip):
         secrets = {}
         for secretfile in os.listdir(dirname):
             secretvalue = None
@@ -58,22 +58,25 @@ class SecretInAFile(SecretProviderBase):
                         secretvalue = source.read()
                     if suffix:
                         secretfile = secretfile[:-len(suffix)]
+                    if strip:
+                        secretvalue = secretvalue.rstrip("\r\n")
                     secrets[secretfile] = secretvalue
         return secrets
 
-    def checkConfig(self, dirname, suffixes=None):
+    def checkConfig(self, dirname, suffixes=None, strip=True):
         self._dirname = dirname
         if suffixes is None:
             suffixes = [""]
         self.checkSecretDirectoryIsAvailableAndReadable(dirname,
                                                         suffixes=suffixes)
 
-    def reconfigService(self, dirname, suffixes=None):
+    def reconfigService(self, dirname, suffixes=None, strip=True):
         self._dirname = dirname
         self.secrets = {}
         if suffixes is None:
             suffixes = [""]
-        self.secrets = self.loadSecrets(self._dirname, suffixes=suffixes)
+        self.secrets = self.loadSecrets(self._dirname, suffixes=suffixes,
+                                        strip=strip)
 
     def get(self, entry):
         """
--- a/master/docs/developer/secrets.rst
+++ b/master/docs/developer/secrets.rst
@@ -49,10 +49,10 @@ File provider
 
 .. code-block:: python
 
-    c['secretsProviders'] = [util.SecretInAFile(dirname="/path/toSecretsFiles"]
+    c['secretsProviders'] = [util.SecretInAFile(dirname="/path/toSecretsFiles")]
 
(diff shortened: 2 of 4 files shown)


master/buildbot/secrets/providers/file.py
```diff
@@ -58,22 +58,25 @@ def loadSecrets(self, dirname, suffixes):
                         secretvalue = source.read()
                     if suffix:
                         secretfile = secretfile[:-len(suffix)]
+                    if strip:
+                        secretvalue = secretvalue.rstrip("\r\n")
```
tardyp: I would strip all the whitespaces the python way . secretvalue.strip()

master/docs/developer/secrets.rst
```diff
@@ -52,7 +52,7 @@ File provider
     c['secretsProviders'] = [util.SecretInAFile(dirname="/path/toSecretsFiles"]
 
 In the master configuration the provider is instantiated through a Buildbot service secret manager with the file directory path.
-File secrets provider reads the file named by the key wanted by Buildbot and returns the contained text value.
+File secrets provider reads the file named by the key wanted by Buildbot and returns the contained text value (removing trailing newlines if present).
```
tardyp: removing leading and trailing whitespace

master/docs/manual/secretsmanagement.rst
```diff
@@ -63,7 +63,8 @@ SecretInAFile
 
 In the passed directory, every file contains a secret identified by the filename.
```
tardyp: we need to document the full signature here

master/buildbot/secrets/providers/file.py
Lekensteyn: That would not work with passwords that start with a space (yes, I have at least one such password).

tardyp:
@Lekensteyn Could you please update the documentation as suggested?

Lekensteyn:
@tardyp Done, is this good enough or do you want the `strip=True` parameter to be included too oin the function call?
```diff
diff --git a/master/docs/manual/secretsmanagement.rst b/master/docs/manual/secretsmanagement.rst
index a6876d74a..344b23070 100644
--- a/master/docs/manual/secretsmanagement.rst
+++ b/master/docs/manual/secretsmanagement.rst
@@ -64,7 +64,7 @@ SecretInAFile
 In the passed directory, every file contains a secret identified by the filename.
 
 e.g: a file ``user`` contains the text ``pa$$w0rd``. Trailing newlines are
-removed from the contents.
+removed from the contents (unless ``strip=False``).
 
 SecretInVault
 `````````````
```

tardyp:
@Lekensteyn I would like the documentation to document the actual signature for SecretsInAFile

Lekensteyn:
@tardyp Done (also fixed a syntax error in the existing example)

master/docs/manual/secretsmanagement.rst
```diff
@@ -59,11 +59,13 @@ SecretInAFile
...
 .. code-block:: python
 
-    c['secretsProviders'] = [util.SecretInAFile(dirname="/path/toSecretsFiles"]
+    c['secretsProviders'] = [util.SecretInAFile(dirname="/path/toSecretsFiles",
+                                                strip=True)]
 
 In the passed directory, every file contains a secret identified by the filename.
 
-e.g: a file ``user`` contains the text ``pa$$w0rd``.
+e.g: a file ``user`` contains the text ``pa$$w0rd``. Trailing newlines are
```
tardyp: please look at other part of the doc for example on how we describe the signature of a step ex:
```

.. py:class:: buildbot.steps.source.git.Git

The :bb:step:`Git` build step clones or updates a `Git <http://git.or.cz/>`_ repository and checks out the specified branch or revision.


.. note::

    The Buildbot supports Git version 1.2.0 and later: earlier versions (such as the one shipped in Ubuntu 'Dapper') do not support the :command:`git init` command that the Buildbot uses.

::

    from buildbot.plugins import steps

    factory.addStep(steps.Git(repourl='git://path/to/repo', mode='full',
                              method='clobber', submodules=True))

The Git step takes the following arguments:

``repourl``
   (required): the URL of the upstream Git repository.

``branch``
   (optional): this specifies the name of the branch to use when a Build does not provide one of its own.
   If this parameter is not specified, and the Build does not provide a branch, the default branch of the remote repository will be used.

[...]
```

Lekensteyn:
Thanks @tardyp , that was a good pointer, I've added a list of arguments.

Issue #3078, opened 2017-03-25
Github change hook broken in 0.9.5 if payload is not ascii (PY2)


vbraun:
I just updated to 0.9.5 and the github changehook doesn't work any more (this is the sagemath.org buildbot). I can simulate it with wget, the github webhook console contains the same Error 400 with body "ascii": 
~~~~
$ wget -S -O - --header="Content-Type: application/json" --header="X-Github-Event: push" --post-file=/tmp/payload.json http://build.sagedev.org/change_hook/github
--2017-03-25 22:49:02--  http://build.sagedev.org/change_hook/github
Resolving build.sagedev.org (build.sagedev.org)... 128.208.160.249
Connecting to build.sagedev.org (build.sagedev.org)|128.208.160.249|:80... connected.
HTTP request sent, awaiting response... 
  HTTP/1.1 400 ascii
  Date: Sat, 25 Mar 2017 21:49:54 GMT
  Content-Length: 5
  Content-Type: text/html
  Server: TwistedWeb/16.3.0
2017-03-25 22:49:03 ERROR 400: ascii.
~~~~
The log doesn't contain any error (I was hoping for a backtrace, but no luck):
~~~~
...
2017-03-25 21:40:54+0000 [_GenericHTTPChannelProtocol,435,91.66.84.148] X-GitHub-Event: push
2017-03-25 21:40:54+0000 [_GenericHTTPChannelProtocol,435,91.66.84.148] New revision: 16c0bfa0
...
~~~~
But buildbot doesn't actually pick up any changes.

If I manually strip out non-ascii characters then the payload goes through.

The payload is http://sagepad.org/pub/payload.json

seankelly:
What was your previous version? 0.9.4?

Checking the JSON file and it decodes as valid UTF-8. Not sure why it would fail now.

seankelly:
FYI, 0.9.5 is when the GitHub change_hook was converted to support Python 3.

This is the [offending line](https://github.com/buildbot/buildbot/blob/v0.9.5/master/buildbot/www/hooks/github.py#L202):

    'author': '{} <{}>'.format(commit['author']['name'],                   
                               commit['author']['email']),                 

The name in this case contains a non-ascii character, ü,

```
>>> "{} <{}>".format(u"Julian Rüth", u"b")                                             
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #3198, opened 2017-05-17
hooks: accept unicode in PR comments

noc0lour:
convert string in format to unicode so unicode characters in PR comments are handled properly and don't generate a silent error

### DIFF
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -150,7 +150,6 @@ class GitHubEventHandler(PullRequestMixin):
 
         properties = self.extractProperties(payload['pull_request'])
         properties.update({'event': event})
-
         change = {
             'revision': payload['pull_request']['head']['sha'],
             'when_timestamp': dateparse(payload['pull_request']['created_at']),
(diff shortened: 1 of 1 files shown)


seankelly:
Reminds me of #3144.

noc0lour:
From what I've read having a unicode string and doing format with an ASCII string on it is ok. The other way around it will throw an error.

seankelly:
@noc0lour Yep. #3078 was opened about that issue, but for the author. Are there any other strings that are formatted?

noc0lour:
I thought about changing all strings who use format to unicode but this might bring harm in other places.  No idea why the unit test fails @rodrigc . There was some thing about lz4 in another issue though.

tardyp:
lz4 issue is due to the new lz4 package I am working on it with #3200

rodrigc:
@noc0lour in future, as long as you can get things to work, I think it is OK to try to convert all strings  to unicode.  This will give consistent behavior between Python 2 and 3, since by default on Python 3 strings are unicode.

For my initial Python 3 work, I only converted enough strings to unicode to get the unit tests to pass.  That's not to say that more can be done.

noc0lour:
K, I'll look for strings with format and try to convert them to unicode mostly... Most important are strings with user input.

rodrigc:
@noc0lour Take a look at https://mail.python.org/pipermail/python-dev/2017-May/147949.html for interesting reading about this.

I think converting all instances of:

```
"foobar {}".format(whatever)
```

to:

```
u"foobar {}".format(whatever)
```

Is the best way to go.

noc0lour:
I agree. Yesterday I just was too lazy to come up with a regex which will safely match all cases ;)

tardyp:
actually, we need this PR to be merged in order to have pyup working..
pyup puts smileys in its PRs, so they wont be accepted by nine.bb.net :-)

noc0lour:
If you want you can already merge this small change, and I come up with a way to change other strings. I already use this commit in https://ci.gnuradio.org/buildbot/

PR #3200, opened 2017-05-17
adds a requirements.txt for smooth dependency update with pyup.io

tardyp:


.bbtravis.yml
```diff
@@ -89,6 +89,7 @@ install:
   - |
       # pip installs
       set -e
+      pip install -r requirements-dev.txt
```
seankelly: This fails on the worker with Python 2.6:
```
ERROR: You need Python 2.7 or 3.3+ to install the typing package.
```

Issue #3188, opened 2017-05-13
doStepIf is not renderable


trefzer:
according to the documentation I expected doStepIf to bee renderable, but it's not !

trefzer:
I don't think this is a documentation issue, I rather think the this step should be rendable, otherwise this option would not make any sense for me since you could replace it with a simple 'if' statement in the configuration code. Or said differently, if this option is not rendrable you could simply delete the option since it has no additional value.

Frodox:
@trefzer I think is it resolved now :) 
I see noting in docs about renderer, [just](http://docs.buildbot.net/current/manual/cfg-buildsteps.html)
 
> To do this, set the step’s doStepIf to a boolean value, or to a function that returns a boolean value or Deferred. 

So, it looks like `def myfunc(step):` didn't work too?

trefzer:
according to me its solved.
the problem was that the rendering did happen after doStepif was used (in short he did: IF doStepIf true, THEN render doStepIf).

PR #3214, opened 2017-05-18
fix for doStepIf not being renderable

trefzer:
- evaluation of doStepIf should be done after rendering

fixes Issue #3188

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -543,12 +543,6 @@ class BuildStep(results.ResultComputingConfigMixin,
             if self.stopped:
                 raise BuildStepCancelled
 
-            # check doStepIf
-            if isinstance(self.doStepIf, bool):
-                doStep = self.doStepIf
-            else:
-                doStep = yield self.doStepIf(self)
-
             # render renderables in parallel
             renderables = []
             accumulateClassList(self.__class__, 'renderables', renderables)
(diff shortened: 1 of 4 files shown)


master/buildbot/process/buildstep.py
```diff
@@ -566,6 +560,14 @@ def setRenderable(res, attr):
             # we describe ourselves only when renderables are interpolated
             self.realUpdateSummary()
 
+            # check doStepIf (after rendering)
```
noc0lour: I agree on putting it after the rendering, but converting str/unicode to bool shouldn't be necessary since the renderable should return bool and nothing else beside that. The last case is just to convert the result from the callable to something usable.
A unit test which fails before the changes and passes after would be much appreciated

trefzer: You are  right, converting str/unicode is not necessary. 
I removed the original two commits, a new one is below.

About unit tests, I agree, but I have no idea about howto write them. So I will not implement. Sorry.

tardyp: Hi @trefzer 
Here is the documentation on how to run/write unit tests for buildbot.

You can start by modifying this test
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/test_process_buildstep.py#L242
and run ```trial buildbot.test.unit.test_process_buildstep.TestBuildStep.test_doStepIf_false```

Please remember that this project is entirely run by volunteers. Unit tests are an essential way to keep our code stable for the benefit of the whole community. We cannot accept contributions without unit test and docs update.

This code is fundamental and run by everybody. It is fundamental it is carefully tested.

Thank you for your understanding

trefzer: @tardyp thanks for doku links and the entry point, that saved lot of time.

PR #3263, opened 2017-05-26
Fix ChangesEndpoint get doesn't limit results

rkashapov:
I have encountered that BuildBot is hanging if I visit the Last Changes page in web UI.
It turned out that when ChangesEndpoint.get method tries to determine which one to use - `db.changes.getRecentChanges` or  `db.changes.getChanges` it compares tuple and list and makes the wrong decision and makes try to return all the changes.

It looks like the bug was introduced in [this](https://github.com/buildbot/buildbot/commit/6d0ebb56175beaa8a3776a071a23a7d0b6f37088#diff-bcc309a389ecb20d626f59e7a656dd48R273) commit. 
## Contributor Checklist:

### DIFF
--- a/master/buildbot/data/changes.py
+++ b/master/buildbot/data/changes.py
@@ -85,7 +85,7 @@ class ChangesEndpoint(FixerMixin, base.Endpoint):
         else:
             # this special case is useful and implemented by the dbapi
             # so give it a boost
-            if (resultSpec.order == ['-changeid'] and resultSpec.limit and
+            if (resultSpec.order == ('-changeid',) and resultSpec.limit and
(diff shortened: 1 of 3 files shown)


tardyp:
Thanks for the fix. I think we should rather switch the code to use the generic resultSpec. applyToSQLQuery()
as for the builds and buildrequests endpoints.

That would be nice if you find some time to work on it, this will allow more performant query combinations for the REST api.
Waiting for that this fix is still better than nothing.

rkashapov:
@tardyp Thank you to accept my pull request. I'll try to find some time to work on that.

PR #3269, opened 2017-05-27
logpreview: allow logpreviews to be collapsed

Lekensteyn:
Make log previews collapsable, useful when you want a summary of the log (number of lines) while reviewing the results of another step.

Additionally, collapse previews by default when when multiple logs are available (stdio, warnings).
___
screencast (1.2MIB)
___
Not sure how to write a test (spec) file for this usability feature, but manual verification resulted in the following:
1. Finished builds with one preview are expanded.
2. Finished builds with multiple previews are collapsed.
3. (observation:) The collapses state of the preview is not remembered (implementation detail of how the Logpreview component is removed when the step is collapsed).
4. Live builds also start expanded, but as soon as the second log is added in a step, both are collapsed.

Another possible (unimplemented) option would be not to fetch the log contents until it is expanded, saving a request.

## Contributor Checklist:

### DIFF
--- a/www/base/guanlecoja/config.coffee
+++ b/www/base/guanlecoja/config.coffee
@@ -45,7 +45,7 @@ config =
                 version: "~3.1.1"
                 files: []
             'buildbot-data':
-                version: '~2.1.0'
+                version: '~2.2.0'
                 files: 'dist/buildbot-data.js'
 
         testdeps:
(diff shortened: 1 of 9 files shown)


tardyp:
Please copy paste screenshot in your PR description so that it is easier to understand what you did

Lekensteyn:
@tardyp Done, added a screencast

tardyp:
Looks good. This would be indeed good to lazy load the log content if full display. keeping it open to get more feedback from others

Lekensteyn:
Rebased on #3277 to allow for cancelling requests and implemented lazy loading and cancelling of pending requests.

tardyp:
thanks!

PR #3274, opened 2017-05-27
www: update guanlecoja to 0.8.3

Lekensteyn:
Fixes installation with newer node versions (v7.10.0). Fixes an issue
with phantomjs when TMPDIR is not set. Adds missing yarn.lock files.

Since the package.json dependencies are the same for codeparameter,
console_view, nestedexample, waterfall_view and wsgi_dashboards, I
expect their yarn.lock also to be identical. This is achieved by running
`yarn upgrade` once for each followed by `yarn upgrade --prefer-offline`
to get a deterministic run where the lock file does not change (due to
different dependency resolution?).
___
Tested by trying to install the frontend in a clean VM (with Arch Linux). Fails before this patch with at least the `console_view` package, after this patch installation completes as expected.

I initially had some issues with deterministic orderings with yarn (https://github.com/yarnpkg/yarn/issues/1168#issuecomment-304464905), but resolved it with two steps:

    cd www
    # First install dependencies (this was repeated at least 10 times, each time something different changed...)
    for i in */;do (cd "$i" && yarn upgrade); done
    # In the end this was tried and it gave a fixed ordering on repetitive runs.
    for i in */;do (cd "$i" && yarn upgrade --prefer-offline); done

## Contributor Checklist:


tardyp:
thanks!

PR #3277, opened 2017-05-27
[www] Make it possible to cancel requests

### LINKED ISSUE #3275: [logpreview] requests should be cancelled when the log is closed
While loading a log, if a step is quickly closed and expanded, multiple lines will appear.

## Steps to reproduce
1. Open Build steps (e.g. https://nine.buildbot.net/#/builders/8/builds/38)
2. (Optional, for faster reproduction:) Enable network throttling (e.g. for Chromium this can be done in the network tab, set to GPRS 5000ms, 500kb/s)
3. Click several times on one step.
4. Open the step.

## Expected results
Trailing lines of the log, e.g. lines 90-129

## Actual results
Duplicated lines, e.g. lines 119-129 where the middle lines are repeated four times.

Lekensteyn:
Extends the RestService interface with a cancel method on pending requests. This fixes #3275 which should help with slow links/servers and larger logs. Additionally, it can be used to reduce data usage when the view is hidden anyway.

Implementation relies on timeout parameter of $http:
https://docs.angularjs.org/api/ng/service/$http
https://stackoverflow.com/questions/35375120/cancelling-ongoing-angular-http-requests-when-theres-a-new-request

If this looks good, please consider publishing an updated buildbot-data package with the first change before merging the second patch. Let me know if you prefer two PRs for this.

## Contributor Checklist:

### DIFF
--- a/www/data_module/yarn.lock
+++ b/www/data_module/yarn.lock
@@ -2083,17 +2083,13 @@ oauth-sign@~0.8.1:
   version "0.8.2"
   resolved "https://registry.yarnpkg.com/oauth-sign/-/oauth-sign-0.8.2.tgz#46a6ab7f0aead8deae9ec0565780b7d4efeb9d43"
 
-object-assign@4.1.0:
+object-assign@4.1.0, object-assign@^4.0.1, object-assign@^4.1.0:
   version "4.1.0"
   resolved "https://registry.yarnpkg.com/object-assign/-/object-assign-4.1.0.tgz#7a3b3d0e98063d43f4c03f2e8ae6cd51a86883a0"
 
 object-assign@^3.0.0:
   version "3.0.0"
   resolved "https://registry.yarnpkg.com/object-assign/-/object-assign-3.0.0.tgz#9bedd5ca0897949bca47e7ff408062d549f587f2"
-
-object-assign@^4.0.1, object-assign@^4.1.0:
-  version "4.1.1"
-  resolved "https://registry.yarnpkg.com/object-assign/-/object-assign-4.1.1.tgz#2109adc7965887cfc05cbbd442cac8bfbb360863"
 
 object-component@0.0.3:
   version "0.0.3"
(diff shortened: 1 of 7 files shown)


tardyp:
looks good. I published the new version of buildbot-data, and updated your branch

www/data_module/yarn.lock
```diff
@@ -2083,18 +2083,14 @@ oauth-sign@~0.8.1:
...
-object-assign@4.1.0:
+object-assign@4.1.0, object-assign@^4.0.1, object-assign@^4.1.0:
   version "4.1.0"
   resolved "https://registry.yarnpkg.com/object-assign/-/object-assign-4.1.0.tgz#7a3b3d0e98063d43f4c03f2e8ae6cd51a86883a0"
 
 object-assign@^3.0.0:
   version "3.0.0"
   resolved "https://registry.yarnpkg.com/object-assign/-/object-assign-3.0.0.tgz#9bedd5ca0897949bca47e7ff408062d549f587f2"
 
-object-assign@^4.0.1, object-assign@^4.1.0:
```
Lekensteyn: Thanks, I indeed had the unittest failure with chromium 58.0.3029.110.

As for this yarn.lock change, I noticed that the same issue happens after `yarn install` for each package. Really wished that this resolution is more deterministic :/

tardyp:
I think we should add those files in gitignore. They don't add anything for
a frontend build system.

Lekensteyn:
These yarn.lock files [should be tracked](https://stackoverflow.com/q/39990017/427545) to ensure reproducible installations. To prevent them from being modified during installation, what about using the `yarn install --pure-lockfile`?

tardyp:
@Lekensteyn In our case, we don't have problems with reproducible installations. We would have to fix the buildbot-pkg package to use --pure-lockfile.
As npm packages are constantly evolving we will have lots of patches to yarn.lock, which are just noise.

PR #3282, opened 2017-05-29
Fix wrong GerritStatusPush.name attribute

rkashapov:
I have following buildbot configuration:
```python
class FooStatusPush(GerritStatusPush):
    name = 'foo'

class BarStatusPush(GerritStatusPush):
   name = 'bar'

config['services'].extend([
    FooStatusPush(server_url, username, etc...),
    BarStatusPush(server_url, username, etc...),
])
```

It used to work for me, but after update to version 0.9.7 it have broken.
After some investigations I have found out that problem was in [`BuildbotService.checkConfig`](https://github.com/buildbot/buildbot/blob/v0.9.7/master/buildbot/util/service.py#L216) method.
It takes the first constructor argument (which is a Gerrit server name) and assignes it to the name attribute. If I have several reporters defined in my configuration with the same Gerrit server only one of them will work eventually, because all of them will have the same name.
## Contributor Checklist:

### DIFF
--- a/master/buildbot/util/service.py
+++ b/master/buildbot/util/service.py
@@ -212,9 +212,8 @@ class BuildbotService(AsyncMultiService, config.ConfiguredMixin, util.Comparable
                 pass
         yield AsyncMultiService.startService(self)
 
-    def checkConfig(self, name=None, *args, **kwargs):
-        if name is not None:
-            self.name = ascii2unicode(name)
+    def checkConfig(self, *args, **kwargs):
+        return defer.succeed(True)
 
     def reconfigService(self, name=None, *args, **kwargs):
         return defer.succeed(None)
--- a/master/buildbot/test/unit/test_reporter_gerrit.py
+++ b/master/buildbot/test/unit/test_reporter_gerrit.py
@@ -509,3 +509,18 @@ class TestGerritStatusPush(unittest.TestCase, ReporterTestMixin):
         self.patch(reactor, 'spawnProcess', spawnProcess)
         gsp.callWithVersion(lambda: self.assertEqual(
             gsp.gerrit_version, LooseVersion('2.14')))
+
+    def test_name_as_class_attribute(self):
+        class FooStatusPush(GerritStatusPush):
+            name = 'foo'
+
+        reporter = FooStatusPush('gerrit.server.com', 'password')
+        self.assertEqual(reporter.name, 'foo')
+
+    def test_name_as_kwarg(self):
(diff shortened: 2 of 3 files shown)


master/buildbot/util/service.py
```diff
@@ -213,7 +213,7 @@ def startService(self):
         yield AsyncMultiService.startService(self)
 
     def checkConfig(self, name=None, *args, **kwargs):
-        if name is not None:
+        if self.name is None and name is not None:
```
tardyp: the name set by the constructor should take priority

Wootanis2531:
Jirawoot agrees.
Spark by Readdle

master/buildbot/reporters/gerrit.py
```diff
@@ -300,6 +300,9 @@ def buildComplete(self, key, build):
             result = _handleLegacyResult(result)
             self.sendCodeReviews(build, result)
 
+    def checkConfig(self, *args, **kwargs):
```
tardyp: checkConfig is usually found at the beginning of the class

tardyp: Looks like I did not realize the problem at first glance.
You say the it did broke in .0.9.7. Do you have an idea of what did create the issue?
would you like to git bisect?

rkashapov: I have searched a commit that introduced the bug. It was [this commit](https://github.com/buildbot/buildbot/commit/3d2d77095e9a21caab0cbd93298adf103136252e). It looks like the reason that caused the bug was following diff in `BuildbotService` code.
```diff
-    def checkConfig(self, *args, **kwargs):
-        return defer.succeed(True)
+    def checkConfig(self, name=None, *args, **kwargs):
+        if name is not None:
+            self.name = ascii2unicode(name)

```

tardyp: Thanks for the root causing. I think we should rather revert this hunk.

I did not realize that this would put name as the first positional argument. :-(

what if we use kwargs.pop('name') instead?

rkashapov: That is already done in [BuildService.__init__](https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/service.py#L174)

I think we should take the previous version of checkConfig.
```python
def checkConfig(self, *args, **kwargs):
    return defer.succeed(True)
```

rkashapov:
I have reverted breaking changes in BuilbotService.

PR #3287, opened 2017-05-30
BuildbotService.__init__() pop() -> get() name

nazarewk:
`kwargs.pop("name", None)` entirely prevented `name` from descending onto subclasses' positional argument.

For example:
```python
>>> worker.Worker(name='example', password='example')
...
builtins.TypeError: checkConfig() missing 1 required positional argument: 'name'
```

### DIFF
--- a/master/buildbot/util/service.py
+++ b/master/buildbot/util/service.py
@@ -171,7 +171,7 @@ class BuildbotService(AsyncMultiService, config.ConfiguredMixin, util.Comparable
     objectid = None
 
     def __init__(self, *args, **kwargs):
-        name = kwargs.pop("name", None)
+        name = kwargs.get("name", None)
         if name is not None:
             self.name = ascii2unicode(name)
         self.checkConfig(*args, **kwargs)


tardyp:
So this may be better to worker, but worse for other services (see unit tests failing)

Now we try to avoid positional arguments.
We only support name as positional argument for worker.
Is this really a problem for you?

tardyp:
is it related to #3282 ?

nazarewk:
@tardyp it is not related to #3282 

The problem is that `Worker` (and probably a lot of other services) have absolutely no way of knowing about keyword argument `name` during initialization. So i guess it requires a lot of refactor to even start fixing the root issue?

tardyp:
@nazarewk It is not really the problem of refactorization. It is the problem of stable API.

name has always been a positional argument for Worker, so we cannot put it as keyword argument, so that it plays better with service.

name is a reserved word for service, that happens to be re-used for authentication of workers.
I don't see the use case of having the service's checkConfig to know about their name passed via keyword instead of class attribute.
They can still do some checks of self.name at the time of checkConfig if really needed, and that would be the class name or the keyword if set.

nazarewk:
Ok, knowing the context let's just close this PR.

tardyp:
@nazarewk thanks for the feedback!

PR #3289, opened 2017-06-02
PY3: paths should be str not bytes (equality)

nazarewk:
Twisted reads paths as bytes (and rightly so), but buildbot codebase
(at least matchers) expect them to be unicode strings and equality
doesn't work between bytes and str


tardyp:
py3 test fail I am not sure why..

nazarewk:
I'm gonna check it after holidays around 12th-13th, errors are quite cryptic.

tardyp:
would be nice to have this fixed for the next release I want to do next week

tardyp:
I fixed the unit test in your branch and added a relnote

PR #3290, opened 2017-06-02
www: fix timestamp insertion for scripts.js

Lekensteyn:
Attribute interpolation syntax has changed, adjust accordingly to
prevent the code from being shown literally.  See
https://pugjs.org/language/attributes.html

## Contributor Checklist:


tardyp:
ah! I wonder why the cache avoidance seems to not work anymore. good catch! thank you so much!

PR #3296, opened 2017-06-05
Waterfall usability fixes

Lekensteyn:
Usability fixes for the waterfall view, see the individual commit messages for details.

Situation before:
screenshot situation before

Notice the scrollbar on the bottom. Notice the wrong position of the tooltip (and its failure to complete loading since it appears and disappears continously).

Situation after:
screenshot situation after

Notice the spacing on the right of the graph.

## Contributor Checklist:

### DIFF
--- a/www/waterfall_view/src/module/main.module.coffee
+++ b/www/waterfall_view/src/module/main.module.coffee
@@ -462,7 +462,8 @@ class Waterfall extends Controller
         p.each -> @parentNode.appendChild(this)
 
         # Show tooltip on the left or on the right
-        r = build.builderid < self.builders.length / 2
+        builderIds = self.builders.map((builder) -> builder.builderid)
+        r = builderIds.indexOf(build.builderid) < self.builders.length / 2
 
         # Create tooltip
         height = 40
--- a/www/waterfall_view/src/module/main.module.coffee
+++ b/www/waterfall_view/src/module/main.module.coffee
@@ -300,6 +300,7 @@ class Waterfall extends Controller
             p = self.d3.select(@parentNode)
             a = p.append('a')
                 .attr('xlink:href', "#/builders/#{builderid}")
+            a.append('title').text(this.textContent)
             a.node().appendChild(this)
 
         # Rotate text
--- a/www/waterfall_view/src/module/main.module.coffee
+++ b/www/waterfall_view/src/module/main.module.coffee
@@ -470,6 +470,7 @@ class Waterfall extends Controller
             else "150,0 170,#{height / 2} 150,#{height} 0,#{height} 0,0"
         tooltip = e.append('g')
             .attr('class', 'svg-tooltip')
+            .style('pointer-events', 'none')
             .attr('transform', "translate(#{mouse[0]}, #{mouse[1]})")
             .append('g')
                 .attr('class', 'tooltip-content')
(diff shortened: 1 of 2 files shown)


Lekensteyn:
Updated the PR with another fix and screenshots.

tardyp:
I did that spacing on purpose. notice the horizontal scollbar.
I think that it is better to fill the whole space for bars and have to scroll a bit to see full name for builders.
Actually I am happy with neither solution.
I realize this is subjective, and I am willing to withdraw if others think your solution is best.

Lekensteyn:
That horizontal scrollbar actually annoys me, at the moment I have only three workers so the scrollbar feels just weird (especially since the builder names are not actually falling off the page). The whitespace on the left is also not something I am happy with, but at least it removes the scrollbar.

Possible alternatives that I have considered:
- rotate 90 degrees such that the text is vertically displayed. Not sure if it would make it much more readable.
- Just remove spacing completely. With the title (tooltip) on top of the builder name it should be manageable.

Note that if you shrink the window (and then reload), the text will fall off the screen anyway. (Reloading is necessary as the dimensions are computed only once.)

Another idea is to change the current bottom-left/top-right orientation to top-left/bottom-right. This would take advantage of the spacing that is already available for the times, but has as disadvantage that the text starts elsewhere.

Another option is to move the time marker from the left to the right. Disadvantage is that it is also unnatural to read.

tardyp:
I didn't think about top left to bottom right orientation. Might be a good compromise.
Can you post a screenshot?

Lekensteyn:
Rotating the text requires more work because the width has to be determined before translation. If you disregard that the text is upside down and in reverse, this is how it would look like (rotate(-155)):
screenshot rotate(-155)

The text does not fully fit either so that does not seem ideal.

This is how it would look like if you rotate(-90):
screenshot rotate(-90)

And this is how the unpatched version looks like with just three builders:
screenshot with three builders

(if the window size is increased, then the text will not fall off.)

tardyp:
I think we can specify the rotation anchor in order to implement rotate top-left to bottom right.

I think this is the best compromise

Frodox:
and `-90` too !

Hnasar:
What about line wrapping like in bb 0.8?
That would mean we could avoid awkward text rotation.

Hnasar:
But I do think having the whitespace in the original fix is slightly preferable to the ever present horizontal scroll bar

tardyp:
> What about line wrapping like in bb 0.8?

buildbot nine manages to fit much more builders in one screen by using this. With current width, line-wrapping would cut words every 2 chars. So In order to make line wrapping work, you would have to increase the minimum column size.

> But I do think having the whitespace in the original fix is slightly preferable to the ever present horizontal scroll bar

Not sure... If we can't decide, probably we would have to put that configurable.

Lekensteyn:
Sorry for the delay, I won't be able to look into this for the at least the next two weeks (and probably longer). Would it be possible to cherry-pick at least the Firefox fix while waiting for the rotation change?

rodrigc:
@Lekensteyn I cherry-picked your firefox fix in https://github.com/buildbot/buildbot/pull/3634

Frodox:
ping @Lekensteyn any time now..?

Lekensteyn:
@Frodox Not yet, I have not kept up-to-date with the latest changes (need to check whether they still apply). Should I just close this PR and open a new one when it's ready?

Lekensteyn:
Sorry, have not gotten back to this.

PR #3174, opened 2017-05-04
Doc: fixing Apache websocket example 

gracinet:
Hello,
after upgrading one of my masters to Debian 9 (Stretch) I found the Apache configuration to be broken, here's an updated working one.

Regards

PR #3299, opened 2017-06-06
[docs] Fix Apache reverse proxy config

Frodox:

### DIFF
--- a/master/docs/manual/cfg-www.rst
+++ b/master/docs/manual/cfg-www.rst
@@ -562,7 +562,7 @@ Here is an nginx configuration that is known to work (nginx 1.6.2):
 
 To run with Apache2, you'll need `mod_proxy_wstunnel <https://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html>`_ in addition to `mod_proxy_http <https://httpd.apache.org/docs/2.4/mod/mod_proxy_http.html>`_. Serving HTTPS (`mod_ssl <https://httpd.apache.org/docs/2.4/mod/mod_ssl.html>`_) is advised to prevent issues with enterprise proxies (see :ref:`SSE`), even if you don't need the encryption itself.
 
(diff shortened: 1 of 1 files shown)


Frodox:
Just found that it breaks #3174. What is correct way to resolve it? Paste one more config?
Add configs into some examples dir?
According to [ws-apache-docs](https://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html) `ProxyPass` should match some Path (as in nginx config), like
```
ProxyPass "/ws2/"  "ws://echo.websocket.org/"
ProxyPass "/wss2/" "wss://echo.websocket.org/"
```

current one
```
ProxyPass ws://127.0.0.1:8020/ws
```

can't be working one because it is not match any path. So it can be only like
```
ProxyPass /ws ws://127.0.0.1:8020/ws
```

Frodox:
from conversation to #3174 it looks like it wrong PR.
Should there be
```
ProxyPass /ws !
```
now?

gracinet:
Agh sorry for the mess, indeed I failed the copy-paste from my running systems. The /ws is indeed missing, that line should have been:

           ProxyPass /ws ws://localhost:8011/ws
          ProxyPassReverse /ws ws://localhost:8011/ws


ProxyPass /ws !, together with <Location> gets back to the prior version, that didn't work for me on Debian 9, hence motivating the prior change.

Just adding the missing /ws would probably be simpler than enclosing in a <Location> (and would match what I have in production on Debian 8 and 9). If that also works for you, then I'd prefer it

Frodox:
@gracinet  what is meaning of `ProxyPass /ws !` ? So, should it be returned with note about Debian9 ? What happens without this option?

gracinet:
I works, without ProxyPass /ws !  for both Debian 8 and 9, but I did not test it with the re-added <Location>

That ! directive is meant to prevent proxying the given URI, and could be used at some point to dispatch between non-proxied / differently proxied nested locations. Everything looks as if its scope changed between the two Debian versions, and it started blocking also what was in <Location /ws>

tardyp:
thanks guys!

Issue #3300, opened 2017-06-07
GitHub authentication not working on re-logins.


s0undt3ch:
I've set my nine installation to use GitHub for authentication.
I login using GitHub, authorize my oauth application to forward my login details to buildbot and I'm authenticated.
I logout and I'm never able to login again. I added some logging statements and when I try to re-login, I see that GitHub calls the right handler on buildbot and sends all the right details and buildbot even populates the groups correctly, but the UI never shows me as authenticated.

To solve this I have to delete all user tokens from my oauth application on GitHub and re-login.

Additionally, after the first successful authentication, I'm unable to login from other devices(browsers).

I also added some debug logs to twisted, namely `twisted.web.http.Request.write()` where I log the cookies and from a device which is not initially authenticated, using the same account which was previously authorized, I try to login, I see the GitHub request to buildbot show up, the `TWISTED_SESSION` cookie get's set for path `/` and then the requests for the several buildbot REST endpoints start to show on logs but they do not contain the `TWISTED_SESSION` cookie header.

This nine install is being proxied by nxing using SSL. The nxing configuration mimics the one provided on buildbot docs.

tardyp:
It might be a problem of secure vs insecure cookie.
secure cookies are only sent to https.

s0undt3ch:
So, buildbot is setting an insecure cookie because it's being served through HTTP and not HTTPS which the browser then rejects because it's accessing the website through HTTPS?

Can we force the secure cookie bit somewhere?

tardyp:
it is handled by twisted web. I think It should work if the reverse proxy is correctly telling that the connection is actually https baked

s0undt3ch:
So, I've set the forward headers in nginx and created https://github.com/buildbot/buildbot/pull/3302, however I notice something.
This works as supposed on Chromium, but doesn't on Firefox so, #3302 might not actually be required?!

seankelly:
Per last comment and from #3306, marking as closed.

PR #3302, opened 2017-06-07
Set the forward flags on the nginx example config

s0undt3ch:
Refs #3300

And especially:

## Contributor Checklist:

### DIFF
--- a/master/docs/manual/cfg-www.rst
+++ b/master/docs/manual/cfg-www.rst
@@ -540,6 +540,13 @@ Here is an nginx configuration that is known to work (nginx 1.6.2):
             add_header Strict-Transport-Security "max-age=31536000; includeSubdomains;";
             spdy_headers_comp 5;
 
+            proxy_set_header HOST $host;
+            proxy_set_header X-Real-IP $remote_addr;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header X-Forwarded-Proto  $scheme;
+            proxy_set_header X-Forwarded-Server  $host;
+            proxy_set_header X-Forwarded-Host  $host;
+
             # you could use / if you use domain based proxy instead of path based proxy
             location /buildbot/ {
                 proxy_pass http://127.0.0.1:5000/;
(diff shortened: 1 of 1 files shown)


master/docs/manual/cfg-www.rst
```diff
@@ -543,20 +543,32 @@ Here is an nginx configuration that is known to work (nginx 1.6.2):
             # you could use / if you use domain based proxy instead of path based proxy
             location /buildbot/ {
                 proxy_pass http://127.0.0.1:5000/;
+                proxy_set_header Host $host;
```
tardyp: I dont think all this complexity is even needed.
Here is what we are using in the nine.buildbot.net

https://github.com/buildbot/buildbot-infra/blob/master/roles/nginx/templates/proxy

I think the whole think could be changed as

    proxy_set_header HOST $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto  $scheme;
    proxy_set_header X-Forwarded-Server  $host;
    proxy_set_header X-Forwarded-Host  $host;
    map $http_upgrade $connection_upgrade {
        default upgrade;
        ''      close;
    }
    location /buildbot/ {
        proxy_pass http://127.0.0.1:5000/;
        proxy_buffering off;
        proxy_read_timeout 600;

        # These three lines are be required for web socket proxying
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection $connection_upgrade;
    }

s0undt3ch:
Updated docs with your suggestion.

As a side note, I continue to be able to login/loout against GitHub using chromium, but not Firefox(this is using the above nginx setup).

tardyp:
@s0undt3ch do you have more info on what is different with FF?

can you compare with https://nine.buildbot.net/#/ which is also implementing github auth?

s0undt3ch:
My guess is something related to the websocket implementation being different on these browsers?

I previously logged in into your nine setup(can't remember which browser I used) successfully but today either using the desktop Firefox or the mobile Firefox I'm unable.

My previous CI process used Jenkins which was also linked to github authentication and I never ran into problems but I believe Jenkins is not using websockets, it's polling.

You can test this behavior by using Firefox, the first login attempt will be successful. Logout and then try to login again.

I'm using Firefox 53.0.2, at least the mobile version of it.

tardyp:
@s0undt3ch Indeed, I can reproduce :-/

s0undt3ch:
I tried looking at the UI code, but I'm to new to this code to find what I was looking for.
Does one need to be explicit about which cookies are used for the websocket connection?
Its a weird question because the websocket connection is an HTTP connection(which the browser should reuse known cookies) which then gets upgraded....

tardyp:
@s0undt3ch see my investigation report at https://github.com/buildbot/buildbot/issues/3306 
It is pretty nasty race condition

Issue #3306, opened 2017-06-09
oauth2 logout + login fails on Firefox


tardyp:
as discussed in #3302

On firefox, the second login with oauth keeps the user in anonymous.

Needs to investigate further

tardyp:
So problem comes from the fact that during login procedure the ws connection is closed, and the JS in charge of detecting that triggers, which triggers polling of / and once a success full xhr on / is done, the JS will reload the page. In the middle of that the github oauth is still going out, but is then interrupted by that JS reload.

![image](https://user-images.githubusercontent.com/109859/26972698-2a455fe2-4d13-11e7-9282-775b7cfd1a83.png)


We need to make sure we disable all angularJS callbacks before going to login page.

s0undt3ch:
Makes sense now.

tardyp:
@s0undt3ch please see the fix in #3307. this fixes the problem for me on nine.buildbot.net. Please report back your own results.

s0undt3ch:
#3307 fixes the problem I was experiencing, Thanks!

PR #3307, opened 2017-06-09
fix issue with ws reconnection logic breaks oauth connection sequence.

tardyp:
When browsing away from the page, Firefox sends the onclose event for the websocket, which eventually triggers a reload of the page.
This races with the oauth connection sequence.
As the browser do not get a document for a while during that sequence, javascript continues to run, and eventually reload the page, cancelling the oauth redirection sequence

fix for #3306 

## Contributor Checklist:


s0undt3ch:
This fixes the problem I was experiencing, Thanks!

PR #3312, opened 2017-06-12
PY3: fix request.method being bytes instead of str

nazarewk:
additionally create tuple to begin with instead of list then changing it to tuple

Yielded `TypeError` in [buildbot.authz.endpointmatchers](https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/authz/endpointmatchers.py#L46)


tardyp:
I'd like to see a test that would trig this problem

PR #3327, opened 2017-06-14
Hook plugins

tardyp:
This is a refactor to make the hook system integrated to the plugin system.
This will allow better customizability, and switch the api to an autodiscovery based on twisted.python.reflect.namedModule to our plugin system.

We also change the API to a single function to a class that can potentially store cache.
The classes have access to self.master, but the getChange function cannot return deferred. This has to be done in a follow up PR.

## Contributor Checklist:


seankelly:
This is great.

PR #3333, opened 2017-06-16
db: allow using in-memory sqlite uri

rjarry:
When using an in-memory sqlite database (`db_url='sqlite://'`), everything is reset at each process restart. This causes the following error:

```
Setting up database with URL 'sqlite:'
The Buildmaster database needs to be upgraded before this version of
buildbot can run.  Use the following command-line

    buildbot upgrade-master ...

to upgrade the database, and try starting the buildmaster again.  You may
want to make a backup of your buildmaster before doing so.
```

Since there is no data persistence, the upgrade-master command would work but everything would be lost when restarting the server.

Handle this special case and initialize the db model at database connection to allow startup to continue.

See the following links for more details:

https://sqlite.org/inmemorydb.html
http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlite

### DIFF
--- a/master/buildbot/db/connector.py
+++ b/master/buildbot/db/connector.py
@@ -127,6 +127,11 @@ class DBConnector(WorkerAPICompatMixin, service.ReconfigurableServiceMixin,
 
         # make sure the db is up to date, unless specifically asked not to
         if check_version:
+            if db_url == 'sqlite://':
+                # Using in-memory database. Since it is reset after each process
+                # restart, `buildbot upgrade-master` cannot be used (data is not
+                # persistent). Upgrade model here to allow startup to continue.
+                self.model.upgrade()
             current = yield self.model.is_current()
             if not current:
                 for l in upgrade_message.format(basedir=self.master.basedir).split('\n'):
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ docker-buildbot-master-ubuntu:
 
 $(VENV_NAME):
 	virtualenv -p $(VENV_PY_VERSION) $(VENV_NAME)
-	$(VENV_NAME)/bin/pip install -U pip
+	$(VENV_NAME)/bin/pip install -U pip setuptools
 	$(VENV_NAME)/bin/pip install -e pkg \
 		-e 'master[tls,test,docs]' \
 		-e 'worker[test]' \
--- a/master/buildbot/test/unit/test_db_connector.py
+++ b/master/buildbot/test/unit/test_db_connector.py
@@ -92,6 +92,9 @@ class DBConnector(db.RealDatabaseMixin, unittest.TestCase):
         return d
 
     def test_setup_check_version_bad(self):
+        if self.db_url == 'sqlite://':
+            raise unittest.SkipTest(
+                'sqlite in-memory model is always upgraded at connection')
         d = self.startService(check_version=True)
         return self.assertFailure(d, exceptions.DatabaseNotReadyError)
 


rjarry:
I guess there was a problem with git clone:

```
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
```

Is there a way to re-trigger a check?

Frodox:
Interesting idea. How much is it faster vs sqlite on HDD or smth like postgresql?

rjarry:
Hi,

It is a lot faster than mysql or postgresql. And several orders of
magnitude faster than sqlite on HDD.

I had big performance issues while setting things up in a dev environment
with sqlite on hdd. Since I don't need persistence at all, in-memory db
makes sense.

However, this change seems to disturb a lot of unit/integration tests....
I'll need to have a closer look.

master/buildbot/db/connector.py
```diff
@@ -125,6 +125,12 @@ def setup(self, check_version=True, verbose=True):
         self.pool = pool.DBThreadPool(
             self._engine, reactor=self.master.reactor, verbose=verbose)
 
+        if db_url == 'sqlite://':
+            # Using in-memory database. Since it is reset after each process
+            # restart, `buildbot upgrade-master` cannot be used (data is not
+            # persistant). Upgrade model here to allow startup to continue.
```
Frodox: Tests a bit fair enough ;D

> buildbot/db/connector.py:131 Wrong spelling of a word 'persistant' in a comment:
> # persistant). Upgrade model here to allow startup to continue.
> Did you mean: ''persistent' or 'persist ant' or 'persist-ant' or 'persistence''? [wrong-spelling-in-comment]

rjarry: Haha, that's cruel :-)

Maybe I should take some english lessons.

rjarry:
Erf there is something I don't understand... How is this change affecting tests with `db_url=mysql://` ?

```
  File "/buildbot/buildbot-job/build/master/buildbot/db/schedulers.py", line 156, in thd
    if row['masterid'] == masterid:
exceptions.TypeError: 'NoneType' object has no attribute '__getitem__'
```
https://nine.buildbot.net/#/builders/15/builds/578

Does that ring a bell to anyone?

wootanis:
Erf there is something I don't understand... How is this change affecting tests with db_url=mysql:// ?

  File "/buildbot/buildbot-job/build/master/buildbot/db/schedulers.py", line 156, in thd
    if row['masterid'] == masterid:
exceptions.TypeError: 'NoneType' object has no attribute '__getitem__'
https://nine.buildbot.net/#/builders/15/builds/578

Does that ring a bell to anyone?

wootanis:
17.06.2017, 01:11, "Robin Jarry" <notifications@github.com>:When using an in-memory sqlite database (db_url='sqlite://'), everything is reset at each process restart. This causes the following error:
Setting up database with URL 'sqlite:'
The Buildmaster database needs to be upgraded before this version of
buildbot can run.  Use the following command-line

    buildbot upgrade-master ...

to upgrade the database, and try starting the buildmaster again.  You may
want to make a backup of your buildmaster before doing so.

Since there is no data persistence, the upgrade-master command would work but everything would be lost when restarting the server.
Handle this special case and initialize the db model at database connection to allow startup to continue.
See the following links for more details:
https://sqlite.org/inmemorydb.html
http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlit…



You can view, comment on, or merge this pull request online at:
  https://github.com/buildbot/buildbot/pull/3333

Commit Summary

  db: allow using in-memory sqlite uri


File Changes

  
    M
    master/buildbot/db/connector.py
    (6)
  


Patch Links:

  https://github.com/buildbot/buildbot/pull/3333.patch
  https://github.com/buildbot/buildbot/pull/3333.diff

rjarry:
I work around the test error but I'm not too happy with the fix...

Could this test `test_setup_check_version_bad` be removed ? It would avoid having an ugly hack.

tardyp:
thanks!

Issue #3339, opened 2017-06-19
canStartBuild example (RemoteCommand) does not work 


nazarewk:
`RemoteCommand` implementation has changed and `canStartBuild` [example](http://docs.buildbot.net/0.9.8/manual/customization.html#canstartbuild-functions) stopped working because `self.step is None` in [here](https://github.com/buildbot/buildbot/blob/a5237a9cc02740e474cb0b3b717ed7296889f708/master/buildbot/process/remotecommand.py#L278)

Frodox:
[permanent link](https://github.com/buildbot/buildbot/blob/a5237a9cc02740e474cb0b3b717ed7296889f708/master/buildbot/process/remotecommand.py#L278)

nazarewk:
thanks, updated both links to be permanent

tardyp:
@nazarewk looks like indeed a patch is needed in cleanup so that it is just a nop if self.step.build is None. Would you like to do this patch?

seankelly:
Fixed in #3343.

PR #3343, opened 2017-06-20
#3339 NOP if RemoteCommand.build is None

nazarewk:
https://github.com/buildbot/buildbot/issues/3339 fix

### DIFF
--- a/master/buildbot/process/remotecommand.py
+++ b/master/buildbot/process/remotecommand.py
@@ -275,7 +275,10 @@ class RemoteCommand(base.RemoteCommandImpl, WorkerAPICompatMixin):
     @defer.inlineCallbacks
     def remoteUpdate(self, update):
         def cleanup(data):
+            if self.step is None:
+                return data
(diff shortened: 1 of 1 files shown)


master/buildbot/process/remotecommand.py
```diff
@@ -275,7 +275,10 @@ def addToLog(self, logname, data):
     @defer.inlineCallbacks
     def remoteUpdate(self, update):
         def cleanup(data):
+            if not self.step.build:
+                return ''
```
tardyp: it should return data in this case

master/buildbot/process/remotecommand.py
```diff
@@ -275,7 +275,10 @@ def addToLog(self, logname, data):
     @defer.inlineCallbacks
     def remoteUpdate(self, update):
         def cleanup(data):
+            if not self.step.build:
```
tardyp: if self.step.build is None

PR #3349, opened 2017-06-21
Add a Python 3.4 builder

rodrigc:


rodrigc:
@tardyp can you add Python 3.4 to the list of Python versions on the buildbots?
There are still some Linux distributions which have Python 3.4 by default.

tardyp:
updated the dockerfile: https://github.com/buildbot/metabbotcfg/pull/86

seankelly:
3.4 builder still failing because of missing Python 3.4. Is there some way to indicate when the Docker image was built to see if it at least happened after the other PR?

tardyp:
dammit. the new docker image now has python 3.4, but it pulled a new version of chrome which is not compatible with our version of selenium :-(

rodrigc:
@tardyp Looks like you've upgraded protractor/selenium, so I think this works now.

PR #3350, opened 2017-06-21
Python 3.4 doesn't support PEP 461 bytes formatting.

rodrigc:


rodrigc:
See: https://github.com/buildbot/buildbot/pull/3349
for adding buildbot support

tardyp:
looks good!

bdbaddog:
Is it worth supporting 3.4 and not set the floor at 3.5?

tardyp:
if 3.4 is only this fix, I would be happy with it, if it makes more distro happy.

rodrigc:
Python 3.4 is still needed because on some Linux distributions that are not that old, Python 3 == 3.4

PR #3357, opened 2017-06-22
Render secrets during service reconfiguration

protatremy:
Secret is rendered when the service reconfig function is called like renderables are during build start.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -454,6 +454,29 @@ class _SecretRenderer(object):
         defer.returnValue(secret_detail.value)
 
 
+class Secret(_SecretRenderer):
+
+    def __repr__(self):
+        return "Secret({0})".format(self.secretKey)
+
+    def __init__(self, secretKey):
+        self.secretKey = secretKey
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, master):
+        secretsSrv = master.namedServices.get("secrets")
+        if not secretsSrv:
+            error_message = "secrets service not started, need to configure" \
+                            " SecretManager in c['services'] to use 'secrets'" \
+                            "in Interpolate"
+            raise KeyError(error_message)
+        credsservice = master.namedServices['secrets']
+        secret_detail = yield credsservice.get(self.secretKey)
+        if secret_detail is None:
+            raise KeyError("secret key %s is not found in any provider" % self.secretKey)
+        defer.returnValue(secret_detail.value)
+
+
 class _SecretIndexer(object):
 
     def __contains__(self, password):
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -68,6 +68,17 @@ class Properties(util.ComparableMixin):
         self._used_secrets = {}
         if kwargs:
             self.update(kwargs, "TEST")
+        self._master = None
+
+    @property
+    def master(self):
+        if self.build is not None:
+            return self.build.master
+        return self._master
+
+    @master.setter
+    def master(self, value):
+        self._master = value
 
     @classmethod
     def fromDict(cls, propDict):
@@ -463,14 +474,14 @@ class Secret(_SecretRenderer):
         self.secretKey = secretKey
 
     @defer.inlineCallbacks
-    def getRenderingFor(self, master):
-        secretsSrv = master.namedServices.get("secrets")
+    def getRenderingFor(self, props):
+        secretsSrv = props.master.namedServices.get("secrets")
         if not secretsSrv:
             error_message = "secrets service not started, need to configure" \
                             " SecretManager in c['services'] to use 'secrets'" \
                             "in Interpolate"
             raise KeyError(error_message)
-        credsservice = master.namedServices['secrets']
+        credsservice = props.master.namedServices['secrets']
         secret_detail = yield credsservice.get(self.secretKey)
         if secret_detail is None:
             raise KeyError("secret key %s is not found in any provider" % self.secretKey)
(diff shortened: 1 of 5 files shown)


master/buildbot/process/properties.py
```diff
@@ -454,6 +454,29 @@ def getRenderingFor(self, properties):
...
+class Secret(_SecretRenderer):
+
+    def __repr__(self):
+        return "Secret({0})".format(self.secretKey)
+
+    def __init__(self, secretKey):
+        self.secretKey = secretKey
+
+    @defer.inlineCallbacks
+    def getRenderingFor(self, master):
```
tardyp: its not a renderable anymore if you give him a master

master/buildbot/test/unit/test_secret_rendered_service.py
```diff
@@ -0,0 +1,64 @@
...
+        self.master = master
+
+
+class FakeServiceUsingSecrets(BuildbotService):
+
+    name = "FakeServiceUsingSecrets"
+    render_secrets = ["secret_to_render"]
+
+    @defer.inlineCallbacks
+    def reconfigService(self, secret):
```
tardyp: def reconfigService(self, secret=None):

master/buildbot/test/unit/test_secret_rendered_service.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, master):
+        super(FakeBuildWithMaster, self).__init__()
+        self.master = master
+
+
+class FakeServiceUsingSecrets(BuildbotService):
+
+    name = "FakeServiceUsingSecrets"
+    render_secrets = ["secret_to_render"]
```
tardyp: +    render_secrets = ["secret"]

master/buildbot/test/unit/test_secret_rendered_service.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @defer.inlineCallbacks
+    def reconfigService(self, secret):
+        self.secret_to_render = secret
```
tardyp: at this point secret must rendered

master/buildbot/test/unit/test_secret_rendered_service.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def setUp(self):
+        self.master = fakemaster.make_master()
+        fakeStorageService = FakeSecretStorage()
+        fakeStorageService.reconfigService(secretdict={"foo": "bar",
+                                                       "other": "value"})
+        self.secretsrv = SecretManager()
+        self.secretsrv.services = [fakeStorageService]
+        self.secretsrv.setServiceParent(self.master)
+        self.srvtest = FakeServiceUsingSecrets(["foo", "other"])
```
tardyp: List of use cases to try

     FakeServiceUsingSecrets(secret=Secret("foo")])

-> reconfigureService is called with secret="bar" 

     FakeServiceUsingSecrets(secret="foo")

-> return the service with secret hardcoded in the master.cfg (reconfigureService is called with secret="foo") 

     FakeServiceUsingSecrets(Secret("foo"))

-> positional arguments cannot work  (reconfigureService is called with secret=Secret("foo")), and the type check will fail)


Nested renderables:

     FakeServiceUsingSecrets(secret=[Secret("foo"), Secret("other")])

-> reconfigureService is called with secret=['bar', 'value'])


For this to work, you just have to use

    def reconfigServiceWithSibling(self, sibling):
        # only reconfigure if sibling is configured differently.
        # sibling == self is using ComparableMixin's implementation
        # only compare compare_attrs
        if self.configured and sibling == self:
            return defer.succeed(None)
        self.configured = True
        p = properties.Properties()
        p.master = self.master
        kwargs = {k: yield p.render(v) for k, v in sibling._config_kwargs}
        return self.reconfigService(*sibling._config_args,
                                    **kwargs)

for this to work, you will have to set a master property in Properties class, and change _SecretRenderable to:

    #  in Properties class:
    @property
    def master(self):
        if self.build is not None:
            return self.build.master
        return self._master

    @master.setter
    def master(self, value):
        self._master = value


- rename to make it API (replace your Secret class)
- use property.master instead of property.getBuild().master

master/buildbot/util/service.py
```diff
@@ -189,19 +191,39 @@ def getConfigDict(self):
...
-                                    **sibling._config_kwargs)
+        # render renderables in parallel
+        # Properties import to resolve cyclic import issue
+        from buildbot.process.properties import Properties
+        p = Properties()
+        p.master = self.master
+        # render renderables in parallel
+        renderables = []
+        kwargs = {}
+        accumulateClassList(self.__class__, 'renderables', renderables)
```
tardyp: why renderables? secrets!

There are some reporters that allows renderable to be configured, and which will render them for each builds. So if they are already rendered at reconfig, the feature will break. That is strange that the unit tests didn't catch this.

master/buildbot/util/service.py
```diff
@@ -189,19 +191,39 @@ def getConfigDict(self):
...
+                kwargs.update({k: v})
+        d = yield self.reconfigService(*sibling._config_args,
+                                       **sibling._config_kwargs)
+        defer.returnValue(d)
 
+    @defer.inlineCallbacks
     def configureService(self):
         # reconfigServiceWithSibling with self, means first configuration
-        return self.reconfigServiceWithSibling(self)
+        d = yield self.reconfigServiceWithSibling(self)
```
tardyp: I would prefer to avoid transforming this into inlineCallbacks since it is not needed. It adds an overhead

master/buildbot/util/service.py
protatremy: done

master/buildbot/util/service.py
protatremy: done

PR #3367, opened 2017-06-23
EC2LatentSlave: spot request state bugfixes and adding tag support

opalmer:
This PR makes a couple of minor bugfixes:

*  Ensure that if the spot request enters the `capacity-not-available` or `capacity-oversubscribed` states that the request is closed by `_wait_for_request()`. The master would already skip to the next worker but it wouldn't close the spot request (which won't expire and counts against a region level limit)
* Create a `_tag_instance()` function and call this for both on demand and spot instances.  Previously spot instances didn't receive tags passed to `__init__()`.

## Contributor Checklist:

* [x] I have updated the unit tests (nothing to update)

### DIFF
--- a/master/buildbot/newsfragments/ec2_latent_worker.bugfix
+++ b/master/buildbot/newsfragments/ec2_latent_worker.bugfix
@@ -1,0 +1,3 @@
+_wait_for_request() would fail to format a log statement due to an invalid type being passed to log.msg (resulting in a broken build)
+_wait_for_request() could leave a spot request open if it entered the capacity-oversubscribed or capacity-not-available states
+Tags are now applies to spot instances too
--- a/master/buildbot/worker/ec2.py
+++ b/master/buildbot/worker/ec2.py
@@ -401,6 +401,17 @@ class EC2LatentWorker(AbstractLatentWorker):
             opts = args[0]
         return dict((k, v) for k, v in iteritems(opts) if v is not None)
 
+    def _tag_instance(self):
+        if not self.tags:
+            return
+
+        self.instance.create_tags(
+            Tags=[
+                {"Key": k, "Value": v}
+                for k, v in iteritems(self.tags)
+            ]
+        )
+
     def _start_instance(self):
         image = self.get_image()
         launch_opts = dict(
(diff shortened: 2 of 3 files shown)


master/buildbot/newsfragments/ec2_latent_worker.bugfix
```diff
@@ -0,0 +1,3 @@
+_wait_for_request() would fail to format a log statement due to an invalid type being passed to log.msg (resulting in a broken build)
```
tardyp: Those sentence will be placed in the middle of the release notes. Please be more user oriented.

    bug fixes in EC2LatentWorker for spot request support.
    Tag support in EC2LatentWorker

master/buildbot/worker/ec2.py
```diff
@@ -523,6 +542,7 @@ def _request_spot_instance(self):
             raise LatentWorkerFailedToSubstantiate()
```
tardyp: these lines are marked uncovered by unit tests. Please add unit tests for it, so that we can keep the code working.

opalmer: Slowly working on it but there's a bit of a problem.

Looking at the existing tests for spot instances they all call `_start_instance()` which is fine except this method calls `self.ec2.create_instance` so the tests, unless I'm missing something obvious, basically don't cover spot instances at all and the existing coverage seems to confirm this:

* https://codecov.io/gh/buildbot/buildbot/src/master/master/buildbot/worker/ec2.py#L471
* https://codecov.io/gh/buildbot/buildbot/src/master/master/buildbot/worker/ec2.py#L491
* https://codecov.io/gh/buildbot/buildbot/src/master/master/buildbot/worker/ec2.py#L564

I did start to work on updating `test_start_spot_instance()` to call `_request_spot_instance()` but found out moto does not implement `describe_spot_price_history()` so we won't be able to get coverage for `_bid_price_from_spot_price_history()`. 

The other issue, which I'd certainly welcome some suggestions on is how to get moto to create a spot request then modify the state of the request so `_wait_for_request()` can return correctly. I tried digging around on moto's own tests and it looks like I might be able to do something like:

```python 
models = get_model('SpotInstanceRequest', 'us-east-1')
# loop over models in a thread and modify the state so _wait_for_request() returns
```

But I'm not sure that's the right approach (or if it will even work) so I'd certainly welcome some suggestions. If I can get `test_start_spot_instance()` working modifying the other tests shouldn't be too difficult which would be great because then spot requests will actually be covered in the tests.

Thoughts?

PR #3368, opened 2017-06-24
Fix bytes/unicode issues in the pollers

### LINKED ISSUE #3366: HGPoller fails with some Mercurial repositories
HGPoller crashes on some repositories when using Python3 based Buildbot. Using the same master.cfg file, Python2 based Buildbot masters don't experience the same crash. I tested this with two different Mercurial repositories on the Python3 based Buildbot and only one of those experienced the crash - I am not certain how to explain that detail, but the permissions on each repository were configured identically.

[master.cfg.txt](https://github.com/buildbot/buildbot/files/1099078/master.cfg.txt)
[err.txt](https://github.com/buildbot/buildbot/files/1099083/err.txt)

rodrigc:
Fixes #3366 #3375

### DIFF
--- a/master/buildbot/changes/p4poller.py
+++ b/master/buildbot/changes/p4poller.py
@@ -235,6 +235,13 @@ class P4Source(base.PollingChangeSource, util.ComparableMixin):
             args.extend(['-m', '1', '%s...' % (self.p4base,)])
 
         result = yield self._get_process_output(args)
+        # decode the result from its designated encoding
+        try:
+            result = bytes2unicode(result, self.encoding)
+        except UnicodeError as ex:
+            log.msg(u"Warning: cannot fully decode {} in {}".format(
+                    repr(result), self.encoding))
+            result = bytes2unicode(result, encoding=self.encoding, errors="replace")
 
         last_change = self.last_change
         changelists = []
--- a/master/buildbot/test/unit/test_changes_hgpoller.py
+++ b/master/buildbot/test/unit/test_changes_hgpoller.py
@@ -26,6 +26,8 @@ from buildbot.test.util import changesource
 from buildbot.test.util import gpo
 
 ENVIRON_2116_KEY = 'TEST_THAT_ENVIRONMENT_GETS_PASSED_TO_SUBPROCESSES'
+LINESEP_BYTES = os.linesep.encode("ascii")
+PATHSEP_BYTES = os.pathsep.encode("ascii")
 
 
 class TestHgPoller(gpo.GetProcessOutputMixin,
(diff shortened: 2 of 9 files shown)


rodrigc:
@stevegcarpenter I have a better fix for #3366 .  Would you be able to test this by installing buildbot master from my branch with:

```
pip install -e "git+https://github.com/rodrigc/buildbot.git@poller-bytes#egg=buildbot&subdirectory=master"
```

and providing feedback?

master/buildbot/changes/p4poller.py
```diff
@@ -235,6 +235,16 @@ def _poll(self):
             args.extend(['-m', '1', '%s...' % (self.p4base,)])
 
         result = yield self._get_process_output(args)
+        # decode the result from its designated encoding
+        try:
+            result = bytes2unicode(result, self.encoding)
```
tardyp: shouldn't we rather use replace instead?

master/buildbot/test/unit/test_changes_hgpoller.py
```diff
@@ -26,6 +26,8 @@
 from buildbot.test.util import gpo
 
 ENVIRON_2116_KEY = 'TEST_THAT_ENVIRONMENT_GETS_PASSED_TO_SUBPROCESSES'
+LINESEP_BYTES = os.linesep.encode("utf-8")
```
tardyp: I think default ascii encoding is enough for those :)

master/buildbot/changes/p4poller.py
rodrigc: I don't think so.  Elsewhere in this file, they use the same pattern of doing a decode and then catching a UnicodeError.

master/buildbot/test/unit/test_changes_hgpoller.py
rodrigc: OK

master/buildbot/changes/p4poller.py
tardyp: Having the same pattern in the file is not always a good reason.
What will happen here is that the change is just ignored. I think it is better, as an overall best practice to be more gentler when there are errors. Try our best to still injest the change even if it looks suspicious instead of just rejecting it.

I think mode="replace" is good at that.

maybe we should do:

```
try:
    result = bytes2unicode(result, self.encoding)
except UnicodeError:
   log.msg("Warning: cannot fully decode {} in {}".format(result, self.encoding))
   result = bytes2unicode(result, self.encoding, mode="replace")
```

rodrigc: OK, I made the change.

PR #3371, opened 2017-06-24
Roles from GitHub teams

s0undt3ch:
Initial draft on getting GH team membership for the authenticated user and use those teams for roles

Need to cleanup the code a bit and add tests, however I just wanted to get this PR opened to start addressing review change requests.
## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_www_oauth.py
+++ b/master/buildbot/test/unit/test_www_oauth.py
@@ -31,6 +31,7 @@ from twisted.web.resource import Resource
 from twisted.web.server import Site
 
 from buildbot.test.util import www
+from buildbot.test.util.config import ConfigErrorsMixin
 
 try:
     import requests
@@ -52,7 +53,7 @@ class FakeResponse(object):
         pass
 
 
-class OAuth2Auth(www.WwwTestMixin, unittest.TestCase):
+class OAuth2Auth(www.WwwTestMixin, ConfigErrorsMixin, unittest.TestCase):
 
     def setUp(self):
         if requests is None:
@@ -64,13 +65,18 @@ class OAuth2Auth(www.WwwTestMixin, unittest.TestCase):
 
         self.googleAuth = oauth2.GoogleAuth("ggclientID", "clientSECRET")
         self.githubAuth = oauth2.GitHubAuth("ghclientID", "clientSECRET")
+        self.githubAuth_v4 = oauth2.GitHubAuth(
+            "ghclientID", "clientSECRET", apiVersion=4)
+        self.githubAuth_v4_teams = oauth2.GitHubAuth(
+            "ghclientID", "clientSECRET", apiVersion=4, getTeamsMembership=True)
         self.githubAuthEnt = oauth2.GitHubAuth(
             "ghclientID", "clientSECRET", serverURL="https://git.corp.fakecorp.com")
         self.gitlabAuth = oauth2.GitLabAuth(
             "https://gitlab.test/", "glclientID", "clientSECRET")
         self.bitbucketAuth = oauth2.BitbucketAuth("bbclientID", "clientSECRET")
 
-        for auth in [self.googleAuth, self.githubAuth, self.githubAuthEnt, self.gitlabAuth, self.bitbucketAuth]:
+        for auth in [self.googleAuth, self.githubAuth, self.githubAuth_v4, self.githubAuth_v4_teams,
+                     self.githubAuthEnt, self.gitlabAuth, self.bitbucketAuth]:
             self._master = master = self.make_master(url='h:/a/b/', auth=auth)
             auth.reconfigAuth(master, master.config)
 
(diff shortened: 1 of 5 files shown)


master/buildbot/www/authz/roles.py
```diff
@@ -86,3 +86,18 @@ def getRolesFromUser(self, userDetails):
...
+class RolesFromTeams(RolesFromBase):
+
+    def __init__(self, teamPrefix=""):
+        RolesFromBase.__init__(self)
+        self.teamPrefix = teamPrefix
+
+    def getRolesFromUser(self, userDetails):
+        roles = []
+        if 'teams' in userDetails:
+            for team in userDetails['teams']:
```
tardyp: you can simplify the two lines:

    for team in userDetails.get('teams', []):

master/buildbot/www/oauth2.py
```diff
@@ -206,22 +211,104 @@ def __init__(self,
 
             self.authUri = '{0}/login/oauth/authorize'.format(serverURL)
             self.tokenUri = '{0}/login/oauth/access_token'.format(serverURL)
-            self.resourceEndpoint = '{0}/api/v3'.format(serverURL)
+        self.serverURL = serverURL or self.resourceEndpoint
+
+        if not isinstance(apiVersion, int):
```
tardyp: this is not needed. the second check is a subset of this check.

master/buildbot/www/oauth2.py
```diff
@@ -206,22 +211,104 @@ def __init__(self,
...
+                config.error(
+                    'Retrieving team membership information using GitHubAuth is only '
+                    'possible using GitHub api v4.')
+            self.resourceEndpoint = '{0}/api/v3'.format(self.serverURL)
+        else:
+            self.resourceEndpoint = '{0}/graphql'.format(self.serverURL)
+        self.getTeamsMembership = getTeamsMembership
+
+    def post(self, session, query):
+        log.info('POST: {endpoint} -> DATA:\n----\n{data}\n----', endpoint=self.resourceEndpoint, data=query)
```
tardyp: We should put those logs under a debug flag

master/buildbot/www/oauth2.py
```diff
@@ -206,22 +211,104 @@ def __init__(self,
...
                 break
         orgs = self.get(c, '/user/orgs')
-
         return dict(full_name=user['name'],
                     email=user['email'],
                     username=user['login'],
                     groups=[org['login'] for org in orgs])
 
+    def getUserInfoFromOAuthClient_v4(self, c):
+        graphql_query = textwrap.dedent('''
```
tardyp: please add some unit tests for this method

master/buildbot/www/oauth2.py
```diff
@@ -206,22 +211,104 @@ def __init__(self,
...
+        data = data['data']
+        log.info('Response: {response}', response=data)
+        user_info = dict(full_name=data['viewer']['name'],
+                         email=data['viewer']['email'],
+                         username=data['viewer']['login'],
+                         groups=[org['node']['login'] for org in
+                                 data['viewer']['organizations']['edges']]
+        )
+        if self.getTeamsMembership:
+            graphql_query = '''
```
tardyp: we already have jinja as a dependency. shall we use jinja two make this code more readable?

master/buildbot/www/authz/roles.py
s0undt3ch: True

master/buildbot/www/oauth2.py
s0undt3ch: Ok

master/buildbot/www/oauth2.py
s0undt3ch: Will do

master/buildbot/www/oauth2.py
s0undt3ch: Invoking jinja just for this? It will be less performant but sure.

Render a string or create a template file and render that?
Do we have helper utilities to prep jinja for rendering?

tardyp: Jinja compiled templates is as performant as pure python. It actually compiles into very similar code as you did.
You would have to precompile that template at init time

s0undt3ch: OK. That makes more sense.

master/buildbot/www/oauth2.py
s0undt3ch: Added, for both v4 and v4 with team membership fetching.

s0undt3ch:
Still need to add tests for `RolesFromTeams`

s0undt3ch:
News fragment added and documentation updated.

master/buildbot/test/unit/test_www_oauth.py
```diff
@@ -210,6 +216,144 @@ def test_GithubAcceptToken(self):
...
+                        }
+                    },
+                }
+            }
+        ])
+        res = yield self.githubAuth_v4_teams.acceptToken("TOK3N")
+        self.assertEqual({'email': 'bar@foo',
+                          'username': 'bar',
+                          'groups': ["hello", "grp"],
+                          'teams': [
```
s0undt3ch: You tell me. The idea was a clean separation. If you wish to have `grp` as a role, you'd pass an empty group prefix and we'd end up with a `grp/committers` role too for example.

tardyp: The thing is the notion of Team is very bound to Github, and the only auth plugin which would implement this. This is why I would prefer not to introduce a new concept unless it is necessary.

I think '/' is already a clean separation. You then can use  regular expression to give rights to grp/* or just grp or grp/foo/*

s0undt3ch:
Should be good to take another look.

common/code_spelling_ignore_words.txt
```diff
@@ -569,6 +569,7 @@ gotperspective
 gotrootobject
 gpo
 gracefulshutdown
+graphql
```
tardyp: this one is for pylint..

the one you need to fix is:
https://github.com/buildbot/buildbot/blob/master/master/docs/spelling_wordlist.txt#L1

tardyp:
:1st_place_medal:

PR #3381, opened 2017-06-26
grid_view: display most recent changes on the left

rjarry:
Add a new setting allowing to change the ordering of changes. By default, most recent changes are displayed on the left (older ones on the right).

### DIFF
--- a/www/base/src/app/workers/workers.route.coffee
+++ b/www/base/src/app/workers/workers.route.coffee
@@ -33,4 +33,9 @@ class State extends Config
                 name:'show_old_workers'
                 caption:'Show old workers'
                 default_value: false
+            ,
+                type:'bool'
+                name:'showWorkerBuilders'
+                caption:'Show list of builders for each worker (can take a lot of time)'
+                default_value: false
             ]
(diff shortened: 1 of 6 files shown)


www/grid_view/src/module/grid.controller.coffee
```diff
@@ -28,6 +28,7 @@ class Grid extends Controller
         @changeFetchLimit = settings.changeFetchLimit.value
         @buildFetchLimit = settings.buildFetchLimit.value
         @compactChanges = settings.compactChanges.value
+        @rightToLeft = settings.rightToLeft.value
 
         @buildsets = @data.getBuildsets(
             limit: @changeFetchLimit
```
tardyp: Here you should rather take @buildFetchLimit. If the trigger step is used, the number of buildset is more on the order of the number of build rather than the order of change.
better to many than too few.

www/base/src/app/workers/workers.route.coffee
```diff
@@ -33,4 +33,9 @@ class State extends Config
                 name:'show_old_workers'
                 caption:'Show old workers'
                 default_value: false
+            ,
+                type:'bool'
+                name:'showWorkerBuilders'
+                caption:'Show list of builders for each worker (can take a lot of time)'
+                default_value: false
```
tardyp: I would need to set this value to true by default or fix the e2e tests

PR #3388, opened 2017-06-28
Fix GerritVerifyStatusPush()'s request body to match Gerrit's expectations

opalmer:
## Contributor Checklist:

* [x] I have updated the appropriate documentation (no change)

### DIFF
--- a/master/buildbot/reporters/gerrit_verify_status.py
+++ b/master/buildbot/reporters/gerrit_verify_status.py
@@ -16,6 +16,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+from six.moves import http_client
 from twisted.internet import defer
 from twisted.python import failure
 
@@ -214,7 +215,7 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
         changes = yield self.getGerritChanges(props)
         for change in changes:
             try:
-                yield self.createStatus(
+                response = yield self.createStatus(
                     change['change_id'],
                     change['revision_id'],
                     name,
@@ -226,6 +227,13 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
                     reporter=reporter,
                     category=category,
                     duration=duration)
+
+                # The verify-status plugin will return 204 NO CONTENT
+                # on success.
+                if response.code != http_client.NO_CONTENT:
+                    body = yield response.text()
+                    raise ValueError('Expected 204 NO CONTENT. Response was %s: %s' % (response.code, body))
+
             except Exception:
                 log.failure(
                     'Failed to send status!', failure=failure.Failure())
--- a/master/buildbot/reporters/gerrit_verify_status.py
+++ b/master/buildbot/reporters/gerrit_verify_status.py
@@ -136,6 +136,8 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
         if duration is not None:
             payload['duration'] = duration
 
+        payload = {'verifications': {name: payload}}
+
         if self._verbose:
             log.debug(
                 'Sending Gerrit status for {change_id}/{revision_id}: data={data}',
--- a/master/buildbot/reporters/gerrit_verify_status.py
+++ b/master/buildbot/reporters/gerrit_verify_status.py
@@ -233,7 +233,10 @@ class GerritVerifyStatusPush(http.HttpStatusPushBase):
                 # The verify-status plugin will return 204 NO CONTENT
                 # on success.
                 if response.code != http_client.NO_CONTENT:
-                    body = yield response.text()
+                    try:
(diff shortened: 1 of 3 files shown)


opalmer:
@tardyp, looks like `format.format` doesn't exist in Twisted 14.0.2? I can go digging to fix it tomorrow unless you happen to know the correct call.

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -226,6 +229,16 @@ def send(self, build):
                     reporter=reporter,
                     category=category,
                     duration=duration)
+
+                # The verify-status plugin will return 204 NO CONTENT
+                # on success.
+                if response.code != http_client.NO_CONTENT:
+                    try:
+                        body = yield response.text()
```
tardyp: response.text is not an API so it will always raise AttributeError. what are you trying to do?

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -135,6 +136,8 @@ def createStatus(self,
         if duration is not None:
             payload['duration'] = duration
 
+        payload = {'verifications': {name: payload}}
```
tardyp: gasp.. I am pretty sure I did test that e2e. Looks like I didn't :(

master/buildbot/reporters/gerrit_verify_status.py
tardyp: not sure why you first try response.text(). For me it does not correspond to any API

opalmer: It works for me and does appear to exist according to the docs:

https://treq.readthedocs.io/en/latest/api.html#treq.response._Response.text

Maybe `.text()` was introduced in a later version of treq? Happy to change this to response.content() only if you think it's appropriate.

master/buildbot/reporters/gerrit_verify_status.py
opalmer: No worries, if it did work then I'm willing to bet the API changed or something.

I'm a little curious how you did your end to end testing? I've got another project where I'm trying to get [gerrit setup with Docker](https://github.com/andygrunwald/go-gerrit/issues/11) and I ended up developing [gerrittest](https://github.com/opalmer/gerrittest) as a result. Might be useful for buildbot for these kinds of changes.

master/buildbot/reporters/gerrit_verify_status.py
rodrigc: `response.text()` returns unicode, while `response.content` returns bytes.
It's probably better to leave this as `response.text()` especially for Python 3,
otherwise we get into the bytes/unicode issues.

master/buildbot/reporters/gerrit_verify_status.py
tardyp: I looked at the history of verify status plugin for several month, and could not see a version with current code's payload.

master/buildbot/reporters/gerrit_verify_status.py
tardyp: see: http://docs.buildbot.net/current/developer/utils.html#buildbot.util.service.IHTTPResponse
Note
IHTTPResponse is a subset of treq Response API described here The API it is voluntarily minimalistic and reflects what is tested and reliable to use with the three backends (including fake). The api is a subset of the treq API, which is itself a superset of twisted IResponse API. treq is thus implemented as passthrough.

Notably:

There is no api to automatically decode content, as this is not implemented the same in both backends.

so please use .content(), and decode manually.

opalmer: Understood, fixed in 6e3583c6566ed1b9c5709169d0adca97a6ae6294.

master/buildbot/util/logger.py
```diff
@@ -35,5 +35,10 @@ def _log(self, format, *args, **kwargs):
         critical = _log
 
         def failure(self, format, failure, *args, **kwargs):
-            log.error(failure, format.format(args, **kwargs))
```
opalmer: Updated in 1e9a5432a6a9ac314186830c138a487e8ab0fdc3. This should solve the AttributeError I was getting initially.

tardyp:
still some unit tests to fix (and pylint)

master/buildbot/reporters/gerrit_verify_status.py
```diff
@@ -135,6 +138,8 @@ def createStatus(self,
...
```
sa2ajj: I'm sorry I do not quite understand: you seem to overwrite whatever `payload` has so far...

seankelly:
@tardyp You marked as okay but now it's "stale", should this be merged still?

PR #3415, opened 2017-07-05
Fix renderability of a build step name

rkashapov:
According to the [documentation](http://docs.buildbot.net/current/manual/cfg-buildsteps.html#common-parameters), a build step name may be a renderable object. But if I pass an Interpolate instance as the name argument I will get configuration error that says "BuildStep name must be a string". 
## Contributor Checklist:


Frodox:
it should have newsfragment I think, it is cool enough feature :)

PR #3418, opened 2017-07-06
Fix python 3 issue with displaying the properties in API call

danieltcv:
When running buildbot with python3, the following API call:                     

/api/v2/builders/<int>/builds&property=*                                        

or any property filter for that matter does not return the expected set of      
filters but instead returns an empty dictionary {}.

The fix is to compare the right types of variables - string vs binary.
In python 2 the b'' prefix is ignored so it shouldn't break changes there. 

## Contributor Checklist:


rodrigc:
testPipeString is failing after this change.  Can you fix the test as part of your commit?

danieltcv:
@rodrigc done

master/buildbot/newsfragments/.bugfix
```diff
@@ -0,0 +1,10 @@
+When running buildbot with python3, the following API call:
```
tardyp: the news fragment should be a single sentence. you can point to :issue:`3418` for people who would like more details.

PR #3421, opened 2017-07-07
Fix ShellArg arguments being rendered only once

ntsiron:
ShellArg arguments were rendered only once, the first time the object
was rendered. All subsequent renderings of a ShellArg instance, i.e. all
subsequent calls to the getRenderingFor() method of the instance,
computed always the same values for its arguments.

This happened because the first time a ShellArg instance was rendered it
substituted it's attributes (even the renderable ones) with their renderings.

For example in the following code snippet: 

```
f = util.BuildFactory()
f.addStep(steps.ShellSequence(
    commands=[
        util.ShellArg(
            command=[
                "echo",
                util.Interpolate("Build number: %(prop:buildnumber)s")
            ],
            logfile="stdio")
    ]))
```

The output of the echo command is the same for every build. It is the number of the first build ran with the the builder using this build factory, when the ShellArg() instance was first rendered.

The fix changes the getRenderingFor() method of ShellArg() to return a rendered copy of itself instead of substituting its own attributes. 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/shellsequence.py
+++ b/master/buildbot/steps/shellsequence.py
@@ -25,6 +25,7 @@ from buildbot import config
 from buildbot.process import buildstep
 from buildbot.process import results
 
+import copy
 
 class ShellArg(results.ResultComputingConfigMixin):
     publicAttributes = (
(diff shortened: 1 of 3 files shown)


master/buildbot/steps/shellsequence.py
```diff
@@ -63,10 +65,11 @@ def validateAttributes(self):
 
     @defer.inlineCallbacks
     def getRenderingFor(self, build):
+        rv = copy.copy(self)
         for p_attr in self.publicAttributes:
             res = yield build.render(getattr(self, p_attr))
-            setattr(self, p_attr, res)
-        defer.returnValue(self)
+            setattr(rv, p_attr, res)
+        defer.returnValue(rv)
```
tardyp: Fix looks good. We require the writing of unit tests for bugs fixed.

ntsiron: Ok i will write the unit test.

PR #3431, opened 2017-07-10
Add dummy primary keys in tables lacking them

vriera:
Some tables from migrate_utils.py and model.py don't have a primary
key. Because of that some RDBMS complain when executing...

  buildbot upgrade-master /path/to/master/directory/

Having primary keys in tables is highly recommended [1], so let's fix
that issue by adding some dummy ones.

1: https://stackoverflow.com/questions/840162/should-each-and-every-table-have-a-primary-key

Signed-off-by: Vicente Olivert Riera <Vincent.Riera@imgtec.com>

### DIFF
--- a/master/buildbot/db/model.py
+++ b/master/buildbot/db/model.py
@@ -105,9 +105,9 @@ class Model(base.DBConnectorComponent):
     buildrequest_claims = sautils.Table(
         'buildrequest_claims', metadata,
         sa.Column('brid', sa.Integer, sa.ForeignKey('buildrequests.id'),
-                  nullable=False),
+                  primary_key=True),
         sa.Column('masterid', sa.Integer, sa.ForeignKey('masters.id'),
-                  index=True, nullable=True),
+                  index=True, nullable=False),
         sa.Column('claimed_at', sa.Integer, nullable=False),
     )
 
@@ -117,8 +117,8 @@ class Model(base.DBConnectorComponent):
     build_properties = sautils.Table(
         'build_properties', metadata,
         sa.Column('buildid', sa.Integer, sa.ForeignKey('builds.id'),
-                  nullable=False),
-        sa.Column('name', sa.String(256), nullable=False),
+                  primary_key=True),
+        sa.Column('name', sa.String(256), primary_key=True),
         # JSON encoded value
         sa.Column('value', sa.Text, nullable=False),
         sa.Column('source', sa.String(256), nullable=False),
@@ -185,10 +185,11 @@ class Model(base.DBConnectorComponent):
 
     logchunks = sautils.Table(
         'logchunks', metadata,
-        sa.Column('logid', sa.Integer, sa.ForeignKey('logs.id')),
+        sa.Column('logid', sa.Integer, sa.ForeignKey('logs.id'),
+                  primary_key=True),
         # 0-based line number range in this chunk (inclusive); note that for
         # HTML logs, this counts lines of HTML, not lines of rendered output
-        sa.Column('first_line', sa.Integer, nullable=False),
+        sa.Column('first_line', sa.Integer, primary_key=True),
         sa.Column('last_line', sa.Integer, nullable=False),
         # log contents, including a terminating newline, encoded in utf-8 or,
         # if 'compressed' is not 0, compressed with gzip, bzip2 or lz4
@@ -202,8 +203,8 @@ class Model(base.DBConnectorComponent):
     buildset_properties = sautils.Table(
         'buildset_properties', metadata,
         sa.Column('buildsetid', sa.Integer, sa.ForeignKey('buildsets.id'),
-                  nullable=False),
-        sa.Column('property_name', sa.String(256), nullable=False),
+                  primary_key=True),
+        sa.Column('property_name', sa.String(256), primary_key=True),
         # JSON-encoded tuple of (value, source)
         sa.Column('property_value', sa.Text, nullable=False),
     )
@@ -304,6 +305,7 @@ class Model(base.DBConnectorComponent):
     # Files touched in changes
     change_files = sautils.Table(
         'change_files', metadata,
+        sa.Column('changefileid', sa.Integer, primary_key=True, autoincrement=True),
         sa.Column('changeid', sa.Integer, sa.ForeignKey('changes.changeid'),
                   nullable=False),
         sa.Column('filename', sa.String(1024), nullable=False),
@@ -313,8 +315,8 @@ class Model(base.DBConnectorComponent):
     change_properties = sautils.Table(
         'change_properties', metadata,
         sa.Column('changeid', sa.Integer, sa.ForeignKey('changes.changeid'),
-                  nullable=False),
-        sa.Column('property_name', sa.String(256), nullable=False),
+                  primary_key=True),
+        sa.Column('property_name', sa.String(256), primary_key=True),
         # JSON-encoded tuple of (value, source)
         sa.Column('property_value', sa.Text, nullable=False),
     )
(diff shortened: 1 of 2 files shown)


master/buildbot/db/model.py
```diff
@@ -104,6 +104,7 @@ class Model(base.DBConnectorComponent):
     # claim is made by the master referenced by masterid.
     buildrequest_claims = sautils.Table(
         'buildrequest_claims', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
         sa.Column('brid', sa.Integer, sa.ForeignKey('buildrequests.id'),
```
tardyp: brid is the primary key for that table

master/buildbot/db/model.py
```diff
@@ -116,6 +117,7 @@ class Model(base.DBConnectorComponent):
     # This table contains the build properties
     build_properties = sautils.Table(
         'build_properties', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: primary key here is buildid + name

master/buildbot/db/model.py
```diff
@@ -185,6 +187,7 @@ class Model(base.DBConnectorComponent):
 
     logchunks = sautils.Table(
         'logchunks', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: primary is logid + first_line

master/buildbot/db/model.py
```diff
@@ -201,6 +204,7 @@ class Model(base.DBConnectorComponent):
     # This table contains input properties for buildsets
     buildset_properties = sautils.Table(
         'buildset_properties', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: buildsetid + property_name

master/buildbot/db/model.py
```diff
@@ -304,6 +308,7 @@ class Model(base.DBConnectorComponent):
     # Files touched in changes
     change_files = sautils.Table(
         'change_files', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: changeid + filename

master/buildbot/db/model.py
```diff
@@ -312,6 +317,7 @@ class Model(base.DBConnectorComponent):
     # Properties for changes
     change_properties = sautils.Table(
         'change_properties', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: changeid + property_name

master/buildbot/db/model.py
```diff
@@ -324,6 +330,7 @@ class Model(base.DBConnectorComponent):
     # and committer, for example.
     change_users = sautils.Table(
         "change_users", metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: changeid + uid

master/buildbot/db/model.py
```diff
@@ -498,6 +505,7 @@ class Model(base.DBConnectorComponent):
     # the change.
     scheduler_changes = sautils.Table(
         'scheduler_changes', metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: schedulerid + changeid

master/buildbot/db/model.py
```diff
@@ -569,6 +577,7 @@ class Model(base.DBConnectorComponent):
     # and the value is a JSON string.
     object_state = sautils.Table(
         "object_state", metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: objectid + name

master/buildbot/db/model.py
```diff
@@ -601,6 +610,7 @@ class Model(base.DBConnectorComponent):
     # particular interface - a version-control system, status plugin, etc.
     users_info = sautils.Table(
         "users_info", metadata,
+        sa.Column('dummy', sa.Integer, primary_key=True, autoincrement=True),
```
tardyp: uid + attr_type

vriera:
Hi @tardyp , the purpose of this pull request wasn't to be accepted as it is. Now I know how the primary keys should be in those tables so I can send a proper pull request. I didn't wrote the model so I wasn't sure which column should be the primary key. Thanks for the info.

master/buildbot/db/model.py
vriera: Do you mean brid + masterid?

vriera: The buildrequest_claims table is weird. If you really want the brid field to be the primary key, then having that table is not necessary since all fields on it could be moved to the buildrequests table.

brid + masterid doesn't make sense to be a composite primary key as masterid can be NULL.

master/buildbot/db/model.py
vriera: filename is a VARCHAR(1024), that is too long to be a (primary) key:

> Got fatal OperationalError on DB
> Traceback (most recent call last):
> Failure: sqlalchemy.exc.OperationalError: (_mysql_exceptions.OperationalError) (1071, 'Specified key was too long; max key length is 3072 bytes') [SQL: u'\nCREATE TABLE change_files (\n\tchangeid INTEGER NOT NULL, \n\tfilename VARCHAR(1024) NOT NULL, \n\tPRIMARY KEY (changeid, filename), \n\tFOREIGN KEY(changeid) REFERENCES changes (changeid)\n)CHARACTER SET utf8\n\n']

https://dev.mysql.com/doc/refman/5.7/en/innodb-restrictions.html

master/buildbot/db/model.py
tardyp: Indeed, it is weird, but it is quite important.

The reason why there is this table is to implement the buildrequest claiming in an atomic way. 

In a multimaster environment, a buildrequest is created in the buildrequest table, and then a message will be sent to all masters to inform them about it.

Each master capable of building this request will then try to insert an entry in the claim table, the rdbms guarantees that only one master will manage to insert that entry. The other master will get that query cancelled, and then it will know that it shouldn't build this request as another master did take it.

master/buildbot/db/model.py
tardyp: probably in that case, we can add an autoincrement changefileid.

tardyp:
> The purpose of this pull request wasn't to be accepted as it is. Now I know how the primary keys should be in those tables so I can send a proper pull request.
Sounds good then! thanks for working on it!

master/buildbot/db/model.py
vriera: Still, that table is not necessary. Look this workflow:

1. The buildrequest is created in the buildrequest table, with the masterid column as NULL.
2. A message will be sent to all masters to inform them about it.
3. Each master capable of building this request will then try to update the masterid column only if is NULL, so other masters will not be able to update it anymore.

vriera: If you still want to keep having a buildrequest_claims, then the masterid column shouldn't be null. First of all because the masterid colum is a foreign key that references a primary key in the buildrequest table, and primary keys cannot be null, so it will never reference a null value. And second because the combination of brid and master id are both unique and identify a claim, so they should be a composite primary key.

When a master inserts a row in buildrequest_claims, it inserts the brid and the masterid both at the same time, so masterid already has a non-null value.

Do you prefer this approach for my next spin of this pull request?

vriera:
New PR sent: https://github.com/buildbot/buildbot/pull/3432

PR #3467, opened 2017-07-23
Addition of a KubeLatentWorker

Zempashi:
Third attempt of Kubernetes Latent Worker.

Splitting the connection to kubernetes in an util module and the LatentWorker in one another.

I need some advice to find a correct solution for kube_extra_spec (param to patch the job spec for kubernetes): I have tried a spec_merge method that merge dict and append to list, but almost unusable.
May be something like jsonpatch (http://jsonpatch.com/), but the python lib ( https://github.com/stefankoegl/python-json-patch) is not compatible with python3.6 and it would be nice to avoid an additional library.

I really would like to have the user provide a complete job template and find a way to add image, password, masterFQDN, etc...

I soon try with a callable.


seankelly:
I think `KubeClientService` should be part of the same file as the worker.

tardyp:
> I need some advice to find a correct solution for kube_extra_spec (param to patch the job spec for kubernetes): 

I think the nodejs community has a great tool for that https://lodash.com/docs#merge which is used everywhere to combine default json config together with other config.

We have a python implementation of the algorithm (MIT, so that's compatible with GPLv2):
https://github.com/dgilland/pydash/blob/develop/pydash/objects.py#L867
perhaps we don't want to depends on pydash, but we could pick the code and tests into buildbot.util

I think this is a problem common to docker/hyper/marathon, and probably other latents. So maybe we should include it in buildbot.

Zempashi:
@seankelly that was for the ability to use in other part of buildbot (for exemple janitor can get list of active master based on kubernetes discovery)

@tardyp good idea, I was looking into this kind of library. I'll give it a try

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+
+    kube_dir = '/var/run/secrets/kubernetes.io/serviceaccount/'
+
+    kube_namespace_file = os.path.join(kube_dir, 'namespace')
+    kube_token_file = os.path.join(kube_dir, 'token')
+    kube_cert_file = os.path.join(kube_dir, 'ca.crt')
+
+    config_loader_list = ['kube_config_loader', 'incluster_config_loader']
+
+    def __init__(self,
```
tardyp: Actually, this class is only a config loader.

You can call it KubernetesConfigLoader(), and have a makeHttpClientService() method.

so that client can do:
```
self._kube_config = KubernetesConfigLoader(...)
self._kube = self._kube_config.makeHttpClientService()
self._kube.get("...".format(namespace=self._kube_config.namespace))
```
With this method, you can use the fake httpclient service easily.
You can also patch the configloader for tests to skip configuration details

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+                 config_method=None):
+        self.kube_config_file = kubeconfig or self.default_kubeconfig
+        self._cert_file = None
+        url, req_kwargs, namespace = self.guess_config(config_method)
+        self._cert = req_kwargs.pop('cert', None)
+        self._verify = req_kwargs.pop('verify', None)
+        HTTPClientService.__init__(self, url, **req_kwargs)
+        self.namespace = namespace
+
+    def guess_config(self, config_method=None):
```
tardyp: this method is highly untested :)

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+                config_loader_iter = iter((config_method,))
+        for config in config_loader_iter:
+            if callable(config):
+                config_loader = config
+            else:
+                try:
+                    config_loader = getattr(self, config)
+                except AttributeError:
+                    raise ValueError(
+                        "Unknown config method '{0}'. "
```
tardyp: don't use {0} formatting, rather use {} as py2.6 is not supported

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+                        "Only {1} are available.".format(
+                            config, self.config_loader_list
+                        )
+                    )
+            if not config_loader:
+                continue
+            try:
+                return config_loader()
+            except KubeClientConfigException:
+                continue
```
tardyp: what if there is a syntax error in my config? this code won't allow to debug that.

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+            url = cluster['server']
+        except KeyError:
+            raise ValueError(
+                "cluster config misses mandatory value: 'cluster.server'"
+            )
+        skip_ssl = cluster.get('insecure-skip-tls-verify', False)
+        cert_data = cluster.get('certificate-authority-data', None)
+        cert_path = cluster.get('certificate-authority', None)
+        if cert_data:
+            tmp = tempfile.NamedTemporaryFile(delete=False)
```
tardyp: I would put that file in ~/.kube/<certsha>.crt and don't bother with deleting it.

master/buildbot/worker/kubernetes.py
tardyp: Also, having this a SharedService makes the old service reused if the parameters are not changed.
This means that the config will not be reloaded on buildbot reconfig!

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+        url = '/apis/batch/v1/namespaces/{namespace}/jobs'.format(
+            namespace=self._kube.namespace
+        )
+        res = yield self._kube.post(url, json=job_spec)
+        if res.code != 201:
+            raise LatentWorkerFailedToSubstantiate(
+                "Unable to create Kubernetes job %s" % res.code
+            )
+        res_json = yield res.json()
```
tardyp:
```
self.instance = yield res.json()
```

master/buildbot/worker/kubernetes.py
```diff
@@ -0,0 +1,368 @@
...
+    instance = None
+
+    kubeclientservice_args = ['kubeconfig', 'config_method']
+
+    def default_job_spec(self):
+        return {
+            "apiVersion": "batch/v1",
+            "kind": "Job",
+            "metadata": {
+                "name": self.getContainerName()
```
tardyp: shouldn't a job name change between each instanciation?

tardyp:
Hi @Zempashi I am currently looking at https://github.com/kubernetes-incubator/client-python and I think it is very well designed. We should definitly depend on it at least for configuration parsing.

Maybe they also have good configuration merging.

tardyp:
FWIW, I am gonna unbury this. As hyper.sh is going down, we need a new source of LatentWorker for our own CI. kube looks like a good cloud agnostic way of sourcing and autoscaling docker workers.

Issue #3474, opened 2017-07-26
Builders view: Tag selected: "Back" button doesn't work as expected


jangmarker:
Steps to reproduce:
1. open https://nine.buildbot.net/#/builders
2. click the "buildbot" tag (now the URL is https://nine.buildbot.net/#/builders?tags=%2Bbuildbot)
3. press the browser's "back" button

Actual result:
1. URL in changed back to https://nine.buildbot.net/#/builders
2. builders are still filtered by the "buildbot" tag

Expected result:
1. is ok
2. builders are not filtered by the tag anymore

Issue #3473, opened 2017-07-26
UI: Builders view: Tags don't indicate that that they can be clicked


jangmarker:
Actual:
![screenshot_20170726_151838](https://user-images.githubusercontent.com/6429953/28609109-e0bfe2d8-71a8-11e7-9c9a-e619e40ee397.png)

Expected (similar):
![screenshot_20170726_152007](https://user-images.githubusercontent.com/6429953/28609142-03730986-71a9-11e7-948a-b49d73a579ae.png)

(notice the different cursor and the increased size of the badge)

PR #3487, opened 2017-08-02
Builders GUI fixes

jangmarker:
Two small fixes in the Builders view:
- use 'pointer' cursor style for tags #3473 
- update tag filter when using the browser's back button #3474 

## Contributor Checklist:

* I have updated the unit tests =>  minor patch
* I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory) =>  minor patch
* I have updated the appropriate documentation =>  minor patch

### DIFF
--- a/www/base/src/app/builders/builders.controller.coffee
+++ b/www/base/src/app/builders/builders.controller.coffee
@@ -1,6 +1,6 @@
 class Builders extends Controller
     constructor: ($scope, $log, dataService, resultsService, bbSettingsService, $stateParams,
-        $location, dataGrouperService) ->
+        $location, dataGrouperService, $rootScope) ->
         # make resultsService utilities available in the template
         _.mixin($scope, resultsService)
         $scope.connected2class = (worker) ->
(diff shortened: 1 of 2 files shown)


www/base/src/styles/styles.less
```diff
@@ -74,6 +74,9 @@
   }
 }
 
+.label {
```
jangmarker: You've merged nonetheless, do you want me to create a follow-up commit?

www/base/src/app/builders/builders.controller.coffee
```diff
@@ -31,16 +31,21 @@ class Builders extends Controller
...
+                $scope.tags_filter = [$scope.tags_filter]
+
+        updateTagsFilterFromLocation()
 
         $scope.$watch  "tags_filter", (tags, old) ->
             if old?
                 $location.search("tags", tags)
         , true
 
+        $rootScope.$on '$locationChangeSuccess', updateTagsFilterFromLocation
```
jangmarker: Do you mind if I report a bug to keep track of it? I don't know whether I have time to properly factorize this in the near future.

www/base/src/styles/styles.less
tardyp: I merged to be able to ship it in 0.9.10. I will see how it works from the release packages, and see if I ship it like this.
I would like indeed a followup to cleanup the code

tardyp: I just did a quick test on nine.bb.net with the release tarball, and it looks okay. I did test that the UI is even better if you change the ``span.label`` by a ``a.label(href="#")`` because there is a mouse over effect on top of the cursor change

Issue #3475, opened 2017-07-26
Waterfall view: status indicator bar of builders is hard to read


jangmarker:
The colorful bars in the screenshot below are tiring to read, especially if red and orange bars are right next to each other.
![spectacle q10081](https://user-images.githubusercontent.com/6429953/28609419-36101a04-71aa-11e7-8c4b-32dd823686eb.png)

Suggested solution:
They could be larger or the builder names could have a background of the indicator's color.

tardyp:
looks like an easy fix. would you like to take a look at it?
https://medium.com/buildbot/buildbot-ui-plugin-for-python-developer-ef9dcfdedac0

PR #3488, opened 2017-08-02
Waterfall GUI fixes

jangmarker:
Two small fixes in the Waterfall view:
- Increase size of build status indicators #3475 
  spectacle s12143

- sort x axis by locale (`b C m Z` instead of `C Z b m`)

## Contributor Checklist:

* I have updated the unit tests => minor patch
* I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory) => minor patch
* I have updated the appropriate documentation => minor patch

### DIFF
--- a/www/waterfall_view/src/module/main.module.coffee
+++ b/www/waterfall_view/src/module/main.module.coffee
@@ -305,8 +305,8 @@ class Waterfall extends Controller
         # Rotate text
         xAxisSelect.selectAll('text')
             .style('text-anchor', 'start')
-            .attr('transform', 'translate(0, -5) rotate(-25)')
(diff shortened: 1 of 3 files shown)


mrbauff:
The change to sort the x-axis by locale only sorts the _buildername_ but neither the href _links_ nor the _builds_ itself, thus leading to inconsistencies as observed in #3819 .

If the builders are to be sorted in a 'locale aware' way, then I think this could be done in a consistent way in the [resultspec.py](https://github.com/buildbot/buildbot/blob/80e4fd0b27b85ce56f00b4227b442c0b39b73ce7/master/buildbot/data/resultspec.py#L122).

PR #3491, opened 2017-08-03
[docs] db - minor unification

Frodox:

### DIFF
--- a/master/docs/manual/cfg-dbconfig.rst
+++ b/master/docs/manual/cfg-dbconfig.rst
@@ -4,20 +4,21 @@
 DbConfig
 --------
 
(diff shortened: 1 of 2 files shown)


seankelly:
Could you fix the DbConfig page too? It has:
> c['db_url'] = 'mysql://user@pass:mysqlserver/buildbot'

The `@` and `:` need to be swapped.

Frodox:
ah, of course. Fixed

Frodox:
Can we write about sqlite in memory case also here? `sqlite://`

@rjarry is it ready enough to use/test it?:)

tardyp:
> Can we write about sqlite in memory case also here? sqlite://

Need to be very very clear that this is only for development, or if you like to loose your data!

PR #3497, opened 2017-08-04
P4encodingfix pull request for fixes #3493

### LINKED ISSUE #3493: P4 with python 3.6 under windows fails on worker
Trying to migrate from python 2.7 to python 3.6 on worker and getting following error in Perforce sync command:

'b'p4'' is not recognized as an internal or external command,
operable program or batch file.

Which is interesting because the command is

b'p4' b'-p' b'192.168.1.10:4000' b'-u' b'ABC' b'-c' b'ABC-buildbot09-build-targets' b'client' b'-i'
 in dir D:\buildbot\build-targets\build (timeout 1200 secs)
 watching logfiles {}
 argv: [b"b'p4'", b"b'-p'", b"b'192.168.1.10:4000'", b"b'-u'", b"b'ABC'", b"b'-c'", b"b'ABC-buildbot09-build-targets'", b"b'client'", b"b'-i'"]
 writing 547 bytes to stdin
 using PTY: False

Which has doubly-quoted binary items here.

Not sure if the problem is the binary string or something else. Master is running 3.4.3

sphet:
I've added a pull request to remove the encoding for P4.

Hope I did this all right. It fixes issue #3493

### DIFF
--- a/master/buildbot/newsfragments/p4encodingfix-3493.bugfix
+++ b/master/buildbot/newsfragments/p4encodingfix-3493.bugfix
@@ -1,0 +1,2 @@
+Removed encoding in p4 as it was causing double encoded bytestrings on Python 3.4
+
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -234,6 +234,20 @@ class P4(Source):
 
         command.extend(doCommand)
 
+        def encodeString(s):
+            if not isinstance(s, str):
+                return s.encode('utf-8')
+            return s
+
+        def encodeArg(arg):
+            if isinstance(arg, tuple):
+                # If a tuple, then the second element is the argument that will
+                # be used when executing the command.
+                return (arg[0], encodeString(arg[1]), arg[2])
+            return encodeString(arg)
+
+        command = [encodeArg(c) for c in command]
+
         return command
 
     def _dovccmd(self, command, collectStdout=False, initialStdin=None):
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -25,7 +25,6 @@ from twisted.python import log
 
 from buildbot import config
 from buildbot import interfaces
-from buildbot import util
 from buildbot.interfaces import WorkerTooOldError
 from buildbot.process import buildstep
 from buildbot.process.properties import Interpolate
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -235,14 +235,6 @@ class P4(Source):
 
         command.extend(doCommand)
 
-        def encodeArg(arg):
-            if isinstance(arg, tuple):
-                # If a tuple, then the second element is the argument that will
-                # be used when executing the command.
-                return (arg[0], util.encodeString(arg[1]), arg[2])
-            return util.encodeString(arg)
-
-        command = [encodeArg(c) for c in command]
         return command
 
     def _dovccmd(self, command, collectStdout=False, initialStdin=None):


rodrigc:
Unfortunately for this kind of change, you need to test it on Python 2.7 and 3.4.

I don't think this fix is correct.  I took a look into this, and I think that the "correct" fix which works properly on Python 2.7 and 3.4 is more involved, and touches more modules than just the p4 module.

sphet:
What's the right move then? I don't want to have to roll back to 2.7. Is
there some insight on the right patch? I can try to setup a virtual envin
2.7 but there is no guarantee my test will be thorough as we're in an ascii
domain, not unicode.

rodrigc:
Unfortunately, I don't think there is a quick and easy patch to fix this.
I think that the unicode/bytes stuff needs to be reworked carefully in p4
and several of the other modules related to SCM.  It takes some careful thought and testing
to: (1)  fix things on Python 3, (2) not break Python 2.

Testing with actual p4 binaries, and not just the mocked out unit tests is also important.

tardyp:
I dont think we really need to test with P4. 
We probably need to be stricter in the remotecommand unit test framework to verify that the commands strings are passed with proper typing.

Then I don't think this patch really explains the bug you are trying to fix.

My take on this bug is that there is something in the code that does 
command = [str(x) for x in command]

str(b'xx') return b'xxx' on py2 and u"b'xxx'" on py3. then there is probably another encoding :-/

I think we should rather hunt this str()

sphet:
I'll try to put together a test script tomorrow and test in 2.7 and 3.x.

One question: is the expectation that we should accept unicode clientspec
and source control names here? What is the history of this kind of escape
-- I mean what was its purpose?

tardyp:
> One question: is the expectation that we should accept unicode clientspec
and source control names here? What is the history of this kind of escape
-- I mean what was its purpose?

I am not sure what is exactly the question. 
There is the problem of legacy. What worked in release 0.9.n should work the same in 0.9.n+1. So if at somepoint a buildbot component accepted the user to send a parameter as unicode, then it should always work in future release. So we cannot say: this parameter has to be bytes starting from now. 

Then at some point in the pipeline, there are some components (twisted or p4 process) that requires properly encoded bytes, and python2 was had a different autocast policies. So we started to add explicit casting code, so that the code is working with python2 and python3. 
Lots of that casting has been done with pre-requisite that the unit tests where properly written, and we couldn't test everything e2e especially when its about proprietary SCM.

bdbaddog:
As far as I can tell unicode is the default for p4 can work with unicode.
Here's the chapter from the manual.

https://www.perforce.com/perforce/r15.1/manuals/
p4guide/chapter.configuration.html#configuration.unicode

"The Perforce service can be run in Unicode mode to activate support for
file names or directory names that contain Unicode characters, and Perforce
identifiers (for example, user names) and specifications (for example,
changelist descriptions or jobs) that contain Unicode characters."

https://www.perforce.com/perforce/r15.1/manuals/cmdref/P4CHARSET.html

sphet:
@bdbaddog Yes - this is true. And in python 3 strings are by default unicode. So I suspect the issue is that we don't want to encode it at all in python3, but we need to encode it in python 2.7. 

There appears to be code in the remote shell process that does a unicode-to-native operation already:

https://github.com/buildbot/buildbot/blob/119a673276ac843c3ece2d181267df3253ad1c83/worker/buildbot_worker/runprocess.py#L319

So perhaps if the string is already encoded we shouldn't encode it again? I will see if there is a reasonable way to do this but the 'text_type' used in a number of places is unicode in py3.6 and simply a str (ascii) in py2.7.

Further notes - the encodeArg found in GitPoller encodes the string as ascii, while the p4 encodes it as the default encoding of utf-8. This measn that the git poller can't handle non-ascii characters and certainly my test shows that.

There is a method in util being used by p4 called encodeString but it is only used for p4 so I don't have much faith in it being correct. https://github.com/buildbot/buildbot/blob/2150ad453bfc76b8e3853d23ad0f7d0f7b022c25/master/buildbot/util/__init__.py#L213

In py2.7 it essentially says if this is a Str then encode it as utf-8.  In 3.6 it says if this is a unicode then encode it as unicode. This essenitally wraps it in the b' ' because it is already encoded. So my guess is we really only want to encode it if the current string encoding differs from the target.

Are there any patterns already here for that?

tardyp:
> This essenitally wraps it in the b' ' because it is already encoded. 
No. .encode() will never wrap something in b''. str() will.

This code is suspicious:
https://github.com/buildbot/buildbot/blob/2150ad453bfc76b8e3853d23ad0f7d0f7b022c25/master/buildbot/util/__init__.py#L219
is it used somehow in your usecase?

rodrigc:
For a long time, buildbot only worked with python2, so the unicode handling was done with only python2 in mind.  While things *just worked*, I don't think things were done consistently for the python3 model of unicode.  I cleaned up a lot of stuff as part of python3 porting work,
but I think there are still some problems in the pollers.

I am going to disagree with Pierre, and say that while it is good to get the unit tests to work,
if you can do some additional testing on your own with a real p4 setup and get it to work with python2 and python3, that would be good.  If you can provide quick instructions on how to set up a p4 repository for testing that would be helpful.

I may get some disagreement on this, but my other suggestion is that in the poller,
as much as possible you should aim to use unicode strings internally in the application
on python2 *and* python3, and when you push things out over the wire, you convert from unicode to bytes, and when you pull things in over the wire, you convert from bytes to unicode.
This pattern was named the "unicode sandwich" by Ned Batchelder: https://nedbatchelder.com/blog/201203/pragmatic_unicode.html

seankelly:
`none_or_str` is only used in tests of itself. That could cause the problem, but it's not used anywhere in the master or worker code.

sphet:
@tardyp I see - Str() is putting it in b'' because it is now a byte string. However, I still think that in python 3.x we don't want to do that -- there is clearly code in the remote shell that handles the unicode strings and converts them appropriately.  It isn't clear to me where the Str() is being called - likely something inside the RemoteShellCommand is causing this to happen.

I'm out of my depth now -- all I really want to do is fix this one place by not encoding the strings if python3 . Some of the commands work without encoding already.

I am thinking this would work locally - as long as the only things passed in as arguments are strings or unicode strings - which is all that seems possible in p4.py

```
def encodeString(s, encoding='utf-8'):
	if not isinstance(s, str):
		return s.encode(encoding)
	return s
```

* In python 2.7, ascii strings are type str, but unicode are not, so it would be encoded.
* In python 3 unicode strings are type str so they would not be encoded.

This seems to work in my test case. 

S

sphet:
@rodrigc The existing function in util was only used by perforce, so I thought bringing it local to perforce made more sense. I will admit I failed to delete the now-unused item in perforce. Can you point me to 'one of the existing functions' to use?

sphet:
Can we find some way forward on this. At this point I am about ready to roll my master and workers all back to 2.7 because I don't want to maintain a custom repository. I'm pretty new to git and it's becoming a bit difficult handling a number of branches and a live install of all those changes (unless someone has a trick to do that)

tardyp:
@sphet I think @rodrigc 's latest patch are proper fixes for your bug. Please close this I am correct

rodrigc:
I think that my fix in https://github.com/buildbot/buildbot/pull/3543 fixes this as @tardyp mentioned.
However, I think that @sphet uncovered one legitimate issue.  I think that `encodeArg`
can just be deleted from `p4.py`.  I don't think it does anything useful, and it makes the code more confusing to follow.  Encoding of arguments to bytes happens in other places in the stack,
so I don't think it is necessary in `p4.py`.
However, I am not going to touch `p4.py` just now.....we can clean this up in a follow-on
PR.

seankelly:
I concur.

tardyp:
ok. waiting for @sphet to close this one then. We can remove that helper in a later PR

PR #3423, opened 2017-07-07
Added support for Roles for Email Domain

pmatos:
## Contributor Checklist:

* [ X ] I have updated the unit tests
* [ X ] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [ X ] I have updated the appropriate documentation

Fixes #3422

PR #3500, opened 2017-08-09
RolesFromDomain: fix typo in documentation and add entry_point information

pmatos:
Fix typo in documentation for RolesFromDomain made in #3423. Add entry_points so we can use class.


pmatos:
I don't really quite get what the problem was here with the tests. It seems to be a selenium test failing but I can't see how much changes affect this.

PR #3501, opened 2017-08-09
Encode recipients to bytes

rodrigc:


rodrigc:
We need a unit test with an e-mail recipient with non-ascii characters in it, such as
*ä*

tardyp:
@rodrigc so you want to add that unit test? Looks from emails that there is another patch needed in twisted to support unicode emails..

rodrigc:
On second thought, this patch is not fully correct.  Although we are encoding things using utf-8 and passing it to Twisted,
it looks like internally the Twisted ESMTPSenderFactory is expecting things to be encoded to ascii.  We cannot encode using utf-8, and pass to ESMTPSenderFactory which internally may try to decode to ascii.

The "ascii" encoding is 7-bit ascii.  Technically this is correct, but in this day and age,
this is very restrictive, because we have users who send e-mails with names set.

The proper fix for this needs to be done inside Twisted, to by default use a more accepting
encoding than "ascii" such as "latin1" or even "utf-8".

There are two things to consider:
(1)  Fields in the mail header, such as *From:*, *To:*, *Subject:* need to be encoded inside Twisted.
(2)  The message body needs to be encoded in Twisted.

I'll see what patch I can push to Twisted, along with some test cases with some non-ascii characters.

tardyp:
as per:
https://en.wikipedia.org/wiki/Email_address

> In mail systems compliant with RFC 6531 and RFC 6532 an email address may be encoded as UTF-8, both a local-part as well as a domain name.

so utf8, no latin1

rodrigc:
ok

PR #3511, opened 2017-08-13
support cancel build during prepare worker

jonnyyu:
In latent worker scenario, depending on the actual latent implementation, worker prepare may take a long time. 
In that case, it's desirable to cancel a build during prepare worker. 
In latent worker implementation, I can add logic to check build.stopped status, and cancel preparation. But when the prepare returns false, the code here will always start a new build to retry which defeats the stop build purpose.
The change here is to really cancel the build if the build has been marked stopped.

## Remove this paragraph

And especially:

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -326,7 +326,7 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
         if ready_or_failure is not True:
             yield self.buildPreparationFailure(ready_or_failure, "worker_prepare")
             if self.stopped:
-                self.buildFinished(["worker", "cancelled"], CANCELLED)
+                self.buildFinished(["worker", "cancelled"], self.results)
             else:
                 self.buildFinished(["worker", "not", "available"], RETRY)
             return
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -325,7 +325,10 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
         # If it returns failure then we don't start a new build.
         if ready_or_failure is not True:
             yield self.buildPreparationFailure(ready_or_failure, "worker_prepare")
-            self.buildFinished(["worker", "not", "available"], RETRY)
+            if self.stopped:
+                self.buildFinished(["worker", "cancelled"], CANCELLED)
+            else:
+                self.buildFinished(["worker", "not", "available"], RETRY)
             return
 
         # ping the worker to make sure they're still there. If they've
--- a/master/buildbot/test/integration/test_latent.py
+++ b/master/buildbot/test/integration/test_latent.py
@@ -30,7 +30,7 @@ from buildbot.interfaces import LatentWorkerFailedToSubstantiate
 from buildbot.interfaces import LatentWorkerSubstantiatiationCancelled
 from buildbot.process.buildstep import BuildStep
 from buildbot.process.factory import BuildFactory
-from buildbot.process.results import RETRY
+from buildbot.process.results import RETRY, CANCELLED
 from buildbot.process.results import SUCCESS
 from buildbot.test.fake.latent import LatentController
 from buildbot.test.fake.reactor import NonThreadPool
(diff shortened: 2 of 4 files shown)


master/buildbot/process/build.py
```diff
@@ -325,7 +325,10 @@ def startBuild(self, build_status, workerforbuilder):
         # If it returns failure then we don't start a new build.
         if ready_or_failure is not True:
             yield self.buildPreparationFailure(ready_or_failure, "worker_prepare")
-            self.buildFinished(["worker", "not", "available"], RETRY)
```
tardyp: I dont think this is suffisiant. I would expect to have some code added to timeout the start of latent worker or something like this. Is that already there?

jonnyyu:
just added some unit tests, will look into test_latent.py for e2e tests.

master/buildbot/process/build.py
jonnyyu: AbstractLatentWorker already has worker missing timer to cover the start timeout.

AbstractLatentWorker.substantiate() will start a missing_timer before calling create_instance(), when it is fired, it will call AbstractLatentWorker._missing_timer_fired which will notify _substantiation_notifier as Failure, so here the workerForBuilder.prepare() will have a Failure exception.

jonnyyu:
@tardyp I've updated tests and added a newsfragment.
please review again when you have time, thanks!

master/buildbot/process/build.py
tardyp: well. I think the cancel should be instant, it does not have to wait until timeout, which can be quite long.

I think stopBuild() should force stop of instantiation somehow, probably by forcing the timer to timeout.

master/buildbot/test/integration/test_latent.py
```diff
@@ -538,3 +539,45 @@ def test_worker_close_connection_while_building(self):
...
+            'multiMaster': True,
+        }
+        master = self.getMaster(config_dict)
+        builder_id = self.successResultOf(
+            master.data.updates.findBuilderId('testy'))
+
+        # Trigger a buildrequest
+        self.createBuildrequest(master, [builder_id])
+
+        unclaimed_build_requests = []
```
tardyp: this is not used in this test

master/buildbot/test/integration/test_latent.py
```diff
@@ -538,3 +539,45 @@ def test_worker_close_connection_while_building(self):
...
+        self.successResultOf(master.mq.startConsuming(
+            lambda key, request: unclaimed_build_requests.append(request),
+            ('buildrequests', None, 'unclaimed')))
+
+        # Stop the build
+        master.mq.produce(("control", "builds", '1', 'stop'),
+                          dict(reason='no reason'))
+
+        # Indicate that the worker can't start an instance.
+        controller.start_instance(False)
```
tardyp: this should not be needed. stop build should do it instead

tardyp:
@jonnyyu will you be able to fix the last small remaining problems? 

We can merge this without the stop build synchronous latent worker suspend, as it is still an improvement over current situation.

I am sure that after some time you (or someone using the stop button with latent workers) will want to implement the synchronous stop :)

jonnyyu:
@tardyp sorry for late reply, I will make the fix for unused variable and I’ll need make another fix that I should return self.results instead of hard coded cancelled when stopped==true. I found I need mark build as retry when stop build is invoked by cleanShutdown.  I don’t have time to do it now, maybe this weekend.

jonnyyu:
@tardyp updated

MikeAlexeev:
Hi. 
@jonnyyu, how do you check if build is stopped in latent worker?

I'm working on such problem too. 
I have tried to add defer to Build object which is fired when build is stopped (in Build.stopBuild method)
And I add callback to it in LatentWorker.start_instance as we have build object as start_instance argument.
In callback I cancel substantiating

Something like this:
```
def start_instance(self, build):
        d = some_substantiate_operation()
        build.stopped_defer.addCallback(lambda x: d.cancel())
        return d
```
Or this cancellation can be done in AbstractLatentWorker._substantiate to make 
this behavior default for all latent workers.

@tardyp, maybe it would be convinient if there are some deffers in Build instance
which are fired when build state is changed?
Or maybe change type of Build.stopped from bool to defer?

tardyp:
@MikeAlexeev I think it should be easier to just have stopBuild function in the Build class detect if its workerforbuildber is in prepare state and then will decide to cancel substanciation i.e. unsubstanciate

MikeAlexeev:
Ok, sounds good. I will try

PR #3512, opened 2017-08-13
Fix secrets file permissions when downloaded to worker

perher:
Secrets downloaded to the worker are using default permissions, meaning that they are readable by other users. This doesn't work well with sending for example ssh keys, (~/.ssh/id_rsa).
The patch forces all secrets to only allow read and write (not execute) by current owner.
Let me know if a more comprehensive solution like allow setting file permissions in the config is needed.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/secrets_perms.bugfix
+++ b/master/buildbot/newsfragments/secrets_perms.bugfix
@@ -1,0 +1,1 @@
+Fix secrets downloaded to worker with too wide permissions
--- a/master/buildbot/steps/download_secret_to_worker.py
+++ b/master/buildbot/steps/download_secret_to_worker.py
@@ -15,6 +15,8 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import stat
+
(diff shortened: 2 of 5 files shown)


tardyp:
Looks good. I can't think of use case where a secret would be other than 600. If somebody has, then we can add the configurability.

Also, not sure how this will work on windows

seankelly:
I can see someone wanting group read permission. But, I think that should be pushed off for a configuration parameter.

tardyp:
@seankelly not sure if you want to do the conf parameter now, or in followup PR.

seankelly:
I'm fine with a follow-up PR.

perher:
Cool, thanks for merging. I can have look at creating a second PR for specifying file permissions.

Not sure what the syntax would look like, maybe something like adding an option tuple parameter
`secrets_list = [('/first/path', Interpolate('write something and %(secret:somethingmore)s'), '755'),
                ('/second/path', Interpolate('%(secret:othersecret)s'), '700']`

Issue #3517, opened 2017-08-14
LineBoundaryFinder struggles when a script does not output new lines


tardyp:
```
<aj355> I finally figured out what was causing the performance issue (https://github.com/buildbot/buildbot/issues/3444)
06:22:50 A<aj355> We have a (innocent looking) build step which copies ~600,000 files (in 20 minutes) from one location to another on the worker machine. This step is run once in a while (that’s why i was seeing performance issue intermittently). The performance issue happens exactly when this step is run.
06:22:50 A<aj355> This step prints a lot of data very quickly on console in a progress bar kind of manner (it keeps over-writing the console output by writing \b (backspace)
```

A script could indeed be showing status report using terminal control flow (or ugly \b) instead of \r.

This could be a DOS issue if a evil script output lots of data without newlines, then master will crash with out of memory.

PR #3521, opened 2017-08-14
Performance improvement for lineboundaries

tardyp:
When a remotecommand sends huge lines, the LBF will eat the data until it finds a boundary.
This could cause a master to OOM if a malicious or buggy script send lots of data to stdout/err

We also add support for other line boundaries (ANSI cursor control and \b+)

fix for #3517 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_util_lineboundaries.py
+++ b/master/buildbot/test/unit/test_util_lineboundaries.py
@@ -96,6 +96,13 @@ class LBF(unittest.TestCase):
         self.assertCallbacks(['1%\n5%\n15%\n100%\n', 'finished\n'])
 
     @defer.inlineCallbacks
+    def test_backspace_folded(self):
+        r"a lot of \b is treated as and converted to a newline"
+        yield self.lbf.append('1%\b\b5%\b\b15%\b\b\b100%\nfinished')
+        yield self.lbf.flush()
+        self.assertCallbacks(['1%\n5%\n15%\n100%\n', 'finished\n'])
+
+    @defer.inlineCallbacks
     def test_mixed_consecutive_newlines(self):
         r"mixing newline styles back-to-back doesn't collapse them"
         yield self.lbf.append('1\r\n\n\r')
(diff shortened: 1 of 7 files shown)


master/buildbot/test/unit/test_util_lineboundaries.py
```diff
@@ -117,6 +124,34 @@ def test_split_newlines(self):
...
+        yield self.lbf.append('1234\033[1;2H4321')
+        yield self.lbf.flush()
+        self.assertCallbacks(['1234\n', '4321\n'])
+        yield self.lbf.append('1234\033[1;2f4321')
+        yield self.lbf.flush()
+        self.assertCallbacks(['1234\n', '4321\n'])
+
+    @defer.inlineCallbacks
+    def test_long_lines(self):
+        """long lines are splitted"""
```
seankelly: "Split" is the right form to use.

master/buildbot/util/lineboundaries.py
```diff
@@ -20,21 +20,45 @@
...
+class LineBoundaryFinder(object):
 
+    __slots__ = ['partialLine', 'callback', 'warned']
+    # split at reasonable line length.
+    # too big lines will fill master's memory, and slow down the UI too much.
+    MAX_LINELENGTH = 4096
     # the lookahead here (`(?=.)`) ensures that `\r` doesn't match at the end
     # of the buffer
-    newline_re = re.compile(r'(\r\n|\r(?=.)|\n)')
+    # we also convert cursor control sequence to newlines
```
seankelly: Does this mean that the control sequences will only be saved for "short" lines? If there's a process that output text with the control sequences that are then colored in the web UI, there would be nothing to color for the long lines?

master/buildbot/test/unit/test_util_lineboundaries.py
tardyp: damn google misguided me !

master/buildbot/util/lineboundaries.py
tardyp: Its only the control sequence used for cursor control. color control sequence are different.

PR #3526, opened 2017-08-15
Fix rest api filtering

jangmarker:
Fixes queries like:
[/api/v2/builds?builderid__eq=1&builderid__eq=2](http://nine.buildbot.net/api/v2/builds?builderid__eq=1&builderid__eq=2)

This would have yielded an empty result with the previous code.

## Contributor Checklist:

       code is fixed to match the documentation


tardyp:
good catch. This needs unit tests!

does that also fix #3527 ?

jangmarker:
I've added unit tests, however, I haven't found a usage of the database `contains` operator. Maybe I've been overlooking something.

It doesn't fix #3527, in fact I discovered #3527 while working on this. For a given list property of a database object the `contains` operator requires *all* values given in the REST call to be in that list. This is a sensible query, but a query where *at least one* of the values given in the REST call have to be in the list property is not possible at the moment (at least I haven't found a way).

tardyp:
Hi @jangmarker the Rest API filtering is featured around the needs for the UI.
I think nobody actually looked at the plural case for contains.

If you have usecases, I think it makes more sense to have the contains operator have a default 'or' behaviour. It better matches the behaviour for 'eq'.
We could add a xcontains operator, which has a 'and' plural behaviour.

master/buildbot/data/resultspec.py
```diff
@@ -49,17 +49,26 @@ class FieldBase(object):
         'contains': lambda d, v: set(v) <= set(d),
     }
 
+    plural_operators_sql = {
+        'eq': lambda d, v: d.in_(v),
+        'ne': lambda d, v: d.notin_(v),
+        'contains': lambda d, vs: (v.in_(d) for v in vs),
```
tardyp: needs a plural test for contains. I am not sure how this code would work

master/buildbot/data/resultspec.py
```diff
@@ -49,17 +49,26 @@ class FieldBase(object):
         'contains': lambda d, v: set(v) <= set(d),
```
tardyp: looks like this 'old' code should provide a OR behaviour at data API level. 
It is just the sql mode which was not implemented

jangmarker: Currently the old code does not provide OR behavior. For instance:
1) https://nine.buildbot.net/api/v2/builders?tags__contains=buildbot&order=builderid
    yields multiple builders, including `builderid: 1` (first one)
2) https://nine.buildbot.net/api/v2/builders?tags__contains=try&order=builderid
    yields one builder `builderid: 1`
3) https://nine.buildbot.net/api/v2/builders?tags__contains=try&tags__contains=buildbot&order=builderid (combination of 1+2)
    yields only one builder (AND behavior), but with OR it would yield all builders shown in the queries 1 and 2

I'll create a new pull request that changes the `contains` behavior to OR and then will adapt this request accordingly.

tardyp:
@jangmarker I think you wanted to rebase this PR?

jangmarker:
Yes, sorry, I didn't make it before vacation. My response times will be rather slow until September.

PR #3534, opened 2017-08-16
schedulers: make builderNames renderable

rjarry:
Add a new `builderFilter` parameter to all schedulers. This callback allows to dynamically select which builders get scheduled when this scheduler is triggered. By default, all builders in `builderNames` are
scheduled.

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -864,9 +864,15 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
         # check that all builders are implemented on this master
         unscheduled_buildernames = set([b.name for b in self.builders])
         for s in itervalues(self.schedulers):
-            for n in s.listBuilderNames():
-                if n in unscheduled_buildernames:
-                    unscheduled_buildernames.remove(n)
+            builderNames = s.listBuilderNames()
+            if interfaces.IRenderable.providedBy(builderNames):
+                unscheduled_buildernames.clear()
+            else:
+                for n in builderNames:
+                    if interfaces.IRenderable.providedBy(n):
+                        unscheduled_buildernames.clear()
+                    elif n in unscheduled_buildernames:
+                        unscheduled_buildernames.remove(n)
         if unscheduled_buildernames:
             error("builder(s) %s have no schedulers to drive them"
                   % (', '.join(unscheduled_buildernames),))
@@ -879,7 +885,12 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
         all_buildernames = set([b.name for b in self.builders])
 
         for s in itervalues(self.schedulers):
-            for n in s.listBuilderNames():
+            builderNames = s.listBuilderNames()
+            if interfaces.IRenderable.providedBy(builderNames):
+                continue
+            for n in builderNames:
+                if interfaces.IRenderable.providedBy(n):
+                    continue
                 if n not in all_buildernames:
                     error("Unknown builder '%s' in scheduler '%s'"
                           % (n, s.name))
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -69,6 +69,8 @@ class Properties(util.ComparableMixin):
         if kwargs:
             self.update(kwargs, "TEST")
         self._master = None
+        self._sourcestamps = None
+        self._changes = None
 
     @property
     def master(self):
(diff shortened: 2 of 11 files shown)


rjarry:
I think I understand the concept but I don't see how it would be used. Could you provide a dummy example on how you would configure such a scheduler ?

tardyp:
```
@util.renderer
def findBuilders(props):
    builders = []
    for f in props.allFiles():
        if f.endswith('.rst'):
            builders.append('check_docs')
        if f.endswith('.c'):
            builders.append('check_code')
            
c['schedulers'].append(schedulers.SingleBranchScheduler(
                            name="all",
                            treeStableTimer=None,
                            builderNames=findBuilders))

```

master/buildbot/process/properties.py
```diff
@@ -80,6 +82,36 @@ def master(self):
     def master(self, value):
         self._master = value
 
+    @property
+    def sourcestamps(self):
+        if self.build is not None:
+            return self.build.getAllSourceStamps()
```
tardyp: return [ss.__ssdict for ss in self.build. getAllSourceStamps()]

master/buildbot/process/properties.py
```diff
@@ -80,6 +82,36 @@ def master(self):
...
+        return self._sourcestamps
+
+    @sourcestamps.setter
+    def sourcestamps(self, value):
+        self._sourcestamps = value
+
+    @defer.inlineCallbacks
+    def allChanges(self):
+        if self.build is not None:
+            defer.returnValue(self.build.allChanges())
```
tardyp: return [ss.__chdict for ss in self.build. getAllChanges()]

master/buildbot/process/properties.py
```diff
@@ -80,6 +82,36 @@ def master(self):
...
+    def sourcestamps(self):
+        if self.build is not None:
+            return self.build.getAllSourceStamps()
+        return self._sourcestamps
+
+    @sourcestamps.setter
+    def sourcestamps(self, value):
+        self._sourcestamps = value
+
+    @defer.inlineCallbacks
```
tardyp: I  think we should not just pre-populate the changes as we usually already have them in reporter or schedulers.
Idea is to avoid requiring people to deal with deferred as much as we can.

master/buildbot/process/properties.py
```diff
@@ -80,6 +82,36 @@ def master(self):
...
+                if isinstance(ss, integer_types):
+                    changes.extend((yield self.master.data.get(
+                        ('sourcestamps', ss, 'changes'))))
+        defer.returnValue(changes)
+
+    def allFiles(self):
+        if self.build is not None:
+            return self.build.allFiles()
+        files = []
+        for chdict in self.getAllChanges():
```
tardyp: see? getAllChange returned deferred, and you are screwed.

master/buildbot/process/properties.py
rjarry: I don't like the idea of accessing a private attribute like this. Maybe we should tune the `asDict()` method so that it takes a `v2` parameter and returns the correct value accordingly?

master/buildbot/process/properties.py
rjarry: I understand the problem, but how do you manage to access the changes if build is not set?

master/buildbot/process/properties.py
tardyp: then make a getSSDict() in TempSourceStamp

master/buildbot/process/properties.py
tardyp: the creator of the properties must is the setter. If it hasn't then we should raise an attribute error or something like this

rjarry: But in the case of a scheduler, there is no build yet. So there is no way to set it.

On the other hand, we have a list of sourcestamps (and/or a list of changes) that triggered the scheduler.

tardyp: I meant the setter of self._sourcestamps

rjarry: So this means that `Properties.allChanges()` and `Properties.allFiles()` must raise an error when `build` is not set.

On the other hand, I hope that it is possible for an `IRenderable` to return a deferred. Otherwise, I am screwed if I want to access the data API from inside it.

rjarry:
I have made a few changes according to the review. There is a failed check with protractor tests which I am not sure I understand...

rjarry:
This now looks OK.

I have one suggestion to make. Would that make sense to combine properties that come from `changes` into those that are used to render `builderNames`?

tardyp:
I agree property combination would make sense

tardyp:
actually. Only change properties really make sense to generate the buildernames.
and I also want schedulers properties to be renderable and depend on change properties.

rjarry:
Haha, that's sneaky :D Ok, I'll see what I can do.

master/buildbot/config.py
```diff
@@ -864,9 +864,15 @@ def check_single_master(self):
         # check that all builders are implemented on this master
         unscheduled_buildernames = set([b.name for b in self.builders])
         for s in itervalues(self.schedulers):
-            for n in s.listBuilderNames():
-                if n in unscheduled_buildernames:
-                    unscheduled_buildernames.remove(n)
+            builderNames = s.listBuilderNames()
+            if interfaces.IRenderable.providedBy(builderNames):
+                unscheduled_buildernames.clear()
```
tardyp: needs test for this..

master/buildbot/test/unit/test_process_properties.py
```diff
@@ -64,6 +65,24 @@ def asDict(self):
         }
         return ds
 
+    def asSSDict(self):
+        return self.asDict()
+
+
+class FakeChange:
```
tardyp: I don't think it makes much sense to use FakeChange/FakeSource. TempChange/SourceStamp is nearly the same, and can be used as is. So that it is also tested by this test.
I believe this code predates TempSourceStamp, so this is why it is weirdly implemented.

master/docs/manual/cfg-schedulers.rst
```diff
@@ -35,6 +35,83 @@ There are several common arguments for schedulers, although not all are availabl
...
+    .. code-block:: python
+
+       from buildbot.plugins import util, schedulers
+
+       @util.renderer
+       def builderNames(props):
+           builders = []
+           for f in props.files:
+               if f.endswith('.rst'):
+                   builders.append('check_docs')
```
tardyp: Actually, this code is shitty (I know I wrote it). We should use a set in order to deduplicate (and turn into a list in return statement)

rjarry:
Ready for round #4 :)

tardyp:
thanks for the great work!

Frodox:
Woooohhhooooo!!! :fireworks:

rjarry:
Yay :-)

Thanks for your thorough review and your patience @tardyp !

PR #3537, opened 2017-08-17
git: use `checkout -B` rather than `branch -M`

mathstuf:
Some projects require local branches to exist (namely `master`) in order
to run their tests. The rename step takes away the `master` branch from
the checkout on the build machine and can cause tests to fail. Rather
than renaming the current branch, checkout a new branch, overwriting any
branch of the previous name.

## Contributor Checklist:

Not sure if there's relevant documentation as to how the local branch gets made anywhere or not.


tardyp:
looks good. keeping it open to let others object if necessary

PR #3543, opened 2017-08-18
Perforce bytes fix

### LINKED ISSUE #3493: P4 with python 3.6 under windows fails on worker

rodrigc:
Fix a case where the Perforce step was passing bytes all the way down
to RemoteCommand.  `buildbot_worker.util.__init__.Obfuscated(b"something")` was returning a string `"b'something'"` instead of `b"something"` which was causing things to break.

Fixes #3493

### DIFF
--- a/master/buildbot/newsfragments/perforce.bugfix
+++ b/master/buildbot/newsfragments/perforce.bugfix
@@ -1,0 +1,1 @@
+Fix the Perforce build step on Python 3 (:issue:`3493`)
--- a/master/buildbot/process/remotecommand.py
+++ b/master/buildbot/process/remotecommand.py
@@ -16,7 +16,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 from future.utils import iteritems
-from future.utils import string_types
+from future.utils import text_type
 
 from twisted.internet import defer
 from twisted.internet import error
@@ -360,7 +360,7 @@ class RemoteShellCommand(RemoteCommand):
         if decodeRC is None:
             decodeRC = {0: SUCCESS}
         self.command = command  # stash .command, set it later
-        if isinstance(self.command, string_types):
+        if isinstance(self.command, (text_type, bytes)):
             # Single string command doesn't support obfuscation.
             self.fake_command = command
         else:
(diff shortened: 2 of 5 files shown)


rodrigc:
@sphet Please try this patch which fixes #3493.
Please apply the patch manually, or obtain the whole tree from git by doing:

`git clone -b perforce-bytes-fix https://github.com/rodrigc/buildbot`

Please report back.

sphet:
Thanks @rodrigc - I will be back in the office on Wednesday PDT and will check it then.

seankelly:
This fixes the issue for me.

rodrigc:
Could you describe your setup, configuration, and what steps you followed to verify the fix?

seankelly:
[The setup](https://github.com/seankelly/buildbot-test-source). It can replicate the problem.

I installed your branch for master and worker on Python 2 and 3. After that, it works as expected.

rodrigc:
Thanks for providing details about your test setup, and testing on Python 2 and 3!

sphet:
Does this change require upgrading all the workers as well, or only the master?

I'm back at this and upgrading only the master still gives me the same bugs.

rodrigc:
You need to upgrade both

PR #3558, opened 2017-08-25
grid_view: handle RETRY builds

rjarry:
When a build is RETRY'd (for example, because a worker has been disconnected before it was finished), there are multiple builds for the same build request.

Handle that special case, and display all builds for each build request.


rjarry:
I'm not sure what went wrong here :)

rjarry:
Gentle bump, I'm not sure what to do with the build errors :-s

PR #3568, opened 2017-09-01
Fix example which fails with "trial pyflakes"

rodrigc:
When calling "trial pyflakes", add current directory to PYTHONPATH.

This is required in Twisted 16.6.0 and higher.

See:
https://stackoverflow.com/questions/42425901/twisted-trial-pythonpath-and-sys-path

In Twisted 16.4.0 and higher, trial does not add the pwd to PYTHONPATH.

A user is required to either:
  1.  Properly install their package via setup.py and then run trial
  2.  Explicitly add '.' to PYTHONPATH.

The decision was made to keep this new behavior:
https://twistedmatrix.com/pipermail/twisted-python/2017-February/031145.html

### DIFF
--- a/smokes/master.cfg
+++ b/smokes/master.cfg
@@ -34,7 +34,7 @@ c['protocols'] = {'pb': {'port': 9989}}
 # about source code changes.  Here we point to the buildbot clone of pyflakes.
 c['change_source'] = []
 c['change_source'].append(changes.GitPoller(
-        'https://github.com/buildbot/pyflakes.git',
+        'https://github.com/buildbot/hello-world.git',
         workdir='gitpoller-workdir', branch='master',
         pollinterval=300))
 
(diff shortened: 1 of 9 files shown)


rodrigc:
I noticed this problem before and meant to submit a patch.
@ssikyou mentioned this on the #buildbot IRC channel and reminded me that this needed to be
fixed.

smokes/master.cfg
```diff
@@ -61,15 +61,14 @@ c['schedulers'].append(schedulers.ForceScheduler(
...
-
+factory.addStep(steps.ShellCommand(command=["trial", "pyflakes"],
+                                   env={"PYTHONPATH": "."}))
 
 slowfactory = util.BuildFactory()
 #check out the source
 slowfactory.addStep(steps.Git(repourl='https://github.com/buildbot/pyflakes.git', mode='incremental'))
-# run the tests (note that this will not actually run the tests are they do not work with py3)
-slowfactory.addStep(steps.ShellCommand(command=["echo", "trial", "pyflakes"]))
+slowfactory.addStep(steps.ShellCommand(command=["trial", "pyflakes"],
```
tardyp: I am not sure we really want to run tests in these smoke tests.This will just make them slower for no reason.

Actually it fails on py3 (https://nine.buildbot.net/#/builders/26/builds/1121/steps/7/logs/stdio). probably the version of pyflakes that we test is not python3 compatible.

rodrigc:
What is the point of https://github.com/buildbot/pyflakes ?
For the purposes of this test, I would be inclined to change this to just clone the real
https://github.com/PyCQA/pyflakes , but at a specific tag, to avoid breakages due to changing upstream code.

tardyp:
I guess the point was to make sure our example is not dependent on another repository. Fetching anything but master would require configuring buildbot to not use master, which would probably confuse users,

djmitche:
Yes, the idea is just to have a small Python repo handy that's easy to download and won't break unexpectedly (because we'll never change it).  It wouldn't be a bad idea to replace it with buildbot/hello-world or something like that.

rodrigc:
OK, now I understand what buildbot/pyflakes is for.
It is just an example repository.

@tardyp can you delete the existing buildbot/pyflakes repository,
and create a new buildbot/pyflakes fork with parent https://github.com/PYCQA/pyflakes ?

That way we can keep using buildbot/pyflakes as our "hello world", we will not
depend on an upstream repo, but we will have newer code that actually has passing
tests on Python 3.

tardyp:
I prefer to go with the hello-world project. I will create one, with some dummy code, and a few tests

tardyp:
I updated your branch to transition to hello-world project

rodrigc:
OK

PR #3555, opened 2017-08-23
Fix regression in oath2 github v3 handling

dragon512:
## Remove this paragraph
Fix regression in GitHub auth2 v3 

http://trac.buildbot.net/wiki/Development
And especially:
http://trac.buildbot.net/wiki/SubmittingPatches

## Contributor Checklist:

* [ ] I have updated the unit tests
* [ ] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [ ] I have updated the appropriate documentation

PR #3575, opened 2017-09-04
fix github apiv3 for saas mode

tardyp:
it looks like https://github.com/buildbot/buildbot/pull/3555
did fix the on prem version of github, but in turn broke the saas version

@dragon512 please review, and test with on prem github (I guess you have access to one)


dragon512:
Give me a few days. I will test it out

tardyp:
Lets merge it as the default github oauth2 is broken.

dragon512:
@tardyp It seems to work for me on our enterprise GitHub

Issue #3572, opened 2017-09-01
LatentDockerWorker: Fail if image is not found and cannot be pulled


btaidm:
For LatentDockerWorker, it would make sense if the build would fail if an image isn't found or can't be downloaded instead on continually retrying.

This could be used in tandem of #3071, so that is the image is not there, it tries a pull, and then if that fails, the build should fail or retry a couples times.

I have looked at the code for latent workers, and I don't really see a way to make it stop retrying.

btaidm:
I could do that. Not sure on how I would do the unit tests.

Brainstorming here:
Would the worker tell the build process that it should give up, or should that be inherent to the
build process itself?

PR #3578, opened 2017-09-05
Setup Docker Latent Worker to autopull image based on config

### LINKED ISSUE #3071: DockerLatentWorker: pull specified image if it is not yet on the system
This is merely an enhancement, but it would be nice if the specified image is pulled from dockerhub if it is not on yet on the host system. 
This should be pretty easy to accomplish using docker-py, but I don't know if this a desired feature.

btaidm:
Added automatic pulling of docker images by setting require config.
The setting is `autopull` and is set to false by default, keeping the original workflow.

Closes #3071.
Starts #3572 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/docker_autopull.feature
+++ b/master/buildbot/newsfragments/docker_autopull.feature
@@ -1,0 +1,1 @@
+Added autopull for Docker images based on config. (:issue:`3071`)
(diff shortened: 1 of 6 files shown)


tardyp:
looks good! thanks, this was a long awaited feature!

We require to have test updated, though, so please add a unit test for this feature.

https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/test_worker_docker.py#L1

btaidm:
@tardyp 
Added Unit Test and update Fake Docker to include a pull function

tardyp:
great work!

chrwen-omicron:
An option to pull the latest could also be interesting, or should autopull simply pull every time even if the image exists to get changes?

PR #3581, opened 2017-09-06
Fixed buildbot master not start error in docker

eshizhan:
I'm running buildbot in docker with volumes(`-v /path/to/buildbot:/var/lib/buildbot`). The PID not change in container everytime, and the removing pid logic not enough when **the pid equal to ppid**. It's always try starting in looping.
```
checking basedir
checking for running master
'/var/lib/buildbot/twistd.pid' exists - is this master still running?
Can't upgrade master yet. Waiting for database ready?
```

## Contributor Checklist:

### DIFF
--- a/master/buildbot/scripts/base.py
+++ b/master/buildbot/scripts/base.py
@@ -69,7 +69,11 @@ def checkPidFile(pidfile):
                 raise OSError("Can't check status of PID {} from pidfile {}: {}".format(
                     pid, pidfile, why))
         else:
-            raise BusyError("'{}' exists - is this master still running?".format(pidfile))
+            # No exception if pid equal ppid, fixed it.
+            if pid == os.getppid():
+                os.remove(pidfile)
+            else:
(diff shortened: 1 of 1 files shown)


master/buildbot/scripts/base.py
```diff
@@ -69,7 +69,11 @@ def checkPidFile(pidfile):
                 raise OSError("Can't check status of PID {} from pidfile {}: {}".format(
                     pid, pidfile, why))
         else:
-            raise BusyError("'{}' exists - is this master still running?".format(pidfile))
+            # No exception if pid equal ppid, fixed it.
+            if pid == os.getppid():
```
tardyp: I think this deserves a bit more comments.
Anyway I am not sure if this will not bite us later. Since this is a docker specific issue, I wonder if we'd rather delete the pidfile in the docker entrypoint shell script.

eshizhan: Agree, new PR at #3606 , I'm close this.

PR #3582, opened 2017-09-06
Remove unnecessary unicode/bytes encoding from p4 modules

rodrigc:
@sphet noticed in https://github.com/buildbot/buildbot/issues/3493#issuecomment-320105740 that encodeArg() in the P4 code probably doesn't belong.

I looked at it, and he was right.  It isn't necessary, because the command is encoded to bytes much later on in https://github.com/buildbot/buildbot/blob/master/worker/buildbot_worker/runprocess.py#L328

### DIFF
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -234,15 +234,6 @@ class P4(Source):
             command.extend(self.p4extra_args)
 
         command.extend(doCommand)
-
-        def encodeArg(arg):
-            if isinstance(arg, tuple):
-                # If a tuple, then the second element is the argument that will
-                # be used when executing the command.
-                return (arg[0], util.encodeString(arg[1]), arg[2])
-            return util.encodeString(arg)
-
-        command = [encodeArg(c) for c in command]
         return command
 
     def _dovccmd(self, command, collectStdout=False, initialStdin=None):
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -25,7 +25,6 @@ from twisted.python import log
 
 from buildbot import config
 from buildbot import interfaces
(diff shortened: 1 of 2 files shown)


rodrigc:
I tested this with Python 2.7 and Python 3.6 on OS X.
I used a real perforce repository.

I created a perforce repository by doing:
```
export P4HOST=localhost
export P4PORT=1666
mkdir /tmp/sample-p4-repo
p4d -r /tmp/sample-p4-repo
```

On client, I was able to do:

```
export P4HOST=localhost
export P4PORT=1666
p4 info
```

In master.cfg, I was able to add:

```
p4source = changes.P4Source(p4base='//depot/project/', 
                                                     p4port="localhost:1666",
                                                     pollInterval=15, pollAtLaunch=True,
                                                     split_file=lambda branchfile: branchfile.split('/',1))
c['change_source'].append(p4source)

factory.addStep(steps.P4(p4port='localhost:1666',
                         p4client=util.WithProperties('localhost'),
                         p4user='crodrigues',
                         p4base='//depot',
                         mode='incremental'))
```

On both Python 2.7 and 3.6, I was able to see the real p4 commands being executed, with the command output in the log.

rodrigc:
@seankelly please test this on your [perforce setup](https://github.com/seankelly/buildbot-test-source). and report back

rodrigc:
This is a follow on commit to https://github.com/buildbot/buildbot/pull/3543

rodrigc:
@tardyp after this change, there is now no user of  `buildbot.util.encodeString()`
in the buildbot master tree.  How do we typically deal with situations like this?  Do we delete the function, or is there some sort of deprecation cycle?

tardyp:
I would be inclined to remove it. Buildbot is not really a string encoding framework. this is not really part of our API.

PR #3583, opened 2017-09-06
Check for duplicates in addURL

### LINKED ISSUE #3554: Buildbot should check for duplicates in addURL
In a step if we call self.addURL() multiple times with same URL, buildbot doesn't check if URL is already added, and adds the duplicate URL. Buildbot should check if the URL is already added before adding the URL.

This happens when a script parses a lot of output and generate links for various things. We can manually maintain a list ourself and check for duplicates before calling addURL(), but it feels so basic that buildbot should be doing this check instead of callers of addURL() checking for duplicates.

ewongbb:
ensure that duplicate entries of the url are not added.

fixes #3554

### DIFF
--- a/master/buildbot/db/steps.py
+++ b/master/buildbot/db/steps.py
@@ -152,11 +152,9 @@ class StepsConnectorComponent(base.DBConnectorComponent):
                 _racehook()
             urls = json.loads(row.urls_json)
 
-            url_exists = False
-            for url_item in urls:
-                url_exists = url_exists or (url_item.get('url') == url)
+            url_item = dict(name=name, url=url)
 
-            if not url_exists:
+            if url_item not in urls:
                 urls.append(dict(name=name, url=url))
                 q = tbl.update(whereclause=wc)
                 conn.execute(q, urls_json=json.dumps(urls))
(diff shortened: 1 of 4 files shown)


master/buildbot/db/steps.py
```diff
@@ -151,10 +151,15 @@ def thd(conn):
             if _racehook is not None:
                 _racehook()
             urls = json.loads(row.urls_json)
-            urls.append(dict(name=name, url=url))
 
-            q = tbl.update(whereclause=wc)
-            conn.execute(q, urls_json=json.dumps(urls))
+            url_exists = False
```
tardyp: this code is not very pythonic.

I would compare the whole object anyway, and this helps to simplify the code 

```
url = dict(name=name, url=url)
if url not in urls:
    [...]
```

ewongbb:
I changed util/__init__.py and source/p4.py?

ewongbb:
at this point, I"m not sure why it's busting in the FakeDB part but ok with the RealDB tests.

ewongbb:
nvm..  found out that fakeDB.py also had an addURL that needed changing.

tardyp:
thanks!

PR #3587, opened 2017-09-06
Latent Worker Stops when it cannot Substantiate

### LINKED ISSUE #3572: LatentDockerWorker: Fail if image is not found and cannot be pulled

btaidm:
This PR adds in a feature that will keep Latent Workers from continually retrying when buildrequest errors, such as a docker image cannot be found. 

Fixes #3572 

Examples include:
* Docker Image not existing

## Contributor Checklist:

### DIFF
--- a/master/buildbot/interfaces.py
+++ b/master/buildbot/interfaces.py
@@ -59,6 +59,9 @@ deprecatedWorkerModuleAttribute(
 class LatentWorkerFailedToSubstantiate(Exception):
     pass
 
+class LatentWorkerCannotSubstantiate(Exception):
+    pass
+
 
 deprecatedWorkerModuleAttribute(
     locals(), LatentWorkerFailedToSubstantiate,
--- a/master/buildbot/interfaces.py
+++ b/master/buildbot/interfaces.py
@@ -59,6 +59,7 @@ deprecatedWorkerModuleAttribute(
 class LatentWorkerFailedToSubstantiate(Exception):
     pass
 
+
 class LatentWorkerCannotSubstantiate(Exception):
     pass
 
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -327,6 +327,8 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
             yield self.buildPreparationFailure(ready_or_failure, "worker_prepare")
             if self.stopped:
                 self.buildFinished(["worker", "cancelled"], self.results)
+            elif ready_or_failure.check([interfaces.LatentWorkerCannotToSubstantiate]):
+                self.buildFinished(["worker", "cannot", "substanciate"], EXCEPTION)
             else:
                 self.buildFinished(["worker", "not", "available"], RETRY)
             return
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -327,7 +327,7 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
             yield self.buildPreparationFailure(ready_or_failure, "worker_prepare")
             if self.stopped:
                 self.buildFinished(["worker", "cancelled"], self.results)
-            elif ready_or_failure.check([interfaces.LatentWorkerFailedToSubstantiate]):
+            elif isinstance(ready_or_failure, Failure) and ready_or_failure.check(interfaces.LatentWorkerCannotSubstantiate):
                 self.buildFinished(["worker", "cannot", "substantiate"], EXCEPTION)
             else:
                 self.buildFinished(["worker", "not", "available"], RETRY)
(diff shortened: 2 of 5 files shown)


master/buildbot/process/build.py
```diff
@@ -327,9 +328,19 @@ def startBuild(self, build_status, workerforbuilder):
             yield self.buildPreparationFailure(ready_or_failure, "worker_prepare")
             if self.stopped:
                 self.buildFinished(["worker", "cancelled"], self.results)
+            elif isinstance(ready_or_failure, Failure) and ready_or_failure.check(interfaces.LatentWorkerFailedToSubstantiate):
```
tardyp: I think we need a CannotSubstanciate in order to distinguish transiant infra issues from issues that come from buildrequest

master/buildbot/worker/latent.py
```diff
@@ -97,6 +100,15 @@ def stop_instance(self, fast=False):
     def substantiated(self):
         return self.conn is not None
 
+    def retry_check(self):
+        # responsible for checking if the build should continue to retry
+        # or stop trying to substantiate
+        self.retry_count += 1
+        return self.retry_count < self.max_retry_count
```
tardyp: we need to make sure that this retry counter is from the same build! Here you are a bit re-implementing the quarantine timer

tardyp:
We already have the quarantine timer, which does the retry mechanism. I think it is good, and we should keep it.

What we don't have is a mean to distinguish a failure coming from the infra and a failure coming from a badly configured buildrequest.
My first idea was to dinstinguish failures, and have two different exceptions

LatentWorkerFailedToSubstantiate -> means transient error, we should retry after timer
LatentWorkerCannotSubstantiate -> means error from buildrequest (e.g cannot pull a non existing image), we should not retry.

We could also  automatically transform a LatentWorkerFailedToSubstantiate to a LatentWorkerCannotSubstantiate if the same buildrequest has been retried n times. For that, we need a new db API, which will tell the number of retry builds associated to the buildrequest associated to the current build. When a build is retried, the buildrequest is still the same, so you eventually have one buildrequest and several retried builds.

Note that we have a quite good integration test in order to test those things:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/integration/test_latent.py#L10

unit tests do not really make sense here as things are pretty well connected together.

let me know if you need more details

tardyp:
Please rebase and cleanup your history..
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

btaidm:
@tardyp 
History has been rewritten.
Integration Tests have been added.

master/buildbot/interfaces.py
```diff
@@ -60,11 +60,20 @@ class LatentWorkerFailedToSubstantiate(Exception):
...
+class LatentWorkerCannotSubstantiate(Exception):
+    pass
+
+
 deprecatedWorkerModuleAttribute(
     locals(), LatentWorkerFailedToSubstantiate,
     compat_name="LatentBuildSlaveFailedToSubstantiate")
 
 
+deprecatedWorkerModuleAttribute(
```
tardyp: no need for that. we never released a LatentBuildSlaveCannotSubstantiate symbol

master/buildbot/test/integration/test_latent.py
```diff
@@ -209,6 +211,44 @@ def test_failed_substantiations_get_requeued(self):
...
+            Failure(LatentWorkerCannotSubstantiate("substantiation failed")))
+        # Flush the errors logged by the failure.
+        self.flushLoggedErrors(LatentWorkerCannotSubstantiate)
+
+        dbdict = yield master.db.builds.getBuildByNumber(builder_id, 1)
+
+        # When the substantiation fails, the result is an exception.
+        self.assertEqual(EXCEPTION, dbdict['results'])
+        controller.auto_stop(True)
+        self.flushLoggedErrors(LatentWorkerFailedToSubstantiate)
```
tardyp: why also a LatentWorkerFailedToSubstantiate. sounds like an issue.

master/buildbot/interfaces.py
btaidm: Will Remove

master/buildbot/test/integration/test_latent.py
btaidm: CopyPasta mistake. Will Remove

btaidm:
@tardyp  Updated with required change

btaidm:
Prevent the staleness

PR #3590, opened 2017-09-07
Builders UI: Load builds based on tags filter

jangmarker:
On large buildbot instances (>400 builders) it depends on luck for which builders a build is shown (because only 200 builds are fetched). When drilling down into the list by filtering for a tag (project in our case) it might well be that no builds are shown for the visible builders.

Increasing the amount of loaded builds is not an option, as the resulting JSON would be too large and performance is already an issue.

This solution reloads builds when tag filtering changes. Additionally, the request for builds asks the REST API to filter the builds by builderid (which is possible with the fixes of #3526).

Drawbacks:
- every change in the tags filter sends a new request to the server (should not happen too often in usual usage)
- changing the tags filter removes the builds and readds them, which can be a visible appearance, disappearance

I could add an option to enable this new behavior (disabled by default) so that large buildbot instances can have the advantages while smaller instances are not affected.

## Contributor Checklist:

* ~I have updated the unit tests~ necessary?
* ~I have updated the appropriate documentation~ - no documentation affected to my knowledge

### DIFF
--- a/www/base/src/app/builders/builders.controller.coffee
+++ b/www/base/src/app/builders/builders.controller.coffee
@@ -1,6 +1,6 @@
 class Builders extends Controller
     constructor: ($scope, $log, dataService, resultsService, bbSettingsService, $stateParams,
-        $location, dataGrouperService, $rootScope) ->
+        $location, dataGrouperService, $rootScope, $filter) ->
         # make resultsService utilities available in the template
         _.mixin($scope, resultsService)
         $scope.connected2class = (worker) ->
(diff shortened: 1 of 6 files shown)


www/base/src/app/builders/builders.controller.coffee
```diff
@@ -102,12 +102,28 @@ class Builders extends Controller
...
+
+            filteredBuilds = $filter('filter')($scope.builders, $scope.isBuilderFiltered) || []
+            builderIds = filteredBuilds.map (builder) -> builder.builderid
+            builderIds = [] if builderIds.length == $scope.builders.length
+
+            builds = data.getBuilds(limit: 200, order: '-started_at', builderid__eq: builderIds)
+            dataGrouperService.groupBy($scope.builders, workers, 'builderid', 'workers', 'configured_on')
+            dataGrouperService.groupBy($scope.builders, builds, 'builderid', 'builds')
+
+        $scope.$watch "builders.$resolved", (resolved) -> requeryBuilds() if resolved
```
tardyp: This adds a delay on the list of builds. We should avoid that at least when no tag is selected

www/base/src/app/builders/builders.controller.coffee
```diff
@@ -102,12 +102,28 @@ class Builders extends Controller
...
+        builds = null
+
+        requeryBuilds = () ->
+            $scope.builders.forEach (builder) -> builder.builds = []
+
+            filteredBuilds = $filter('filter')($scope.builders, $scope.isBuilderFiltered) || []
+            builderIds = filteredBuilds.map (builder) -> builder.builderid
+            builderIds = [] if builderIds.length == $scope.builders.length
+
+            builds = data.getBuilds(limit: 200, order: '-started_at', builderid__eq: builderIds)
```
tardyp: while we are at it, we should really get this hardcoded value in the settings, like we do in console and grid

jangmarker: done, also added news fragment for this

www/base/src/app/builders/builders.controller.coffee
jangmarker: done for the case when no tag is selected

tardyp:
Looks like this change adds unstability to the e2e smoke tests. I keep restarting it, and get either error for smokes/2.7 or smokes/3.5.

As much as I would like to get this in 0.9.12, I think we can't take it now.

jangmarker:
@tardyp I've rebased on current master and it seems now the build pass. Please note that I haven't changed anything. The last failed build that I could see before pushing didn't fail because of my changes but because something in the package management seemed to be broken.

tardyp:
ok lets see if this really increase the build failure ratio

PR #3592, opened 2017-09-08
Allows more UTF-8 characters in builder names, worker names, and step names

rodrigc:
On Python 3, I was able to use the following:

image

On Python 2, I wasn't able to use a builder with Cyrillic characters.
This is due to a bug in Twisted's logger in Python 2 which does not fully support unicode.
An attempt was made to fix it on Python 2 here: https://twistedmatrix.com/trac/ticket/8864 , but it was reverted.  The Twisted logger has no problems with unicode on Python 3.

### DIFF
--- a/master/buildbot/util/identifiers.py
+++ b/master/buildbot/util/identifiers.py
@@ -22,7 +22,7 @@ import re
 
 from buildbot import util
 
-ident_re = re.compile('^[a-zA-Z_-][a-zA-Z0-9_-]*$')
+ident_re = re.compile(u'^[a-zA-Z\u00a0-\U0010ffff_-][a-zA-Z0-9\u00a0-\U0010ffff_-]*$', flags=re.UNICODE)
 initial_re = re.compile('^[^a-zA-Z_-]')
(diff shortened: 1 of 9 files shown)


master/buildbot/util/identifiers.py
```diff
@@ -22,7 +22,7 @@
 
 from buildbot import util
 
-ident_re = re.compile('^[a-zA-Z_-][a-zA-Z0-9_-]*$')
+ident_re = re.compile(u'^[a-zA-Z\u00a0-\U0010ffff_-][a-zA-Z0-9\u00a0-\U0010ffff_-]*$', flags=re.UNICODE)
```
seankelly: That's a very large range. What about combining characters and other complications in Unicode? How will that affect this?

seankelly: I also see this regex is in two places. It appears to be the same. Any way to use it only once to prevent deviations of the two?

tardyp:
https://github.com/buildbot/buildbot/issues/3614

This is one of the reason why I don't think it is a good idea to go beyond ascii for identifiers.

PR #3606, opened 2017-09-09
Fixed buildbot master not start error in docker

eshizhan:
#3581 Fixed in entrypoint script.

### DIFF
--- a/master/docker/start_buildbot.sh
+++ b/master/docker/start_buildbot.sh
@@ -45,6 +45,8 @@ if [ ! -f $B/buildbot.tac ]
 then
     cp /usr/src/buildbot/docker/buildbot.tac $B
 fi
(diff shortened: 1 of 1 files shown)


master/docker/start_buildbot.sh
```diff
@@ -45,6 +45,11 @@ if [ ! -f $B/buildbot.tac ]
 then
     cp /usr/src/buildbot/docker/buildbot.tac $B
 fi
+# Fixed buildbot master not start error in docker
+if [ -f $B/twistd.pid ]
+then
+    rm -f $B/twistd.pid
```
tardyp: rm -f is enough.

eshizhan: Repush

PR #3612, opened 2017-09-12
Renderable builder locks

p12tic:
Hi,

Currently step locks can be renderables whereas builder locks can't. This pull request implements renderable support for builder locks.

Could you please look and decide whether this PR is good in principle? I'd like not to spend time refining the PR if the implementation goes to completely wrong direction :-)

Currently there are no tests for the new functionality, some existing tests have been broken by changes of internal APIs and some other things are missing. These all will be fixed on the next iteration if you like the PR.

Thanks a lot for your time!

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -38,6 +38,7 @@ from zope.interface import implementer
 from buildbot import interfaces
 from buildbot import locks
 from buildbot import util
+from buildbot.interfaces import IRenderable
 from buildbot.revlinks import default_revlink_matcher
 from buildbot.util import config as util_config
 from buildbot.util import identifiers as util_identifiers
@@ -911,7 +912,7 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
                 lock_dict[lock.name] = lock
 
         for b in self.builders:
-            if b.locks:
+            if b.locks and not IRenderable.providedBy(b.locks):
                 for lock in b.locks:
                     check_lock(lock)
 
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -195,7 +195,11 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
     deprecatedWorkerClassMethod(locals(), getWorkerName)
 
     @staticmethod
-    def setupProperties(props, requests, builder):
+    def setupPropertiesKnownBeforeBuildStarts(props, requests, builder,
+                                              workerforbuilder):
+        # Note that this function does not setup the 'builddir' worker property
+        # It's not possible to know it until before the actual worker has
+        # attached.
 
         # start with global properties from the configuration
         props.updateFromProperties(builder.master.config.properties)
@@ -205,12 +209,19 @@ class Build(properties.PropertiesMixin, WorkerAPICompatMixin):
         for change in Build.allChangesFromSources(sources):
             props.updateFromProperties(change.properties)
 
-        # and finally, get any properties from requests (this is the path
-        # through which schedulers will send us properties)
+        # get any properties from requests (this is the path through which
+        # schedulers will send us properties)
         for rq in requests:
             props.updateFromProperties(rq.properties)
 
+        # get builder properties
         builder.setupProperties(props)
+
+        # get worker properties
+        # navigate our way back to the L{buildbot.worker.Worker}
+        # object that came from the config, and get its properties
+        worker_properties = workerforbuilder.worker.properties
+        props.updateFromProperties(worker_properties)
 
     def setupOwnProperties(self):
         # now set some properties of our own, corresponding to the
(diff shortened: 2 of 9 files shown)


tardyp:
I think this might not be a very good idea to start breaking the APIs like this.

I would like to have more comments on why you are doing each change. Maybe just inside github comments. 

can't we just take the lock at a time where it is easier to render them?

master/buildbot/config.py
```diff
@@ -900,7 +901,7 @@ def check_lock(lock):
                 lock_dict[lock.name] = lock
 
         for b in self.builders:
-            if b.locks:
+            if b.locks and not IRenderable.providedBy(b.locks):
```
p12tic: This check is needed because renderable locks are not known until builder start and we can't do any checks at config time for these locks.

master/buildbot/process/build.py
```diff
@@ -190,25 +194,23 @@ def getWorkerName(self):
         return self.workerforbuilder.worker.workername
     deprecatedWorkerClassMethod(locals(), getWorkerName)
 
-    def setupProperties(self):
```
p12tic: This was moved to a static method so that we don't need an instance of a Build to know its properties. The BuildRequestDistributor needs the final build properties in order to figure out whether a certain build request will satisfy the lock constraits before creating a build for it.

master/buildbot/process/build.py
```diff
@@ -190,25 +194,23 @@ def getWorkerName(self):
...
+    @staticmethod
+    def setupProperties(props, requests, builder):
 
         # start with global properties from the configuration
-        props.updateFromProperties(self.master.config.properties)
+        props.updateFromProperties(builder.master.config.properties)
 
         # from the SourceStamps, which have properties via Change
-        for change in self.allChanges():
+        sources = requests[0].mergeSourceStampsWith(requests[1:])
```
p12tic: This is the way we compute Build sources in its __init__. If there are more places this is used we could move the sources setup to a separate function, but for now I think the amount of duplication of logic is too small.

master/buildbot/process/build.py
```diff
@@ -225,19 +227,22 @@ def setupOwnProperties(self):
             props.setProperty("codebase", source.codebase, "Build")
             props.setProperty("project", source.project, "Build")
 
-    def setupWorkerForBuilder(self, workerforbuilder):
```
p12tic: Previously setupWorkerForBuilder did 2 things:
 - Figured out the worker properties to add to build properties
 - Assigned the worker to the build

Since the BuildRequestDistributor also needs worker properties to figure out whether a build can be started for specific build request, this method was split into two so that it's possible to retrieve worker properties without creating a build.

We could probably merge this method into setupProperties, because there's no reason for these methods to be separate.

p12tic:
Sorry for a late reply.

I've added several comments to the diff. When implementing this change I tried to reduce the amount of changes to the absolute minimum possible. The primary reason why we change the Build API is that before this PR there was no way to retrieve properties for a build without creating an instance of a Build. This is needed by the BuildRequestDistributor in order to figure out whether a certain build request will satisfy the lock constraits.

If API breakage is the primary concern I could rename the new static Build APIs to something else and bring back the non-static APIs as wrappers.

Thanks a lot!

master/buildbot/process/builder.py
```diff
@@ -261,10 +263,21 @@ def getAvailableWorkers(self):
         return [wfb for wfb in self.workers if wfb.isAvailable()]
     deprecatedWorkerClassMethod(locals(), getAvailableWorkers)
 
-    def canStartWithWorkerForBuilder(self, workerforbuilder):
+    @defer.inlineCallbacks
+    def canStartWithWorkerForBuilder(self, workerforbuilder, buildrequests):
```
p12tic: If we want absolutely preserve this API as is, buildrequests could be set to None by default and the method would fail if any of the locks were renderable when the method was called without build request information.

tardyp:
This looks pretty reasonable to me.

Now there is another exercise needed is the inevitable update of tests, this might show how much deep those APIs are used

p12tic:
I've made the following changes:
 - Split setup of the 'builddir' worker property into a separate method, as it's only available once a worker is attached, which is quite late in the case of latent workers.
 - Because of the above, the number of property setup methods in the Build class started to become confusing. So I've merged Build.setupProperties and Build.setupWorkerProperties into Build.setupPropertiesKnownBeforeBuildStarts. Both of these methods were used in two places throughout the codebase next to each other, so no reason for them to be separate.
 - There were comparatively large use of Builder.canStartWithWorkerForBuilder in tests and it looked like something external code might actually use, so I fixed compatibility with the old API. Now it shows a warning when the `buildrequests` parameter is not provided and in case a renderable lock is configured, an exception is thrown.
 - Added new tests for the new behaviors.
 - Documented and added release notes.

It looks like there are some minor issues remaining, I'll fix them soon. Aside from that, what other problems could you identify in the PR?

Thanks!

p12tic:
Ping :) The tests results are all green, for some reason github has missed an update.

tardyp:
Looks like good work! Please come up with more of those!

PR #3618, opened 2017-09-14
Sourcestamps doc

artagnon:
Simple enough.


rodrigc:
@tardyp looks like some of the buildbots cannot find `docker`: https://nine.buildbot.net/#/builders/13/builds/1281

PR #3621, opened 2017-09-15
PatchParameter

tardyp:
Initial design to allow sending patch via the force scheduler UI.

For now, we add the necessary blocks to the force framework, and make sure a patch is properly applied with an integration test.

UI, relnotes, and docs will be in followup PRs (to keep things reviewable).

## Contributor Checklist:

* [x] I have updated the unit tests
* [x] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [x] I have updated the appropriate documentation

PR #3622, opened 2017-09-15
Force patch ui

tardyp:
UI on top of #3621 

looks like this:
image


tardyp:
closed in favor of #3621

PR #3623, opened 2017-09-15
Running with --nodaemon and pid files

### LINKED ISSUE #3012: Running with --nodaemon and PID files
This ticket is a migrated Trac ticket [3608](http://trac.buildbot.net/ticket/3608)

People contributed to the original ticket: @unknown_contributor
Ticket created on: `Sep 12 2016`
Ticket last modified on: `Sep 12 2016`

---

Both the master and the worker still use PID files while running with --nodaemon. I've had issues with workers running with --nodaemon not starting because of an old, stale PID file, corresponding to an actual process belonging to another user.

This can be thought to be inconsistent, and harms the writing ofreally simple non forking configuration files to manage the process through an external service manager (systemd, supervisord etc.).

It seems that the cousin ``twistd`` executable has a --pidfile argument, that can be used to mean "no pid file". This is explained notably in https://twistedmatrix.com/documents/current/core/howto/systemd.html.

We could expose that, or we could decide that --nodaemon implies no pid file.

I believe this is true in Eight and Nine, and that a fix could be easily backported
(marking this as minor, because it's easy to circumvent by putting appropriate rms in unit files or other confs)

---

mscam:
Add empty pidfile option to master and worker start script when `--nodaemon` option is on. 
Fixes #3012

### DIFF
--- a/master/buildbot/newsfragments/nodaemon-pidfile.bugfix
+++ b/master/buildbot/newsfragments/nodaemon-pidfile.bugfix
@@ -1,0 +1,1 @@
+Add empty pidfile option to master and worker start script when --nodaemon option is on. (:issue:`3012`).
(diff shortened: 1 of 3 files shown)


mscam:
@tardyp newsfragment added :+1:

tardyp:
thanks!

PR #3634, opened 2017-09-20
waterfall: prevent tooltip from flashing in Firefox

rodrigc:
Cherry-picked from https://github.com/buildbot/buildbot/pull/3296 by @Lekensteyn

### DIFF
--- a/www/waterfall_view/src/styles/styles.less
+++ b/www/waterfall_view/src/styles/styles.less
@@ -219,3 +219,7 @@
     }
   }
 }
+
+.svg-tooltip {
+  pointer-events: none;
+}


Lekensteyn:
Thanks, does 0cfa6c2bd6e351a3c479e54661ba84649e6ef2d1 also need to be applied? Last time I checked the tooltip showed on the wrong side of the cursor because of the wrong order

rodrigc:
@Lekensteyn I'm not sure about that.  What would be helpful is if you could submit only https://github.com/buildbot/buildbot/commit/0cfa6c2bd6e351a3c479e54661ba84649e6ef2d1 on a new branch in a new PR, and paste a before and after screenshot showing what it fixes.
For this UI kind of stuff, it is hard to understand the impact without screenshots.

www/waterfall_view/src/module/main.module.coffee
```diff
@@ -470,6 +470,7 @@ class Waterfall extends Controller
             else "150,0 170,#{height / 2} 150,#{height} 0,#{height} 0,0"
         tooltip = e.append('g')
             .attr('class', 'svg-tooltip')
+            .style('pointer-events', 'none')
```
tardyp: I'd rather put this style in styles.less
```css
.svg-tooltip {
pointer-events: 'none';
}
```

rodrigc: OK, I changed it

www/waterfall_view/src/styles/styles.less
```diff
@@ -219,3 +219,7 @@
     }
   }
 }
+
+.svg-tooltip {
+  pointer-events: 'none';
```
Lekensteyn: Any reason why this is not just:

    pointer-events: none;

https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

rodrigc: @Lekensteyn I changed the syntax to match the format on the developer.mozilla.org web site

PR #3636, opened 2017-09-20
Fix GitPoller using peeled tags in the ls-remote output

obynio:
While trying to use the [`only_tags`](http://docs.buildbot.net/current/manual/cfg-changesources.html#gitpoller) option of GitPoller, I ran into a strange issue as GitPoller was unable to detect new git tags and throw a [`GitError`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L37).

It appears that the GitPoller is parsing the output of `git ls-remove` in [`_getBranches`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L138) without using the option `--refs` which hides peeled tags and pseudorefs. In my case, [`_getBranches`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L138) was parsing both the lightweight and annotated refs of the same tag and passing both of them to [`git fetch`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L188) which raised an error 128.

Without using [`--refs`,](https://git-scm.com/docs/git-ls-remote.html#git-ls-remote---refs) with `git ls-remote`, [`_getBranches`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L138) returns the following result
```
da39a3ee5e6b4b0d3255bfef95601890afd80709	refs/tags/mytag
d74880752203616894cfb3db7ad644c086497992	refs/tags/mytag^{}
```
Note the `^{}` which indicates the annotated ref of the tag.

Using [`--refs`,](https://git-scm.com/docs/git-ls-remote.html#git-ls-remote---refs) we can hide peeled tags so [`_getBranches`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L138) returns the following result
```
da39a3ee5e6b4b0d3255bfef95601890afd80709	refs/tags/mytag
```
Which is the result expected. This PR fix this issue.

Regards,

### DIFF
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -135,7 +135,7 @@ class GitPoller(base.PollingChangeSource, StateMixin):
         return str
 
     def _getBranches(self):
-        d = self._dovccmd('ls-remote', [self.repourl])
+        d = self._dovccmd('ls-remote', ['--refs', self.repourl])
 
         @d.addCallback
         def parseRemote(rows):
--- a/master/buildbot/test/unit/test_changes_gitpoller.py
+++ b/master/buildbot/test/unit/test_changes_gitpoller.py
@@ -706,7 +706,7 @@ class TestGitPoller(gpo.GetProcessOutputMixin,
     def test_poll_allBranches_single(self):
         self.expectCommands(
             gpo.Expect(b'git', b'init', b'--bare', b'gitpoller-work'),
-            gpo.Expect(b'git', b'ls-remote', self.REPOURL)
+            gpo.Expect(b'git', b'ls-remote', b'--refs', self.REPOURL)
             .stdout(b'4423cdbcbb89c14e50dd5f4152415afd686c5241\t'
                     b'refs/heads/master\n'),
             gpo.Expect(b'git', b'fetch', self.REPOURL,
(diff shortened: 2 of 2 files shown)


seankelly:
Tests need to be updated.

Do you know when `--refs` was included?

obynio:
According to their man, the option `--refs` was added to `ls-remote` since Git 2.8.0.

seankelly:
Note: Ubuntu 16.04 includes git 2.7.4. Debian 9 includes git 2.11. CentOS 7 includes git 1.8. I don't think this can be merged without a version check. Too many stable distros don't include a new enough git yet.

rodrigc:
The actual command `--refs` flag was added here:
https://github.com/git/git/commit/2718ff098a1cd1bf48fc230164259b56e3e7178f

This lines up with git v1.4.2

The ls-remote command was first documented here (after 2718ff):
https://github.com/git/git/commit/972b6fe746f0d9a3ea13118d80d9371a9ff68dba
but `--refs` wan't mentioned in the docs for ls-remote, but that was an omission, since
the flag was already there.

Documentation for the `--refs` flag to `ls-remote` was added here:
https://github.com/git/git/commit/40a88529080295dd9a4218e1a72d1e8143ee4f7b

rodrigc:
@seankelly Based on reading the source code of git, `ls-remote --refs` is in all the versions of git that you mentioned.

With this version of git:
```
git version 1.8.3.1
```

I tried:

```
git ls-remote --refs https://github.com/git/git
```

and it worked.

I don't think a version check is so critical for this change, since this flag has been in git for quite a while, and all the mainstream distributions have it.

rodrigc:
@obynio you definitely do need to fix the unit tests which are broken by your change.
Can you fix those tests and push those changes to your branch so that the buildbots can run it?

obynio:
@rodrigc yes, I will fix that, thanks

obynio:
Tests seems to be fixed now :)

obynio:
Tests seems to be fixed and commits are squashed.

seankelly:
No problem here since the docs were very late about documenting the option.

PR #3640, opened 2017-09-21
Handle case where git step runs on worker with no git binary installed

### LINKED ISSUE #3637: Git step tries to extract the git version, even if output is not from Git
If on a system you have a git command (but somehow it is a rogue one and don't output anything) or if your Twisted spawn_process goes wild (ex on Windows)

you can end up with this error on Buildbot

```
buildbot/steps/source/git.py", line 546, in checkSupport
    version = stdout.strip().split(' ')[2]
exceptions.IndexError: list index out of range
```

Here 
https://github.com/buildbot/buildbot/blob/5f995bf3f82bade2eab9a95cade97e6864b560f2/master/buildbot/steps/source/git.py#L588

all the check for specific git versions should be done only if git is installed, that is the lines should be indented.

-----

To reproduce it, set up a git command which is not git, but a bogus one.

Buildbot should detect that it is not git, and fail with the more appropriate error

```
    raise BuildSlaveTooOldError("git is not installed on slave")
buildbot.interfaces.BuildSlaveTooOldError: git is not installed on slave
```

rodrigc:
Fixes #3637

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -583,9 +583,13 @@ class Git(Source):
         stdout = yield self._dovccmd(['--version'], collectStdout=True)
 
         gitInstalled = False
+        version = "0.0.0"
         if 'git' in stdout:
             gitInstalled = True
-        version = stdout.strip().split(' ')[2]
+            try:
+                version = stdout.strip().split(' ')[2]
+            except IndexError:
+                gitInstalled = False
         if LooseVersion(version) < LooseVersion("1.6.5"):
             self.supportsBranch = False
         if LooseVersion(version) < LooseVersion("1.7.8"):
--- a/master/buildbot/test/unit/test_steps_source_git.py
+++ b/master/buildbot/test/unit/test_steps_source_git.py
@@ -16,10 +16,12 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+from twisted.internet import defer
 from twisted.internet import error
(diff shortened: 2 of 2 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -583,9 +583,13 @@ def checkBranchSupport(self):
...
 
         gitInstalled = False
+        version = "0.0.0"
         if 'git' in stdout:
             gitInstalled = True
-        version = stdout.strip().split(' ')[2]
+            try:
+                version = stdout.strip().split(' ')[2]
+            except IndexError:
+                gitInstalled = False
```
tardyp: I think we should also handle the case where split(' ')[2] exists but do not return a correct version

rodrigc: I don't know what an incorrect version is.  This code https://github.com/buildbot/buildbot/blob/master/master/buildbot/steps/source/git.py#L589

uses LooseVersion.  According to http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html ,
with LooseVersion, there is no such thing as an invalid version number.

ewongbb: would "a1" be considered an invalid version #?

rodrigc: `LooseVersion` accepts anything as a version, including "a1"

tardyp: okay. If there is another use case later, we can fix it at that time.

PR #3650, opened 2017-09-24
Fixed a deadlock bug related to distinct but equal BuildSteps

egirault:
If a build factory has a buildstep with a lock, this will trigger a deadlock whenever 3 or more builds of this factory are started at the same time. The main reason is that BaseLock uses the == operator to compare the owners of a given lock. But if owners are BuildSteps, it doesn't work because they inherit from ComparableMixin (#d6c6b550), so two distinct BuildStep objects sharing the same properties will be considered as equal, although they are distinct Python objects: `b0 == b1` but `b0 is not b1`. Because of this, the first two buildsteps acquiring the lock were able to complete, but then the `waiting` list was empty and the remaining buildsteps were waiting forever. It was not possible to stop them with the `Stop` button of the web interface; only a master restart stopped them.
We use the `is` operator instead of `==` to fix this.

### DIFF
--- a/master/buildbot/data/buildsets.py
+++ b/master/buildbot/data/buildsets.py
@@ -26,6 +26,7 @@ from twisted.python import log
 from buildbot.data import sourcestamps as sourcestampsapi
 from buildbot.data import base
 from buildbot.data import types
+from buildbot.db.buildsets import AlreadyCompleteError
 from buildbot.process.buildrequest import BuildRequestCollapser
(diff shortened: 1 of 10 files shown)


tardyp:
Hi @egirault I added an integration test in your branch, which found another case of ==, which I also fixed.

egirault:
Yes indeed I forgot one `!=`. Thanks!

egirault:
Thank you!

PR #3148, opened 2017-04-25
allow github reporter to work with git@ urls

dragon512:
Allow GitHub to parse https:// and git@github.com:xxx style repository URLS correctly
Add some verbose messages that can be useful.

## Contributor Checklist:

* [x] I have updated the unit tests
* [x] I have created a file in the master/buildbot/newsfragment directory (and read the README.txt in that directory)
* [x] I have updated the appropriate documentation

PR #3682, opened 2017-10-07
Minor fix : On rebuild, giturl is empty. Don't backtrace out

anish:
After https://github.com/buildbot/buildbot/pull/3148 merged, the fix in https://github.com/buildbot/buildbot/pull/2724 needs to extended a bit

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -141,6 +141,8 @@ class GitHubStatusPush(http.HttpStatusPushBase):
             repoOwner, repoName = project.split('/')
         else:
             giturl = giturlparse(sourcestamps[0]['repository'])
+            if not giturl:
+                return
             repoOwner = giturl.owner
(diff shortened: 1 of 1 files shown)


master/buildbot/reporters/github.py
```diff
@@ -141,6 +141,8 @@ def send(self, build):
             repoOwner, repoName = project.split('/')
         else:
             giturl = giturlparse(sourcestamps[0]['repository'])
+            if not giturl:
```
tardyp: Well, it is not normal that sourcestamp is not set on rebuild. github reporter should work on rebuild.

It does work on nine.buildbot.net(I often rebuild, and it updates the status on github), so I think there is an issue with your configuration.

anish: ok, I will investigate this more. Might just be a side effect of #3683 even ( the last few changes to github reports/hooks seems to have broken GHE a lot )

Issue #3661, opened 2017-09-29
Builds doesn't get scheduled even when worker is idle


aj062:
I have noticed many times, that some workers stay idle for long time, even when the corresponding builder has a lot of pending builds. The builder have just one worker associated, and that worker is idle (not running any build). Still, builds are not getting scheduled on the worker even though builder have pending buildrequests.

twistd.log on worker/master doesn't have any error. Just have connection events. e.g.: "Worker bot11 attached to mybuilder-release-builder"

master running: Buildbot 0.9.11, Twisted: 17.5.0, Python: 2.7.5
worker running: buildbot-worker 0.9.11 or buildslave-0.8.12

What can I do to debug this further?

aj062:
Can this issue be caused due to not doing a clean "buildbot stop"? 
https://github.com/buildbot/buildbot/issues/3535

Maybe buildbot still thinks that the worker is building something.

Is there is a way we can check in database (postgres) what's the current state of the worker, to debug this further?

aj062:
Looking at this https://git.io/vdEAm, I think there are multiple issues. This happens more frequently in case of large number of builders.

1. There is delay(~5-30s) between each iteration of activityLoop (probably acquiring locks or _maybeStartBuildsOnBuilder (https://git.io/vdEs5) takes time)
2. _pending_builders is not processed in FIFO order, so activityLoop might never reach some of the builders, because it just keep processing new builders added to _pending_builders. It seems to be sorted here: https://git.io/vdEGL
3. efficiency: buildrequestdistributor is processing just one builder per iteration in activityLoop https://git.io/vdEGY, probably it can process multiple one.

tardyp:
> There is delay(~5-30s) 
I think it might be related to  the discussion we had a little while ago:
https://github.com/buildbot/buildbot/issues/3395
How many pending buildrequests do you have?
Is it a bit better if you flush out old ones?

>It seems to be sorted here:
I think it is sorted only when there is a new buildrequest or a finished build (a new worker available)

> one builder per iteration 
It well processing several builders per iteration would just add a second loop, and this will actually be the same sequence of processing.
pending_builders.pop(0) is actually poping the processed builder, so next loop will process the rest unless there is another sort in between. 
Maybe indeed this is our problem and we should have the pending_builder locked until it is empty.

aj062:
> How many pending buildrequests do you have?

~10,000 pending buildrequests.

[...]

PR #3685, opened 2017-10-09
Distributor optim

tardyp:
WIP fix for #3661

### DIFF
--- a/master/buildbot/newsfragments/getOldersBuildrequest.bugfix
+++ b/master/buildbot/newsfragments/getOldersBuildrequest.bugfix
@@ -1,0 +1,1 @@
+Improve ``getOldestRequestTimeOptim`` function of buildrequest distributor to to sorting and paging in the database layer.
--- a/master/buildbot/process/buildrequestdistributor.py
+++ b/master/buildbot/process/buildrequestdistributor.py
@@ -16,6 +16,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import copy
 import random
 from datetime import datetime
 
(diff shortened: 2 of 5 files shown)


master/buildbot/newsfragments/getOldersBuildrequest.bugfix
```diff
@@ -0,0 +1 @@
+Improve ``getOldestRequestTimeOptim`` function of buildrequest distributor to to sorting and paging in the database layer.
```
seankelly: Doubled "to". Is the second one supposed to be "do" instead?

master/buildbot/process/buildrequestdistributor.py
```diff
@@ -472,21 +473,28 @@ def _activityLoop(self):
...
 
-            # bail out if we shouldn't keep looping
-            if not self.running or not self._pending_builders:
+                # bail out if we shouldn't keep looping
+                if not self._pending_builders:
+                    self.pending_builders_lock.release()
+                    self.activity_lock.release()
+                    break
+                # take that builder list, and run it until the end
+                pending_builders = copy.copy(self._pending_builders)
```
seankelly: Why the copy?

tardyp: If you look at how locks are setup, it is not guaranteed that pending_builders will not be messed up while we are running the loop. 

Current implementation is not doing it, as it is playing with set and sorts. The locks looks useless as far as I can read. But as this stuff is pretty central, I prefer to be overcautious.

tardyp:
included in #3684

PR #3686, opened 2017-10-09
Improve Builder infrastructure for priority functions

jsonn:
The default priority function sorts by `getOldestRequestTime` and that differs subtly from the documentation. If two requests are collapsed, the submission time of the winner stays, so with the default behavior, starvation can occur. The new `getNewestCompleteTime` can be used to avoid that by sorting on finished jobs instead, i.e. preferring the builder that has waited for the longest.

The second part cleans up `getOldestRequestTime` to better use the data layer, i.e. when many unclaimed builds are around.

Documentation changes can be done separately, including an example using gNCT.

### DIFF
--- a/master/docs/manual/customization.rst
+++ b/master/docs/manual/customization.rst
@@ -175,6 +175,26 @@ A simple ``prioritizeBuilders`` implementation might look like this::
 
     c['prioritizeBuilders'] = prioritizeBuilders
 
+If the change frequency is higher than the turn-around of the builders,
+the following approach might be helpful:
+
+.. code-block:: python
+
+    def prioritizeBuilders(buildmaster, builders):
+        """Prioritize builders. First, prioritize inactive builders.
+        Second, consider the last time a job was completed (no job is infinite past).
+        Third, consider the time the oldest request has been queued.
+        This provides a simple round-robin scheme that works with collapsed builds."""
+
+        def isBuilding(b):
+            return bool(b.building) or bool(b.old_building)
+
+        builders.sort(key = lambda b: (isBuilding(b), b.getNewestCompleteTime(), b.getOldestRequestTime()))
+        return builders
+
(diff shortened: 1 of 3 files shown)


tardyp:
> The new getNewestCompleteTime can be used to avoid that by sorting on finished jobs instead, i.e. preferring the builder that has waited for the longest.

I am not sure if this will really work.Won't it un prioritize a builder that has never built?

jsonn:
Correct, a new builder will get priority. That's part of the reason why I want to keep the question of default behavior separate -- it may or may not work for everyone. A combination i.e. sorting by `(b.getNewestCompleteTime(), b.getOldestRequestTime())` might work best,  but I'm still experimenting with that part.

Frodox:
Request also for docs updating/examples :)

jsonn:
I give up. I have no idea how to appease the spell checker for code...

tardyp:
looks good. sorry you had to try and error for rst fix. could you please squash those commits?

master/docs/manual/customization.rst
```diff
@@ -175,6 +175,26 @@ A simple ``prioritizeBuilders`` implementation might look like this::
 
     c['prioritizeBuilders'] = prioritizeBuilders
 
+If the change frequency is higher than the turn-around of the builders,
+the following approach might be helpful:
+
```
tardyp: isn't it trying to workaround the issue that is fixed here: https://github.com/buildbot/buildbot/pull/3685 ?

jsonn: I don't think so. In my case, the master is not busy at all. It's just that the default priority function interacts badly with the collapse logic. New changes will have the same submission time for all builders of that scheduler and a collapsed build will retain its submission time. As such, the later builders are never considered as long as there is a steady flow of new changes.

Issue #3603, opened 2017-09-08
Grid View broken for Virtual Builders


btaidm:
Grid View is currently broken with virtual builders. 

New (0.9.11):
![image](https://user-images.githubusercontent.com/3837347/30231666-fd6991b6-94b9-11e7-860e-e493ff49d220.png)

Old (0.9.10):
![image](https://user-images.githubusercontent.com/3837347/30231708-1f8106a8-94ba-11e7-986e-dc106b753188.png)

tardyp:
arg. Meanwhile, you can upgrade all other components, but keep gridview 0.9.10 :-/

tardyp:
@rjarry do you have an idea what could be wrong?

rjarry:
No, I don't see what could be the problem. It seems to be working on https://nine.buildbot.net/#/grid

Does it run on the latest version?

btaidm:
No. It is on 0.9.10

rodrigc:
nine.buildbot.net is running grid_view plugin 0.9.10

rodrigc:
Fixed in https://github.com/buildbot/buildbot/pull/3689

Issue #3688, opened 2017-10-09
[grid_view] updating 0.9.10 to 0.9.12 leaves the grid with a single row


aureliendavid:
Hello, 

with 0.9.10 I have a grid view that looks like this:

![bbgrid1](https://user-images.githubusercontent.com/7168263/31344179-3637e388-ad12-11e7-9885-e9a55c58d668.png)

after updating to 0.9.12 (with pip) I get

![bbgrid2](https://user-images.githubusercontent.com/7168263/31344195-4355c56c-ad12-11e7-81f6-6a3538824f44.png)

any idea what happened?

Thanks in advance.

bdbaddog:
I didn't have grid enabled under 0.9.10, but I see the same with 0.9.12.

rodrigc:
I can confirm this problem.  For example, look at: http://buildbot.python.org/all/#/grid

It looks like maybe this patch is causing problems:
https://github.com/buildbot/buildbot/pull/3558

@rjarry can you comment?

rodrigc:
Same issue reported here:
https://github.com/buildbot/buildbot/issues/3603

rjarry:
Hi @rodrigc ,

After a quick `git bisect`:

```
ad04194769415097deabdd8f249103e9db4bfbf7 is the first bad commit
```

Indeed, the [merge commit](https://github.com/buildbot/buildbot/commit/ad04194769415097deabdd8f249103e9db4bfbf7#diff-841a6c04c85578283ace7ae9cd9525eaL142) looks bad :)

I'll fix that right away.

tardyp:
I'll release 0.9.12.post1 for grid_view when we have a working fix.

tardyp:
0.9.12.post1 is on pypi now.

aureliendavid:
great, it's back to normal now

thanks all

PR #3689, opened 2017-10-10
Fix broken grid_view

rjarry:
There was a glitch during a merge.

This fixes issues #3603 and #3688.


rodrigc:
@rjarry: I confirm, this fix works.  Thanks!

Issue #3363, opened 2017-06-23
[UI][Console View] Previous changes converted into unknown revision (a few seconds ago)


Frodox:
I love Console view plugin a lot. It is what I need for my workflow (with some improvements).

But I see noisy bug. All is fine when I just started bb and have <=25 changes (I use GitLab web hook only, no fake changes). But when I have 26, 27, 28 ++ changes, first one (1, 2, 3) becomes 'unknown revision'. But in fact it was fine, and build logs show it is fine. I think previous changes should be just hidden and not pull-ed in.
But in best way I want to increase list if changes from 25 to 50, 100, and have pagination for old one.
![selection_011](https://user-images.githubusercontent.com/3653069/27477594-48e5a15c-5815-11e7-8ac8-7dc0f329b166.png)
![selection_012](https://user-images.githubusercontent.com/3653069/27477595-48e5dde8-5815-11e7-9675-fb8381c5a7ca.png)
![selection_013](https://user-images.githubusercontent.com/3653069/27477596-48ee5090-5815-11e7-8424-4b1bfde310cf.png)

estan:
I think we're suffering from this as well. For example:

![console](https://user-images.githubusercontent.com/86675/29994194-18ef346c-8fca-11e7-8864-4a9725677501.png)

tardyp:
You can increase the number of fetched changes in the settings page. We
should eventually find an heuristique to automatically fetch me changes.

estan:
@tardyp: Ah,  but the problem is not the number of changes (that's OK for me), but that revisions are shown as "unknown".

Frodox:
@estan do you use ForceScheduler also?

estan:
@Frodox Yes, not sure if all those unknown revisions in my screenshot are forced builds or not though.

But in any case, surely Buildbot should be able to get the info about the `got_revision` (the one that was actually fetched by the forced build) and present that?

s0undt3ch:
We use the force scheduler but we have way more unknown revisions the the number of forced builds we triggered.
We do however base most triggering on GH events and not polling...

tardyp:
So it looks like there is an issue with the UI decoding the got_revision value. Anybody want to debug this?

This needs to set a breakpoint here:
https://github.com/buildbot/buildbot/blob/master/www/console_view/src/module/main.module.coffee#L270

estan:
Sorry for the late reply, I followed the quick start for getting a `master` version of the UI running against our master.

[...]

PR #3691, opened 2017-10-13
console_view: match forced builds on got_revision if possible (#3363)

aureliendavid:
This is a quick implementation of what was discussed here: https://github.com/buildbot/buildbot/issues/3363#issuecomment-336092464

**Before:** 

---

buildbot-issue-before

---

Both builds 140 and 141 are building the same `got_revision`. 

140 is triggered by a `SingleBranchScheduler`
141 is triggered by a `ForceScheduler`

**After:**

---

buildbot-issue-after

---
Both builds are properly grouped under the same `change`.

There may be better ways to do this, it's just a try.

### DIFF
--- a/www/console_view/src/module/main.module.coffee
+++ b/www/console_view/src/module/main.module.coffee
@@ -103,8 +103,10 @@ class Console extends Controller
         @sortBuildersByTags(@all_builders)
 
         @changesBySSID ?= {}
+        @changesByRevision ?= {}
         for change in @changes
             @changesBySSID[change.sourcestamp.ssid] = change
+            @changesByRevision[change.revision] = change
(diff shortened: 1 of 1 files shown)


jsonn:
I think the correct way forward consists of two parts:
(1) Make it possible to fetch the changes lazily.
(2) Use the indirection build -> buildrequest -> buildset -> sourcestamps -> ssids

The second part should be easy to integrate here, the first half is more involved but helps with active masters in general.

tardyp:
fetching changes lazily is a great followup indeed.
Having to go through all the indirection is what we tried to avoid by using this shortcut. We try to limit the number of rest api calls

PR #3700, opened 2017-10-17
Give renderables in reporters access to master singleton

### LINKED ISSUE #3082: Renderable in reporters shall have access to master singleton
At the moment, they dont have, which makes it impossible to access data api.

https://github.com/buildbot/buildbot/pull/3064

anish:
Also needs build props to actually be able to address steps

This works to get build step output in GitHubCommentPush, but via a complicated renderer. I'm wondering if there's
some variation of getDetailsForBuildset that should be used instead.

Fixes #3082 @tardyp

### DIFF
--- a/master/buildbot/newsfragments/reporters_master_singleton.feature
+++ b/master/buildbot/newsfragments/reporters_master_singleton.feature
@@ -1,0 +1,1 @@
+Give reporters access to master single in renderables. This allows access to build logs amongst other things
--- a/master/buildbot/reporters/bitbucketserver.py
+++ b/master/buildbot/reporters/bitbucketserver.py
@@ -75,6 +75,8 @@ class BitbucketServerStatusPush(http.HttpStatusPushBase):
     @defer.inlineCallbacks
     def send(self, build):
         props = Properties.fromDict(build['properties'])
+        props.master = self.master
+
         results = build['results']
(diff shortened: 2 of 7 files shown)


master/buildbot/process/properties.py
```diff
@@ -83,6 +84,14 @@ def master(self, value):
         self._master = value
 
     @property
+    def buildProps(self):
```
tardyp: I dont think that we should do this.

If a renderer needs to use the steps, then it will need to call the data api to fetch the steps.
using 
```
props.master.data.get(('builders', props.getProperty('buildername'), 'builds', props.getProperty('buildernumber'), 'steps')
```

anish: I had to change it to 
`yield props.master.data.get(('builders', props.getProperty('buildername'), 'builds', props.getProperty('buildnumber'), 'steps'))`

but that works, will update the code. I'm still a little lost on why the step logs have an `h` or `e` before them. Is there any guide to this, or an easy way to strip that ?

tardyp:
looks good. This needs a newsfragment, though, and remove [WIP] if you are happy to merge.

anish:
@tardyp good to go now, added news fragment and an example

tardyp:
thanks!

PR #3712, opened 2017-10-24
Fix getBuildRequests results when using multiple sourcestamps

obynio:
I recently faced a nasty bug when using multiple codebases. It appears that [`getBuildResquests`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/buildrequests.py#L84) returns build requests with the same `id` several times when using multiple codebases in the same build. After some debugging, I discovered that when you use several codebases (which generate several sourcestamps) in the build, `getBuildRequests` does not group them and thus returns several build requests with the same `id`.

As always, an example is more meaningful... Here, we have a buildrequest which use several sourcestamps. The SQL query used in `getBuildRequests` is generated in [`_saSelectQuery`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/buildrequests.py#L48) and join together several models including [`buildrequests`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/model.py#L73), [`buildset_sourcestamps`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/model.py#L451) and [`sourcestamps`](https://github.com/buildbot/buildbot/blob/master/master/buildbot/db/model.py#L416). However, when using multiple sourcestamps, the join will duplicate the buildrequest and `getBuildRequests` will return them.

Here is the current SQL query result of a buildset which use two codebases:

```
id          buildsetid  builderid   priority    complete    results     submitted_at  complete_at  waited_for  brid        masterid    claimed_at  branch      repository  codebase    buildername
----------  ----------  ----------  ----------  ----------  ----------  ------------  -----------  ----------  ----------  ----------  ----------  ----------  ----------  ----------  -----------
435         331         40          0           1           0           1508832228    1508832236   1           435         1           1508832228                          production  sdk-android
435         331         40          0           1           0           1508832228    1508832236   1           435         1           1508832228                          internal    sdk-android
```

But if we perform a `GROUP BY buildrequests.id` on the query, we ensure that the same build request is not duplicated by the joined sourcestamps before returning the `getBuildRequests` result:

```
id          buildsetid  builderid   priority    complete    results     submitted_at  complete_at  waited_for  brid        masterid    claimed_at  branch      repository  codebase    buildername
----------  ----------  ----------  ----------  ----------  ----------  ------------  -----------  ----------  ----------  ----------  ----------  ----------  ----------  ----------  -----------
435         331         40          0           1           0           1508832228    1508832236   1           435         1           1508832228                          production  sdk-android
```

From my tests, it solves two bugs that I noticed:
* When using multiple codebases in your builset and a mailnotifier with the option `buildSetSummary=True`, buildbot was sending an email with the same text repeated within due to multiple buildrequests returned to the notifier [right here](https://github.com/buildbot/buildbot/blob/master/master/buildbot/reporters/utils.py#L54).
* When using multiple codebases in your builset and fetching the buidrequests on [http://localhost:8020/api/v2/buildrequests](http://localhost:8020/api/v2/buildrequests), the same buildrequests was returned several times.

### DIFF
--- a/master/buildbot/db/buildrequests.py
+++ b/master/buildbot/db/buildrequests.py
@@ -83,6 +83,10 @@ class BuildRequestsConnectorComponent(base.DBConnectorComponent):
 
     def getBuildRequests(self, builderid=None, complete=None, claimed=None,
                          bsid=None, branch=None, repository=None, resultSpec=None):
+
+        def deduplicateBrdict(brdicts):
+            return list(({b['buildrequestid']: b for b in brdicts}).values())
+
         def thd(conn):
             reqs_tbl = self.db.model.buildrequests
             claims_tbl = self.db.model.buildrequest_claims
@@ -116,14 +120,13 @@ class BuildRequestsConnectorComponent(base.DBConnectorComponent):
                 q = q.where(sstamps_tbl.c.repository == repository)
 
             if resultSpec is not None:
-                return resultSpec.thd_execute(
+                return deduplicateBrdict(resultSpec.thd_execute(
                     conn, q,
-                    lambda r: self._brdictFromRow(r, self.db.master.masterid))
+                    lambda r: self._brdictFromRow(r, self.db.master.masterid)))
 
             res = conn.execute(q)
 
-            return [self._brdictFromRow(row, self.db.master.masterid)
-                    for row in res.fetchall()]
+            return deduplicateBrdict([self._brdictFromRow(row, self.db.master.masterid) for row in res.fetchall()])
         return self.db.pool.do(thd)
 
     def claimBuildRequests(self, brids, claimed_at=None, _reactor=reactor):
--- a/master/buildbot/newsfragments/duplicate-brid.bugfix
+++ b/master/buildbot/newsfragments/duplicate-brid.bugfix
@@ -1,0 +1,1 @@
+Fix duplicate build requests results in :py:class:`~buildbot.db.buildrequests.BuildRequestsConnectorComponent` when querying the database  (:issue:`3712`).
--- a/master/buildbot/test/unit/test_db_buildrequests.py
+++ b/master/buildbot/test/unit/test_db_buildrequests.py
@@ -319,6 +319,11 @@ class Tests(interfaces.InterfaceTests):
                                branch='branch_A', repository='repository_A'),
             fakedb.BuildsetSourceStamp(buildsetid=self.BSID + 3,
                                        sourcestampid=self.BSID + 3),
+            # multiple sourcestamps on the same buildset are possible
+            fakedb.SourceStamp(id=self.BSID + 4,
+                               branch='branch_B', repository='repository_B'),
+            fakedb.BuildsetSourceStamp(buildsetid=self.BSID + 3,
+                                       sourcestampid=self.BSID + 4),
         ])
         d.addCallback(lambda _:
                       self.db.buildrequests.getBuildRequests(**kwargs))


rodrigc:
Some of the unit tests are breaking with this change such as this:

https://travis-ci.org/buildbot/buildbot/jobs/292071129#L641

Can you fix this?
Also if you can add a test to either test_db_buildrequests.py, or test_process_buildrequest.py or tests_data_buildrequests.py to validate your fix?

obynio:
Tests seems to be fixed and commits are squashed.

tardyp:
> As always, an example is more meaningful...

We find unit tests even more meaningful :) Could you please add a unit test for your use case?

obynio:
I didn't have time to write some. I'll take care of that tomorrow.

master/buildbot/db/buildrequests.py
```diff
@@ -111,9 +112,24 @@ def thd(conn):
                 q = q.where(reqs_tbl.c.buildsetid == bsid)
 
             if branch is not None:
-                q = q.where(sstamps_tbl.c.branch == branch)
+                q = q.column(sstamps_tbl.c.branch).where(sstamps_tbl.c.branch == branch)
             if repository is not None:
-                q = q.where(sstamps_tbl.c.repository == repository)
+                q = q.column(sstamps_tbl.c.repository).where(sstamps_tbl.c.repository == repository)
+
+            # ONLY_FULL_GROUP_BY enabled by default in MySQL obliges us to indicate all the group_by columns
```
tardyp: obliges looks franglish :) requires looks better to me. That said, it is not clear to me why that huge group_by is needed.
Isn't distinct more appropriate?
http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.distinct

obynio:
Yes, it seems to be an issue with ONLY_FULL_GROUP_BY enabled by default

master/buildbot/db/buildrequests.py
obynio: I thought about that yesterday, but unfortunately using dinstinct() had no effect :/

tardyp:
I think lazily joining only if branch or repo filtering is enabled is a first idea, but then you still have the risk of having duplicate if e.g you only filter on branch, and both codebase are on the same branch.

I googled a bit, and it looks like the proper way would be do go with subqueries, which have lots of db implementation specifics. So I doubt we can really make a portable query with sqlalchemy.
So what I would do is to lazily join, and in case of join do the dedupe in python (by putting the results into a dictionary keyed by brid)

obynio:
Okay, I get it. Seems to be a great idea. I will explore that direction asap, thanks :)

obynio:
Nice, seems that the python dedupe is working well :)

master/buildbot/db/buildrequests.py
```diff
@@ -64,8 +64,7 @@ def _saSelectQuery(self):
         from_clause = from_clause.join(builder_tbl,
```
tardyp: I think we said we should only join in case of sourcestamp filtering (branch or repository)

tardyp:
@obynio I think this is a good change, but it needs a bit more work. Will you be able to finish it?

obynio:
@tardyp yup, here are the latest squashed changes, seems to work nicely.

tardyp:
@obynio thanks! there is a need for a newsfragment and that is good to go!

obynio:
@tardyp just added a newsfragment and the build seems to fail, maybe an issue with the ci ?

tardyp:
thanks!

PR #3721, opened 2017-10-27
Transform arguments from bytes to strings.

### LINKED ISSUE #3720: REST args in GitLab push notification do not work
This is fundamentally the same problem we had at #3506 (@rodrigc). In Python3, some strings are now byte strings and are not identified properly by Buildbot.

In this case, I have a local gitlab instance with the webhook for push events: `https://mybuildbot/change_hook/gitlab?project=lthmb`.

However, when I test this the Buildbot logs show no `project` parameter set:
```
2017-10-27 18:15:14+0200 [-] injected change Change(revision='2afc1ab3ab28980db47d64714852aa0d586be7f9', who='Paulo Matos <pmatos@linki.tools>', branch='master', comments="Merge branch 'foobar'\n", when=1506599131, category='push', project='', repository='git@linki:pmatos/lthmb.git', codebase='')
```
I went ahead and printed the request received from gitlab in `www/hooks/gitlab.py` and noticed that `request.args` in `getChanges` are byte strings: `2017-10-27 18:15:14+0200 [-] {b'project': [b'lthmb']}`

I think in Buildbot nine with Py3 we might need string conversion here.

pmatos:
This change works for me locally. Transform bytes in rest args to normal strings. Fixes #3720.

### DIFF
--- a/master/buildbot/newsfragments/gitlab-args.bugfix
+++ b/master/buildbot/newsfragments/gitlab-args.bugfix
@@ -1,0 +1,1 @@
+Arguments passed to GitLab push notifications now work with Python 3.  Fixes (:issue:`3720`).
--- a/master/buildbot/test/unit/test_www_hooks_gitlab.py
+++ b/master/buildbot/test/unit/test_www_hooks_gitlab.py
@@ -280,7 +280,7 @@ class TestChangeHookConfiguredWithGitChange(unittest.TestCase):
     def testGitWithChange_WithProjectToo(self):
         self.request = FakeRequest(content=gitJsonPayload)
         self.request.uri = b"/change_hook/gitlab"
-        self.request.args = {'project': ['MyProject']}
+        self.request.args = {b'project': [b'MyProject']}
         self.request.received_headers[_HEADER_EVENT] = b"Push Hook"
         self.request.method = b"POST"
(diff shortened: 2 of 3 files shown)


pmatos:
@rodrigc  sure, I will take a look.

pmatos:
Now using `bytes2unicode`. No failures in python 2.7.13:
```
PASSED (skips=23, successes=5162)
```

tardyp:
you should also make sure the strings in the test reflects the type of the real thing

rodrigc:
@pmatos thanks for the feedback.
Can you do a quick test against an actual GitLab setup and provide feedback?  I don't have a GitLab setup.
In the past, I found that for GitHub, although the tests passed on Python 3, when run against an actual GitHub system, there were errors, and I fixed those.

Thank you for your help.

pmatos:
OK, let me take a look at these issues during the workweek.

rodrigc:
BTW, request.args is used all over the buildbot codebase, and we need to make the same fix that you did in other places.  If you want to help me find and fix those places, that would be much appreciated. :)

pmatos:
@rodrigc Thanks for sorting this out. With my children on holidays this week I have been on the move and without much time to look into it.

PR #3730, opened 2017-11-02
request.args is of the form {b"arg": [b"val"}

rodrigc:


master/buildbot/test/unit/test_www_hooks_bitbucket.py
```diff
@@ -150,12 +150,12 @@ def setUp(self):
 
     @inlineCallbacks
     def testGitWithChange(self):
-        change_dict = {'payload': [gitJsonPayload]}
+        change_dict = {b'payload': [gitJsonPayload]}
 
         request = FakeRequest(change_dict)
```
tardyp: I think we should also change FakeRequest.test_render to ensure that all its attributes are never using native strings. (method, args, uri, prepath, received_headers)

As per twisted spec, all Request args shall be bytes if I understand correctly.

That way we make sure we don't forget another place for this kind of bug.

master/buildbot/test/fake/web.py
```diff
@@ -92,14 +93,28 @@ def _get_child_mock(self, **kw):
 
     # cribed from twisted.web.test._util._render
     def test_render(self, resource):
-        result = resource.render(self)
+        if self.args:
```
tardyp: if self.args is not necessary (with the for just after)

master/buildbot/test/fake/web.py
```diff
@@ -92,14 +93,28 @@ def _get_child_mock(self, **kw):
...
+                        "values which are not bytes".format(args))
+
+        if self.uri and not isinstance(self.uri, bytes):
+                raise ValueError("self.uri: {!r} is {}, not bytes".format(
+                    type(self.uri), self.uri))
+
+        if self.method and not isinstance(self.method, bytes):
+                raise ValueError("self.method: {!r} is {}, not bytes".format(
+                    type(self.method), self.method))
+
```
tardyp: 👍

master/buildbot/test/fake/web.py
```diff
@@ -92,14 +93,28 @@ def _get_child_mock(self, **kw):
 
     # cribed from twisted.web.test._util._render
     def test_render(self, resource):
-        result = resource.render(self)
```
tardyp: I think you forgot to re-add this line.

master/buildbot/newsfragments/poller-bytes.bugfix
```diff
@@ -0,0 +1 @@
+Pollers which use twisted.web.http.Request have been fixed to use bytes, not unicode.  This fixes pollers on Python 3.
```
tardyp: its actually web hooks, not pollers :) I'll fix that when I release

Issue #3733, opened 2017-11-03
TestRunProcess.testPipeString worker tests failing in Win2012/Win64 command prompt .


ewongbb:
while I have managed to install buildbot worker on a win2012 and running in
a command prompt, running the worker tests fail with the following:

===============================================================================
[FAIL]
Traceback (most recent call last):
  File "c:\projects\buildbot\worker\buildbot_worker\test\unit\test_runprocess.py
", line 309, in check
    self.assertTrue({'stdout': nl('a\nb\n')} in b.updates, b.show())
  File "c:\Program Files\Git\python27\lib\site-packages\twisted\trial\_synctest.
py", line 395, in assertTrue
    super(_Assertions, self).assertTrue(condition, msg)
  File "c:\Program Files\Git\python27\lib\unittest\case.py", line 422, in assert
True
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: [{u'header': u'c:\\Program Files\\Git\\python27
\\python.exe -c "import sys; sys.stdout.write(\'b\\na\\n\')" | sort\n in dir tes
t-basedir\n watching logfiles {}\n argv: c:\\Program Files\\Git\\python27\\pytho
n.exe -c "import sys; sys.stdout.write(\'b\\na\\n\')" | sort\n environment:\n  A
LLUSERSPROFILE=C:\\ProgramData\n  APPDATA=C:\\Users\\Administrator\\AppData\\Roa
ming\n  COMMONPROGRAMFILES="C:\\Program\\ Files\\Common\\ Files"\n  COMMONPROGRA
MFILES(X86)="C:\\Program\\ Files\\ (x86)\\Common\\ Files"\n  COMMONPROGRAMW6432=
"C:\\Program\\ Files\\Common\\ Files"\n  COMPUTERNAME=TESTWINBB\n  COMSPEC=C:\\W
indows\\system32\\cmd.exe\n  FP_NO_HOST_CHECK=NO\n  HOMEDRIVE=C:\n  HOMEPATH=\\U
sers\\Administrator\n  LOCALAPPDATA=C:\\Users\\Administrator\\AppData\\Local\n
LOGONSERVER=\\\\TESTWINBB\n  NUMBER_OF_PROCESSORS=2\n  OS=Windows_NT\n  PATH="C:
\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System3
[...]

PR #3734, opened 2017-11-03
Fix issue with windows's sys.executable contains spaces

ewongbb:
This 'fixes' Issue #3733.  

Though I'm not sure I'm going about in the right manner.

### DIFF
--- a/worker/buildbot_worker/runprocess.py
+++ b/worker/buildbot_worker/runprocess.py
@@ -490,7 +490,9 @@ class RunProcess(object):
                 argv = os.environ['COMSPEC'].split()
                 if '/c' not in argv:
                     argv += ['/c']
-                argv += list(self.command)
+                cmd_list = list(self.command)
+                cmd_list[0] = shell_quote(cmd_list[0], self.builder.unicode_encoding)
+                argv += list(cmd_list)
                 self.using_comspec = True
             else:
                 argv = self.command
--- a/worker/buildbot_worker/test/unit/test_util.py
+++ b/worker/buildbot_worker/test/unit/test_util.py
@@ -99,6 +99,21 @@ class TestObfuscated(unittest.TestCase):
         self.assertEqual(1, util.Obfuscated.get_real(cmd))
         self.assertEqual(1, util.Obfuscated.get_fake(cmd))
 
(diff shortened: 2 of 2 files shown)


ewongbb:
What I'm trying to 'parse' for the first command and then to check whether or not it has spaces
in it and if it doesn't have quotes already, we add the quotes for the command.  Since this is my first foray 
into this part of the code, I'm kinda confused as to what the real command is and what the fake command is.  And whether or not this approach is even workable.

tardyp:
please don't refer to trac issues, and avoid requiring to look at the issue to know what is being fixed

tardyp:
I think the problem is around that line:

https://github.com/buildbot/buildbot/blob/master/worker/buildbot_worker/runprocess.py#L499

                display = shell_quote(self.fake_command, self.builder.unicode_encoding)

Looks like at some point, sombody wanted to shell quote the command for windows, but at some other point, somebody only kept the quoting for fake_display (which makes few sense)

anyway, we should probably keep the solution to this problem to the "if windows" block I just pointed

ewongbb:
fwiw.. it wasn't a trac issue that I mentioned. It was a github one.

tardyp:
> fwiw.. it wasn't a trac issue that I mentioned. It was a github one.

Sorry for that grumpiness of mine :) You can always link a github issue by using a '#' before the number.

Anyway, you did not address the other more important remark.
https://github.com/buildbot/buildbot/pull/3734#issuecomment-341705554

Your implementation should be windows specific I think. It will probably fail on linux if somebody is using space in its virtualenv (which we do not test), because linux does not expect the quote its paths.

real_command is the command that is actually executed, and fake_command is the one that is used for any logging. That one is supposed to be cleared from any passwords or secrets.

Please have a look at this shell_quote function I was pointing to you. I think this is the solution, rather than implementing your own.

worker/buildbot_worker/util/__init__.py
```diff
@@ -74,6 +74,30 @@ def to_text(s):
             return s
         return str(s)
 
+    @staticmethod
```
tardyp: I think this is not needed anymore

PR #3740, opened 2017-11-07
Set tags for spot instances in EC2LatentWorker

morrone:
EC2LatentWorker ignores its "tags" parameter when spot_instance is
True.  We move the call to create_tags() from _start_instance(),
which only applies to non-spot instances, to _wait_for_instance(),
which is part of both the normal and spot instance call paths.


tardyp:
looks good! thanks

PR #3743, opened 2017-11-07
fix relnote

tardyp:


tardyp:
@seankelly thanks for heads up!

seankelly:
I need to be quicker on the review. Missed the previous merge by 30 seconds.

tardyp:
actually, I added that new feature in the last minute, so the delay was quite short :-/

PR #3759, opened 2017-11-16
Change container name by moving the hash to the end

dragon512:
Avoids issue when the filter to find container is matching on the start of the string. This causes cases in which we find more than one worker container and it force removes running instances.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -92,7 +92,7 @@ class DockerBaseWorker(AbstractLatentWorker):
         return AbstractLatentWorker.reconfigService(self, name, password, **kwargs)
 
     def getContainerName(self):
-        return ('%s-%s' % ('buildbot' + self.masterhash, self.workername)).replace("_", "-")
+        return ('buildbot-{worker}-{hash}'.format(worker=self.workername, hash=self.masterhash)).replace("_", "-")
 
     @property
     def shortid(self):
@@ -213,11 +213,15 @@ class DockerLatentWorker(DockerBaseWorker):
 
     def _thd_start_instance(self, image, dockerfile, volumes):
         docker_client = self._getDockerClient()
+        container_name = self.getContainerName()
         # cleanup the old instances
         instances = docker_client.containers(
             all=1,
-            filters=dict(name=self.getContainerName()))
+            filters=dict(name=container_name))
+        container_name = "/{0}".format(container_name)
         for instance in instances:
+            if container_name not in instance['Names']:
+                continue
             try:
                 docker_client.remove_container(instance['Id'], v=True, force=True)
             except NotFound:
--- a/master/buildbot/newsfragments/docker-container-name.bugfix
+++ b/master/buildbot/newsfragments/docker-container-name.bugfix
@@ -1,0 +1,1 @@
+Fixes issue with Buildbot DockerLatentWorker, where Buildbot can kill running workers by mistake based on the form the worker name (:issue:`3800`)
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -27,7 +27,8 @@ class Client(object):
         self.call_args_create_container = []
         self.call_args_create_host_config = []
         self.called_class_name = None
-        self._images = [{'RepoTags': ['busybox:latest', 'worker:latest', 'tester:latest']}]
+        self._images = [
+            {'RepoTags': ['busybox:latest', 'worker:latest', 'tester:latest']}]
         self._pullable = ['alpine:latest', 'tester:latest']
         self._pullCount = 0
         self._containers = {}
@@ -60,9 +61,19 @@ class Client(object):
 
     def containers(self, filters=None, *args, **kwargs):
         if filters is not None:
+            if 'existing' in filters.get('name', ''):
+                self.create_container(
+                    image='busybox:latest',
+                    name="buildbot-existing-87de7e"
+                )
+                self.create_container(
+                    image='busybox:latest',
+                    name="buildbot-existing-87de7ef"
+                )
+
             return [
                 c for c in self._containers.values()
-                if c['name'] == filters['name']
+                if c['name'].startswith(filters['name'])
             ]
         return self._containers.values()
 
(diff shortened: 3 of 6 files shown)


tardyp:
@dragon512 looks like the unitests need a fix

dragon512:
@tardyp is this test issue something I should fix? I know should add a news fragment.

tardyp:
the marathon tests are failing because marathon implementation is based on docker. Could you please update the assertion?
see details in https://nine.buildbot.net/#builders/51/builds/1309

then indeed news fragment will be mandatory

tardyp:
@dragon512 would you have time to work on it?

dragon512:
@tardyp I should... I been busy fixing some testing issues on ATS, I will fix this as soon as that is out of the way

dragon512:
@tardyp I have added everything needed for this PR. I also added an issue to link the bug fix with. Let me know if you need anything else for this PR.

tardyp:
Sorry for not thinking of this before, but I already fixed that bug in hyper:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/worker/hyper.py#L138

I think that solution is a bit cleaner.

tardyp:
I mean. changing the name to enhance the filter is good, but we should *also* compare the actual name of the resulting containers like I did for hyper

dragon512:
I see what you mean. Give me a moment to add that in

dragon512:
Ok added a check to test is and of the names in the "Names" list matched the container name

dragon512:
@tardyp Not sure what to do with the reduced codecov failure

master/buildbot/worker/docker.py
```diff
@@ -213,11 +213,15 @@ def _image_exists(self, client, name):
...
         docker_client = self._getDockerClient()
+        container_name = self.getContainerName()
         # cleanup the old instances
         instances = docker_client.containers(
             all=1,
-            filters=dict(name=self.getContainerName()))
+            filters=dict(name=container_name))
+        container_name = "/{0}".format(container_name)
         for instance in instances:
+            if container_name not in instance['Names']:
```
tardyp: The problem with codecov is that this line is not tested. Are you sure that ['Names'] is really in instance (I would have said it is ['Name']

dragon512: I did a dump and tested it locally. I was surprised that it is Names and it is a list. I did a local test to validate it works, by spinning up 20 or so builds, killing build bot and start it up again with some extra log.msg() to make sure it killed the correct containers. Not sure how we could fake this in the current test framework

dragon512: Here is a sample from the log file.

```2017-12-05 11:37:10-0600 [-] substantiating worker <LatentWorkerForBuilder builder=u'genBuilder' worker=u'docker-worker-9' state=DETACHED>
2017-12-05 11:37:10-0600 [-] ************* /buildbot-docker-worker-9-8c9f9f [u'/buildbot-docker-worker-9-8c9f9f']
2017-12-05 11:37:10-0600 [-] KILLL!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
2017-12-05 11:37:11-0600 [-] Starting factory _HTTP11ClientFactory(<function quiescentCallback at 0x7f4a500859b0>, <twisted.internet.endpoints._WrapperEndpoint object at 0x7f4a4bfe0b50>)```
oh and here is a stripped dump of the data structure we get back:
```
2017-12-05 11:24:23-0600 [-] {u'Command': u"/bin/bash --login -c 'buildbot-worker start --nodaemon'",
	 u'Created': 1512494662,
	 u'HostConfig': {u'NetworkMode': u'default'},
	 u'Id': u'36da48ba18d7c6283bdec7b32aa241379ab1a432ebaa2f56e24f2ed9fb23c4d5',
	 u'Image': u'ubuntu-17_10_bb:9ae0debffc28caf625958984f959ba1e',
	 u'ImageID': u'sha256:a8acab60e2e57240714aa3da03bad62b17e543490188ce696a686e36169f78f9',
	 u'Labels': {},
	 u'Mounts': [{u'Destination': u'...',
	              u'Mode': u'rw',
	              u'Propagation': u'rprivate',
	              u'RW': True,
	              u'Source': u'....',
	              u'Type': u'bind'}],
	 u'Names': [u'/buildbot-docker-worker-19-8c9f9f'],
	 u'NetworkSettings': {u'Networks': {u'bridge': {u'Aliases': None,
	                                                u'DriverOpts': None,
	                                                u'EndpointID': u'cf2ef9fa9dad943efa48e3818823701ea560bff4f7b74a0304c9f9fa0c209a37',
	                                                u'Gateway': u'...',
	                                                u'GlobalIPv6Address': u'...',
	                                                u'GlobalIPv6PrefixLen': 64,
	                                                u'IPAMConfig': None,
	                                                u'IPAddress': u'...',
	                                                u'IPPrefixLen': 16,
	                                                u'IPv6Gateway': u'...',
	                                                u'Links': None,
	                                                u'MacAddress': u'...',
	                                                u'NetworkID': u'ed8863131a5619a4d5de0552d46f5460edb3cfc8c82a3f46f96e8e89497e7507'}}},
	 u'Ports': [],
	 u'State': u'running',
	 u'Status': u'Up Less than a second'} 
```

Comment (tardyp) [reply master/buildbot/worker/docker.py]: You can look at the hyper test here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/test_worker_hyper.py#L135

I think you can easily adapt it on docker.
Here is the full commit, you will need to update the fake:
https://github.com/buildbot/buildbot/commit/29e564dfccda956b721992131b7167777aae2d0d

**Comment by dragon512** [comment]
@tardyp Looks like I fixes the testing issue, but there is some other issue with formatting in:
```
www/badges/setup.py:44:67: E261 at least two spaces before inline comment
            'VERSION', 'templates/*.svg.j2', 'static/.placeholder' # dist is required by buildbot_pkg
```

I did not change this code it seems to have been merged in outside my PR. I can fix it if you want as well.

**Comment by tardyp** [comment]
> I did not change this code it seems to have been merged in outside my PR. I can fix it if you want as well.
please yes.

**Comment by dragon512** [comment]
You might want to add a setup.cfg with 
```
[pep8]
max-line-length = 132
```
 and some script to make it easy to autopep (or what ever tool you want to use) all the source before checkin code

dragon512:
@tardyp Ok it is all ready for you :-)

tardyp:
thanks!

PR #3764, opened 2017-11-17
bitbucketserver.py: Fixes to interoperate with Bitbucket Cloud

willnewton:
The current implementation doesn't work with Bitbucket Cloud,
there's a number of exceptions thrown due to mismatched
JSON key names. Fix those and fix the commit web view URL.

Signed-off-by: Will Newton <will.newton@andromedadigital.com>

### DIFF
--- a/master/buildbot/www/hooks/bitbucketserver.py
+++ b/master/buildbot/www/hooks/bitbucketserver.py
@@ -73,8 +73,8 @@ class BitbucketServerEventHandler(object):
     def handle_repo_push(self, payload):
         changes = []
         project = payload['repository']['project']['name']
-        repo_url = payload['repository']['links']['self'][0]['href']
-        repo_url = repo_url.rstrip('browse')
+        repo_url = payload['repository']['links']['self']['href']
+        web_url = payload['repository']['links']['html']['href']
 
         for payload_change in payload['push']['changes']:
             if payload_change['new']:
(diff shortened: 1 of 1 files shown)


tardyp:
I believe there is no such thing as old bitbucket server version, right?
so no need to maintain backward compat?

willnewton:
@tardyp it's hard to say. There are two products "Bitbucket Cloud" which is the hosted version which I am using. Then there is "Bitbucket Server" which is the self-hosted version.

Bitbucket Cloud requires these changes to work and has no backward compatibility concerns.

Bitbucket Server does allow installing a range of versions. However from the changelogs it isn't clear when or if anything in this area changed.

The bitbucket.py change hook seems to be totally obsolete (passing request params rather than JSON in the body).

hir12111:
That file is meant to support Bitbucket Server hooks and IMHO, this change shouldn't break it (there are teams depending on it), if both API are very similar, it may be worth supporting both by catching a KeyError exception.
The downsides are: tests will be more complicated and the code will not adapt easily with API changes, so it may also be worth having this in a different file

willnewton:
@hir12111 are you aware of any documentation of these APIs? I have been following this document which seems to match what I am seeing in the wild:

https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html

willnewton:
It's possible we could copy this code over to bitbucket.py, which as I mentioned above seems to be obsolete:

https://confluence.atlassian.com/bitbucket/post-service-management-223216518.html

tardyp:
I can find the doc for bitbucketserver 5.5, which references 'scmId' instead of 'scm':
https://confluence.atlassian.com/bitbucketserver/post-service-webhook-for-bitbucket-server-776640367.html

I would be in favor of having another hook called bitbucketcloud, with the modification that you passed.

bitbucket cloud and bitbucket server are two different products that are working on trying to converge, but not achieving it. Until that, imho we should have differentiated support for them as if they were two products.

willnewton:
@tardyp  I'll close this PR in favour of https://github.com/buildbot/buildbot/pull/3775 which implements the suggested approach.

Issue #3701, opened 2017-10-17
Update existing change feature


rjarry:
Sometimes, it is convenient to re-send a change to the server to "rebuild" a set of builds. The reason behind that can be:

* There was an infrastructure problem and we need to run everything again.
* We need to rebuild the same revision but with different change properties

For now, when doing that, a new change is created but linked to the same sourcestampid than the old change. And builds are re-scheduled but linked to the old change (so new properties are ignored).

It would be better to update the existing change with the new properties and re-trigger schedulers with the same changes/sourcestamp.

Hyask:
Just for the record, after some discussions with @rjarry.

A hack can be done to fix the problem there:
https://github.com/buildbot/buildbot/blob/02d8fe20aee02788281042733a24b4d11486ab24/master/buildbot/db/changes.py#L222 
Just append `.order_by(changes_tbl.c.changeid.desc())` to the query, so that the last change corresponding the to the asked sourcestampid is fetched, rather than the first.

This works, at least in my personal simple case, but will likely never get upstream, since this is just a hack, and a better solution should be found.

PR #3786, opened 2017-11-30
db: always return the most recent change from ssid

rjarry:
When submitting multiple times the same revision via change hooks, a new change is created each time but linked to the same sourcestampid.

These changes may carry different properties but when schedulers are notified of a new change, they access to the change(s) via sourcestampid via `self.master.data.get(('sourcestamp', ssid, 'changes'))` which only returns the first one.

Change this and return the most recent one instead. This will not affect nominal cases where there only is one change by ssid.

Fixes issue #3701.

### DIFF
--- a/master/buildbot/db/changes.py
+++ b/master/buildbot/db/changes.py
@@ -221,6 +221,9 @@ class ChangesConnectorComponent(base.DBConnectorComponent):
             changes_tbl = self.db.model.changes
             q = changes_tbl.select(
                 whereclause=(changes_tbl.c.sourcestampid == sourcestampid))
+            # if there are multiple changes for this ssid, get the most recent one
+            q = q.order_by(changes_tbl.c.changeid.desc())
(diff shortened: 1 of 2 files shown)


master/buildbot/db/changes.py
```diff
@@ -221,6 +221,8 @@ def thd(conn):
             changes_tbl = self.db.model.changes
             q = changes_tbl.select(
                 whereclause=(changes_tbl.c.sourcestampid == sourcestampid))
+            # if there are multiple changes for this ssid, get the most recent one
+            q = q.order_by(changes_tbl.c.changeid.desc())
```
tardyp: I think if we are going to fetch only one, then we should also do the ```.limit(1)```

rjarry: Done. Keep in mind that this is will only occur when rebuilding the same revision.

rjarry:
yay, thanks :)

PR #3792, opened 2017-12-01
Reporter and WebHooks doc tweaks

WayneKeenan:
Changes to docs:

- GitHubCommentPush max paylod size note.
- Example code clarifications.
- Reverse proxy note.

### DIFF
--- a/master/docs/manual/cfg-wwwhooks.rst
+++ b/master/docs/manual/cfg-wwwhooks.rst
@@ -89,7 +89,7 @@ GitHub hook
 .. note::
 
    There is a standalone HTTP server available for receiving GitHub notifications as well: :contrib-src:`master/contrib/github_buildbot.py`.
-   This script may be useful in cases where you cannot expose the WebStatus for public consumption.
+   This script may be useful in cases where you cannot expose the WebStatus for public consumption. Alternatively, you can setup a reverse proxy :ref:`Reverse_Proxy_Config` 
 
 The GitHub hook has the following parameters:
 
@@ -135,7 +135,8 @@ The simplest way to use GitHub hook is as follows:
 
 .. code-block:: python
 
-    c['www'] = dict(...,
+    c['www'] = dict(    
+        'base': True
         change_hook_dialects={'github': {}},
     )
 
@@ -161,7 +162,6 @@ The parameters are:
                 change_hook_dialects={
                     'github': {
                         'secret': 'MY-SECRET',
-                        'strict': True,
                     },
                 },
             )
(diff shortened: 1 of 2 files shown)


master/docs/manual/cfg-wwwhooks.rst
```diff
@@ -89,7 +89,7 @@ GitHub hook
 .. note::
 
    There is a standalone HTTP server available for receiving GitHub notifications as well: :contrib-src:`master/contrib/github_buildbot.py`.
-   This script may be useful in cases where you cannot expose the WebStatus for public consumption.
+   This script may be useful in cases where you cannot expose the WebStatus for public consumption. Alternatively, you can setup a reverse proxy :ref:`Reverse_Proxy_Config`
```
tardyp: well the Reverse_Proxy_Config exposes the whole web status to public.

You need a different config than the one that is described there.

master/docs/manual/cfg-wwwhooks.rst
```diff
@@ -135,7 +135,8 @@ The simplest way to use GitHub hook is as follows:
 
 .. code-block:: python
 
-    c['www'] = dict(...,
+    c['www'] = dict(    
+        'base': True
```
tardyp: why? no need to have base!

WayneKeenan: it didn't work without it.

master/docs/manual/cfg-wwwhooks.rst
WayneKeenan: so, either it's a documented security warning? or an example config?

tardyp: I think you can add an example here of an nginx config which implements the redirection just for the hook.

master/docs/manual/cfg-wwwhooks.rst
tardyp: what didn't work? What is the symptom?

c['www']['base'] is not an available config
c['www']['change_hook_dialects']['base'] is, but only enables  the base hook. 

The github hook should not be dependent on the base hook. I double checked on the code that there is no code smell in that direction. Please double check on your side that this is not a confusion.

WayneKeenan: my confusion, I reverted back to PyPi release and it's ok. I must have tested (what I thought was) the endpoint using http://cam1.local:8010/change_hook/

tardyp: can you fix your patch then?

Issue #3796, opened 2017-12-02
Error creating Windows Service, encoding issue calling Win32 Service API


WayneKeenan:
buildbot: latest from github
Python: 3.6.3
Windows: 10

The error when trying to start the service is :
```python
Error starting service: The service did not respond to the start or control request in a timely fashion.
Debugging service BuildBot - press Ctrl+C to stop.
****************** b'...\\worker_dir'
Error 0xC0000003 - The instance's SvcRun() method failed

Traceback (most recent call last):
  File "...\env\lib\site-packages\win32\lib\win32serviceutil.py", line 835, in SvcRun
    self.SvcDoRun()
  File "..\env\lib\site-packages\buildbot_worker-2017.12.2-py3.6.egg\buildbot_worker\scripts\windows_service.py", line 226, in SvcDoRun
    if not self._checkConfig():
  File "..\env\lib\site-packages\buildbot_worker-2017.12.2-py3.6.egg\buildbot_worker\scripts\windows_service.py", line 211, in _checkConfig
    dir_string)
  File "..\env\lib\site-packages\win32\lib\win32serviceutil.py", line 272, in SetServiceCustomOption
    win32api.RegSetValueEx(key, option, 0, win32con.REG_SZ, value);
TypeError: Objects of type 'bytes' can not be converted to Unicode.
```

A quick and dirty monkey patch to get past the error:

Change :
```python
dir_string = os.pathsep.join(self.dirs).encode("mbcs")
```
To
```python
dir_string = os.pathsep.join(self.dirs)
```

rodrigc:
@WayneKeenan actually your change is correct.  Values supplied to `win32api.RegSetValueEx` must be unicode not bytes.  Is that the only fix you needed to get things working?

Can you submit a pull request with this change?

WayneKeenan:
That was the only monkey patch, but I could only get the worker service to start in `debug` mode, and even then there was another exception exiting `debug` mode. (less critical)

So I abounded it for a dirty workaround: 
```bash
sc create buildbot binPath= "%BUILDBOT_WORKER_EXE% start %WORKER_DIR%" error= ignore obj= %YOUR_DOMAIN_OR_DOT%\%USERNAME% password= %PASSWORD start= auto depend= tcpip DisplayName= "Buildbot Worker"
sc start buildbot
```

But with that method windows says the service doest start, because it not correctly responding to the Windows management signals, however, the worker is running.  hacky I know.

[...]

PR #3797, opened 2017-12-03
Fix incorrect encoding for #3796

WayneKeenan:
Values supplied to win32api.RegSetValueEx must be unicode not bytes.

### DIFF
--- a/master/buildbot/scripts/windows_service.py
+++ b/master/buildbot/scripts/windows_service.py
@@ -205,7 +205,7 @@ class BBService(win32serviceutil.ServiceFramework):
                        "Stopping the service.")
             return False
         if save_dirs:
-            dir_string = os.pathsep.join(self.dirs).encode("mbcs")
+            dir_string = os.pathsep.join(self.dirs)
             win32serviceutil.SetServiceCustomOption(self, "directories",
                                                     dir_string)
         return True


tardyp:
@rodrigc any opinion?

pmisik:
Hi, there is yet another place where is used this construction
https://github.com/buildbot/buildbot/blob/master/worker/buildbot_worker/scripts/windows_service.py#L207

I do not remember seeing a similar track on W10 or W2k16.
Probably due to we followed http://trac.buildbot.net/wiki/RunningBuildbotOnWindows.
And if I remember correctly there was step to explicitly setup registry keys Directories.

We don't use cygwin and we install it by 
`.\Scripts\buildbot_windows_service.exe  --user {bbuser} --password {password} --startup auto install`

or in case of worker just

`.\Scripts\buildbot_worker_windows_service.exe --user {BBUser} --password {BBpassword} --startup auto install`

pmisik:
Hi, there is yet another place where is used this construction
https://github.com/buildbot/buildbot/blob/master/worker/buildbot_worker/scripts/windows_service.py#L207

I do not remember seeing a similar track on W10 or W2k16.
Probably due to we followed http://trac.buildbot.net/wiki/RunningBuildbotOnWindows.
And if I remember correctly there was step to explicitly setup registry keys Directories.

We don't use cygwin and we install it by 
`.\Scripts\buildbot_windows_service.exe  --user {bbuser} --password {password} --startup auto install`

or in case of worker just

`.\Scripts\buildbot_worker_windows_service.exe --user {BBUser} --password {BBpassword} --startup auto install`

Issue #3804, opened 2017-12-05
coffeescript -> javascript ES6 (ES2015) ?


simevo:
Coffeescript complicates the build, adds dependencies and makes contributing harder.

Coffeescript does not appear in the [top 100 programming languages tracked by the TIOBE index](https://www.tiobe.com/tiobe-index/), nor in the 23 tracked by [PYPL](https://pypl.github.io/PYPL.html) and not in the 15 tracked by [github](https://octoverse.github.com/).

Other projects are migrating from CS to JS ES6, for example:
https://medium.com/adstage-engineering/our-journey-from-coffeescript-to-es6-af297b0990ac

According to:
```
find www | grep 'coffee$' | xargs wc | tail -n 1
```
buildbot contains about 10 kLOC of coffeescript (that's conservative as it also counts empty lines and comments...).

tardyp:
Hi @simevo ,
I think coffeescript was a good choice 4 years ago (we started in 2013), but it is certainly not anymore, indeed.

As the blog post title suggests, migrating out of coffeescript will be a Journey.

I am all for that if people want to start that journey, and propose a proper plan. e.g.

- [ ] switch from guanlecoja build system to webpack
- [ ] re-write the plugins system in webpack (being compatible with the old guanlecoja based looks useful)
- [ ] convert the coffee code to es6 (with decaffeinate)
   - [ ] will probably mean to modify decaffeinate to understand ng-classify.

simevo:
javascript is a "vibrant ecosystem" which also means that is risky: a new and fashionable dialect / tool may be abandoned in 3 years or less

buildbot is a server component and should be much stabler than that; this means that you should be extra careful when choosing new dependencies / tools

having said that, for the sake of packageability (for debian/ubuntu, see: https://bugs.debian.org/790241):
- getting rid of guanlecoja would be great
- webpack is fine
- bower is also deprecated but yarn is not yet in debian so it would be best to wait a bit on that

it'd be interesting to hear as many comments as possible from the other team members on this issue !

dragon512:
I tend to find that typescript and web components ( like https://www.polymer-project.org/) have been good models for me in the Web space dev area. 
Typescript gives us some type safety and feature of what is next in Javascript es6 and es7 land today.
Web component allows us to make modular/component items that can be easily added to existing page. Communication happens in a standard way. 

I have found this gives me the best of what is new but in a safe flexible way

tardyp:
@simevo

[...]

PR #3806, opened 2017-12-05
WIP: Decaffeinate the data-module

tardyp:
following up on #3804 , here is a first try at decaffeinate.

I first ran ng-classify to remove the ng-classify annotations, and then ran decaffeinate.
I then took https://github.com/preboot/angularjs-webpack as a reference to build a webpack library

I wasn't able to run karma tests yet, so I don't know if the results are equivalent.

The resulting code is 101k compared to the 41k of coffeescript, but I am not sure if I use the correct webpack strategy, as I never really dived into webpack.


uglycoyote:
I appreciate you looking in to this so I thought I would take a stab to see if I could help a bit.  On another project I have used Webpack with Mocha tests (but not Karma).   But the setup here is more ornate than what i had with Mocha, I guess because Karma has its own .conf.js and it seems to want to run webpack underneath it.  (whereas with Mocha I think I was just running `webpack` to make my bundles -- one of which was the application code and the other a bundle of tests -- and then running a separate mocha command afterwards on the test bundle)

How far did you get towards running the karma tests?   It seems like what you have here is close to working.

When I cloned your repo and switched to your branch I ran `npm test` in the `www\data_module` directory.  The first few times I tried this it kept quitting out early because it was unabled to start or communicate with PhantomJS (not sure why).   For comparison I cloned the `angularjs-webpack` repo and did `npm test` in there and had the same issue, so whatever was going on it wasn't a problem with how you set it up.

When I did an `npm start` in `angularjs-webpack`, and then later did a `npm test` again, PhantomJS started working for me and I got test results in `angularjs-webpack`  (I tweaked their app.spec.js to make a failure and then a success again).  After that I went back to `buildbot\www\data_module' again and PhantomJS was working there. 

Then I was getting the following in the `npm test` output:

```
05 12 2017 16:44:07.136:WARN [reporter]: SourceMap position not found for trace: undefined 
05 12 2017 16:44:07.137:WARN [reporter]: SourceMap position not found for trace: undefined 
PhantomJS 2.1.1 (Windows 7 0.0.0) ERROR                                                    
  TypeError: Cannot call a class as a function             
  at src/tests.webpack.js:883                              
```

I wanted to debug which test sources were being used, so in tests.webpack.js I added

	context.keys().forEach( key => console.log("context in tests.webpack.js:", key) );

and then when running I could see the log of all of the js files that it was bringing in.

Despite the comment in karma.conf.js which says:

      // Grab all files in the app folder that contain .spec.

, I didn't see anything that was restricting it to just the .spec.js files so I tried changeding tests.webpack.js to say

	const context = require.context('./', true, /\.spec.js$/);

After making that change `npm test` had output showing that it's trying to run each of the tests in the various `.spec` files in the `data_module`.  There are a whole bunch of separate blocks like this example from the `dataquery.service.spec.js`:

```
PhantomJS 2.1.1 (Windows 7 0.0.0): Executed 3 of 17 (3 FAILED) (0 secs / 0.003 secs)
05 12 2017 16:52:01.221:WARN [reporter]: SourceMap position not found for trace: undefined
05 12 2017 16:52:01.221:WARN [reporter]: SourceMap position not found for trace: undefined
05 12 2017 16:52:01.222:WARN [reporter]: SourceMap position not found for trace: undefined
PhantomJS 2.1.1 (Windows 7 0.0.0) dataquery service encountered a declaration exception FAILED
        TypeError: Object is not a constructor (evaluating 'module('bbData')') in src/tests.webpack.js (line 40639)
        webpack:///src/services/data/collection/dataquery.service.spec.js:8:22 <- src/tests.webpack.js:40639:22
        webpack:///src/services/data/collection/dataquery.service.spec.js:6:9 <- src/tests.webpack.js:40635:9
        src/tests.webpack.js:40835:34
        __webpack_require__@webpack:/webpack/bootstrap%2070b3b97624f8e650bd02:19:0 <- src/tests.webpack.js:20:34
        webpackContext@webpack:/src%20/.spec.js$:13:0 <- src/tests.webpack.js:3591:28
        forEach@[native code]
        webpack:///src/tests.webpack.js:13:23 <- src/tests.webpack.js:41690:47
        __webpack_require__@webpack:/webpack/bootstrap%2070b3b97624f8e650bd02:19:0 <- src/tests.webpack.js:20:34
        webpack:/webpack/bootstrap%2070b3b97624f8e650bd02:65:0 <- src/tests.webpack.js:66:37
        global code@src/tests.webpack.js:67:12
```

followed by 

```
PhantomJS 2.1.1 (Windows 7 0.0.0): Executed 17 of 17 (17 FAILED) ERROR (0.128 secs / 0.039 secs)

=============================== Coverage summary ===============================
Statements   : 100% ( 7/7 )
Branches     : 100% ( 0/0 )
Functions    : 100% ( 2/2 )
Lines        : 100% ( 6/6 )
================================================================================
```

So... it is trying.   I suspect that the original 

	const context = require.context('./', true, /\.js$/);

is the correct thing to do but there's something somewhere making the `TypeError: Cannot call a class as a function` and it's hard to tell where because sourcemaps don't seem to work right in that situation.

uglycoyote:
A few other things i discovered after fiddling a bit more:

You can get better debugging using Chrome instead of PhantomJS
`npm install --save-dev karma-chrome-launcher`, then change to `browsers: ['Chrome']` in `karma.conf.js`

also switch to `singleRun: false,` otherwise the browser will vanish and the test process will finish.

The `TypeError: Cannot call a class as a function` mentioned earlier is because of class constructors being called without new, e.g. in `data.constant.js` it needs to be `new Api()` and `new Endpoints()` 

Errors saying `TypeError: module is not a function` are due to lines like 

`beforeEach(module('bbData'));` 

in the `.spec.js` files.  This needs to be changed to 

`beforeEach(angular.module('bbData'));` or `beforeEach(angular.mock.module('bbData'));`,

 I'm not sure which. (either way I was getting `Error: [$injector:nomod] Module 'bbData' is not available!`)

tardyp:
Note that there is also this work in progress:
https://github.com/tardyp/buildbot-data

I am not sure yet which approach is the best :-/

PR #3814, opened 2017-12-08
config: check for duplicate service names

### LINKED ISSUE #3813: Two mail.MailNotifier instances with the same auto-generated name
Thank you guys, BuildBot is great!

I have found something like a bug.
I have **two** mail.MailNotifiers in my master.cfg with only **two differences**:
- a list of **recipients**
- a boolean parameter **buildSetSummary**

but BuildBot generates the same name for both:

`buildbot/util/service.py:64: self.name = MailNotifier_builders_Check REST [debug]+Check REST [debug]/1 Check clients/test-otc-1658failing_passing_warnings
`

I have fixed it setting theirs names explicitly.
But the behavior is **misleading**.

Many thanks for the people from freenode/#buildbot

rjarry:
Most services have a default `name` attribute. When instantiating the same service class multiple times (e.g. multiple email notifiers), if the user does not override the name, only the last service is actually
configured.

Make sure that all services have a unique name.

Fixes #3813 

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -848,6 +848,10 @@ class MasterConfig(util.ComparableMixin, WorkerAPICompatMixin):
 
                 continue
 
+            if _service.name in self.services:
+                error('Duplicate service name %r' % _service.name)
+                continue
+
             self.services[_service.name] = _service
 
     def check_single_master(self):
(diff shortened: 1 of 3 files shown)


rjarry:
The flake errors seem unrelated to my changes. Is this PR good for merge?

PR #3815, opened 2017-12-08
db: update schema to allow easier cleanup

rjarry:
Some buildbot installs handle a lot of builders (more than 500) and therefore a lot of builds. This can cause the database to grow over time and take too much disk space (also, this can increase complex SQL queries response time).

`logchunks` cleanup is already available via the `JanitorConfigurator`. But there is no easy way to cleanup old `sourcestamps`, `changes`, `buildsets`, `buildrequests`, `builds`, `buildsteps` and `logs` since the database constraints will require that the cleanup procedure starts from the leafs. This can be very complex and time consuming.

In order to allow deleting any database record along with any other records that may link to it, modify the foreign key constraints to "cascade" delete records when linked records are deleted.

### DIFF
--- a/master/buildbot/db/migrate/versions/050_cascading_deletes_all.py
+++ b/master/buildbot/db/migrate/versions/050_cascading_deletes_all.py
@@ -1,0 +1,90 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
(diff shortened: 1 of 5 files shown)


rjarry:
The migration procedure works with prostgres, but it does not with sqlite. There may be something that I missed...

tardyp:
You will need to add a test for the migration script so that we verify it works on every db system.
Please look at examples in the test/unit directory..

rjarry:
The flake errors seem unrelated to my changes. Can we merge this?

PR #3831, opened 2017-12-15
Add host_conf value "init"

dragon512:
This prevents zombie and abandoned process issues in worker while it runs

## Contributor Checklist:


dragon512:
@tardyp We don't really need any tests here I think. This is about adding the smart init process given the version of python docker supports adding this. We need python docker package 2.2.0 or better. The failures from before are because the CI system is running the old docker-py 1.x version.

tardyp:
yes, need_tests was for finxing the tests.

Thanks!

PR #3827, opened 2017-12-14
Introduce a common GithubApiService

s0undt3ch:
## Contributor Checklist:

* [x] I have updated the unit tests
* [x] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [ ] I have updated the appropriate documentation

PR #3835, opened 2017-12-17
Passed headers overwrite base headers

s0undt3ch:
Any header that is defined at service instantiation should be overridden by any header of the same explicitly passed when making a request.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_util_httpclientservice.py
+++ b/master/buildbot/test/unit/test_util_httpclientservice.py
@@ -121,6 +121,18 @@ class HTTPClientServiceTestTxRequest(HTTPClientServiceTestBase):
                                                                 'Content-Type': 'application/json'
                                                             })
 
+    def test_passed_headers_overwrite_base_headers(self):
+        self.base_headers.update({'X-TOKEN': 'XXXYYY'})
+        self._http.post('/bar', json={'foo': 'bar'}, headers={'X-TOKEN': 'AABBCC'})
+        jsonStr = json.dumps(dict(foo='bar'))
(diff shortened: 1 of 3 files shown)


s0undt3ch:
Do we have integration tests for the HTTP client service?
I need an actual response, even if from a local server to test the  headers attribute access for.both backends.

master/buildbot/util/httpclientservice.py
```diff
@@ -63,6 +65,17 @@ def json(self):
     def code(self):
         return self._res.status_code
 
+    @property
+    def headers(self):
+        if self._headers is None:
```
tardyp: according to codecov, this method is not tested. I think it is not actually needed/used

s0undt3ch: It will be needed for my other PR. Decided to include I here because its also about headers. Can pull it out for now if you prefer.

s0undt3ch:
Removed that headers code. Will add it back when needed.

master/buildbot/test/unit/test_util_httpclientservice.py
```diff
@@ -121,6 +121,18 @@ def test_post_auth(self):
                                                                 'Content-Type': 'application/json'
                                                             })
 
+    def test_passed_headers_overwrite_base_headers(self):
```
tardyp: I reviewed that chunk in the other PR. we need tests for the two other backends. (and documentation two document that new feature)

s0undt3ch:
Closing this PR. This feature will be included in https://github.com/buildbot/buildbot/pull/3827 where its needed.

PR #3855, opened 2017-12-27
Fix #3032 - Setting timeout=None to the CopyDirectory step has no effect

### LINKED ISSUE #3032: Setting timeout=None to the CopyDirectory step has no effect.
This ticket is a migrated Trac ticket [3669](http://trac.buildbot.net/ticket/3669)

People contributed to the original ticket: `benoit.allard@...`
Ticket created on: `Feb 28 2017`
Ticket last modified on: `Feb 28 2017`

---

According to the documentation, this should disable the timeout altogether.

Unfortunately, the timeout value is given to the worker only if the value evaluates to 'True-ish'. The worker on its side, when not given a value uses 120sec. Making the timeout=None useless.

---

_Comment from_: @Ben
_Date_: `Feb 28 2017`

The workaround is to use a bigger value, like 1200.

jhonatandarosa:
This PR fix the issue #3032 - Setting timeout=None to the CopyDirectory step has no effect.

### DIFF
--- a/worker/buildbot_worker/commands/fs.py
+++ b/worker/buildbot_worker/commands/fs.py
@@ -161,7 +161,7 @@ class CopyDirectory(base.Command):
         fromdir = os.path.join(self.builder.basedir, self.args['fromdir'])
         todir = os.path.join(self.builder.basedir, self.args['todir'])
 
-        self.timeout = args.get('timeout', 120)
+        self.timeout = args.get('timeout', None)
         self.maxTime = args.get('maxTime', None)
 
         if runtime.platformType != "posix":


worker/buildbot_worker/commands/fs.py
```diff
@@ -161,7 +161,7 @@ def start(self):
         fromdir = os.path.join(self.builder.basedir, self.args['fromdir'])
         todir = os.path.join(self.builder.basedir, self.args['todir'])
 
-        self.timeout = args.get('timeout', 120)
+        self.timeout = args.get('timeout', None)
```
tardyp: I don't think this is the right fix.
```
>>> args = {'timeout': None}
>>> print(args.get('timeout', 120))
None
```
args.get should be returning None is args['timeout'] is None.

aren't we rather removing None argument before?

PR #3856, opened 2017-12-27
model change for worker state

### LINKED ISSUE #3429: How to graceful shutdown one/all workers?
> In previous version, like 0.8.7, there is way to graceful shutdown one workers through web UI, but in v0.9.8, I didn't find the button. Is it not implemented or removed?

The core functionality is here but not yet mapped to data API / UI.
A similar functionality is the stop build/cancel buildrequest #1656

- It is about adding 2 actions to the worker endpoint.
- In order to make multi-master work, those actions should send a message to the mq
- Then a button should be added to the UI

Work is ongoing. Remaining tasks:

- [x] DB model and API for status of worker (add paused status, shutdown status, connected status, while we are at it to avoid check for the master worker table)
- [x] Data API for status of worker (map status db api, send mq updates about them)
- [x] UI for reflecting worker status

(edited by @tardyp)

tardyp:
Now the pause and graceful states are stored in DB, and thus available in the UI.

Fixes #3429

### DIFF
--- a/master/buildbot/data/resultspec.py
+++ b/master/buildbot/data/resultspec.py
@@ -440,3 +440,9 @@ class ResultSpec(object):
             rv.offset, rv.total = offset, total
             rv.limit = limit
             return rv
+
+# a resultSpec which does not implement filtering in python (for tests)
+class OptimisedResultSpec(ResultSpec):
+
+    def apply(self, data):
+        return data
--- a/master/buildbot/data/resultspec.py
+++ b/master/buildbot/data/resultspec.py
@@ -441,6 +441,7 @@ class ResultSpec(object):
             rv.limit = limit
             return rv
 
+
 # a resultSpec which does not implement filtering in python (for tests)
 class OptimisedResultSpec(ResultSpec):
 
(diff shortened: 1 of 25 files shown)


tardyp:
> Would that make sense to add a "quarantined" state?

Indeed, I think this would be useful, there is also the need to store the info on when it is supposed to go out of quarantine.

> Instead of multiple columns, could you use a single one like for Build results? This may ease adding new states. Unless multiple states can coexist.

indeed multiple states can coexists, I also wanted it to be easily searchable.

> I don't see any use case for this in the docs. Did I miss something? forget that, I saw the related issue #3429.

I think indeed, more explaination in the release note can be useful

> Finally, could you squash & reorganize commits and write proper commit messages? I miss a high level description of why this feature is needed. directly in git logs. GitHub issues may not be here forever.

I tried at least in the beginning of this dev to keep commits incremental, as a way to show what is the process to develop such changes (keep small changes with unit tests passing).
I think the commit went a bit more eratic in the end, I will try to clean them up.

tardyp:
@rjarry I think I've addressed your points. Sober enough to have a look?

rjarry:
Yes it looks good.

I did not see a way to see if a worker is paused (manually) or quarantined. No biggie, it can be added later.

PR #3857, opened 2017-12-29
vstudio: Support lists in env values

krf:
## Remove this paragraph

And especially:

## Contributor Checklist:


krf:
I can't reproduce the issue, for this I'd need to revert the patch on the production system.

We use lists in environment variables basically everywhere, this is why this problem showed up. As far as I know the problem was that the vstudio step attempted to amend a env var (a list) as populated by `BuilderConfig(env=...)`.

PR #3858, opened 2017-12-29
Upgrade to Guanlecoja-ui 1.8.0; Waterfall body height fix

uglycoyote:
@tardyp, here are the changes which upgrade to `guanlecoja-ui` version 1.8.0 on buildbot, and also keep those changes from causing the waterfall view to be broken.

With regards to the waterfall view, I took a different approach then previously discussed.  The `.waterfall { height : 100vh }` trick did not work perfectly.   When testing the waterfall view with this 100vh rule in place, I noticed that when you clicked a build and brought up the dialog, everything in the background would disappear.   I think what was happening is that the angular-bootstrap modal has some javascript which dynamically adds the class `modal-open` onto the `body`, and this class sets `overflow:hidden`.   I suppose setting the `100vh` on the waterfall div was not causing the body to grow, and therefore most of the contents of the div were considered "overflow".

So instead of using `.waterfall{height:100vh}` I went back to the idea of setting waterfall-specific rules on the body.   Originally I was thinking that was a dirty thing to do, but upon discovering the way that angular-bootstrap dynamically adds/removes the`modal-open` class to `body`, I loosened my judgement of the morality of doing the same thing myself.

I accomplished this by having the waterfall controller's constructor add the class `waterfall` to the body.  It also needs to remove this when you leave the waterfall page so I did this on the waterfall's scope`s $destroy event.  The `body.waterfall` rule sets the `height:100%` which makes the waterfall work correctly but was causing the scroll-jump problem in Console.    Interestingly, even with `height:100%`, the waterfall does not exhibit the problem.

### DIFF
--- a/www/waterfall_view/src/module/main.module.coffee
+++ b/www/waterfall_view/src/module/main.module.coffee
@@ -31,9 +31,9 @@ class Waterfall extends Controller
         #  class needs to removed upon exiting the waterfall via the $destroy
         #  event below.)
         body = @$rootElement.find("body")
-        body.addClass("waterfall")
+        body.addClass("hundredpercent")
         @$scope.$on("$destroy", ()=>
-            body.removeClass("waterfall")
+            body.removeClass("hundredpercent")
         );
 
         glTopbarContextualActionsService.setContextualActions(actions)
--- a/www/waterfall_view/src/module/main.module.coffee
+++ b/www/waterfall_view/src/module/main.module.coffee
@@ -10,7 +10,7 @@ class WaterfallView extends App
 
 class Waterfall extends Controller
     self = null
-    constructor: (@$scope, $q, $timeout, @$window, @$log,
+    constructor: (@$rootElement, @$scope, $q, $timeout, @$window, @$log,
                   @$uibModal, dataService, d3Service, @dataProcessorService,
                   scaleService, @bbSettingsService, glTopbarContextualActionsService) ->
         self = this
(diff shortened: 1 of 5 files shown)


www/waterfall_view/src/module/main.module.coffee
```diff
@@ -23,6 +23,19 @@ class Waterfall extends Controller
...
         ]
+
+        # 'waterfall' class needs to be dynamically added to the body in order 
+        #  to support waterfall-specific styling of the body.  (this is a bit
+        #  awkward since the body is provided by guanlecoja-ui and is the same
+        #  element as you switch between different plugin pages, therefore the
+        #  class needs to removed upon exiting the waterfall via the $destroy
+        #  event below.)
+        body = @$rootElement.find("body")
+        body.addClass("waterfall")
```
tardyp: I would rather call that class hundredpercent (so that other dashboard can use the same trick.)

uglycoyote: Sounds good.  I'll make that change.

uglycoyote: I'll move it the hundredpercent style into somewhere more global as well I suppose.  `/buildbot/www/base/src/styles/styles.less` is what I was thinking.

uglycoyote: @tardyp, good thinking: You were right to be suspicious of the css rule name `body.waterfall` since there was already a rule for `.waterfall` below that, so my `body.waterfall` was inadvertently picking up the rules that were meant for the the `div.waterfall`, and in fact that was what was making it work.  Once I renamed it to `hundredpercent` it stopped working again -- I realized setting `body { height : 100% }` does nothing without also setting `html { height : 100% }` which was the other rule that I removed the `guanlecoja-ui`.  

However, making the new `.hundredpercent` rule use { height : 100vh } seems to work well.  When I set `100vh` on the body it does not cause the "disappearing upon opening dialog" problem that happened when setting `100vh` on the `div.waterfall`

On a side note, whenever I try to add my new changes with `git add -A .`, I end up accidentally adding a bunch of package-lock.json files that apparently aren't already in the repository.  Should we add that to .gitignore or should we just start adding those files to the repo?  Internet says you should add them to the repo, but I don't know much about it or have an opinion about it.

tardyp: we do commit the yarn.lock from time to time, but imho this is just noise.

uglycoyote:
Merci beaucoup, Pierre!  Et, bonne année!

tardyp:
Merci a toi pour ta contribution!

PR #3860, opened 2017-12-30
Fix version when installing from git archive

rjarry:
When installing from a git archive downloaded from github, the file `master/buildbot/VERSION` is not present since it is generated when building source archive with:

```
./setup.py sdist
```

We also do not have access to git.

To work around the problem, use the export-subst git attribute to make git archive write the current revison information in various files and parse this to try and detect a version.

> From: https://git-scm.com/docs/gitattributes
> 
> `export-subst`
> 
> If the attribute `export-subst` is set for a file then Git will expand several placeholders when adding this file to an archive. The expansion depends on the availability of a commit ID, i.e., if [git-archive[1]](https://git-scm.com/docs/git-archive) has been given a tree instead of a commit or a tag then no replacement will be done. The placeholders are the same as those for the option `--pretty=format:` of [git-log[1]](https://git-scm.com/docs/git-log), except that they need to be wrapped like this: `$Format:PLACEHOLDERS$` in the file. E.g. the string `$Format:%H$` will be replaced by the commit hash.

If the archived revision is tagged, the tag will be used as version otherwise `99.dev.$commitid` will be used. If the code is not issued from a `git archive`, the version will be extracted from git describe as it was already done before.

### DIFF
--- a/master/buildbot/__init__.py
+++ b/master/buildbot/__init__.py
@@ -61,6 +61,34 @@ def mTimeVersion(init_file):
     return d.strftime("%Y.%m.%d")
 
 
+def getVersionFromArchiveId(git_archive_id='$Format:%ct %d$'):
+    """ Extract the tag if a source is from git archive.
+
+        When source is exported via `git archive`, the git_archive_id init value is modified
+        and placeholders are expanded to the "archived" revision:
+
+            %ct: committer date, UNIX timestamp
+            %d: ref names, like the --decorate option of git-log
+
+        See man gitattributes(5) and git-log(1) (PRETTY FORMATS) for more details.
+    """
+    # mangle the magic string to make sure it is not replaced by git archive
+    if not git_archive_id.startswith('$For''mat:'):
+        # source was modified by git archive, try to parse the version from
+        # the value of git_archive_id
+
+        match = re.search(r'tag:\s*v([^,)]+)', git_archive_id)
+        if match:
+            # archived revision is tagged, use the tag
+            return gitDescribeToPep440(match.group(1))
+
+        # archived revision is not tagged, use the commit date
+        tstamp = git_archive_id.strip().split()[0]
+        d = datetime.datetime.fromtimestamp(int(tstamp))
+        return d.strftime('%Y.%m.%d')
+    return None
+
+
 def getVersion(init_file):
     """
     Return BUILDBOT_VERSION environment variable, content of VERSION file, git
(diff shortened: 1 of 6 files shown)


rjarry:
Debian packages require to have a source archive (ideally generated from git). Buildbot needs to be able to determine its own version when outside of its own git repo and without a distutils generated archive.

master/buildbot/__init__.py
```diff
@@ -80,6 +80,16 @@ def getVersion(init_file):
     except IOError:
         pass
 
+    git_archive_id = '$Format:%ct %d$'
+    if not re.search(r'^\$Format:', git_archive_id):
```
tardyp: I don't see how this test could fail. Are you sure you want to set git_archive_id just before?

tardyp: ahh. I downloaded the archive, and saw that git archive will actually replace that code!
We surely need to explain that with a comment in the code.

we can also rather say:
```
if not git_archive_id.startswith("$Format:"):
```

rjarry: This special `$Format:PLACEHOLDERS$` string is replaced by the expanded `PLACEHOLDERS` when building the archive (see https://git-scm.com/docs/gitattributes and https://git-scm.com/docs/pretty-formats).

If we are executing this code from the git repo, the export-subst thingy will not take place and `git_archive_id` **will** begin with `$Format:`. I only want to look into the contents of `git_archive_id` if the expansion took place.

rjarry: I can think of a unit test but that would require adding a tag and checking that git_archive_id is indeed modified after `git archive` by looking in the files.

rjarry:
@tardyp wrote:
> ahh. I downloaded the archive, and saw that git archive will actually replace that code!
> We surely need to explain that with a comment in the code.

You're right, otherwise that sounds like cosmic ray computing :)

> we can also rather say:
> 
> ```python
> if not git_archive_id.startswith("$Format:"):
> ```

Ok that's simpler indeed.

Are you still having second thoughts about this patch or should I fix these ?

rjarry:
Ok I added comments and used `startswith` instead of a regexp.

As for the unit test, I'm not sure on how to approach this.

master/buildbot/__init__.py
```diff
@@ -80,6 +80,29 @@ def getVersion(init_file):
...
         pass
 
+    # When source is exported via `git archive`, the following line is modified
+    # and placeholders are expanded to the "archived" revision:
+    #
+    #     %ct: committer date, UNIX timestamp
+    #     %d: ref names, like the --decorate option of git-log
+    #
+    # See man gitattributes(5) and git-log(1) (PRETTY FORMATS) for more details.
+    git_archive_id = '$Format:%ct %d$'
```
tardyp: what we can unit test is the extraction method. You can rework that code like this:
```
def getVersionFromArchiveId(git_archive_id):
    if not git_archive_id.startswith('$Format:'):
        # source was modified by git archive, try to parse the version from
        # the value of git_archive_id

        match = re.search(r'tag:\s*v([^,)]+)', git_archive_id)
        if match:
            # archived revision is tagged, use the tag
            return match.group(1)

        # archived revision is not tagged, use the commit date
        tstamp = git_archive_id.strip().split()[0]
        d = datetime.datetime.fromtimestamp(int(tstamp))
        return d.strftime('%Y.%m.%d')
    return None
```
```
def getVersion():
    [...]
    version = getVersionFromArchiveId('$Format:%ct %d$')
    if version is not None:
        return version
```
then the unit test will test results for the different format expected from git-archive (the one without expantion, the one with only timestamp, and the one with timestamp and ref-name)

You can use the same unit test to test the 3 copies of that method (in each module), and put that unit test in buildbot module only. (so that we keep code coverage high, and don't only fix the main one if there is an incompatibility later)

tardyp:
@rjarry as I want to try and release today, I did the refactor and the unit tests for you.

rjarry:
Thanks @tardyp ! Sorry I didn't have time these days.

rjarry:
@tardyp wrote:
> I don't think you answered why BUILDBOT_VERSION method is not usable.

Indeed, I forgot that. I was not sure it was possible to know the version from within Debian packaging scripts. I checked and it looks doable.

Nevertheless, I feel that being able to know the version from a git archive can be useful. This patches do not hurt.

PR #3864, opened 2018-01-02
improve release scripts to use github releases

tardyp:
Now that we use circle ci to create the tarball, we don't build them locally anymore.

eventually, more and more of this process should transition to the CI (still have to figure out how to sign securely)


tardyp:
The problem is I want to make everything automated, including the signing. That means I have to share the private key to CircleCI (if I continue to use them for release), which won't happen...

grafana maintainer is doing it though..
https://github.com/grafana/grafana/blob/master/scripts/build/sign_packages.sh

Even if we do that on our own infra, I am not very found of signing outside of my computer.. 

What would be nice is buildbot being able to relay gpg-agent.

Issue #3878, opened 2018-01-06
umask for master process is set to 0


eugene-katsevman:
On my system when I do `buildbot restart master` twisted is run as a daemon and its umask is set to 0.

This leads to all work and build directories of local workers and master itself having 777 permissions

tardyp:
Hi,
I cannot reproduce this.

I would appreciate a little bit more details.

in your buildbot.tac, do you have those lines:
```
# Default umask for server
umask = None
```

There is no difference between ```buildbot restart``` and ```buildbot start``` in how the process is starting.
You can add a log here to see if this is buildbot which is setting the umask:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/master.py#L230

eugene-katsevman:
Hey, thanks for the quick response!
I've modified umask in buildbot.tac to have 0o22 value and by logging I can confirm that this value is being set (where you pointed at).
But!
I've connected with gdb to twisted process and `call/o umask(0)` gives me 0 immediately. Repo workdir is being created with 777 permission.
If it helps, I'm using a single LocalWorker.

I can provide any other info you need, I just don't know what to.

eugene-katsevman:
an update - umask for localworker is False for whatever reason

    2018-01-06 17:37:16+0300 [-] Starting Worker -- version: 0.9.15
    2018-01-06 17:37:16+0300 [-] with the umask False

tardyp:
can you try the fix in #3881 ?

eugene-katsevman:
It works! Thanks!

PR #3881, opened 2018-01-06
fix umask is set to 0 when using localWorker

tardyp:
since 9f2d60d7c3e4fa7b862d96ce5c627b40397b234e usepty is no more an argument of
BaseWorker, so usepty set as a positional argument was actually setting the umask to 0

Fixes: #3878


eugene-katsevman:
it works! I've applied that patch to my local venv and now everything is good

PR #3890, opened 2018-01-14
fix yarn mutex to work on multiplatform

tardyp:
buildbot_pkg was reportedly broken on windows since the yarn mutex fix

### DIFF
--- a/master/buildbot/newsfragments/windows_pkg.bugfix
+++ b/master/buildbot/newsfragments/windows_pkg.bugfix
@@ -1,0 +1,1 @@
+Fixed issue with ``buildbot_pkg``` not hanging on yarn step on windows (:issue:`3890`).
--- a/pkg/buildbot_pkg.py
+++ b/pkg/buildbot_pkg.py
@@ -214,8 +214,7 @@ class BuildJsCommand(distutils.cmd.Command):
 
             # if we find yarn, then we use it as it is much faster
             if yarn_version != "":
-                # --mutex allows to build in parallel
-                commands.append(['yarn', 'install', '--pure-lockfile', '--mutex', 'file:/tmp/.yarn-mutex'])
+                commands.append(['yarn', 'install', '--pure-lockfile'])
             else:
                 commands.append(['npm', 'install'])
 


tardyp:
@uglycoyote can you test on windows?

uglycoyote:
Sorry, no that seems not to have solved it, running ```pip install -e .``` on my plugin just hangs as it did before, although it's clear from the output that it is now using a OS-specific temp file location as your new code intends.

Here's the output I get from ```pip install -v -e .``` (verbose mode)

```
Obtaining file:///D:/buildbot/buildbot-react-plugin-boilerplate
  Running setup.py (path:D:\buildbot\buildbot-react-plugin-boilerplate\setup.py) egg_info for package from file:///D:/buildbot/buildbot-react-plugin-boilerplate
    Running command python setup.py egg_info
    running egg_info
    running build_js
    yarn: 0.27.5 npm:  5.6.0
    Running command: yarn install --pure-lockfile --mutex file:/C:\Users\admin\AppData\Local\Temp\yarn-mutex
    yarn install v0.27.5
    warning Waiting for the other yarn instance to finish

```

Also, simply running the command ```yarn install --pure-lockfile --mutex file:/C:\Users\admin\AppData\Local\Temp\yarn-mutex``` gives

```
yarn install v0.27.5
warning Waiting for the other yarn instance to finish
```

and hangs indefinitely.

I can confirm that the path ```C:\Users\admin\AppData\Local\Temp\``` exists and is where many temp files are stored already on my computer.  I also tried:

* ```--mutex file:/d:/temp/yarn-mutex``` to see if a different folder would work
* ```--mutex d:/temp/yarn-mutex``` to see if it doesn't really need ```file:/```, but it says ```error unknown single instance type d```
* ```--mutex file://d:/temp/yarn-mutex``` to see if it's really supposed to have two slashes in ```file://``` but it still hangs

also noteworthly is that I had a windows open watching the contents of d:\temp while doing all of this and never actually saw any temp files get created in there.

uglycoyote:
Simply running ```yarn install --pure-lockfile --mutex file``` seems to work fine without locking up.

from https://yarnpkg.com/lang/en/docs/cli/:
> When using file Yarn will write/read a mutex file .yarn-single-instance in the current working directory by default.

maybe that's the best option?

tardyp:
@uglycoyote the goal is to be able to run several package builds in parallel. Each of those packages being in another directory.
If doing yarn install in parallel, the yarn cache files gets corrupted.
It looks like this --mutex option is supposed to fix, that but is pretty half backed. The default value should be the cache directory itself, and not the current directory (makes absolutely no sense to me why would they do that).
Anyway, I'll remove that mutex option and give up parallel build :-/

PR #3893, opened 2018-01-15
master/buildbot/worker/docker.py: Fix test fail when docker is not in…

dol-sen:
…stalled

This fixes numerous tracebacks (note, only partial traceback):

  File "/usr/lib64/python3.6/site-packages/buildbot/worker/docker.py", line 259, in _thd_start_instance
    if docker_py_version >= 2.2:
builtins.NameError: name 'docker_py_version' is not defined

With this patch, the tests pass.

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -41,6 +41,7 @@ try:
 except ImportError:
     docker = None
     client = None
+    docker_py_version = 0.0
 
 
 def _handle_stream_line(line):


dol-sen:
the first 2 test fails are due to an unexpected failure, not due to the patch

Cloning into '.'...
fatal: early EOF
fatal: index-pack failed

I suspect the codecov fails are due to the previous 2.

master/buildbot/worker/docker.py
```diff
@@ -41,6 +41,7 @@
 except ImportError:
     docker = None
     client = None
+    docker_py_version = 0.0
 
 
 def _handle_stream_line(line):
```
tardyp: the test should not pass, it will be skipped.
That means buildbot +  docker will not be supported on gentoo.

I don't think this is what you want. The patch is good though, as docker should not be mandatory for running buildbot (and running the tests).

dol-sen: I have added the docker dependency to the test deps, and made it optional for normal run time.
setup.py dopes not list the docker-py dependency though.  You probably should list it as an optional
dependency.

tardyp: right, we have a lot of optional dependencies, which we do not describe very well (its in the doc, in the relevent parts)

PR #3898, opened 2018-01-19
Only use --force with "git submodules update" if supported

krf:
Support for --force option was added only in Git 1.7.6, so don't use it with
earlier version.

Update the test for the correct commands being generated for both the
currently used 1.7.5 and 1.7.6.

## Contributor Checklist:


tardyp:
cool thanks!

Issue #3901, opened 2018-01-20
Builder page should display worker name instead of id


aj062:
Builders page (http://nine.buildbot.net/#/builders) has workers column on the right which is very useful. However, it displays an id, instead of the worker name. On hover-over it displays the worker name. The worker name should be displayed by default.

Builders page is useful in providing a quick summary of all the builders and worker status. It shows a quick overview of which workers are up and which are down. Displaying worker name would be better as it would quickly show which particular worker are down, and which particular workers are associated with which builder. We can't expect people to remember the id of each worker, whereas name of worker is easily identifiable.

I believe it should be a simple change in https://git.io/vNg3p to always show worker.name

rodrigc:
I agree with you!

PR #3902, opened 2018-01-20
Builder page should display worker-name or worker-id based on settings

ila321:
Fixed https://github.com/buildbot/buildbot/issues/3901

### DIFF
--- a/master/buildbot/newsfragments/worker_name_instead_of_id.bugfix
+++ b/master/buildbot/newsfragments/worker_name_instead_of_id.bugfix
@@ -1,0 +1,1 @@
+Fixed Builder page should display worker name instead of id (:issue:`3901`).
--- a/www/base/src/app/builders/builders.route.coffee
+++ b/www/base/src/app/builders/builders.route.coffee
@@ -36,6 +36,11 @@ class State extends Config
                 caption:'Show old builders'
                 default_value: false
             ,
+                type:'bool'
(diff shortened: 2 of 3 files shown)


tardyp:
My fear, when I designed this page, is that this would totally screw up the page if the number of workers are bigger.

ila321:
> My fear, when I designed this page, is that this would totally screw up the page if the number of workers are bigger.

I believe the benefit (functionality) overweights slightly not-so-good-looking UI issue. The benefit would be in every case, while the issue with longer columns would be only in some cases, when people have large number of workers for a builder. 

Also, when people have large number of workers, they would reasonably understand/expect the display to be longer. However, it is not reasonable to hide the worker names for everyone just because someone can have large worker name or large number of workers.

aj062:
I agree that this would be a useful change. The UI wouldn't really 'screw' up, it would just be wider columns for few builders (which have lot of workers).

Frodox:
Ideally - implement it as option in web settings. Because 10-20+ workers would really screw up page. And in production, imho, teams have more then 1-10 workers.

ila321:
> Ideally - implement it as option in web settings. Because 10-20+ workers would really screw up page. And in production, imho, teams have more then 1-10 workers.

@Frodox Good idea. Updated the patch to implement it as an option in web settings.

tardyp:
Than you all for your collaboration!

Issue #3904, opened 2018-01-21
Owner column is always empty on Builder page


ila321:
Owner column seems to be always empty on the Builder page. e.g.: https://nine.buildbot.net/#/builders/3

Also, we should remove this column and add columns for more meaningful information like revision, worker-name on this page.

Frodox:
dup of #3311 actually

krf:
Bump. Close this since it's a dup?

p12tic:
Fixed.

Issue #3903, opened 2018-01-21
Builder page should show revisions


ila321:
In Buildbot 0.8, builder page displayed revision information for each build. It made very convenient and quick to see which revision each build is building. If a developer want to open the build corresponding to his recent commit, he can quickly look at the revision column and find the build.

'revision' column is missing in Buildbot 0.9 (e.g.: https://nine.buildbot.net/#/builders/3). That seems like a regression. This column should be added again.

Frodox:
there is another tabs/pages for viewing revision, imho

heyakyra:
@Frodox what tab/page would that be?

HashemNasarat-es:
![image](https://user-images.githubusercontent.com/57108885/80421762-fab1de00-88aa-11ea-8768-9e37c3983bcf.png)
perhaps?

I do wish it was on the builer overview page though...
![image](https://user-images.githubusercontent.com/57108885/80421810-11583500-88ab-11ea-80f8-65a447c55c84.png)

tardyp:
Hi @HashemNasarat-es 
There are two problems with showing all properties in builder page 
1) this could add significant query size overhead, as the list of property is unlimited
2) the table would be very large making the UI ugly.

We are open to any suggestion, bonus point if they come with a coded proof of concept!

Issue #3546, opened 2017-08-19
Builder page should show the worker information


aj062:
In Buildbot 0.8, builder page used to show the slaves configured for that builder (and status of the slave, e.g.: "Connected", "Offline" etc.).

However, in Buildbot 0.9, builder page do not show the worker information at all. This is very useful information. This information should be displayed by default (or at least controlled by an option on settings page).

aakashja:
Can this be configured in the template?

Frodox:
"Builders" page show this information at the right column. Also, "Workers" page is useful enough.
Yes, it can be configured (not like in 0.8 exactly) - http://docs.buildbot.net/latest/developer/www-base-app.html + https://medium.com/buildbot/buildbot-ui-plugin-for-python-developer-ef9dcfdedac0

aakashja:
I (and most of the people in my team) strongly feel that missing workers information on builder page is significant regression from buildbot 0.8.

It was very convenient to directly see the workers associated with the builder in 0.8, especially while debugging issue with a builder. In Buildbot 0.9, we need to open builders page and search for that builder. This introduces more steps in the workflow to get basic information.

It would be great if Buildbot can bring this feature back.

tardyp:
please remember that Buildbot is 100% run by its community, which you are an important member. This feature is quite easy to add, if this will help your team to have it, please consider contributing it.

PR #3905, opened 2018-01-21
Display revision and worker column on Builder page

ila321:
Fixed https://github.com/buildbot/buildbot/issues/3903
Fixed https://github.com/buildbot/buildbot/issues/3904
Partially fix https://github.com/buildbot/buildbot/issues/3546

### DIFF
--- a/www/base/src/app/common/directives/builds/buildstable.tpl.jade
+++ b/www/base/src/app/common/directives/builds/buildstable.tpl.jade
@@ -8,7 +8,8 @@
         td(width='100px') #
         td(width='150px') Started At
         td(width='150px') Duration
-        td(width='200px') Owner
+        td(width='150px') Revision
+        td(width='150px') Worker
         td Status
       tr(ng-repeat='build in builds | orderBy:"-started_at"')
           td(ng-show="builders") {{ builders.get(build.builderid).name }}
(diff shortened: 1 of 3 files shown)


tardyp:
I am worried that got revision is not always set.

Also sha1 revision might not be useful unless it is associated to change data (like in console)

We had lots of back and forth bugs for console view in order to get proper revision that works for all the cases.

As we are into fetching more properties, why not add the 'owner' and property, which should be set at beg of build (actually only when build is forced, but that can be fixed):
https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/build.py#L473

Also please copy paste screenshots in your description to show what it looks like.

tardyp:
see https://github.com/buildbot/buildbot/pull/3909 for a fix making the owners property have the full blamelist.

www/base/src/app/common/directives/builds/buildstable.tpl.jade
```diff
@@ -25,6 +26,12 @@
             span(ng-show="build.complete", title="{{(build.complete_at - build.started_at)| durationformat:'LLL' }}")
               | {{(build.complete_at - build.started_at)| duration }}
           td
+            span(title="Revision")
+              p(ng-repeat='(codebase, revision) in build.properties.got_revision[0]')
```
tardyp: looks like you need to add track_by (see smoke tests

Issue #3545, opened 2017-08-19
Builds column on builders page doesn't display recent builds for many builders 


aj062:
builder page (e.g.: https://nine.buildbot.net/#/builders) displays the recent builds corresponding to each builder. However, for many of the builders, this column is empty, even though the builder have many builds in the past.

In below screenshots builds are not displayed for many builders (e.g.  "buildbot py:3.6 sqla:l t:trial tw:trunk"). This seems to happens more frequently for the builders whose last build was little old as compared to rest of the builders. In these screenshot, the last build was few months ago, but this happens in many other cases as well. Build info would be missing for builders which had last build few hours ago, if there are many other builders which had their last build few minutes ago.

<img width="909" alt="screen shot 2017-08-19 at 8 03 52 am" src="https://user-images.githubusercontent.com/3098702/29487944-68abc296-84b6-11e7-8529-e16886456291.png">

<img width="785" alt="screen shot 2017-08-19 at 8 04 16 am" src="https://user-images.githubusercontent.com/3098702/29487950-81a6e29e-84b6-11e7-8c7b-2e2d3634d4c0.png">

Missing this important information makes it very hard to look at the builder page and see status of all the builders. In Buildbot 0.8, builders page always displayed last build of every builder.

aakashja:
Is there a overall limit of 200 recent builds across all builders? 
as defined here: https://git.io/v5fnk

aakashja:
This will be an issue on any buildbot instance have large number of builders. Even on https://nine.buildbot.net/#/builders with 50+ builders, this is clearly an issue.

Ideally, recent build information should be fetched and displayed for every builder. But if it affects performance significantly, then  instead of a fixed number '200', a multiple of NUMBER_OF_BUILDERS should be used (e.g.: 15 x number_of_builders).

tardyp:
Indeed, in order to reduce the number of data api needed to display the builder page, we choose to only display the 200 last builds.

We should have this value configurable like we do for the other dashboards. I dont think we should try to be too smart at multiplying the number of builders. There are too many usecases this could hurt

aakashja:
Yeah, having this value configurable in settings (or master.cfg) file would be good.

For now, how can we change this value in our CI. By using custom template?

tardyp:
you can't really, as it is hardcoded in the coffeescript :-(

[...]

PR #3907, opened 2018-01-21
Fix for Builds column on builders page doesn't display recent builds for many builders

ila321:
Fixed https://github.com/buildbot/buildbot/issues/3545

200 builds was an extremely small number of builds to fetch. Even https://nine.buildbot.net/#/builders doesn't have bubbles for many builders. Large instances with large number of builders see this issue prominently.

Given that data output from Buildbot 0.9 API is very compact, fetching 1000 builds should be fine.

### DIFF
--- a/master/buildbot/newsfragments/increase_num_bubbles.bugfix
+++ b/master/buildbot/newsfragments/increase_num_bubbles.bugfix
@@ -1,0 +1,1 @@
+Fixed Builds column on builders page doesn't display recent builds for many builders (:issue:`3545`).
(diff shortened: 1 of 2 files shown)


tardyp:
On nine.buildbot.net, the builder load page goes from 1.39 s to 2s. I think this starts to be very sloppy.
I think as we are addressing the largest instance, we need a better solution.

ila321:
I like proposed solution in  #3908.

Should we still have a better default of 500 builds? So that people who do not configure this in master.cfg still have less chances of noticing this bug.

Updated this pull-request for 500 builds.

tardyp:
I think 200 builds is already a big value. So I'll close that for now in favor of #3924

PR #3924, opened 2018-01-28
Add capability to override the default UI settings

### LINKED ISSUE #3908: Web UI User settings should have defaults configurable in master.cfg
c['www']['default_ui_settings'] = {
'Builders.buildFetchLimit': 1000
}

This can be done easily by whitelisting in https://github.com/buildbot/buildbot/blob/master/master/buildbot/config.py#L807
And the the settings module in the UI can fetch its default value there (overriding the default hardcoded in the coffee)

tardyp:
Fixes #3908


aj062:
This is awesome. Thanks :)

PR #3938, opened 2018-02-06
steps/source/repo.py: Use --force-sync for sync step

willnewton:
Pass the --force-sync option to repo sync. This ensures that if
we change a module location in the manifest then the sync will
still succeed.

This option is not the default behaviour as it can cause data loss
in the working directory but this is not a concern for buildbot
as no changes are ever made in the checked out copy.

## Remove this paragraph

And especially:

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/repo_force_sync.feature
+++ b/master/buildbot/newsfragments/repo_force_sync.feature
@@ -1,0 +1,1 @@
+The repo source step now syncs with the ``--force-sync`` flag which allows the sync to proceed when a source repo in the manifest has changed.
--- a/master/buildbot/steps/source/repo.py
+++ b/master/buildbot/steps/source/repo.py
@@ -311,7 +311,7 @@ class Repo(Source):
         for command in self.manifestDownloads:
             yield self._Cmd(command, workdir=self.build.path_module.join(self.workdir, ".repo", "manifests"))
 
-        command = ['sync']
+        command = ['sync', '--force-sync']
         if self.jobs:
             command.append('-j' + str(self.jobs))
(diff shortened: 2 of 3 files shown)


tardyp:
change is okay, but you really need to do your usual homework...

- update the tests
- add newsframent
- update documentation (add a warning in the repo section)

willnewton:
@tardyp sorry, I fixed the tests and added the news fragment. I didn't add anything to the docs as I was unsure how to word it. From a user's perspective all this should do is make syncs that failed previously now succeed, I can't imagine a scenario where a user would have modifications in their checked out copy (it would also be overwritten if they used the tarball option).

master/buildbot/newsfragments/repo_force_sync.feature
```diff
@@ -0,0 +1 @@
+The repo source step now syncs with the --force-sync flag which allows the sync to proceed when a source repo in the manifest has changed.
```
tardyp: you need to double-back-quote ``--force-sync``

willnewton:
@tardyp I updated the new fragment following your suggestion.

vvscloud:
I feel this should be updated in document, as this feature will make repo step fail with old workers. as --force-sync flag is unrecognized by old versions of repo. Did we test the compatibility of this change with old workers?

tardyp:
Is there any reason to use older version of repo? repo is supposed to be automatically updated to the latest version.
--force-sync has been added in 2014..

https://gerrit.googlesource.com/git-repo/+/abaa7f312f1b6c8d11d7c757fe909900ce5788b5

Issue #3771, opened 2017-11-21
ShellSequence lack of logging


pohmelie:
I have `ShellSequence` in my `BuildFactory`:
``` python
commands = [
    util.ShellArg(command=["git", "-C", str(p), "checkout", repo.branch or "master"]),
    util.ShellArg(command=["git", "-C", str(p), "pull"]),
    util.ShellArg(command=["git", "-C", str(p), "checkout", revision]),
    util.ShellArg(command=["git", "-C", str(p), "submodule", "update", "--init"]),
]
...
steps.ShellSequence(name="update", haltOnFailure=True, commands=commands, timeout=None, env=env),
...
```
But I see no output in web ui:
![screenshot-2017-11-21 buildbot console view](https://user-images.githubusercontent.com/2385765/33079180-fdc7e250-cee5-11e7-847e-933fc3bcf6ac.png)

It looks like `makeRemoteShellCommand` does not set `collectStdout` and `collectStderr`
https://github.com/buildbot/buildbot/blob/master/master/buildbot/steps/shellsequence.py#L117
https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/buildstep.py#L1196

chrspeich:
I had the same problem.

Turns out you need so specify logfile='something' for each ShellArg to get the stdio log of that process. That name is somewhat misleading, as it does not specify an actual file to be read on the worker, but specifies the name of the stdio log captured.

tardyp:
Thanks for the hint. I agree that maybe we should rename that parameter logname=
We should however keep the backward compatibility not to break people's master.cfg.

would you like to contribute that?

pohmelie:
@chrspeich 
Can't test until tomorrow, but what it will looks like in web ui?

tardyp:
@pohmelie if you choose the same logname, it will use the same log for all commands.
but you can choose different names, and there will be one logfile per command if you wish

pohmelie:
@tardyp 
Sounds good, will try this tomorrow :+1:

pohmelie:
Unfortunately can't test this without extra work, since right now we have no `ShellSequence` in our configuration.

PR #3950, opened 2018-02-12
Rename logfile to logname in ShellArg

chrspeich:
The name logfile suggests that a file should be specified, while it is just the log name to use for the stdio output of the command beeing run. This commit renames logfile to logname while keeping the old logfile parameter around for compatibility.

This addresses the problem mentioned in #3771.

I'm not sure about the following part in the doc:

> Note that if ``logname`` name does not start with the prefix ``stdio``, that prefix will be set like ``stdio <logname>``

I don't see this on my buildbot. When I specify ``logfile='test'`` the log is just displayed as ``test`` in the web ui.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/shellarg-rename-logfile-logname.bugfix
+++ b/master/buildbot/newsfragments/shellarg-rename-logfile-logname.bugfix
@@ -1,0 +1,1 @@
+Rename ShellArg's logfile to logname to reflect the actual meaning. (:issue:`3771`)
--- a/master/buildbot/newsfragments/shellarg-rename-logfile-logname.bugfix
+++ b/master/buildbot/newsfragments/shellarg-rename-logfile-logname.bugfix
@@ -1,1 +1,1 @@
-Rename ShellArg's logfile to logname to reflect the actual meaning. (:issue:`3771`)
+Deprecated ShellArg's ``logfile`` argument over ``logname`` to reflect the actual meaning. (:issue:`3771`)
--- a/master/buildbot/steps/shellsequence.py
+++ b/master/buildbot/steps/shellsequence.py
@@ -42,6 +42,7 @@ class ShellArg(results.ResultComputingConfigMixin):
 
         self.logname = logname
         if logfile is not None:
+            config.warnDeprecated('1.0.1', "logfile is deprecated, use logname")
             if self.logname is not None:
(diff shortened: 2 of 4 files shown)


master/buildbot/steps/shellsequence.py
```diff
@@ -31,15 +31,22 @@
...
+    def __init__(self, command=None, logname=None, logfile=None, **kwargs):
         name = self.__class__.__name__
         if command is None:
             config.error("the 'command' parameter of %s "
                          "must not be None" % (name,))
         self.command = command
-        self.logfile = logfile
+
+        self.logname = logname
+        if logfile is not None:
```
tardyp: you can add a deprecation message here.

chrspeich:
I've added a ``config.warnDeprecated``. I guessed ``1.0.1`` as version, I hops this is okay.

tardyp:
1.0.1 is fine. but you need to fix the tests, and make sure to swallow the warning in the test

chrspeich:
So, sorry, I forgot to push this. I've fixed the failing tests.

master/buildbot/newsfragments/shellarg-rename-logfile-logname.bugfix
```diff
@@ -0,0 +1 @@
+Deprecated ShellArg's ``logfile`` argument over ``logname`` to reflect the actual meaning. (:issue:`3771`)
```
tardyp: I slightly reworded the newsfragment

master/buildbot/steps/shellsequence.py
```diff
@@ -31,15 +31,23 @@
...
+            config.warnDeprecated('1.0.1', "logfile is deprecated, use logname")
+            if self.logname is not None:
```
tardyp: I can see those lines are uncovered. We must add some unit tests to make sure that logfile is still working (i.e no mistake error in the warning)

tardyp: You can see here example how to flush and verify the warnings:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/test_worker_transition.py#L107

p12tic:
Removing `stalled` label as this PR is going to be merged as part of #5681.

PR #3955, opened 2018-02-16
buildbot.reporters.hipchat: Include as reporter

cmouse:
The documentation had old example that no longer worked


seankelly:
This is because it's missing from the `reporters` list in `setup.py`. I think we should always document using `buildbot.plugins` and fix any plugins that can't be imported.

cmouse:
so... why is it failing to import in the first...

seankelly:
Because it isn't listed in `setup.py`. The `plugins` importer uses that.

cmouse:
ok, I'll repair this fix to do that then.

Issue #3413, opened 2017-07-04
Relax builder name length restriction


aj062:
As per http://docs.buildbot.net/latest/manual/installation/nine-upgrade.html#identifiers, builder name can be only up to 20 characters. 

This is not at all practical. For builder names to be readable, it usually contains many keywords (like platform name, os, architecture, version, type etc.). In our instance we have over 100+ builders and builder-names of upto 65 characters. Reducing them to 20 characters is just not possible.

Looking at http://trac.buildbot.net/wiki/SuccessStories, a lot of people use long builder names. Here are list of few buildbot instance and corresponding maximum builder name length:

53 characters: http://lab.llvm.org:8011/builders
52 characters: https://buildbot.openlighting.org/builders
44 characters: https://gdb-build.sergiodj.net/builders
40 characters: https://build.webkit.org/builders/
35 characters: http://testing.evergreen-ils.org/buildbot/builders
34 characters: http://buildbot.python.org/all/builders
34 characters: http://buildbot.openafs.org:8010/builders
32  characters: https://build.macports.org/builders
30 characters: https://buildbot.askmonty.org/buildbot/builders
32 characters: https://ci.apache.org/builders

Having this impractical restriction would make it more difficult for people to upgrade to buildbot 0.9 (only 1 out of ~50 buildbots linked on http://trac.buildbot.net/wiki/SuccessStories has upgraded yet).

I suggest increasing the limit to 100 (or atleast 70) characters for builder names.

tardyp:
Having larger buildername gives more problems in term of UI. It is already quite hard to make a UI able to display up to 20 characters.
For our nine.buildbot.net, we decided to use abbreviation for our builder names. see
http://nine.buildbot.net/#/console
having 70 characters buildernames would have made this console view completly unreadable.

There are already the builder description and builder tags for the needed additional information for a builder.

actually the restriction is 50 chars as far as I can see from here: https://github.com/buildbot/buildbot/blob/master/master/buildbot/data/builders.py#L86
You can try to increase it, but I dont think it is a good idea for the above reasons

seankelly:
I think documenting the length is actually 50 is fine. Only two Buildbot instances from the list above exceed 50 characters.

ila321:
I think we should relax the restriction since there would be many more private Buildbot instances hitting this limit.

Also, is there really a restriction in the first place?  Builder name seems to be stored as sa.Text in db without a length restriction: https://git.io/vA8ZF

PR #3957, opened 2018-02-18
Relax builder name length restriction

ila321:
Fix https://github.com/buildbot/buildbot/issues/3413

There are clearly some open-source instances which uses builder names longer than 50 character. There would be many more private instances which would be hit by this issue.

The only concern with longer builder names is that some UI like console view might not look so good. I believe this decision should be left with the admin of Buildbot instead of forcing them to shorten the names, some Buildbot instances do not even enable console/waterfall view, even in those cases we are enforcing this strict limit.

### DIFF
--- a/master/buildbot/data/builders.py
+++ b/master/buildbot/data/builders.py
@@ -87,7 +87,7 @@ class Builder(base.ResourceType):
 
     class EntityType(types.Entity):
         builderid = types.Integer()
-        name = types.Identifier(50)
+        name = types.Identifier(70)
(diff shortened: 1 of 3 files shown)


tardyp:
> The only concern with longer builder names is that some UI like console view might not look so good. I believe this decision should be left with the admin of Buildbot instead of forcing them to shorten the names, some Buildbot instances do not even enable console/waterfall view, even in those cases we are enforcing this strict limit.

Lets agree to disagree :)

Issue #3414, opened 2017-07-04
Relax step name length restriction


aakashja:
As per http://docs.buildbot.net/latest/manual/installation/nine-upgrade.html#identifiers, step name can be up to 50 characters.

step names are usually descriptive and can easily go over 50 characters. In our instance we have step names upto 60 characters. I believe other people might also have faced this problem.

I suggest increasing the limit to 70 characters for step names, so that it makes it easier for people to upgrade to buildbot 0.9

seankelly:
While 50 characters is a lot, I do think we should make it possible to allow a much larger limit for usability. It would have to be transformed somehow before going in the database.

PR #3958, opened 2018-02-18
Relax step name length restriction

ila321:
Fix https://github.com/buildbot/buildbot/issues/3414

### DIFF
--- a/master/buildbot/test/unit/test_db_migrate_versions_044_add_step_hidden.py
+++ b/master/buildbot/test/unit/test_db_migrate_versions_044_add_step_hidden.py
@@ -40,7 +40,7 @@ class Migration(migration.MigrateTestMixin, unittest.TestCase):
             'steps', metadata,
             sa.Column('id', sa.Integer, primary_key=True),
             sa.Column('number', sa.Integer, nullable=False),
-            sa.Column('name', sa.String(50), nullable=False),
+            sa.Column('name', sa.String(70), nullable=False),
             sa.Column('buildid', sa.Integer),
             sa.Column('started_at', sa.Integer),
             sa.Column('complete_at', sa.Integer),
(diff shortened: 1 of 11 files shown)


master/buildbot/test/unit/test_db_migrate_versions_044_add_step_hidden.py
```diff
@@ -40,7 +40,7 @@ def create_tables_thd(self, conn):
             'steps', metadata,
             sa.Column('id', sa.Integer, primary_key=True),
             sa.Column('number', sa.Integer, nullable=False),
-            sa.Column('name', sa.String(50), nullable=False),
+            sa.Column('name', sa.String(70), nullable=False),
```
tardyp: You will have to create a new migration script in order to increase this restriction.
I believe this was a bad idea to limit the size of name inside the db. probably it will be better to change the column to sa.Text() so that there is no limitation.

sphet: I know this pull is failing and I'm with you -- no restriction would be better. Can we put the desire to change the step length to a Text field instead on a list somewhere? What would it take to simply make that change? We would need some kind of upgrade process?

tardyp: @sphet not sure I fully get your comment. I updated my review in order to take in account what I understand.

sphet: @tardyp All I meant was 1) please no restrictions - let's move to a variable-length text field 2) Can we ensure this change is on the roadmap somewhere. Thanks!

ila321: @tardyp @sphet Agree. However, I am not much familiar with database migration script,  and I am afraid that I wouldn't have much time to learn about it and implement the changes.

Would anyone of you have time to implement this?

PR #3961, opened 2018-02-19
Fix IRC output for Python3 master

zware:
A Python3 master outputs `bytes` reprs to IRC:

```
[py-bb] b'Build x86 Windows7 3.6 #209 is complete: Warnings [\x038warnings test (warnings)\x0f] - http://buildbot.python.org/all/#builders/90/builds/209'
```

Stripping explicit encodes from `master/buildbot/reporters/irc.py` and just letting Twisted handle the encoding seems to fix it complete with color:

```
[py-bb] Build AMD64 Windows8 3.x #659 is complete: Success [build successful] - http://buildbot.python.org/all/#builders/32/builds/659
```

I don't understand the buildbot codebase or Twisted framework well enough to say whether this fix is actually *correct*, and can't get the tests to work properly on my machine anyway, so please feel free to fix up this PR as necessary or close it outright in favor of the correct fix.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -60,13 +60,13 @@ class IrcStatusBot(StatusBot, irc.IRCClient):
 
     # The following methods are called when we write something.
     def groupChat(self, channel, message):
-        self.notice(channel, message.encode('utf-8', 'replace'))
+        self.notice(channel, message)
 
     def chat(self, user, message):
-        self.msg(user, message.encode('utf-8', 'replace'))
+        self.msg(user, message)
 
     def groupDescribe(self, channel, action):
-        self.describe(channel, action.encode('utf-8', 'replace'))
+        self.describe(channel, action)
 
     def getContact(self, user=None, channel=None):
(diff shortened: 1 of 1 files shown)


rodrigc:
Unfortunately, this needs to be fixed lower down in Twisted.
There is a ticket for this: http://twistedmatrix.com/trac/ticket/6320
It will require some work, and my queue is a bit full right now. :(

tardyp:
@rodrigc Can we find a workaround just for this problem?

Looks like the problems comes from https://github.com/twisted/twisted/blob/twisted-17.9.0/src/twisted/words/protocols/irc.py#L1732
```
        fmt = 'PRIVMSG %s :' % (user,)
```
in py3, it looks like ``str(b'foo')`` is ``"b'foo'"`` so this is not really repr(), but actually str()

I think, @zware workaround is correct as indeed the current twisted code is doing the encoding eventually:
https://github.com/twisted/twisted/blob/twisted-17.9.0/src/twisted/words/protocols/irc.py#L264

So here we would just have to fix the unit test.

PR #3974, opened 2018-02-27
Appveyor: install build requirements

xoviat:


appveyor.yml
```diff
@@ -15,6 +15,8 @@ install:
   - "python -c \"import sys; print(sys.exec_prefix)\""
   - "python -c \"import sys; print(sys.executable)\""
   - "python -V -V"
+  - "python -m pip install -U pip setuptools wheel"
+  - "python -m pip install -U git+https://github.com/pypa/pip.git"
```
tardyp: So we need git pip in order to make pypiwin32 work? I guess I prefer to stick with pypiwin32 < 222

PR #3975, opened 2018-02-28
remove ramlfications from dependencies

tardyp:
Raml being too much flexible, it turns out very difficult to make generic tooling.
So ramlfication struggles to reach raml 1.0 support.

The Buildbot raml spec as it is now still is useful, we just include the basic parser able to generate the doc.

This will help @rjarry work of pushing buildbot to debian


rjarry:
@tardyp can you squash the 3 commits ? they don't have much sense on their own.

tardyp:
done

PR #3991, opened 2018-03-12
Make the title of the builder blocks on the homepage a link

alexlarsson:
This makes it easy to go directly to the builder, rather than a specific
build.


alexlarsson:
Wow, thats a new one:
```
```

tardyp:
> error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.

Yeah this is a github instability. Buildbot Git step is supposed to retry, but that doesn't work :-(

PR #4017, opened 2018-03-27
Delete batch file accidentally added with pull request #3997

uglycoyote:
Apologies, this was a batch file I had added to make it convenient to start "gulp dev proxy" but has hardcoded server name.  I have moved this outside my repo directory so it doesn't accidentally get added anymore.


tardyp:
ahah my bad not to spot it in review!

Issue #4039, opened 2018-04-11
Official buildbot-master docker image lacks "requests" for github auth


kallisti5:
From image here: https://hub.docker.com/r/buildbot/buildbot-master/

Config:
```python
        'auth': util.GitHubAuth("XXX", util.Secret("github-auth-secret"), getTeamsMembership=True),
        'authz': util.Authz(
                allowRules=[
                        util.AnyControlEndpointMatcher(role="infrastructure"),
                ],
                roleMatchers=[
                        util.RolesFromGroups(groupPrefix='XXX/')
                ]
        )
```

Error:
```
exceptions.AttributeError: Unable to load buildbot.util:GitHubAuth: No module named requests
```

tardyp:
The official master image is by design minimalistic. If you need additional features, you need to create your own image like we do for e.g buildbot_travis:

https://github.com/buildbot/buildbot_travis/blob/master/Dockerfile

I am not completly against doing a full featured Docker image though.

kallisti5:
> The official master image is by design minimalistic.

Sure, but the stock plugins that buildbot provides don't work because of a missing python module. It seems kind of odd to say "we don't ship this python module we require because we're minimalistic"

tardyp:
Indeed, there are a lot of stocks plugins. Some require various python dependencies not present in the default "setup.py". Those additional dependencies are usually described in the doc.

kallisti5:
@tardyp  Except they're not per #3455 :-)

PR #4040, opened 2018-04-11
master/Dockerfile: Add py-requests to solve #4039

kallisti5:


rodrigc:
Fixed in https://github.com/buildbot/buildbot/pull/4041

PR #4046, opened 2018-04-13
Webpackoja preparation: embed guanlecoja-ui, and remove ng-classify

tardyp:
For now, this is only for main UI.

@rjarry


tardyp:
@rjarry I have now modified buildbot_pkg to build the javascripts with python instead of gulp.

It still needs coffee-script node-uglifyjs and node-less installed to build.
jade is not available in debian, so I used pyjade (vendored in buildbot_pkg) in order to build.

The problem with this solution is that it will be obsolete as soon as we merge #4056 from @uglycoyote and start using webpack, because we start using require(), and then we will depend on webpack to build.

PR #4049, opened 2018-04-15
Decaffeinate smokes

rodrigc:
## Contributor Checklist:


uglycoyote:
Those changes all look pretty straightforward.  It seems like it's mainly the Angular aspects of the data module that we were struggling with, whereas this code doesn't use angular so it is more straightforward.  Did these tests all run for you fine after the conversion (i did not test, just looked over your diffs a bit)

rodrigc:
I observed the tests passed here: https://nine.buildbot.net/#/builders/10/builds/2318/steps/6/logs/stdio

and I ran them locally too.

As you mentioned, this stuff is simple.  I don't see Coffeescript being significantly easier to understand than Javascript for this stuff, so I think this is fine.

rodrigc:
@tardyp OK, I removed a bunch of *return* that were unnecessary.  Tests are failing because chromium version is outdated, so the Docker image of metabbotcfg needs to be rebuilt to get the newer chrome version.

rodrigc:
@tardyp OK, smoke tests now pass after you upgraded metabbotcfg to have newer chromium browser

PR #4097, opened 2018-05-15
gitlab web hook: don't schedule build on non-interesting changes; fixes #3635

### LINKED ISSUE #3635: Buildbot schedules build when a GitLab merge request is deleted
My master installation is running 28958cffbdf703e29b76fb24344848bd65e8ea8a.
In the Buildbot configuration a "gitlab" change hook is registered and triggered by a GitLab configured "Merge Requests Event".

Currently builds are triggering when a merge request is created, updated or deleted.
Build should not be scheduled when a merge request is deleted.

dankegel:
Idea is simple: don't rebuild when somebody e.g. closes the MR or assigns it to somebody else;
save rebuilding for when we know there's something worth rebuilding.

This MR causes builds to happen only when a MR is originally opened, when it is reopened, or when new code is pushed.

### DIFF
--- a/master/buildbot/newsfragments/issue3635-filter-gitlab-mr-better.bugfix
+++ b/master/buildbot/newsfragments/issue3635-filter-gitlab-mr-better.bugfix
@@ -1,0 +1,1 @@
+Don't schedule a build when a GitLab merge request is deleted or edited (:issue:3635)
--- a/master/buildbot/newsfragments/issue3635-filter-gitlab-mr-better.bugfix
+++ b/master/buildbot/newsfragments/issue3635-filter-gitlab-mr-better.bugfix
@@ -1,1 +1,1 @@
-Don't schedule a build when a GitLab merge request is deleted or edited (:issue:3635)
+Don't schedule a build when a GitLab merge request is deleted or edited (:issue:`3635`)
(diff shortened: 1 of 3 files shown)


master/buildbot/www/hooks/gitlab.py
```diff
@@ -108,6 +108,19 @@ def _process_merge_request_change(self, payload, project, event, codebase=None):
...
+
+        # Filter out uninteresting events
+        state = attrs['state']
+        if re.match('^(closed|merged|approved)$', state):
+            log.msg("GitLab MR#{}: Ignoring because state is {}".format(attrs['iid'], state))
+            return []
+        # I'd like to be able to let pushes trigger a retry, but don't know how to do that.
+        # So only allow opens or reopens of an MR to trigger a try build.
+        action = attrs['action']
+        if not re.match('^(open|reopen)$', action):
```
tardyp: I think indeed, we should rebuild in case of merge request update. I think the action for that is just "update": https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/models/event.rb#L20

dankegel:
That appveyor problem looks unrelated...?

Review comments addressed, unmarking as WIP.

master/buildbot/newsfragments/issue3635-filter-gitlab-mr-better.bugfix
```diff
@@ -0,0 +1 @@
+Don't schedule a build when a GitLab merge request is deleted or edited (:issue:3635)
```
tardyp: :issue:`3635`

PR #4106, opened 2018-05-18
Allow Perforce change sources to resolve the 'who' field into a more verbose format using a callable

sphet:

### DIFF
--- a/master/buildbot/changes/p4poller.py
+++ b/master/buildbot/changes/p4poller.py
@@ -125,7 +125,7 @@ class P4Source(base.PollingChangeSource, util.ComparableMixin):
                  pollInterval=60 * 10, histmax=None, pollinterval=-2,
                  encoding='utf8', project=None, name=None,
                  use_tickets=False, ticket_login_interval=60 * 60 * 24,
-                 server_tz=None, pollAtLaunch=False, revlink=lambda branch, revision: (u''), resolvewho=lambda who: who):
+                 server_tz=None, pollAtLaunch=False, revlink=lambda branch, revision: (u''), resolvewho=lambda who: (who)):
 
         # for backward compatibility; the parameter used to be spelled with 'i'
         if pollinterval != -2:
@@ -148,6 +148,10 @@ class P4Source(base.PollingChangeSource, util.ComparableMixin):
         if not callable(revlink):
             config.error(
                 "You need to provide a valid callable for revlink")
+
+        if not callable(resolvewho):
+            config.error(
+                "You need to provide a valid callable for resolvewho")
 
         self.p4port = p4port
         self.p4user = p4user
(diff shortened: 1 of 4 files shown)


master/buildbot/changes/p4poller.py
```diff
@@ -125,7 +125,7 @@ def __init__(self, p4port=None, p4user=None, p4passwd=None,
                  pollInterval=60 * 10, histmax=None, pollinterval=-2,
                  encoding='utf8', project=None, name=None,
                  use_tickets=False, ticket_login_interval=60 * 60 * 24,
-                 server_tz=None, pollAtLaunch=False, revlink=lambda branch, revision: (u'')):
+                 server_tz=None, pollAtLaunch=False, revlink=lambda branch, revision: (u''), resolvewho=lambda who: (who)):
```
seankelly: Any way this could be done as a renderable instead?

sphet: @seankelly - I'm not sure I understand how a renderable would be appropriate here. I'm open to any suggestions though. 

In our case we take the 'who' in the form of user@workspace - for example Charles@SecondWorkspace - and need to map it to charles.oneal@workspace.com. We pipe it through a callable that has LDAP and perforce user list knowledge, so it is a couple of dictionary checks and some additional logic. How do renderables work in this case?

seankelly: A renderable can be a function to do that. It meshes better with how Buildbot handles dynamic text. The issue is that its input is the build's properties.

I'm hoping @tardyp might have an idea.

sphet:
Sean

I suspect that because this happens at the poller/changes level it is "too
soon" to be part of the build renderables?

master/buildbot/changes/p4poller.py
tardyp: yeah, I've thought of that too, but renderable really do not have anything to interpolate in pollers

sphet:
@tardyp Can you point me in the direction (an example) of writing a unit test to test the config error? Thanks.

tardyp:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/test_changes_gitpoller.py#L1340

master/buildbot/test/unit/test_changes_p4poller.py
```diff
@@ -441,7 +441,9 @@ def check(res):
         d.addCallback(check)
         return d
 
-
+    def test_resolveWho_callable(self):
+        self.assertRaisesConfigError("You need to provide a valid callable for resolvewho",
```
tardyp: you need to inherit configmixin to have this check available

sphet: Thanks - got sidetracked in the office today and forgot to check my update PR

sphet:
Not sure what those remaining failures are caused by.

tardyp:
Looks like the remaining failures are unusually high false positive rate. Can you please squash your history and force push? This will restart the CI

tardyp:
thanks!

PR #4115, opened 2018-05-25
Improve Trigger step documentaion

peterboba:
Add more info to the set_properties param of Trigger step as it was lacking the important info


tardyp:
thanks

Issue #4077, opened 2018-05-05
GitHubAuth: 'NoneType' object has no attribute '__getitem__'


ibuclaw:
Config:
```
        c['www']['auth'] = util.GitHubAuth(os.environ['BUILDMASTER_CLIENT_ID'],
                                           os.environ['BUILDMASTER_CLIENT_SECRET'],
                                           apiVersion=4, getTeamsMembership=True)
```

Traceback when attempting to login using oauth.
```
buildbot_1      | 2018-05-05 10:25:48+0000 [-] /usr/lib/python2.7/site-packages/urllib3/connectionpool.py:858: urllib3.exceptions.InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
buildbot_1      | 2018-05-05 10:25:49+0000 [-] /usr/lib/python2.7/site-packages/urllib3/connectionpool.py:858: urllib3.exceptions.InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
buildbot_1      | 2018-05-05 10:25:49+0000 [-] While rendering resource:
buildbot_1      | 	Traceback (most recent call last):
buildbot_1      | 	  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 434, in errback
buildbot_1      | 	    self._startRunCallbacks(fail)
buildbot_1      | 	  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 501, in _startRunCallbacks
buildbot_1      | 	    self._runCallbacks()
buildbot_1      | 	  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
buildbot_1      | 	    current.result = callback(current.result, *args, **kw)
buildbot_1      | 	  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 1184, in gotResult
buildbot_1      | 	    _inlineCallbacks(r, g, deferred)
buildbot_1      | 	--- <exception caught here> ---
buildbot_1      | 	  File "/usr/lib/python2.7/site-packages/twisted/internet/defer.py", line 1126, in _inlineCallbacks
buildbot_1      | 	    result = result.throwExceptionIntoGenerator(g)
[...]

PR #4116, opened 2018-05-25
Supplimentary fix for GH team membership check.

ibuclaw:
Github had a bug in their GraphQL API where if _one_ organization restricted access for OAuth apps to read internal team data, then the returned response would have no data for _all_ organizations the user is a member of.  This has now been fixed, and check for None has been moved accordingly to accomodate the new response that github returns.

See #4077.


ibuclaw:
FYI @tardyp

PR #4122, opened 2018-05-28
Use .winerror accessor instead of indexing

estan:
Seems at some point pywin32 silently dropped the support for indexing.
This fixes the error brought up by @rodrigc here:

  https://github.com/buildbot/buildbot/issues/3796#issuecomment-349159727


tardyp:
cool thanks!

PR #4128, opened 2018-06-01
hooks: allow remote ref to be specified for GitHub hook closes #3998

### LINKED ISSUE #3998: GitHub source step fails on missing magic `/merge` ref of GitHub
Currently I run into an issue where the changehook is too fast for the magic `merge` ref in GitHub to be created on Pull Requests.

An easy fix would be to try again in this case or wait a reasonable amount in `startVC()` of the GitHub step if the remote ref ends in `merge`.

Thoughts?

noc0lour:

This will allow bigger repositories with a lot of merge conflicts to avoid
spurious build failures due to not/late available `/merge` ref


tardyp:
cool!

PR #4131, opened 2018-06-03
Fix bug in worker disconnected notice

zware:
This bug causes worker disconnection notices to not be sent to our CPython buildbot worker owners. The fix is straightforward and apparently not covered by existing tests; I haven't attempted to add one.

## Contributor Checklist:

* [x] ~~I have updated the appropriate documentation~~


tardyp:
thanks @zware Could you please add a newsfragment, so that this fix is not forgotten in the releasenotes?

zware:
@tardyp Done (I think :)).

PR #4154, opened 2018-06-17
WIP: Add support for specifying private SSH keys for Git and GitPoller

p12tic:
Currently there's no way to clone ssh:// git repositories if public key authentication is needed. E.g. Gitlab (at least the self-hosted instances) don't allow cloning via user+password combo. The workaround is to put the private key into the worker along with appropriate ssh config, which is inconvenient when many workers are involved and keys need to be changed or introduced.

I've hacked a proof of concept code for the first approach. The Git source step will upload the private SSH key when needed, use appropriate `core.sshCommand` config option to specify the key, and delete the key file when done. This way keys can be updated by just changing the secret on the master without any changes on the workers.

Similar approach has been adopted for GitPoller. This allows the keys to be specified as a secret and avoids the need to modify the buildbot master when changing SSH keys.

The private keys are exposed as sshPrivateKey parameter to both Git and GitPoller. The value may be either secret or just a string.

I'd like to add support for old git versions too by creating a temporary wrapper SSH script on the worker just like this PR does with the private SSH key. Do you think it's worthwhile to add this to the official buildbot project? I need this for myself anyway, so I'll likely maintain my own patches if this is deemed too hacky :-)

Please tell me what do you think about this and I will finish the PR according to the feedback.

## Contributor Checklist:

* [partially] I have updated the unit tests
* [not yet] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [not yet] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -20,6 +20,7 @@ from future.utils import itervalues
 
 import os
 import re
+import stat
 
 from twisted.internet import defer
 from twisted.internet import utils
@@ -44,7 +45,9 @@ class GitPoller(base.PollingChangeSource, StateMixin):
     compare_attrs = ("repourl", "branches", "workdir",
                      "pollInterval", "gitbin", "usetimestamps",
                      "category", "project", "pollAtLaunch",
-                     "buildPushesWithNoCommits")
+                     "buildPushesWithNoCommits", "sshPrivateKey")
+
+    secrets = ("sshPrivateKey", )
 
     def __init__(self, repourl, branches=None, branch=None,
                  workdir=None, pollInterval=10 * 60,
@@ -52,7 +55,8 @@ class GitPoller(base.PollingChangeSource, StateMixin):
                  category=None, project=None,
                  pollinterval=-2, fetch_refspec=None,
                  encoding='utf-8', name=None, pollAtLaunch=False,
-                 buildPushesWithNoCommits=False, only_tags=False):
+                 buildPushesWithNoCommits=False, only_tags=False,
+                 sshPrivateKey=None):
 
         # for backward compatibility; the parameter used to be spelled with 'i'
         if pollinterval != -2:
@@ -63,7 +67,8 @@ class GitPoller(base.PollingChangeSource, StateMixin):
 
         base.PollingChangeSource.__init__(self, name=name,
                                           pollInterval=pollInterval,
-                                          pollAtLaunch=pollAtLaunch)
+                                          pollAtLaunch=pollAtLaunch,
+                                          sshPrivateKey=sshPrivateKey)
 
         if project is None:
             project = ''
@@ -92,6 +97,7 @@ class GitPoller(base.PollingChangeSource, StateMixin):
         self.project = bytes2unicode(project, encoding=self.encoding)
         self.changeCount = 0
         self.lastRev = {}
+        self.sshPrivateKey = sshPrivateKey
 
         if fetch_refspec is not None:
             config.error("GitPoller: fetch_refspec is no longer supported. "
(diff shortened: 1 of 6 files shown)


master/buildbot/changes/gitpoller.py
```diff
@@ -63,7 +67,8 @@ def __init__(self, repourl, branches=None, branch=None,
 
         base.PollingChangeSource.__init__(self, name=name,
                                           pollInterval=pollInterval,
-                                          pollAtLaunch=pollAtLaunch)
+                                          pollAtLaunch=pollAtLaunch,
+                                          sshPrivateKey=sshPrivateKey)
```
tardyp: why do you need to pass it to parent if your store it in this class?

master/buildbot/changes/gitpoller.py
```diff
@@ -338,27 +344,67 @@ def _process_changes(self, newRev, branch):
...
+        commandsThatNeedKey = [
+            'fetch',
+            'ls-remote',
+        ]
+        if self.sshPrivateKey is not None and command in commandsThatNeedKey:
+            return True
+        return False
+
+    def _downloadSshPrivateKey(self, keyPath):
+        with open(keyPath, 'w') as keyFile:
```
tardyp:
```
    try:
        oldmask = os.umask(0o077)
        with open(keyPath, 'w') as keyFile:
            keyFile.write(self.sshPrivateKey)
    finally:
        os.umask(oldmask)
```

tardyp: additional merit points for writing an utility context manager in buildbot.utils
..and make sure this works on Windows, hint: tests!

master/buildbot/steps/source/base.py
```diff
@@ -220,6 +220,36 @@ def evaluateCommand(_):
             return cmd.rc
         return d
 
+    def _downloadFile(self, buf, filename, workdir=None, mode=None):
```
tardyp: There is the CompositeStepMixin, which provide this method

master/buildbot/changes/gitpoller.py
tardyp: I now realize that you stole this code from the worker/transfer code.

That code actually is 12 years old (https://github.com/buildbot/buildbot/commit/05317ec1a0bc96f8c66910235b2de228e173a29a#diff-af6cb5744179a15d0117a8201da0995dR819), so I think it is time to not trust the comment from @warner, and just use the try: finally method.

tardyp:
BTW, I see you are from unity.
Is this a hint katana is moving to nine? That would be great!

master/buildbot/changes/gitpoller.py
p12tic: It turns out that the BuildbotService requires any secrets to be rendered to be passed as kwargs to __init__. See https://github.com/buildbot/buildbot/blob/dfad61a971c696a4688181a06ae139241799deda/master/buildbot/util/service.py#L211.

master/buildbot/changes/gitpoller.py
p12tic: The way I've written the code definitely has a race condition, thus the comment :-) But even the umask method has a race condition: if the attacker creates keyPath just after umask, but before open, the permissions won't be changed and the key file would retain the original permissions. Also, if something on another thread depends on os.umask value, then this would result in a race condition that might affect correctness and would be a hell to debug.

Wouldn't it be a better idea just to rely on the permissions of the parent directory?

master/buildbot/steps/source/base.py
p12tic: Yeah, noticed that. I just wanted to leave refactoring of this for another PR.

p12tic:
> Would you mind splitting the PR into two?
Agreed. Maybe even a third one for downloadFile refactor.

>  Is this a hint katana is moving to nine? That would be great!
This PR is done on my own personal time :-) I'm from a different team and don't even really know the internal roadmaps of Katana.

By the way, do you think it's worthwhile to add support for git versions from 1.0 to 2.10 by using a script wrapper for SSH? Knowing how much to include into the official PR would save a lot of time for me. Thanks!

master/buildbot/changes/gitpoller.py
tardyp: > if the attacker creates keyPath just after umask, but before open, the permissions won't be changed

that is what O_EXCL is made for. I agree umask is not thread safe and no good solution.
```

import os

flags = os.O_WRONLY | os.O_CREAT | os.EXLC
with os.fdopen(os.open('testfile.txt', flags, 0o600), 'w') as fout:
    fout.write("secrets!")
```
I fear that might be posix only.

> Wouldn't it be a better idea just to rely on the permissions of the parent directory?

Then we would have to check that.

p12tic: What do you think about something like this?
```
os.mkdir('path/to/keydir', mode=0o700)
    with open('path/to/keydir/key, 'w') as keyFile:
        keyFile.write(self.sshPrivateKey)
    <do stuff>
shutil.rmtree('path/to/keydir')```

p12tic:
> By the way, do you think it's worthwhile to add support for git versions from 1.0 to 2.10 to official buildbot by using a script wrapper for SSH? Knowing how much to include into the PR and how much I will need to keep for myself would save a lot of time for me. Thanks!

Ping :-)

seankelly:
Ubuntu 16.04 has git 2.7. Debian 9 has 2.11 while oldstable has 2.1. My vote would be to support 2.3 and later. That was released in February 2015.

p12tic:
Will split into three PRs: one for CompositeStepMixin.downloadFileContentToWorker (which has already landed), one for the Git part and one for the Gitpoller part.

PR #4160, opened 2018-06-21
Add support for specifying private SSH keys for Git

p12tic:
Currently there's no way to clone ssh:// git repositories if public key authentication is needed. E.g. Gitlab (at least the self-hosted instances) don't allow cloning via user+password combo. The workaround is to put the private key into the worker along with appropriate ssh config, which is inconvenient when many workers are involved and keys need to be changed or introduced.

This PR implements the first two approaches and supports git 2.3 and newer. The Git source step will upload the private SSH key when needed, use appropriate `core.sshCommand` config or environment option to specify the key, and delete the key file when done. This way keys can be updated by just changing the secret on the master without any changes on the workers.

A similar approach is being implemented for GitPoller.

The private keys are exposed as sshPrivateKey parameter to Git step. The value may be either secret or just a string.

## Contributor Checklist:

* [yes] I have updated the unit tests
* [yes] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [yes] I have updated the appropriate documentation

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -94,6 +94,7 @@ backoff
 backporting
 backslashing
 basedir
+basename
 basepath
 baserev
 basetgz
--- a/master/buildbot/newsfragments/git-support-private-ssh-key.feature
+++ b/master/buildbot/newsfragments/git-support-private-ssh-key.feature
@@ -1,0 +1,2 @@
+-:bb:step:`Git` now supports `sshPrivateKey` parameter to specify private ssh
+key for fetch operations. This feature is supported on git 2.3 and newer.
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -28,6 +28,7 @@ from buildbot import config as bbconfig
 from buildbot.interfaces import WorkerTooOldError
 from buildbot.process import buildstep
 from buildbot.process import remotecommand
+from buildbot.process.properties import Properties
 from buildbot.steps.source.base import Source
 
 RC_SUCCESS = 0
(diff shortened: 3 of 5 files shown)


p12tic:
Please don't merge just yet, testing on Windows shows issues.

seankelly:
Addendum to my comment in #4154, the git in CentOS 7 is only 1.8. I think supporting 2.3+ is still fine; this is to document where various distros are.

tardyp:
which kind of windows issues?

Is it possible to write a test to show the issue?  a test that appveyor could catch

p12tic:
> which kind of windows issues?

There were my own environment issues causing ssh host key verification to fail. Also, downloading ssh key failed due to strange permission error which I assume were caused by wrong use of relative paths, e.g. `path\to\key\nonexisting-path\..\keyname`. 5529a96 fixes this by properly computing relative paths.

In addition to that the temporary file for the ssh key is now downloaded into a temporary directory. This will allow to better protect the private key in the future: I'd like to add `mode` parameter for the worker `mkdir` command so that it's possible to create new directories with the permissions different from default `0o777`. We could then create the new directories for the ssh key with the `0o700` permission set.

By the way, could the test suite be rerun for this PR? The test failures seem to be caused by unrelated reasons and are likely instabilities.

I think the PR is good to go now.

p12tic:
Closing to restart the tests upon reopen.

tardyp:
looks like there is an issue with npm right now, or something like this. I (or someone) have to investigate

tardyp:
https://github.com/tardyp/gulp-bower-deps/issues/2

PR #4174, opened 2018-06-27
Simplify test code by using @defer.inlineCallbacks where possible

p12tic:
Currently there are plenty of places in tests that use `d.addCallback()` or `@d.addCallback()` for little benefit in my opinion. Using `@defer.inlineCallbacks` decorator makes the code more consistent with the rest of the buildbot codebase and easier to understand. This style is promoted in the buildbot documentation. Since there's a better alternative exists, using `d.addCallback()` and `@d.addCallback()` is a kind of small tech debt.

I propose we gradually change `d.addCallback()` or `@d.addCallback()` in the test code to use `@defer.inlineCallbacks` decorator. While such churn of code probably does not make sense in the main codebase due to risk of regressions, changing tests brings no such downside.

What do you think? If you don't agree, I would really like to learn the reasons why :-)

The PR is just an illustration of what changes would be done.

## Contributor Checklist:

* [does not apply] I have updated the unit tests
* [does not apply] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [does not apply] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/test/unit/test_changes_hgpoller.py
+++ b/master/buildbot/test/unit/test_changes_hgpoller.py
@@ -35,12 +35,13 @@ class TestHgPoller(gpo.GetProcessOutputMixin,
                    unittest.TestCase):
     usetimestamps = True
 
+    @defer.inlineCallbacks
     def setUp(self):
         # To test that environment variables get propagated to subprocesses
         # (See #2116)
         os.environ[ENVIRON_2116_KEY] = 'TRUE'
         self.setUpGetProcessOutput()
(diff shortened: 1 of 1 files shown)


p12tic:
Unrelated question: looks like the [protractor tests](https://nine.buildbot.net/#/builders/26/builds/2562) are quite unstable. Shouldn't the timeout be increased as the error message suggests? Instabilities bring reduced attention to tests and real failures may be accidentally missed.

PR #4176, opened 2018-06-27
pylint: Fix invalid spelling errors when running tests locally

p12tic:
For some reason pylint tests fail with spelling errors when run locally via `make virtualenv; . .venv/bin/activate; make pylint`. Presumably `make virtualenv` should download the correct versions of tools, but there's unfortunately some environment difference between the CI setup and what's downloaded locally.

The spelling errors look invalid though, this PR adds appropriate blacklist entries for the affected words.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -226,6 +226,7 @@ cb
 cbc
 cbd
 cd
+ce
 cfg
 cgi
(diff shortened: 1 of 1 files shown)


tardyp:
This looks like a dictionary issue, so probably some aspell dictionaries difference, which are not managed by make virtualenv :-(

PR #4173, opened 2018-06-26
Add support for specifying private SSH keys for Git older than v2.3

p12tic:
This PR adds support for specifying private SSH keys for Git older than v2.3. The outcome of the discussions in https://github.com/buildbot/buildbot/pull/4160 and https://github.com/buildbot/buildbot/pull/4154 was that the official buildbot doesn't necessarily need support for Git older than v2.3. However, I need this feature anyway, so the code was written regardless. The implementation is simple and I personally think that the benefits outweigh the minimal increase of maintenance burden for the buildbot project. But feel free to reject the PR, at least it's public so someone who badly needs this feature might cherry pick the commits :-)

Major distributions that are still supported and ship git older than 2.3:
 - Debian Jessie, [git 2.1.4](https://packages.debian.org/jessie/git), [supported till 2020-06-30](https://wiki.debian.org/LTS)
 - Ubuntu Trusty, [git 1.9.1](https://packages.ubuntu.com/trusty-updates/git), [supported till 2019-04](https://wiki.ubuntu.com/Releases)
 - Centos 6, [git 1.7.1](http://mirror.centos.org/centos/6/os/x86_64/Packages/), [supported till 2020-11-30](https://wiki.centos.org/About/Product)
 - Centos 7, [git 1.8.3](http://mirror.centos.org/centos/7/os/x86_64/Packages/), [supported till 2024-06-30](https://wiki.centos.org/About/Product)

The changes have been tested on Windows and Linux.

## Contributor Checklist:

* [x] I have updated the unit tests
* [x] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [x] I have updated the appropriate documentation

PR #4178, opened 2018-06-27
Add support for specifying private SSH keys for GitPoller

p12tic:
Currently there's no way to poll ssh:// git repositories if public key authentication is needed. E.g. Gitlab (at least the self-hosted instances) don't allow cloning via user+password combo. The workaround is to put the private key into the worker along with appropriate ssh config, which is inconvenient when many workers are involved and keys need to be changed or introduced.

This PR implements the first two approaches and supports git 2.3 and newer. The GitPoller source step will upload the private SSH key when needed, use appropriate `core.sshCommand` config or environment option to specify the key, and delete the key file when done. This way keys can be updated by just changing the secret in the secret provider without changing anything on the master.

A similar approach has been implemented for Git step. See https://github.com/buildbot/buildbot/pull/4160 and https://github.com/buildbot/buildbot/pull/4173.

The private keys are exposed as sshPrivateKey parameter to GitPoller step. The value may be either secret or just a string.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1086,6 +1086,7 @@ requestsubmitted
 requeue
 requiredargs
 reschedulenextbuild
+resetmocks
 resourceneedsreconfigs
 resourcetype
 restrmatcher
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -338,27 +338,17 @@ class GitPoller(base.PollingChangeSource, StateMixin):
                 repository=bytes2unicode(self.repourl, encoding=self.encoding),
                 category=self.category, src=u'git')
 
+    @defer.inlineCallbacks
     def _dovccmd(self, command, args, path=None):
-        d = utils.getProcessOutputAndValue(self.gitbin,
+        res = yield utils.getProcessOutputAndValue(self.gitbin,
             [command] + args, path=path, env=os.environ)
-
-        def _convert_nonzero_to_failure(res,
-                                        command,
-                                        args,
-                                        path):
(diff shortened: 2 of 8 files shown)


seankelly:
I haven't looked through all of the code, but it looks similar to what's in the Git step. I'm fine merging with it duplicated but seems like it would be good to later have a "gitutil" (and probably similar items for Mercurial) that the step and poller can use.

p12tic:
Yeah, I agree wrt. duplication. The reason why I haven't done that yet is that in the Git step everything happens in the worker and in the GitPoller step everything happens on master. So various details are different even though the structure is the same. Deduplicating the "structure" part may make everything much more complex.

tardyp:
the windows tests fail with:

```
[ERROR]
Failure: exceptions.OSError: (2, 'CreateProcess', 'The system cannot find the file specified.')
buildbot.test.integration.test_worker_transition.RunMaster.test_config_0_9_0b5
buildbot.test.integration.test_worker_transition.RunMaster.test_config_0_9_0b5_api_renamed
Ran 5534 tests in 146.685s
FAILED (skips=60, errors=2, successes=5472)
```
does that test use GitPoller?

tardyp:
The test passes on mainline: https://ci.appveyor.com/project/djmitche/buildbot/build/1.0.6592

p12tic:
Turns out the test failure was caused by doing git feature checks in `activate()` function of gitpoller. `activate()` started depending on `git` being installed and this broke the tests that fully start and shutdown the master. I've moved the checks to `poll()` instead and the test should now pass. Bonus is that GitPoller will now pick up newly installed or upgraded git instead of needing a master restart.

PR #4181, opened 2018-06-30
GitHub hook fixes

### LINKED ISSUE #4180: Use hmac.compare_digest in the GitHub webhook if possible
Currently buildbot uses naive `!=` to compare digests. 

https://github.com/buildbot/buildbot/blob/f522a22a88d6f2d39e8ff1afe639dc7ff6ef4a39/master/buildbot/www/hooks/github.py#L117-L120

I propose to use hmac.compare_digest() if possible. For example: (not tested yet)

```Python
def _cmp(a, b):
    try:
        # try the more secure compare_digest() first
        from hmac import compare_digest
        return compare_digest(a, b)
    except ImportError:
        # and fallback to the insecure simple comparison otherwise
        return a == b

if not _cmp(mac.hexdigest(), hexdigest):
    raise ValueError('Hash mismatch') 
```

If this sounds a good idea, I'll try to open a pull request.

BTW, HMAC verification has been implemented long time ago (https://github.com/buildbot/buildbot/commit/baf1f47a5521dae8752cc4a236cbf35ff4db0e18), but the doc implies it's missing (https://github.com/buildbot/buildbot/blob/master/master/docs/manual/cfg-wwwhooks.rst)

> Patches are welcome to implement: https://developer.github.com/webhooks/securing/

EDIT: a grammar fix

yan12125:
## Contributor Checklist:

     - Not sure if such a change is worth a news item or not

Closes #4180

### DIFF
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -124,7 +124,7 @@ class GitHubEventHandler(PullRequestMixin):
                     # and fallback to the insecure simple comparison otherwise
                     return a == b
 
-            if not _cmp(mac.hexdigest(), unicode2bytes(hexdigest)):
+            if not _cmp(bytes2unicode(mac.hexdigest()), hexdigest):
                 raise ValueError('Hash mismatch')
 
         content_type = request.getHeader(b'Content-Type')
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -114,9 +114,17 @@ class GitHubEventHandler(PullRequestMixin):
             mac = hmac.new(unicode2bytes(self._secret),
                            msg=unicode2bytes(content),
                            digestmod=sha1)
-            # NOTE: hmac.compare_digest should be used, but it's only available
-            # starting Python 2.7.7
-            if mac.hexdigest() != hexdigest:
+
+            def _cmp(a, b):
+                try:
+                    # try the more secure compare_digest() first
+                    from hmac import compare_digest
+                    return compare_digest(a, b)
+                except ImportError:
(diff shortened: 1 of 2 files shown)


yan12125:
Buildbot failures are due to network issues:
```
Cloning into '.'...
fatal: early EOF
fatal: index-pack failed
```
(https://nine.buildbot.net/#/builders/6/builds/2573)
```
fatal: early EOF
fatal: index-pack failed
```
https://nine.buildbot.net/#/builders/5/builds/2625

And the Appveyor job is pending in a queue with more than 100 jobs.

BTW, should I squash the follow-up fixes or keep them as-is for now?

master/buildbot/www/hooks/github.py
```diff
@@ -114,9 +114,17 @@ def _get_payload(self, request):
...
                            msg=unicode2bytes(content),
                            digestmod=sha1)
-            # NOTE: hmac.compare_digest should be used, but it's only available
-            # starting Python 2.7.7
-            if mac.hexdigest() != hexdigest:
+
+            def _cmp(a, b):
+                try:
+                    # try the more secure compare_digest() first
+                    from hmac import compare_digest
```
tardyp: we don't support python 2.6, so that's fine to always use  compare_digest

yan12125: Hmm. Are Python 2.7.0~2.7.6 unsupported, either?

tardyp: well, we don't go to that precision, indeed. But we don't test python minor versions, so I don't . I guess lets keep it that way, as it is not much of a burden. Stats show that there are around 111 unique installs with python 2.7.5..

PR #4193, opened 2018-07-04
WIP: Merge requests are like patches; they should not be merged.

dankegel:
See issue 4190.

Please don't review yet, this is a WIP.


tardyp:
I think it makes sense to collapse mergerequests. If you push some crap, they do the mergerequest, then review yourself and immediately push another fix, then in this case you want to collapse with the crap. You even would like to cancel the previous MR test if it has started.

dankegel:
The tests fail locally because I can't seem to set buildset properties in my test case; the fake database is a bit of a mystery to me.

master/buildbot/process/buildrequest.py
```diff
@@ -267,7 +267,17 @@ def canBeCollapsed(master, br1, br2):
         if br1['buildsetid'] == br2['buildsetid']:
             defer.returnValue(True)
 
-        # get the buidlsets for each buildrequest
+        # Merge requests should only be collapsed with other
```
ngladitz: I think this fails testing because when the new buildset is not a merge request it will not run the new logic and will collapse with non merge requests.

I think the following changes might make sense (they fix the test for me at least)? :
```patch
diff --git a/master/buildbot/process/buildrequest.py b/master/buildbot/process/buildrequest.py
index a6f83ce9f..2923d49e8 100644
--- a/master/buildbot/process/buildrequest.py
+++ b/master/buildbot/process/buildrequest.py
@@ -270,8 +270,12 @@ class BuildRequest(object):
         # Merge requests should only be collapsed with other
         # merge requests from the same branch of the same source project
         bsp1 = yield master.db.buildsets.getBuildsetProperties(br1['buildsetid'])
+        bsp2 = yield master.db.buildsets.getBuildsetProperties(br2['buildsetid'])
+
+        if bsp1.get('event') != bsp2.get('event'):
+            defer.returnValue(False)
+
         if bsp1.get('event') == ('merge_request', 'Change'):
-            bsp2 = yield master.db.buildsets.getBuildsetProperties(br2['buildsetid'])
             if (bsp2.get('event') != ('merge_request', 'Change') or
                     bsp1.get('source_branch') != bsp2.get('source_branch') or
                     bsp1.get('source_git_ssh_url') != bsp2.get('source_git_ssh_url')):
```

master/buildbot/process/buildrequest.py
```diff
@@ -267,7 +267,18 @@ def canBeCollapsed(master, br1, br2):
         if br1['buildsetid'] == br2['buildsetid']:
             defer.returnValue(True)
 
-        # get the buidlsets for each buildrequest
+        # Don't collapse e.g. GitLab merge requests with normal builds
+        bsp1 = yield master.db.buildsets.getBuildsetProperties(br1['buildsetid'])
```
tardyp: I would prefer that you use the data API to access this (as people may copy this code in their own master config as an example)

master/buildbot/process/buildrequest.py
```diff
@@ -267,7 +267,18 @@ def canBeCollapsed(master, br1, br2):
...
+        bsp2 = yield master.db.buildsets.getBuildsetProperties(br2['buildsetid'])
```
tardyp: please get the properties for bsp2 only if the bsp1 change has merge_request property (for optimisation)

dankegel: I did that originally, but had trouble avoiding improper collapsing then, as normal builds would then collapse themselves with merge requests :-(

master/buildbot/process/buildrequest.py
dankegel: Tried that just now, but:
```
-        bsp1 = yield master.db.buildsets.getBuildsetProperties(br1['buildsetid'])
-        bsp2 = yield master.db.buildsets.getBuildsetProperties(br2['buildsetid'])
+        bsp1 = yield master.data.get(('buildsets', br1['buildsetid'], 'properties'))
+        bsp2 = yield master.data.get(('buildsets', br1['buildsetid'], 'properties'))
```
fails with
```
twisted.trial.unittest.FailTest: Lists differ: [20] != [17, 18, 19, 20]
```
in test_collapseRequests_collapse_default_with_a_merge_request

dankegel:
I originally wanted to use the standard sourceset comparison method, but did not know how to encode the additional info.  Having magic values for 'patch_body' seemed wrong somehow, and adding fields sounded above my pay grade.  How would you suggest approaching it?

master/buildbot/process/buildrequest.py
ngladitz: @dankegel for `bsp2` you probably meant to use `br2['buildsetid']` rather than `br1['buildsetid']`?

dankegel: Quite.  Can I blame "coding immediately after waking up for a midnight snack"?  :-) Fixed.

master/buildbot/process/buildrequest.py
```diff
@@ -267,7 +267,18 @@ def canBeCollapsed(master, br1, br2):
         if br1['buildsetid'] == br2['buildsetid']:
             defer.returnValue(True)
 
-        # get the buidlsets for each buildrequest
+        # Don't collapse e.g. GitLab merge requests with normal builds
```
tardyp: I really think it is not a good idea to include this gitlab specific code in here.

If we can't get it to work with the sourcestamp method, we should put it in another function that should be configured as collapseRequests in the mergerequest processing builder. Then modify the gitlab example master.cfg

dankegel: Oh, I didn't say we couldn't get it to work with sourcestamps, just that I hesitated to overload the patch field.  Let me give that a shot, it would be easier for users than requiring them to add a collapseRequests field in master.cfg...

master/docs/manual/cfg-builders.rst
```diff
@@ -152,7 +152,8 @@ If ``collapseRequests`` is given, it completely overrides the global configurati
 Possible values for both ``collapseRequests`` configurations are:
 
 ``True``
-    Requests will be collapsed if their sourcestamp are compatible (see below for definition of compatible).
+    Requests will be collapsed if they and their sourcestamps are
```
sa2ajj: do not break one-sentence-per-line please

master/docs/manual/cfg-builders.rst
```diff
@@ -167,6 +168,13 @@ Sourcestamps are compatible if all of the below conditions are met:
 * Neither source stamp has a patch (e.g., from a try scheduler)
 * Either both source stamps are associated with changes, or neither are associated with changes but they have matching revisions.
 
+Build requests are compatible unless they have
```
sa2ajj: please comply with one-sentence-per-line policy

master/docs/manual/cfg-builders.rst
dankegel: Had not heard of that rule, I'll fix.

sa2ajj: Hmm... Then we need to fix it.

When changes are following this policy, it's much easier to review the changes (especially "typos").

sa2ajj: Oh, found it: http://docs.buildbot.net/latest/developer/style.html#documentation

sa2ajj:
On a somewhat related note, you may want to rebase the branch.

PR #4200, opened 2018-07-08
Move code duplicated across git source step and gitpoller to util.git

p12tic:
This prepares for ssh host key support.

## Contributor Checklist:

* [not applicable] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)

### DIFF
--- a/master/buildbot/util/git.py
+++ b/master/buildbot/util/git.py
@@ -1,0 +1,52 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+from distutils.version import LooseVersion
+
+
+class GitMixin(object):
+
+    def setupGit(self):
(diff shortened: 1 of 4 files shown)


master/buildbot/util/git.py
```diff
@@ -0,0 +1,77 @@
...
+#
+# Copyright Buildbot Team Members
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+from distutils.version import LooseVersion
+
+
+class GitFeatures(object):
```
tardyp: We usually use the Mixin design pattern in order to handle this.
https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-are-they-useful

I think this would be more coherent to continue use this pattern

p12tic: Thanks for suggestion. Fixed now.

master/buildbot/util/git.py
```diff
@@ -0,0 +1,71 @@
...
+            self.supportsSubmoduleForce = True
+        if LooseVersion(version) >= LooseVersion("1.7.8"):
+            self.supportsSubmoduleCheckout = True
+        if LooseVersion(version) >= LooseVersion("2.3.0"):
+            self.supportsSshPrivateKeyAsEnvOption = True
+        if LooseVersion(version) >= LooseVersion("2.10.0"):
+            self.supportsSshPrivateKeyAsConfigOption = True
+
+
+def adjustCommandParamsForSshPrivateKey(command, env, gitFeatures,
```
tardyp: why did you put this outside of the Mixin?

master/buildbot/util/git.py
```diff
@@ -0,0 +1,71 @@
...
+                                        keyPath, sshWrapperPath=None):
+    if gitFeatures.supportsSshPrivateKeyAsConfigOption:
```
tardyp: you could just use self.supportsSshPrivateKeyAsConfigOption if you were inside the mixin

p12tic: Right. I was thinking about the mixin as strictly containing functionality related to supported feature mask. Perhaps the class should be renamed to GitMixin if it contains additional functions.

tardyp: right

p12tic: Should be good now.

PR #4201, opened 2018-07-08
Add support for explicit SSH host key checking to Git and Gitpoller

p12tic:
The PR adds support for checking SSH host keys explicitly when SSH private key is specified. This avoids any need to maintain `known_hosts` files in any way or change the SSH configuration. If SSH config is modified to accept new host keys by default, this increase the risk of MITM in containerized test environments where depending on container setup the saved host key may be lost each time container state is purged.

## Contributor Checklist:


tardyp:
needs rebase

p12tic:
Rebased and ran smoketest on my buildbot instance.

PR #4212, opened 2018-07-11
Badge rendering overrides, and fixes

gregorydlogan:
## Contributor Checklist:

## Description

We were experiencing issues where rendered badges looked like  rather than the expected common formats.  We are using the Buildbot Docker image for the master, so I'm guessing this is somehow related to the Cairo implementation within that image.  This patch implements:

- An optional override to Cairo's autodetection.  
- A modification to the badges to render text in the right place

## Known TODOs:
- Is this a feature, or a bugfix?  I have not created the newsfragment item.
- Unclear how I might test this.  `trail buildbot.test` passes, but I'm guessing there's more to do?

### DIFF
--- a/master/docs/manual/cfg-www.rst
+++ b/master/docs/manual/cfg-www.rst
@@ -268,6 +268,8 @@ The default templates are very much configurable via the following options.
     {
         "left_text": "Build Status",  # text on the left part of the image
         "left_color": "#555",  # color of the left part of the image
+        "left_size": "40", # override autodetected font width, omit if not required
+        "right_size": "50", # override autodetected font width, omit if not required
         "style": "flat",  # style of the template availables are "flat", "flat-square", "plastic"
         "template_name": "{style}.svg.j2",  # name of the template
         "font_face": "DejaVu Sans",
--- a/www/badges/buildbot_badges/templates/flat-square.svg.j2
+++ b/www/badges/buildbot_badges/templates/flat-square.svg.j2
@@ -5,6 +5,6 @@
   <rect width="{{ left.width + right.width }}" height="20" fill-opacity=".1"/>
   <g fill="#fff" text-anchor="middle" font-family="{{config.font_face}},Verdana,Geneva,sans-serif" font-size="{{config.font_size}}">
     <text x="{{ left.width/2+1 }}" y="14">{{ left.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="14">{{ right.text }}</text>
+    <text x="{{ right.width/2+left.width-1 }}" y="14">{{ right.text }}</text>
   </g>
 </svg>
--- a/www/badges/buildbot_badges/templates/flat.svg.j2
+++ b/www/badges/buildbot_badges/templates/flat.svg.j2
@@ -10,7 +10,7 @@
   <g fill="#fff" text-anchor="middle" font-family="{{config.font_face}},Verdana,Geneva,sans-serif" font-size="{{config.font_size}}">
     <text x="{{ left.width/2+1 }}" y="15" fill="#010101" fill-opacity=".3">{{ left.text }}</text>
     <text x="{{ left.width/2+1 }}" y="14">{{ left.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="15" fill="#010101" fill-opacity=".3">{{ right.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="14">{{ right.text }}</text>
+    <text x="{{ right.width/2+left.width-1 }}" y="15" fill="#010101" fill-opacity=".3">{{ right.text }}</text>
+    <text x="{{ right.width/2+left.width-1 }}" y="14">{{ right.text }}</text>
   </g>
 </svg>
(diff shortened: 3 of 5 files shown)


gregorydlogan:
Will add the newsfragment when I get a moment.

I had not considered that it might be a font issue - it renders so the font has to be present, right?  Maybe it's only present in Python (or something falling back to a default font), and not present for Cairo.  I'll change the default font and see if that changes anything.

gregorydlogan:
Ok, definitely related to fonts in some way.  Installing the `ttf-dejavu` package into the docker image remedies the mangled render.  The finished badge is still tighter than I would make it (
) but is acceptable.

Moving forward, what should the resolution here be?  IMO it's best if the width can be auto-detected, but to do that properly we'd need to embed a lot of fonts, or restrict user choise.  I'm fine with embedding, since I'm going to be doing that downstream anyway, but that might rub some folks the wrong way.  Likewise with requiring the use of specific fonts. Thoughts?

tardyp:
> Moving forward, what should the resolution here be? IMO it's best if the width can be auto-detected, but to do that properly we'd need to embed a lot of fonts, or restrict user choise. I'm fine with embedding, since I'm going to be doing that downstream anyway, but that might rub some folks the wrong way. Likewise with requiring the use of specific fonts. Thoughts?

I think you are right the best is to keep it auto-detected. We should probably fail to produce a badge if the font is not present, with a 500 error with proper error message telling the admin to install the right font.

www/badges/buildbot_badges/templates/flat-square.svg.j2
```diff
@@ -5,6 +5,6 @@
   <rect width="{{ left.width + right.width }}" height="20" fill-opacity=".1"/>
   <g fill="#fff" text-anchor="middle" font-family="{{config.font_face}},Verdana,Geneva,sans-serif" font-size="{{config.font_size}}">
     <text x="{{ left.width/2+1 }}" y="14">{{ left.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="14">{{ right.text }}</text>
+    <text x="{{ right.width/2+left.width-1 }}" y="14">{{ right.text }}</text>
```
sa2ajj: I wonder if there could be a bit more descriptive commit message justifying the change: if I read it right, the expression looks the same, so a bit more reasoning behind the change would be more than welcome.

www/badges/buildbot_badges/templates/flat.svg.j2
```diff
@@ -10,7 +10,7 @@
   <g fill="#fff" text-anchor="middle" font-family="{{config.font_face}},Verdana,Geneva,sans-serif" font-size="{{config.font_size}}">
     <text x="{{ left.width/2+1 }}" y="15" fill="#010101" fill-opacity=".3">{{ left.text }}</text>
     <text x="{{ left.width/2+1 }}" y="14">{{ left.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="15" fill="#010101" fill-opacity=".3">{{ right.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="14">{{ right.text }}</text>
+    <text x="{{ right.width/2+left.width-1 }}" y="15" fill="#010101" fill-opacity=".3">{{ right.text }}</text>
```
sa2ajj: ditto for these two lines.

www/badges/buildbot_badges/templates/plastic.svg.j2
```diff
@@ -12,7 +12,7 @@
   <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11">
     <text x="{{ left.width/2+1 }}" y="14" fill="#010101" fill-opacity=".3">{{ left.text }}</text>
     <text x="{{ left.width/2+1 }}" y="13">{{ left.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="14" fill="#010101" fill-opacity=".3">{{ right.text }}</text>
-    <text x="{{ left.width+right.width/2-1 }}" y="13">{{ right.text }}</text>
+    <text x="{{ right.width/2+left.width-1 }}" y="14" fill="#010101" fill-opacity=".3">{{ right.text }}</text>
```
sa2ajj: and ditto for these two as well.

master/docs/manual/cfg-www.rst
```diff
@@ -268,6 +268,8 @@ The default templates are very much configurable via the following options.
     {
         "left_text": "Build Status",  # text on the left part of the image
         "left_color": "#555",  # color of the left part of the image
+        "left_size": "40", # override autodetected font width, omit if not required
```
sa2ajj: could you please elaborate a bit more how this change improves things? compared to the current state.

www/badges/buildbot_badges/templates/flat-square.svg.j2
gregorydlogan: The expression is actually quite different:  You have two parts, `left` and `right`.  We want the text for `right` to be centred within `right`.  So the X position of right's font is `right.width/2 - 1`, offset from zero by `left.width`.  The previous expression took with width of the final image, and plonked the text down in the middle of it :)

I can take a look at redoing the commits with the above explanation.

gregorydlogan:
@tardyp testing if the font exists and then bailing out if it does not sounds like the best case solution, but I don't really have time to learn the Cairo API right now.  Should I file this as an issue, and then we can merge the rest of this in when the remaining bits of the PR are complete?

www/badges/buildbot_badges/templates/flat-square.svg.j2
sa2ajj: So this is a JavaScript thingie?..  Assuming it is, resolving this conversation.

PR #4224, opened 2018-07-20
Support extra state in buildbot.process.properties.renderer

p12tic:
Currently `buildbot.process.properties.renderer` supports only simple renderings that are dependent only on the properties passed to it. In more complex buildbot configurations where large part of it is generated it's sometimes needed to have a renderer that has access to full properties set of the build and also depends on some other data which could vary. Unfortunately, `Transform` does not satisfy this use case because it only allows predefined properties to be combined and does not give both properties and extra arguments to the same function. Thus the only solution is to implement custom renderer that satisfies the `IRenderable` interface and stores that extra data as part of state.

I propose that we allow supplying extra arguments to the object returned by `buildbot.process.properties.renderer` decorator that are later passed to the renderer function. For example:

```
@util.renderer
def my_test_renderer(props, extra_arg, extra_kwarg='default'):
    return '%s-%s-%s' % (props.getProperty('myprop'), extra_arg, extra_kwarg)

...

SomeStep(some_renderable_arg=my_test_renderer.withArgs('arg', extra_kwarg='kw'), ...)
```

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -909,8 +909,11 @@ class _Renderer(util.ComparableMixin, object):
 
     @defer.inlineCallbacks
     def getRenderingFor(self, props):
+        args = yield props.render(self.args)
+        kwargs = yield props.render(self.kwargs)
+
         # We allow the renderer fn to return a renderable for convenience
-        result = yield self.fn(props, *self.args, **self.kwargs)
+        result = yield self.fn(props, *args, **kwargs)
         result = yield props.render(result)
         defer.returnValue(result)
 
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -907,11 +907,12 @@ class _Renderer(util.ComparableMixin, object):
         new_renderer.kwargs.update(kwargs)
         return new_renderer
 
+    @defer.inlineCallbacks
     def getRenderingFor(self, props):
         # We allow the renderer fn to return a renderable for convenience
-        d = defer.maybeDeferred(self.fn, props, *self.args, **self.kwargs)
-        d.addCallback(props.render)
-        return d
+        result = yield self.fn(props, *self.args, **self.kwargs)
+        result = yield props.render(result)
+        defer.returnValue(result)
 
     def __repr__(self):
         if self.args or self.kwargs:
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -897,14 +897,26 @@ class _Renderer(util.ComparableMixin, object):
 
     def __init__(self, fn):
         self.fn = fn
+        self.args = []
+        self.kwargs = {}
+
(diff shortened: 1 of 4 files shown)


p12tic:
Looks like pywin is not install for some reason on Appveyor CI server.

master/buildbot/process/properties.py
```diff
@@ -897,14 +897,30 @@ class _Renderer(util.ComparableMixin, object):
...
+    @defer.inlineCallbacks
     def getRenderingFor(self, props):
+        args = yield props.render(self.args)
+        kwargs = yield props.render(self.kwargs)
+
         # We allow the renderer fn to return a renderable for convenience
-        d = defer.maybeDeferred(self.fn, props)
-        d.addCallback(props.render)
-        return d
+        result = yield defer.maybeDeferred(self.fn, props, *args, **kwargs)
```
tardyp: no need for maybeDeferred if the result is yield in inlineCallbacks

```
        result = yield self.fn(props, *args, **kwargs)
```

yield accepts both deferred or non deferred result.

master/buildbot/process/properties.py
```diff
@@ -897,14 +897,30 @@ class _Renderer(util.ComparableMixin, object):
...
+        result = yield props.render(result)
+        defer.returnValue(result)
 
     def __repr__(self):
+        if self.args or self.kwargs:
+            return 'renderer(%r, args=%s, kwargs=%s)' % (self.fn, self.args,
```
tardyp: I would rather use %r for args and kwargs

master/buildbot/test/unit/test_process_properties.py
```diff
@@ -1366,53 +1366,56 @@ def setUp(self):
...
 
+    @defer.inlineCallbacks
     def test_renderer(self):
         self.props.setProperty("x", "X", "test")
-        d = self.build.render(
-            renderer(lambda p: 'x%sx' % p.getProperty('x')))
-        d.addCallback(self.assertEqual, 'xXx')
-        return d
 
-    def test_renderer_called(self):
```
tardyp: Why did you remove this test? looks like an overlook.

master/buildbot/process/properties.py
p12tic: Fixed now, thanks!

master/buildbot/process/properties.py
p12tic: Indeed. Fixed now.

master/buildbot/test/unit/test_process_properties.py
p12tic: Yeah, the first proposal that I had in mind was making the renderable callable instead of adding the `withArgs` method. I've removed the test and forgot the add it after figuring out that the callable API would be completely unreadable and unexpected. Thanks for noticing this! Fixed now.

PR #4227, opened 2018-07-23
GitPush step

p12tic:
This is an initial PR for GitPush step for pushing git commits to remote repository. The PR is WIP, this specific implementation has not yet been tested, only the unit tests pass. I've created the PR to gather early feedback.

The PR consists of two parts: first, the guts of Git step that are reusable have been extracted to util.git.GitWorkerCommonMixin (please advise of better name and/or location if possible); then the GitPush step has been implemented on top. The GitPush step is relatively simple for now, for example, it does not support the codebase parameter.

I've put the GitPush to steps.source.git module, but I think it's a little bit confusing because until now everything in steps.source is related to source checkout, not other operations. It probably makes sense to use the opportunity think about other non-checkout related steps, e.g. pushes on other VCs or commit operation.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -589,3 +589,116 @@ class Git(Source, GitStepMixin):
             defer.returnValue("clobber")
         else:
             defer.returnValue("clone")
+
+
+class GitPush(buildstep.BuildStep, GitStepMixin, CompositeStepMixin):
+
+    description = None
+    descriptionDone = None
+    descriptionSuffix = None
+
+    ''' Class to perform Git push commands '''
+
+    name = 'gitpush'
+    renderables = ['repourl', 'branch']
+
+    def __init__(self, workdir=None, repourl=None, branch=None, force=False,
+                 env=None, timeout=20 * 60, logEnviron=True,
+                 sshPrivateKey=None, sshHostKey=None,
+                 config=None, **kwargs):
+        """
+        @type  workdir: string
+        @param workdir: local directory (relative to the Builder's root)
+                        where the tree should be placed
+
+        @type  repourl: string
+        @param repourl: the URL which points at the git repository
+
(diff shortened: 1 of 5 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -28,11 +28,12 @@
 from buildbot.process import remotecommand
 from buildbot.process.properties import Properties
 from buildbot.steps.source.base import Source
-from buildbot.util.git import GitMixin
+from buildbot.steps.worker import CompositeStepMixin
+from buildbot.util.git import GitWorkerCommonMixin
```
tardyp: Common and Mixin looks like redounded. I would say GitStepMixin

master/buildbot/util/git.py
```diff
@@ -87,3 +98,192 @@ def getSshWrapperScriptContents(keyPath, knownHostsPath=None):
 def getSshKnownHostsContents(hostKey):
     host_name = '*'
     return '{0} {1}'.format(host_name, hostKey)
+
+
+class GitWorkerCommonMixin(GitMixin):
+
+    def setupGitWorkerCommon(self):
```
tardyp: setupGitStep

tardyp:
image

tardyp:
appart from this naming, the code looks good to me.

master/buildbot/util/git.py
p12tic: Fixed now.

master/buildbot/steps/source/git.py
p12tic: Fixed now.

p12tic:
The PR has been tested in production. The only issues observed were that sshPrivateKey does not work together with git submodules due to use of relative paths (the same issue affects regular Git step). I'll open a PR with a fix once this PR lands.

PR #4254, opened 2018-08-07
use py3.7 for travis

tardyp:
less try if we are 3.7 compatible..

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -78,6 +78,12 @@ matrix:
 
     - python: "3.6"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=trial
+    - python: "3.6"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
+
+    - python: "3.7"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=trial
+    - python: "3.7"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
 
 # Dependencies installation commands
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -84,6 +84,8 @@ matrix:
     - python: "3.7"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=trial
     - python: "3.7"
+      env: TWISTED=trunk SQLALCHEMY=latest TESTS=trial
+    - python: "3.7"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
 
 # Dependencies installation commands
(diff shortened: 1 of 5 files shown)


yan12125:
Python 3.7 requires git-master of Twisted - even 18.7.0 is not compatible yet.

tardyp:
@yan12125 thanks for info. I integrated your patches temporary for testing.
Not sure if https://github.com/python/cpython/pull/1650 is going for 3.7.1 in that case, probably we want to wait and avoir if python==3.7

tardyp:
@yan12125 what issue do you have with twisted 18.7.0. Here, it looks like the only issue remaining is the one resolved in #4246

yan12125:
> Not sure if python/cpython#1650 is going for 3.7.1 in that case, probably we want to wait and avoid adding 'if python==3.7'

As stated by the Python developer, it's in 3.7.0, so exception-related patches are needed.

> @yan12125 what issue do you have with twisted 18.7.0? Here, it looks like the only issue remaining is the one resolved in #4246

There are some commits fixing 3.7 incompatibilities after the twisted-18.7.0 tag [1]. Looks like those issues do not affect buildbot. I'm just worried about potential breakages.

[1] https://github.com/twisted/twisted/compare/release-18.7.0-9475...trunk

yan12125:
Thanks! Tested this patch locally - no uncaught Python warnings.

.bbtravis.yml
```diff
@@ -78,6 +78,14 @@ matrix:
 
     - python: "3.6"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=trial
+    - python: "3.6"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
+
+    - python: "3.7"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=trial
```
cclauss:
```
    dist: xenial    # required for Python 3.7 (travis-ci/travis-ci#9069)
    sudo: required  # required for Python 3.7 (travis-ci/travis-ci#9069)
```

.bbtravis.yml
```diff
@@ -78,6 +78,14 @@ matrix:
...
+    - python: "3.7"
+      env: TWISTED=trunk SQLALCHEMY=latest TESTS=trial
```
cclauss: Also here...

.bbtravis.yml
```diff
@@ -78,6 +78,14 @@ matrix:
...
+    - python: "3.7"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
```
cclauss: Also here.

tardyp:
@cclauss thanks for hint, we use buildbot_travis for our CI, so those are .bbtravis.yml

PR #4264, opened 2018-08-17
WIP: Add pywin32 to requirements-ci.txt on Windows

p12tic:
This may fix python3.6 test failures on Appveyor.

See https://ci.appveyor.com/project/djmitche/buildbot/build/1.0.6934.

### DIFF
--- a/requirements-ci.txt
+++ b/requirements-ci.txt
@@ -62,6 +62,7 @@ pyOpenSSL==18.0.0
 pyparsing==2.2.0
 python-dateutil==2.7.3
 pytz==2018.5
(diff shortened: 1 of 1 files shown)


tardyp:
pywin32 is installed manually in the appveyor.yml. From the log, we can see that it is indeed installed even on py3.6

https://github.com/buildbot/buildbot/blob/master/appveyor.yml#L19

Maybe we should release the constraint < 222

tardyp:
see https://github.com/buildbot/buildbot/pull/4266

p12tic:
Thanks!

PR #4268, opened 2018-08-18
git: Use absolute paths for ssh data so that git submodules work

p12tic:
Turns out git submodule commands are executed from different directories than the git repository root and thus relative paths to any data given to original git command no longer work. This PR fixes that by using absolute paths.

Tested with both Git and GitPush step.

## Contributor Checklist:

* [not necessary] I have updated the appropriate documentation


terencehonles:
I created an issue #4316, but it doesn't seem like github likes linking to PRs based on #s in the title.

As mentioned in #4316: This does not work correctly because the buildbot worker may not have the same directory structure as the buildbot master. I'll look at putting a patch in place to resolve this on the worker side.

PR #4269, opened 2018-08-18
Enable max-line-length warning in flake8

p12tic:
Enabling max-line-length flake8 warning as discussed on IRC. Starting with 140 chars in this PR.

I suggest eventually targeting PEP-8 limit of 80 characters or the default of pylint which is 100 characters. Personally I prefer the former because on 1920x1080 monitors it's not possible to fit 2 windows with 100 columns each (plus sidebars and line numbers) side by side without sacrificing font readability.

Also updating pylint spellcheck ignore list because it started failing on Ubuntu 18.04 again.


tardyp:
Yeah I am not too found of max line length to be 80 chars, as it often results to complicated code layout in my experience. 100-110 looks better to me, and I have a large screen that can easily fit 250 chars :)

I admit that this is subjective opinion from me and previous maintainer, and I am willing to reduce line width progressively

PR #4276, opened 2018-08-26
Fix IRC output for Python 3 master

zware:
Twisted takes care of any necessary encoding; doing the encoding in
buildbot when running on Python 3 causes the message to show up at the
server as a bytes object repr.

This is another shot at #3961, this time with tests fixed and newsfragment added.

## Contributor Checklist:

* [x] ~~I have updated the appropriate documentation~~

### DIFF
--- a/master/buildbot/newsfragments/python3-irc-bytes.bugfix
+++ b/master/buildbot/newsfragments/python3-irc-bytes.bugfix
@@ -1,0 +1,3 @@
+:py:class:`buildbot.reporters.irc.IrcStatusBot` no longer encodes messages
+before passing them on to methods of its Twisted base class to avoid posting
+the ``repr()`` of a bytes object when running on Python 3.
--- a/master/buildbot/test/unit/test_reporters_irc.py
+++ b/master/buildbot/test/unit/test_reporters_irc.py
@@ -53,6 +53,14 @@ class TestIrcStatusBot(unittest.TestCase):
             args = ('nick', 'pass', ['#ch'], [], [], {})
         return irc.IrcStatusBot(*args, **kwargs)
 
+    def test_groupDescribe(self):
+        b = self.makeBot()
+        b.describe = lambda d, m: evts.append(('n', d, m))
+
+        evts = []
+        b.groupDescribe('#chan', 'hi')
+        self.assertEqual(evts, [('n', '#chan', 'hi')])
+
     def test_groupChat(self):
         b = self.makeBot()
         b.notice = lambda d, m: evts.append(('n', d, m))
--- a/master/buildbot/test/unit/test_reporters_irc.py
+++ b/master/buildbot/test/unit/test_reporters_irc.py
@@ -59,7 +59,7 @@ class TestIrcStatusBot(unittest.TestCase):
 
         evts = []
         b.groupChat('#chan', 'hi')
-        self.assertEqual(evts, [('n', '#chan', b'hi')])
+        self.assertEqual(evts, [('n', '#chan', 'hi')])
 
     def test_chat(self):
         b = self.makeBot()
@@ -67,7 +67,7 @@ class TestIrcStatusBot(unittest.TestCase):
 
         evts = []
         b.chat('nick', 'hi')
-        self.assertEqual(evts, [('m', 'nick', b'hi')])
+        self.assertEqual(evts, [('m', 'nick', 'hi')])
 
     def test_getContact(self):
         b = self.makeBot()
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -60,13 +60,13 @@ class IrcStatusBot(StatusBot, irc.IRCClient):
 
     # The following methods are called when we write something.
     def groupChat(self, channel, message):
-        self.notice(channel, message.encode('utf-8', 'replace'))
+        self.notice(channel, message)
 
     def chat(self, user, message):
-        self.msg(user, message.encode('utf-8', 'replace'))
+        self.msg(user, message)
 
     def groupDescribe(self, channel, action):
-        self.describe(channel, action.encode('utf-8', 'replace'))
+        self.describe(channel, action)
 
     def getContact(self, user=None, channel=None):
         # nicknames and channel names are case insensitive


zware:
It does not appear to on my local machine.

zware:
@rodrigc As far as I can tell, that's exactly the change this is making.  The `repr()` is happening within Twisted, because we're passing it `bytes` instead of `str`.

tardyp:
If the api wants str, then I guess @zware patch does the right thing.

zware:
To clarify, it's not `repr()` that's called, but rather `str()` is called by `'some_string %s' % message_passed_by_buildbot_which_is_a_bytes` within Twisted: see for example https://github.com/twisted/twisted/blob/trunk/src/twisted/words/protocols/irc.py#L500.

My newsfragment message says `repr()` because that's effectively what it is (`str(b'test') == repr(b'test')`).

I'll also note that we've been running this patch on buildbot.python.org for months with no issues, the only problem is remembering to reapply it after updating :)

rodrigc:
@tardyp no it doesn't, the patch calls `.encode()` which passes bytes

zware:
@rodrigc I'm afraid you're reading the patch backwards: this patch *removes* the `.encode()` calls :)

rodrigc:
OK, wait a sec, for some reason, I saw a backwards version of this patch where encode was added.

zware:
What's the preferred method for re-running CI checks here?  As far as I can tell, none of the failures are the fault of the most recent commit (which was designed to appease the CodeCov check).

master/buildbot/newsfragments/python3-irc-bytes.bugfix
```diff
@@ -0,0 +1,3 @@
+:py:class:`buildbot.reporters.irc.IrcStatusBot` no longer encodes messages
```
sa2ajj: I believe it's the best to follow one-sentence-per-line in every bit of the documentation

sa2ajj: ReST is spelled like that

master/buildbot/test/unit/test_reporters_irc.py
```diff
@@ -53,6 +53,14 @@ def makeBot(self, *args, **kwargs):
             args = ('nick', 'pass', ['#ch'], [], [], {})
         return irc.IrcStatusBot(*args, **kwargs)
 
+    def test_groupDescribe(self):
+        b = self.makeBot()
+        b.describe = lambda d, m: evts.append(('n', d, m))
+
+        evts = []
```
sa2ajj: should `evts` be define *before* it's used?  (as a side note: what prevents calling it `events`?)

master/buildbot/test/unit/test_reporters_irc.py
```diff
@@ -53,6 +53,14 @@ def makeBot(self, *args, **kwargs):
             args = ('nick', 'pass', ['#ch'], [], [], {})
         return irc.IrcStatusBot(*args, **kwargs)
 
+    def test_groupDescribe(self):
+        b = self.makeBot()
```
sa2ajj: `b` is not really descriptive and, while I understand it stands for `bot`, I do not see why it could not be named so...

master/buildbot/test/unit/test_reporters_irc.py
```diff
@@ -53,6 +53,14 @@ def makeBot(self, *args, **kwargs):
...
+        b.describe = lambda d, m: evts.append(('n', d, m))
```
sa2ajj: what `d` and `m` are?

master/buildbot/test/unit/test_reporters_irc.py
zware: This added test is a direct copy of the two following tests, with minimal changes as required to test the different method.

master/buildbot/newsfragments/python3-irc-bytes.bugfix
zware: I don't understand these comments:

> I believe it's the best to follow one-sentence-per-line in every bit of the documentation

Sphinx takes care of line wrapping for output, and input is much more readable with sane linebreaks.

> ReST is spelled like that

I don't see where this should apply.

sa2ajj: (I cannot see now why ReST comment was made, sorry)

As for one-sentence-per-line, it's easier during the reviews.  I believe some insights could be seen in http://rhodesmill.org/brandon/2012/one-sentence-per-line/

master/buildbot/test/unit/test_reporters_irc.py
sa2ajj: I see no reason not to improve things.  Blind copy-n-paste is not really helping.  If I have added this test, I'd also add a "cosmetic changes" commit to make things more readable.  It's up to you and other reviewers though...

tardyp: I think it is better to keep consitency within the file, and add style improvments on other PR. This helps the code style and help stuff getting in faster, IMHO. 
@p12tic and @mokibit are currently doing such excelent work in their latest submitted PRs

master/buildbot/test/unit/test_reporters_irc.py
rodrigc: I think it is unfair to penalize @zware, when he is just trying to keep the existing style of the file.

master/buildbot/test/unit/test_reporters_irc.py
rodrigc: `d` is for destination such as an IRC user or channel, `m` is for message.
It's not the best variable names, but I don't think it is fair to penalize @zware for this, because this convention is followed all over the place in this file.

master/buildbot/newsfragments/python3-irc-bytes.bugfix
zware: Thanks for an interesting read; I'll try to keep that in mind in future.  However I don't think it makes much sense here as this is a short-lived file and the simpler diffs argument doesn't really apply when it's all additions no matter where the linebreaks are.

If flaky CI forces me to push again, I will consider additional linebreaks here as a simple change to allow a fresh push :)

master/buildbot/test/unit/test_reporters_irc.py
sa2ajj: Ok.

PR #4291, opened 2018-09-01
Move from yarn registry to the npmjs one

yan12125:
registry.yarnpkg.com is going to be deprecated [1]

This PR attempts to address recent CI failures. For example, in [2], the build fails at getaddrinfo registry.yarnpkg.com. The patch is generated with the following command:

```
sed -i 's#registry.yarnpkg.com#registry.npmjs.org#' www/*/yarn.lock
```

[1] https://github.com/yarnpkg/yarn/issues/5891
[2] https://nine.buildbot.net/#/builders/9/builds/2767

## Contributor Checklist:

* [x] I have updated the unit tests (fixing currently broken tests)
* [ ] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory) - not necessary IMO as that's an internal change
* [ ] I have updated the appropriate documentation - not necessary IMO as that's an internal change


rodrigc:
Today's yarnpkg.com outage seems to be a known problem due to DNS: https://twitter.com/yarnpkg/status/1035896058616991744

PR #4296, opened 2018-09-03
Create Dockerfile.py3 files for creating python3-based buildbot docker images

rodrigc:


tardyp:
Frankly I am not sure it is worth to maintain the four dockerfile.

py2 is dead, lets move our Dockerfiles to py3 we already have a lot of users on python3 (about 25%).
We have about 32% of our install base that is from those dockerfiles, so this changing default will put python3 base to >50%.

image

tardyp:
lets merge this, and we will make it the default for subsequent release

tardyp:
created v1.4.0-py3 tag on docker hub
https://hub.docker.com/r/buildbot/buildbot-master/builds/bu2sqjutbjsbljqh5qakbdf/

https://hub.docker.com/r/buildbot/buildbot-worker/builds/bzmbew2mypv4zezjo475xml/

PR #4298, opened 2018-09-04
Server side redirect on main page is added.

ANtlord:
## Contributor Checklist:


ANtlord:
Please consider new relations. I'm not sure that it's good place for class `InternalResource`. Hope you have @tardyp an opinion about that. Thanks

master/buildbot/www/resource.py
```diff
@@ -117,3 +119,19 @@ def render(self, request):
         redir = self.base_url + self.basepath
         request.redirect(redir)
         return redir
+
+
+class InternalResource(Resource):
+    """
+    Resource available only for authorized users if authorization is enabled.
+    """
+    def render(self, request):
```
tardyp: I don't think it is needed to factorize an InternalResource if this is not needed. (principle of Minimum Viable Product)

master/buildbot/www/resource.py
```diff
@@ -117,3 +119,19 @@ def render(self, request):
...
+        from .oauth2 import OAuth2Auth
```
tardyp: avoid this kind of import

master/buildbot/www/resource.py
```diff
@@ -117,3 +119,19 @@ def render(self, request):
...
+        wwwconfig = self.master.www
+        userinfos = wwwconfig.getUserInfos(request)
+        auth = wwwconfig.auth
+        if 'anonymous' in userinfos and isinstance(auth, OAuth2Auth) and auth.autologin:
```
tardyp: I would rather say:
```
 if 'anonymous' in userinfos and auth and auth.autologin:
     return auth.do_autologin(self, request)
```
and then implement the details of the autologin in the auth plugin.

master/buildbot/www/resource.py
tardyp: so you can just implment this in the render function of IndexResource

tardyp:
any luck on finishing this? 90% of the PR not finished after 3 days never finish :)

ANtlord:
@tardyp I'm of the those who can't sleep with unfinished deals

tardyp:
looks like the tests shows some corner cases need to be taken care of.
We also need a new test for this, and a newsfragment. 
http://docs.buildbot.net/1.4.0/developer/quickstart.html for how to run tests locally

ANtlord:
Sorry for the delay. I don't forget about this PR. I'll take a time for it ASAP.

ANtlord:
@tardyp There is an error related to condition checks anonymous and check `auth` object. The text the error has is `exceptions.AttributeError: 'NoAuth' object has no attribute 'autologin'`. It would be avoided If I remain my check of class of an object. If it shouldn't be in the code still I have no idea how to prevent the error. Here is a link to the build https://nine.buildbot.net/#/builders/8/builds/2866. Checkout it please. Any idea?

tardyp:
@ANtlord I would rather add the attribute autologin=False by default in the BaseAuth class.

ANtlord:
> @ANtlord I would rather add the attribute autologin=False by default in the BaseAuth class.

Are you sure? I can but it would change the entire interface

tardyp:
> @ANtlord I would rather add the attribute autologin=False by default in the BaseAuth class.

> Are you sure? I can but it would change the entire interface

yes, by modifying config.py you are already changing the interface.

ANtlord:
@tardyp Hello! Would you mind to look at the [output](https://nine.buildbot.net/#/builders/10/builds/3036/steps/7/logs/stdio)? I'm not sure how could I affect that.

ANtlord:
@tardyp It doesn't do. `bb` build fails and I don't understand why, do you? I've attached link to the error in the previous comment. Look at there please.

p12tic:
Sorry for chiming in this late into the PR. Would a new test make sense? There are similar tests in test/unit/test_www_oauth.py, shouldn't be hard to add..

p12tic:
@ANtlord The failure was caused by old Chrome being installed on the test images. This shouldn't be a problem anymore on latest master.

p12tic:
Also, can we have at least a little documentation and a release note? I would have added the unit test myself, but I was unsure what exact problem this PR is solving.

PR #4300, opened 2018-09-04
Fixes #4299 `headerRegex` should be a string pattern, as the `header` is

### LINKED ISSUE #4299: Authentication error due to string/bytes confusion
Hi,

I use buildbot in a Docker swarm setup.

I use the buildbot-master image provided on dockerhub, in which I just add my config files and some ssh keys.
I use an NGinx container as a reverse proxy to provide HTTPS access and LDAP authentication. The buildbot-master configuration uses `util.RemoteUserAuth()` and the `REMOTEUSER` header.

This works great with the buildbot-master image v1.2.0

When v1.4.0 got out, I decided to give it a shot and noticed the image had switched to python3, nice!

I fixed my configuration files for Python3 and the server started as expected.

But when I tried to connect to the WebUI, I got this error:

```
 2018-09-04 10:01:09+0000 [-] Loading /var/lib/buildbot/buildbot.tac...
 2018-09-04 10:01:10+0000 [-] Loaded.
 2018-09-04 10:01:10+0000 [-] twistd 18.7.0 (/usr/bin/python3.6 3.6.5) starting up.
 2018-09-04 10:01:10+0000 [-] reactor class: twisted.internet.epollreactor.EPollReactor.
 2018-09-04 10:01:10+0000 [-] Starting BuildMaster -- buildbot.version: 2018.09.03
 2018-09-04 10:01:10+0000 [-] Loading configuration from '/var/lib/buildbot/master.cfg'
 2018-09-04 10:01:10+0000 [-] Creating Prometheus metrics
 2018-09-04 10:01:10+0000 [-] Setting up database with URL 'postgresql+psycopg2://buildbot:xxxx@db/buildbot'

[...]

 2018-09-03 14:27:59+0000 [_GenericHTTPChannelProtocol,0,10.0.4.2] While rendering resource:
 	  File "/usr/lib/python3.6/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
 	    result = g.send(result)
 	  File "/usr/lib/python3.6/site-packages/buildbot/www/config.py", line 126, in renderIndex
 	    yield self.config['auth'].maybeAutoLogin(request)
 	  File "/usr/lib/python3.6/site-packages/twisted/internet/defer.py", line 1613, in unwindGenerator
 	  File "/usr/lib/python3.6/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
 	  File "/usr/lib/python3.6/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
 	  File "/usr/lib/python3.6/site-packages/buildbot/www/resource.py", line 91, in failHttpError
 	    f.trap(Error)
 	  File "/usr/lib/python3.6/site-packages/twisted/python/failure.py", line 439, in trap
 	    self.raiseException()
 	  File "/usr/lib/python3.6/site-packages/twisted/python/failure.py", line 467, in raiseException
 	    raise self.value.with_traceback(self.tb)
 	  File "/usr/lib/python3.6/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
 	  File "/usr/lib/python3.6/site-packages/buildbot/www/resource.py", line 84, in failHttpRedirect
 	    f.trap(Redirect)
 	  File "/usr/lib/python3.6/site-packages/twisted/python/failure.py", line 439, in trap
 	    self.raiseException()
 	  File "/usr/lib/python3.6/site-packages/twisted/python/failure.py", line 467, in raiseException
 	    raise self.value.with_traceback(self.tb)
 	  File "/usr/lib/python3.6/site-packages/buildbot/www/config.py", line 126, in renderIndex
 	    yield self.config['auth'].maybeAutoLogin(request)
 	  File "/usr/lib/python3.6/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
 	    result = g.send(result)
 	  File "/usr/lib/python3.6/site-packages/buildbot/www/auth.py", line 123, in maybeAutoLogin
 	    res = self.headerRegex.match(header)
 	builtins.TypeError: cannot use a bytes pattern on a string-like object
 ```

 It looks like the header provided by `request.getHeader()` is a string and the `headerRegex` has been converted to a bytes pattern in 734c7c52ae7b0c1e6f1de57e49fb56813b8d5b10 which leads to this error.

```diff
diff --git a/master/buildbot/www/auth.py b/master/buildbot/www/auth.py
index 86a88bd91..4c3cb6189 100644
--- a/master/buildbot/www/auth.py
+++ b/master/buildbot/www/auth.py
@@ -104,7 +104,7 @@ class NoAuth(AuthBase):
 
 class RemoteUserAuth(AuthBase):
     header = b"REMOTE_USER"
-    headerRegex = re.compile(br"(?P<username>[^ @]+)@(?P<realm>[^ @]+)")
+    headerRegex = re.compile(r"(?P<username>[^ @]+)@(?P<realm>[^ @]+)")
 
     def __init__(self, header=None, headerRegex=None, **kwargs):
         AuthBase.__init__(self, **kwargs)
@@ -113,7 +113,7 @@ class RemoteUserAuth(AuthBase):
         if header is not None:
             self.header = header
         if headerRegex is not None:
-            self.headerRegex = re.compile(unicode2bytes(headerRegex))
+            self.headerRegex = re.compile(headerRegex)
 
     @defer.inlineCallbacks
     def maybeAutoLogin(self, request):
```

This patch seems to solve the problem but I'm not sure it is entirely correct or complete.

Also, the Dockerfile included in the v1.4.0-py3 image is not the right one, it is for python2.

ebouaziz:
I did not manage to pass all tests in `./common/validate.sh`. It does not seem related to the changes in the PR though.

## Contributor Checklist:

I don't think there is a documentation about this.

### DIFF
--- a/master/buildbot/www/auth.py
+++ b/master/buildbot/www/auth.py
@@ -111,7 +111,7 @@ class RemoteUserAuth(AuthBase):
         if self.userInfoProvider is None:
             self.userInfoProvider = UserInfoProviderBase()
         if header is not None:
-            self.header = header
+            self.header = unicode2bytes(header)
         if headerRegex is not None:
             self.headerRegex = re.compile(unicode2bytes(headerRegex))
 


ebouaziz:
Sorry I'm not that familiar with the github workflow, I should probably have made a branch for this.

master/buildbot/www/auth.py
```diff
@@ -113,17 +113,20 @@ def __init__(self, header=None, headerRegex=None, **kwargs):
...
+            self.headerRegex = re.compile(headerRegex)
 
     @defer.inlineCallbacks
     def maybeAutoLogin(self, request):
         header = request.getHeader(self.header)
         if header is None:
-            raise Error(403, b"missing http header " + self.header + b". Check your reverse proxy config!")
+            err = "missing http header " + self.header + \
+                  ". Check your reverse proxy config!"
+            raise Error(403, err.encode('utf-8'))
```
rodrigc: `unicode2bytes()` helper function should be used instead of bare `.encode('utf-8')`

master/buildbot/www/auth.py
```diff
@@ -113,17 +113,20 @@ def __init__(self, header=None, headerRegex=None, **kwargs):
         if header is not None:
             self.header = header
         if headerRegex is not None:
-            self.headerRegex = re.compile(unicode2bytes(headerRegex))
+            self.headerRegex = re.compile(headerRegex)
 
     @defer.inlineCallbacks
     def maybeAutoLogin(self, request):
         header = request.getHeader(self.header)
```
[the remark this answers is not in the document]
ebouaziz: Indeed, I did not change the type of `self.header` that is passed to `request.getHeader()`, it is still `bytes`.

master/buildbot/www/auth.py
ebouaziz: You're right.

I made a lot of mistakes, I wish I could run the tests on my machine.

master/buildbot/www/auth.py
```diff
@@ -113,17 +113,19 @@ def __init__(self, header=None, headerRegex=None, **kwargs):
...
-            raise Error(403, b"missing http header " + self.header + b". Check your reverse proxy config!")
+            raise Error(403, b"missing http header " + self.header +
+                             b". Check your reverse proxy config!")
         res = self.headerRegex.match(header)
         if res is None:
-            raise Error(
-                403, b'http header does not match regex! "' + header + b'" not matching ' + self.headerRegex.pattern)
+            err = 'http header does not match regex! "' + header + \
+                  '" not matching ' + self.headerRegex.pattern
+            raise Error(403, unicode2bytes(err))
```
rodrigc: This change is still wrong and unnecessary.  You should just revert this entire line.  `header` and `headerRegex` need to be bytes

master/buildbot/www/auth.py
```diff
@@ -113,17 +113,19 @@ def __init__(self, header=None, headerRegex=None, **kwargs):
         if header is not None:
             self.header = header
```
rodrigc: It might be worth change this line to:
       `self.header = unicode2bytes(header)`

master/buildbot/www/auth.py
```diff
@@ -104,7 +104,7 @@ class NoAuth(AuthBase):
 
 class RemoteUserAuth(AuthBase):
     header = b"REMOTE_USER"
-    headerRegex = re.compile(br"(?P<username>[^ @]+)@(?P<realm>[^ @]+)")
+    headerRegex = re.compile(r"(?P<username>[^ @]+)@(?P<realm>[^ @]+)")
```
rodrigc: This change needs to be reverted.

master/buildbot/www/auth.py
ebouaziz: Damn, you're right, I don't know why I went this way, this is all a mess. It's way simplier to convert the header from string to bytes from the start and change nothing else. I'll start over!

master/buildbot/test/unit/test_www_auth.py
```diff
@@ -113,7 +113,7 @@ def setUp(self):
 
     @defer.inlineCallbacks
     def test_maybeAutoLogin(self):
-        self.request.input_headers[b'HDR'] = b'rachel@foo.com'
+        self.request.input_headers[b'HDR'] = 'rachel@foo.com'
```
tardyp: I don't think this change is correct, headers are dict of bytes indexed by bytes

tardyp: @rodrigc what do you think? headers are dict of bytes of bytes, or dict of bytes of strings?

rodrigc: @tardyp I think all the changes to `test_www_auth.py` should be reverted.

master/buildbot/www/auth.py
```diff
@@ -111,13 +111,13 @@ def __init__(self, header=None, headerRegex=None, **kwargs):
...
         if header is not None:
-            self.header = header
+            self.header = unicode2bytes(header)
         if headerRegex is not None:
             self.headerRegex = re.compile(unicode2bytes(headerRegex))
 
     @defer.inlineCallbacks
     def maybeAutoLogin(self, request):
-        header = request.getHeader(self.header)
+        header = unicode2bytes(request.getHeader(self.header))
```
rodrigc: request.getHeader() returns bytes, this change should be reverted

master/buildbot/www/auth.py
```diff
@@ -111,13 +111,13 @@ def __init__(self, header=None, headerRegex=None, **kwargs):
         if self.userInfoProvider is None:
             self.userInfoProvider = UserInfoProviderBase()
         if header is not None:
-            self.header = header
+            self.header = unicode2bytes(header)
```
rodrigc: This line is the only line worth keeping from this patch

master/buildbot/www/auth.py
ebouaziz: I'm sorry but the backtrace in #4299 tells us otherwise:
```
 	File "/usr/lib/python3.6/site-packages/buildbot/www/auth.py", line 123, in maybeAutoLogin
 	    res = self.headerRegex.match(header)
 	builtins.TypeError: cannot use a bytes pattern on a string-like object
```

ebouaziz: Now, if `request.getHeader()` should absolutly not return a string (which it does according to this backtrace), maybe there is a bug elsewhere to discover.

rodrigc:
`getHeader()` is documented to return `bytes` https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http.py#L970 ,
so we should go with that and fix the place which is returning `str` to be consistent in the buildbot code.

ebouaziz:
Oh it makes sense now! You're right. I'll deliver the patch for L114 and report to them then.

ebouaziz:
Hello,

I digged a little more into Twisted.

While `Request.getHeader()` is documented to return `bytes`, it uses `self.requestHeaders.getRawHeaders(key)`.

As you can see at https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http_headers.py#L226 and https://github.com/twisted/twisted/blob/trunk/src/twisted/web/http_headers.py#L29, it will return the same type as the type of the name parameter it is given.

And I did instanciate the `RemoteUserAuth` object with `header='REMOTEUSER'` which is a string and remained a string because of line 114, which is why `Request.getHeader()` returned a string although it is documented to take and return only bytes.

So in the end, @rodrigc was right and line 114 is the only line worth keeping.

Sorry I did not dig deep enough to begin with, I'll update the PR now.

tardyp:
@ebouaziz good findings! So we need to enforce RemoteUserAuth.header is a byte string

ebouaziz:
@tardyp yes and that is what the `self.header = unicode2bytes(header)` does (as advised by @rodrigc in the begining).

Is it worth to put a newsfragment for this?

PR #4332, opened 2018-09-18
Git step documentation update

sa2ajj:
## Contributor Checklist:

----

I finally found an internal project to use the latest Buildbot and while the only important change is about being able to use tags, I decided to polish the step's documentation.

Requests for more documentation updates are more than welcome :smile:

### DIFF
--- a/master/buildbot/newsfragments/git-source-step-update.doc
+++ b/master/buildbot/newsfragments/git-source-step-update.doc
@@ -1,0 +1,9 @@
+Update Git source checkout build step documentation:
+
+* remove redundant empty lines
+* make the indentation consistent
+* make the code example explicitly Python snippet
+* make the SSH related parameter documentation compliant with one-sentence-per-line policy
+* re-format require/optional status and default values for the Git source checkout step
+* fix the Git spelling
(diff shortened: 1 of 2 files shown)


master/buildbot/newsfragments/git-source-step-update.doc
```diff
@@ -0,0 +1,9 @@
+Update Git source checkout build step documentation:
+
+* remove redundant empty lines
```
tardyp: I guess the newsfragment should be reasonably concise :)

sa2ajj: would you like me to make it one liner?  (do we have a policy about this somewhere?)

tardyp: There is a readme file in this directory, which explains the policy. But yes, this is what is made into the changelog. So it should be concise and contain information useful for a changelog.

sa2ajj:
I'm sorry, I did not provide a fix for the news item :(

tardyp:
no problem :) I often fix news fragment. Github makes it very easy

PR #4339, opened 2018-09-25
WIP: make urlText renderable for the FileUpload build step

sa2ajj:
## Contributor Checklist:

* [?] I have updated the unit tests
* [?] I have updated the appropriate documentation
-----
@tardyp do you think we'd need a test case for this change and/or documentation: the former would rely on the renderables test suite and the latter would stand out from the other parameter documentation for this build step.


sa2ajj:
/me . o O (**now** I read that `README.txt`)

Issue #4326, opened 2018-09-14
Builders page doesn't list workers in multi-master configuration


aj062:
If two masters have exactly same configuration (only difference being pb port number), https://localhost:8010/#/builders doesn't list workers corresponding to the builders.

Browser console have following error (multiple times):

> Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: worker in builder.workers, Duplicate key: object:8685, Duplicate value: {"_endpoint":"workers","$accessor":{},"configured_on":[{"builderid":803,"masterid":4},{"builderid":803,"masterid":5}],"connected_to":[],"graceful":false,"paused":false,"workerid":466,"workerinfo":{},"_id":466,  ...

If I set:`c['builders'] = []` in one of the master, then https://localhost:8010/#/builders page works fine. However, that master doesn't start any build anymore.

Can two masters in multi-master configuration have same c['builders'] and c['workers']?

tardyp:
Looks like a typical angularJS issue, easy to fix. Please have a look!

You created 78 issues, and 0 merge requests. It is time for your company to start giving back! Please tell that to your management. They can't rely on such huge CI system without having some developers working on it.

aj062:
> Looks like a typical angularJS issue, easy to fix. Please have a look!

I am not an angularJS expert. I tried to have a look. It seems like the ng-repeat which is having problems about duplicate workers is: https://github.com/buildbot/buildbot/blob/master/www/base/src/app/builders/builders.tpl.jade#L55

It seems to get data here:
https://github.com/buildbot/buildbot/blob/master/www/base/src/app/builders/builders.controller.coffee#L109

I printed out output of "data.getBuilders()" and it contains duplicate entries for workers. So, I wonder if the issue is in multi-master.

Do you know if this was working before (with symmetric multi-master) or a recent issue?

Is it fine to have same c['builders'] and c['workers'] in two masters?

tardyp:
The addition of worker list in builder page is relatively recent (came after I did the multimaster tests)

There is this code which is suspicious:

https://github.com/tardyp/buildbot/blob/master/www/base/src/app/common/services/datagrouper/datagrouper.service.coffee#L6

with multimaster, the configured_on attribute will contains several builderid, masterid tuples, with only masterid changing, so this is probably why the worker appear twice.

PR #4340, opened 2018-09-25
Fixed Builders page doesn't list workers in multi-master configuration

aj062:
Fixed  https://github.com/buildbot/buildbot/issues/4326: 'Builders page doesn't list workers in multi-master configuration'.


sa2ajj:
@aj062 did you check the failures?  are they related to your change?

tardyp:
I think this fix is breaking the worker page.

aj062:
I haven't seen any issue on workers page in our environment after deploying this fix. What does this test indicate?

Also, all the tests passed after I created this PR. I think they were re-run on approval, and then this test failed. This makes me think if this is an intermittent issue with the test. Can you please 'Rebuild' failed builds?

Issue #4105, opened 2018-05-17
GitHubStatusPush problems when 2 or more codebases


vladtcvs:
github.py: 136 `branch = props['branch']`

But when I add ForceScheduler with 2 codebases, I don't have global 'branch' property, I have 2 sources with 'branch' in each of them.

I didn't find some description, how GitHubStatusPush should be used in this case.

steven-johnson:
I think https://github.com/buildbot/buildbot/pull/4347 would address this issue.

PR #4347, opened 2018-10-02
Improve resiliency of GitHubStatusPush

steven-johnson:
If you attempt to use GitHubStatusPush with a builder that specifies multiple codebases, you can currently get runtime failures:
- the existing code assumes that the [0]'th entry of sourcestamps always has valid 'branch' property (it might not, if the codebase is one that is pulled without a branch specification)
- the existing code assumes that *all* entries of sourcestamps have a valid 'revision' property (same reason as above)

This revises the code to be more defensive, and skip the update attempts for sourcestamps that have no revision data (logging the skip in verbose mode).

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -126,50 +126,53 @@ class GitHubStatusPush(http.HttpStatusPushBase):
         context = yield props.render(self.context)
 
         sourcestamps = build['buildset'].get('sourcestamps')
-
-        if not sourcestamps or not sourcestamps[0]:
+        if not sourcestamps:
             return
 
-        project = sourcestamps[0]['project']
-
-        branch = props['branch']
-        m = re.search(r"refs/pull/([0-9]*)/merge", branch)
-        if m:
-            issue = m.group(1)
-        else:
-            issue = None
-
-        if "/" in project:
-            repoOwner, repoName = project.split('/')
-        else:
-            giturl = giturlparse(sourcestamps[0]['repository'])
-            repoOwner = giturl.owner
-            repoName = giturl.repo
-
-        if self.verbose:
-            log.msg("Updating github status: repoOwner={repoOwner}, repoName={repoName}".format(
-                repoOwner=repoOwner, repoName=repoName))
-
         for sourcestamp in sourcestamps:
+            project = sourcestamp['project']
+
+            # If branch is None, so is issue
+            issue = None
+            if 'branch' in props:
+                m = re.search(r"refs/pull/([0-9]*)/merge", props['branch'])
+                if m:
+                    issue = m.group(1)
+
+            repoOwner = None
+            repoName = None
+            if "/" in project:
+                repoOwner, repoName = project.split('/')
+            else:
+                giturl = giturlparse(sourcestamp['repository'])
+                if giturl:
+                    repoOwner = giturl.owner
+                    repoName = giturl.repo
+
             sha = sourcestamp['revision']
+            # If the scheduler specifies multiple codebases, don't bother updating
+            # the ones for which there is no revision
+            if not sha:
+                log.msg(
+                    'Skipped status update for codebase {codebase}, '
+                    'context "{context}", issue {issue}.'.format(
+                        codebase=sourcestamp['codebase'], issue=issue, context=context))
+                continue
+
             try:
-                repo_user = unicode2NativeString(repoOwner)
-                repo_name = unicode2NativeString(repoName)
-                sha = unicode2NativeString(sha)
-                state = unicode2NativeString(state)
-                target_url = unicode2NativeString(build['url'])
-                context = unicode2NativeString(context)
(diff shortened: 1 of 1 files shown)


tardyp:
I would prefer to get some unit tests for those use cases

steven-johnson:
Thanks. It will probably take me a few days to add the test cases.

master/buildbot/reporters/github.py
```diff
@@ -126,50 +126,53 @@ def send(self, build):
...
-        project = sourcestamps[0]['project']
+        for sourcestamp in sourcestamps:
+            project = sourcestamp['project']
 
-        branch = props['branch']
-        m = re.search(r"refs/pull/([0-9]*)/merge", branch)
-        if m:
-            issue = m.group(1)
-        else:
+            # If branch is None, so is issue
```
sa2ajj: this line probably does not belong to the change

master/buildbot/reporters/github.py
```diff
@@ -126,50 +126,53 @@ def send(self, build):
...
             issue = None
+            if 'branch' in props:
+                m = re.search(r"refs/pull/([0-9]*)/merge", props['branch'])
+                if m:
+                    issue = m.group(1)
+
+            repoOwner = None
```
sa2ajj: i understand we follow Twisted naming scheme in a lot of places, there's no need to do it here: `repo_owner` would be just fine

master/buildbot/reporters/github.py
```diff
@@ -126,50 +126,53 @@ def send(self, build):
...
+            repoName = None
```
sa2ajj: ditto (wrt `repo_name`)

master/buildbot/reporters/github.py
```diff
@@ -126,50 +126,53 @@ def send(self, build):
...
-
-        for sourcestamp in sourcestamps:
             sha = sourcestamp['revision']
+            # If the scheduler specifies multiple codebases, don't bother updating
+            # the ones for which there is no revision
+            if not sha:
+                log.msg(
+                    'Skipped status update for codebase {codebase}, '
+                    'context "{context}", issue {issue}.'.format(
+                        codebase=sourcestamp['codebase'], issue=issue, context=context))
```
sa2ajj: this line **could** be formatted as one parameter per line w/ `context=...` having a trailing `,` for easier future reviews

master/buildbot/reporters/github.py
steven-johnson: These are pre-existing names (my change just shuffled them around).

krf:
@steven-johnson I'd be interested in getting this fixed, esp. wrt #4105 -- any incentives to pick this up again?

steven-johnson:
> @steven-johnson I'd be interested in getting this fixed, esp. wrt #4105 -- any incentives to pick this up again?

I don't have the bandwidth to add the unit tests at this time, unfortunately. I'd welcome someone taking over this patch and finishing it, though.

steven-johnson:
FYI: after a long-overdue upgrade of our buildbot from 1.x to 2.8, I find that this failure has moved from "once in a while" to "pretty much every single time" -- GitHubStatusPush no longer seems to work *at all* for our setup; the log files indicate that ~every usage of it fails with some variation of

```
          File "/path/to/python/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
            result = g.send(result)
          File "/path/to/python/lib/python3.7/site-packages/buildbot/reporters/http.py", line 80, in getMoreInfoAndSen
d
            yield self.send(build)
          File "/path/to/python/lib/python3.7/site-packages/twisted/internet/defer.py", line 1613, in unwindGenerator
          File "/path/to/python/lib/python3.7/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInli
neCallbacks
          File "/path/to/python/lib/python3.7/site-packages/buildbot/reporters/http.py", line 80, in getMoreInfoAndSen
d
            yield self.send(build)
          File "/path/to/python/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
            result = g.send(result)
          File "/path/to/python/lib/python3.7/site-packages/buildbot/reporters/github.py", line 132, in send
            branch = props['branch']
          File "/path/to/python/lib/python3.7/site-packages/buildbot/process/properties.py", line 140, in __getitem__
            rv = self.properties[name][0]
        builtins.KeyError: 'branch'
```

master/buildbot/reporters/github.py
p12tic: Fixed now.

master/buildbot/reporters/github.py
p12tic: Fixed now.

master/buildbot/reporters/github.py
p12tic: Fixed now.

p12tic:
This as been merged as part of #5316, removing stale label.

PR #4350, opened 2018-10-06
reporter/github: checking the response status code

iblis17:
- and show the github response if error

----
TODO

- [x] unit test

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -161,7 +161,7 @@ class GitHubStatusPush(http.HttpStatusPushBase):
                 context = unicode2NativeString(context)
                 issue = unicode2NativeString(issue)
                 description = unicode2NativeString(description)
-                yield self.createStatus(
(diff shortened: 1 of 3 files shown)


master/buildbot/reporters/github.py
```diff
@@ -171,19 +171,24 @@ def send(self, build):
                     issue=issue,
                     description=description
                 )
+
+                assert response.code // 100 == 2
```
tardyp: Please do not use assert. assert is supposed to be disabled in production.

Also // 100 == 2 is a bit hard to read

iblis17: Which method do you prefer to check the status code? (I personally always use `//`

tardyp: I would create a helper method isStatus2XX(response)

iblis17: ha, okay.

iblis17:
I added some test cases.

Issue #4177, opened 2018-06-27
buildbot.test.unit.test_worker_libvirt.TestWorkQueue.test_work test fails under certain conditions


p12tic:
Environment:
OSX 10.12, python from MacPorts, v2.7.15, python environment from `make virtualenv`.

buildbot.test.unit.test_worker_libvirt.TestWorkQueue.test_work will fail when `trial -j8 buildbot.test.unit` is launched, but not when `trial buildbot.test.unit` or `trial -j8 buildbot` is launched.

Error output:
```
Traceback (most recent call last):
Failure: twisted.internet.defer.TimeoutError: <buildbot.test.unit.test_worker_libvirt.TestWorkQueue testMethod=test_work> (test_work) still running at 5.0 secs

buildbot.test.unit.test_worker_libvirt.TestWorkQueue.test_work
```

PR #4359, opened 2018-10-10
Fix broken tests (#4177)

p12tic:
This PR fixes libvirt tests to be able to run them in parallel mode.

Additionally, seems that ec2 tests fail on my machine due to missing credentials. This PR changes the code to skip the tests if such errors are encountered.

### DIFF
--- a/master/buildbot/test/unit/test_worker_libvirt.py
+++ b/master/buildbot/test/unit/test_worker_libvirt.py
@@ -185,9 +185,6 @@ class TestLibVirtWorker(unittest.TestCase):
 
 class TestWorkQueue(unittest.TestCase):
 
-    def setUp(self):
-        self.queue = libvirtworker.WorkQueue()
-
     def delayed_success(self):
         def work():
             d = defer.Deferred()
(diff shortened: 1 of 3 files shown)


master/buildbot/test/unit/test_worker_ec2.py
```diff
@@ -572,6 +581,10 @@ def botoSetup(self):
         except NotImplementedError:
             raise unittest.SkipTest("KeyPairs.create_key_pair not implemented"
                                     " in this version of moto, please update.")
+        except (botocore.exceptions.ClientError,
```
tardyp: I think unit tests should not require EC2 credentials. As those are fully mocked by moto we shouln't get a credential error

p12tic:
Actually, ec2 is completely non-mocked, see https://github.com/spulec/moto/issues/1793. boto3 should be pinned to 1.7.x as a workaround.

p12tic:
Updated the PR with a change that pins botocore to older version when creating virtualenv with `make virtualenv` as suggested in https://github.com/spulec/moto/issues/1793.

PR #4370, opened 2018-10-22
Show buildrequest owners in the builder page (fixes #4207, #3904)

### LINKED ISSUE #4207: Missing owner in buildrequest table on builder page
image

It seems problem is here 
https://github.com/buildbot/buildbot/blob/master/www/base/src/app/builders/builder/builder.tpl.jade#L24
but adding `` span | {{ br.owner}} `` is not enough as it seems ``.owner`` neither ``.properties`` are not set.

p12tic:
Fixes #4207, #3904.

This change is not tested by a e2e test, should I create one?

## Contributor Checklist:

### DIFF
--- a/www/base/src/app/builders/builder/builder.controller.coffee
+++ b/www/base/src/app/builders/builder/builder.controller.coffee
@@ -100,5 +100,9 @@ class Builder extends Controller
                 limit: $scope.numbuilds
                 order: '-number'
             $scope.buildrequests = builder.getBuildrequests(claimed:false)
+            $scope.buildrequests.onNew = (buildrequest) ->
+                data.getBuildsets(buildrequest.buildsetid).onNew = (buildset) ->
+                    buildset.getProperties().onNew = (properties) ->
+                        buildrequest.properties = properties
             $scope.builds.onChange=refreshContextMenu
             $scope.buildrequests.onChange=refreshContextMenu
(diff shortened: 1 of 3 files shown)


pmisik:
I just tried it and it work well to me.
Thank You @p12tic

www/base/src/app/builders/builder/builder.controller.coffee
```diff
@@ -100,5 +100,9 @@ class Builder extends Controller
                 limit: $scope.numbuilds
                 order: '-number'
             $scope.buildrequests = builder.getBuildrequests(claimed:false)
+            $scope.buildrequests.onNew = (buildrequest) ->
```
tardyp: this will generate a large number of http requests

p12tic: Yes. Hopefully there are few unclaimed buildrequests for each builder. What would be the proper way to expose this? Create a new data API endpoint or adapt the getBuildrequests one to include the needed information?

tardyp: I would rather keep the buildbot API generic. The problem is actually not so much the rest API. build rest api already has a properties parameter to fetch more properties. We could add this to buildrequest as well.
The problem is for message queues messages which does not contain the information. So new buildrequests or new build events do not contain the properties required for the page.

A generic workaround would be in the data module. If there is additional query parameter, then the data module will automatically run the rest API for that single build/element in order to fetch the additional stuff. This would be a nice a generic way to handle the problem, while it generates more rest api calls, it simplifies the code on client side, by avoiding all those onNew specific calls.

PR #4375, opened 2018-10-25
Failed ticket parsing should raise exception

lancelodliu:
## Contributor Checklist:


master/buildbot/changes/p4poller.py
```diff
@@ -201,6 +201,7 @@ def _acquireTicket(self, protocol):
     def _parseTicketPassword(self, text):
         lines = text.split("\n")
         if len(lines) < 3:
+            raise P4PollerError("Failed to parse ticket in that output is less than 3 lines: {}".format(text))
             return None
```
p12tic: Do we need this return if we raise exception immediately after?

lancelodliu: > Do we need this return if we raise exception immediately after?

No, apparently. I will fix this.

aisoard:
@lancelodliu Sorry about that conflict, I didn't see your branch.

We do have a test for ticket parsing failure now, not sure how to "expect" an exception in testing though.

p12tic:
@aisoard: Use `with self.assertRaises(P4PollerError):` block.

PR #4380, opened 2018-10-27
smokes: Use older chromedriver for old Chrome

p12tic:
Currently we always download newest chromedriver. Unfortunately it usually supports only a couple of latest releases of Chrome which is problematic as they get released each month.

To work around this we parse the chromedriver release notes, extract the chrome releases that each chromedriver release supports and use the newest for the version of Chrome on the system.

Hopefully this fixes recent e2e failures on buildbot CI. The other solution is to update Chrome there every few months which would result in forgotten updates time to time.

### DIFF
--- a/common/smokedist-download-compatible-chromedriver.py
+++ b/common/smokedist-download-compatible-chromedriver.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+from __future__ import absolute_import
+from __future__ import print_function
+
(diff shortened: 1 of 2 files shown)


tardyp:
Looks good. I have been indeed rebuilding the base metabbotcfg image every other month in order to keep chrome updated, and also to keep a recent cache of pypi packages.

PR #4381, opened 2018-10-28
do pylint with -j2

tardyp:
to try to workaround instabilities

### DIFF
--- a/master/Makefile
+++ b/master/Makefile
@@ -1,6 +1,6 @@
 # developer utilities
 pylint:
-	pylint -j4 --rcfile=../common/pylintrc buildbot
+	pylint -j2 --rcfile=../common/pylintrc buildbot
 	@test ! -f fail
 
 tutorial:


p12tic:
Could it be that there's some I/O bottleneck in the buildbot CI build farm? I get a feeling that these kinds of instabilities happen more often when more builds are running concurrently. Maybe a lock to limit the maximum number of builds per physical server could help?

tardyp:
We use hyper.sh for our testfarm. The load is normally spread over a big bunch of hosts. 

pylint is very cpu intensive, and not very much io bound. maybe this is some neighborhood noise indeed, but it is hard to tell. :-/
It is also memory intensive so we might get OOMed. with -j2 the pylint run in 5min

p12tic:
A related question: would you accept a PR that makes the number of jobs configurable? E.g. `NUM_JOBS := $(shell get_num_jobs.sh)` and then use $NUM_JOBS in the rules. For local development it's useful to utilize all CPUs and for CI we can hardcode something depending on environment variables.

tardyp:
yes. However, I think it is best for local development to setup automatic pylint upon save

for me make pylint is really for those who don't want to tweak things so it make not much sense to add NUM_JOBS.

PR #4383, opened 2018-10-28
Increase jasmine timeout

p12tic:
I think I finally figured out the cause of at least some of the instabilities of smokes tests.

Turns out that jasmine's `defaultTimeoutInterval` also defines the maximum duration of whole test. If we accidentally exceed this timeout, jasmine will start the `afterEach` callback, but also not stop execution of running test itself. The test will then fail as the page is pulled from under its feet and produce additional messages which were very confusing.

I was able to reliably reproduce this failure mode running smoke tests on a cgroup limited to 5% of CPU. There might be other issues causing instabilities still, but at least this one will be fixed.

This PR also updates yarn.lock to include the `integrity` line.

### DIFF
--- a/smokes/protractor-headless.conf.js
+++ b/smokes/protractor-headless.conf.js
@@ -21,7 +21,15 @@ exports.config = {
     framework: 'jasmine',
 
     jasmineNodeOpts: {
-        defaultTimeoutInterval: 50000,
+        // jasmine requires that whole test is completed within
+        // defaultTimeoutInterval. If we accidentally exceed this timeout,
+        // jasmine will not stop the execution of the test method, but will
+        // simply start afterEach() callback and fail the test. The test code
+        // will likely fail too as the page was pulled from under its feet.
+        // The test error messages will make the cause of the failure very
+        // confusing. Thus we increase the timeout value to one that hopefully
+        // will never be exceeded.
+        defaultTimeoutInterval: 1000000,
         print: function() {}
     },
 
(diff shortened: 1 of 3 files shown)


tardyp:
I am not sure if this is the rootcause. 
Right, If there is a timeout, then more test might fail because things are not cleanedup properly, but the testsuite would have failed anyway because of the timeout.

My guts feeling is that at some point, there is something that is actually making the system hang, and the test will never finish. 
In that case, increasing the timeout will just make the test fail after more time in this case.

We already increased that timeout several time IIRC.

p12tic:
>  Right, If there is a timeout, then more test might fail because things are not cleaned up properly, but the testsuite would have failed anyway because of the timeout.

I think that I expressed myself incorrectly. The test itself does not fail due to, e.g. 5 or 10 second timeout in some step taking too long. It's jasmine who terminates the test because it exceeded `defaultTimeoutInterval`. So even if everything is alright within the test execution, jasmine might terminate it just because it's slow.

I've verified that this is exactly what happens locally by adding a bunch of console.info(). I was able to see the test happily executing alongside the `afterEach` callback just until the effects of `afterEach` interferes with the expectations of the test.

Having said that, I agree that most instabilities are not caused by this problem right now even though I remember exactly the same problems in the past. I'd say the downsides of having a change like this are fewer than the possibility of instabilities, even if small. But feel free to close the PR.

tardyp:
ok lets try it then.

rodrigc:
I think that @p12tic's patch to increase the Jasmine timeout is a reasonable way to go.  I've done similar things to this when writing other protractor-based tests for other projects.  Thanks for this fix!

PR #4372, opened 2018-10-22
Create a "Pending Buildrequests" page that lists all unclaimed buildrequests (fixes #4239)

p12tic:
Relatively simple addition done by the approach suggested in #4239. This change is not tested by a e2e test, should I create one?

## Contributor Checklist:

* [n/a] I have updated the unit tests
* [x] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [n/a] I have updated the appropriate documentation

PR #4384, opened 2018-10-28
Add e2e test for pending build requests page

p12tic:
Improves #4372.

Also did some minimal refactoring to reduce code duplication.


tardyp:
cool!

PR #4389, opened 2018-11-01
www: Don't use external resources in dashboard smoke test

p12tic:
Probably 90% of recent random smoke test failures have been caused by the dashboard test not loading dashboard page in 11s (e.g. https://nine.buildbot.net/#/builders/10/builds/3085). Turns out that we load an external resource from github (`https://api.github.com/repos/buildbot/buildbot/stats/code_frequency`) which may take a long time to generate the first time.

While I have not proved this is the actual cause of the test instability, I did reproduce the failure the first time I ran the test locally and could not reproduce it again, which is consistent with github generating the data the first time and timing out the test and being able to serve the cached version thereafter.

This PR uses local data in the repository for the test instead of loading it dynamically. While it may not fix the instability, it still does not make sense to load external data in tests.

### DIFF
--- a/smokes/mydashboard.py
+++ b/smokes/mydashboard.py
@@ -4,7 +4,6 @@ from __future__ import print_function
 import os
 import time
 
-import requests
 from flask import Flask
 from flask import render_template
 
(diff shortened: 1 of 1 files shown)


smokes/mydashboard.py
```diff
@@ -31,18 +31,16 @@ def main():
...
 
-    # Example on how to use requests to get some info from other web servers
-    code_frequency_url = "https://api.github.com/repos/buildbot/buildbot/stats/code_frequency"
-    results = requests.get(code_frequency_url)
-    while results.status_code == 202:
-        # while github calculates statistics, it returns code 202.
-        # this is no problem, we just sleep in our thread..
-        time.sleep(500)
-        results = requests.get(code_frequency_url)
+    # try not to fetch data from external sources because they might become
```
tardyp: well, the actual goal of this code is to serve as example in the doc for dashboard.
So this comment will get in the doc, and can be hard to understand.

Can you please keep the requests example as a comment?

p12tic: Indeed, thanks for pointing that out. I didn't realize that we have a symlink there. 

I think it's better to have a separate copy in docs as the fact that we use the file in CI should not make them worse. I've made docs have a separate copy in 2f3dc6e30ac09bb9d8578158f551354f210925a8 and added a mention that they should be kept in sync in 708042200775af444195273738358a7f5de4bfaf.

tardyp: The original goal of this is to provide an example of custom python dashboard.

As this is a fair amount of code, my goal there was to make sure it is always tested by having it in the e2e tests as well. I agree with you that this shouldn't affect the test reliability

tardyp: Well, the warning comment is typically the thing I wanted to avoid. 
who reads comment on top of a file? If there is a change in the middle of the file, github won't even show the comment to the reviewers.

I would rather remove the requests stuff altogether rather than duplicate this huge code.

p12tic: OK, will implement that.

PR #4390, opened 2018-11-02
docs: Resave *_src.svg files with new version of inkscape

p12tic:
Opening *_src.svg files with recent version of Inkscape produces a warning: 

> <filename> was created in an older version of Inkscape (90 DPI) and we need to make it compatible with newer versions (96 DPI). Tell us about this file:
>
>  - This file contains digital artwork for screen display (Choose if unsure).
>  - This file is intended for physical output such as paper or 3D prints

I've selected the first option and resaved without any other changes. It's probably worth to commit the updated files to save future headache. I've regenerated the png files and confirmed that there are only minimal rasterization changes.


tardyp:
looks good!

PR #4391, opened 2018-11-02
docs: Add diagram showing the relationships between source concepts

p12tic:
Add the diagram visually documenting the relationships between source concepts to the "Source Stamps" section in the "Concepts" chapter. This documents the current situation and is not related my proposal of enhancing project support in config files and the frontend.

changes


p12tic:
By the way, would you agree to merge the Concepts chapter with Introduction. It seems that both of them talk about the same thing and duplicate each other in various ways.

tardyp:
Looks good. 
I think it is good to have a simple introduction chapter, and a more detailed Concepts chapter.

It does not hurt to rephrase the introduction chapter in concepts IMHO. learning is about repetition.

PR #4396, opened 2018-11-04
www: Print webdriver stdio to console in e2e tests

p12tic:
This PR turns on redirection of webdriver stdio to console. This way we can debug intermittent webdriver startup failures like https://nine.buildbot.net/#/builders/10/builds/3108. The option is not documented and comes from webdriver js bindings. See https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/remote/index.js where `opt_options` to `SeleniumServer` constructor is `localSeleniumStandaloneOpts` from protractor config.

### DIFF
--- a/smokes/protractor-headless.conf.js
+++ b/smokes/protractor-headless.conf.js
@@ -9,6 +9,12 @@ exports.config = {
 
     SELENIUM_PROMISE_MANAGER: false,
 
+    localSeleniumStandaloneOpts: {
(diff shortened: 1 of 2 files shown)


p12tic:
I ran the smokes tests around 15 times total on buildbot testfarm and unfortunately did not get any failures. The failure is not stupid and knows when to hide.. :-)

Issue #3027, opened 2017-03-10
[Nine] HttpPushStatus fails on CheckConfig


tardyp:
This ticket is a migrated Trac ticket [3625](http://trac.buildbot.net/ticket/3625)

People contributed to the original ticket: @unknown_contributor, @tardyp
Ticket created on: `Oct 12 2016`
Ticket last modified on: `Feb 08 2017`

---

I've been trying to add an `[[HttpPushStatus]]` service to my buildbot but there was no way I could get it to work.

Using the very simple [from buildbot.plugins import reporters
sp = reporters.[[HttpStatusPush]](serverUrl="http://example.com/submit")
c[[example shown in the doc|'services'].append(sp)]], I added the following lines to my master.cfg file:

    from buildbot.plugins import reporters
    
    c[[= [|'services']]]
    sp = reporters.[[HttpStatusPush]](serverUrl="http://localhost:16161/")
    c['services'].append(sp)

This causes the master.cfg file to fail on $ buildbot checkconfig master.cfg :

    $ buildbot checkconfig master.cfg
    :0: [[UserWarning]]: You do not have a working installation of the service_identity module: 'No module named service_identity'.  Please install it from <https://pypi.python.org/pypi/service_identity> and make sure all of its dependencies are satisfied.  Without the service_identity module and a recent enough pyOpenSSL to support it, Twisted can perform only rudimentary TLS client hostname verification.  Many valid certificate/hostname mappings may be rejected.
    error while parsing config file:
    Traceback (most recent call last):
      File "/Library/Python/2.7/site-packages/twisted/internet/defer.py", line 149, in maybeDeferred
        result = f(*args, **kw)
      File "/Library/Python/2.7/site-packages/buildbot/scripts/checkconfig.py", line 57, in checkconfig
        return _loadConfig(basedir=basedir, configFile=configFile, quiet=quiet)
      File "/Library/Python/2.7/site-packages/buildbot/scripts/checkconfig.py", line 27, in _loadConfig
        config.[[FileLoader]](basedir, configFile).loadConfig()
      File "/Library/Python/2.7/site-packages/buildbot/config.py", line 161, in loadConfig
        self.basedir, self.config[[FileName]])
    --- <exception caught here> ---
      File "/Library/Python/2.7/site-packages/buildbot/config.py", line 118, in load[[ConfigDict]]
        exec(f, localDict)
[...]

PR #4400, opened 2018-11-04
test: Add test for non-authenticated HttpStatusPush

p12tic:
Add test against bug filed in #3027.

### DIFF
--- a/master/buildbot/test/unit/test_reporter_http.py
+++ b/master/buildbot/test/unit/test_reporter_http.py
@@ -70,12 +70,18 @@ class TestHttpStatusPush(unittest.TestCase, ReporterTestMixin):
         yield self.master.startService()
 
     @defer.inlineCallbacks
-    def createReporter(self, **kwargs):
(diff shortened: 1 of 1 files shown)


p12tic:
Looks like the previous version had a genuine test failure, which I've fixed now.

master/buildbot/test/unit/test_reporter_http.py
```diff
@@ -70,12 +70,18 @@ def setUp(self):
...
             self.master,
-            "serv", auth=("username", "passwd"))
-        self.sp = sp = HttpStatusPush("serv", auth=(
-            "username", Interpolate("passwd")), **kwargs)
+            "serv", auth=auth)
+
+        interpolated_auth = None
+        if auth is not None:
+            username, passwd = auth
+            passwd = Interpolate(passwd)
```
tardyp: Interpolate is immutable, so you can put it in default args

p12tic: Non-interpolated passwd was used in getService so I couldn't reuse the interpolated value.

PR #4405, opened 2018-11-06
Disable flaky windows tests

p12tic:
Adds a onPlatform attribute to `@flaky` decorator and uses it to disable buildbot.test.integration.interop.test_interruptcommand.InterruptCommand.test_setProp on Windows.

Also, add files created during smoke tests to .gitignore.

### DIFF
--- a/master/buildbot/test/util/decorators.py
+++ b/master/buildbot/test/util/decorators.py
@@ -42,13 +42,15 @@ def todo(message):
 
 def flaky(bugNumber=None, issueNumber=None):
     def wrap(fn):
-        if not os.environ.get(_FLAKY_ENV_VAR):
-            if bugNumber is not None:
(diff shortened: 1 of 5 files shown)


master/buildbot/test/util/decorators.py
```diff
@@ -40,9 +40,13 @@ def wrap(func):
     return wrap
 
 
-def flaky(bugNumber=None, issueNumber=None):
+def flaky(bugNumber=None, issueNumber=None, onPlatform=None):
     def wrap(fn):
-        if not os.environ.get(_FLAKY_ENV_VAR):
+        match_platform = True
```
tardyp:
```suggestion
     if onPlatform is not None and sys.platform != onPlatform:
         return fn
```

p12tic: Agreed. Wanted to preserve the present style of the function.

PR #4409, opened 2018-11-07
Move upgrading docs to single chapter

p12tic:
This PR moves the upgrading docs to a single chapter at the end. Previously 0.9 upgrading information has been dispersed into 3 sections. The new chapter is moved to the end as most users will probably read it only once and then never ever need it again.

Also, this PR moves all configuration docs to a single folder. This does not change output, only the source organization.


master/docs/relnotes/index.rst
```diff
@@ -10,6 +10,161 @@ Release Notes
 
 .. towncrier release notes start
 
+Buildbot ``1.5.1-dev117`` ( ``2018-11-07`` )
```
seankelly: Should this be here? Seems like it shouldn't be filled out until the release happens.

p12tic: Fixed now.

PR #4414, opened 2018-11-08
Simplify buildrequestdistributor tests

p12tic:
This PR removes the `_quiet()` function that is called when build request distributor's activity loop finishes and instead just stores deferred to the activity loop itself. A separate function to wait on that deferred is added `_waitForFinish()` and tests can use that instead of the `_quiet` hack. This is IMO a cleaner approach.

Also, this PR add more uses of `inlineCallbacks` and adds waits for deferreds returned from functions that were not waited for before.

### DIFF
--- a/master/buildbot/process/buildrequestdistributor.py
+++ b/master/buildbot/process/buildrequestdistributor.py
@@ -325,6 +325,7 @@ class BuildRequestDistributor(service.AsyncMultiService):
         self.active = False
 
         self._pendingMSBOCalls = []
+        self._activity_loop_deferred = None
 
     @defer.inlineCallbacks
     def stopService(self):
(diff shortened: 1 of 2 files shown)


master/buildbot/process/buildrequestdistributor.py
```diff
@@ -543,6 +547,8 @@ def createBuildChooser(self, bldr, master):
         # just instantiate the build chooser requested
         return self.BuildChooser(bldr, master)
 
-    def _quiet(self):
+    @defer.inlineCallbacks
+    def _waitForFinish(self):
         # shim for tests
```
seankelly: Was the `pass` the shim or is the function itself a shim for tests? If the former, then I think this comment should be removed now that there's content in this function.

p12tic: Fixed in a8f36ecee17f8795384df48a2048c0d7d7200f6f.

Issue #4413, opened 2018-11-07
pylint sometimes ends up on wrong hyper worker


p12tic:
Looks like pylint job is configured to be run on m3 hyper instance which has 4GB, but for some reason it runs on instance with 1GB of RAM. This is what I got out of some pylint job just before it timed out. 

```top - 21:37:34 up 55 min,  0 users,  load average: 4.91, 3.20, 2.22
Tasks:  15 total,   1 running,  14 sleeping,   0 stopped,   0 zombie
%Cpu(s):  3.4 us, 95.5 sy,  0.0 ni,  0.0 id,  0.0 wa,  0.0 hi,  1.1 si,  0.0 st
KiB Mem :  1030604 total,    10692 free,   991892 used,    28020 buff/cache
KiB Swap:        0 total,        0 free,        0 used.     8056 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S %CPU %MEM     TIME+ COMMAND                                          
 1811 buildbot  20   0   90932  21624      0 D  3.2  2.1   0:29.39 python                                           
 1842 root      20   0   40408    812    272 R  1.0  0.1   0:00.44 top                                              
    4 buildbot  20   0  170908  29176      0 S  0.8  2.8   0:03.61 twistd                                           
 1823 buildbot  20   0  491100 400312      0 D  0.6 38.8   6:13.57 pylint                                           
 1824 buildbot  20   0  597340 499316      0 D  0.6 48.4   6:16.80 pylint                                           
 1815 buildbot  20   0  412776  31424      0 D  0.5  3.0   0:01.39 pylint                                           
    1 root      20   0    4420    120      0 S  0.0  0.0   0:00.00 init                                             
    3 buildbot  20   0     200      4      0 S  0.0  0.0   0:00.00 dumb-init                                        
 1803 buildbot  20   0    8424    256      4 S  0.0  0.0   0:00.00 make                                             
 1804 buildbot  20   0    4512     80      0 S  0.0  0.0   0:00.00 sh                                               
 1805 buildbot  20   0    8424    260      4 S  0.0  0.0   0:00.00 make                                             
 1809 buildbot  20   0    4512     76      0 S  0.0  0.0   0:00.00 sh                                               
 1810 buildbot  20   0   19712    232      0 S  0.0  0.0   0:00.00 bash                                             
 1812 buildbot  20   0  121324  34640      0 S  0.0  3.4   0:00.98 pylint                                           
[...]

PR #4416, opened 2018-11-10
Don't attempt to start builds that will have unsatisfied locks

p12tic:
This is the first step of fixing #4413. In order to prevent builds from being built on wrong workers we must have a reliable way of doing so. The place we currently do such checks is `Builder.canStartWithWorkerForBuilder` but it turns out that the caller disregards the rejection if there are no other available workers and proceeds regardless as a fallback. This PR fixes that by removing this fallback behavior and merging `Builder.canStartWithWorkerForBuilder` into `Builder.canStartBuild` as there's no longer a conceptual difference between the two. Bonus result is that we no longer start builds that are known to have unsatisfiable locks at the time. 

`Builder.canStartWithWorkerForBuilder` and `Builder.canStartBuild` are two ways to check whether a buildrequest may be built on specific worker for builder. The first checks the builder locks that will be
acquired by build, the second just calls config.canStartBuild().

The original intention before renderable builder locks have been added in bcf85bb was that `canStartWithWorkerForBuilder` checks properties that are independent of the buildrequest and `canStartBuild` checks properties that are dependent on the buildrequest. At the time nothing could be really buildrequest specific, thus the fallback algorithm of allowing builds past the `canStartWithWorkerForBuilder` check as fallback made sense, because we weren't exactly sure whether the build could be built or not. This allows builds with known unsatisfiable
locks to be started on busy instances of Buildbot.

Once renderable builder locks have been implemented, both `canStartWithWorkerForBuilder` and `canStartBuild` started to depend on the actual buildrequest but that was not noticed at the time. Thus there's no longer a reason to keep `canStartWithWorkerForBuilder` and `canStartBuild` separate. Combining them and getting rid of the fallback build path both avoids the unsatisfied lock issue and simplifies the implementation.

The PR also makes `Build.canStartWithWorkerForBuilder` private as it leaks internal implementation detail.

No fallback has been provided for the removed functions as it's not part of publicly documented interface and keeping them unchanged would prevent us from moving forward.


seankelly:
>  Builder.canStartWithWorkerForBuilder method has been removed. Use Builder.canStartBuild. 
Can that be done? Is that method considered part of the API and thus semver's guarantees or not?

p12tic:
> Can that be done? Is that method considered part of the API and thus semver's guarantees or not?

The problem is that not changing it will constrain us a lot, even in this case specifically. The old canStartWithWorkerForBuilder did not have buildrequests attribute and we still have an API break, because canStartBuild requires it. I agree that if we do think there are users that use that API then it makes sense to provide some fallback.

Please correct me if I'm wrong, but it seems to me that 90% of users don't hack buildbot core internals. Would it make sense to start telling that all APIs are not supposed to be used if they are not documented and then make APIs public by users request? Otherwise any code that is being written is part of public API and that means we can't easily change it.

seankelly:
My question is more, do we have a defined API and if not, which I believe is the case, we need to start defining it so things like this aren't going to cause a 2.0. Alternatively, assume the worse and release 2.0 anyway. I doubt many people are calling this function so the impact would be minimal.

master/buildbot/process/builder.py
```diff
@@ -279,38 +279,31 @@ def getAvailableWorkers(self):
...
-        if buildrequests is None:
-            warnings.warn(
-                "Not passing corresponding buildrequests to "
-                "Builder.canStartWithWorkerForBuilder is deprecated")
-
         locks = [(self.botmaster.getLockFromLockAccess(access), access)
                  for access in locks]
-        can_start = Build.canStartWithWorkerForBuilder(locks, workerforbuilder)
-        defer.returnValue(can_start)
-    deprecatedWorkerClassMethod(locals(), canStartWithWorkerForBuilder,
```
tardyp: this one you can remove.

master/buildbot/test/unit/test_process_builder.py
```diff
@@ -215,55 +213,128 @@ def test_getCollapseRequestsFn_builder_False(self):
...
+        yield self.makeBuilder()
+
+        self.bldr.botmaster.getLockFromLockAccess = \
+            mock.Mock(return_value='dummy')
+
+        with mock.patch(
+                'buildbot.process.build.Build._canAcquireLocks',
+                mock.Mock(return_value=False)):
+            startable = yield self.bldr.canStartBuild('worker', 100)
+            self.assertEqual(True, startable)
```
tardyp: why it is reversed?

I hope you're not one of those who prefer to write

```c++
if (0 == value){
/* "Yoda conditionals" ("If zero the value is, execute this code you must!") */
}
```

master/buildbot/newsfragments/build-canStartWithWorkerForBuilder.removal
```diff
@@ -0,0 +1 @@
+Build.canStartWithWorkerForBuilder static method has been made private.
```
tardyp:
```suggestion
Build.canStartWithWorkerForBuilder static method has been made private and renamed _canAcquireLocks
```

master/buildbot/process/builder.py
```diff
@@ -279,38 +279,31 @@ def getAvailableWorkers(self):
     deprecatedWorkerClassMethod(locals(), getAvailableWorkers)
 
     @defer.inlineCallbacks
-    def canStartWithWorkerForBuilder(self, workerforbuilder, buildrequests=None):
+    def canStartBuild(self, workerforbuilder, buildrequest):
```
tardyp: Builder cannot be overriden or customized unless via patching, so I guess that one is okay

master/buildbot/newsfragments/build-canStartWithWorkerForBuilder.removal
tardyp: Build class can be changed via BuilderConfig's factory, so it is not cool to deprecate it.
I admit this would be a very deep hack, and most probably this change won't impact anybody

master/buildbot/test/unit/test_process_builder.py
p12tic: Agreed. I've just used C# testing frameworks a lot and there the comparison order is `expected`, `actual`.

p12tic: Fixed now.

master/buildbot/newsfragments/build-canStartWithWorkerForBuilder.removal
p12tic: Fixed now.

p12tic:
@tardyp: Is there anything from what you requested in the review not fixed? Or did you actually want to leave `Build.canStartWithWorkerForBuilder` as is?

PR #4418, opened 2018-11-11
Improve documentation of abstract latent worker

p12tic:
Improves the formatting to follow other classes.

### DIFF
--- a/master/docs/manual/customization.rst
+++ b/master/docs/manual/customization.rst
@@ -539,25 +539,33 @@ Aside from the service methods, the other concerns in the previous section apply
 Writing a New Latent Worker Implementation
 ------------------------------------------
 
-Writing a new latent worker should only require subclassing :class:`buildbot.worker.AbstractLatentWorker` and implementing :meth:`start_instance` and :meth:`stop_instance`.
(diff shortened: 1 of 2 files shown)


master/docs/manual/customization.rst
```diff
@@ -539,26 +539,34 @@ Aside from the service methods, the other concerns in the previous section apply
...
 
-    def stop_instance(self, fast=False):
-        # responsible for shutting down instance. Return a deferred. If `fast`,
-        # we're trying to shut the master down, so callback as soon as is safe.
-        # Callback value is ignored.
-        raise NotImplementedError
+This class is the base class of all latent workers and implements some common functionality.
+A custom worker should only need to override :meth:`start_instance` and :meth:`stop_instance` methods.
 
 See :class:`buildbot.worker.ec2.EC2LatentWorker` for an example.
```
seankelly: The EC2 worker is one of the first latent workers and has grown in complexity to be the largest latent worker. I think we should find a better example.

The Hyper worker might be better since @tardyp wrote it recently. It's the second shortest after Marathon and includes putting any `hyper_sh` method calls in a thread pool to handle them being blocking calls.

p12tic: Agreed, and also https://github.com/buildbot/buildbot/pull/4419 implements the new API only on hyper.

PR #4425, opened 2018-11-12
Display build times with d3

tardyp:
This shows the perf regression we had:
toto

### DIFF
--- a/master/buildbot/newsfragments/build_times.feature
+++ b/master/buildbot/newsfragments/build_times.feature
@@ -1,0 +1,1 @@
+Builder page now has a plot displaying the evolution of build times over time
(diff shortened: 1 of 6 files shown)


p12tic:
I was wondering how much server-side resources does this graph use on larger installations. Could it be that large number of users overwhelm the master?

p12tic:
Yet at the same time my own PR did completely stupid things in wrt. performance :-)

tardyp:
The graph is purely generated on the browser, with the data that it already downloaded for the build table. So no additional load for server.

tardyp:
You can see the result on metabuildbot: https://nine.buildbot.net/#/builders/68
(and the increase of pylint run time from 10 to 17 minutes)

p12tic:
Thanks for explanation. If we're only using the data we download anyway, then my comment does not apply. Good to know.

PR #4387, opened 2018-10-30
Introduce class-level method getResponsibleUsersForBuild (instead of …

churlinpr:
…performing a call to utils.getResponsibleUsersForBuild) in notifier.py. It was the only way to send mails ONLY to people specified in the owners property (allowing to bypass the changes' users).

## Contributor Checklist:

* [ ] I have updated the unit tests <- Unfortunately I could not get the tests to run. make virtualenv (from http://docs.buildbot.net/latest/developer/tests.html#quick-start) fails. I've tested the patch in my companies' bots though.
* [X] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [X] I have updated the appropriate documentation

PR #4426, opened 2018-11-13
Introduce class-level method getResponsibleUsersForBuild (instead of …

churlinpr:
…performing a call to utils.getResponsibleUsersForBuild) in notifier.py. It was the only way to send mails ONLY to people specified in the owners property (allowing to bypass the changes' users).

## Contributor Checklist:

* [ ] I have updated the unit tests <- Unfortunately I could not get the tests to run. make virtualenv (from http://docs.buildbot.net/latest/developer/tests.html#quick-start) fails. I've tested the patch in my companies' bots though.


churlinpr:
This is a rebased version of https://github.com/buildbot/buildbot/pull/4387 which I accidentally closed, as requested by @tardyp.

PR #4437, opened 2018-11-16
rework pending buildrequests

tardyp:
small rework to polish release

- optimise the number of request
- display all properties

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -18,10 +18,6 @@ env:
   global:
   - BUILDBOT_TEST_DB_URL=sqlite://
   - HYPER_SIZE=m1
-  # Current moto (1.3.7) requires dummy credentials to work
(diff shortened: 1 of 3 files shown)


p12tic:
I think for the future it would be useful to think about somehow changing the way properties are displayed so that the table wouldn't become very wide if there are a lot of different properties. That could become a problem on buildbot instances hosting a bunch of semi-related "projects" that all define their own, different properties for builders. If each project has, say, 3 unique property keys, then having 10 such projects would already make the table too wide IMO.

p12tic:
Many thanks for noticing this!

tardyp:
I agree with the display all property thing. 
There is a horizonal scroll bar which appear automatically, so it does not stretch the table too much.

Normally there is not too much buildrequests property (except with buildbot travis)

PR #4440, opened 2018-11-17
tests: Fire events registered to eventual in tearDown in libvirt tests

p12tic:
This fixes random timeouts of TestWorkQueue.test_work test when running tests in parallel.


master/buildbot/test/unit/test_worker_libvirt.py
```diff
@@ -185,6 +186,10 @@ def test_canStartBuild_domain_and_connected(self):
 
 class TestWorkQueue(unittest.TestCase):
 
+    @defer.inlineCallbacks
```
tardyp: no need for inlineCallbacks there
```suggestion
    def tearDown(self):
        return eventual.flushEventualQueue()
```

p12tic: I just wanted to be explicit - it's immediately obvious that the function returns a deferred when it's marked `@defer.inlineCallbacks`. Additionally, it's immediately obvious that it does not return value as `defer.returnValue` is not present. Also, it does not need refactor when we start adding more stuff to the `tearDown` function.

p12tic: Fixed now.

tardyp: keep in mind that inlineCallbacks has a non-negligible overhead. This is why it is better to just return the deferred if the function is just a simple function call.

If you want to make it obvious, you can always add a docstring.

p12tic: I haven't considered that.

p12tic: Just FYI, did some performance testing and it seems that each `@defer.inlineCallbacks` decoration takes a bit less time than 3 additional callbacks in the callbacks chain via `addCallback`.

See https://gist.github.com/p12tic/8b78560731d1324e50780bcc2a5bde91

tardyp: So, according to your test what is the overhead of inlineCallbacks compared to simple function with return of the deferred? (hint, there is a trap!)


I have made a bunch of contribution to inlineCallbacks
https://github.com/twisted/twisted/pull/813
https://github.com/twisted/twisted/pull/814
while it is very important for readabilty, it is also a huge hack with lots of downsides. This is why I like async/await, which is just simple and implemented in the langage so no need for hacks.

p12tic: > So, according to your test what is the overhead of inlineCallbacks compared to simple function with return of the deferred?

Approximately 1000 times :-) Maybe more.

It becomes interesting when comparing performance of functions that have several calls of addCallback. Looks like inlineCallbacks is maybe 10 to 30% slower.

async/await + ensureDeferred is 6 times faster.

p12tic: By the way, why don't we have a decorator like this:

```def asyncToDeferred(f):
    def wrapper(*args, **kwargs):
        return defer.ensureDeferred(f(*args, **kwargs))
    return wrapper```

Then use the new async/await syntax and enjoying the performance improvements while providing the same interfaces? There's no case when it's slower compared to inlineCallbacks at least, and we could use bare async functions internally.

p12tic: `Deferred` does have `__await__`, now I wonder if it works correctly. At least in the performance tests in the gist it does not crash.

p12tic: By the way, the PR should be good to go now :-)

PR #4443, opened 2018-11-18
Use inline callbacks where possible (part 2)

mokibit:
Use of defer.inlineCallbacks leads to more readable code.

### DIFF
--- a/master/buildbot/db/users.py
+++ b/master/buildbot/db/users.py
@@ -19,6 +19,8 @@ from __future__ import print_function
 import sqlalchemy as sa
 from sqlalchemy.sql.expression import and_
 
+from twisted.internet import defer
+
 from buildbot.db import base
 from buildbot.util import identifiers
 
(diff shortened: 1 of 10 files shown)


master/buildbot/db/users.py
```diff
@@ -216,9 +219,9 @@ def thd(conn):
                         return
 
             transaction.commit()
-        d = self.db.pool.do(thd)
-        return d
+        defer.returnValue((yield self.db.pool.do(thd)))
```
p12tic: No need to return?

master/buildbot/test/unit/test_data_buildsets.py
```diff
@@ -181,18 +170,15 @@ def do_test_addBuildset(self, kwargs, expectedReturn,
...
-            (bsid, brids) = xxx_todo_changeme
-            self.assertEqual((bsid, brids), expectedReturn)
-            # check the correct message was received
-            self.master.mq.assertProductions(
-                expectedMessages, orderMatters=False)
-            # and that the correct data was inserted into the db
-            self.master.db.buildsets.assertBuildset(bsid, expectedBuildset)
-        d.addCallback(check)
-        return d
+        xxx_todo_changeme = yield self.rtype.addBuildset(_reactor=clock, **kwargs)
```
p12tic: `xxx_todo_changeme` could have been collapsed simply to `(bsid, brids)`

PR #4448, opened 2018-11-18
Use inline callbacks where possible (part 7)

mokibit:
Use of defer.inlineCallbacks leads to more readable code.

### DIFF
--- a/master/buildbot/test/util/db.py
+++ b/master/buildbot/test/util/db.py
@@ -212,11 +212,12 @@ class RealDatabaseMixin(object):
         yield self.db_pool.do(self.__thd_clean_database)
         yield self.db_pool.do(self.__thd_create_tables, table_names)
 
+    @defer.inlineCallbacks
     def tearDownRealDatabase(self):
         if self.__want_pool:
-            return self.db_pool.do(self.__thd_clean_database)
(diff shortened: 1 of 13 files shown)


master/buildbot/test/util/db.py
```diff
@@ -212,11 +212,12 @@ def setUpRealDatabase(self, table_names=None, basedir='basedir',
         yield self.db_pool.do(self.__thd_clean_database)
         yield self.db_pool.do(self.__thd_create_tables, table_names)
 
+    @defer.inlineCallbacks
     def tearDownRealDatabase(self):
         if self.__want_pool:
-            return self.db_pool.do(self.__thd_clean_database)
-        return defer.succeed(None)
+            yield self.db_pool.do(self.__thd_clean_database)
```
tardyp: @p12tic This is the one that I was telling you about. Lets not use inlineCallbacks when it is not necessary.

Typically only use inlineCallbacks when this is replacing an addCallbacks

p12tic: I fully agree with your comment, especially after seeing the performance results. The problem was that lot of such cases have landed even before we had that discussion :-( I thought that it makes sense to fix that in bulk and let this case in, especially as it's test code.

PR #4467, opened 2018-12-06
Add support for v2 of the kv engine of vault secret provider

### LINKED ISSUE #4466: SecretInVault secret provider integration tests no longer work
The test `buildbot.test.integration.test_integration_secrets_with_vault.SecretsConfig.test_secret` no longer works.

See https://travis-ci.org/buildbot/buildbot/jobs/464401540.

Looks like the default kv engine shipping with the `vault` engine is now v2 which we don't support yet.

p12tic:
## Remove this paragraph
This PR adds support for v2 of the Vault key-value secret engine in the `SecretInVault` secret provider. 

I've also clarified documentation that the key-value engine is the only one supported. Looking into the documentation of Vault in does not look that other secret engines are using the same API at all. Supporting them properly needs a separate secret provider.

Fixes #4466.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/secrets/providers/vault.py
+++ b/master/buildbot/secrets/providers/vault.py
@@ -64,10 +64,14 @@ class HashiCorpVaultSecretProvider(SecretProviderBase):
         """
         if self.apiVersion == 1:
             path = self.secretsmount + '/' + entry
-            proj = yield self._http.get('/v1/{0}'.format(path))
         else:
             path = self.secretsmount + '/data/' + entry
-            proj = yield self._http.get('/v1/{0}'.format(path))
+
+        # note that the HTTP path contains v1 for both versions of the key-value
+        # secret engine. Different versions of the key-value engine are
+        # effectively separate secret engines in vault, with the same base HTTP
+        # API, but with different paths within it.
+        proj = yield self._http.get('/v1/{0}'.format(path))
         code = yield proj.code
         if code != 200:
             raise KeyError("The key %s does not exist in Vault provider: request"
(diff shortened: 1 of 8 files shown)


master/buildbot/secrets/providers/vault.py
```diff
@@ -57,11 +62,19 @@ def get(self, entry):
...
         get the value from vault secret backend
         """
-        path = self.secretsmount + '/' + entry
-        proj = yield self._http.get('/v1/{0}'.format(path))
+        if self.apiVersion == 1:
+            path = self.secretsmount + '/' + entry
+            proj = yield self._http.get('/v1/{0}'.format(path))
+        else:
+            path = self.secretsmount + '/data/' + entry
+            proj = yield self._http.get('/v1/{0}'.format(path))
```
tardyp: v1 as well?

master/buildbot/test/integration/test_integration_secrets_with_vault.py
```diff
@@ -66,9 +69,12 @@ def masterConfig():
             name="force",
             builderNames=["testy"])]
 
+    # note that as of December 2018, the vault docker image default to kv
```
tardyp: should we fix the version of vault in the tests?

master/buildbot/secrets/providers/vault.py
tardyp: at least add a comment.

tardyp:
appart from that looks good! thanks for working on it!

master/buildbot/test/integration/test_integration_secrets_with_vault.py
p12tic: Looking at https://hub.docker.com/_/vault/ it does not seem that there's some tag with v1 kv secret engine as default.

master/buildbot/secrets/providers/vault.py
p12tic: Agreed.

PR #4469, opened 2018-12-08
add a warning about python2

tardyp:
@p12tic @rodrigc


master/buildbot/scripts/start.py
```diff
@@ -129,10 +129,37 @@ def launch(config):
             pidfile.write("{0}".format(proc.pid))
 
 
+def py2Warning(config):
+    if sys.version[0] == '2' and not os.path.exists(os.path.join(config['basedir'], "no_python2_warning")):
```
p12tic: Maybe we should add a way to run buildbot tests by checking whether some key like `BUILDBOT_RUNNING_PYTHON_27_TESTS` is in os.environ? That way we could still run tests on python 2.7 right until we actually drop support.

master/buildbot/scripts/start.py
```diff
@@ -129,10 +129,37 @@ def launch(config):
...
+        -----------------------------------------------------
+
+
+        Python 2 is going unmaintained as soon as 2020: https://pythonclock.org/
+
+
+        Buildbot open source project is as well deprecating Python 2 for better maintainability.
+
+
+        Buildbot 2.0 going to be released in 2019 will remove support for Python < 3.5  https://github.com/buildbot/buildbot/issues/4439
```
p12tic: Buildbot 2.0 going to be ->  Buildbot 2.0 that is going to be

seankelly: I would phrase it as:

> Buildbot 2.0, scheduled for release in 2019, will remove support ...

When in 2019 will it be released?

rodrigc: @seankelly @tardyp My suggestion is to choose a date to announce dropping Python 2.7 support and advertise that.   In this e-mail,

https://lists.buildbot.net/pipermail/users/2018-November/001756.html

I suggested June 2019.  That is 7 months from now.  If we advertise 7 months in advance, that will give notice to people that they should consider upgrading their buildbot setup, or freezing on an older version.

djmitche:
It would be great to see a suggestion to upgrade now.  Something like "To prepare for that transition, we recommend upgrading your buildmaster to run on Python 3.6 now"

master/buildbot/scripts/start.py
tardyp: In the weekly meeting, we decided February, in order to put more pressure to upgrade. We can still delay it if we see in the telemetry that the community is not ready

rodrigc: @tardyp @djmitchie February may seem far away, but that's a bit of an aggressive date.
If you consider that it took nearly 1 year to get buildbot working on Python 3.....there are many companies and organizations using buildbot that don't move as fast.
Also, keep in mind that there are distributions such as FreeBSD that package up buildbot, and they will need to adapt.

However, at some point you need to pick a date, and just go for it.

I see you have changed the https://buildbot.net web site which is good.

I advise you to do the following ASAP (even out of cycle of any release):

1.  Send an e-mail to users@buildbot.net and announce@buildbot.net announcing the tenative date for dropping Python 2.7 support in the buildbot master.
2.  Tweet the same

master/buildbot/scripts/start.py
```diff
@@ -129,10 +130,35 @@ def launch(config):
             pidfile.write("{0}".format(proc.pid))
 
 
+def py2Warning(config):
+    if sys.version[0] == '2' and not config['quiet']:
+        print(textwrap.dedent("""\
+
+
```
rjarry: Why the empty lines?

master/buildbot/scripts/start.py
```diff
@@ -129,10 +130,35 @@ def launch(config):
...
+        -----------------------------------------------------
+
+        Python 2 is going unmaintained as soon as 2020: https://pythonclock.org/
+
+        To prepare for that transition, we recommend upgrading your buildmaster to run on Python 3.6 now!
+
+        Buildbot open source project is as well deprecating Python 2 for better maintainability.
+
+
+        Buildbot 2.0 going to be released in February 2019 will remove support for Python < 3.5
```
rjarry: Same here.

master/buildbot/scripts/start.py
```diff
@@ -129,10 +130,35 @@ def launch(config):
...
+        https://github.com/buildbot/buildbot/issues/4439
+
+        On most installation, switching to python3 is just about running the tool 2to3 over the master.cfg.
```
rjarry: On most installationS.

Over the master.cfg *file*.

seankelly: I would rephrase it to drop the "just". In general I try to avoid it when writing documentation because it can discourage people that don't fit in that use case but don't realize it.

> On most installations, switching to Python 3 can be accomplished by running the 2to3 tool over the master.cfg file.

tardyp:
I think I resolved all discussions. please tell me if I forgot something

PR #4474, opened 2018-12-12
change_hooks: fix json decode error on python3

rjarry:
Fix the following error with python3:

```
   File "buildbot/www/change_hook.py", line 107, in getAndSubmitChanges
     changes, src = yield self.getChanges(request)
   File "twisted/internet/defer.py", line 1418, in _inlineCallbacks
     result = g.send(result)
   File "buildbot/www/change_hook.py", line 168, in getChanges
     changes, src = yield handler.getChanges(request)
   File "buildbot/www/hooks/base.py", line 62, in getChanges
     properties = json.loads(args.get(b'properties')[0])
   File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
     s.__class__.__name__))
 builtins.TypeError: the JSON object must be str, not 'bytes'
```

Make sure to properly decode to unicode before parsing JSON data.

While there, also decode all other fields to unicode.

## Contributor Checklist:

* [x] I have updated the unit tests.
* [ ] I have updated the appropriate documentation: *N/A*

### DIFF
--- a/master/buildbot/newsfragments/change-hook-properties-python3.bugfix
+++ b/master/buildbot/newsfragments/change-hook-properties-python3.bugfix
@@ -1,0 +1,1 @@
+Fixed JSON decoding error when sending build properties to www change hooks on Python 3.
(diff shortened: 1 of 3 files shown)


tardyp:
care to make a unit test, and make sure this always works?

PR #4415, opened 2018-11-09
Remove version upper bounds for boto3/botocore and update

yan12125:
Version upper bounds are no longer necessary after the latest moto3
supports the latest boto3/botocore [1].

[1] https://github.com/spulec/moto/pull/1907

This pull request partially reverts https://github.com/buildbot/buildbot/pull/4359.

## Contributor Checklist:

* [ ] I have updated the unit tests
* [ ] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [ ] I have updated the appropriate documentation

(None of above is applicable for such changes, IMO)

PR #4497, opened 2018-12-31
ec2: Don't require test runner to set dummy credentials in environment

p12tic:
PR #4415 added an additional step required for running unit tests, namely requiring specific environment variables for unit tests. This PR sets them inside the tests themselves.


tardyp:
shouldn't we rather revert the hack, now that the issue is fixed?

p12tic:
The fix is not yet released. I agree that once moto 1.3.8 is released we should get rid of os.environ changes too.

Issue #4504, opened 2019-01-03
logMaxSize global param seems to be defunct


rgerhards:
The `logMaxSize` global parameter (and friends) seem to be defunctional in buildbot current and at least also in 1.1.

We have issues with some CI tests that generate excessively large log files in some error cases. This leads to Out Of Memory on the buildbot master machine and in turn often leads to total machine hangup. Configuration is
```
c = BuildmasterConfig = {}
...
# Global Log Settings
c['logCompressionMethod'] = 'gz'
c['logMaxSize'] = 1024*1024*32 # 32MiB
c['logMaxTailSize'] = 64*1024
c['logEncoding'] = 'utf-8'
```
Nevertheless, buildbot accepts far longer logs, Here is a sample of a ~580MiB log generated while these settings were active: https://build.rsyslog.com/#/builders/181/builds/909

Grepping over buildbot current (master branch in this case) shows that `logMaxSize` is read from config but never used. Commit https://github.com/buildbot/buildbot/commit/b8ccc08d4bb4e80244503cf37b6c8e73fb764f48 introduced the global parameter. However, buildbot current seems to not have any remains of that code. Might it have been removed accidently in the mean time. I tried to track this down in git history, but not being a Python programmer I could not really find a trace of where the code was moved to (or lost).

I would appreciate if someone could have a look at this issue.

tardyp:
Indeed, this is fairly possible that tis config option has been overlooked during conversion to data api.
would you like to contribute a fix to that?

I think the most straighforward way to implement this is to count logged bytes here:
https://github.com/buildbot/buildbot/blob/3ea4fdb113c58093c662562bf75c1be385e5d33e/master/buildbot/process/log.py#L79

You would just count the number of chars passed here, and stop calling self.master.data.updates.appendLog when the maxLogSize has been reached.

rgerhards:
[...]

PR #4505, opened 2019-01-03
re-enable global parameter "logMaxSize"

rgerhards:
The global parameter "logMaxSize" seems to be defunctional for
quite a while (at least back to 1.1). It simply is ignored. This
can trigger out of memory conditions and buildbot aborts/system
hangs when build steps generate overly large log files.

This patch re-enables it. However, this patch does NOT re-enable
the "logMaxTailSize" global parameter, which is also defunctional.
This may be done by a later PR (my current understanding of Python
is too weak to do this reliable in a single shot).

Many thanks to @alorbach for helping with the implementation and
@tardyp for pointing me to the right spot.

closes https://github.com/buildbot/buildbot/issues/4504

I have not updated the unit tests, as they also seem desfunctional and I do not know how to make them work. I still think the PR is worth merging as it fixes a critical problem.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/log.py
+++ b/master/buildbot/process/log.py
@@ -42,6 +42,8 @@ class Log(object):
         self.finishWaiters = []
         self.lock = defer.DeferredLock()
         self.decoder = decoder
+        self.numChars = 0
+        self.truncated = False
 
     @staticmethod
     def _decoderFromString(cfg):
@@ -81,7 +83,12 @@ class Log(object):
         # formatted for the log type, and newline-terminated
         assert lines[-1] == '\n'
         assert not self.finished
-        yield self.lock.run(lambda: self.master.data.updates.appendLog(self.logid, lines))
+        if not self.truncated:
+            self.numChars = self.numChars + len(lines)
+            if self.master.config.logMaxSize is not None and self.numChars > self.master.config.logMaxSize:
+                lines = ' \n\n#### LOG TOO LARGE - TRUNCATED #### - Configured maxLogSize is %d\n' % self.master.config.logMaxSize
+                self.truncated = True
+            yield self.lock.run(lambda: self.master.data.updates.appendLog(self.logid, lines))
 
     # completion
 
--- a/master/buildbot/newsfragments/logMaxSize.bugfix
+++ b/master/buildbot/newsfragments/logMaxSize.bugfix
@@ -1,0 +1,1 @@
+bugfix global parameter `logMaxSize` did no longer work


rgerhards:
Unset param is not handled - will fix

rgerhards:
spell checker failure is in code not touched by this PR

rgerhards:
OK, I was wrong on the spell checker, this is fixed as well. But for the remaining small drop in Coverage I really don't know how to add a test that checks these two lines. If you can guide me a bit, I'll try to craft a test.

p12tic:
See test_process_log.py for similar tests. You could store the truncation length and truncation template as instance variables of Log and change them in a test for simpler testing.

master/buildbot/process/log.py
```diff
@@ -81,7 +83,12 @@ def addRawLines(self, lines):
         # formatted for the log type, and newline-terminated
         assert lines[-1] == '\n'
         assert not self.finished
-        yield self.lock.run(lambda: self.master.data.updates.appendLog(self.logid, lines))
+        if not self.truncated:
+            self.numChars = self.numChars + len(lines)
+            if self.master.config.logMaxSize is not None and self.numChars > self.master.config.logMaxSize:
```
p12tic: I think it makes sense to truncate at line granularity. Right now we truncate the granularity of a single batch submitted to `addRawLines`. If batches of different lengths were submitted, then we would truncate at different points even if the logs themselves were identical which would be really confusing to humans reading the logs.

To not to regress on performance it makes sense to only do anything special if self.master.config.logMaxSize is not None and split the incoming `lines` into a list of actual lines whenever we detect that the current chunk will be truncated somewhere.

rgerhards: Thx for the suggestions, but I think the patch has a more severe problem: buildbot still aborts if large log files are sent, more info at https://github.com/buildbot/buildbot/issues/4504#issuecomment-451194269

If I can't fix that, it makes no sense to fix this regression here.

rgerhards:
based on @p12tic's comments on performance, I think we are taking the wrong path here in any case: The first and foremost place to limit the log file is the worker. Why send  very large log files to buildbot master just to discard them there? Granted, a check at master guards against DoS (and old workers), but the prime check should be at the worker side IMHO.

Am I on the right path if I think that `worker/buildbot_worker/runprocess.py` method `_sendBuffers` is a good place to add this? I just do not have any idea how the worker could obtain the server's `maxLogSize` setting.

tardyp:
The problem with worker is that it is quite hard to update. So we prefer not to add feature in the worker.
You will need to change the protocol for the master to tell what is the maxlogsize, and then change the worker code, and update all of them.
The DOS argument is not bad, but for me quite unlikely, and can be managed master side with the code you just sent.
The performance issue pointed by @p12tic is quite negligible imho compared to the rest of the stack.

rgerhards:
> The problem with worker is that it is quite hard to update. So we prefer not to add feature in the worker.
> You will need to change the protocol for the master to tell what is the maxlogsize, and then change the worker code, and update all of them.

understood - makes a lot of sense!

> The DOS argument is not bad, but for me quite unlikely, and can be managed master side with the code you just sent.

Unfortunately, my code seems not to work. I can still DoS buildbot master easily (details in https://github.com/buildbot/buildbot/issues/4504#issuecomment-451194269).

@tardyp If you like, I can craft a PR which does the same for your master, in case this helps understanding the issue. Not sure if it is related to my patch or some other part, but the plain fact is that buildbot master silently goes away.

> The performance issue pointed by @p12tic is quite negligible imho compared to the rest of the stack.

Well, I am always pro-performance and I had hoped the change in unintrusive enough. But of course there is always room for improvement. Without any idea of how to fix the abort of buildbot, I think there is little point in the patch.

PR #4507, opened 2019-01-03
KubeLatentWorker

tardyp:
This is a continuation of #3467   by @Zempashi

Changes:
* [x] Remove parsing of .kube/config It is actually quite complex to parse for an initial support
* [x] Replace by usage of kubectl proxy. Parsing of config, and authentication is delegated to the go cmdline
* [x] separate config management in its own classes for easier customization
* [x] Replace the usage of Jobs by Pods. Pods can be identified by name, this is necessary to kill previous lost latent worker in case of master crash or bug.

## Contributor Checklist:


tothandras:
Hi @tardyp :slightly_smiling_face: 

> Replace the usage of Jobs by Pods. Pods can be identified by name, this is necessary to kill previous lost latent worker in case of master crash or bug.

[labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) are perfect for this scenario, have you considered using them to find the jobs/pods?

tardyp:
@tothandras because names are unique identifiers. 
It is easy to delete a put by name, you just need one delete REST api, while for labels, you first need to search for it/them.

This also makes the logic more similar to docker or hyper.

master/buildbot/util/kubeclientservice.py
```diff
@@ -0,0 +1,249 @@
...
+
+    class LocalPP(ProcessProtocol):
+        def __init__(self, config_loader):
+            self.config_loader = config_loader
+            self.got_output_deferred = defer.Deferred()
+            self.terminated_deferred = defer.Deferred()
+
+        def outReceived(self, data):
+            if not self.got_output_deferred.called:
+                self.got_output_deferred.callback(data)
```
p12tic: Do we have any guarantees of how chunked input to `outReceived` is? We could get a single character worst case, couldn't we? I think ﻿`utils.getProcessOutputAndValue` would be better here.

master/buildbot/util/kubeclientservice.py
```diff
@@ -0,0 +1,249 @@
...
+    Parsing the config and setting up SSL is complex.
+    So for now, we use kubectl proxy to load the config and connect to master.
+    This will run the kube proxy as a subprocess, and return configuration with
+    http://localhost:PORT
+    """
+    kube_ctl_proxy_cmd = ['kubectl', 'proxy']  # for tests override
+
+    class LocalPP(ProcessProtocol):
+        def __init__(self, config_loader):
+            self.config_loader = config_loader
```
p12tic: config_loader is unused as far as I can see

master/buildbot/util/kubeclientservice.py
p12tic: Maybe we should wait for the first line break and call `got_output_deferred` only then?

master/buildbot/util/kubeclientservice.py
```diff
@@ -0,0 +1,249 @@
...
+class KubeInClusterConfigLoader(KubeConfigLoaderBase):
+    kube_dir = '/var/run/secrets/kubernetes.io/serviceaccount/'
+
+    kube_namespace_file = os.path.join(kube_dir, 'namespace')
+    kube_token_file = os.path.join(kube_dir, 'token')
+    kube_cert_file = os.path.join(kube_dir, 'ca.crt')
+
+    def checkConfig(self):
+        if not os.path.exists(self.kube_dir):
+            return config.error(
```
p12tic: `config.error` does not return anything

master/buildbot/worker/latent.py
```diff
@@ -307,6 +308,9 @@ def disconnect(self):
 
     @defer.inlineCallbacks
     def stopService(self):
+        # the worker might be insubstantiating from buildWaitTimeout
+        while self.insubstantiating:
+            yield asyncSleep(0.1)
```
p12tic: I think it's worth having an unit test to cover the shutdown of latent workers.

tardyp:
hyper.sh did shutdown yesterday as per plan. I didn't have time to finish this with the quality I wanted.

I will need to merge anyway, in order to enable kube on metabbot. I will address the review comment in a later PR.

PR #4470, opened 2018-12-10
Eliminate Python 2.7 support on buildbot master only

rodrigc:
Update documentation, installer, and CI to remove support for Python 2.7 from the buildbot master.

Fixes #4439

PR #4513, opened 2019-01-06
Replace range(len(foo)) with enumerate(foo).

rodrigc:
This eliminates some pylint warnings on Python 3.

This is a subset of the changes in #4470

### DIFF
--- a/master/buildbot/data/base.py
+++ b/master/buildbot/data/base.py
@@ -15,7 +15,6 @@
 
 from __future__ import absolute_import
 from __future__ import print_function
-from future.builtins import range
 from future.moves.collections import UserList
 
 import copy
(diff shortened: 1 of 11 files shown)


master/buildbot/data/base.py
```diff
@@ -53,8 +52,7 @@ def compileEventPathPatterns(self):
 
     def getEndpoints(self):
         endpoints = self.endpoints[:]
-        for i in range(len(endpoints)):
-            ep = endpoints[i]
+        for (i, ep) in enumerate(endpoints):
```
yan12125: I think parentheses around `i, ep` are usually omitted? That's what I see in other Python projects, and that's also how `enumerate()` is used in [PEP 279](https://www.python.org/dev/peps/pep-0279/), the PEP proposing `enumerate()`.

rodrigc: OK I got rid of the parentheses next to the enumerate() statements.

seankelly:
I see a number of instances where `foo[i]` isn't replaced with the new `val`.

tardyp:
> I see a number of instances where foo[i] isn't replaced with the new val.


indeed. We need to be careful, because the codecov for this code is not very good.

rodrigc:
@tardyp I didn't use `val` in all the places, because in some places they do some things where they really need to get at the index variable `i`, so just left it like that

p12tic:
I would like to see `foo[i]` to be replaced with `val` where possible too. Otherwise LGTM.

Issue #4521, opened 2019-01-10
Docs §1.1.4 not valid behind certain proxies/firewalls


0cjs:
In the "First Run" documentation at the end of [§1.1.4](https://docs.buildbot.net/1.7.0/tutorial/firstrun.html#creating-a-worker) (after the instructions to start the worker) it says you should see something under the waterfall view and "Your master is now quietly waiting for new commits to hello-world."

My waterfall view shows a blank panel, rather than a line with "runtests" as in the documentation image, and poking around does not provide any indication that anything is configured for `hello-world` at all.

The reason for this is likely that I tried this in a heavily firewalled environment where access to `git://github.com/buildbot/hello-world.git` is blocked:

    2019-01-10 17:01:22+0900 [-] command ['fetch', 'git://github.com/buildbot/hello-world.git', '+master:refs/buildbot/git%3A%2F%2Fgithub.com%2Fbuildbot%2Fhello-world.git/master'] in /home/cjs/co/py-virtualenv/buildbot/buildbot/master/gitpoller-workdir on repourl git://github.com/buildbot/hello-world.git failed with exit code 128: fatal: unable to connect to github.com:
            github.com[0: 192.30.255.113]: errno=Connection timed out                   
            github.com[1: 192.30.255.112]: errno=Connection timed out

Changing the URL to use `https:` instead of `git:` and running `buildbot reconfig` doesn't fix the problem even though both the master and the worker were started with working `http_proxy` etc. environment variables. (`git clone https://github.com/buildbot/hello-world.git` does work fine with these set.) Searching the web for how to configure the master and builder to use a proxy is difficult because of the enormous number of hits on pages describing how to use a reverse proxy in front of the buildbot server.

### Suggested Actions

1. Where you have the user check the buildbot master web server for the `hello-world` app, add a note that network access to `github.com` port TCP/9418 (Git protocol) is required for this to be working.
2. Consider using `https:` instead of `git:` as the default protocol for that repo if that would work ok. This will fix the problem for people with port 443 available but 9418 blocked to `github.com`.
3. From the same place, link to documentation on how to get the master and builder to use a proxy, if that can be done, for people who don't have port 443 access to `github.com`.

tardyp:
Hy @0cjs I think those are good recommendations. Can you send a PR so you can get credits for your findings?

0cjs:
[...]

PR #4522, opened 2019-01-10
Change demo repo and docs to fetch via HTTPS instead of Git protocol

0cjs:
This does little to help people without this access, but at least it
should save new users time debugging what's going wrong.

In the long run we should consider the following:
1. Change the fetch of `hello-world` to use `https:` instead of
   `git:`, for those whom port 9148 is blocked.
2. Document how to get the master and builder to use proxy, for those
   whom port 443 is blocked.

## Contributor Checklist:

* [N/A] I have updated the unit tests
* [N/A] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [YES] I have updated the appropriate documentation


tardyp:
Hey, thanks for the PR.
Didn't you want to update the sample.cfg as well?

https://github.com/buildbot/buildbot/blob/f9530f6543eb45b9de1fc6ccecc14c98d9f8137d/master/buildbot/scripts/sample.cfg#L34

seankelly:
The git protocol version was [picked eight years ago](https://github.com/buildbot/buildbot/commit/85eb1b0e3b3ddb5ea7073db7d3bd5668b91ec2be). I believe there was an HTTP endpoint at the time but I recall git didn't work well with it back then. It's safe to switch now.

0cjs:
Ok. I've now verified that `requests` should be using a proxy if  the correct variables are set and tested this myself, using a proxy. I've added a commit to this branch to change to `https` for fetching.

Other potential changes I can do if you want:
1. Squash the two commits if you don't want to keep the "need to be able to connect to 9148" documentation change in the history.
2. Update the initial comment in the PR to match the new stuff this branch now does, if that's how you roll.

In #4521 I'm also discussing a separate potential documentation problem; if I end up making commits to fix that I'm happy with either adding them to this PR or making a new one.

p12tic:
@0cjs I think the first suggestion makes sense. Regarding to the second, you can do if you want, but I guess it's not worth spending additional time.

Thanks!

p12tic:
@0cjs Could you also add a newsfragment so that your changes shows in the release notes?

0cjs:
I have squashed the commits and added a newsfragment file (marked as a "bugfix" for lack of a better term that I can find), leaving a single commit that I think is ready to go on to `master`. (There are other suggestions on #4521 for further doc changes, but I see no reason to delay this for those.)

p12tic:
LGTM, thank you! Someone needs to restart the Appveyor test as it fails due to unrelated reasons.

tardyp:
❤️

Issue #4491, opened 2018-12-25
Builds exceeding lock not scheduled


vbraun:
We have a lock limiting buildbot to 3 concurrent build jobs. Submitted tasks exceeding the limit used to be shown with a yellow badge on /#/builders (just like builds in progress), but now not any more. This may be an intentional change, though it doesn't really make sense to me. 

The pending tasks are shown on /#/pendingbuildrequests. The pending tasks are not being started when the lock is being released by a completed task, so the pending tasks never start. This is certainly a bug. 

This is the sagemath buildbot http://build.sagemath.org running buildbot-1.7.0 on python 2.7

There are two exceptions it twistd.log (attached), though they don't seem to be obviously related:
[bulidbot-twistd.log](https://github.com/buildbot/buildbot/files/2708967/bulidbot-twistd.log)

vbraun:
For the record, the the debug tab on #/buildrequests/<n> for such a pending-but-never-starting request says
~~~~
Buildrequest
builderid 71 
buildrequestid 23 
buildsetid 23 
claimed false 
claimed_at
claimed_by_masterid
complete false 
complete_at
priority 0 
properties
results -1 
submitted_at 1545760877 
waited_for false 

Buildset 
bsid 23 
complete false 
complete_at
external_idstring
parent_buildid
parent_relationship
reason A build was forced by 'sage': force build 
results -1 
sourcestamps [{"branch":"","codebase":"","created_at":1545735534,"patch":null,"project":"","repository":"","revision":"","ssid":1}]
submitted_at 1545760877 

Builder
_endpoint builders 
builderid 71 
description
masterids [1] 
name Zpkg slow AIMS bu18_64 (Ubuntu 18.04 64 bit) pkg-binary 
tag ["binary-pkg"] 
_id 71 

Builds
~~~~

vbraun:
Downgrading to buildbot-1.6.0 fixes it for me

p12tic:
Likely caused by my changes. Will look ASAP.

p12tic:
Hey @vbraun, could you share the buildbot config that you're using? So far I've been unable to reproduce the issue of builds being not scheduled on a simple setup (3 workers, 1 builder and a master lock limiting build concurrency to 2). Once the lock becomes available, subsequent builds are scheduled as expected.

[...]

PR #4525, opened 2019-01-12
Try to issue builds on lock release (case 4491)

p12tic:
Fixes bug #4491. The regression has been caused by changes in #4416. Before that PR we always started buildrequests even if they had unsatisfied locks. After the PR the BuildRequestDistributor started to check whether the build to be issued could satisfy the lock requirements. Unfortunately, we didn't notify BuildRequestDistributor about released locks in all cases, e.g. when build finishes we just notified it to run builders on the same worker. If a buildrequest uses completely unrelated builder, then it won't be scheduled.

This PR fixes this in a bit hacky way - we just inform the botmaster to recheck all builders for buildrequests to be scheduled whenever a build with locks finishes. That could be a performance problem for larger setups.

A better solution would be to track the locks that caused a buildrequest to be not scheduled and only recheck the buildrequest whenever that specific lock got released.

I plan to look into the performance of the BuildRequestDistributor in the case of a large number of pending build requests being present. Performance problems in this specific scenario is one of the reasons why Unity Technologies is moving away from buildbot. It's not the only setup experiencing problems as I remember more reports of that in the issues list.

## Contributor Checklist:

* [x] (n/a) I have updated the appropriate documentation

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1096,6 +1096,7 @@ regexs
 registeradapter
 reid
 rejectedworkers
+releaselocks
 releasers
 remoteaddresses
 remotecommand
@@ -1444,6 +1445,7 @@ unabbreviated
 unambiguity
 unclaim
 unclaimedbrdicts
+unclosed
 unconfigure
 underpowered
 unencrypted
@@ -1559,6 +1561,7 @@ whitespace
 wiki
 wikipedia
 wildcard
+winerror
 wkdir
 workdir
 workdirs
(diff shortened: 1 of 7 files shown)


p12tic:
> Can you disclose in which other direction Unity Technologies is moving to? Maybe we can get inspiration from it.

Unfortunately I'm not an area I know well :-(

p12tic:
> Couldn't this performance problem be fixed by reworking the algorithm/implementation of the current BuildRequestDistributor?

I believe it could.

rodrigc:
@p12tic if Unity moves away from buildbot, will you stop contributing to buildbot?  It's understandable if you have to do this, but it will be a loss for the buildbot team, since you are doing some very solid contributions!

p12tic:
I will continue working on Buildbot. I'm contributing on my free time on things unrelated to business needs of Unity.

rodrigc:
@p12tic Glad to keep you on the team!

PR #4527, opened 2019-01-13
Fix latent worker integration tests

p12tic:
Looks like some latent worker integration tests returned a Deferred. Since they derived the `SynchronousTestCase` test case class, that meant that any assertion errors got silently ignored.

This PR fixes this by expanding TestReactor to support `callLater(0, ...)` and not lead to deadlocks in tests even if the reactor is not manually kicked for iteration. Additionally, the latent worker integration tests were fixed to not return a Deferred.

The changes to the reactor undo the changes in #4525.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/master.py
+++ b/master/buildbot/master.py
@@ -199,6 +199,7 @@ class BuildMaster(service.ReconfigurableServiceMixin, service.MasterService,
                                                      masterid=self.masterid)
             yield self.data.updates.expireMasters()
         self.masterHeartbeatService = internet.TimerService(60, heartbeat)
+        self.masterHeartbeatService.clock = self.reactor
         # we do setServiceParent only when the master is configured
         # master should advertise itself only at that time
 
--- a/master/buildbot/test/fake/reactor.py
+++ b/master/buildbot/test/fake/reactor.py
@@ -23,6 +23,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+from twisted.internet import reactor
 from twisted.internet.base import _ThreePhaseEvent
 from twisted.internet.interfaces import IReactorCore
 from twisted.internet.interfaces import IReactorThreads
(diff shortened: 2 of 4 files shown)


tardyp:
I think it is a good idea to keep those tests synchronous as much as we can.

Can we patch successResultOf to run the eventually calls until the deferred is called?

p12tic:
Why, is this because of performance concerns? My own view is that using a consistent coding style allows higher quality of tests due to lower barrier of entry.

tardyp:
Synchronous tests are easier to reason about and indeed faster.

This is why @tomprince introduced them, and I think we should move toward that as a target consistent coding style rather than away as much as we can.

p12tic:
I think there's a bit of misunderstanding on both sides :-)

I believe the current solution already addresses your concerns, just in a non-obvious way.

As we are using `TestReactor`, the "asynchronous" tests are not actually fully asynchronous. The deferred must still have the result at the point of `yield`. The only thing that differs is that anything registered to the real `reactor` may have a chance to run, otherwise there's deadlock. The current modifications to the `TestReactor` allow anything that has been added to it via `callLater(0, ...)` to run at the point of `yield`. Otherwise, the test is still fully synchronous as long as nothing uses the real `reactor`.

The benefits of this approach is better readability, as `@inlineCallbacks` is very well known compared to `successResultOf` which is used only in few tests. Additionally, this reduces the chance of silent test failures, as `SynchronousTestCase` does not handle tests marked with `@inlineCallbacks` well - it just ignores any assertion errors. Finally, this is much more robust - in case of `successResultOf` sometimes one needs to somehow workaround the fact that the deferred does not have the result yet.

As for downsides, I don't have any data yet about the performance.

What do you think about the points I raised? Do you believe they are correct?

p12tic:
Update: the performance of the test_latent integration tests is reduced only very slightly.

On a Ryzen 2 CPU @ 3.00GHz with turbo boost disabled, `time trial buildbot.test.integration.test_latent` returns the following `real` timings:

Before the PR: between 3.31-3.37s, average 3.35s.
After the PR: between 3.46-3.51s, average 3.48s.

The regression is less than 5%.

tomprince:
> in case of `successResultOf` sometimes one needs to somehow workaround the fact that the deferred does not have the result yet.

While, using `successResultOf` requires one to think about when the deferred fires, and what is required to make that happen, that is a strength, rather than a weakness. I'm no longer involved with buildbot, so I don't know the current state of these tests; but when I originally wrote them, I needed the flexibility provided by synchronous tests. I was fixing subtle races, that couldn't easily be triggered without using the flexibility of `successResultOf`.

> The benefits of this approach is better readability, as `@inlineCallbacks` is very well known compared to `successResultOf` which is used only in few tests.

`inlineCallbacks` is older than `successResultOf`, so most tests predate `successResultOf`. I don't think you can conclude anything based on the relative occurrence of the idioms.

p12tic:
> While, using successResultOf requires one to think about when the deferred fires, and what is required to make that happen, that is a strength, rather than a weakness [...]  I was fixing subtle races, that couldn't easily be triggered without using the flexibility of successResultOf.

I definitely agree that it useful to have a way to trigger hiding race conditions. What I'm arguing is that the approach I'm proposing would still have the same property just by continuing the use of the fake `TestReactor`.

The major actual difference is that `successResultOf` currently cannot resolve callbacks registered via `reactor.callLater(0, ...)`. That's a problem, because this limits testability of code as in certain cases we absolutely need to defer code to the next iteration of the reactor. If we fix this within `successResultOf`, we'll effectively have the same behavior as what I'm proposing. The rest of the differences are basically usability, they don't affect how the test actually runs.

tardyp:
I updated the branch as a way to re-run the tests on kube.

I still would like to keep those integrations tests with successResultOf as an example on how to use this way of testing. 
This does not mean that we want to convert all code to SynchronousTestCase, but I think our codebase would benefit doing more of those.

p12tic:
@tardyp: Got that. I've fixed the latent worker tests to keep using `SynchronousTestCase`.

PR #4536, opened 2019-01-17
db: fix foreign key constraints for "nullable" columns

rjarry:
Some foreign key columns have `nullable=True` (which is the default). When a record on which such a foreign key points to is deleted, it should not be deleted in cascade (since it is valid to have a `NULL` value).

Update the `ondelete` fields of all these `nullable` foreign keys to `'SET NULL'`. To make things clearer, explicitly add `nullable=True` statements where missing. Also, some foreign keys columns were missing `nullable=False` statements. Add them.

Add a migration script and a unit test.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/migrate/versions/052_cascading_set_null.py
+++ b/master/buildbot/db/migrate/versions/052_cascading_set_null.py
@@ -1,0 +1,95 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+from __future__ import absolute_import
+from __future__ import print_function
+from future.utils import iteritems
+
+import sqlalchemy as sa
+from migrate.changeset.constraint import ForeignKeyConstraint
+from migrate.exceptions import NotSupportedError
+
+from buildbot.util import sautils
+
+
+def upgrade(migrate_engine):
+    metadata = sa.MetaData()
+    metadata.bind = migrate_engine
+
+    table_names = set(TABLES_FKEYS_SET_NULL.keys())
+    table_names.update(TABLES_COLUMNS_NOT_NULL.keys())
+
+    tables = {}
+    for t in table_names:
+        tables[t] = sautils.Table(t, metadata, autoload=True)
+
+    fks_to_change = []
+    # We need to parse the reflected model in order to find the automatic
+    # fk name that was put.
+    # Mysql and postgres have different naming convention so this is not very
+    # easy to have generic code working.
+    for t, keys in iteritems(TABLES_FKEYS_SET_NULL):
+        table = tables[t]
+        for fk in table.constraints:
+            if not isinstance(fk, sa.ForeignKeyConstraint):
+                continue
+            for c in fk.elements:
+                if str(c.column) in keys:
+                    # migrate.xx.ForeignKeyConstraint is changing the model
+                    # so initializing here would break the iteration
+                    # (Set changed size during iteration)
+                    fks_to_change.append((
+                        table, (fk.columns, [c.column]),
+                        dict(name=fk.name, ondelete='SET NULL')))
+
+    for table, args, kwargs in fks_to_change:
+        fk = ForeignKeyConstraint(*args, **kwargs)
+        table.append_constraint(fk)
+        try:
+            fk.drop()
+        except NotSupportedError:
+            # some versions of sqlite do not support drop,
+            # but will still update the fk
+            pass
+        fk.create()
+
+    for t, cols in iteritems(TABLES_COLUMNS_NOT_NULL):
+        table = tables[t]
+        if table.dialect_options.get('mysql', {}).get('engine') == 'InnoDB':
+            migrate_engine.execute('SET FOREIGN_KEY_CHECKS = 0;')
+        try:
+            for c in table.columns:
+                if c.name in cols:
+                    c.alter(nullable=False)
+        finally:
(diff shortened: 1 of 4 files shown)


rjarry:
I tried to setup the cleanup using the cascade deletes and I noticed that when removing old sourcestamps, all changes are systematically deleted.

This is caused by the fact that `changes.parent_changeids` (which is allowed to be `NULL`) has a `ON DELETE CASCADE` trigger. If you delete one change, all its "child" changes are deleted. This is a problem :)

However, by putting `SET NULL` this basically makes the "oldest after delete" changes the new "root" changes (i.e. the change without a `parent_changeids`).

I applied the same reasoning to other tables.

In any case, when we have `nullable=False` we must have `ondelete='CASCADE'`. When `nullable=True` I think we should always have `ondelete='SET NULL'` but there are some columns where that may require discussion.

master/buildbot/db/migrate/versions/052_cascading_set_null.py
```diff
@@ -0,0 +1,90 @@
...
+        try:
+            fk.drop()
+        except NotSupportedError:
+            # some versions of sqlite do not support drop,
+            # but will still update the fk
+            pass
+        fk.create()
+
+
+TABLES_FKEYS_SET_NULL = {
```
tardyp: I don't see parent change ids in this list.

I am not sure how to read those tables.

Will the buildrequest_claims associated with a masterid be deleted when I delete that master?

I expect so, but reading the code it looks like it will just nullify the FK

rjarry: Yes the parent change ids is:

```python
```

And yes, when deleting a master, the related `buildrequest_claims.masterid` column will be set to `NULL`. This is one of the cases where I was not sure what to do.

Maybe the correct solution is to forbit `buildrequest_claims.masterid` to be `NULL`. So move it from `TABLES_FKEYS_SET_NULL` to `TABLES_COLUMNS_NOT_NULL`.

rjarry:
For the record, a thorough cleanup script looks like this:

```sql
BEGIN;

-- Delete all sourcestamps without any changes.
DELETE FROM sourcestamps WHERE id NOT IN (SELECT sourcestampid FROM changes);

-- Modify the foreign key constraint on changes.parent_changeids.
-- This is mandatory to allow deleting old changes without deleting all
-- "child" changes in cascade.
-- This hack may be removed once this PR is merged.
ALTER TABLE ONLY changes DROP CONSTRAINT changes_parent_changeids_fkey;
ALTER TABLE ONLY changes ADD CONSTRAINT changes_parent_changeids_fkey
FOREIGN KEY (parent_changeids) REFERENCES changes(changeid) ON DELETE SET NULL;

-- Delete soucestamps older than 30 days.
-- The starting point is not "now" but the most recent sourcestamp.
DELETE FROM sourcestamps
WHERE created_at < (SELECT max(created_at) - (30 * 24 * 60 * 60) FROM sourcestamps);

-- Delete buildsets that have no link to a sourcestamp.
DELETE FROM buildsets
WHERE id NOT IN (SELECT buildsetid FROM buildset_sourcestamps);

-- Delete builders that are not referenced by any build.
DELETE FROM builders
WHERE id NOT IN (SELECT builderid FROM builder_masters)
AND id NOT IN (SELECT builderid FROM builds);

COMMIT;
```

seankelly:
> This is caused by the fact that `changes.parent_changeids` (which is allowed to be `NULL`) has a `ON DELETE CASCADE` trigger. If you delete one change, all its "child" changes are deleted. This is a problem :)

Shouldn't all of those changes belong to the same sourcestamp however? If you're deleting a sourcestamp, there's no need for the changes related to it.

That said, I ran into a similar problem but from the opposite side. I wanted to prune the number of changes and ran into a lot of difficulty with the `parent_changeids` column. I needed to keep the existing sourcestamps but only delete old changes. The self-referential nature of `parent_changeids` made it challenging.

rjarry:
> Shouldn't all of those changes belong to the same sourcestamp however? If you're deleting a sourcestamp, there's no need for the changes related to it.

Yes and the changes are deleted in cascade because of their foreign key to sourcestamps.id. However, deleting these changes also causes the deletion of their "child" changes (which are completely unrelated to the deleted sourcestamps).

> Your cleanup script is very interesting. Adding that to the docs, or code as a janitor, would be extremely helpful.

I do not like the "janitor" concept as it is at the moment. I prefer such low-level administration tasks to be run outside of the normal build workflow. I might submit something like an improvement of the `buildbot cleanupdb` command. Something like

```
~$ buildbot cleanupdb --max-days 30
```

would be nice :)

And it may actually work on all database engines. Not only on PostgreSQL and MySQL with InnoDB engine.

master/buildbot/db/migrate/versions/052_cascading_set_null.py
tardyp: Your reasoning here is that CASCADE_DELETE and NOT_NULL are equivalent. I disagree with this.

There are cases where you want to cascade delete even if the FK can be null.

if you delete a patch, you want to cascade delete the sourcestamps, and then the builds associated with it. because the build makes no sense without the patch.
But.. a sourcestamp can be made without patch, and this is meaningful.

in this list, I think only  following make sense
```
   'buildsets': ['parent_buildid'],
    'changes': ['changes.changeid'],
```
The other shall be ON DELETE CASCADE and NULLABLE imho

rjarry: Ok. But when you delete a worker, do you want to delete all its builds?

tardyp: I think a build without worker makes no sense indeed. It should not be NULLABLE.
It might be NULLABLE temporarily for technical reasons

rjarry: What should we do about this comment then?

```python
        # worker which performed this build
        # keep nullable to support worker-free builds
```

rjarry: I went back in time as far as I could, the comment was added here:

https://github.com/buildbot/buildbot/commit/9dbb106fc2b274527a08b1b945aa0897a471cee3#diff-58b3908894b0546af761475a0243ffd1R98

With no specific reason. By looking at the code that creates builds. There is no way we can have builds without any worker. I'll change `builds.workerid` to `nullable=False` and leave `ondelete='CASCADE'`.

seankelly:
> However, deleting these changes also causes the deletion of their "child" changes (which are completely unrelated to the deleted sourcestamps).

My experience was that the child changes were all related to the same sourcestamp. I did not check extensively however. I'll try to verify that sometime later.

PR #4540, opened 2019-01-18
fix and improve usage of KubeHardcodedConf

doudou:
This pull request fixes the usage of the `verify` argument in KubeHardcodedConf, and of the passing of the Authorization header - which was not forwarded to the HTTP layer.

In addition, it improves the usage of the authorization token, by creating a separate dedicated `authorization` keyword arg, and allowing usage of a secret provider.

The documentation is updated, with some GKE-specific info (since that's what we use, I hope that's OK, it's hard to get when you're not used to Kubernetes).

### DIFF
--- a/master/buildbot/util/httpclientservice.py
+++ b/master/buildbot/util/httpclientservice.py
@@ -1,4 +1,4 @@
-# This file is part of Buildbot. Buildbot is free software: you can)uth
+# This file is part of Buildbot. Buildbot is free software: you can)
 # redistribute it and/or modify it under the terms of the GNU General Public
 # License as published by the Free Software Foundation, version 2.
 #
--- a/master/buildbot/util/httpclientservice.py
+++ b/master/buildbot/util/httpclientservice.py
@@ -1,4 +1,4 @@
-# This file is part of Buildbot. Buildbot is free software: you can
+# This file is part of Buildbot. Buildbot is free software: you can)uth
 # redistribute it and/or modify it under the terms of the GNU General Public
 # License as published by the Free Software Foundation, version 2.
 #
@@ -159,8 +159,9 @@ class HTTPClientService(service.SharedService):
             kwargs['data'] = jsonBytes
         return url, kwargs
 
+    @defer.inlineCallbacks
     def _doTxRequest(self, method, ep, **kwargs):
-        url, kwargs = self._prepareRequest(ep, kwargs)
+        url, kwargs = yield self._prepareRequest(ep, kwargs)
         if self.debug:
             log.debug("http {url} {kwargs}", url=url, kwargs=kwargs)
 
@@ -170,25 +171,25 @@ class HTTPClientService(service.SharedService):
             if self.debug:
                 log.debug("==> {code}: {content}", code=res.status_code, content=res.content)
             return res
+
         # read the whole content in the thread
         kwargs['background_callback'] = readContent
         if self.verify is False:
             kwargs['verify'] = False
-        d = self._session.request(method, url, **kwargs)
-        d.addCallback(TxRequestsResponseWrapper)
-        d.addCallback(IHttpResponse)
-        return d
 
+        res = yield self._session.request(method, url, **kwargs)
+        return defer.returnValue(IHttpResponse(TxRequestsResponseWrapper(res)))
+
+    @defer.inlineCallbacks
     def _doTReq(self, method, ep, **kwargs):
-        url, kwargs = self._prepareRequest(ep, kwargs)
+        url, kwargs = yield self._prepareRequest(ep, kwargs)
         # treq requires header values to be an array
         kwargs['headers'] = {k: [v]
                              for k, v in kwargs['headers'].items()}
         kwargs['agent'] = self._agent
 
-        d = getattr(treq, method)(url, **kwargs)
-        d.addCallback(IHttpResponse)
-        return d
+        res = yield getattr(treq, method)(url, **kwargs)
+        return defer.returnValue(IHttpResponse(res))
 
     # lets be nice to the auto completers, and don't generate that code
     def get(self, ep, **kwargs):
--- a/master/buildbot/util/httpclientservice.py
+++ b/master/buildbot/util/httpclientservice.py
@@ -178,7 +178,7 @@ class HTTPClientService(service.SharedService):
             kwargs['verify'] = False
 
         res = yield self._session.request(method, url, **kwargs)
-        return defer.returnValue(IHttpResponse(TxRequestsResponseWrapper(res)))
+        return IHttpResponse(TxRequestsResponseWrapper(res))
 
     @defer.inlineCallbacks
     def _doTReq(self, method, ep, **kwargs):
@@ -189,7 +189,7 @@ class HTTPClientService(service.SharedService):
         kwargs['agent'] = self._agent
 
         res = yield getattr(treq, method)(url, **kwargs)
-        return defer.returnValue(IHttpResponse(res))
+        return IHttpResponse(res)
 
     # lets be nice to the auto completers, and don't generate that code
     def get(self, ep, **kwargs):
--- a/master/buildbot/util/kubeclientservice.py
+++ b/master/buildbot/util/kubeclientservice.py
@@ -66,7 +66,8 @@ class KubeConfigLoaderBase(BuildbotService):
 class KubeHardcodedConfig(KubeConfigLoaderBase):
     def reconfigService(self,
                         master_url=None,
-                        authorization=None,
+                        bearerToken=None,
+                        basicAuth=None,
                         headers=None,
                         cert=None,
                         verify=None,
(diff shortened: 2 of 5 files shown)


doudou:
Note: no news item as it's small improvements on the very new K8S support

master/docs/manual/configuration/workers-docker.rst
```diff
@@ -469,10 +469,34 @@ With :class:`KubeHardcodedConfig`, you just configure the necessary parameters t
...
+    you could store e.g. a bearer token in a secret provider and do
+
+        authorization=util.Interpolate("Bearer %(secret:k8s-token)")
+
+    When using the Google Kubernetes Engine (GKE), a bearer token for the
+    default service account can be had with:
+
+        gcloud container clusters get-credentials --region [YOURREGION] YOURCLUSTER
+        kubectl describe sa
+        kubectl describe secret [SECRET_ID]
```
tardyp: My understanding is that this secret is temporary, and will only last for a limited time. kubectl is supposed to automatically renew that token.

So whereas this method might work, if I understand correctly, it is not suitable for prod.

doudou: Well ... I'm running with that, I'll report my findings by the next week ;-)

My understanding - but it is very unclear in the docs - is that the token for the user has a limited time, not the token for a service account. `get-credentials` recommends using service accounts for external applications that need to access the Kubernetes cluster.

master/buildbot/util/httpclientservice.py
```diff
@@ -175,25 +176,25 @@ def readContent(session, res):
...
         kwargs['background_callback'] = readContent
         if self.verify is False:
             kwargs['verify'] = False
-        d = self._session.request(method, url, **kwargs)
-        d.addCallback(TxRequestsResponseWrapper)
-        d.addCallback(IHttpResponse)
-        return d
 
+        res = yield self._session.request(method, url, **kwargs)
+        return IHttpResponse(TxRequestsResponseWrapper(res))
```
tardyp: this breaks on python2. You need to use defer.returnValue inside inlineCallbacks

master/buildbot/util/kubeclientservice.py
```diff
@@ -190,19 +193,31 @@ def __init__(self, kube_config=None):
...
+        if config['headers']:
+            if not req_kwargs['headers']:
+                req_kwargs['headers'] = {}
+            else:
+                req_kwargs['headers'] = req_kwargs['headers'].dup
+            req_kwargs['headers'].update(config['headers'])
+
+        auth = config.get('authorization', None)
+        if auth is not None:
+            req_kwargs['headers']['Authorization'] = yield self.config.renderSecrets(auth)
```
tardyp: renderSecrets can be slow (contact Vault for example). generally, the secrets are cached. It would be better if the secrets were calculated at the time of the reconfigService, so that we don't have to put prepareRequest as returning deferred.

master/docs/manual/configuration/workers-docker.rst
tardyp: Why don't you use `KubeProxyConfigLoader`? It was written especially for the weird renewable token of GKE.
That is what we are using for metabuildbot.

master/buildbot/util/httpclientservice.py
seankelly: Will that matter though?

tardyp: right. We can start removing defer.returnValue() from the codebase after we drop 2.7

master/docs/manual/configuration/workers-docker.rst
doudou: I'm running the master on-premise, only the slaves on GKE. Or have I misunderstood how KubeProxyConfigLoader works ?

master/buildbot/util/kubeclientservice.py
doudou: As far as I can see, that would require a major infrastructure change in buildbot - reconfigService is not expected to return a deferred. I'm not ready to spend the time needed for that change given that "slow" is not really an issue here - starting the builds on GKE for me take already ~10s, I don't mind having a few more.

master/buildbot/util/httpclientservice.py
doudou: Will change

master/docs/manual/configuration/workers-docker.rst
doudou: OK ... not right. KubeCtlProxyConfigLoader is using kubectl. I've added a commit to describe the three options at a high level.

I find it a lot more convenient than having to build a container with gcloud setup and logged in, as well as with kubectl installed, to be honest. I agree that it would be a solution, but for what it's worth the bearer token seems to work so far.

I can't find any documentation on automatic rotation of tokens on GKE. Do you have a link ?

doudou:
- added bearerToken/basicAuth in place of authorization
- added the missing defer.returnValue
- I still have not changed the documentation regarding the service account. I'm happy to do so if it gets the PR merged ;-)

master/docs/manual/configuration/workers-docker.rst
tardyp: I have implied this token rotation from that github issue: https://github.com/LeastAuthority/txkube/issues/139

master/buildbot/util/kubeclientservice.py
tardyp: yes, reconfigService is optionally deferable. checkConfig is not.

The only thing is to make sure secretManager is available at that time. The other option is to just cache those secrets. I agree this can be done in a later PR if you encounter some performance issues

master/buildbot/util/kubeclientservice.py
```diff
@@ -190,19 +198,37 @@ def __init__(self, kube_config=None):
...
+            req_kwargs['headers'].update(config['headers'])
+
+        bearer = config.get('bearerToken', None)
+        if bearer is not None:
+            bearer = yield self.config.renderSecrets(bearer)
+            req_kwargs['headers']['Authorization'] = "Bearer {token}".format(token=bearer)
+
+        basic = config.get('basicAuth', None)
+        if basic is not None:
+            basic = yield self.config.renderSecrets(basic)
```
tardyp: In python, usually, the basic auth conf is in the form ('login', 'password'). Secret rendering is recursive, so you can write

```python
KubeHardcodedConfig(basic=('user1', Secret('user1_passwd'))
```
So I think we should implement the ":".join() and the base64 encode.
And we should put that logic inside the base config loader class instead of here.

I would rather put getConfig a deferable, so that we can implement this logic, and also implement caching.

master/docs/manual/configuration/workers-docker.rst
tardyp: I agree that gcloud creds provisioning might not be convenient. FWIW This is what I've done in our ansible setup: https://github.com/buildbot/buildbot-infra/blob/master/roles/bb-master/tasks/main.yml#L141

master/buildbot/util/kubeclientservice.py
doudou: > And we should put that logic inside the base config loader class instead of here.

By "base config loader", I'm assuming that you mean HardcodedConfig ?

tardyp: I was thinking about KubeConfigLoaderBase  https://github.com/buildbot/buildbot/blob/15826fcf9bc0e99d0a8640bf02a32911cdaf90fa/master/buildbot/util/kubeclientservice.py#L42

doudou: You know best of course, but given that the authorization headers are specific to the hardcoded config class, I'm not so sure how to integrate them in the base config loader.

Anyways, moving things to getConfig is a can of worms, I've got to refactor the fake kube service to handle that :(

tardyp: right. The core reason is that I think _prepareRequest start to become too large, so maybe we can just split that into several methods a call it a day

doudou: :+1: I'll do that instead.

tardyp:
@doudou Thanks for updating this branch!
the unit tests and coding style checks are failing. can you please fix those ?

doudou:
I'll get started with the style checks ... could you help me with the HTTP service failures ? https://buildbot.buildbot.net/#/builders/13/builds/104/steps/9/logs/stdio

I'm really not sure how to fix those ...

doudou:
I force-pushed to fix the small style errors, but the e.g. flake8 job seem to be still using the old code

For example:
- https://buildbot.buildbot.net/#builders/42/builds/56
- refers the right commit (157b02d)
- but reports an error that does not exist on this commit (https://github.com/buildbot/buildbot/pull/4540/files#diff-4240fcc745ea9f8297809bb45a833282R192)

seankelly:
Check the Properties tab for the `got_revision` property. That is the revision the build is using.

    {"buildbot":"54d4423b233beb7856fbf1d9b4e56af65377b16c"}

It's stuck on the [54d4423](https://github.com/buildbot/buildbot/commit/54d4423b233beb7856fbf1d9b4e56af65377b16c) commit.

tardyp:
The trial errors are a leak from a former test, they appear only when you run the whole test suites. Those issues are pretty challenging to root cause. working on it.

tardyp:
alright, I found the issue, it was an ancient issue triggered by the fact that _doRequest* are now inlineCallbacks.

Thanks for your contribution!

doudou:
> Thanks for your contribution!

Sure. Thanks for buildbot ! :)

doudou:
Just noticed, but getAuthorization is defined in KubeHardcodedConfig and used in KubeClientService. This won't work with the other subclasses, or ?

tardyp:
Oups. Indeed. And this broke our CI for a while..

PR #4542, opened 2019-01-19
Allows more UTF-8 characters in builder names, worker names, and step names

rodrigc:
## Contributor Checklist:

I am reopening this from https://github.com/buildbot/buildbot/pull/3592 because stalebot closed that and I can't reopen it

On Python 3, I was able to use the following:

image

On Python 2, I wasn't able to use a builder with Cyrillic characters.
This is due to a bug in Twisted's logger in Python 2 which does not fully support unicode.
An attempt was made to fix it on Python 2 here: https://twistedmatrix.com/trac/ticket/8864 , but it was reverted.  The Twisted logger has no problems with unicode on Python 3.

### DIFF
--- a/master/buildbot/newsfragments/utf8-identifier.feature
+++ b/master/buildbot/newsfragments/utf8-identifier.feature
@@ -1,0 +1,1 @@
+Identifiers can now contain UTF-8 characters which are not ASCII.  This includes worker names, builder names, and step names.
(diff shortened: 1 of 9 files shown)


p12tic:
Note, breaks python2, should only be merged to Buildbot v2.0.

rodrigc:
@p12tic Actually this change does not break Python2.  One of the tests failed earlier because it was flaky, but that had nothing to do with this change.  I updated the branch, and all the tests now pass, even on Python 2.

p12tic:
As far as I understood, we're relaxing the regexes for the names on both python2 and python3, so a user may try to use UTF-8 on python2 and get strange errors, whereas before he'd get a clear error message from our checks. If that's the case, we do break python2 in certain way :-)

AFAIK 1.7.1 will be released soon, so we'll be able to merge this PR for v2.0 right after that. It'll be just several days of delay, depending on when Pierre cuts the next release, so I think it's worth waiting.

PR #4544, opened 2019-01-19
Fix logwatcher tests

p12tic:
Fixes logwatcher tests as they used `SynchronousTestCase` even though the functions returned a Deferred. This causes any assertion errors to be ignored on python3.


tardyp:
I think you convinced me.

PR #4546, opened 2019-01-20
Reduce duplication in tests by introducing TestReactorMixin

p12tic:
The duplication of `TestReactor` setup and chance of errors there has been reduced by introducing `TestReactorMixin`.

Also, some minor fixes.

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -583,6 +583,7 @@ github
 gitlab
 gitorious
 gitpoller
+globals
 gmail
 gmt
 google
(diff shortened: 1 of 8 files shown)


rodrigc:
Thanks for submitting this patch, but plus make sure you follow all the steps in this procedure:

https://twistedmatrix.com/trac/wiki/TwistedDevelopment#SubmittingaPatch

Otherwise the Twisted devs won't look at this.  Specifically focus on steps 1., 6., 10., and 11.

tardyp:
cool. The cleanup was also no found in several of the tests. I spend a few hours this afternoon about this because removing the hyper tests did trigger this issues.
Then I rebased onver your latest work and it disapear. I think this fix is the real one. Thanks a bunch!

PR #4557, opened 2019-01-23
Unpin CairoSVG as buildbot is now py3 only

yan12125:
## Contributor Checklist:

Upstream an Arch Linux patch https://git.archlinux.org/svntogit/community.git/tree/trunk/cairosvg2.patch?h=packages/buildbot-www

I've used that patch for 2 months. PNG rendering works just fine :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
             env
             python3 -m venv .venv
             . .venv/bin/activate
-            pip install -U pip setuptools
+            pip install -U 'pip<19' setuptools
             pip install -r requirements-ci.txt
             pip install -r requirements-cidocs.txt
             pip install pyinstaller
--- a/www/badges/setup.py
+++ b/www/badges/setup.py
@@ -38,7 +38,7 @@ setup_www_plugin(
     packages=['buildbot_badges'],
     install_requires=[
         'klein',
-        'CairoSVG==1.0.22',  # cairoSVG 2+ is not py2 compatible
+        'CairoSVG',
         'cairocffi', 'Jinja2'
     ],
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             env
             python3 -m venv .venv
             . .venv/bin/activate
-            pip install -U pip setuptools
+            pip install -U 'pip<19' setuptools
```
rodrigc: Please remove this part of your patch.  I removed pyinstaller from requirements-ci.txt here: https://github.com/buildbot/buildbot/commit/093069f6ec727acbbb40da48de5b86e3e2846522#diff-3cd54b3c956e4935722ec3c8a8f09be6

www/badges/setup.py
```diff
@@ -38,7 +38,7 @@
     packages=['buildbot_badges'],
     install_requires=[
         'klein',
-        'CairoSVG==1.0.22',  # cairoSVG 2+ is not py2 compatible
+        'CairoSVG',
```
rodrigc: This is OK

.circleci/config.yml
yan12125: Wow, @tardyp is as fast as light 👍. Let me still answer it: pyinstaller is still used in CircleCI builds: https://github.com/buildbot/buildbot/blob/master/.circleci/config.yml#L32, and there's a broken build due to the referred pip issue: https://circleci.com/gh/buildbot/buildbot/7562. I've subscribed to the pip issue and will revert this change if the pip team fixes it and releases a new version.

BTW, apparently previous builds succeeded as pyinstaller was cached.

PR #4563, opened 2019-01-24
Remove 'u' prefix for unicode strings

rodrigc:
This is not needed since we now support only Python 3.

I generated this patch by doing:

```
2to3-3.7 -w -f unicode .
```


tardyp:
flake8 is unhappy about code formatting for one line appart from that okay from me

rodrigc:
@tardyp: Fixed

PR #4572, opened 2019-01-29
fix issue #3819

### LINKED ISSUE #3819: Waterfall displays wrong builder names
In the waterfall view of my project (using buildbot 0.9.14, updated from 0.8.12), some of the workers columns get the name from other ones. E.g.: the **console view** column "A" points to "http://host:8010/#/builders/14", but the **waterfall view** column "A" points to "http://host:8010/#/builders/10" which is in fact "B".

This happens for the 6 workers coming from a given git project but not for the 4 coming from another one.

mrbauff:
The locale-aware sorting introduced in PR #3488 only sorted the _builder names_. Therefore the axis labels (builder names) did not match the builds in the waterfall display. This PR reverses that change.

Fixes #3819


tardyp:
I guess we should properly sort the whole thing eventually

PR #4576, opened 2019-01-30
Fix RemoteUserAuth.maybeAutLogin consumes bytes object as str

### LINKED ISSUE #4402: util.RemoteUserAuth raises builtins.TypeError
RemoteUserAuth (nginx http basic auth) fails with Type error, because the result of the regex match has values of type `bytes`, e.g.: `{'username': b'foobar'}`
Full stacktrace
```python
          File "/.../python3.5/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
            result = g.send(result)
          File "/.../python3.5/site-packages/buildbot/www/auth.py", line 131, in maybeAutoLogin
            yield self.updateUserInfo(request)
          File "/.../python3.5/site-packages/twisted/internet/defer.py", line 1613, in unwindGenerator 
          File "/.../python3.5/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
          File "/.../python3.5/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
          File "/.../python3.5/site-packages/buildbot/www/resource.py", line 92, in failHttpError
            f.trap(Error)
          File "/.../python3.5/site-packages/twisted/python/failure.py", line 439, in trap
            self.raiseException()
          File "/.../python3.5/site-packages/twisted/python/failure.py", line 467, in raiseException
            raise self.value.with_traceback(self.tb)
          File "/.../python3.5/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
          File "/.../python3.5/site-packages/buildbot/www/resource.py", line 85, in failHttpRedirect
            f.trap(Redirect)
          File "/.../python3.5/site-packages/twisted/python/failure.py", line 439, in trap
            self.raiseException()
          File "/.../python3.5/site-packages/twisted/python/failure.py", line 467, in raiseException
            raise self.value.with_traceback(self.tb)
          File "/.../python3.5/site-packages/buildbot/www/config.py", line 126, in renderIndex
            yield self.config['auth'].maybeAutoLogin(request)
          File "/.../python3.5/site-packages/buildbot/www/auth.py", line 131, in maybeAutoLogin
            yield self.updateUserInfo(request)
          File "/.../python3.5/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
            result = g.send(result)
          File "/.../python3.5/site-packages/buildbot/www/auth.py", line 78, in updateUserInfo
            session.updateSession(request)
          File "/.../python3.5/site-packages/buildbot/www/service.py", line 110, in updateSession
            request.addCookie(cookiename, self.uid, path=b"/",
          File "/.../python3.5/site-packages/buildbot/www/service.py", line 137, in uid
            return jwt.encode(claims, self.site.session_secret, algorithm=SESSION_SECRET_ALGORITHM)
          File "/.../python3.5/site-packages/jwt/api_jwt.py", line 62, in encode
            cls=json_encoder
          File "/usr/lib/python3.5/json/__init__.py", line 237, in dumps
            **kw).encode(obj)
          File "/usr/lib/python3.5/json/encoder.py", line 198, in encode
            chunks = self.iterencode(o, _one_shot=True)
          File "/usr/lib/python3.5/json/encoder.py", line 256, in iterencode
            return _iterencode(o, 0)
          File "/usr/lib/python3.5/json/encoder.py", line 179, in default
            raise TypeError(repr(o) + " is not JSON serializable")
        builtins.TypeError: b'foobar' is not JSON serializable
```
following line [here](https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/auth.py#L127) resolves the problem:
```python
user_info = {k: buildbot.util.bytes2unicode(v) for k, v in res.groupdict().items()}
```
Sorry if this is a duplicate, but I failed to find a ticket for this issue. 
I will create a PR as soon as I have spare time.

cheshirekow:
The headerRegex is held as a bytes object so the output of match() is a bytes object. Prior to this change it was consumed as if it were a str object, resulting in a `TypeError` later on when trying to serialize to JSON.

Closes #4402
Also related to #4299 I think.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/fix_remoteuserauth_typestr.bugfix
+++ b/master/buildbot/newsfragments/fix_remoteuserauth_typestr.bugfix
@@ -1,4 +1,1 @@
-:issue:`4402`
-
-Fix RemoteUserAuth.maybeAutLogin consumes bytes object as str leading to 
-TypeError during JSON serialization. 
+Fix RemoteUserAuth.maybeAutLogin consumes bytes object as str leading to TypeError during JSON serialization. :issue:`4402`
--- a/master/buildbot/newsfragments/fix_remoteuserauth_typestr.bugfix
+++ b/master/buildbot/newsfragments/fix_remoteuserauth_typestr.bugfix
@@ -1,0 +1,4 @@
+:issue:`4402`
+
+Fix RemoteUserAuth.maybeAutLogin consumes bytes object as str leading to 
+TypeError during JSON serialization. 
--- a/master/buildbot/test/unit/test_www_auth.py
+++ b/master/buildbot/test/unit/test_www_auth.py
@@ -13,6 +13,7 @@
 #
 # Copyright Buildbot Team Members
 
+import json
 import mock
 
 from twisted.cred.checkers import InMemoryUsernamePasswordDatabaseDontUse
@@ -117,6 +118,7 @@ class RemoteUserAuth(www.WwwTestMixin, unittest.TestCase):
                          'username': 'rachel',
                          'realm': 'foo.com',
                          'email': 'rachel'})
+        _ = json.dumps(self.request.session.user_info)
 
     @defer.inlineCallbacks
     def test_maybeAutoLogin_no_header(self):
--- a/master/buildbot/test/unit/test_www_auth.py
+++ b/master/buildbot/test/unit/test_www_auth.py
@@ -13,7 +13,6 @@
 #
 # Copyright Buildbot Team Members
 
-import json
 import mock
 
 from twisted.cred.checkers import InMemoryUsernamePasswordDatabaseDontUse
@@ -118,7 +117,6 @@ class RemoteUserAuth(www.WwwTestMixin, unittest.TestCase):
                          'username': 'rachel',
                          'realm': 'foo.com',
                          'email': 'rachel'})
-        _ = json.dumps(self.request.session.user_info)
 
     @defer.inlineCallbacks
     def test_maybeAutoLogin_no_header(self):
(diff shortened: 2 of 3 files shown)


p12tic:
Looks good to me, but it would be great to have a unit test for this bug.

cheshirekow:
@p12tic: a wonderful aspiration, but I know literally nothing about the buildbot codebase so I wouldn't even know where to start. Feel free to pick up this change though if you know where to add it.

master/buildbot/newsfragments/fix_remoteuserauth_typestr.bugfix
```diff
@@ -0,0 +1,4 @@
+:issue:`4402`
+
+Fix RemoteUserAuth.maybeAutLogin consumes bytes object as str leading to
```
p12tic: Could you put entire sentence in one line and then append `( :issue:`4402` )` at the end?

cheshirekow:
@p12tic yeah the build failure was helpful. I can add a check that the output is json serializable in the existing test if that sounds good? 

I will update the newsfragment.

master/buildbot/newsfragments/fix_remoteuserauth_typestr.bugfix
cheshirekow: Done

p12tic:
I think there's no point checking for json encode because it's actually implementation detail of jwt. We'd need to check whether the uid property works, that would be a good test.

I think the version without any test was better. Thanks!

cheshirekow:
Ok I will remove.

Issue #4574, opened 2019-01-30
P4 polling with ticket has incorrect parsing


aisoard:
## Issue

P4 polling with tickets uses the `p4 login -p` command and expect 3 lines or more on stdout, the third line being the ticket number.

However, with my version of the CLI tools from Perforce:

    $ p4 -V
    Rev. P4/LINUX26X86_64/2018.2/1740258 (2018/12/11).

the ticket number appears on the 2nd line! (and there are therefore only 2 lines of stdout):

    $ p4 login -p
    Enter password: 
    D25B979...3741E

## Patch

Here is my proposed patch:

    --- buildbot/changes/p4poller.py	2019-01-29 18:30:41.596740370 -0800
    +++ buildbot/changes/p4poller.py	2019-01-28 17:23:42.110348463 -0800
    @@ -200,7 +200,9 @@
     
         def _parseTicketPassword(self, text):
    -        lines = text.split("\n")
    +        lines = text.splitlines()
    -        if len(lines) < 3:
    -            return None
    -        return lines[2].strip()
    +        return lines[-1].strip()
     
         def _getPasswd(self):
             if self.use_tickets:

To make it short: **I just pick the last line of `stdout` no matter the number of lines.**
I'm not sure if that works on all `p4` versions, and I fell that is not the best way of doing this.

## Version found
Buildbot 1.8.0
Python 2.7.15rc1 & Python 3.6.5

aisoard:
Fixed.

PR #4581, opened 2019-01-31
Fix for issue with p4 ticket parsing

aisoard:
Proposed fix for **P4 polling with ticket has incorrect parsing** https://github.com/buildbot/buildbot/issues/4574.


master/buildbot/test/unit/test_changes_p4poller.py
```diff
@@ -334,7 +334,7 @@ def spawnProcess(pp, cmd, argv, env):
                              ['p4', [b'p4', b'login', b'-p']])
             pp.makeConnection(transport)
             self.assertEqual('pass\n', transport.msg)
-            pp.outReceived('Enter password:\nSuccess:  Password verified.\nTICKET_ID_GOES_HERE\n')
+            pp.outReceived('Enter password:\nTICKET_ID_GOES_HERE\n')
```
seankelly: Does this still work with older versions of p4?

aisoard: It is written so that it does (I index relative to the last line for that purpose). That is, both previous and new test case should pass (I didn't bother duplicating the unit test just for that).

But I do not have an older version of p4 to actually try it in production.

p12tic: I think it's worth duplicating the test to ensure that both new and old versions still work.

aisoard:
I just duplicated the test. I'm trying to write a negative test too (for failed identification) but I have no idea how to run the testsuite.

aisoard:
Rebased on top of current HEAD.

master/buildbot/test/unit/test_changes_p4poller.py
```diff
@@ -344,6 +344,89 @@ def spawnProcess(pp, cmd, argv, env):
...
+            P4Source(p4port=None, p4user=None, p4passwd='pass',
+                     p4base='//depot/myproject/',
+                     split_file=lambda x: x.split('/', 1),
+                     use_tickets=True))
+        self.expectCommands(
+            gpo.Expect('p4', '-P', 'TICKET_ID_GOES_HERE',
+                       'changes', '-m', '1', '//depot/myproject/...').stdout(first_p4changes)
+        )
+
+        class FakeTransport:
```
p12tic: Could we extract this duplicated class to the top of the file? E.g. just above `class TestP4Poller`.

p12tic:
@aisoard: Could you rebase on top of the current master again? During the release of v2.0 we broke the smokes test which is now fixed there.

master/buildbot/test/unit/test_changes_p4poller.py
aisoard: Ah, yes. Just did so.

aisoard:
Ok, I just rebased. Crossing fingers...

aisoard:
@p12tic @seankelly Am I missing something?

p12tic:
I think you don't miss anything. @tardyp is the one who usually merges PRs, but there's several day latency most of the time.

tardyp:
@aisoard thanks for the fix and tests!

Issue #4600, opened 2019-02-08
Database corrupted after interrupting upgrade-master


rjarry:
When interrupting the `upgrade-master` command, the database may end up corrupted. This is especially true for SQLite and MySQL (with MyISAM tables).

PR #4601, opened 2019-02-08
scripts: ignore signals in upgrade-master

rjarry:
This command should not be interrupted to avoid database data (or worse, schema) corruption.

Make sure to trap all possible signals and ignore them. Print an explicit warning so that users are aware that the signal was not lost in translation.

Fixes: #4600

### DIFF
--- a/master/buildbot/newsfragments/upgrade-master-avoid-database-corruption.bugfix
+++ b/master/buildbot/newsfragments/upgrade-master-avoid-database-corruption.bugfix
@@ -1,0 +1,1 @@
+To avoid database corruption, the ``upgrade-master`` command now ignores all signals except ``SIGKILL``. It cannot be interrupted with ``ctrl-c`` (:issue:`4600`).
--- a/master/buildbot/scripts/upgrade_master.py
+++ b/master/buildbot/scripts/upgrade_master.py
@@ -15,6 +15,7 @@
 
 
 import os
+import signal
 import sys
 import traceback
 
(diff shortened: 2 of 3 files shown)


master/buildbot/scripts/upgrade_master.py
```diff
@@ -77,6 +78,19 @@ def upgradeDatabase(config, master_cfg):
...
+        WARNING: ignoring signal %s.
+        This process should not be interrupted to avoid database corruption.
+        If you really need to terminate it, use SIGKILL.
+        """.split())
+        print(msg % signum)
+
+    for signame in ("SIGTERM", "SIGINT", "SIGQUIT", "SIGHUP", "SIGUSR1",
+                    "SIGUSR2", "SIGBREAK", "CTRL_C_EVENT", "CTRL_BREAK_EVENT"):
+        if hasattr(signal, signame):
+            signal.signal(getattr(signal, signame), sighandler)
```
tardyp: I guess you need to catch ValueError to ignore error in windows .

rjarry: I think it would be better to know what signal are supported.

I don't have any Windows machine to test this :(

rjarry: Ok, I found a laptop somewhere in a meeting room. Installed python on it and tested :)

`CTRL_C_EVENT` and `CTRL_BREAK_EVENT` are only usable as `os.kill` arguments. They are not to be used to register a signal handler.

master/buildbot/newsfragments/upgrade-master-avoid-database-corruption.bugfix
```diff
@@ -0,0 +1 @@
+To avoid database corruption, the ``upgrade-master`` command now ignores all signals. It cannot be interrupted with ``ctrl-c`` (:issue:`4600`).
```
p12tic: now ignores all signals -> now ignores all signals except SIGKILL.

Issue #4412, opened 2018-11-07
Hyper workers are not unsubstantiated if they crash in certain conditions


p12tic:
This build https://nine.buildbot.net/#/builders/68/builds/274 disconnected when running make pylint step. The master for some reason did not stop the steps, even though it did register the worker as offline. End of worker log:

```2018-11-07 20:51:27+0000 [HangCheckProtocol,client]  startCommand:shell [id 118964]
2018-11-07 20:51:27+0000 [HangCheckProtocol,client] RunProcess._startCommand
2018-11-07 20:51:27+0000 [HangCheckProtocol,client]  bash -c 'make pylint'
2018-11-07 20:51:27+0000 [HangCheckProtocol,client]   in dir /buildbot/buildbot-job/build (timeout 1200 secs)
2018-11-07 20:51:27+0000 [HangCheckProtocol,client]   watching logfiles {}
2018-11-07 20:51:27+0000 [HangCheckProtocol,client]   argv: ['bash', '-c', 'make pylint']
2018-11-07 20:51:27+0000 [HangCheckProtocol,client]  environment:
        {u'BUILDBOT_TEST_DB_URL': u'sqlite://',
         'BUILDMASTER': 'nine.buildbot.net',
         'BUILDMASTER_PORT': '9989',
         'HOME': '/home/buildbot',
         'HOSTNAME': '57c6c7fa6b79',
         u'HYPER_SIZE': u'm3',
         'LANG': 'en_US.utf8',
         'PATH': u'/buildbot/buildbot-job/build/sandbox/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
         'PWD': u'/buildbot/buildbot-job/build',
         u'SQLALCHEMY': u'latest',
         u'TESTS': u'pylint',
         u'TRAVIS_PULL_REQUEST': u'True',
         u'TWISTED': u'latest',
         'USER': 'buildbot',
         'WORKERNAME': 'hyper27',
         u'branch': u'refs/pull/4411/merge',
         u'builddir': u'/buildbot/buildbot-job',
         u'buildername': u'buildbot-job',
         u'buildnumber': u'274',
         u'codebase': u'buildbot',
         u'commit-description': u"{'buildbot': u'v1.5.0-120-g116fb59'}",
         u'event': u'pull_request',
         u'got_revision': u"{'buildbot': u'116fb592b0ea5c4d42d0ff48f7efac776e5dc256'}",
[...]

PR #4606, opened 2019-02-11
Improve behavior of AbstractLatentWorker by implementing a state machine

p12tic:
Previously the latent worker tracked its state implicitly by looking into the following variables: self.conn, self._substantiation_notifier, self.insubstantiating. As a result, the implementation was seemingly simple, but it didn't handle the edge cases well. For example, in a lot of cases `stop_instance()` was called incorrect number of times and integration tests didn't even check that. The latent worker implementations were less robust as a result (e.g. causing issues like #4412).

This PR switches `AbstractLatentWorker` to a state machine and additionally enables proper tracking of latent worker state in its integration tests. This has led to more robust behavior which is covered by 10 new latent worker integration tests. We now document what the LatentWorker implementations may expect in terms of guarantees of when `start_instance()` and `stop_instance()` is called.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/latent.py
+++ b/master/buildbot/test/fake/latent.py
@@ -42,15 +42,18 @@ class LatentController(SeverWorkerConnectionMixin):
     stop_instance() is executed.
     """
 
+    STATE_NOT_STARTED = 0
+    STATE_STARTING = 1
+    STATE_STARTED = 2
+    STATE_STOPPING = 3
+
     def __init__(self, case, name, kind=None, build_wait_timeout=600, **kwargs):
         self.case = case
         self.build_wait_timeout = build_wait_timeout
         self.worker = ControllableLatentWorker(name, self, **kwargs)
         self.remote_worker = None
 
-        self.starting = False
-        self.started = False
-        self.stopping = False
+        self.state = self.STATE_NOT_STARTED
         self.auto_stop_flag = False
         self.auto_start_flag = False
         self.auto_connect_worker = True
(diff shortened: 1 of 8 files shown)


master/buildbot/test/fake/latent.py
```diff
@@ -42,15 +42,18 @@ class LatentController(SeverWorkerConnectionMixin):
     stop_instance() is executed.
     """
 
+    STATE_NOT_STARTED = 0
```
tardyp: would it make sense to use automat to implement this?

https://pypi.org/project/Automat/

p12tic: I think that using automat would make code less readable. It could probably work for `LatentController`, but in `AbstractLatentWorker` the '.state' member does not represent entire state and we make decisions about the next state taking into account more data than that. I'm probably biased, but in this case procedural paradigm allows easier building of the mental model than a declarative one.

By the way, I saw somewhere within Buildbot state constants being represented in a separate class as `NamedConstant` members. Should I do that here too?

seankelly: Since it's 3.5+, what about using enum?

p12tic: OK, will use enum.

p12tic: Should be fixed now.

p12tic:
@tardyp: Are there any issues or risks with this PR that still need to be addressed before merging?

Issue #4592, opened 2019-02-05
prioritizeBuilders order is not properly enforced


rjarry:
The order specified by the `prioritizeBuilders` function is completely ignored. Here is a minimalist `master.cfg` to reproduce the problem:

```python
from buildbot.plugins import schedulers, steps, util, worker
from buildbot.process import factory

c = BuildmasterConfig = {}

c['buildbotNetUsageData'] = None
c['workers'] = [worker.LocalWorker('local', max_builds=1)]
c['protocols'] = {'null': {}}
c['builders'] = []

BRANCHES = ('master', 'maint-1.x', 'maint-2.x', 'maint-3.x')
CONFIGS = {
    'x86-full': ['for-tests'],
    'x86-ipsec': ['experiental'],
    'x86-customer': ['cust'],
    'x86-doc': [],
    'x86-debug': ['internal'],
    'aarch64-full': ['for-tests'],
    'aarch64-ipsec': ['experiental'],
    'aarch64-customer': ['cust'],
    'aarch64-doc': [],
    'aarch64-debug': ['internal'],
}

FACTORY = factory.BuildFactory(
    steps=[steps.ShellCommand(name='do it', command='sleep 120')])

for branch in BRANCHES:
    for name, tags in CONFIGS.items():
        c['builders'].append(
            util.BuilderConfig(
                name='%s-%s' % (branch, name),
                tags=tags,
                workernames=['local'],
                properties={'branch': branch},
                factory=FACTORY))

LABELS_PRIO = {'for-tests': 0, 'internal': 1024, 'experimental': 5000}
BRANCH_PRIO = {'master': 0}

def sort_builders(buildmaster, builders):
    def _key(builder):
        if builder.config.tags:
            labels_prio = min(LABELS_PRIO.get(t, 512) for t in builder.config.tags)
        else:
            labels_prio = 512
        branch = builder.config.properties.get('branch')
        branch_prio = BRANCH_PRIO.get(branch, 512)
        return (labels_prio, branch_prio)
    builders.sort(key=_key)
    return builders

c['prioritizeBuilders'] = sort_builders

c['schedulers'] = []
for branch in BRANCHES:
    builder_names = []
    for builder in c['builders']:
        if branch == builder.properties.get('branch'):
            builder_names.append(builder.name)
    c['schedulers'].append(
        schedulers.SingleBranchScheduler(
            name=branch, branch=branch, builderNames=builder_names))

c['title'] = 'test'
[...]

PR #4558, opened 2019-01-23
Make undocumented apis private

p12tic:
As discussed in IRC meeting. This PR makes explicit that any undocumented API is internal and subject to change.

PR #4612, opened 2019-02-13
BotMaster and BuildRequestDistributor improvements for scalability

rjarry:
Allow using a custom `BuildRequestDistributor` class instead of the default one.

This is the first step to address bug #4592. The current implementation of the build request distributor is overly complex and suboptimal. However, modifying the algorithm would make the `BuilderConfig.nextBuild` and `BuilderConfig.nextWorker` functions obsolete.

There may be people using them and we cannot break the API. The idea is to propose multiple built-in `BuildRequestDistributor` implementations (will be done in separate commits).

### DIFF
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -204,6 +204,7 @@ class MasterConfig(util.ComparableMixin):
         self.collapseRequests = None
         self.codebaseGenerator = None
         self.prioritizeBuilders = None
+        self.buildRequestDistributorClass = None
         self.multiMaster = False
         self.manhole = None
         self.protocols = {}
@@ -250,6 +251,7 @@ class MasterConfig(util.ComparableMixin):
         "buildCacheSize",
         "builders",
         "buildHorizon",
+        "buildRequestDistributor",
         "caches",
         "change_source",
         "codebaseGenerator",
@@ -472,6 +474,18 @@ class MasterConfig(util.ComparableMixin):
             error("prioritizeBuilders must be a callable")
         else:
             self.prioritizeBuilders = prioritizeBuilders
+
+        # local import to avoid circular imports
+        from buildbot.process.buildrequestdistributor import BuildRequestDistributor
+
+        brd_cls = config_dict.get('buildRequestDistributorClass')
+        if brd_cls is not None and (
+                not inspect.isclass(brd_cls) or
+                not issubclass(brd_cls, BuildRequestDistributor)):
+            error('buildRequestDistributorClass must be a subclass of %s'
+                  % BuildRequestDistributor)
+        else:
+            self.buildRequestDistributorClass = brd_cls
 
         protocols = config_dict.get('protocols', {})
         if isinstance(protocols, dict):
--- a/master/buildbot/newsfragments/config-buildrequestdistributor.feature
+++ b/master/buildbot/newsfragments/config-buildrequestdistributor.feature
@@ -1,0 +1,1 @@
+A custom build request distributor class can now be configured via :bb:cfg:`buildRequestDistributorClass`.
--- a/master/buildbot/process/botmaster.py
+++ b/master/buildbot/process/botmaster.py
@@ -185,6 +185,9 @@ class BotMaster(service.ReconfigurableServiceMixin, service.AsyncMultiService):
         timer = metrics.Timer("BotMaster.reconfigServiceWithBuildbotConfig")
         timer.start()
 
+        # reconfigure build request distributor
+        yield self.reconfigBuildRequestDistributor(new_config)
+
         # reconfigure builders
         yield self.reconfigServiceBuilders(new_config)
 
@@ -195,6 +198,23 @@ class BotMaster(service.ReconfigurableServiceMixin, service.AsyncMultiService):
         # try to start a build for every builder; this is necessary at master
         # startup, and a good idea in any other case
         self.maybeStartBuildsForAllBuilders()
+
+        timer.stop()
+
+    @defer.inlineCallbacks
+    def reconfigBuildRequestDistributor(self, new_config):
+
+        timer = metrics.Timer("BotMaster.reconfigBuildRequestDistributor")
+        timer.start()
+
+        brd_cls = new_config.buildRequestDistributorClass
+        if brd_cls is None:
+            brd_cls = BuildRequestDistributor
+        if not isinstance(self.brd, brd_cls):
+            if self.brd is not None:
+                yield self.brd.disownServiceParent()
+            self.brd = brd_cls(self)
+            self.brd.setServiceParent(self)
 
         timer.stop()
 
--- a/master/buildbot/test/unit/test_config.py
+++ b/master/buildbot/test/unit/test_config.py
@@ -35,6 +35,7 @@ from buildbot import locks
 from buildbot import revlinks
 from buildbot import worker
 from buildbot.changes import base as changes_base
+from buildbot.process import buildrequestdistributor
 from buildbot.process import factory
 from buildbot.process import properties
 from buildbot.schedulers import base as schedulers_base
@@ -65,6 +66,7 @@ global_defaults = dict(
     properties=properties.Properties(),
     collapseRequests=None,
     prioritizeBuilders=None,
+    buildRequestDistributorClass=None,
     protocols={},
     multiMaster=False,
     manhole=None,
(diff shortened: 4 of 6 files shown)


p12tic:
Does it make sense to expose functionality at this low level? We may want to change BuildRequestDistributor in various way and exposing its behaviour will lock it in.

rjarry:
@p12tic I'm not sure how to improve the situation. The current API exposes 4 different functions to control build scheduling:

- `prioritizeBuilders` (which does not work)
- `BuilderConfig.nextBuild` (which I think kind of overlaps with `prioritizeBuilders`)
- `BuilderConfig.nextWorker`
- `BuilderConfig.canStartBuild` (which I also think overlaps with `nextWorker`).

These overlaps are (in my opinion) the main performance problem with build scheduling. One way to make things better would be to replace them with only 2 functions:

- `nextBuild(list of pending breqs)` (global, not per-builder)
- `nextWorker(breq)` (global or per-builder, does not matter)

The problem is that these functions are all entwined into the `BuildRequestDistributor` and `BuildChooser` classes. And I don't see how we can merge `prioritizeBuilders` and `BuilderConfig.nextBuild` into a global `nextBuild` without breaking users config. Same goes for merging `BuilderConfig.nextWorker` and `BuilderConfig.canStartBuild` but to a lesser extent.

That's why my idea was to allow users to get rid of all that complexity and write their own scheduling system. To me the simplest solution would of course be to deprecate `prioritizeBuilders`,  `BuilderConfig.nextBuild` and `BuilderConfig.canStartBuild` and change the API of `BuilderConfig.nextWorker`.

What do you think?

p12tic:
I agree with your observation that the current situation is bad, that the current APIs tie too much into `BuildRequestDistributor` and that the APIs that you propose make sense. My main worry is that if we allow users to override `BuildRequestDistributor`, then naturally we'll need to expose APIs needed for it to work. That will cement our internal architecture in build request distribution and make refactors much harder. Currently we can do whatever we want in this area (see #4558).

We will need almost complete refactor of `BuildRequestDistributor` because:
 - its performance is bad. I know that the Buildbot instance at Unity Technologies did not scale to more than around 1000-2000 of pending build requests. Situation got so bad at times that there was a hard limit that prevented developers from issuing new builds. I think there were similar reports in Buildbot bugtracker. 
 - locks are not handled well and this hides another potentially huge performance problem. Basically, currently `BuildRequestDistributor` does not know anything about locks at all. So whenever a lock becomes released, we must inform the distributor that some builders may run new builds. Unfortunately in the case of renderable builder locks we basically need to tell the `BuildRequestDistributor` to retry build checks on _all_ builders. Right now we don't do that, which results in builds being not run sometimes even if there are free workers.

I was thinking about these problems some time ago and my plan was to introduce a data structure that would store a list of pending buildrequest and for each of them a list of reasons why it doesn't run yet. The result would be that the `BuildRequestDistributor` would know what buildrequests are for sure not runnable and won't spend time on them, potentially resulting in much better performance. On the other hand, if e.g. a lock becomes released or a worker becomes available, then the distributor would immediately know what buildrequests may potentially be runnable, also saving time.

The main point where we could hook into the buildrequest distributor would be where it decides which of the buildrequests that are potentially runnable to start. I think that both of your proposed APIs fit well here: `nextBuild(list of pending breqs)` could decide which of the buildrequests we should try to start build for. `BuilderConfig.nextWorker(breq)` could decide which worker to run buildrequest on. If a worker is not available, then the build request will wait for that worker. If `nextWorker` returns `False`, then we can add a condition not to retry the buildrequest for, say, 10 seconds or a minute. I think we could pass a list of currently available workers so that `nextWorker` does not try to use busy worker for no reason.

What do you think about the above? I think we could potentially even hook the old APIs into the above buildrequest distributor until we release Buildbot v3.0 and can drop them and switch to your proposal.

rjarry:
If we can manage to shoehorn legacy functions into a complete refactor I think that would indeed be the best option. I have virtually zero knowledge on locks. Can you share an overview of the data structure you were talking about ? Especially the part with reasons for not starting builds.

tardyp:
I really like this idea and I think this is the way to go for smooth transition to a scalable API. The current BDR is quite stable, and this does not add maintainance burden to add another one.
We can the deprecate the old one, and maybe try to implement old APIs using it.

We should really think about a proper API wich would enable scalability.
Rather than a 'nextBuild' API, I think it would be wiser to use a 'sortBuild' API.

We can also maintain a list of current buildrequest already pre-sorted. The sort algorithm will only re-sort when new buildrequest arrived, and re-sort is faster than sorting from scratch

We can also run the sort algorithm only once per activity loop, and then pop the buildrequests in order

rjarry:
@tardyp wrote:
> I really like this idea and I think this is the way to go for smooth transition to a scalable API. The current BDR is quite stable, and this does not add maintainance burden to add another one.
We can the deprecate the old one, and maybe try to implement old APIs using it.

After what @p12tic said, I kind of agree that we should avoid make such an API "public". The BRD is an important piece of buildbot and if the API is public, we will have a hard time making radical changes if required.

> Rather than a 'nextBuild' API, I think it would be wiser to use a 'sortBuild' API.

Why not, the concept is the same but as you said, we may have better performance since re-sorting an already sorted list is fast.

In light of the above, how do you think we should expose the "new" BRD and how users may enable it instead of the old one?

How about making it dependent on which configuration items are set: if `sortBuilds` is set, use the new BRD (and ignore+display warning for `prioritizeBuilders`, `BuilderConfig.nextBuild`, `BuilderConfig.nextWorker`, `BuilderConfig.preferedWorkers` and `BuilderConfig.canStartBuild`), else use the old one.

I'm not sure what to do of `BuilderConfig.nextWorker`. Since it's semantics would change with the new BRD (combining `BuilderConfig.canStartBuild`). Maybe we should add another function. What about `BuilderConfig.findWorker(breq)`?

This makes API exposure minimal while preserving current behavior until we think the new BRD is stable enough to make it the default.

tardyp:
It think the BRD maybeStartBuildForBuilders API is good, so we could commit it for 2.0.0 range.
The we can just say we only support the two BRD (SimpleBRD and ScalableBRD)

The simpleBRD has a simple API, which is convenient, and use the current builder based priority methods, and the ScalableBRD provide only a few priority functions, which are master wide.

This will allow much better optimisation, but of course at the price of priority function largely less convenient.

rjarry:
I don't really agree with that. This exposes the BRD classes as public API. I'll try to work on a draft which I'll submit here.

p12tic:
I think current situation shows the danger of exposing build request distributor APIs quite well. `maybeStartBuildForBuilders` was good API until the problems with locks surfaced. To solve them we need to refactor build request distributor and `maybeStartBuildForBuilders` API probably doesn't fit well anymore because in case of renderable locks for example, we would need to start builds for all builders. So even if that API was simple and stable, it does not mean that it's good candidate for exposure, as new information may change the situation completely.

p12tic:
@rjarry wrote:
> Can you share an overview of the data structure you were talking about ?

I don't have a rigorous definition at this point yet. What I was thinking about was something like:

The buildrequest distributor tracks runnable and blocked build requests. Each buildrequest has associated information about what constraints it needs to satisfy before being turned into a build. Such constraints could be a list of locks that need to be taken or the worker(s) a buildrequest may run on and so on. The distributor attemps to start runnable build requests and gets various events that could move buildrequests from runnable state to blocked and vice-versa. As an optimization, each event source has a list of build requests that could be affected, so whenever an event arrives, we don't need to traverse all pending buildrequests.

To be more concrete, some pseudocode follows:

```

class BuildRequestConstraint:
    # represents a constraint for a build request

class BuildRequestConstraintWorker:
    workers # the name of the workers a buildrequest may run on

class BuildRequestConstraintLocks:
    # represents the locks that need to be acquired
    # in case of renderable locks the list is updated just before the buildrequest
    # is turned into a build. If some lock can't be acquired then the list of locks
    # is not re-rendered again until all locks are acquirable.
    locks 

class BuildRequestConstraintDelay:
    time # represents the earliest time a buildrequest may become runnable

class TrackedBuildRequest:
    # represents a build request in buildrequest distributor
    buildrequest # the buildrequest
    constraits # a list of constraits for the build request.
    satisfied_constraints # a list of bool saying whether corresponding 
                          # constraint is satisfied or not 
    runnable # a bool saying whether the buildrequest may be turned into a build or not 
               # (maybe just the above list is sufficient)

class BuildRequestTracker:
    # represents data structure for tracking build requests, 
    # could be just part of BuildRequestDistributor

    blocked # a list of TrackedBuildRequest
    runnable # a list of TrackedBuildRequest

    # a map between worker name and a list of TrackedBuildRequest references
    worker_to_buildrequests 
                            
    # a map between lock name and a list of TrackedBuildRequest references
    lock_to_buildrequests 

    # a list of tuples each containing time when buildrequest 
    # may run and a reference to TrackedBuildRequest
    delay_to_buildrequests 
```

The best thing of the above is that we are not limited by types of constraints that prevent a buildrequest to run, we could potentially add any number of them: the only changes are addition of new constraint type, a new map between events related to the constraint and buildrequests affected by it and some additional code.

In case of a large number of pending build requests it's likely that each constraint could have a large list of affected buildrequests and updating them would take a lot of time on each constraint change (e.g. lock release). However it would still be faster than the current situation when we blindly iterate over all buildrequests only to see that they can't run for one reason or another.

How does the above sound? @rjarry, @tardyp ?

rjarry:
That looks like what I had in mind after looking at the code this morning.

rjarry:
@p12tic, since you will work on the first steps of this, should I close this PR or do you want to reuse it?

p12tic:
I think we can close it.

PR #4613, opened 2019-02-15
Python 2 removed rather than deprecated

ari:

### DIFF
--- a/master/docs/conf.py
+++ b/master/docs/conf.py
@@ -97,10 +97,7 @@ blocdiag_transparency = True
 
 # add a loud note about python 2
 rst_prolog = textwrap.dedent("""\
-.. caution:: Buildbot is deprecating Python 2.7.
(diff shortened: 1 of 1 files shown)


master/docs/conf.py
```diff
@@ -97,10 +97,7 @@
 
 # add a loud note about python 2
 rst_prolog = textwrap.dedent("""\
-.. caution:: Buildbot is deprecating Python 2.7.
-    This is one of the last releases supporting it on the buildmaster.
-    `More info <https://github.com/buildbot/buildbot/issues/4439>`_.
-
+.. caution:: Buildbot no longer supports Python 2.7 on the build master.
```
rodrigc: You are absolutely correct.  Python 2.7 has been *dropped*, not just deprecated.

Can you change this sentence to:

`Buildbot no longer supports Python 2.7 on the buildbot master`,

Issue #4578, opened 2019-01-30
Reboot worker as build step without failing build


jawalking:
As part of my build I need to reboot the remote worker as part of the build step. However every time the worker disconnects from the master the build fails or acts unpredictably. Is there anyway to ignore this brief worker disconnect, or make the master/build factory aware that the worker will be   disconnected?

tardyp:
There is no such feature yet.

Code that handles disconnection is here.

https://github.com/buildbot/buildbot/blob/6f4466eb0251c4671762681521cd074a58d3d3f2/master/buildbot/process/build.py#L574

As you can see it will mark the build retry.

You could overwrite the BuildStep's interrupt command, with a hack like this (untested)

```python
class DisconnectingShellCommand(ShellCommand):
    def interrupt(self, reason):
             self.stopped = True
             self.build.results = 0  # SUCCEED
             if self.cmd:
                 self.cmd.results = lambda: 0
                 d = self.cmd.interrupt(reason)
                 d.addErrback(log.err, 'while cancelling command')
```

vx-i:
I would like to do the same thing with master:

1. Checkout git
2. buildbot checkconfig
3. if OK, restart master

What do you think about? Does it make sense?

tardyp:
@vx-i please create another issue, as this one is about rebooting the worker

vx-i:
@tardyp Done -> #5788.

Issue #4434, opened 2018-11-15
RFC: Support suspend and wake machines running workers on demand


p12tic:
Hey,

I have several local build servers that are rather rarely used but have non-insignificant power usage on idle. To reduce waste of power I currently suspend and resume them manually. Ideally I'd like buildbot to do this: suspend machine when there have been no builds for a specified time and resume it on demand using a wake-on-lan packet, probably from another machine than the master is running.

Does anyone have any ideas and suggestions on how this could be best implemented?

My current plan is as follows:

- There's new `SuspendableWorker` class that overrides `AbstractLatentWorker`.

- There's a new class `SuspendableWorkerGroupManager` (naming is WIP :-) ) that coordinates when to shutdown or wake up a specific machine that hosts multiple `SuspendableWorker`s.

- Multiple `SuspendableWorker` are linked to `SuspendableWorkerGroupManager`. `SuspendableWorker`s and `SuspendableWorkerGroupManager` communicate about what everyone is doing. Once `SuspendableWorkerGroupManager` notices that workers have been idle for specified time, it insubstantiates them to prevent further incoming builds. Once that completes, it puts the machine to sleep. Any substantiations of `SuspendableWorker`s after that will result in `SuspendableWorkerGroupManager` resuming the machine.

- `SuspendableWorkerGroupManager` is extensible by two actions: `suspend` and `wake`.

- `wake` and `suspend` are implemented via two different class hierarchies. The actions are almost completely independent, for example, suspend will run from different host than wake. Thus we can't just extend `SuspendableWorkerGroupManager` as we`d like to mix and match different implementations of `wake` and `suspend`.

- The implementations of `wake` and `suspend` are implemented by talking to hosts directly without reusing any existing worker infrastructure. At least `suspend` will need elevated privileges and using workers will increase the chance of credential leakage.

Does the above architecture make sense? In particular, I'm interested in complications that might potentially arise in multimaster setups. Do I need add a new table for `SuspendableWorkerGroupManager` to the DB? @tardyp, @seankelly ? :-)

seankelly:
This feels like a latent worker, except with terrible startup time compared to a VM or container. But, the flow feels similar:

1. Need to do builds so start the worker (machine).
2. Do builds.
3. Wait some amount of time (`build_wait_timeout`) for additional builds.
4. Stop the worker (and machine).

[...]

PR #4616, opened 2019-02-19
Suspendable latent workers

p12tic:
This PR implements a new latent worker class first suggested in #4434. The new backends allows users to run latent workers on physical infrastructure they control. The machines can be shut down or suspended when not used and woken using wakeonlan functionality. This makes Buildbot much more efficient in test farms with low utilization rate allowing use of old and/or cheap equipment that is not energy-efficient. Most old server or desktop computers use at least 60W when idle, so suspending them when not needed yields cost savings of at least 80$ per year per instance at electricity cost of 0.15$/kWh.

The PR also adds groundwork for implementing on-demand shutdown/reboot of workers requested in #4578.

## Comments and questions for reviewers
In #4434 I have been advised that the functionality may be better implemented as a plugin. Unfortunately after researching the problem it seems that suspendable workers couple too much into Buildbot, so making it a plugin would be counter-productive. Additionally, given limited developer resources of Buildbot ecosystem, it could be a wise idea to include even relatively rarely used functionality into Buildbot proper as long as it does not compromise its architecture. Splitting functionality into multiple repositories increases maintenance overhead a lot due to the need to coordinate releases and also the user confidence is decreased because from their point of view there's much higher risk that the plugin won't be maintained.

Additionally, it looks like that it could make sense to introduce a new concept of a "machine" within Buildbot. A machine is a physical host that runs one or more workers that share processing resources. If a single machine runs a lot of workers the maximum number of builds often needs to be limited. This currently can only be done by using master locks. I think this could be streamlined by introducing "machine locks". If this point makes sense to explore in the future, then the `suspendableMachines` configuration key in this PR should probably be renamed just to `machines` and perhaps the workers should be assigned a machine in their own configuration instead of having a list of all workers as a parameter to `SuspendableMachine`.

## Contributor Checklist:

* [not yet] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [not yet] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/worker/suspendable.py
+++ b/master/buildbot/worker/suspendable.py
@@ -20,22 +20,53 @@ from twisted.internet import reactor
 from twisted.internet import utils
 from twisted.python import log
 
+from twisted.internet import defer
+from twisted.python import log
 from zope.interface import implementer
 
 from buildbot import interfaces
+from buildbot.util import Notifier
 from buildbot.util import service
+from buildbot.worker import AbstractWorker
 from buildbot.worker.latent import AbstractLatentWorker
 
 
 @implementer(interfaces.ISuspendableWorker)
 class SuspendableWorker(AbstractLatentWorker):
 
-    def checkConfig(name, password, **kwargs):
-        AbstractLatentWorker.checkConfig(name, password, **kwargs)
+    def checkConfig(self, name, password, **kwargs):
+        self.machine_manager = None
+        AbstractLatentWorker.checkConfig(self, name, password,
+                                         build_wait_timeout=-1,
+                                         **kwargs)
 
     def reconfigService(self, name, password, **kwargs):
         return AbstractLatentWorker.reconfigService(self, name, password,
+                                                    build_wait_timeout=-1,
                                                     **kwargs)
+
+    @defer.inlineCallbacks
+    def start_instance(self, build):
+        if not self.machine_manager:
+            log.err(('SuspendableWorker {} does not have a manager and could '
+                     'not start build').format(self.name))
+            ret = False
+        else:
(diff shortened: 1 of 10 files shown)


master/buildbot/worker/suspendable.py
```diff
@@ -0,0 +1,364 @@
...
+from buildbot.util import private_tempdir
+from buildbot.util import service
+from buildbot.util.git import getSshArgsForKeys
+from buildbot.util.git import getSshKnownHostsContents
+from buildbot.worker import AbstractWorker
+from buildbot.worker.latent import AbstractLatentWorker
+
+
+@implementer(interfaces.ISuspendableWorker)
+class SuspendableWorker(AbstractLatentWorker):
```
tardyp: I think this logic should be part of the base LatentWorker.

We should be able to attach a machine_manager to all the latent workers.

For example this can be something we use for Kubernetes cluster.

We could use a machine manager to completely shutdown the cluster, and avoid costs during long interruption times.


Same for docker latent worker, one could start an EC2 machine for 4 docker latent worker.

All those use cases for me make sense for the wide range of user while your usecase with ssh and WOL, I am not very convinced will have more than one user (no offense)

p12tic: No offence taken :) Your suggestions make sense, will implement them. What do you think about the possibility of having machine locks? If that makes sense, I would think about it while reworking the PR so that future expansion is easier.

tardyp: > machine locks

why not, but lets implement it with baby steps

p12tic: Yes, sure.

p12tic:
Superseded by #4691.

PR #4621, opened 2019-02-21
Convert bytes to unicode on complete data.

kanarip:
Avoid translating to unicode a possibly interrupted stream of bytes, because packet size.

Otherwise, buildbot may trigger the following exception:

```
 builtins.UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 32766-32767: unexpected end of data 
```

### DIFF
--- a/master/buildbot/process/remotetransfer.py
+++ b/master/buildbot/process/remotetransfer.py
@@ -179,11 +179,23 @@ class StringFileWriter(base.FileWriterImpl):
     def __init__(self):
         self.buffer = ""
 
+        # Preserve the data if converting from bytes to unicode fails because of packet size
+        # truncation, so we may attempt later.
+        self._data_tail = b""
+
     def remote_write(self, data):
-        self.buffer += bytes2unicode(data)
+        self._data_tail += data
+
+        try:
+            self.buffer += bytes2unicode(self._data_tail)
+            self._data_tail = b""
+        except UnicodeError:
+            pass
 
     def remote_close(self):
-        pass
+        # Shouldn't, but could happen.
+        if self._data_tail != b"":
+            self.buffer += bytes2unicode(self._data_tail)
 
 
 class StringFileReader(FileReader):


rodrigc:
@kanarip can you change your patch to use `getincrementaldecoder`?  See: https://github.com/buildbot/buildbot/blob/bfbc0ad19a738328cdf32aac1d93352627c7f889/worker/buildbot_worker/runprocess.py#L192

p12tic:
I think it's worth going a little bit extra than what @rodrigc requested. Can we make `buffer` a property and throw an exception if its value is attempted to be read when `remote_close` has not yet been called? This way we'll avoid confusing silent failures if timing is not right in another part of the code.

kanarip:
> I think that this change is somewhat in the right direction. However, you are changing the
> type of `self.buffer` to bytes, and in `remote_close` changing the type again to unicode `str`.
> 
> I don't think this is a good approach.
> 
> Also, a lot of the tests are failing with your change, so I think this patch needs a bit more work

I don't really understand why the tests started failing, but I suppose @p12tic's comment illustrates buffer is read before remote_close (and would therefore need to be type str).

I suppose a private property _data_tail might be appended data to if bytes2native on it fails, while the next chunk of data supplied would be appended to it, and bytes2native would be executed on the two combined pieces of data, and so forth, and so forth, while remote_close() finally wraps all remaining pieces if any.

Does that approach make sense?

Alternatively, I'll need to understand whether I'm to use some self.command.builder.unicode_encoding (encod**er**?) or whichever other such.

kanarip:
Grmbl. Both the previous version of this patch and this iteration work for me in my deployment -- I'm not understanding why the tests fail or what else I might be able to do about it.

rodrigc:
Your new patch is worse than your original patch.  Can you follow the advice I gave here: https://github.com/buildbot/buildbot/pull/4621#issuecomment-466198457

PR #4626, opened 2019-02-26
Cache claimed counts in BaseLock

p12tic:
`BaseLock` traverses the list of owners every time it wants to know how many `exclusive` and `counting` accesses have been claimed. This is unnecessary, as we could just introduce two numbers to track that and modify them whenever entries are added or removed from the owners.

For tests, the PR adds parameterized library as a dependency which will allow to reduce code duplication in unit tests.

This PR also adds unit tests for `BaseLock` as they were completely missing before.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -49,6 +49,7 @@ api
 apimaster
 app
 appdata
+applicative
 apps
 araujo
 arg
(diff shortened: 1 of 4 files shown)


lopsided98:
I think `parameterized` should be added to `test_deps` in setup.py.

PR #4629, opened 2019-03-01
Feature/expand logs

vibbo:
Hi,
I have a little webUI patch to enable configuration of logs, that should be expanded (unless they are empty).
By default it expands "summary" logs, to honor the current behavior.

---

### DIFF
--- a/www/base/src/app/common/directives/buildsummary/buildsummary.directive.coffee
+++ b/www/base/src/app/common/directives/buildsummary/buildsummary.directive.coffee
@@ -12,7 +12,7 @@ class Buildsummary extends Directive('common')
         }
 
 class _buildsummary extends Controller('common')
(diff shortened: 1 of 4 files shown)


www/base/src/app/common/directives/buildsummary/buildsummary.directive.coffee
```diff
@@ -71,6 +72,9 @@ class _buildsummary extends Controller('common')
         @isSummaryLog = (log) ->
```
tardyp: So this function is not necessary anymore?

vibbo: https://github.com/buildbot/buildbot/blob/27abf3e0c95ddd64f67be3c87d90a323c1560977/www/base/src/app/common/directives/buildsummary/buildsummary.directive.coffee#L74-L78
Still used here, to pop summary log to the top of the step view.

PR #4634, opened 2019-03-05
Improve lock tests

p12tic:
This PR enhances lock integration tests. The new tests comprehensively test different lock types and access modes.

One new test of interest is a test that exercises the behavior of locks during reconfig. Because the botmaster tracks locks by lockid which includes maxCount member, changing maxCount and reconfiguring master will effectively create a new lock. This allows two builds referring to the same lock to claim more than one 'exclusive' access or more 'counting' accesses than specified. This hopefully will be fixed in near future.

Finally, the PR reduces some duplication in integration tests that run a fake master.

## Contributor Checklist:


master/buildbot/test/integration/test_locks.py
```diff
@@ -120,11 +142,292 @@ def test_builder_lock_release_wakes_builds_for_another_builder(self):
...
+        master = yield self.getMaster(config_dict)
+
+        builder_ids = []
+        for i in range(builder_count):
+            builder_ids.append((
+                yield master.data.updates.findBuilderId('builder{}'.format(i))))
+
+        return stepcontrollers, master, config_dict, lock, builder_ids
+
+    @parameterized.expand([
```
tardyp: this looks like lots of new integration tests. Wouldn't that raise the test time too much? Normally, you would not test lots of different cases in integration, but rather do it in unit testing

p12tic: Normally I would agree, but in this case I think that the benefits are worth the overhead. The tested functionality is a core part of buildbot and is also prone to subtle issues. Each of the different cases test different interaction of various components and that interaction is hard to properly cover by unit tests. I believe that given how time consuming is investigation of lock-related issues, if these integration tests catch even a single bug then they would still pay off a year of their overhead for the entire buildbot development team. Furthermore, as time goes by and computing resources become cheaper, the cost of these tests will become relatively smaller and smaller.

On my machine, the new tests slowed down test execution by around 2-4% when running tests on single core.

On buildbot test farm the current test with longest runtime is `pylint` by a wide margin (around 2x), so the new tests should not affect the total latency of the entire test suite.

tardyp: ok

p12tic: Thanks!

Issue #4636, opened 2019-03-06
bytes2unicode tries to convert types other than bytes


vibbo:
Hi,
as far as I can tell, force scheduler WebUI is not showing validation errors properly. It ends up with generic error: "internal error - see logs" in UI and
``builtins.TypeError: coercing to str: need a bytes-like object, dict found``
in twistd.log
This originates in https://github.com/buildbot/buildbot/blob/0727f2c585a6e9cce4e841c8cb955a747ab8e93b/master/buildbot/www/rest.py#L217
when msg is a dict with errors for force fields (as it should).

Solution could be in fixing function bytes2unicode:
https://github.com/vibbo/buildbot/commit/e802c0db3aadc7cceefc0aa082dcaea632b17a09
But I am not sure, whether this can have hidden unintended consequences.

tardyp:
This is not normal to have msg being a dict at that point. Do you have a full stack trace?

vibbo:
> 2019-03-07 14:09:33+0100 [-] While rendering resource:
	Traceback (most recent call last):
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/twisted/internet/defer.py", line 1475, in gotResult
	    _inlineCallbacks(r, g, status)
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/twisted/internet/defer.py", line 1464, in _inlineCallbacks
	    status.deferred.errback()
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/twisted/internet/defer.py", line 501, in errback
	    self._startRunCallbacks(fail)
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/twisted/internet/defer.py", line 568, in _startRunCallbacks
	    self._runCallbacks()
	--- <exception caught here> ---
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	    current.result = callback(current.result, *args, **kw)
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/buildbot/www/resource.py", line 96, in failHttpError
	    f.trap(Error)
	  File "/home/vibbo/bbm/sandbox/lib/python3.5/site-packages/twisted/python/failure.py", line 439, in trap
	    self.raiseException()
[...]

PR #4640, opened 2019-03-07
bytes2unicode now handles all types correctly. #4636

vibbo:


master/buildbot/test/unit/test_util.py
```diff
@@ -280,9 +280,15 @@ def test_None(self):
     def test_bytes2unicode(self):
         rv1 = util.bytes2unicode(b'abcd')
         rv2 = util.bytes2unicode('efgh')
+        rv3 = util.bytes2unicode({})
```
tardyp: why should bytes2unicode not crash when given a dict?

vibbo: See issue: https://github.com/buildbot/buildbot/issues/4636
In short, it is used in that way, at least once.

tardyp:
superseeded by #4641

PR #4649, opened 2019-03-11
Don't allow extra claims on changed locks during reconfig

p12tic:
This PR fixes wrong behavior identified in #4634. Previously, it was possible to change `maxCount` parameter on a lock, reconfigure master and take two exclusive claims on the lock.

Also, locks are now tracked by name, this is a prerequisite for build request distributor improvements.

Finally, several worker tests have been fixed to wait on deferreds, as previously they didn't do that and any failures weren't being reported if tests were ran in multithreaded mode.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/locks.py
+++ b/master/buildbot/locks.py
@@ -41,7 +41,7 @@ class BaseLock:
 
     def __init__(self, name, maxCount=1):
         # Name of the lock
-        self.name = name
+        self.lockName = name
         # Current queue, tuples (waiter, LockAccess, deferred)
         self.waiting = []
         # Current owners, tuples (owner, LockAccess)
@@ -214,14 +214,14 @@ class RealMasterLock(BaseLock):
         self._updateDescription()
 
     def _updateDescription(self):
-        self.description = "<MasterLock({}, {})>".format(self.name,
+        self.description = "<MasterLock({}, {})>".format(self.lockName,
                                                          self.maxCount)
 
     def getLockForWorker(self, workername):
         return self
 
     def updateFromLockId(self, lockid, config_version):
-        assert self.name == lockid.name
+        assert self.lockName == lockid.name
         assert isinstance(config_version, int)
 
         self.config_version = config_version
@@ -232,7 +232,7 @@ class RealMasterLock(BaseLock):
 class RealWorkerLock:
 
     def __init__(self, lockid):
-        self.name = lockid.name
+        self.lockName = lockid.name
         self.maxCount = lockid.maxCount
         self.maxCountForWorker = lockid.maxCountForWorker
         self.config_version = 0
@@ -246,23 +246,23 @@ class RealWorkerLock:
         if workername not in self.locks:
             maxCount = self.maxCountForWorker.get(workername,
                                                   self.maxCount)
-            lock = self.locks[workername] = BaseLock(self.name, maxCount)
+            lock = self.locks[workername] = BaseLock(self.lockName, maxCount)
             self._updateDescriptionForLock(lock, workername)
             self.locks[workername] = lock
         return self.locks[workername]
 
     def _updateDescription(self):
         self.description = \
-            "<WorkerLock({}, {}, {})>".format(self.name, self.maxCount,
+            "<WorkerLock({}, {}, {})>".format(self.lockName, self.maxCount,
                                               self.maxCountForWorker)
 
     def _updateDescriptionForLock(self, lock, workername):
         lock.description = \
-            "<WorkerLock({}, {})[{}] {}>".format(lock.name, lock.maxCount,
+            "<WorkerLock({}, {})[{}] {}>".format(lock.lockName, lock.maxCount,
                                                  workername, id(lock))
 
     def updateFromLockId(self, lockid, config_version):
-        assert self.name == lockid.name
+        assert self.lockName == lockid.name
         assert isinstance(config_version, int)
 
         self.config_version = config_version
--- a/master/buildbot/locks.py
+++ b/master/buildbot/locks.py
@@ -18,6 +18,7 @@ from twisted.internet import defer
 from twisted.python import log
 
 from buildbot import util
+from buildbot.util import service
 from buildbot.util import subscription
 from buildbot.util.eventual import eventually
 
@@ -40,6 +41,8 @@ class BaseLock:
     description = "<BaseLock>"
 
     def __init__(self, name, maxCount=1):
+        super().__init__()
+
         # Name of the lock
         self.lockName = name
         # Current queue, tuples (waiter, LockAccess, deferred)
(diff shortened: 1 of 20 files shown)


master/buildbot/process/botmaster.py
```diff
@@ -28,6 +28,54 @@
 from buildbot.util import service
 
 
+class BotLockTracker:
+    ''' Tracks real lock instances given lock ID and config_version. The latter
+        is used as a way to track most recent properties of real locks.
+
+        This approach is needed because there's no central registry of locks
```
tardyp: Central registry? this looks like a job for the SharedService manager.

Can we implement this as a sharedservice?

http://docs.buildbot.net/latest/developer/utils.html#buildbot.util.service.SharedService

each Lock would become a SharedService.

tardyp:
what should we do with this one?

p12tic:
I simply did not have time to continue on the PR yet.

p12tic:
@tardyp: I don't think we need to go to shared service approach. The BotMaster is already a shared service, the lock tracker is just a subclass that handles the locks. Making the lock tracker a shared service tracker would just complicate things.

Additionally, the locks themselves can't probably be shared services, at least as I understand now. New lock access can appear anywhere in the configuration, and can also be interpolated. This means we don't know what locks the configuration will attempt to take and when. It currently makes sense to make the locks just lightweight objects and track them as part as BotMaster.

master/buildbot/test/fake/botmaster.py
```diff
@@ -24,22 +25,16 @@ class FakeBotMaster(service.AsyncMultiService):
...
-        # if the master.cfg file has changed maxCount= on the lock, the next
-        # time a build is started, they'll get a new RealLock instance. Note
-        # that this requires that MasterLock and WorkerLock (marker) instances
-        # be hashable and that they should compare properly.
-        return self.locks[lockid]
-
-    def getLockFromLockAccess(self, access):
-        return self.getLockByID(access.lockid)
+    def getLockByID(self, lockid, config_version):
+        return self.lock_tracker.getLockByID(lockid, config_version)
```
tardyp: BotMaster is not a SharedService, it is a BuildbotService.
SharedService are used to manage singletons. The only instance of SharedService now are HttpClientService and its fake.

To use it, you just have to change this line to:
```suggestion
        return SharedService.getService(lockid.lockClass, self, lockid)
```

then add SharedService parent class to RealWorkerLock and RealMasterLock

p12tic: Okay, I completely misunderstood how exactly SharedService works. It's much clearer now. I've rewritten the code to use `SharedService` now.

master/buildbot/process/botmaster.py
```diff
@@ -28,7 +28,53 @@
...
+        ''' Convert a Lock identifier into an actual Lock instance.
+            @lockid: a locks.MasterLock or locks.WorkerLock instance
+            @config_version: The version of the config from which the list of locks has been
+                acquired by the downstream user.
+            @return: a locks.RealMasterLock or locks.RealWorkerLock instance
+
+            The real locks are tracked using lock ID and config_version. The latter is used as a
+            way to track most recent properties of real locks.
+
+            This approach is needed because there's no central registry of lock access instances
```
tardyp: well. now there is

master/buildbot/process/botmaster.py
```diff
@@ -28,7 +28,53 @@
...
+            that are used within a Buildbot configuration. All lock accesses bring all lock
```
tardyp: Ah.. I now understand your point

```suggestion
            that are used within a Buildbot master.cfg (like there is for e.g c['builders']). All lock accesses bring all lock
```

master/buildbot/locks.py
```diff
@@ -206,31 +209,40 @@ def isOwner(self, owner, access):
...
-        self.maxCount = lockid.maxCount
-        self.maxCountForWorker = lockid.maxCountForWorker
+    def __init__(self, name):
+        super().__init__()
+
+        # the caller will want to call updateFromLockId after initialization most of the time
+        self.lockName = name
+        self.maxCount = None
+        self.maxCountForWorker = None
+        self.config_version = 0
```
tardyp: why not use self.config_version = -1 to force updateFromLockId at creation, without needing the new hasService api

master/buildbot/newsfragments/locks-reconfig-no-extra-claims.bugfix
```diff
@@ -0,0 +1 @@
+Locks will no longer allow extra counting or exclusive claims to be taken if the `maxCount` parameter is changed and the master is reconfigured.
```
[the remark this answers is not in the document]
p12tic: Previously you could have `MasterLock(1)` and it could be taken twice concurrently during reconfig. How the description could be improved?

master/buildbot/locks.py
p12tic: Fixed now.

master/buildbot/process/botmaster.py
p12tic: Fixed now.

master/buildbot/locks.py
```diff
@@ -206,31 +209,40 @@ def isOwner(self, owner, access):
...
+class RealWorkerLock(service.SharedService):
 
-    def __init__(self, lockid):
-        self.name = lockid.name
-        self.maxCount = lockid.maxCount
-        self.maxCountForWorker = lockid.maxCountForWorker
+    def __init__(self, name):
+        super().__init__()
+
+        # the caller will want to call updateFromLockId after initialization most of the time
```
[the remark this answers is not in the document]
p12tic: Fixed now.

master/buildbot/newsfragments/locks-reconfig-no-extra-claims.bugfix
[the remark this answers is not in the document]
p12tic: Fixed now.

master/buildbot/locks.py
```diff
@@ -294,7 +309,7 @@ def __init__(self, lockid, mode, _skipChecks=False):
             assert mode in ['counting', 'exclusive']
 
 
-class BaseLockId(util.ComparableMixin):
+class BaseLockId:
```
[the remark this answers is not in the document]
p12tic: Where do we do that? I think we need a test to prevent similar changes from breaking buildbot.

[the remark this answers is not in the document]
tardyp: ComparableMixin allows to specify which attributes are important for comparaison with ==
It is recursive so that if an item specifies locks in its list of compare_attrs, then it will detect a change based on name change or maxCount. If you are not using this mixin, then == by default compare the id(), then it will always declare to instance different. Please have a look at the source code for more understanding. It is a very neat feature of buildbot.

master/buildbot/newsfragments/locks-reconfig-no-extra-claims.bugfix
```diff
@@ -0,0 +1 @@
+Locks will no longer allow being acquired more times than the `maxCount` parameter if this parameter is changed during master reconfiguration.
```
tardyp: much better thanks

master/buildbot/locks.py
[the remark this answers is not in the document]
p12tic: Reverted the commit.

p12tic:
@tardyp: I reverted the ComparableMixin removal.

p12tic:
@tardyp I think this PR is good to go.

PR #4652, opened 2019-03-11
fix use of verify and debug parameters for the HttpStatusPush reporter

aureliendavid:
to skip ssl checks, `verify` is set in the constructor but not passed to the `HTTPClientService` for the `HttpStatusPush` reporter so it didn't work

same for the `debug` parameter

## Contributor Checklist:

* [✓] I have created a file in the `master/buildbot/newsfragment` directory (and read the `README.txt` in that directory)
* [✓] I have updated the appropriate documentation


aureliendavid:
argh I thought that for such a minor change the tests wouldn't be impacted.. classic mistake

I'm not sure why the `fakehttpclientservice` failed without the extra parameters, but it looks like it works when adding them

PR #4653, opened 2019-03-11
steps: Add GitTag buildstep

chapipo:
Add a buildstep to be able to tag a local Git repository.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -183,7 +183,7 @@ class Git(Source, GitStepMixin):
         self.stdio_log = self.addLogForRemoteCommands("stdio")
 
         try:
-            gitInstalled = yield self.checkBranchSupport()
+            gitInstalled = yield self.checkFeatureSupport()
 
             if not gitInstalled:
                 raise WorkerTooOldError("git is not installed on worker")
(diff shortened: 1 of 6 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -697,3 +697,78 @@ def _doPush(self):
...
+
+        if self.messages and not isinstance(self.messages, list):
+            bbconfig.error('GitTag: messages should be a list')
+
+    @defer.inlineCallbacks
+    def run(self):
+        self.stdio_log = yield self.addLog("stdio")
+        gitInstalled = yield self.checkTagSupport()
+
+        if not gitInstalled:
```
p12tic: Coverage says this branch is not covered. Could we have a small unit test for this condition?

chapipo:
I don't know if the adding the function `checkTagSupport` in `GitStepMixin` is the good way. Maybe we could rename the `checkBranchSupport` to something else ?

master/buildbot/steps/source/git.py
chapipo: Yeah I spotted it, I will do it and add some more case to current unit test to cover identical branch in other buildstep.

p12tic:
Indeed, I missed that. Since #4558 we can freely rename internal APIs. I think `checkFeatureSupport` could be a candidate for a better name.

chapipo:
Function `checkBranchSupport`renamed to `checkFeatureSupport` and unit tests added.

PR #4656, opened 2019-03-12
Remove extraneous `import os` in buildbot_tac.tmpl.

akath20:
This is a very small change of removing an extraneous import in `Buildbot_tac.tmpl`.

The detail provided in this description are very minimal because there isn't much that happened here, but please do let me know if there is anything additional I should do.


tardyp:
thanks!!

PR #4657, opened 2019-03-13
display owners instead of owner in the UI

tardyp:
owners also reflects the owners from changes


YngveNPettersen:
You might want to check other users/setters of "owner"/"owners" as well, for example in the ForcedScheduler

tardyp:
> You might want to check other users/setters of "owner"/"owners" as well, for example in the ForcedScheduler

I did, and it should work as far as I read the code

PR #4658, opened 2019-03-13
Remove uses of future.utils imports.

mokibit:
Remove uses of future.utils imports.


tardyp:
+1 for a future removal of future in setup.py!

PR #4667, opened 2019-03-20
Fix check to make it more clear that bytes2unicode is just for bytes

rodrigc:


master/buildbot/www/rest.py
```diff
@@ -212,7 +212,7 @@ def renderJsonRpc(self, request):
 
         def writeError(msg, errcode=399,
                        jsonrpccode=JSONRPC_CODES["internal_error"]):
-            if not isinstance(msg, dict):
+            if isinstance(msg, bytes):
                 msg = bytes2unicode(msg)
```
tardyp: well, in this case, why not just use `msg.decode()` ?

rodrigc: I wanted to keep the change simple.  It is more correct to check for the type `bytes` than `dict` in this case.

In terms of replacing `bytes2unicode()` with straight `decode()`, that can be done throughout the buildbot tree in a separate pass.  Since buildbot master supports Python 3.5+ that is OK to do, since the default encoding for `bytes.decode()` is utf-8: https://docs.python.org/3.5/library/stdtypes.html#bytes.decode

This is different from Python 2.7:
https://docs.python.org/2/library/stdtypes.html#str.decode

where the default encoding is the one returned by https://docs.python.org/2/library/sys.html#sys.getdefaultencoding which was not always guaranteed to be utf-8

PR #4672, opened 2019-03-21
Add a password store secrets provider

alucryd:
Hi guys, needed to access passwords from a password store so I wrote a plugin just for that. Hope you'll find it helpful and acceptable.

### DIFF
--- a/master/buildbot/secrets/providers/passwordstore.py
+++ b/master/buildbot/secrets/providers/passwordstore.py
@@ -17,7 +17,6 @@ password store based provider
 """
 
 import os
-import subprocess
 from pathlib import Path
 
 from buildbot import config
--- a/master/buildbot/secrets/providers/passwordstore.py
+++ b/master/buildbot/secrets/providers/passwordstore.py
@@ -1,0 +1,68 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+"""
+password store based provider
+"""
+
+import os
+import subprocess
+from pathlib import Path
+
+from buildbot import config
+from buildbot.secrets.providers.base import SecretProviderBase
+
+
+class SecretInPass(SecretProviderBase):
+    """
+    secret is stored in a password store
+    """
+    name = "SecretInPass"
+
+    def checkPassIsInPath(self):
+        if not any([(Path(p) / "pass").is_file() for p in os.environ["PATH"].split(":")]):
+            config.error("pass does not exist in PATH")
+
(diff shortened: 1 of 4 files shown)


master/buildbot/secrets/providers/passwordstore.py
```diff
@@ -0,0 +1,68 @@
...
+            self._env["PASSWORD_STORE_GPG_OPTS"] = "--passphrase %s" % gpgPassphrase
+        if dirname:
+            self._env["PASSWORD_STORE_DIR"] = dirname
+
+    def get(self, entry):
+        """
+        get the value from pass identified by 'entry'
+        """
+        try:
+            proc = subprocess.run(
```
tardyp: I am worried that this is synchronous call. This will stop the whole buildbot process until the pass process finishes.
We should use async for that:
https://twistedmatrix.com/documents/12.1.0/core/howto/process.html

which will need to refactor the whole code to async

Another way to make this acceptable is to implement a cache, so that the pass is only retrieved once near boot

alucryd: I didn't go the async way because even on a modest server this takes less than one second, but I'll rewrite this to be asynchronous.

Not so keen on having a cache, while the file based provider is dedicated to buildbot, the pass one can very well contain a lot more passwords than what is actually needed by buildbot, and we'd also need either a ttl or an invalidation mechanism so we don't have to reload buildbot with every password update.

tardyp: less than one second is quite a lot actually. If you look at the GCE Latent Worker PR which is ongoing: https://github.com/buildbot/buildbot/pull/4663

It is querying the secret manager for every request to GCE, so even 400ms will slow down a lot buildbot. Doing is sync will additionaly make the UI feel unresponsive.

Probably the cache should be done generically, and not at the secret drivers, indeed.

master/buildbot/test/unit/test_secret_in_passwordstore.py
```diff
@@ -0,0 +1,102 @@
...
+        with patch.object(Path, "is_file", return_value=True):
+            self.tmp_dir = self.createTempDir("temp")
+            self.srvpass = SecretInPass("password", self.tmp_dir)
+            yield self.srvpass.startService()
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        yield self.srvpass.stopService()
+
+    def testCheckConfigSecretInPassService(self):
```
p12tic: Can we use `test_check_config_secret_in_pass_service` naming? Over 90% of buildbot tests use this approach.

master/buildbot/secrets/providers/passwordstore.py
doudou: > It is querying the secret manager for every request to GCE, so even 400ms will slow down a lot buildbot. Doing is sync will additionaly make the UI feel unresponsive.

For the record, it's not.

GCE auth is providing us with an access token that is valid for an hour. Just the round trip to google servers would already take ~1s for authentication otherwise.

alucryd: Rewrote using Twisted's getProcessOutput. There's the option of using getProcessOutputAndValue as well, please let me know if you'd prefer that. AFAICT getProcessOutput raises IOError when stderr receives sth so it should be enough in this case.

master/buildbot/test/unit/test_secret_in_passwordstore.py
alucryd: Sure, I went with camel because the file provider test used that case, I prefer snake in Python anyway!

alucryd:
One thing I wasn't sure about, getProcessOutput returns bytes instead of a string, so I went ahead and decoded it. What's the best practice with secrets, should I just return the byte output?

p12tic:
> What's the best practice with secrets, should I just return the byte output?

In this case existing APIs contrain what we can do. All other secret providers return a string.

tardyp:
thanks!

PR #4683, opened 2019-03-27
Don't interpret zero epoch differently in DB layer

p12tic:
Currently we don't convert dates that evaluate to zero between epoch and datetime. This leads to inconsistent data as APIs expecting datetime now must handle None, int==0 and datetime objects correctly. This PR always compares epoch to None, so the downstream code only needs to handle None and datetime. In addition to that, we can use fake reactors freely in test code without having to advance them out of the initial zero time.

This change is safe, because there's no way to actually send zero epoch to the database currently if it means something like None, i.e. not set epoch. Any 0 epochs in old databases will be converted to 1970-01-01 00:00 datetime, but I believe this is completely insignificant issue.

### DIFF
--- a/master/buildbot/test/unit/test_data_masters.py
+++ b/master/buildbot/test/unit/test_data_masters.py
@@ -249,7 +249,8 @@ class Master(TestReactorMixin, interfaces.InterfaceTests, unittest.TestCase):
         yield self.rtype.expireMasters(_reactor=clock)
         master = yield self.master.db.masters.getMaster(14)
         self.assertEqual(master, dict(id=14, name='other',
-                                      active=False, last_active=None))
+                                      active=False,
+                                      last_active=epoch2datetime(0)))
         self.rtype._masterDeactivated. \
             assert_called_with(14, 'other')
 
--- a/master/buildbot/test/unit/test_db_buildsets.py
+++ b/master/buildbot/test/unit/test_db_buildsets.py
@@ -176,7 +176,8 @@ class Tests(interfaces.InterfaceTests):
                                       reason='rsn', sourcestamps=[234],
                                       submitted_at=datetime.datetime(1978, 6, 15, 12, 31, 15,
                                                                      tzinfo=UTC),
-                                      complete=False, complete_at=None, results=-1,
+                                      complete=False,
+                                      complete_at=epoch2datetime(0), results=-1,
                                       bsid=91,
                                       parent_buildid=None, parent_relationship=None))
 
--- a/master/buildbot/test/unit/test_db_buildsets.py
+++ b/master/buildbot/test/unit/test_db_buildsets.py
@@ -143,25 +143,6 @@ class Tests(interfaces.InterfaceTests):
         return self.do_test_getBuildsetProperties(91, [], dict())
 
     @defer.inlineCallbacks
-    def test_getBuildset_incomplete_None(self):
-        yield self.insertTestData([
-            fakedb.Buildset(id=91, complete=0,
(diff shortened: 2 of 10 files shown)


master/buildbot/test/unit/test_db_buildsets.py
```diff
@@ -175,7 +175,8 @@ def test_getBuildset_incomplete_zero(self):
                                       reason='rsn', sourcestamps=[234],
                                       submitted_at=datetime.datetime(1978, 6, 15, 12, 31, 15,
                                                                      tzinfo=UTC),
-                                      complete=False, complete_at=None, results=-1,
+                                      complete=False,
+                                      complete_at=epoch2datetime(0), results=-1,
```
tardyp: why not None?

master/buildbot/test/unit/test_data_masters.py
```diff
@@ -247,7 +247,8 @@ def test_expireMasters(self):
         yield self.rtype.expireMasters(_reactor=clock)
         master = yield self.master.db.masters.getMaster(14)
         self.assertEqual(master, dict(id=14, name='other',
-                                      active=False, last_active=None))
+                                      active=False,
+                                      last_active=epoch2datetime(0)))
```
tardyp: why not none?

tardyp:
looks good, but the epoch2datetime(0) looks slike a smell

master/buildbot/test/unit/test_db_buildsets.py
p12tic: Because 0 comes from the fake reactor - it's still a correct date/time, the only thing special with it is that it's the default start time for fake reactor. `None` would mean that the date is not set altogether which is not the case.

master/buildbot/test/unit/test_data_masters.py
p12tic: Replied above.

p12tic:
I think that in cases when we send datetimes as epoch in the tests, then we should compare them to data at the same "layer". That is, if we send e.g. epoch==12341234, then we should use `epoch2datetime(12341234)` in comparisons, because `datetime.datetime(whatever date, time corresponds to 12341234` is very non-intuitive. In most tests with reactor, implicitly send datetimes as epoch just by using `reactor.seconds`, that's why I used epoch2datetime(0) in the above cases. It's not obvious that `epoch2datetime(0) == datetime.datetime(1970, 1, 1, 0, 0, 0)`

master/buildbot/test/unit/test_db_buildsets.py
tardyp: yes, but when complete=False, the complete_at should be None, not 0.
So I guess this is hidding something not good.
We have to differentiate null values (unset) to 0 value (start of times)

p12tic: Indeed, I completely missed that point. Agreed, needs investigation.

p12tic: @tardyp: Alright, so what's happening here is that we intentionally push `complete=0, complete_at=0` to the DB and check what `self.db.buildsets.getBuildset()` returns. Do we want to perform cleanup at the DB layer to return data that makes sense even if the data in the DB doesn't? As far as I see there's no other cases of such cleanup being performed.

tardyp: This test verifies that the data api converts zero value to null value. So as we are removing this feature, I would remove the test case.

p12tic: Fixed in 543ad404de5.

PR #4689, opened 2019-03-30
Render GitHub change hook secret

yan12125:
## Contributor Checklist:

Mirroring #4561 for GitHub :)

### DIFF
--- a/master/buildbot/newsfragments/render-github-change-hook-secret.feature
+++ b/master/buildbot/newsfragments/render-github-change-hook-secret.feature
@@ -1,0 +1,1 @@
+The GitHub change hook secret can now be rendered.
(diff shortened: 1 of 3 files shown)


master/buildbot/newsfragments/render-github-change-hook-secret.bugfix
```diff
@@ -0,0 +1 @@
+The GitHub change hook secret is now rendered correctly.
```
p12tic: Can we reword the release note to something like `The GitHub change hook secret can now be rendered`? It's not that the rendering was incorrect, it was not supported at all.

yan12125: Thanks! Fixed the news fragment as well as renaming this file from `.bugfix` to `.feature`.

tardyp:
thanks for your PR!

PR #4690, opened 2019-03-30
Basic reporter for zulip. 

mayankmadan:
https://github.com/zulip/zulip/issues/5639

## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/zulip.py
+++ b/master/buildbot/reporters/zulip.py
@@ -1,22 +1,39 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+
 from twisted.internet import defer
 
 from buildbot import config
-from buildbot.reporters import utils
 from buildbot.reporters.http import HttpStatusPushBase
 from buildbot.util import httpclientservice
 from buildbot.util.logger import Logger
 
-logger = Logger()
+log = Logger()
 
 
 class ZulipStatusPush(HttpStatusPushBase):
     name = "ZulipStatusPush"
+    neededDetails = dict(wantProperties=True)
 
     def checkConfig(self, endpoint, token, stream=None, **kwargs):
         if not isinstance(endpoint, str):
             config.error("Endpoint must be a string")
         if not isinstance(token, str):
             config.error("Token must be a string")
+        super().checkConfig(**kwargs)
 
     @defer.inlineCallbacks
     def reconfigService(self, endpoint, token, stream=None, **kwargs):
@@ -27,15 +44,9 @@ class ZulipStatusPush(HttpStatusPushBase):
         self.token = token
         self.stream = stream
 
-    def buildStarted(self, key, build):
-        return self.send(build, key[2])
-
-    def buildFinished(self, key, build):
-        return self.send(build, key[2])
-
     @defer.inlineCallbacks
-    def getBuildDetails(self, build, event):
-        yield utils.getDetailsForBuild(self.master, build, wantProperties=True)
+    def send(self, build):
+        event = ("new", "finished")[0 if build["complete_at"] is None else 1]
         jsondata = dict(event=event, buildid=build["buildid"], buildername=build["builder"]["name"], url=build["url"],
                         project=build["properties"]["project"][0])
         if event == "new":
@@ -43,16 +54,12 @@ class ZulipStatusPush(HttpStatusPushBase):
         elif event == "finished":
             jsondata["timestamp"] = build["complete_at"]
             jsondata["results"] = build["results"]
-        return jsondata
 
-    @defer.inlineCallbacks
-    def send(self, build, event):
-        jsondata = yield self.getBuildDetails(build, event)
         if self.stream is not None:
             url = "/api/v1/external/buildbot?api_key={}&stream={}".format(self.token, self.stream)
         else:
             url = "/api/v1/external/buildbot?api_key={}".format(self.token)
-        res = yield self._http.post(url, json=jsondata)
-        if res.code != 200:
-            content = yield res.content()
-            logger.error("{}: Error pushing build status to Zulip: {}".format(res.code, content))
+        response = yield self._http.post(url, json=jsondata)
+        if response.code != 200:
+            content = yield response.content()
+            log.error("{}: Error pushing build status to Zulip: {}".format(response.code, content))
--- a/master/buildbot/reporters/zulip.py
+++ b/master/buildbot/reporters/zulip.py
@@ -61,4 +61,4 @@ class ZulipStatusPush(HttpStatusPushBase):
         response = yield self._http.post(url, json=jsondata)
         if response.code != 200:
             content = yield response.content()
-            log.error("{code}: Error pushing build status to Zulip: {content}".format(code=response.code, content=content))
+            log.error("{code}: Error pushing build status to Zulip: {content}", code=response.code, content=content)
--- a/master/buildbot/reporters/zulip.py
+++ b/master/buildbot/reporters/zulip.py
@@ -8,6 +8,7 @@ from buildbot.util.logger import Logger
 
 logger = Logger()
 
+
 class ZulipStatusPush(HttpStatusPushBase):
     name = "ZulipStatusPush"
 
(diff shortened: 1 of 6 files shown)


master/buildbot/reporters/zulip.py
```diff
@@ -0,0 +1,58 @@
+from twisted.internet import defer
```
p12tic: A copyright header is missing. You can copy it from other python files in the repository, such as buildbot/master/buildbot/changes/github.py.

master/buildbot/reporters/zulip.py
```diff
@@ -0,0 +1,58 @@
...
+
+from buildbot import config
+from buildbot.reporters import utils
+from buildbot.reporters.http import HttpStatusPushBase
+from buildbot.util import httpclientservice
+from buildbot.util.logger import Logger
+
+logger = Logger()
```
p12tic: Can we use `log` name like on other reporters?

master/buildbot/reporters/zulip.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class ZulipStatusPush(HttpStatusPushBase):
+    name = "ZulipStatusPush"
```
p12tic: Was there a particular reason why more of the `HTTPStatusPush` functionality has not been used? Can Zulip support additional properties like those supported in `utils.getDetailsForBuild`? If that's the case, I think it would be great if we allowed to send that data too.

master/buildbot/reporters/zulip.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def checkConfig(self, endpoint, token, stream=None, **kwargs):
```
p12tic: super().checkConfig is missing

master/buildbot/reporters/zulip.py
```diff
@@ -0,0 +1,58 @@
...
+    @defer.inlineCallbacks
+    def reconfigService(self, endpoint, token, stream=None, **kwargs):
+        super().reconfigService(**kwargs)
+        self._http = yield httpclientservice.HTTPClientService.getService(
+            self.master, endpoint,
+            debug=self.debug, verify=self.verify)
+        self.token = token
+        self.stream = stream
+
+    def buildStarted(self, key, build):
```
p12tic: I think you should be able to override `wantedDetails` class variable to get the needed data to be sent to `send()`. This way you wouldn't need to override `buildStarted` and `buildFinished` and extract build details yourself in `send()`.

p12tic:
Looks promising as a first step, thanks for looking into this! In order to land the PR, we will need unit tests and documentation.

master/buildbot/reporters/zulip.py
p12tic: For an example, look at `GerritVerifyStatusPush`.

master/buildbot/reporters/zulip.py
mayankmadan: Sorry about that. Rushed into it and forgot about changing the names

p12tic:
The code looks good now, thanks!

mayankmadan:
> Code looks great! The only things missing are tests and documentation.
I was looking at other tests to get a better idea but I'm a little confused.
https://github.com/buildbot/buildbot/blob/78189a18c43b9ff4dd02a8e89a5ee07b969da77a/master/buildbot/test/unit/test_reporters_hipchat.py#L126
What does the expect function do? Does it fill up a list with the expected http response?

mayankmadan:
It doesn't really send a lot of information about the build right now but that can be easily added(I was not sure what should be sent).

mayankmadan:
@p12tic Sorry about all the force pushes. I was having some trouble with my local environment. It should be good to go now

master/docs/manual/configuration/reporters.rst
```diff
@@ -1364,3 +1364,49 @@ This property must be a list of dictionaries, containing ``change_id`` and ``rev
...
+.. py:class:: ZulipStatusPush(endpoint, token, stream=None)
+
+    :param string endpoint: URL of your Zulip server.
+    :param string token: Private API token
+    :param string stream: The stream in which the build status is to be sent. Defaults to None
+
+
+.. note::
+
+   A private message is sent if steam is set to None.
```
p12tic: steam: Typo

master/docs/relnotes/index.rst
```diff
@@ -10,6 +10,25 @@ Release Notes
 
 .. towncrier release notes start
 
+Buildbot ``2.1.1-dev87`` ( ``2019-04-02`` )
```
p12tic: You don't need to commit release notes, that's done only when actual release goes out. Could you revert this file back?

The documentation build process is inconvenient in this regard, I opened #4695 for that.

master/buildbot/test/unit/test_reporter_zulip.py
```diff
@@ -0,0 +1,90 @@
...
+                "buildid": 20,
+                "buildername": "Builder0",
+                "url": "http://localhost:8080/#builders/79/builds/0",
+                "project": "testProject",
+                "timestamp": 1554161923
+            })
+        self.sp.buildStarted(('build', 20, 'new'), build)
+
+    @defer.inlineCallbacks
+    def test_build_finished(self):
```
p12tic: I think we also need tests for the following cases:
 - when stream is `None`
 - when the http request does not succeed, verify that we log an error
 - when either `endpoint` or `stream` parameters aren't strings, verify that config error is thrown.

p12tic:
Looks almost done, thank you! Don't worry about force-pushes, it's preferred that the history does not include mistakes as they're only relevant during the PR phase, so it does not make sense to include that into project git history which is permanent.

master/buildbot/test/unit/test_reporter_zulip.py
p12tic: For tests, if you hate duplicating code, you could look into the `parameterized` library, see e.g. https://github.com/buildbot/buildbot/blob/053d6a8a0ad56221ad9fe1908f4cea5df81b1a5d/master/buildbot/test/integration/test_locks.py#L314.

mayankmadan:
@p12tic I'm not sure about the parameterized library since the test is not to be tested with different values. Won't it be better to just store the json object in a dictionary and pass it wherever needed?

master/buildbot/reporters/zulip.py
```diff
@@ -0,0 +1,67 @@
...
+            jsondata["results"] = build["results"]
+        if self.stream is not None:
+            url = "/api/v1/external/buildbot?api_key={}&stream={}".format(self.token, self.stream)
+        else:
+            url = "/api/v1/external/buildbot?api_key={}".format(self.token)
+        response = yield self._http.post(url, json=jsondata)
+        if response.code != 200:
+            content = yield response.content()
+            # Just a hack since str.format can't work with '{' in the string
+            content = str(content).replace("{", "{{")
```
p12tic: str.format can't work with `{` in the format string, not the replacements. We don't need this here.

master/buildbot/test/unit/test_reporter_zulip.py
```diff
@@ -0,0 +1,183 @@
...
+            '/api/v1/external/buildbot?api_key=123&stream=xyz',
+            json={
+                "event": "new",
+                "buildid": 20,
+                "buildername": "Builder0",
+                "url": "http://localhost:8080/#builders/79/builds/0",
+                "project": "testProject",
+                "timestamp": 1554161923
+            }, code=500)
+        self.setUpLogging()
```
p12tic: `setUpLogging` should go to the `setUp()` method.

p12tic:
>  I'm not sure about the parameterized library since the test is not to be tested with different values.

You could pass, e.g. started_at, completed_at, expected_url, expected_post_event, and so on. Anyway, that was just a suggestion.

So right now the PR is almost perfect, there are I think 3 trivial issues that need to be addressed. Good work!

master/buildbot/reporters/zulip.py
mayankmadan: But the error or log function in the Logger class takes the string passed as the parameter and calls format function on it again
https://github.com/buildbot/buildbot/blob/b54c4286281667d403b4e7cc7ac1627517178898/master/buildbot/util/logger.py#L24

master/buildbot/test/unit/test_reporter_zulip.py
mayankmadan: Do we really need to setup logging for each test? Only a couple of the test use it so won't it be better to just set it up inside those tests?

master/buildbot/reporters/zulip.py
p12tic: Oh, I didn't know that... Thanks for pointing that out!

I think a reasonable solution would be to use the format in log.error().

E.g. `log.error("{code}: Error pushing build status to Zulip: {content}", code=response.code, content=content)`

Something like that should work, shouldn't it?

master/buildbot/test/unit/test_reporter_zulip.py
p12tic: Agreed, now I see that a lot of tests already do that.

master/buildbot/reporters/zulip.py
mayankmadan: I dont really think it would. Because in this case, content is a json object. So the string passed to the error function would become
`"401: Error pushing: {'result': '...', 'msg': '...'} "`
Which again contains the curly braces which would make the string unformatable.

mayankmadan: A function just to log json could also be useful

p12tic: My idea was to remove log error in our code and reuse the one in `log.error`. That is, we pass the format string and parameters to log.error and it then formats it. The format string does not yet contain `{` and `}` characters from json, so that should be fine.

mayankmadan: I'm sorry about that. I don't know how I missed it :)

mayankmadan:
@p12tic I'm sorry but I might need some help with this. The build errors are from test_pipe_string and test_pb_keepalive. I'm not sure I fully understand whats going on [here](https://buildbot.buildbot.net/#/builders/23/builds/616/steps/7/logs/stdio)

p12tic:
> If its good to go, I'll squash all the commits down to 1

You can do that if you want, the feature is indeed self-contained. Could you fix the typo in the docs, that the last thing I'd like to see.

mayankmadan:
I might need a little help here. I'm not quite sure why the test is failing here

p12tic:
Looks like the test is unstable.

mayankmadan:
> Looks like the test is unstable.

I'm not sure how its unstable but the test for some reason is passing now

PR #4691, opened 2019-03-31
Add support for grouping workers to machines

p12tic:
This PR implements a way to associate workers with entities called "machines". A machine groups one or more workers sharing physical resources. Exposing this allows the following possibilities:

 - latent machines (implemented in this PR). Currently there's no way to shutdown resources that are shared by multiple latent workers. For example, ideally a physical machine or an EC2 instance would be shut down when there are no builds and woken up on new builds.

 - per-machine locks. Often more than one worker runs on the same host and this makes `WorkerLock` useless to prevent excessive usage of resources. In the future we could have `MachineLock` which would prevent e.g. excessive usage of memory.

This PR supersedes #4616 which implemented latent machines in a much more limited way.
 
The PR currently adds neither the newsfragment nor updates the documentation. I think that there's possibility the API may need slight changes once various types of machines are actually implemented. Once we have support for e.g. EC2 machine then it's a good time to expose the functionality to the end users.

Questions:
Is "machine" a good naming? It does not necessarily need to correspond to a physical machine, so the current naming is potentially a bit confusing.

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1099,6 +1099,7 @@ reconfiguring
 reconnection
 recurse
 redhat
+reentrant
 refactor
 refcount
 refetch
--- a/master/buildbot/config.py
+++ b/master/buildbot/config.py
@@ -228,6 +228,7 @@ class MasterConfig(util.ComparableMixin):
         self.builders = []
         self.workers = []
         self.change_sources = []
+        self.machines = []
         self.status = []
         self.user_managers = []
         self.revlink = default_revlink_matcher
@@ -263,6 +264,7 @@ class MasterConfig(util.ComparableMixin):
         "logMaxSize",
         "logMaxTailSize",
         "manhole",
+        "machines",
         "collapseRequests",
         "metrics",
         "mq",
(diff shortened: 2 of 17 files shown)


seankelly:
Machine seems fine. It's virtual machine after all.

p12tic:
@tardyp: A major problem with the approach of having everything as a latent worker is that AbstractLatentWorker already includes a lot of functionality that's rather complex. I think we should think hard if we want to include even more code there.

The major difference between a Worker and a Machine is that a Machine does not connect to buildbot and does not execute builds. A Machine is something that's just can be brought online and taken down, it does not connect to the master like workers. I think that it's enough of a difference to warrant a separate class hierarchy.

A separate class hierarchy won't add a lot of code in cases of e.g. EC2LatentMachine and EC2LatentWorker. Any reusable code can be split off to a separate mixin or class.

As for the idea of arbitrary hierarchy, I agree. We could indeed have arbitrary hierarchy eventually, the current design does not preclude that. I just didn't want to make everything complex from the start.

p12tic:
@tardyp: Rebased the PR and fixed an issue you've identified when we talked several weeks ago.

fischerling:
@p12tic Thank you for working on this feature :)
For our small lab I want to build a very similar setup to the one you described in #4434.
Do you have a configuration using `LatentMachine`, wol and suspend you could share?

PR #4694, opened 2019-04-01
Feature: add more git buildstep (git add and git commit)

chapipo:
This PR add two buildsteps:
- GitAdd to add some files in the staging area
- GitCommit to commit the files in the staging area

I have removed old styles docs from the source (like describe in the docs) and add some trivial test cases in other git related steps.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -770,3 +770,88 @@ class GitTag(buildstep.BuildStep, GitStepMixin, CompositeStepMixin):
 
         ret = yield self._dovccmd(cmd)
         return ret
+
+
+class GitCommit(buildstep.BuildStep, GitStepMixin, CompositeStepMixin):
+
+    description = None
+    descriptionDone = None
+    descriptionSuffix = None
+
+    name = 'gitcommit'
+    renderables = ['paths', 'messages']
+
+    def __init__(self, workdir=None, paths=None, messages=None, env=None,
+                 timeout=20 * 60, logEnviron=True,
+                 config=None, **kwargs):
+
+        self.workdir = workdir
+        self.messages = messages
+        self.paths = paths
+        self.env = env
+        self.timeout = timeout
+        self.logEnviron = logEnviron
+        self.config = config
+        # The repourl, sshPrivateKey and sshHostKey attributes are required by
+        # GitStepMixin, but aren't needed by git add and commit operations
+        self.repourl = " "
+        self.sshPrivateKey = None
+        self.sshHostKey = None
+
+        super().__init__(**kwargs)
+
+        self.setupGitStep()
+
+        if not self.messages:
+            bbconfig.error('GitCommit: must provide messages')
+
+        if not isinstance(self.messages, list):
+            bbconfig.error('GitCommit: messages must be a list')
+
+        if not self.paths:
+            bbconfig.error('GitCommit: must provide paths')
+
+        if not isinstance(self.paths, list):
+            bbconfig.error('GitCommit: paths must be a list')
+
+    @defer.inlineCallbacks
+    def run(self):
+        self.stdio_log = yield self.addLog("stdio")
+        gitInstalled = yield self.checkFeatureSupport()
+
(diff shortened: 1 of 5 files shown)


p12tic:
Thanks for the PR! 

What do you think if there was a single step to add files and commit, i.e. `GitCommit` would accept a list of files and would do both `git add` and `git commit`? Did you have a use case where `GitAdd` does not come with a paired `GitCommit`?

chapipo:
When I start the work on this, I ask myself the same question !

I don't have use case where `GitAdd` step do not come with `GitCommit` step. It's more to keep it close to git command line and have one git command by step.

But I can rework the PR to do only one step if you tell me so ;)

p12tic:
Buildbot does not necessarily need to map command line 1:1. It would be easier for the end users to configure only single step, not two and also reduce the chance of accidental errors. In addition to that, the staging area in git is pretty unique capability in version control systems so not exposing it would make the interface of the step more similar to e.g. HgCommit or SvnCommit once we have them.

I'll review the current PR to look for any problems that are likely to persist after the refactor, so don't hurry in rewriting it :-) Thanks!

tardyp:
I agree with @p12tic and I would even go further.

We should have have a GitPush command, which can do the git branch, git add, git commit, git tag, and git push in one step.
With a nice API so that it is easy to do all those things.

p12tic:
> We should have have a GitPush command, which can do the git branch, git add, git commit, git tag, and git push in one step.
I think that users may want to do multiple commits before pushing, that would be hard in such an API. Push is an atomic event, so multiple `GitPush` may not work in some cases.

tardyp:
> I think that users may want to do multiple commits before pushing,

okay

chapipo:
I have updated the code and the tests.

@p12tic it will be easier to review. A lot of code as duplicated with 2 steps instead of 1.

master/buildbot/steps/source/git.py
```diff
@@ -770,3 +661,87 @@ def _doTag(self):
...
+                 config=None, **kwargs):
+
+        self.workdir = workdir
+        self.messages = messages
+        self.files = files
+        self.env = env
+        self.timeout = timeout
+        self.logEnviron = logEnviron
+        self.config = config
+        # These attributes are not required for a git add command
```
p12tic: I think this comment could be reworded to something like: "The repourl, sshPrivateKey and sshHostKey attributes are required by GitStepMixin, but aren't needed by git add and commit operations"

master/buildbot/steps/source/git.py
```diff
@@ -770,3 +661,87 @@ def _doTag(self):
...
+class GitCommit(buildstep.BuildStep, GitStepMixin, CompositeStepMixin):
+
+    description = None
+    descriptionDone = None
+    descriptionSuffix = None
+
+    name = 'gitcommit'
+    renderables = ['files', 'messages']
+
+    def __init__(self, workdir=None, files=None, messages=None, env=None,
```
p12tic: Are directories supported? I think better naming would be `paths`.

master/setup.py
```diff
@@ -296,7 +296,7 @@ def define_plugin_entries(groups):
             ('buildbot.steps.source.cvs', ['CVS']),
             ('buildbot.steps.source.darcs', ['Darcs']),
             ('buildbot.steps.source.gerrit', ['Gerrit']),
-            ('buildbot.steps.source.git', ['Git', 'GitTag']),
+            ('buildbot.steps.source.git', ['Git', 'GitAdd', 'GitTag']),
```
p12tic: GitCommit

p12tic:
Looks good, thank you! I've added several comments on minor issues.

master/buildbot/steps/source/git.py
chapipo: Yes, paths is a better naming.

chapipo:
I just update the PR to take care of your comments.

I have updated docs too to reflect the merge of the 2 steps.

p12tic:
Looks great!

PR #4705, opened 2019-04-06
Use fake reactor in tests (part 6)

p12tic:
This PR uses fake reactor in even more tests. The DB and data API implementations now use reactor from master which can be modified in tests. This way there's no need to pass an additional `_reactor` argument.

There's interesting fact that the tests became around 2.5% faster after moving to fake reactor.

### DIFF
--- a/master/buildbot/data/buildrequests.py
+++ b/master/buildbot/data/buildrequests.py
@@ -14,7 +14,6 @@
 # Copyright Buildbot Team Members
 
 from twisted.internet import defer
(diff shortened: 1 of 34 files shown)


tardyp:
Shall we add a self.reactor property in BuildbotService, like we have for master, which would point to self.master.reactor?

p12tic:
> Shall we add a self.reactor property in BuildbotService, like we have for master, which would point to self.master.reactor?

I think that makes sense.

PR #4708, opened 2019-04-07
Miscelaneous latent worker fixes

p12tic:
This PR fixes several bugs in latent worker:
 - Latent workers with negative build_wait_timeout will be shutdown on master shutdown. Previously potentially costly resource would have been left running. Some people possibly don't shutdown latent workers due to long startup time, but it's not acceptable to leave them running when master itself is shutdown.

 - Master will now wait until `start_instance` or `stop_instance` finish during master shutdown and also before other `start_instance` or `stop_instance` call on the same worker is started. This makes proper lifetime management in latent workers backends easier, because calls to `start_instance` or `stop_instance` will never overlap.

 - Latent worker will now correctly handle synchronous exception from `stop_instance()`.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1465,6 +1465,7 @@ unclaim
 unclaimedbrdicts
 unclosed
 unconfigure
+unconfigured
 underpowered
 unencrypted
 unexpectedsuccesses
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1140,6 +1140,7 @@ requestcancelled
 requestjson
 requestsubmitted
 requeue
+requeued
 requiredargs
 reschedulenextbuild
 resetmocks
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1483,6 +1483,7 @@ unlink
 unlinked
 unparsable
 unparseable
+unpatch
 unregister
 unregisters
 unserialized
--- a/master/buildbot/newsfragments/latent-negative-build-timeout-stop-during-master-shutdown.bugfix
+++ b/master/buildbot/newsfragments/latent-negative-build-timeout-stop-during-master-shutdown.bugfix
@@ -1,0 +1,1 @@
+Latent workers with negative build_wait_timeout will be shutdown on master shutdown.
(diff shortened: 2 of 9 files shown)


seankelly:
> but it's not acceptable to leave them running when master itself is shutdown.
What if this behavior is desired however?

p12tic:
>  What if this behavior is desired however?

Good question. I think we need a separate config parameter for that. The current behavior was not exactly documented and it lets users to easily accidentally leave e.g. EC2 instance running.

seankelly:
Agreed. I think deleting the instances is the right choice as a default. However, this changes the default and doesn't let someone opt into the previous behavior.

tardyp:
The problem is that we can be stopping while in the middle of creating a
worker. So we really need some kind of lock here.

p12tic:
@seankelly: During implementation of the additional option for "really never shutdown the worker", it got me thinking: is there indeed a real use case for never shutting down a latent worker? I did some digging and it looks like this behavior got there by chance.

There was never a guarantee that such latent worker would be never stopped - we would stop it if it was started during master shutdown, and we would restart it if it ever lost connection.

I did some archeology and found that the initial implementation of this functionality has been added in 7d3183faefd034ddfe1c1a2609e34f9ac999fea3 on 2010 as a separate option `insubstantiate_after_build`. The initial implementation did shutdown latent workers when the master shuts down. Later, there was a refactor 895ef265d07396fecba99cee3006a35bb1049eaa on 2012 that merged `insubstantiate_after_build` with `build_wait_timeout`. That introduced a change of behavior, in that master would not stop latent workers with this option enabled during its shutdown.

I have a suspicion that there are no users that rely on this behavior by more than an accident. I would argue that we don't need this option at all as this is good way to leak resources with no way to get them back easily. If users really want to start and forget a latent worker, there are other ways to do that, such as overriding `stop_instance` and doing nothing there.

p12tic:
@tardyp: Could this PR still go in with the unnecessary complexity that could be fixed by a simple lock? I did the cleanup and it looks like it would be best to put it into subsequent PR for the sake of easier review.

seankelly:
> is there indeed a real use case for never shutting down a latent worker?
I was thinking about instances where it's hard to get the latent worker again. For example, it takes a long time to boot or not guaranteed to be able to get that specific VM.

However, the worker couldn't connect to the master when the master comes back because the master isn't expecting it. So I agree, we can skip the behavior with this refactor. If anyone does need that behavior, open an issue and we can add it back.

p12tic:
> However, the worker couldn't connect to the master when the master comes back because the master isn't expecting it.

Even worse, master does not know such already running worker exists and will try to start another one.

Issue #4711, opened 2019-04-08
Cannot "upgrade-master" state.sqlite using version 2.0


FHTheron:
I have a state.sqlite DB (55MiB) with a few years worth of build details. I can do `buildbot upgrade-master` from buildbot version 1.6 to 1.8. I can also run `buildbot cleanupdb` successfully. The master has a Janitor to clear old logs weekly.

If I try to upgrade past version 2.0, I get the following error:
```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: logs.stepid
[SQL: INSERT INTO logs SELECT * from migration_tmp]
```

I think this is related to changes made for #4539 and is caused by two `cancelled` entries:
```
$ sqlite3 backup.sqlite 'SELECT * FROM logs WHERE stepid IS NULL' 
273|cancelled|cancelled||1|1|t
48140|cancelled|cancelled||1|1|t
```

There are 565 `cancelled` entries in the table, but only two have lost their stepid values.

Is this considered a corrupt DB? I do not recall ever interrupting an upgrade. Could this be handled by upgrade-master in a graceful manner or should I locally modify the DB to either remove those entries or give them dummy stepid values?

FHTheron:
Prior to upgrade, I deleted the bad rows:
`sqlite3 state.sqlite 'DELETE FROM logs WHERE stepid IS NULL'`
Then used the latest buildbot version and upgrade-master and it succeeded.

Would it be possible to make the error either print some details about the rows, such as the contents of the row? Or even better, perhaps allow `--force` to also mean ignore any errors that could be ignored without complete DB breakage.

Issue #4539, opened 2019-01-18
Cannot delete old changes without wiping all of them


rjarry:
I tried to setup the cleanup using the cascade deletes and I noticed that when removing old sourcestamps, all changes are systematically deleted.

Here is how I experienced the problem:

```sql
SELECT count(*) FROM sourcestamps;
-- 737
SELECT count(*) FROM changes;
-- 737

-- Delete soucestamps older than 30 days.
-- The starting point is not "now" but the most recent sourcestamp.
DELETE FROM sourcestamps
WHERE created_at < (SELECT max(created_at) - (30 * 24 * 60 * 60) FROM sourcestamps);
-- DELETE 577

SELECT count(*) FROM sourcestamps;
-- 160
SELECT count(*) FROM changes;
-- 2
```

rjarry:
@tardyp: damn, I forgot to reference this issue in the news fragment. Do you want me to do another pull request for this? Or should I push it directly to the master branch?

tardyp:
please just create a news fragment in new PR. so that we are sure of the spelling

PR #4713, opened 2019-04-08
Fix db migration errors on sqlite (and potentially others)

rjarry:
When a build is cancelled while a step is starting up. The stepid has not been initialized yet. Before #4539, there was no integrity constraints on the stepid column which may have caused "invalid" rows to be created.

This may cause migration number 52 to fail since the constraint is violated.

In case such offending rows are found during the migration, issue a warning and delete them since they are indeed invalid.

Fixes: #4711 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -789,6 +789,8 @@ class BuildStep(results.ResultComputingConfigMixin,
         return self.build.getWorkerName()
 
     def addLog(self, name, type='s', logEncoding=None):
+        if self.stepid is None:
+            raise BuildStepCancelled
         d = self.master.data.updates.addLog(self.stepid,
                                             util.bytes2unicode(name),
                                             str(type))
@@ -819,6 +821,8 @@ class BuildStep(results.ResultComputingConfigMixin,
     @_maybeUnhandled
     @defer.inlineCallbacks
     def addCompleteLog(self, name, text):
+        if self.stepid is None:
+            raise BuildStepCancelled
         logid = yield self.master.data.updates.addLog(self.stepid,
                                                       util.bytes2unicode(name), 't')
         _log = self._newLog(name, 't', logid)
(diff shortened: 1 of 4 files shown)


master/buildbot/process/buildstep.py
```diff
@@ -789,6 +789,8 @@ def getWorkerName(self):
         return self.build.getWorkerName()
 
     def addLog(self, name, type='s', logEncoding=None):
+        if self.stepid is None:
```
tardyp: looks like unrelated

rjarry: This is the root cause of the problem. 

Since stepid is None until it is actually added to the db, if a build is cancelled before the build step is properly initialized, there may be a call to addLog while stepid is still None.

Now that there are NOT NULL constraints on the logs.stepid column, we should not experience db integrity problems. However, this would cause errors to be risen.

rjarry: Please note that I took care of putting this in a separate commit.

p12tic: A better PR description would have prevented this misunderstanding :-)

rjarry: You're being polite :-)

```bash
| sed 's/better //'
```

PR description updated.

rjarry:
> Code looks good. Could we cover the behavior of not adding logs to DB if stepid is None with unit tests?

Yes that would be better. I did not have much time to work on this (hence the absence of PR description :-))

I'll try to do this soon.

p12tic:
Thanks!

master/buildbot/process/buildstep.py
p12tic: @tardyp: Could you re-review the PR? Then I'll merge it as soon as we have tests.

tardyp:
As this fix some migration issue. I am merging it.

PR #4719, opened 2019-04-14
Fix bug with git submodules

PodnimatelPingvinov:
There are some caveats with git submodules, which are described in git-scm (https://git-scm.com/book/en/v2/Git-Tools-Submodules) book (see section "Issues with Submodules"). One of them is checkouting from branch with submodule to a branch without that submodule. After such action we still have the submodule directory as an untracked directory.

Source git step doesn't consider that and in "fresh" and "clean" methods we can get untracked directories. The solution, which I introduce in my commit, is pretty straightforward: execute "git clean -ddf" not only before checkout, but also after.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/git-submodules.bugfix
+++ b/master/buildbot/newsfragments/git-submodules.bugfix
@@ -1,2 +1,1 @@
-Fixed bug with git checkouting from branch with submodule to branch without
-that submodule in :py:class:`~buildbot.steps.source.git.Git`
+Fixed :py:class:~buildbot.steps.source.git.Git to clean the repository after the checkout when submodules are enabled. Previously this action could lead to untracked module directories after changing branches.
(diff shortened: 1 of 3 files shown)


master/buildbot/newsfragments/git-submodules.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed bug with git checkouting from branch with submodule to branch without
```
p12tic: Users may be confused about what specific bug the release note refers to. I would suggest something like "Fixed :py:class:`~buildbot.steps.source.git.Git` to clean the repository after the checkout when submodules are enabled. Previously this action could lead to untracked module directories after changing branches".

master/buildbot/steps/source/git.py
```diff
@@ -204,6 +204,11 @@ def clean(self):
         rc = yield self._cleanSubmodule()
         if rc != RC_SUCCESS:
             raise buildstep.BuildStepFailed
+
+        if self.submodules:
+            rc = yield self._dovccmd(clean_command)
```
p12tic: Test coverage says that we don't have a test for this code path. Could you maybe add it? `test_mode_full_clean` is probably a good candidate to base the effort on.

PodnimatelPingvinov:
Thanks for review! I hope I understand you right with those fixes.

Issue #3809, opened 2017-12-06
Debian packaging discussions


tardyp:
followup on the packaging discussions happening on #3804

tardyp:
@simevo

> The plan is to redo your build process on debian infrastructure and with debianized tooling, starting from the github tarball as source (not the "output" tarball that you produce which for us is a "binary" artifact)

I am not sure which plan you are referring. Are you planning to be the debian maintainer for us?

We currently have other discussions with debian folks about doing this:
```
From: Matthias Klose <doko@debian.org>
To: Robin Jarry <robin.jarry@6wind.com>, jolly_roger@itblog.org.ua, "W. Martin Borgert" <debacle@debian.org>
Subject: Re: Buildbot debian packaging

On 14.11.2017 15:51, Robin Jarry wrote:
> Hi fellows,
>
> I noticed that the debian packaged version of buildbot and
> buildbot-slave is still on the 0.8.x branch.
>
> A new 0.9.x version is available since october 2016 and I really think
> it would be a good thing to upgrade the debian package to a more recent
> version. Here are some links to the new web dashboard:
>
> https://nine.buildbot.net/
> https://nine.buildbot.net/#/console
> https://nine.buildbot.net/#/grid
> https://nine.buildbot.net/#/waterfall
>
> Please note that this new version changes a lot of things and the
> upgrade procedure can be tricky.
>
> https://docs.buildbot.net/current/manual/installation/nine-upgrade.html
>
> Let me know if there is something I can do to help.

Hi,

I just filed #883529 to orphan the packages. I didn't get any feedback from
Andrij since July 2016, and feedback from before that date was scarce as well.
Please feel free to adopt the packages, and coordinate maybe with Martin about
sponsoring/co-maintaining the packages.  I'm available for initial uploads as
well, but won't have the time to maintain these packages. And buildbot-0.9.x
should use Python3 from the start ...

Matthias

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=883529
```
We discussed it during our weekly meeting yesterday, and @rjarry was candidating to be the debian downstream maintainer for us.

There is few chance that we can remove guanlecoja/buildbot-pkg build system in a reasonable timeframe. 
I would like we can work on a solution to integrate into debian without reworking that.
We are already packaged into gentoo, nix and freebsd (AFAIK, they are using the pypi tarballs)

I am not sure what is the current debian tooling on building nodejs applications.
[...]

PR #4720, opened 2019-04-14
www: Convert all CoffeeScript code to JavaScript

p12tic:
This PR converts all CoffeeScript code within Buildbot to plain JavaScript and fixes errors identified by unit tests.

I've tested the updated frontend locally with locally updated versions of guanlecoja and guanlecoja-ui and did not see any issues in my limited testing.

The PR is roughly based on the ideas in #3806, #4056 and #4046, but goes in a different direction. The PR only includes conversion of CoffeeScript to JavaScript code and minimal cleanup needed to pass tests. Only minimal changes have been done to dependencies. The PR does not switch build system to webpack yet as I did not find an obvious way to support Jade, but we can do that in the future.

Currently existing Buildbot unit tests will fail due to missing upgrades to dependencies.

Depends on https://github.com/buildbot/guanlecoja/pull/18 and on https://github.com/buildbot/guanlecoja-ui/pull/14.


tardyp:
The problem with this approach is that you end up with an ES6 app. 
The code looks fine an maintainable enough, at first sight.

https://kangax.github.io/compat-table/es6/ says that all is es6 compatible for a while, but I am not sure what to expect..

tardyp:
looks like error comes from ngAnnotate, which is only used in production build.

https://stackoverflow.com/questions/45980598/ngannotate-warning-stringmap-expected-string-key
can you take a look?

p12tic:
I should've added that I still needed modifications to guanlecoja to overcome several issues. I think the ngannotate one is caused by ES6 code being not supported, at least that was what the internet said. Converting to ES5 with babel should  be the way to go.

tardyp:
after looking at http://kangax.github.io/compat-table/es6/ , I was wondering if it really make sense to transpile to es5. Only IE 11 does not support es6, and that is 1.23% of our buildbot.net audience, according to google analytics (not counting ppl who use adblock, which block GA)

tardyp:
anyway, uglifyjs cannot minify es6..

tardyp:
I removed annotate and uglify from guanlecoja, then ran the output with uglify-es https://www.npmjs.com/package/uglify-es

The result scripts.js is 1074149 bytes (coffeescript compiled to 668175 bytes)
:(

Then if you gzip it, it goes to 265567 vs 206214

p12tic:
I tried webpack at first and was hit with some differences of behavior plus unclear path for supporting Jade. I think it's good idea to limit this PR to coffeescript only and migrate to webpack as a separate PR, so that there's less risk.

With regards to ES6/5 I'd like to do the same - patch whatever is needed now to support ES5 and then migrate to newer standard later. According to our version policy this probably needs a major version bump as we're removing features.

p12tic:
The current version of the PR works without extra hacks in the dependencies. Unit tests pass and I could do `make frontend` and play with a test buildbot instance.

The PR depends on the following PRs being merged and npm packages released: 

- https://github.com/buildbot/guanlecoja/pull/19 - enables running js sources through babel.

- https://github.com/buildbot/guanlecoja-ui/pull/14 - this one needs to be upgraded and published to bower so that newer gulp-bower-deps is used which includes https://github.com/tardyp/gulp-bower-deps/pull/3 which fixes support for strict mode. The current vendor.js acquired through bower is broken for the current environment.

p12tic:
The PR now depends on https://github.com/buildbot/guanlecoja/pull/20 and https://github.com/buildbot/guanlecoja-ui/pull/14 landing. I was able to test the following locally:
 - unit tests via `gulp prod`. 
 - smoke tests via `common/smoketest.sh whl`
 - manual tests by playing a test master instance and visiting most, if not all, pages on it

The following local modifications were needed for the above to pass:
 - reverting the bower dependency updates for `buildbot-data` and `guanlecoja-ui` so that it does not complain that the required versions are not available.
 - symlinking of the updated `buildbot-data` bower distributable buildbot-data.js across the modules
 - symlinking of the updated `guanlecoja-ui` bower distributable vendor.js and scripts.js from https://github.com/buildbot/guanlecoja-ui/pull/14
 - installation of updated `guanlecoja` from https://github.com/buildbot/guanlecoja/pull/20. package.json needed to be modified to fetch package from local path.
 - installation of updated `guanlecoja-ui` from https://github.com/buildbot/guanlecoja-ui/pull/14. package.json needed to be modified to fetch package from local path.

uglycoyote:
This is excellent, thanks for continuing down this path!

With regards to es5 vs es6:  my gut feeling is that there's no need to compile it down to es5, because of the intended audience being developers (who would tend to be running an ES6-supporting browser or be willing to update to one if their team was using buildbot and it was a requirement of their work), in contrast with shipping a website for marketing purposes to a general audience.

With regards to webpack and jade:  It seems like they either renamed jade to "pug", or pug is a successor to jade, I'm not sure which, but there is a [pug-loader](https://github.com/pugjs/pug-loader).  

Also, as much as I hate to be a flavour-of-the-month recommender, I have had a lot of success with [Parcel](https://parceljs.org/) which is a less-popular, newer, much-faster, zero-configuration alternative to Webpack.   Webpack does have a much larger mindshare at the moment it seems, but Parcel might be worth looking at.   It has out-of-the-box support for Pug, Typescript, LESS/SASS, and the build-in browser hotloading works great and makes iterative turnaround much faster than any setup that i have been able to get with Webpack.

BTW, off-topic a bit, but I occasionally try to follow efforts on the buildbot front end stuff by searching for terms like "javascript", or "coffee" in the IRC logs at https://irclogs.jackgrigg.com/, but it seems like that server has stopped archiving stuff recently, so I was wondering if there is anywhere else where the IRC chats are archived?  see [this search](https://irclogs.jackgrigg.com/search.pl?server=irc.freenode.net&channel=buildbot&nick=&q=coffee) for an example, it says the last mention of coffee was May 2018

tardyp:
@uglycoyote yes, we are targetting developers, but we are also targetting big companies, where sometimes IE is still default browser.
image


>  pug

We already use pug inside guanlecoja, I just didn't bother changing the name of the files. We also use pyjade as a way to customize the templates. There are subtle differences between python implementation and node implementation. I think the best would be to use the python implementation. I have made a test script which converts all all jade files to precompiled templates, and it takes 200ms to run. So maybe we want to drop the pug dependency as well.


> parcel vs webpack

One of the goal of all this is also to be able to get buildbot_www into debian. For that we need to reduce as much as possible the dependencies on node packages which are not yet into debian. This is probably why a simple webpack build system is needed.

We can have a nice dev environment, with lots of dependencies (for hot reload, and all those nice feats). but we must have a production build with very minimal deps.

p12tic:
> parcel vs webpack

I'd like to add that given that within buildbot there's very little time to onboard new developers. All uncommon things are hoops new developers need to figure out themselves. Therefore it's important to use as standard setup as possible, so that there are plenty of information for the developer to learn from.

tardyp:
rebased + update yarn modules + configure data_module as a library.

I reproduce the build pass, and the size of the generated file is actually smaller..

rodrigc:
@p12tic nice!

uglycoyote:
> I'd like to add that given that within buildbot there's very little time to onboard new developers.  All uncommon things are hoops new developers need to figure out themselves. 

I would tend to agree with this, except that my experience contradicts it.  I'm not an expert with any particular JS build system, but have spent tons of time cursing at webpack config files, trying to the plethora of plugins to work together, versus my experience with Parcel being that it "just works" and you don't need a config file or any plugins.  So I would say that Parcel has a vastly better onboarding experience, especially if you are talking about onboarding developers who are not JS experts, and from what I have gathered most people who encounter the Buildbot front end are coming at it from a python background with minimal JS experience and therefore would presumably not benefit from vast amounts of prior knowledge about webpack.

> One of the goal of all this is also to be able to get buildbot_www into debian

I have heard mention of Debian packaging a bunch of times related to the front end stuff, but I'm not quite sure what this means, is debian for some reason difficult to install buildbot_www onto?  We use Ubuntu and haven't had any issues building the frontend there.    I have read through #3809, #3804 mentions Debian, but I still don't fully understand.   Sorry if it is a dumb question.

tardyp:
We are talking about packaging for debian.

Debian policy requires every dependencies to be packaged as well.

guanlecoja has 800 dependencies. That would mean probably create 700 new
debian packages..
webpack, babel (and their deps) are all already packaged as they are
ubiquitous.

Pierre


Le jeu. 25 avr. 2019 à 21:51, uglycoyote <notifications@github.com> a
écrit :

PR #4726, opened 2019-04-18
Fix statistics reconfigure

jangmarker:
Before, the consumers were not correctly unregistered, so each
reconfig would create new consumers *additional* to the old ones.

## Contributor Checklist:

* ~[ ] I have updated the appropriate documentation~

### DIFF
--- a/master/buildbot/statistics/stats_service.py
+++ b/master/buildbot/statistics/stats_service.py
@@ -26,6 +26,10 @@ class StatsService(service.BuildbotService):
     A middleware for passing on statistics data to all storage backends.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.consumers = []
+
     def checkConfig(self, storage_backends):
         for wfb in storage_backends:
             if not isinstance(wfb, StatsStorageBase):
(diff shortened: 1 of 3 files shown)


master/buildbot/statistics/stats_service.py
```diff
@@ -43,7 +43,6 @@ def reconfigService(self, storage_backends):
         for svc in storage_backends:
             self.registeredStorageServices.append(svc)
 
-        self.consumers = []
```
p12tic: I agree with @krf, we should just move `removeConsumers()` here.

master/buildbot/test/unit/test_stats_service.py
```diff
@@ -85,6 +85,15 @@ def test_reconfig_with_fake_storage_backend(self):
         self.stats_service.reconfigService(new_storage_backends)
         self.checkEqual(new_storage_backends)
 
+    def test_reconfig_with_consumers(self):
+        backend = fakestats.FakeStatsStorageService(name='One')
+        backend.captures = [capture.CaptureProperty('test_builder', 'test')]
+        new_storage_backends = [ backend ]
```
p12tic: pylint doesn't like the spaces above between `[` and `]`.

jangmarker:
Is there a place to see the status of all tests? I have the feeling that the tests for the statistics service have not been running before, because they failed when I ran them locally and one of these commits fixes many of the statistics service tests by adding a class variable to a fixture.

jangmarker:
Any comments on my update? I cannot restart the builds, as far as I can tell they do not fail because of my changes.

p12tic:
@jangmarker : Could you rebase on newest master?

p12tic:
Sorry for not noticing your comments after the review.

PR #4729, opened 2019-04-23
www: Import guanlecoja repository

p12tic:
This PR imports the guanlecoja repository from https://github.com/buildbot/guanlecoja at the most recent commit https://github.com/buildbot/guanlecoja/commit/5235f44241ce2fc9d04cf0f0b58d4069d257fb27 to the www/guanlecoja directory. Whole history including merges is preserved for future reference.

This PR is not expected to cause any impact to the rest of the buildbot codebase. Nothing is using the local path as a dependency, so the PR should be relatively safe.


tardyp:
I am not sure what is the point of importing this in the mono repo.
For me the goal is to switch to webpack, and thus drop guanlecoja, so why importing it just before dropping?

p12tic:
Ok, makes sense.

PR #4739, opened 2019-04-27
latent: Implement "generic" latent machines

p12tic:
This PR implements a type of latent machine that can do any action for wake and shutdown. Several actions geared towards management of physical hosts via ssh are implemented. This functionality is unlikely to be used in any commercial build farm as there are more convenient tools to manage large number of machines. However it's still valuable as a toy implementation of the concept and also for small build farms where reduction of noise and/or power use is relatively more important.

The API is currently not intended to be used by the end users, therefore there are no API documentation nor release notes yet. My plan is to verify that the APIs are relatively future-proof by implementing and testing `EC2LatentMachine` in a subsequent PR. Documentation and release notes will be written after that.

### DIFF
--- a/master/buildbot/interfaces.py
+++ b/master/buildbot/interfaces.py
@@ -191,6 +191,14 @@ class IMachine(Interface):
     pass
 
 
+class IMachineAction(Interface):
+    def perform(self, manager):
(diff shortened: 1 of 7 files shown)


tardyp:
Interresting approach

PR #4748, opened 2019-05-03
worker: add test for invalid utf-8 output

vbraun:
Add a unit test for worker receiving invalid utf-8 output, and update tox

### DIFF
--- a/worker/tox.ini
+++ b/worker/tox.ini
@@ -4,7 +4,8 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py24, py25, py26, py27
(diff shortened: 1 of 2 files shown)


worker/tox.ini
```diff
@@ -4,7 +4,8 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py24, py25, py26, py27
+envlist = py24, py25, py26, py27, py35, py36, py37
```
seankelly: Pretty sure tox isn't used and that's why this hasn't been updated.

vbraun: Yes, but I also didn't figure out how else the worker is being tested for py2.x regressions

PR #4751, opened 2019-05-04
WIP: www: Use webpack as frontend bundler

p12tic:
This PR replaces gulp and bower with webpack as the frontend resource bundler and build manager. Tests and UI work locally, but so far I only did minimal manual testing. The PR is relatively risky, so it's a good time to review e2e tests for missing coverage. E.g. I think codeparameter plugin is not covered at all.

The PR removes the dependency of guanlecoja that had lots of goodies that could be reused. As a result each plugin implements webpack config in its own way and thus we have a lot of duplicate configuration right now. We'll need something like guanlecoja in the future to deduplicate that. This time it probably makes sense to keep the package within Buildbot repository so that we don't need to publish to npm.


www/base/src/app/d3/d3.service.js
```diff
@@ -10,10 +10,9 @@ class D3 {
     constructor($document, $q, config, $rootScope) {
         const d = $q.defer();
 
-        // Resolve function
-        $.getScript('d3.min.js', () =>
-            $rootScope.$apply(() => d.resolve(window.d3))
-        );
+        import('d3').then(module => {
+            d.resolve(module);
```
tardyp: the goal was to put d3 as a separate webpack bundle (as it is pretty big)
did you plan to do that?

www/base/src/styles/styles.less
```diff
@@ -1,8 +1,11 @@
 // Import bootstrap and font awesome stylesheets
-@import "../../libs/bootstrap/less/bootstrap.less";
-@import "../../libs/font-awesome/less/font-awesome.less";
-@import (inline) "../../libs/guanlecoja-ui/styles.css";
-@fa-font-path: "./fonts";
+@import "../app/**/*.less";
+@import "./animations.less";
+@import "./colors.less";
+@import "./mobile.less";
+@import "~bootstrap/less/bootstrap.less";
```
tardyp: I'd expect bootstrap to be first. in css, order is important, and can lead with subtle issues

www/base/src/app/app.module.js
```diff
@@ -13,3 +21,8 @@ class App {
 
 
 angular.module('app', new App());
+
+require('./common/common.module.js');
+
+const context = require.context('./', true, /^(?!.*(?:module|spec|webpack.js$)).*\.js$/);
```
tardyp: I think this is better to keep it explicit.

I know this trick is import(*.js) but it is really not obvious.

So I would just import stuff line by line

www/wsgi_dashboards/webpack.config.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
```
tardyp: can we factorize this into buildbot_pkg?

tardyp:
needs rebase

www/base/src/app/d3/d3.service.js
p12tic: Thids is "dynamic" import, it should import d3 from a separate bundle only when this code is run. This is why the import statement returns a promise.

www/base/src/styles/styles.less
p12tic: Agreed.

www/base/src/app/app.module.js
p12tic: Agreed.

www/wsgi_dashboards/webpack.config.js
p12tic: Agreed, WIP.

www/base/src/styles/styles.less
p12tic: Fixed now.

www/wsgi_dashboards/webpack.config.js
p12tic: Fixed now.

www/base/src/app/app.module.js
p12tic: Fixed now.

.circleci/config.yml
```diff
@@ -30,6 +30,7 @@ jobs:
             pip install -r requirements-ci.txt
             pip install -r requirements-cidocs.txt
             pip install pyinstaller
+            make frontend_deps
```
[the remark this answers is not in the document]
p12tic: We need to build guanlecoja-ui and buildbot_data in order to be able to do `pip install -e www/base`, because setup.py assumes these packages have already been built. I agree that the situation can be improved - we can extract only the `yarn build` part to a separate target.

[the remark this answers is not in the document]
p12tic: Should be fixed now.

master/docs/developer/quickstart.rst
```diff
@@ -54,66 +61,78 @@ Prerequisites
 
 .. note::
 
-  Buildbot UI is only tested to build on node 4.x.x.
+  Buildbot UI requires at least node 4 or newer and yarn.
 
 * Install LTS release of node.js.
 
-  http://nodejs.org/ is a good start for windows and osx
+  http://nodejs.org/ is a good start for Windows and OSX.
```
mojca: Nowadays they call it macOS :)

rodrigc:
Nice work!

rajdeepbharati:
LGTM 👍

uglycoyote:
It is great to see things moving forward with this.   One question.  I noticed that the docs for running ```gulp dev proxy``` were removed without mentioning a new proxy alternative in the replacement documentation.  I use ```gulp dev proxy``` all the time, for me it is very important to be able to test the front end on my own machine off the data from our production buildbot server (dummy test data on a local buildbot master usually will not suffice).   

Is there a solution under the new webpack build for doing this kind of proxying?

p12tic:
See #4848.

uglycoyote:
@p12tic  Thanks, `buildbot-dev-proxy` works great.  

Took me a little while to figure out why my plugin was not appearing on the left-menu.  In the old `gulp-dev-proxy` it seemed to just use any plugins that you installed with `pip install`.  

I found the command line option eventually (`--plugins`)

PR #4759, opened 2019-05-08
worker: test for invalid utf-8 in log file and fix the LogFileWatcher

vbraun:
The log file watcher does not utf-8 decode the file content, so
* RunProcessPP.outReceived safely decodes (ignoring errors) before calling RunProcess.addStdout
* LogFileWatcher.poll does not decode before calling RunProcess.addLogfile
So if the log file contains invalid utf-8, once the buffered logs are flushed, we trigger a UnicodeDecodeError

The added unit case fails with UnicodeDecodeError without the change to LogFileWatcher

### DIFF
--- a/worker/buildbot_worker/test/unit/test_runprocess.py
+++ b/worker/buildbot_worker/test/unit/test_runprocess.py
@@ -21,7 +21,6 @@ import os
 import re
 import signal
 import sys
-import tempfile
 import time
 
 from mock import Mock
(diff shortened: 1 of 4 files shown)


worker/buildbot_worker/test/unit/test_runprocess.py
```diff
@@ -857,3 +857,20 @@ def test_statFile_exists(self):
...
+        INVALID_UTF8 = b'\xff'
+        with open('invalid_utf8.log', 'wb') as f:
+            f.write(INVALID_UTF8)
+        # the watcher picks up the changed log
+        lf.poll()
+        # flush she buffer
+        rp._sendBuffers()
+        # cleanup
+        lf.stop()
+        os.remove('invalid_utf8.log')
```
p12tic: I think we need to put this into `finally:`. If test fails then we'd leave the file.

worker/buildbot_worker/test/unit/test_runprocess.py
```diff
@@ -857,3 +857,20 @@ def test_statFile_exists(self):
         self.assertEqual(
             st and st[2], 2, "statfile.log exists and size is correct")
         os.remove('statfile.log')
+
+    def test_invalid_utf8(self):
+        # create the log file watcher first
+        rp = self.makeRP()
+        lf = runprocess.LogFileWatcher(rp, 'test', 'invalid_utf8.log',
```
p12tic: Could we use `tempfile.NamedTemporaryFile`? If tests are ran in parallel and any other test uses this path we'll get rare failures that will be hard to track down.

worker/buildbot_worker/test/unit/test_runprocess.py
```diff
@@ -857,3 +857,20 @@ def test_statFile_exists(self):
...
+                                       follow=False, poll=False)
+        # now write to the log file
+        INVALID_UTF8 = b'\xff'
```
p12tic: Could we check that we don't drop entire log contents and just remove the bad UTF-8? I.e. let's have some valid text here too and then check what we send just after `rp._sendBuffers()`. See `testSendBuffered()` test in `test_runprocess.py` for an example. We'll need to get `FakeWorkerForBuilder` instance out of `makeRP`.

p12tic:
Also, could you add a release note in master/buildbot/newsfragments directory?

worker/buildbot_worker/test/unit/test_runprocess.py
vbraun: The current tests apparently use pytest-trialtemp to run in the _trial_temp subdirectory, which is also listed in .gitignore. The directry is cleared during each test run so leaving stale files in case of error is not a problem, and arguably a debugging feature. 

A NamedTemporaryFile(..., delete=True) can't be reopened on Windows

p12tic: Alright, now I see that this creates more problems than it solves.. I'll try to see if it's possible to address my concern in another way.

p12tic:
@vbraun: Does https://github.com/buildbot/buildbot/pull/4759/commits/28842c6484cad739d25e884d5f81a57f4dc794bb make sense?

vbraun:
Looks good to me!

PR #4764, opened 2019-05-09
add api token feature

tardyp:
The api token copies the user infos from the session into a new JWT, encoded with different key.
this api token cannot be used as a session cookie, and vice versa.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/service.py
+++ b/master/buildbot/www/service.py
@@ -36,6 +36,7 @@ from buildbot.plugins.db import get_plugins
 from buildbot.util import bytes2unicode
 from buildbot.util import service
 from buildbot.util import unicode2bytes
+from buildbot.www import api_token
 from buildbot.www import auth
 from buildbot.www import avatar
 from buildbot.www import change_hook
@@ -153,6 +154,8 @@ class BuildbotSite(server.Site):
 
     def setSessionSecret(self, secret):
         self.session_secret = secret
+        # change the key for api tokens
+        self.api_token_secret = secret[SESSION_SECRET_LENGTH//8:] + secret[:SESSION_SECRET_LENGTH//8]
 
     def makeSession(self):
         """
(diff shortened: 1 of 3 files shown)


master/buildbot/www/service.py
```diff
@@ -153,6 +154,8 @@ def _openLogFile(self, path):
 
     def setSessionSecret(self, secret):
         self.session_secret = secret
+        # change the key for api tokens
+        self.api_token_secret = secret[SESSION_SECRET_LENGTH//8:] + secret[:SESSION_SECRET_LENGTH//8]
```
p12tic: I think it would be better to just hash the secret.

p12tic:
I think we'll need a proper way to store the tokens to DB and for users to revoke them. Right now we're effectively using the user session cookie (albeit coded under a different key) as an API key, with no way to rotate it in case of potential data compromise.

master/buildbot/www/service.py
phi1010: You want to use another secret to sign session cookies and api tokens, please use an established salted Key Derivation Function (KDF, e.g. an scrypt pwhash as provided by Libsodium or a simple HMAC HKDF, if there is no other python reference implementation).

By providing a salt (ideally some random bytes, hardcoded as a constant) you can avoid accidentially deriving this same key for another less secure usecase somewhere else.

You also might want to offer adding a secret from the filesystem config file to avoid key leaks when database contents can be seen.

master/buildbot/www/service.py
```diff
@@ -386,6 +392,18 @@ def getUserInfos(self, request):
         session = request.getSession()
         return session.user_info
 
+    def getUserInfosFromSessionOrApiToken(self, request):
+        token = request.getHeader(b'X-API-TOKEN')
+        if token:
+            try:
+                session = jwt.decode(token, self.site.api_token_secret, algorithm=APITOKEN_SECRET_ALGORITHM)
+                return session.user_info
```
phi1010: Is the content of user_info subject to change on update or profile change? Tokens might live a long time; you could restrict the claim to maybe a version number, a unique and constant user identifier (username? database entry id?) , a token identifier (e.g. sequential or random serial) for revocation, and get every other user data from the database after successful authentication, otherwise you might decode Realnames, Mailadresses, etc. that are not up-to-date anymore.

tardyp:
I agree with you this approach is not good and will make more harm than help med-long term

Issue #4746, opened 2019-05-02
buildbot try (try_vc=git) fails with "Data must not be unicode"


yhager:
Running a command like `buildbot try -w me -b <builder>` results in a stack trace before even connecting to the master:
```
using 'ssh' connect method
job created
Unhandled Error
Traceback (most recent call last):
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/buildbot/clients/tryclient.py", line 907, in <lambda>
    d.addCallback(lambda res: deliver())
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/buildbot/clients/tryclient.py", line 669, in deliverJob
    reactor.spawnProcess(pp, argv[0], argv, os.environ)
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/posixbase.py", line 340, in spawnProcess
    processProtocol, uid, gid, childFDs)
--- <exception caught here> ---
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/process.py", line 770, in __init__
    self.proto.makeConnection(self)
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/protocol.py", line 514, in makeConnection
    self.connectionMade()
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/buildbot/clients/tryclient.py", line 512, in connectionMade
    self.transport.write(self.job)
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/process.py", line 927, in write
    self.pipes[0].write(data)
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/process.py", line 203, in write
    abstract.FileDescriptor.write(self, data)
  File "/home/yuval/.virtualenv/buildbot/lib/python3.7/site-packages/twisted/internet/abstract.py", line 349, in write
    raise TypeError("Data must not be unicode")
builtins.TypeError: Data must not be unicode
```

The patch is very simple:
```
diff --git a/myproject/version.py b/myproject/version.py
index 68d7165..61ad8a0 100644
--- a/myproject/version.py
[...]

PR #4765, opened 2019-05-11
Fixed try client by replaceing str functions with decode

umar456:
Fixes errors that were occuring with the try scheduler on Python 3
with git vc that was caused by invalid conversions from bytes to
strings.

I am not sure if this is the ideal solution. I tried to run the test scripts but many of them failed even on master. The patch seems to work on my machine&trade; but the tests seem to be failing. I am not very familiar with python so please let me know if something can be improved. This address the errors in #4746 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/clients/tryclient.py
+++ b/master/buildbot/clients/tryclient.py
@@ -460,9 +460,9 @@ def createJobfile(jobid, branch, baserev, patch_level, patch_body, repository,
     if version < 5:
         job += ns(jobid)
         job += ns(branch)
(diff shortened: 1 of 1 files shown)


p12tic:
Did you see the test failures even after following the development quick start tutorial [here](http://docs.buildbot.net/2.3.0/developer/quickstart.html#create-a-buildbot-python-environment)?

p12tic:
The issue has been addressed in PR #4886, so this PR is no longer needed.

PR #4771, opened 2019-05-17
Change git ssh data dir to be more unique

M4T1A5:
Currently if you have multiple builders running on the same worker
with the same workdir, the ssh keys are trying to be downloaded to the
same path, which obviously causes problems with parallel builds on the same worker.

This change tries to use the builddir parameter to make the path more unique
for each builder so they won't interfere with each other.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
+++ b/master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
@@ -1,1 +1,1 @@
-:py:class:`~buildbot.util.git.GitStepMixin`: Change ssh data path to be more unique for a given builder
+:py:class:`~buildbot.util.git.GitStepMixin`: Prevent builders from corrupting temporary ssh data path by using builder name as part of the path
--- a/master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
+++ b/master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
@@ -1,0 +1,1 @@
+:py:class:`~buildbot.util.git.GitStepMixin`: Change ssh data path to be more unique for a given builder
--- a/master/buildbot/util/git.py
+++ b/master/buildbot/util/git.py
@@ -151,24 +151,22 @@ class GitStepMixin(GitMixin):
         # temporary directory for that data to ensure the confidentiality of it.
         # So instead of
         # '{path}/{to}/{workdir}/.buildbot-ssh-key' we put the key at
-        # '{path}/{to}/.{workdir}.buildbot/ssh-key'.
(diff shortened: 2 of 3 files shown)


master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
```diff
@@ -0,0 +1 @@
+:py:class:`~buildbot.util.git.GitStepMixin`: Change ssh data path to be more unique for a given builder
```
p12tic: I'd word this like: "Prevent builders from corrupting temporary ssh data by using `builldir` as part of the temporary path for that data". Right now it's not apparent what exact issue we're fixing.

master/buildbot/util/git.py
```diff
@@ -157,14 +157,18 @@ def _getSshDataPath(self):
         # ending with a slash
         path_module = self.build.path_module
 
-        workdir = self._getSshDataWorkDir().rstrip('/\\')
-        if path_module.isabs(workdir):
-            parent_path = path_module.dirname(workdir)
+        if self.build.hasProperty("builddir"):
```
p12tic: Would using builder name as part of the path solve the issue? That would make the solution not depend on whether `builddir` is set or not.

master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
M4T1A5: Yes I agree, I just wasn't sure this information was available here, but I will change it as you suggested.

master/buildbot/util/git.py
M4T1A5: Yes I agree, I just wasn't sure this information was available here, but I will change it

master/buildbot/newsfragments/git-ssh-data-path-fix.bugfix
M4T1A5: Yes, I see, I will change the wording to be more specific/explicit

M4T1A5:
Yes I agree the builder name is a better fit. I am not a professional Python developer so I'm not very familiar navigating Python codebases of this size, but I have now changed the code to use the builder name instead.

p12tic:
Thanks, the code looks perfect now. Now the tests fail, as expected. To run these locally, you can do the steps in the tutorial: https://docs.buildbot.net/current/developer/quickstart.html#create-a-buildbot-python-environment.

You'll just need to change the expected paths in the git unit tests.

M4T1A5:
Thanks for the information on how to setup the testing environment. I got it working have updated all the relevant tests(and looks all tests are passing now).

Issue #4770, opened 2019-05-15
MasterShellCommand should not print master's environment variables by default


aj062:
MasterShellCommand logs all the environment variables on master by default. Many times environment variables contains passwords and API Keys. This is pretty bad to print all the environment variable by default. It should be printed only when someone explicitly sets logEnviron=True.

PR #4772, opened 2019-05-17
MasterShellCommand should not print master's environment variables by default

aj062:
Fixed https://github.com/buildbot/buildbot/issues/4770

### DIFF
--- a/master/buildbot/newsfragments/mastershellcommand_logenviron.bugfix
+++ b/master/buildbot/newsfragments/mastershellcommand_logenviron.bugfix
@@ -1,0 +1,1 @@
+MasterShellCommand should not print master's environment variables by default (:issue:`4770`)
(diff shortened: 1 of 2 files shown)


p12tic:
We can't change API like that, we would need an API break with Buildbot 3.0 for change like this.

I think a better and more flexible solution would be to implement a config key that would set the default value of the logEnviron settings throughout the master. In addition to that, we could have a blacklist of environment variables that are never printed.

Let's hear from @tardyp before committing to any approach.

tardyp:
I am no big fan of storing secrets in environment variables. Sometimes it is however the most straightforward solution.

For worker, we have a special builtbot.tac that I wrote with support for blacklisting environement variables

https://github.com/buildbot/buildbot/blob/master/worker/docker/buildbot.tac#L25

For Master, we have secret manager, which should handle most of the secrets. The key to the vault can be transmitted as a file.
I would expect those with real need to use secret in environnement variables to destroy them with same method in master.cfg

Environment variables are such a common issue when debugging difference between CI and developer environment that I am not really in favor of changing the default.
People would have to manually debug it anyway by using env in their script, and then leak the keys.

PR #4781, opened 2019-05-22
GitTag: Fix tagName arguments not in renderable list

chapipo:
The parameter tagName of the GitTag step is not in the renderable list.

I don't know if I should create a unit test for testing that all parameters are renderables ? I looked in the other git buildstep unit test and I don't found any.

## Contributor Checklist:


p12tic:
> I don't know if I should create a unit test for testing that all parameters are renderables ? I looked in the other git buildstep unit test and I don't found any.

I think we'll eventually need some generic way to check whether a class accepts parameters correctly. That is, whether they're renderable, whether the class checks parameter types and so on.

PR #4798, opened 2019-05-29
docs: add guidelines for database server usage

rjarry:
I was not sure where such a guide would fit best in the docs. Please tell me if there is a more appropriate section.


tardyp:
Looks good (build issue though)
I would also add a note in the global configuration page.

https://github.com/buildbot/buildbot/blob/b251035aca12be84397d59394403f9a15dd17a8b/master/docs/manual/configuration/global.rst#id1

Issue #4809, opened 2019-06-05
GitPoller only triggers on pushes with no commits when the new revision is the tip of another branch


adbull:
Configure buildbot with `GitPoller(buildPushesWithNoCommits=True)` to monitor branches A and B of a git repo. Push commits 1 and 2 to branch A. Then make B point to commit 1, and push.

This should trigger the poller, as it's a new push (despite having no new commits). However, it won't, because the logic in `GitPoller._process_changes` only triggers when the new revision is the tip of another branch.

PR #4810, opened 2019-06-05
gitpoller: fix pushes with no commits, not a branch tip (#4809)

adbull:
## Contributor Checklist:


adbull:
Think this should fix the issue, but not sure the best way to test it?

p12tic:
Thanks for the work fixing this. For tests I guess your best bet is test_changes_gitpoller.py file.

adbull:
Tests added, should be good to go.

PR #4811, opened 2019-06-05
Virtual builders in scheduler

tardyp:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_process_properties.py
+++ b/master/buildbot/test/unit/test_process_properties.py
@@ -60,7 +60,7 @@ class FakeSource:
             'revision': self.revision
         }
         return ds
-
+    asSSDict = asDict
 
 @implementer(IRenderable)
 class DeferredRenderable:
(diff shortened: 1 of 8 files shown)


master/buildbot/test/unit/test_process_properties.py
```diff
@@ -990,9 +990,6 @@ def testUpdateFromPropertiesNoRuntime(self):
         self.assertEqual(self.props.getPropertySource('x'), 'old')
 
     def test_setProperty_notJsonable(self):
-        with self.assertRaises(TypeError):
```
p12tic: What happens if we set non-jsonable renderable? I think we should have a test that verifies the effect in that case.

master/docs/manual/configuration/builders.rst
```diff
@@ -211,4 +211,15 @@ The virtual builder metadata is configured with the following properties:
 
 * ``virtual_builder_tags``: The tags for the virtual builder.
 
+You can also use virtual builders with :bb:sched:`SingleBranchScheduler`. For example::
```
ari: `You can also use virtual builders with :bb:sched:`SingleBranchScheduler` for example if you want to automatically build all branches in your project without having to manually create a new builder each time one is added.`

Something like that might make it clearer as to how to solve this common use-case.

master/buildbot/test/unit/test_process_properties.py
tardyp: There will be an exception during the build.
The point of this check was to raise an exception as soon as possible to ease the debug, but now we accept renderable, the check inside db api is done:

https://github.com/buildbot/buildbot/blob/4ef65c3ab815db1b997cf359c74b0ab1185a3d68/master/buildbot/db/builds.py#L205

PR #4820, opened 2019-06-08
www: Update dependencies

p12tic:
This PR updates Angular to latest version and fixes data_module version to not to clash with Bower versioning.

### DIFF
--- a/www/base/package.json
+++ b/www/base/package.json
@@ -12,7 +12,7 @@
     },
     "devDependencies": {
         "@babel/plugin-syntax-dynamic-import": "^7.2.0",
-        "angular-mocks": "1.5.3",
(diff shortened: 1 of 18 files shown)


tardyp:
Did you make extensive check that everything is working?

Upgrading 2 versions of angular is usually not this harmless.

p12tic:
I did not make a more extensive check than when doing the webpack migration. I think that PR got merged too early, but I did not have time to finish it at that moment and it was blocking Rajdeep so merging was the the best way forward. My plan was to do extensive testing and to fix bugs before releasing 2.4, so I thought that we could do an Angular upgrade too, so that we only need to test once, not twice - after webpack PR and also after Angular upgrade.

tardyp:
fair enough

Issue #4823, opened 2019-06-10
Accidental multiple starts of force build


vibbo:
When on slow client or when force scheduler is too slow, force build can be started multiple times by pressing "Start Build" button. This consumes unnecessary resources and can be confusing for users as they are notified from multiple builds.

p12tic:
Fixed in #4827.

PR #4827, opened 2019-06-11
www: Prevent accidental start of multiple force builds

vibbo:
Multiple force builds can be started if client is slow or if force
scheduler takes too long to execute. We disable the start button after it is
pressed to prevent this behavior.

Issue #4823

## Contributor Checklist:


tardyp:
I think our JS test suite is at a state where we can require unit tests. They don't detect much issues, ans are complex to run/understand. I would merge this as is.

vibbo:
Can we merge this without unittest as tardyp suggested? It would get too much time for me to familiarize myself with JS test suite...

PR #4832, opened 2019-06-13
test: Fix order of parameterized import

p12tic:
isort started changing the order of imports of parameterized. This PR commits it to code.


p12tic:
Looks like I have a different version of isort locally than in the CI.

PR #4834, opened 2019-06-14
www: Reinstante support for proxy for frontend development.

p12tic:
This PR brings back support for proxy for frontend development that has been removed in v2.3.0.

## Contributor Checklist:

### DIFF
--- a/www/base/frontend_proxy.js
+++ b/www/base/frontend_proxy.js
@@ -1,0 +1,79 @@
+#!/usr/bin/env node
+'use strict';
+
+const argv = require('minimist')(process.argv);
+const fs = require('fs');
+const path = require('path');
+const http = require('http');
+const httpProxy = require('http-proxy');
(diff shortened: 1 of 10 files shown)


www/base/frontend_proxy.js
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env node
+'use strict';
+
+const argv = require('minimist')(process.argv);
```
tardyp: I think this would be less hacky if we implemented this proxy in python aiohttp.
ex: https://github.com/oetiker/aio-reverse-proxy/blob/master/paraview-proxy.py

This would avoid the process index hack

p12tic: Thanks for suggestion, indeed python proxy would be better. Could we have the old implementation in the interim though?

tardyp: ok

PR #4835, opened 2019-06-14
optimize janitor by splitting an expensive UPDATE SQL query

bertranddemiddelaer:
optimize janitor by splitting an expensive SQL query (the list
feeding the IN may contain millions of entries) into 2 quick
queries, using the fact that steps.id is autoincremented *and indexed*

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -356,11 +356,21 @@ class LogsConnectorComponent(base.DBConnectorComponent):
 
             # update log types older than timestamps
             # we do it first to avoid having UI discrepancy
+
+            # SELECT steps.id from steps WHERE steps.started_at < older_than_timestamp ORDER BY steps.id DESC LIMIT 1;
+            res = conn.execute(
+                sa.select([model.steps.c.id])
+                .where(model.steps.c.started_at < older_than_timestamp)
+                .order_by(model.steps.c.id.desc())
+                .limit(1)
+            )
+            stepid_max = res.fetchone()[0]
+            res.close()
+
+            # UPDATE logs SET logs.type = 'd' WHERE logs.stepid <= stepid_max;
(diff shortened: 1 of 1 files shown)


master/buildbot/db/logs.py
```diff
@@ -356,11 +356,21 @@ def thddeleteOldLogs(conn):
...
             # we do it first to avoid having UI discrepancy
+
+            # SELECT steps.id from steps WHERE steps.started_at < older_than_timestamp ORDER BY steps.id DESC LIMIT 1;
+            res = conn.execute(
+                sa.select([model.steps.c.id])
+                .where(model.steps.c.started_at < older_than_timestamp)
+                .order_by(model.steps.c.id.desc())
+                .limit(1)
+            )
+            stepid_max = res.fetchone()[0]
```
p12tic: Could we add a comment explaining this? Something like: "we utilize the fact that steps.id is autoincrement, thus step start times are effectively sorted and we only need to find the step at the upper bound of steps to update.

tardyp:
Looks like the main mistake is that timestamps are not indexed.
I think this would help a lot of api usecases to get that indexed.
Waiting for that this looks good.

bertranddemiddelaer:
> Looks like the main mistake is that timestamps are not indexed.
> I think this would help a lot of api usecases to get that indexed.
> Waiting for that this looks good.

This is still an issue, and adding an index did not help much in my case because the original subrequest fed the IN list with millions of items :( Changing the requests made the cost a matter of seconds where it was previously minutes (and a "show full processlist" in mysql console shown that the original UPDATE request spent nearly all its time in preparing status)

master/buildbot/db/logs.py
bertranddemiddelaer: Will do :)

PR #4842, opened 2019-06-15
GitHubStatusPush: Gracefully handle null 'branch'

zware:
## Contributor Checklist:

* ~~I have updated the appropriate documentation~~ No documentation update necessary.

This fixes an issue that we've had on buildbot.python.org, presumably with builders triggered by a daily scheduler rather than a push scheduler.  It seems to just leave an error message in the log, with no ill effects beyond not reporting the status.

I have not updated tests for this, as I'm not familiar enough with how the tests are set up (or how exactly GitHubStatusPush even works) to know what to update or how.  I'm happy to take suggestions, or the branch is open to you :)

### DIFF
--- a/master/buildbot/newsfragments/fix-githubstatuspush-null-branch.bugfix
+++ b/master/buildbot/newsfragments/fix-githubstatuspush-null-branch.bugfix
@@ -1,0 +1,3 @@
+Fixed handling
+of null ``branch`` property
+in :py:class:`~buildbot.reporters.github.GitHubStatusPush`'s ``send()`` method.
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -129,12 +129,12 @@ class GitHubStatusPush(http.HttpStatusPushBase):
 
         project = sourcestamps[0]['project']
 
+        issue = None
         branch = props['branch']
-        m = re.search(r"refs/pull/([0-9]*)/merge", branch)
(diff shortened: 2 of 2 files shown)


master/buildbot/reporters/github.py
```diff
@@ -129,12 +129,12 @@ def send(self, build):
 
         project = sourcestamps[0]['project']
 
+        issue = None
         branch = props['branch']
```
tardyp: probably if there is no branch property, there should be a branch key in the sourcestamp disctionary.

p12tic: Indeed, we should check the sourcestamp for the branch. It's not guaranteed to be there though.

p12tic:
The tests live in test_reporter_github.py file, insertTestData in test/util/reporter.py will need to be updated with a new argument to not insert branch property. Long term we'll need to refactor that function to allow full customization of test data, but for now let's just do the minimum.

master/buildbot/newsfragments/fix-githubstatuspush-null-branch.bugfix
```diff
@@ -0,0 +1,3 @@
+Fixed handling
```
sa2ajj: should this be a single line?

p12tic:
Removing `stalled` label as the code has ended up in the codebase by some other PR.

PR #4843, opened 2019-06-16
Update PULL_REQUEST_TEMPLATE.md

tardyp:
this changes the PR template now that we have a nicer documentation


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,9 +1,14 @@
 ## Remove this paragraph
+
+If you don't remove this paragraph from the pull request description, this means you didn't read our contributor documentation, and your patch might be rejected!
```
p12tic: Maybe change "and your patch might be rejected" to "is likely to have issues and in that case will not be merged as is"?

PR #4847, opened 2019-06-18
fix Exception thrown when there is no log to delete

bertranddemiddelaer:
## Contributor Checklist:


bertranddemiddelaer:
> Hello, as per PR template, we need a unit test for this case, a newsfragment. The doc update is not relevent.

This is done, now. :)

PR #4854, opened 2019-06-20
add an index to steps.started_at

bertranddemiddelaer:
## Contributor Checklist:

* [x] I have added a unit test


bertranddemiddelaer:
I am not sure if there is something to add to update existing buildbot dbs and if you need/want to do that.

bertranddemiddelaer:
This should be complete with upgrade script and unit test as well, now. Do not hesitate to tell me if something is missing as I am not used to change buildbot db model :)

Issue #4745, opened 2019-05-02
Don't timeout during clone / checkout


mojca:
During `git clone` buildbot (2.2) seems to supress all output. A big disadvantage is that it can no longer differentiate between situation when git (or another VCS) is still happily doing its job and situations when something actually hanged. If the command kept outputting something, buildbot wouldn't need to time out.

In my case I wanted to fetch https://github.com/tex-live/luatex.git which is apparently more than 1 GB during checkout.

Note that also #3281 is true for me, but it's orthogonal and I'm actually glad that this other ticket was not addressed at this very moment. I figured out that I had 16 instances of `git` running, all started by buildbot in the last two hours. The internet connection where I'm now is horrible, so I was glad that those thousand megabytes kept being fetched after the timeout occured :) :) :) Of course, it would be nice to fix both.

mojca:
Oh, wait. I now noticed the option

> progress (optional)
>     Passes the (--progress) flag to (git fetch). This solves issues of long fetches being killed due to lack of output, but requires Git 1.7.2 or later.

which presumably solves just that. This makes me wonder: doesn't buildbot immediately check for git version at the beginning? It could automatically turn this on when git is new enough.

rajdeepbharati:
git version can be checked and then the value of progress can be set accordingly here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/steps/source/git.py#L84

rajdeepbharati:
@p12tic do you want me to send a PR?

p12tic:
@rajdeepbharati : That would be great! We already have some version parsing in buildbot/master/buildbot/util/git.py.

rajdeepbharati:
@p12tic I probably can't use that method since progress needs to be set inside `__init__` constructor of `class Git`

p12tic:
> @p12tic I probably can't use that method since progress needs to be set inside `__init__` constructor of `class Git`

You don't need to. You just need to call it before you make a decision on whether to add `--progress` flag. If you check the code flow you'll see that it's already done for you.

PR #4857, opened 2019-06-22
Set progress=True for supported Git versions.

rajdeepbharati:
For Buildsteps, use --progress flag on Git 1.7.2 or later.

Fixes https://github.com/buildbot/buildbot/issues/4745

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -74,10 +74,15 @@ class Git(Source, GitStepMixin):
     def __init__(self, repourl=None, branch='HEAD', mode='incremental', method=None,
                  reference=None, submodules=False, shallow=False, progress=False, retryFetch=False,
                  clobberOnFailure=False, getDescription=False, config=None,
-                 origin=None, sshPrivateKey=None, sshHostKey=None, sshKnownHosts=None, **kwargs):
+                 origin=None, sshPrivateKey=None, sshHostKey=None, sshKnownHosts=None, supportsProgress=None, **kwargs):
 
         if not getDescription and not isinstance(getDescription, dict):
             getDescription = False
+
+        self.supportsProgress = supportsProgress
+
+        if supportsProgress:
+            progress = True
 
         self.branch = branch
         self.method = method
--- a/master/buildbot/test/unit/test_steps_source_git.py
+++ b/master/buildbot/test/unit/test_steps_source_git.py
@@ -1225,7 +1225,7 @@ class TestGit(sourcesteps.SourceStepMixin,
     def test_mode_full_clean_no_existing_repo_with_origin(self):
         self.setupStep(
             self.stepClass(repourl='http://github.com/buildbot/buildbot.git',
-                           mode='full', method='clean', origin='foo'))
+                           mode='full', method='clean', origin='foo', progress=True))
         self.expectCommands(
             ExpectShell(workdir='wkdir',
                         command=['git', '--version'])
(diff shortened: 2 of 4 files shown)


rajdeepbharati:
@p12tic could you please review

master/buildbot/test/unit/test_steps_source_git.py
```diff
@@ -1225,7 +1225,7 @@ def test_mode_full_clean_no_existing_repo_branch(self):
     def test_mode_full_clean_no_existing_repo_with_origin(self):
         self.setupStep(
             self.stepClass(repourl='http://github.com/buildbot/buildbot.git',
-                           mode='full', method='clean', origin='foo'))
+                           mode='full', method='clean', origin='foo', progress=True))
```
p12tic: Why do we add progress param to most tests? This will ensure that we don't test the code path when progress is not set.

master/buildbot/steps/source/git.py
```diff
@@ -74,11 +74,16 @@ class Git(Source, GitStepMixin):
     def __init__(self, repourl=None, branch='HEAD', mode='incremental', method=None,
                  reference=None, submodules=False, shallow=False, progress=False, retryFetch=False,
                  clobberOnFailure=False, getDescription=False, config=None,
-                 origin=None, sshPrivateKey=None, sshHostKey=None, sshKnownHosts=None, **kwargs):
+                 origin=None, sshPrivateKey=None, sshHostKey=None, sshKnownHosts=None, supportsProgress=None, **kwargs):
```
p12tic: This param exposed to the user, but one shouldn't need to specify if the git on the worker supports the flag. Better, just pass `progress=True` by default and don't add the `--progress` flag if it turns out that we don't support progress. A warning should be logged in that case.

master/buildbot/test/unit/test_steps_source_git.py
p12tic: We should instead leave the setup part of most tests as is, update the changed tests with new behavior and also add a test for checking whether explicit progress=True is disabled on old git and whether progress=False disables `--progress` on new it.

master/docs/manual/configuration/buildsteps.rst
```diff
@@ -385,6 +385,7 @@ The Git step takes the following arguments:
 ``progress`` (optional)
    Passes the (``--progress``) flag to (:command:`git fetch`).
    This solves issues of long fetches being killed due to lack of output, but requires Git 1.7.2 or later.
+   Its value is True on Git 1.7.2 or later.
```
tardyp: please don't default this value to true on recent git. It defaults to false not for compat reason, but for performance reason. 
If we set progress to true, then suddently a lot more unneeded log is done.

p12tic: I wonder whether the extra logs are actually a performance problem. On long running fetches the 1-per-second ping will be the most significant factor which results in roughly 3600 lines of log per hour or something like 300kB. Compiling small C++ projects with compiler warnings produces more log output than that. Cloning small project like Buildbot produces only several hundreds lines of log.

By the way, other CIs like Travis or Circle enables this behavior by default by presenting a fake a terminal for stderr.

tardyp:
@rajdeepbharati can we finish this?

rajdeepbharati:
Sure

rajdeepbharati:
@p12tic setting progress=True by default is leading to a ton of tests failing

mojca:
Can we prevent the stale bot from closing this one? :)

PR #4861, opened 2019-06-24
Allow dots in worker names

mojca:
I would like to be able to use dots in worker names.

I'm opening a draft pull request just to get some basic feedback, but I still need to set up my environment and figure out the process of running and modifying the unit tests (I'll do that next), as well as write the newsframents of course.

The inline comments would of course be removed.

There are a lot of different regular expressions. I was pointed at the one in `buildbot/data/types.py`, but that one is not used for checking worker names. I did not yet fully investigate where it is used. I had to change `buildbot/util/identifiers.py` for the change to have the desired effect.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/util/identifiers.py
+++ b/master/buildbot/util/identifiers.py
@@ -17,7 +17,8 @@ import re
 
 from buildbot import util
 
-ident_re = re.compile('^[a-zA-Z\u00a0-\U0010ffff_-][a-zA-Z0-9\u00a0-\U0010ffff_-]*$', flags=re.UNICODE)
+# this line is needed to allow the dot in worker name
+ident_re = re.compile('^[a-zA-Z\u00a0-\U0010ffff._-][a-zA-Z0-9\u00a0-\U0010ffff._-]*$', flags=re.UNICODE)
 initial_re = re.compile('^[^a-zA-Z_-]')
 subsequent_re = re.compile('[^a-zA-Z0-9_-]')
 trailing_digits_re = re.compile('_([0-9]+)$')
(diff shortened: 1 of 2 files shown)


master/buildbot/util/identifiers.py
```diff
@@ -17,7 +17,8 @@
 
 from buildbot import util
 
-ident_re = re.compile('^[a-zA-Z\u00a0-\U0010ffff_-][a-zA-Z0-9\u00a0-\U0010ffff_-]*$', flags=re.UNICODE)
+# this line is needed to allow the dot in worker name
+ident_re = re.compile('^[a-zA-Z\u00a0-\U0010ffff._-][a-zA-Z0-9\u00a0-\U0010ffff._-]*$', flags=re.UNICODE)
```
tardyp: Looking at the error message, looks like this one is the correct one.

https://github.com/buildbot/buildbot/blob/d913f1ec10dc7b751d414c090b83644c192d85a4/master/buildbot/config.py#L683

probably we should not depluciate this re, and import the one from util identifier everywhere

mojca: Yes, this tiny change addresses my problem, as explained in the comment(s).
I added the others just because they seem somehow related. But if some code refactoring is needed, I'm probably not the best candidate to do it.

I'm also surprised by the number of failures this change leads to. I was definitely expecting
```
twisted.trial.unittest.FailTest: no messages for 'a.b.c.d'
```
(which I saw before), but not as many others.

Let me know if I should try to pursue and check unit tests or if there is something more fundamental that you would need to change first (like decreasing the number of regexp copies etc.).

I definitely don't need to change the other two locations for now, and I assume that the number of failed unit tests might decrease after I remove the dot from `identRe = re.compile('^[a-zA-Z._-][a-zA-Z0-9._-]*$')`.

PR #4863, opened 2019-06-24
limit the number of logs row to update during janitor

bertranddemiddelaer:
## Contributor Checklist:

None


bertranddemiddelaer:
Not sure this needs a newsfragment, but this change has a big impact for a logs table with more than 10 millions rows matching the WHERE condition during the UPDATE query.

bertranddemiddelaer:
> Looks good to me, thank you. Could we still have a newsfragment? We forgot to add one for your last optimization, this could be a chance to add a release note for both.

I've just added the newsfragment :)

PR #4864, opened 2019-06-25
www: Warn users who connect using unsupported browsers

p12tic:
This PR adds a widget that warns users who connect using unsupported browsers. Additionally, the duplicated babel config has been moved to a single place.

browser-unsupported-widget

It's possible to close that widget. Most browsers that we don't support will have significant functionality problems, so there's probably no point in making the warning less obstructive.


p12tic:
@tardyp: I moved babel config to build_common, which added a new dependency there. Since it's included to other packages via link: dependency, I needed to refresh the dependency lists of the including packages. This pulled a number of upgrades.

Issue #4786, opened 2019-05-22
Feature Request: GerritEventLogPoller and GerritChangeSource don't duplicate events


cheshirekow:
Currently if both `GerritEventLogPoller` and `GerritChangeSource` are used, they will both generate changes for the same event, leading to duplicate changes (and builds). It would be nice if they could be used together because:

1. `GerritChangeSource` is event-based and can respond instantaneously to events on the Gerrit server, but it is stateless. If the connection to the Gerrit server goes down, any events that happen in the meantime are lost. 
2. `GerritEventLogPoller` is stateful so that any event that happens during a network connection failure will be picked up on the next poll(). However, it's a polling loop so polling rate either introduces overhead or can lead to a delay in picking up changes. 

Fortunately the underlying data source in both mechanisms is pretty similar. Both API's provide a changes one-per-line as a JSON object. A quick test for the "patchset-created" event suggests that they are pretty similar but not quite the same. In my test the `events-log` event contains `"project": {"name": "buildbot"}` while the `stream-events` line contains `"project": "buildbot"`. 

I'm willing to try and tackle this feature but I could use some help. I'm thinking that the easiest thing to do is to hash canonicalized event lines and store those hashes in the database, and skip events that hash to something we've already seen. This is what I came up with for the hash:

```.py
def hash_event(event):
  """
  Return a sha1 message digest of the canonical json for the given event
  dictionary
  """

  # For "patchset-created" the events-log JSON looks like:
  #   "project": {"name": "buildbot"}
  # while the stream-events JSON looks like:
  #   "project": "buildbot"
  # so we canonicalize them to the latter
  if ("project" in event and
      isinstance(event["project"], dict) and
      "name" in event["project"]):
    event = dict(event)
    event["project"] = event["project"]["name"]
  return hashlib.sha1(json.dumps(event, sort_keys=True).encode("utf-8")).hex_digest()
```
Then `GerritChangeSourceBase.addChangeFromEvent` (`gerritchangesource.py:176`) could change to something like:

```.py
def addChangeFromEvent(self, properties, event):
  event_hash = hash_event(event)
  try:
    cursor.execute(
      "INSERT INTO gerrit_event_hashes SET event_hash=?", (event_hash,))
  except sqlite3.IntegrityError:
    return
# ...
```
But I'm not sure what the buildbot way of executing this sqlstatement and catching the error would be. Ideally I would like to do the following sqlite setup as well:
```.sql
[...]

PR #4865, opened 2019-06-26
De-duplicate gerrit change events between the Stream and Poller

cheshirekow:
This change adds a new database table storing a message digest of the normalized Gerrit event JSON for both `GerritEventLogPoller` and `GerritChangeSource`. If either source encounters an event which has already been logged, it will discard the event. 

In addition, if the `GerritEventLogPoller` has no recorded "last successful timestamp" it performs the next query using a lookback period (rather than "now") such that queued events are actually picked up. 

Closes: #4786 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -13,6 +13,7 @@
 #
 # Copyright Buildbot Team Members
 
+import copy
 import datetime
 import json
 
@@ -28,6 +29,35 @@ from buildbot.changes import base
 from buildbot.changes.filter import ChangeFilter
 from buildbot.util import bytes2unicode
 from buildbot.util import httpclientservice
+
+
+def _canonicalize_event(event):
+    """
+    Return an event dictionary which is consistent between the gerrit
+    event stream and the gerrit event log formats.
+    """
+    # For "patchset-created" the events-log JSON looks like:
+    #   "project": {"name": "buildbot"}
+    # while the stream-events JSON looks like:
+    #   "project": "buildbot"
+    # so we canonicalize them to the latter
+    if "change" not in event:
+        return event
+
+    change = event["change"]
+    if "project" not in change:
+        return event
+
+    project = change["project"]
+    if not isinstance(project, dict):
+        return event
+
+    if "name" not in project:
+        return event
+
+    event = copy.deepcopy(event)
+    event["change"]["project"] = project["name"]
+    return event
 
 
 class GerritChangeFilter(ChangeFilter):
@@ -127,23 +157,52 @@ class GerritChangeSourceBase(base.ChangeSource):
 
         properties = {}
         flatten(properties, "event", event)
-        event_with_change = "change" in event and "patchSet" in event
+        properties["event.source"] = self.__class__.__name__
         func_name = "eventReceived_%s" % event["type"].replace("-", "_")
         func = getattr(self, func_name, None)
-        if func is None and event_with_change:
+        if func is None:
             return self.addChangeFromEvent(properties, event)
-        elif func is None:
-            if self.debug:
-                log.msg("unsupported event %s" % (event["type"],))
-            return defer.succeed(None)
-        else:
-            return func(properties, event)
 
+        return func(properties, event)
+
+    @defer.inlineCallbacks
     def addChange(self, chdict):
-        d = self.master.data.updates.addChange(**chdict)
-        # eat failures..
-        d.addErrback(log.err, 'error adding change from GerritChangeSource')
-        return d
+        stampdict = {
+            "branch": chdict["branch"],
+            "revision": chdict["revision"],
+            "patch_author": chdict["author"],
+            "patch_comment": chdict["comments"],
+            "repository": chdict["repository"],
+            "project": chdict["project"],
+        }
+
+        stampid, found_existing = yield(
+             self.master.db.sourcestamps.findOrCreateId(**stampdict))
+
+        if found_existing:
+            if self.debug or True:
+                eventstr = "{}/{} -- {}:{}".format(
+                    self.gitBaseURL, chdict["project"], chdict["branch"],
+                    chdict["revision"])
+                message = (
+                    "gerrit: duplicate change event {} by {}"
+                    .format(eventstr, self.__class__.__name__))
+                log.msg(message.encode("utf-8"))
+            defer.returnValue(None)
+
+        if self.debug:
+            eventstr = "{} -- {}:{}".format(
+                chdict["repository"], chdict["branch"], chdict["revision"])
+            message = (
+                "gerrit: adding change from {} in {}"
+                .format(eventstr, self.__class__.__name__))
+            log.msg(message.encode("utf-8"))
+
+        try:
+            yield self.master.data.updates.addChange(**chdict)
+        except Exception:
+            # eat failures..
+            log.err('error adding change from GerritChangeSource')
 
     def getGroupingPolicyFromEvent(self, event):
         # At the moment, buildbot's change grouping strategy is hardcoded at various place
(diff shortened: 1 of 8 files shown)


cheshirekow:
How the heck did coverage of these two files go *down*?
```
master/buildbot/scripts/cleanupdb.py
master/buildbot/worker_transition.py
```

master/buildbot/db/model.py
```diff
@@ -678,6 +678,15 @@ class Model(base.DBConnectorComponent):
         sa.Column('last_active', sa.Integer, nullable=False),
     )
 
+    # gerrit change hashes
+    gerriteventhashes = sautils.Table(
```
tardyp: sourcestamp table has exactly this purpose, and is generic. Please use this table instead of adding another one.
http://docs.buildbot.net/current/developer/database.html#buildbot.db.sourcestamps.SourceStampsConnectorComponent
You will need to use findSourceStampId()

and you will need to modify it so that the returned structure properly reports that a new sourcestamp has been created (instead of having to test if created_at ~= now())

tardyp: There is also the state table, which can be used for similar purpose.
http://docs.buildbot.net/current/developer/database.html#module-buildbot.db.state

We don't add tables specific to a particular purpose.

tardyp:
Hi, will you have time to finish this very useful change?

cheshirekow:
I will, but not probably not until next week. I am hoping to work on it then :/.

master/buildbot/changes/gerritchangesource.py
```diff
@@ -29,6 +30,33 @@
 from buildbot.util import httpclientservice
 
 
+def canonicalize_event(event):
```
sa2ajj: is this function a part of the public API for this module?  if not, please use `_canonicalize_event`

master/buildbot/changes/gerritchangesource.py
```diff
@@ -29,6 +30,33 @@
...
+    """
+    Return an event dictionary which is consistent between the gerrit
+    event stream and the gerrit event log formats.
+    """
+    # For "patchset-created" the events-log JSON looks like:
+    #   "project": {"name": "buildbot"}
+    # while the stream-events JSON looks like:
+    #   "project": "buildbot"
+    # so we canonicalize them to the latter
+    if ("change" in event and "project" in event["change"] and
```
sa2ajj: what bothers me a bit is a multiple lookups of the same key... it's probably not a problem in a production environment, however it's not a good style in general...

master/buildbot/changes/gerritchangesource.py
```diff
@@ -29,6 +30,33 @@
...
+            isinstance(event["change"]["project"], dict) and
+            "name" in event["change"]["project"]):
+        event = dict(event)
+        event["change"] = dict(event["change"])
+        event["change"]["project"] = event["change"]["project"]["name"]
+    return event
```
sa2ajj: you probably want to add an empty line before the `return` statement

master/buildbot/changes/gerritchangesource.py
```diff
@@ -123,19 +151,23 @@ def flatten(properties, base, event):
...
         func_name = "eventReceived_%s" % event["type"].replace("-", "_")
         func = getattr(self, func_name, None)
-        if func is None and event_with_change:
+        if func is None:
             return self.addChangeFromEvent(properties, event)
-        elif func is None:
-            if self.debug:
-                log.msg("unsupported event %s" % (event["type"],))
-            return defer.succeed(None)
         else:
```
sa2ajj: you do not need this `else:` after the `return` statement

master/buildbot/changes/gerritchangesource.py
```diff
@@ -371,10 +436,14 @@ def now():
...
         last_event_ts = yield self.master.db.state.getState(self._oid, 'last_event_ts', None)
         if last_event_ts is None:
-            last_event = self.now()
+            # If there is not last event time stored in the database, then set
+            # the last event time to some historical look-back
+            last_event = self.now() - datetime.timedelta(days=self._first_fetch_lookback)
         else:
             last_event = datetime.datetime.utcfromtimestamp(last_event_ts)
         last_event_formatted = last_event.strftime("%Y-%m-%d %H:%M:%S")
+        if self.debug:
```
sa2ajj: empty line before this condition check?

master/buildbot/changes/gerritchangesource.py
```diff
@@ -371,10 +436,14 @@ def now():
...
+            log.msg("Polling gerrit: {}".format(last_event_formatted).encode("utf-8"))
         res = yield self._http.get("/plugins/events-log/events/", params=dict(t1=last_event_formatted))
```
sa2ajj: empty line after that debug block?

master/buildbot/changes/gerritchangesource.py
```diff
@@ -154,22 +186,49 @@ def getGroupingPolicyFromEvent(self, event):
...
+                    self.getGroupingPolicyFromEvent(event),
+                    event["patchSet"]["revision"])
+                message = (
+                    "gerrit: duplicate change event {} by {}"
+                    .format(eventstr, self.__class__.__name__))
+                log.msg(message.encode("utf-8"))
+            return defer.returnValue(None)
+
+        cb = yield self.addChange({
+            'author': _gerrit_user_to_author(event_change["owner"]),
```
sa2ajj: nitpick: some consistency about quote usage would be appricated

cheshirekow:
Yes, yes... Thank you stale-bot. I this is still in my queue but I haven't gotten to it yet.

master/buildbot/db/model.py
cheshirekow: What is the meaning of the `codebase` parameter to the `sourcestamp`? Documentation says it's never `None` but it does not appear to be part of the gerrit `chdict` (unless it's somewhere in `properties`).

master/buildbot/changes/gerritchangesource.py
cheshirekow: Done

master/buildbot/changes/gerritchangesource.py
cheshirekow: I disagree regarding style but I don't really care one way or the other. Done.

master/buildbot/changes/gerritchangesource.py
cheshirekow: Done

master/buildbot/changes/gerritchangesource.py
cheshirekow: Done

master/buildbot/changes/gerritchangesource.py
cheshirekow: Done

master/buildbot/changes/gerritchangesource.py
cheshirekow: Done

master/buildbot/changes/gerritchangesource.py
cheshirekow: Done

master/buildbot/db/model.py
cheshirekow: Sorry, I don't quite understand this statement:

> and you will need to modify it so that the returned structure properly reports that a new sourcestamp has been created (instead of having to test if created_at ~= now())

The structure returned from where? Are you saying that I need to modify `findSourceStampId()` itself?

cheshirekow: Oh, now I see. findSourceStampId always returns an id. If it doesn't already exist, it creates one. So we need a way to know if it created one.

tardyp:
Thanks for working on it! there are still a little bit of flake8 and pylint issues to fix.

cheshirekow:
> Thanks for working on it! there are still a little bit of flake8 and pylint issues to fix.

:+1: I also need to test it on a real deployment and make sure it actually works.

cheshirekow:
@tardyp Is there a reason that `master[test]` has different depends listed in `setup.py` than `requirements-ci.txt`? For some reason it requires a very old `'pylint<1.7.0'`. Are there instructions on executing the linters locally? This is what I tried:

```
make virtualenv
source .venv/bin/activate
make -C master pylint
```

But pylint crashes out. This seemed to work though:
```
make virtualenv
source .venv/bin/activate
pip install requirements-ci.txt
make -C master pylint
```

cheshirekow:
So... how about a review then?

p12tic:
@cheshirekow: Many thanks for the effort put into this PR and sorry for the delays in reviewing.

> Is there a reason that master[test] has different depends listed in setup.py than requirements-ci.txt

I believe that `master[test]` dependency list only defines the minimum requirements and `requirements-ci.txt` tracks the latest packages as we use the latter for the CI, so we want to notice issues early. The requirements file is updated automatically, so it's easy for `master[test]` dependencies to become out of date.

PR #4848, opened 2019-06-18
dev-proxy implementation in asyncio and aiohttp

tardyp:
Here is an aiohttp dev-proxy. 

## Contributor Checklist:

* [x] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)
* [x] I have updated the appropriate documentation
* [ ] I have updated the unit tests

PR #4866, opened 2019-06-26
www: Enable frontend unit tests on CI and fix them

p12tic:
Looks like that with the landing of Webpack rework we no longer run unit tests on the CI. I added a Makefile rule to run all frontend unit tests, fixed them and added a new rule to .bbtravis.


p12tic:
I'm looking into the tests, they're broken by #4848. I'll have fixes soon.

PR #4884, opened 2019-07-09
mtrlogobserver: make parallel argument renderable

shinnok:
This way we can pass renderable parallel parameters to MTR step class instantiations.


shinnok:
Context here:
https://github.com/MariaDB/mariadb.org-tools/commit/97182ad1a4927668f52edf14e23754b45c233bcc

PR #4886, opened 2019-07-09
Fix various errors in try client with Python 3

gbenson:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/try-client-python3-fixes.bugfix
+++ b/master/buildbot/newsfragments/try-client-python3-fixes.bugfix
@@ -1,0 +1,1 @@
+:issue:`4765`: Fix various errors in try client with Python 3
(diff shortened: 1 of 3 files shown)


gbenson:
FWIW I checked 'I have updated the appropriate documentation', but I don't know there is any other than the news fragment...

p12tic:
Bug fixes don't need documentation updates, so everything is perfect.

master/buildbot/newsfragments/try-client-python3-fixes.bugfix
```diff
@@ -0,0 +1 @@
+:issue:`4765`: Fix various errors in try client with Python 3
```
p12tic: I think this should've been `4746`, because 4765 points to a PR.

gbenson:
> Thanks a lot for the contribution!

Thanks for making it easy to contribute :D

PR #4896, opened 2019-07-18
Renew bitbucket's hook read

Dispetcher14:
I has set up bitbucket webhook on my buildbot instanse, and it not working. 

Okay, i read the logs, and see, that upstream bitbucket plugin use to parse another keys, not the bitbucket's from response. 
Also, i didn't understand, what means on 
`payload = json.loads(bytes2unicode(request.args[b'payload'][0]))`
There are not a body of request, isn't it? 

Summary:
getChanges returns a list of dictionaries with keys from original files. Values matchs actual body of bitbucket's requests.


Dispetcher14:
Gone to read developers guide.

PR #4897, opened 2019-07-19
Add regex support in buildstep DirectoryUpload

angeloudy:
It's quite useful to copy files selectively from a directory instead of the whole directory.
For example, `gradle` would generate build results in lots of `.xml` files which need to be copied to the master for anlalysis.

All subdirectories will still be uploaded, but for files, only paths that match the regex will be uploaded

### DIFF
--- a/master/buildbot/newsfragments/add_regex_step_DirectoryUpload.feature
+++ b/master/buildbot/newsfragments/add_regex_step_DirectoryUpload.feature
@@ -1,0 +1,1 @@
+Add support for regex in buildstep DirectoryUpload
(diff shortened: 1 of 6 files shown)


angeloudy:
Come someone have a look at this.
I tried to get tests passed, but I don't know what it's trying to test.

p12tic:
I think this is good idea, but the implementation would be much more involved, as we would need to bump worker protocol version.

angeloudy:
I've achieved the same goal with `glob=True` in `MultipleFileUpload`. Closing this PR

Issue #3471, opened 2017-07-25
Waterfall view should be filterable by tags


jangmarker:
Similar to #3470 the waterfall view would be more useful in large buildbot configurations if it could be filtered by tags.

tardyp:
grid view already implements this. This should be easy to port

alorbach:
Any news on this? 
I would really like to be able to filer by tags like in old buildbot 0.8

tardyp:
@alorbach nobody has taken the ticket AFAIK. Feel free to help. Here is a tutorial that should put you on the right direction:
https://medium.com/buildbot/buildbot-ui-plugin-for-python-developer-ef9dcfdedac0

alorbach:
Thanks for the details.

mojca:
This seems to be the same as #2892.

rajdeepbharati:
@p12tic @jangmarker Do you want me to add negative filtering too, like in https://buildbot.buildbot.net/#/builders ?

p12tic:
@rajdeepbharati I think negative filtering makes sense, but I guess that's low priority.

PR #4901, opened 2019-07-20
Allow filtering by tags in waterfall view

rajdeepbharati:
Closes https://github.com/buildbot/buildbot/issues/3471

Demo: https://drive.google.com/file/d/1LGxecjHkf3qUpG9aT5ha_9BUiTRMm5tw/view

http://ec2-18-191-33-221.us-east-2.compute.amazonaws.com/#/waterfall

Screenshot:

Screenshot 2019-07-20 at 8 45 42 PM

### DIFF
--- a/www/waterfall_view/src/module/main.module.js
+++ b/www/waterfall_view/src/module/main.module.js
@@ -195,9 +195,7 @@ var WaterfallController = (function() {
             });
 
             $rootScope.$on('$locationChangeSuccess', function() {
-                if (self.$scope.tags_filter.length) {
-                    self.renderNewData(self.$scope.tags_filter);
-                }
+                self.renderNewData(self.$scope.tags_filter);
             });
         }
 
@@ -655,7 +653,6 @@ var WaterfallController = (function() {
                 });
             }
             this.$location.search("tags", this.$scope.tags_filter);
-            this.renderNewData(this.$scope.tags_filter);
         }
 
         isTagFiltered(tag) {
@@ -669,7 +666,6 @@ var WaterfallController = (function() {
         clearTags() {
             this.$scope.tags_filter = this.tags_filter = [];
             this.$location.search("tags", this.$scope.tags_filter);
-            this.renderNewData(this.$scope.tags_filter);
         }
 
         makeTagBuilders(currentTags, builders) {
@@ -678,13 +674,8 @@ var WaterfallController = (function() {
             if (typeof currentTags != 'string') {
                 anyTagSelected = true;
                 for (const builder of builders) {
-                    let v = currentTags.every(currentTag => {
-                        if (!builder.tags.includes(currentTag)) {
-                            return false;
-                        } else {
-                            return true;
-                        }
-                    });
+                    let v = currentTags.every(currentTag =>
+                        builder.tags.includes(currentTag));
                     if (v) {
                         tag_builders.push(builder);
                     }
@@ -718,7 +709,7 @@ var WaterfallController = (function() {
             var all_tags = [];
             for (let builder of this.builders) {
                 for (let tag of builder.tags) {
-                    if (all_tags.indexOf(tag) < 0) {
+                    if (!all_tags.includes(tag)) {
                         all_tags.push(tag);
                     }
                 }
--- a/www/waterfall_view/src/module/main.module.js
+++ b/www/waterfall_view/src/module/main.module.js
@@ -11,6 +11,7 @@ import 'angular-animate';
 import '@uirouter/angularjs';
 import 'guanlecoja-ui';
 import 'buildbot-data-js';
+import _ from 'lodash';
 
 // Register new module
 class WaterfallView {
@@ -34,7 +35,8 @@ var WaterfallController = (function() {
         }
         constructor($rootElement, $scope, $q, $timeout, $window, $log,
                       $uibModal, dataService, d3Service, dataProcessorService,
-                      scaleService, bbSettingsService, glTopbarContextualActionsService) {
+                      scaleService, bbSettingsService, glTopbarContextualActionsService,
+                      $location, $rootScope) {
             this.zoomPlus = this.zoomPlus.bind(this);
             this.zoomMinus = this.zoomMinus.bind(this);
             this.renderNewData = this.renderNewData.bind(this);
@@ -43,6 +45,9 @@ var WaterfallController = (function() {
             this.$window = $window;
             this.$log = $log;
             this.$uibModal = $uibModal;
+            this.$location = $location;
+            this.$rootScope = $rootScope;
+            this.$scope.tags_filter = this.tags_filter = [];
             this.dataProcessorService = dataProcessorService;
             this.bbSettingsService = bbSettingsService;
             self = this;
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
             });
         }
 
-        renderNewData() {
+        toggleTag(tag) {
+            if (!this.$scope.tags_filter.includes(tag)) {
+                this.$scope.tags_filter.push(tag);
+            } else {
+                this.$scope.tags_filter = _.remove(this.$scope.tags_filter, function(currentTag) {
+                    return currentTag != tag;
+                });
+            }
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        isTagToggled(tag) {
+            if (this.$scope.tags_filter.includes(tag)) {
+                return 'label-success';
+            } else {
+                return 'label-default';
+            }
+        }
+
+        clearTags() {
+            this.$scope.tags_filter = this.tags_filter = [];
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        renderNewData(currentTags) {
+            currentTags = this.$location.search()['tags'];
+            if (currentTags != null) {
+                if (typeof currentTags == 'string') {
+                    this.$scope.tags_filter = this.tags_filter = [currentTags];
+                } else {
+                    this.$scope.tags_filter = this.tags_filter = currentTags;
+                }
+            }
             this.groups = this.dataProcessorService.getGroups(this.all_builders, this.builds, this.c.threshold);
             if (this.s.show_builders_without_builds.value) {
                 this.$scope.builders = this.all_builders;
             } else {
                 this.$scope.builders = (this.builders = this.dataProcessorService.filterBuilders(this.all_builders));
             }
+            var all_tags = [];
+            for (let builder of Array.from(this.builders)) {
+                for (let tag of Array.from(builder.tags)) {
+                    if (all_tags.indexOf(tag) < 0) {
+                        all_tags.push(tag);
+                    }
+                }
+            }
+            all_tags.sort();
+            this.$scope.all_tags = this.all_tags = all_tags;
             this.dataProcessorService.addStatus(this.builders);
+            let tag_builders = [];
+            let anyTagSelected = false;
+
+            if (currentTags != null) {
+                if (typeof currentTags != 'string') {
+                    anyTagSelected = true;
+                    this.$scope.builders.forEach(element => {
+                        let v = currentTags.every(currentTag => {
+                            if (!element.tags.includes(currentTag)) {
+                                return false;
+                            } else {
+                                return true;
+                            }
+                        });
+                        if (v) {
+                            tag_builders.push(element);
+                        }
+                    });
+                } else if (typeof currentTags == 'string') {
+                    anyTagSelected = true;
+                    this.$scope.builders.forEach(element => {
+                        if (element.tags.includes(currentTags)) {
+                            tag_builders.push(element);
+                        }
+                    });
+                }
+            }
+
+            if (anyTagSelected) {
+                this.$scope.builders = this.builders = tag_builders;
+            }
             this.render();
             return this.loadingMore = false;
         }
@@ -674,8 +754,11 @@ var WaterfallController = (function() {
 
 
 angular.module('waterfall_view', new WaterfallView())
-.controller('waterfallController', ['$rootElement', '$scope', '$q', '$timeout', '$window', '$log', '$uibModal', 'dataService', 'd3Service', 'dataProcessorService', 'scaleService', 'bbSettingsService', 'glTopbarContextualActionsService',
-                                    WaterfallController]);
+.controller('waterfallController', ['$rootElement', '$scope', '$q', '$timeout', '$window', '$log', '$uibModal', 'dataService', 'd3Service', 'dataProcessorService', 'scaleService', 'bbSettingsService', 'glTopbarContextualActionsService', '$location', '$rootScope',
+                                    WaterfallController])
+.config(['$locationProvider', function($locationProvider) {
+    $locationProvider.hashPrefix('');
+}]);
 
 require('./dataProcessor/dataProcessor.service.js');
 require('./main.module.js');
--- a/www/waterfall_view/src/module/main.module.js
+++ b/www/waterfall_view/src/module/main.module.js
@@ -672,7 +672,7 @@ var WaterfallController = (function() {
             this.renderNewData(this.$scope.tags_filter);
         }
 
-        setTagBuilders(currentTags, builders) {
+        makeTagBuilders(currentTags, builders) {
             let tag_builders = [];
             let anyTagSelected = false;
             if (typeof currentTags != 'string') {
@@ -730,7 +730,7 @@ var WaterfallController = (function() {
             let anyTagSelected, tag_builders;
 
             if (currentTags != null) {
-                [anyTagSelected, tag_builders] = this.setTagBuilders(currentTags, this.$scope.builders);
+                [anyTagSelected, tag_builders] = this.makeTagBuilders(currentTags, this.$scope.builders);
             }
 
             if (anyTagSelected) {
--- a/www/waterfall_view/src/module/main.module.js
+++ b/www/waterfall_view/src/module/main.module.js
@@ -652,11 +652,11 @@ var WaterfallController = (function() {
             this.renderNewData(this.$scope.tags_filter);
         }
 
-        isTagToggled(tag) {
+        isTagFiltered(tag) {
             if (this.$scope.tags_filter.includes(tag)) {
-                return 'label-success';
+                return true;
             } else {
-                return 'label-default';
+                return false;
             }
         }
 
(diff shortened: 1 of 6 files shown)


www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        renderNewData(currentTags) {
+            currentTags = this.$location.search()['tags'];
+            if (currentTags != null) {
+                if (typeof currentTags == 'string'){
+                    this.$scope.tags_filter = this.tags_filter = [currentTags];
+                } else {
+                    this.$scope.tags_filter = this.tags_filter = currentTags;
```
p12tic: Are we guaranteed to get an array if `currentTags` is not a string?

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+            var all_tags = [];
+            for (let builder of Array.from(this.builders)) {
+                for (let tag of Array.from(builder.tags)) {
+                    if (all_tags.indexOf(tag) < 0) {
+                        all_tags.push(tag);
+                    }
+                }
+            }
+            all_tags.sort();
+            this.$scope.all_tags= this.all_tags = all_tags
```
p12tic: Spacing.

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+            this.$scope.tags_filter = this.tags_filter = [];
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        renderNewData(currentTags) {
+            currentTags = this.$location.search()['tags'];
+            if (currentTags != null) {
+                if (typeof currentTags == 'string'){
+                    this.$scope.tags_filter = this.tags_filter = [currentTags];
```
p12tic: We should set bidirectional data biding shouldn't we? @tardyp I'm not 100% confident in my understanding of AngularJS concepts.

p12tic:
Looks good to me, but I had several questions.

www/waterfall_view/src/module/main.module.js
rajdeepbharati: In fact I'm getting an object in the else case

www/waterfall_view/src/module/main.module.js
rajdeepbharati: I tried using `$rootScope.$on('$locationChangeSuccess', ()=>{})` but inside that I wasn't able to change `$scope.tags_filter`, so I put the logic inside `renderNewData`

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+            } else {
+                this.$scope.tags_filter = _.remove(this.$scope.tags_filter, function(currentTag) {
+                    return currentTag != tag;
+                });
+            }
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        isTagToggled(tag) {
```
tardyp: nit: toggle is an action, not a state.

Then, you method do not return a boolean, but returns a class.

I would call this method isTagFiltered, and would push the class logic into the template

www/waterfall_view/src/module/main.module.js
tardyp: So tags_filter can be either array or object? sounds bad.

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+                }
+            }
             this.groups = this.dataProcessorService.getGroups(this.all_builders, this.builds, this.c.threshold);
             if (this.s.show_builders_without_builds.value) {
                 this.$scope.builders = this.all_builders;
             } else {
                 this.$scope.builders = (this.builders = this.dataProcessorService.filterBuilders(this.all_builders));
             }
+            var all_tags = [];
+            for (let builder of Array.from(this.builders)) {
```
tardyp: Array.from should rarely be needed. It is often a remain from decaffeinate that never been cleaned up

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+                    }
+                }
+            }
+            all_tags.sort();
+            this.$scope.all_tags = this.all_tags = all_tags;
             this.dataProcessorService.addStatus(this.builders);
+            let tag_builders = [];
+            let anyTagSelected = false;
+
+            if (currentTags != null) {
```
tardyp: cyclomatic complexity warning!

please split this into another function

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+                if (typeof currentTags != 'string') {
+                    anyTagSelected = true;
+                    this.$scope.builders.forEach(element => {
```
tardyp: its better to use for .. of instead of method calls. much faster.

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +640,89 @@ var WaterfallController = (function() {
...
+                                return true;
+                            }
+                        });
+                        if (v) {
+                            tag_builders.push(element);
+                        }
+                    });
+                } else if (typeof currentTags == 'string') {
+                    anyTagSelected = true;
+                    this.$scope.builders.forEach(element => {
```
tardyp: same here

www/waterfall_view/src/module/main.module.js
rajdeepbharati: `typeof tags_filter` is giving object but it's actually an array

rajdeepbharati: This is because javascript considers `typeof Array()` as `object`

rajdeepbharati: These issues would be solved if we use typescript

mojca:
@rajdeepbharati deployed the example [here](http://ec2-18-191-33-221.us-east-2.compute.amazonaws.com/#/waterfall).

I would like to hear your feedback. (On 0.8 it was only possible to have one tag active at any given time. On the builders page you are using a different strategy to display tags.)

From my side the feedback is approximately the following, but I would really like to see yours first:
* On my local machine I get an "infinitely long" line of tags, esp. with lots of virtual builders being present. While I could say that this could be limited to one screen width and span multiple lines, the waterfall itself is also very wide, so it's hard to say what the best solution is there.
* The "`x`" for resetting tags is so far on the right that it's useless.
* I only seem to be seeing tags from the same set of builds that are already shown in the waterfall. I need to manually type a tag that's not visible in the current waterfall view, and when I type it manually, I don't seem to get any results, at least not until I "browse back in time" to a sufficient extent that the build would be displayed on the waterfall anyway.
* It would definitely make sense to implement tag search as the next step, but that's a relatively independent task.
* Low priority (and not sure about others either): I would prefer to have the tags sorted irrespective of lower/upper case.
* Let's say that one selects the tag `python37`. Clicking almost any other tag next (say, `rust`) would result in an empty set in the next step. It could be nice to only show the 'relevant tags" when some tags are already selected, but I have no idea how difficult this would be to implement.

Overall I'm very happy that this functionality has been implemented, it's already very helpful, but I would like to get some of your feedback as well, as I'm not sure what's reasonably doable.

rajdeepbharati:
> I only seem to be seeing tags from the same set of builds that are already shown in the waterfall. I need to manually type a tag that's not visible in the current waterfall view, and when I type it manually, I don't seem to get any results, at least not until I "browse back in time" to a sufficient extent that the build would be displayed on the waterfall anyway.

Yes, this is similar to that in `/builders` page. I was wondering what you wanted (all tags or latest one), so I reproduced the behavior of the one in builders page. For example, some ports may have multiple platform tags (10_13 and 10_14). Suppose a builder has latest tag 10_14 and an old tag 10_13. In that case, when the user clicks 10_13 should we show that builder or not?

rajdeepbharati:
> On my local machine I get an "infinitely long" line of tags, esp. with lots of virtual builders being present. While I could say that this could be limited to one screen width and span multiple lines, the waterfall itself is also very wide, so it's hard to say what the best solution is there.

I can add a setting in waterfall to disable showing particular tags on the top of the view. (or only show particular tags). For example, in macports setup, we can enable showing tags which are about the platform (and not portname)

mojca:
On the `/builders` page usually only the builders which are "obsolete" are hidden, while on waterfall I'm not seeing any builders which didn't build anything for a sufficient amount of time and they just happened to have disappeared from the sight.

From the perspective of MacPorts: you most likely won't see any tags for builders that have last built a package one day ago. On the other hand, if you showed all tags without some precautions about what to do with too many tags, the list of tags would get uselessly long.

Maybe showing just visible tags + adding a search field for others would work good enough. I would like to hear feedback from others.

rajdeepbharati:
> On the /builders page usually only the builders which are "obsolete" are hidden, while on waterfall I'm not seeing any builders which didn't build anything for a sufficient amount of time and they just happened to have disappeared from the sight.

This is the old builders page (2.3.1) which uses gulp, while the new one (using webpack) doesn't work that way and old tags disappear. See: (10_15 should be there but isn't because the latest build of rust doesn't have it)

Screenshot 2019-07-23 at 2 58 21 PM
Screenshot 2019-07-23 at 2 58 49 PM

rajdeepbharati:
There are a few critical issues in the `www/base` that have come up after the webpack pr. One of them is that the dev server doesn't work in the files in this [directory](https://github.com/buildbot/buildbot/blob/master/www/base/src/app/layout.jade) (they do work in the subdirectories though). Then the breadcrumbs and icons aren't there, which is again related to `/base`

mojca:
> I can add a setting in waterfall to disable showing particular tags on the top of the view. (or only show particular tags). For example, in macports setup, we can enable showing tags which are about the platform (and not portname)

Yes, that would definitely make sense. (Obviously we would need this blacklisting or whitelisting to happen somewhere in `master.cfg` rather than asking users to manually list those tags in browser settings.)

One way would also be to define the tag list in some way similar to this:
```python
virtual_builder_tags = [{tag: "rust", isHidden: True}, "10.15", "virtual"]
```
or
```python
virtual_builder_tags = ["10.15", "virtual"]
virtual_builder_hidden_tags = ["rust"]
```
Disclaimer: I don't know if doing something like this make sense or not, or whether it is feasible to do, so don't listen to me until someone from buildbot devs comments on it :)

www/waterfall_view/src/module/main.module.js
tardyp: I think you should test more on locationChangeSuccess.

For example, if you bookmark and jump directly, it does not work on the test server:
http://ec2-18-191-33-221.us-east-2.compute.amazonaws.com/#/waterfall?tags=nodejs10

here there is not filtering whereas we should expect to see only nodejs10 builds

rajdeepbharati:
> For example, if you bookmark and jump directly, it does not work on the test server:
http://ec2-18-191-33-221.us-east-2.compute.amazonaws.com/#/waterfall?tags=nodejs10
here there is not filtering whereas we should expect to see only nodejs10 builds


Oops, I think this isn't working on the test server because of the buildbot base is of gulp (it's working fine in my local server). I wasn't able to install complete frontend on aws (it was freezing), so I just installed the waterfall.

I will write tests, nevertheless.

tardyp:
Indeed, e2e tests is a good idea.

rajdeepbharati:
> here there is not filtering whereas we should expect to see only nodejs10 builds

@tardyp I fixed this problem, but it is leading to a warning:
`TypeError: Cannot read property 'node' of undefined`
Also, the color of tags isn't vanishing when I remove all tags from the search bar manually.

An easy workaround is to set `reloadOnSearch: true`, but do you think that's a good idea?

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +646,96 @@ var WaterfallController = (function() {
...
+            }
+        }
+
+        clearTags() {
+            this.$scope.tags_filter = this.tags_filter = [];
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        setTagBuilders(currentTags, builders) {
```
tardyp: setTagBuilders does not set anything makeTagBuilders?

rajdeepbharati: you mean rename it?

rajdeepbharati:
ping: @tardyp updated.

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +646,96 @@ var WaterfallController = (function() {
...
+        toggleTag(tag) {
+            if (!this.$scope.tags_filter.includes(tag)) {
+                this.$scope.tags_filter.push(tag);
+            } else {
+                this.$scope.tags_filter = _.remove(this.$scope.tags_filter, function(currentTag) {
+                    return currentTag != tag;
+                });
+            }
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
```
tardyp: is this mandatory? I would expect $location.search eventually calls $locationChangeSuccess callback.

Please add a comment if it does not

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +646,96 @@ var WaterfallController = (function() {
...
+                return true;
+            } else {
+                return false;
+            }
+        }
+
+        clearTags() {
+            this.$scope.tags_filter = this.tags_filter = [];
+            this.$location.search("tags", this.$scope.tags_filter);
+            this.renderNewData(this.$scope.tags_filter);
```
tardyp: same here

www/waterfall_view/src/module/main.module.js
```diff
@@ -188,6 +193,12 @@ var WaterfallController = (function() {
                     return window.unbind('resize', resizeHandler);
                 });
             });
+
+            $rootScope.$on('$locationChangeSuccess', function() {
+                if (self.$scope.tags_filter.length) {
```
tardyp: if you remove all tag filters, then the view won't be updated

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +646,96 @@ var WaterfallController = (function() {
...
+            this.renderNewData(this.$scope.tags_filter);
+        }
+
+        makeTagBuilders(currentTags, builders) {
+            let tag_builders = [];
+            let anyTagSelected = false;
+            if (typeof currentTags != 'string') {
+                anyTagSelected = true;
+                for (const builder of builders) {
+                    let v = currentTags.every(currentTag => {
```
tardyp: simplify..
```suggestion
                    let v = currentTags.every(currentTag => builder.tags.includes(currentTag))
```

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +646,96 @@ var WaterfallController = (function() {
...
             this.groups = this.dataProcessorService.getGroups(this.all_builders, this.builds, this.c.threshold);
             if (this.s.show_builders_without_builds.value) {
                 this.$scope.builders = this.all_builders;
             } else {
                 this.$scope.builders = (this.builders = this.dataProcessorService.filterBuilders(this.all_builders));
             }
+            var all_tags = [];
+            for (let builder of this.builders) {
+                for (let tag of builder.tags) {
+                    if (all_tags.indexOf(tag) < 0) {
```
tardyp: includes?

rajdeepbharati:
Thank you @tardyp I have made the changes that you asked

mojca:
Thank you!

www/waterfall_view/src/module/main.module.js
```diff
@@ -635,14 +644,89 @@ var WaterfallController = (function() {
...
+            this.$location.search("tags", this.$scope.tags_filter);
+        }
+
+        makeTagBuilders(currentTags, builders) {
+            let tag_builders = [];
+            let anyTagSelected = false;
+            if (typeof currentTags != 'string') {
+                anyTagSelected = true;
+                for (const builder of builders) {
+                    let v = currentTags.every(currentTag =>
```
tardyp: The every is actually weird for me, after testing. I would rather expect any, so that you can viiew several tag at the same time.

www/waterfall_view/src/module/waterfall.tpl.jade
```diff
@@ -3,6 +3,12 @@
         .spinner
             i.fa.fa-circle-o-notch.fa-spin.fa-2x
             p loading
+    p Tags: 
+        span(ng-show="w.all_tags.length > 0" ng-repeat="tag in w.all_tags")
```
tardyp: with metabuildbot, the list of tags is on only one line. I fear that for many tags you can't see all of them without scrolling. Can we fit them in several rows?

rajdeepbharati: Ok, I will send a PR to split them across rows (responsive)

www/waterfall_view/src/module/main.module.js
rajdeepbharati: It does work with several tags at the same time

www/waterfall_view/src/module/waterfall.tpl.jade
mojca: We discussed that already, but I wasn't sure about the best solution. We expect to have 20k+ tags in which case not even multiple lines will help to avoid scrolling, and I suggested that we may need to implement functionality to hide some of those tags (hiding would be done in `master.cfg`), only show the really relevant ones, and then add search-for-tags functionality.

rajdeepbharati: would the portnames be given tags in macports?

www/waterfall_view/src/module/main.module.js
mojca: I saw that you offer three states of tags for builders: required (`and`), forbidden (`and not`), "optional" (`or`).
My ideal scenario (but not super easy to implement) might be:
* If the tag is required (with a `+`) or forbidden, hide all other tags which would result in an empty set of builders being shown.
* If a tag is optional, still show all other tags.
I suggested to ask you for opinion, but I still felt that some basic (but working) functionality would be more helpful than trying and failing to cover all the possible complex scenarios?

www/waterfall_view/src/module/waterfall.tpl.jade
mojca: I would like to be able to see just a single port in the waterfall view. But a prerequisite for that is that the tag with port name is hidden by default, else it would be impossible to find it anyway (even better if also searchable with some auto-complete functionality).

www/waterfall_view/src/module/main.module.js
mojca: Both `and` and `or` are valid requests and there are use cases for both. Say that you want to see both builders on Mac OS X 10.6 building wine (32-bit and 64-bit). Then `or` is not all that helpful, you need `+wine +10.6` (`'wine' and '10.6'`); `('wine' or '10.6')` would be pretty useless.

On the other hand you may want to see one port building on 10.8 and 10.9. So you would need something like `+wine 10.8 10.9` (`'wine' and ('10.8' or '10.9')`).

I don't want to argue too much for either option or either complexity of implementation. I leave it up to developers to say what they want to see implemented. Even what's there now is more helpful than no tags (the only confusing thing is that you currently all to easily end up with an empty set of builders after clicking two tags).

rajdeepbharati: Yes that would be really cool. I will implement it if @tardyp is fine with this.

tardyp:
I think for the filtering, we should use the multiselect.

We actually already have a widget for that in the worker action page:

https://github.com/buildbot/buildbot/blob/7658c965b3bedb3e5532446e1352f62fe7c102ab/www/base/src/app/workers/workeractions.tpl.jade#L7-L8

rajdeepbharati:
> I think for the filtering, we should use the multiselect.

Okay, and should we have the tags as ng-model or then builders themselves?

tardyp:
> Okay, and should we have the tags as ng-model or then builders themselves?
>
tags

Issue #4903, opened 2019-07-21
UnboundLocalError in /api/v2/changes/:n/builds


rajdeepbharati:
```
{"error": "UnboundLocalError(\"local variable 'filters' referenced before assignment\")"}
```

PR #4904, opened 2019-07-21
Handle exception in BuildsEndpoint

rajdeepbharati:
Check whether `filters` variable is defined.

Fixes https://github.com/buildbot/buildbot/issues/4903

### DIFF
--- a/master/buildbot/data/builds.py
+++ b/master/buildbot/data/builds.py
@@ -170,12 +170,15 @@ class BuildsEndpoint(Db2DataMixin, base.BuildNestingMixin, base.Endpoint):
         for b in builds:
             data = yield self.db2data(b)
             # Avoid to request DB for Build's properties if not specified
-            if filters:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


master/buildbot/data/builds.py
```diff
@@ -170,12 +170,17 @@ def get(self, resultSpec, kwargs):
         for b in builds:
             data = yield self.db2data(b)
             # Avoid to request DB for Build's properties if not specified
-            if filters:  # pragma: no cover
-                props = yield self.master.db.builds.getBuildProperties(b['id'])
-                filtered_properties = self._generate_filtered_properties(
-                    props, filters)
-                if filtered_properties:
-                    data['properties'] = filtered_properties
+            try:
```
tardyp: hell! no!

filters don't belong to the else: part. You may want to filter by properties even if you already filter by change (why not?)

rajdeepbharati:
@tardyp I guess this can be merged now

Issue #4880, opened 2019-07-05
images not displayed (icon, favicon, etc.)


rajdeepbharati:
Also, the dev environment isn't working properly in `www/base`

PR #4917, opened 2019-07-24
Configure webpack to load images.

rajdeepbharati:
Import img files in `app.module.js` and configure webpack.

Fixes https://github.com/buildbot/buildbot/issues/4880

### DIFF
--- a/www/base/src/app/app.module.js
+++ b/www/base/src/app/app.module.js
@@ -90,4 +90,8 @@ require('./workers/worker/worker.route.js');
 require('./workers/workeraction.dialog.js');
 require('./workers/workers.controller.js');
 require('./workers/workers.route.js');
-
+require('../img/favicon.ico');
+require('../img/icon.png');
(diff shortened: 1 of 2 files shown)


p12tic:
Thanks for figuring this issue out. I think that instead of requiring the images from javascript side, we should do that where the image is actually used, i.e. the jade templates. E.g. pug loader docs have this example: `img(src=require("./my/image.png"))`

tardyp:
advantage of .jade require is that the file name can be changed to <sha>.png so that cache is ignored on change.
This is arguably barely the case for our usage, but this he a thing to know for a young developer

rajdeepbharati:
I'm not able to use `require` inside my .jade files. I looked up issues on pug-loader and file-loader on their repos and SO but none of the fixes seem to work for me. 😢

p12tic:
Indeed, looks like this does not work. See https://github.com/pugjs/pug/issues/2604.

PR #4922, opened 2019-07-29
Support recursive matching MultipleFileUpload with `glob=True`

angeloudy:
This will support `**` in python `glob.glob`.

By default `recursive` is set to `False`. Setting it to `True` wouldn't hurt anything.
The only disadvantage introduced by recursive matching is performance downgrade, which can be avoided by not using `**`.

### DIFF
--- a/worker/buildbot_worker/commands/fs.py
+++ b/worker/buildbot_worker/commands/fs.py
@@ -235,7 +235,11 @@ class GlobPath(base.Command):
         pathname = os.path.join(self.builder.basedir, self.args['path'])
 
         try:
-            files = glob.glob(pathname)
+            # recursive matching is only support in python3.5+
+            if sys.version_info[:2] >= (3, 5):
+                files = glob.glob(pathname, recursive=True)
+            else:
+                files = glob.glob(pathname)
             self.sendStatus({'files': files})
             self.sendStatus({'rc': 0})
         except OSError as e:
--- a/master/buildbot/newsfragments/support_recursive_glob_MultipleFileUpload.feature
+++ b/master/buildbot/newsfragments/support_recursive_glob_MultipleFileUpload.feature
@@ -1,0 +1,2 @@
+Support recursive matching ('**') in MultipleFileUpload when `glob=True`.
+This is only supported in python3.5+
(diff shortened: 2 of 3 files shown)


tardyp:
can you please create a newsfragment, and be clear that worker needs to be updated

tardyp:
Almost done!

looks like recursive is not supported by python 2.7.

Can you please add a test for 2.7, and document that glob supports recursive only if the worker is on python3?

tardyp:
Can you please look why unit tests are failing?

Also eventually, please squash your commit to cleanup history.
thanks

worker/buildbot_worker/commands/fs.py
```diff
@@ -235,7 +235,11 @@ def start(self):
         pathname = os.path.join(self.builder.basedir, self.args['path'])
 
         try:
-            files = glob.glob(pathname)
+            # recursive matching is only support in python3.5+
+            if sys.version_info[:2] >= (3, 5):
+                files = glob.glob(pathname, recursive=True)
```
sa2ajj: just saying (i'm not sure if it's easier to read): `files = glob.glob(pathname, recursive=True if sys.version_info[:2] >= (3, 5) else False)`

angeloudy: The problem with your code is that python2 does not support `recursive` at all, you can't just pass `false` to it.

sa2ajj: right, this is the build worker code, sorry

tardyp:
Could you please squash your commits?

angeloudy:
@tardyp Sorry, I don't use git quite often. What's squash? what command I use to do it?

tardyp:
Hi squashing is delicate but important skill to have when contributing to open source.

https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

angeloudy:
I tried the `rebase` command. There are lots of commits that are not mine. Are they from updating branch?

angeloudy:
@tardyp I've squashed the commits. Can you merge it now?

angeloudy:
@tardyp Can you help get this one merged?

angeloudy:
Can someone merge this?

Issue #3926, opened 2018-01-28
Buildbot should have a webpage to display all builds for a revision


aj062:
One of the primary purpose of Buildbot is to build and test a "commit" to the repository. Developers often want to know whether their commit build and test successfully  (and did not break anything). However, there is no easy way to check build status for a specific commit.

Buildbot does provide console view which displays this information for recent few commits. However, there are multiple reasons because of which a commit might not show up in console view immediately (e.g.: lot of commits to repository, builds start after a long time after commit due to large number of pending build-requests, console view fetching limited number of recent builds). So, it is not guaranteed that the console view will show builds immediately after the commit. Also, if the developer checks console view after long time, it might be showing build for other recent commits. 

So, if a developer want to check builds for his commit, there is no webpage in Buildbot UI which shows this information for sure (console view may or may not show it). There should be some webpage in Buildbot UI where developer can enter a commit-id (revision) and see all corresponding builds.

It can be implemented in multiple ways, some ideas:
- new page having option to enter revision and see corresponding builds
- new page whose url contains revision, e.g.: http://nine.buildbot.net/#/changes/revision-id
- console view having a search button for revision

tardyp:
This is a nice idea. It will probably need some more data API / db API in order to allow to search builds by change/sourcestamp

uglycoyote:
I would like to see this as well.   We are a game development studio which employs a significant number of artists and other less-technical people who commit content of various kinds which can break the build.   Some of them find it difficult to look at the console page and understand whether or not they have broken the build.

The basic rule is "if every builder has had a green build after your change, then your change is cleared", but even when this rule is understood, it is not always easy to see.

One issue with the stock ```console_view``` is that it doesn't show all changes, only showing changes which have builds associated with them (due to the bit of code below)

            for builder in change.builders
                if builder.builds.length > 0
                    @filtered_changes.push(change)

[...]

PR #4927, opened 2019-07-30
Add webpage to display all builds for a revision

rajdeepbharati:
Closes https://github.com/buildbot/buildbot/issues/3926

- [x] Allow bookmarking individual changes
Use $location to bookmark changes url.
- [x] Display builds for each change

### DIFF
--- a/www/base/src/app/changes/changebuilds/changebuilds.controller.js
+++ b/www/base/src/app/changes/changebuilds/changebuilds.controller.js
@@ -1,0 +1,40 @@
+class ChangeBuildsController {
+    constructor($scope, dataService, bbSettingsService, $stateParams, resultsService, $interval, restService) {
+
+        _.mixin($scope, resultsService);
+        $scope.settings = bbSettingsService.getSettingsGroup('ChangeBuilds');
+        $scope.$watch('settings', () => bbSettingsService.save()
+        , true);
+        const buildsFetchLimit = $scope.settings.buildsFetchLimit.value;
+
+        const dataAccessor = dataService.open().closeOnDestroy($scope);
+        $scope.builders = dataAccessor.getBuilders();
+
+        const changeId = $scope.changeId = $stateParams.changeid;
+
+        dataAccessor.getChanges(changeId).onNew = function(change) {
+            $scope.change = change;
+        }
+
+        const getBuildsData = function() {
+            let requestUrl = `changes/${changeId}/builds`;
+            if (!buildsFetchLimit == '') {
+                requestUrl = `changes/${changeId}/builds?limit=${buildsFetchLimit}`;
+            }
+            restService.get(requestUrl).then((data) => {
+                $scope.builds = data.builds;
+            });
+        }
+
+        getBuildsData();
+
+        const stop = $interval(() => {
+            getBuildsData();
+        }, 5000);
+
+        $scope.$on('$destroy', () => $interval.cancel(stop));
+    }
+}
+
+angular.module('app')
+.controller('changebuildsController', ['$scope', 'dataService', 'bbSettingsService', '$stateParams', 'resultsService', '$interval', 'restService', ChangeBuildsController]);
--- a/www/base/src/app/changes/changebuilds/changebuilds.tpl.jade
+++ b/www/base/src/app/changes/changebuilds/changebuilds.tpl.jade
@@ -1,0 +1,5 @@
+.container
+  .row
+    changedetails(change="change")
+    div(ng-if="builds")
+        builds-table(builds="builds", builders="builders", ng-if="builds")
--- a/www/base/src/app/common/directives/changelist/changelist.tpl.jade
+++ b/www/base/src/app/common/directives/changelist/changelist.tpl.jade
@@ -15,4 +15,6 @@
     .row
         ul.list-group
             li.list-group-item(ng-repeat="change in changes")
+                a(ui-sref="changebuilds({changeid: change.changeid})")
+                    | See builds
                 changedetails(change="change")
--- a/www/base/src/app/app.module.js
+++ b/www/base/src/app/app.module.js
@@ -50,6 +50,8 @@ require('./builders/step/step.controller.js');
 require('./builders/step/step.route.js');
 require('./buildrequests/pendingbuildrequests.controller.js');
 require('./buildrequests/pendingbuildrequests.route.js');
+require('./changes/changebuilds/changebuilds.controller.js');
+require('./changes/changebuilds/changebuilds.route.js');
 require('./changes/changes.controller.js');
 require('./changes/changes.route.js');
 require('./common/common.constant.js');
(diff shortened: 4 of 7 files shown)


www/base/src/app/common/directives/changelist/changelist.tpl.jade
```diff
@@ -16,3 +16,4 @@
         ul.list-group
             li.list-group-item(ng-repeat="change in changes")
                 changedetails(change="change")
+                div {{expandChange()}}
```
tardyp: what is the purpose of this line?

rajdeepbharati: We should ideally be calling this function from the directive, but when I do so, then ALL of the changes are getting expanded.

rajdeepbharati:
@tardyp I had another doubt:
When I use dataservice:
`let mydata = dataService.get(`changes/${changeid}/builds`, {limit:30});`
I am getting an object that contains builds and other items (but I want an array of builds, so that I can iterate).
This is what I'm getting:
```
[listener: ƒ, restPath: "changes/1/builds", query: {…}, accessor: undefined, socketPath: "changes/1/builds/*/*", …]
0: BuildInstance {_endpoint: "builds", $accessor: undefined, builderid: 3, buildid: 6, buildrequestid: 6, …}
1: BuildInstance {_endpoint: "builds", $accessor: undefined, builderid: 6, buildid: 7, buildrequestid: 7, …}
$resolved: true
WrapperClass: class BuildInstance
accessor: undefined
endpoint: "changes/1/builds"
id: "buildid"
listener: ƒ ()
onChange: ƒ noop()
onNew: ƒ noop()
onUpdate: ƒ noop()
query: {limit: 30}
queryExecutor: DataQueryClass {query: {…}, filters: {…}}
restPath: "changes/1/builds"
socketPath: "changes/1/builds/*/*"
socketPathRE: /^changes/1/builds/[^/]+/[^/]+$/
type: "builds"
_byId: {6: BuildInstance, 7: BuildInstance}
_new: []
_updated: []
constructor: (...)
length: 2
__proto__: Array
```

www/base/src/app/common/directives/changelist/changelist.tpl.jade
tardyp: I think your problem is a race between locationChangeSuccess and change loaded.
You need to add a $scope.$watch in order to watch for change list updates

rajdeepbharati: Thank you, I was able to fix the issue using a watcher :)

rajdeepbharati:
@tardyp could you review it now

www/base/src/app/common/directives/changedetails/changedetails.directive.js
```diff
@@ -8,10 +8,82 @@ class Changedetails {
                 compact: '=?'
             },
             template: require('./changedetails.tpl.jade'),
+            controller: '_changeDetailsController'
         };
     }
 }
 
+class _changeDetails {
```
tardyp: This is a Directive. It is supposed to be reusable, and I think it is reused in the build's change tab.

So here you are adding some location logic to a directive. This is not a good practice.
I think you should rather make a proper page for change detail. You would click on the change, and this would go to that page, and provide the details as the list of builds associated with the change.
This is simular to the builders page, which display the list of builders, and then have their own page for more details.

Please refactor this code with this approach.

rajdeepbharati: Should I also have a new page for changeslist instead of the directive?

rajdeepbharati: Can I simply have a href from changeslist directive to a new changes page?

tardyp: There is already a page for changelist, which uses this directive. That is fine.
If you change the directive, then the other pages which show a change list will benefit your improvement (i.e. link to the change detail page), which is the point of a directive compared to implementing everything in a single page.

You can use ui-sref directive, which automatically generate the right href for a state change. Please look at builders page as I suggested.

https://github.com/buildbot/buildbot/blob/7658c965b3bedb3e5532446e1352f62fe7c102ab/www/base/src/app/builders/builders.tpl.jade#L42

rajdeepbharati: Thanks

rajdeepbharati: Thanks

rajdeepbharati:
@tardyp @p12tic I am facing trouble due to this issue: https://github.com/buildbot/buildbot/issues/4940

rajdeepbharati:
ping: @tardyp

www/base/src/app/changes/changebuilds/changebuilds.controller.js
```diff
@@ -0,0 +1,38 @@
...
+
+        const stop = $interval(() => {
+            let requestUrl = `changes/${changeId}/builds`;
+            if (!buildsFetchLimit == '') {
+                requestUrl = `changes/${changeId}/builds?limit=${buildsFetchLimit}`;
+            }
+            restService.get(requestUrl).then((data) => {
+                $scope.builds = data.builds;
+            });
+        }, 5000);
```
tardyp: I think with this, you only get first update after 5s.
Lets factorize the update function, and call it once at init.

www/base/src/app/changes/changebuilds/changebuilds.tpl.jade
```diff
@@ -0,0 +1,4 @@
+.container
+  .row
+    div(ng-if="builds")
```
tardyp: probably we also want the change details in this page.

www/data_module/src/classes/change.service.js
```diff
@@ -8,7 +8,11 @@ class Change {
         let ChangeInstance;
         return (ChangeInstance = class ChangeInstance extends Base {
             constructor(object, endpoint) {
-                super(object, endpoint);
+                const endpoints = [
```
tardyp: I guess this is not needed anymore, and confusing because of #4940

rajdeepbharati:
Made the changes @tardyp

mojca:
I somehow cannot manage to see the builds.

* When clicking on "See builds" it sometimes takes, say, 4 seconds during which time "invalid time" error message is being shown (I thought I found a bug after clicking to the "wrong" commit, but it was apparently just a matter of time).
* The `See builds` is a bit heavy compared to other text. Maybe putting it to the same line as commit message would make more sense, but I'm not sure. It could be nice to have it clickable somewhere on the page next to committed files.
* When `See builds` is clicked, the resulting page remains "closed" (folded). When Just one changeset is being shown, I would say that it makes sense to have everything unfolded / open by default. (See example URL in the next point.)
* I seem to be unable to see builds on a page like https://buildbot.buildbot.net/#/changes/2710 (both locally and on your server).

rajdeepbharati:
> Most important point: I seem to be unable to see builds on a page like https://buildbot.buildbot.net/#/changes/2710 (both locally and on your server), but maybe I'm just not sure how to use it properly.

I think there weren't any builds that's why. And I haven't deployed it to my demo server yet

Issue #4926, opened 2019-07-30
deleteOldLogChunks too slow


angeloudy:
I am using `LogChunksJanitor` to delete old logs.
The job runs once a week.
But every time it took over 3 hours to run, during which period, all the other builder cannot run because the SQL `delete` query locks the database.
The problem is that there are just two many rows in `logs` table. 
Can we delete the logs by chunks, like 1000 rows a time? It will be much faster.

tardyp:
Hello,

Can you tell which database you are using?
How many log per week to you have?

I am not sure how deleting rows by chunk will make it faster.

On the other hand, If you run the janitor everyday, it will obviously be 7 times faster.

angeloudy:
I am using mariadb 10.3.16.

This link explains why it's slow to delete from large tables.

https://dba.stackexchange.com/questions/147103/mysql-delete-becomes-exceptionaly-slow-with-a-large-number-of-rows

I will change it to daily see how fast it can get.

angeloudy:
I just ran my janitor again(4hours since last run) and it took the same amount of time.
So changing it to daily wouldn't help.
I think the problem is I am only deleting logs that are one months old. 
The `logs` table is still huge even with one month's data.
The snowiness could be in the queries to find logs to delete.
It was 3 times slow before I made this change #4893

angeloudy:
The SQL that does the job is
```
DELETE FROM logchunks WHERE logchunks.logid IN (SELECT logs.id  FROM logs  WHERE logs.type = 'd'
```
This query is extremely slow even the `SELECT` subquery returns empty.

There are known issues with sql `DELETE`  with subqueries.
It is recommended to use `JOIN`

If I change the above query to 
```
DELETE t1 FROM logchunks t1 JOIN (SELECT logs.id  FROM logs  WHERE logs.type = 'd'
logs.type = 'd') t2 ON t1.logid = t2.id
```
, it returns immediately.

tardyp:
Hi.
Your thinking makes sense

The code is here"

https://github.com/buildbot/buildbot/blob/7658c965b3bedb3e5532446e1352f62fe7c102ab/master/buildbot/db/logs.py#L395
Would you like to expirement and change the ``in`` into a ``join``, and report your findings?

angeloudy:
I don't know how to write the `join` query. But I did the following changes and now my builder only takes 30 seconds to run, which is more than 100 times faster.
```
[...]

PR #4933, opened 2019-08-02
improve log cleaning performance

angeloudy:
Fixing issue #4926

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -385,15 +385,21 @@ class LogsConnectorComponent(base.DBConnectorComponent):
                 res.close()
 
             # query all logs with type 'd' and delete their chunks.
-            q = sa.select([model.logs.c.id])
-            q = q.select_from(model.logs)
-            q = q.where(model.logs.c.type == 'd')
+            if self.db._engine.dialect.name == 'sqlite':
+                # sqlite does not support delete with a join, so for this case we use a subquery,
+                # which is much slower
+                q = sa.select([model.logs.c.id])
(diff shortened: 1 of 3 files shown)


tardyp:
looks like your approach does not work with sqlite.
Please see error in the CI:
```
builtins.NotImplementedError: This backend does not support multiple-table criteria within DELETE
```

p12tic:
@angeloudy Could you check if `self.db._engine.dialect.name == 'sqlite'` holds and use the old code path in that case?

angeloudy:
@p12tic 
That would be a bit ugly. I am sure there's a way to write the query that works for both.

angeloudy:
I was wrong. Looks like SQLite doesn't support join delete at all.

p12tic:
The unit tests still fail. You can run them locally, the Buildbot documentation has information on how to do that http://docs.buildbot.net/latest/developer/quickstart.html.

angeloudy:
The error looks unrelated to my change. https://buildbot.buildbot.net/#/builders/5/builds/2258

p12tic:
Interesting, I will look into this. Most of the time failures like that turn out to be unexpected side effect of the PR.

angeloudy:
@p12tic  I've reverted my previous change to reveal the original error.
After adding "if self.db._engine.dialect.name == 'sqlite'" as you suggested, I got this error
```
builtins.AttributeError: 'FakeDBConnector' object has no attribute '_engine'
```

p12tic:
@angeloudy Turns out it was a different `FakeDBConnector` that needed modification. Thanks for your PR, the tests now pass.

PR #4943, opened 2019-08-05
Changed files support for GerritChangeSource and GerritEventLogPoller

### LINKED ISSUE #3996: No "Changed files" in Console view while using GerritChangeSource
We have 2 instances of Buildbot. First instance gets changes from Github (using GitPoller) and the second gets it from Gerrit (using GerritChangeSource).
In case of 1st instance, which uses Github, we have filled "changed files" section. Example:
image
It is ok.

In case of 2nd instance, which uses Gerrit, "changed files" section is empty:
image

We use GerritChangeSource like this:
```
from buildbot.plugins import changes

c['change_source'] = changes.GerritChangeSource(
    "gerrit.example.com",
    "gerrit_user",
    handled_events=["patchset-created", "change-merged"]) 
```

Is it a bug?

All Buildbot instances uses version: 0.9.15post1

gvalkov:
Set the `files` property for changes emitted by the Gerrit change sources. It's disabled by default as the extra traffic to the Gerrit server may come as an unpleasant surprise to some. The feature can be enabled by setting `get_files` to true.

Fixes #3996.

## Contributor Checklist:

**Side-note:**
I last contributed (mostly subversion features and fixes) to buildbot 9 years ago - it's great to see how far the project has come since then! I also can't believe the project is still using some of the [figures](4339e66d11f6f547d9711e14600c86938f1b988d) I made for the documentation back then :)

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -298,17 +298,30 @@ class GerritChangeSource(GerritChangeSourceBase):
             self.startStreamProcess()
             self.streamProcessTimeout = self.STREAM_BACKOFF_MIN
 
+    def _buildGerritCommand(self, *gerrit_args):
+        '''Get an ssh command list which invokes gerrit with the given args on the
+        remote host'''
+
+        cmd = [
+            "ssh",
+            "%s@%s" % (self.username, self.gerritserver),
+            "-p", str(self.gerritport)
+        ]
+
+        if self.identity_file is not None:
+            cmd.extend(["-i", self.identity_file])
+
+        cmd.append("gerrit")
+        cmd.extend(gerrit_args)
+        return cmd
+
     def startStreamProcess(self):
         if self.debug:
             log.msg("starting 'gerrit stream-events'")
+
+        cmd = self._buildGerritCommand("stream-events")
         self.lastStreamProcessStart = util.now()
-        uri = "%s@%s" % (self.username, self.gerritserver)
-        args = [uri, "-p", str(self.gerritport)]
-        if self.identity_file is not None:
-            args = args + ['-i', self.identity_file]
-        self.process = reactor.spawnProcess(
-            self.LocalPP(self), "ssh",
-            ["ssh"] + args + ["gerrit", "stream-events"], env=None)
+        self.process = reactor.spawnProcess(self.LocalPP(self), "ssh", cmd, env=None)
 
     def activate(self):
         self.wantProcess = True
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -18,6 +18,7 @@ import json
 
 from twisted.internet import defer
 from twisted.internet import reactor
+from twisted.internet import utils
 from twisted.internet.protocol import ProcessProtocol
 from twisted.python import log
 
@@ -74,7 +75,8 @@ class GerritChangeSourceBase(base.ChangeSource):
     def checkConfig(self,
                     gitBaseURL=None,
                     handled_events=("patchset-created", "ref-updated"),
-                    debug=False):
+                    debug=False,
+                    get_files=False):
 
         if gitBaseURL is None:
             config.error("gitBaseURL must be specified")
@@ -82,9 +84,11 @@ class GerritChangeSourceBase(base.ChangeSource):
     def reconfigService(self,
                         gitBaseURL=None,
                         handled_events=("patchset-created", "ref-updated"),
-                        debug=False):
+                        debug=False,
+                        get_files=False):
         self.gitBaseURL = gitBaseURL
         self.handled_events = list(handled_events)
+        self._get_files = get_files
         self.debug = debug
 
     def lineReceived(self, line):
@@ -154,11 +158,19 @@ class GerritChangeSourceBase(base.ChangeSource):
                               event_change['number'])
         return event_change["branch"]
 
+    @defer.inlineCallbacks
     def addChangeFromEvent(self, properties, event):
-
         if "change" in event and "patchSet" in event:
             event_change = event["change"]
-            return self.addChange({
+
+            files = ["unknown"]
+            if self._get_files:
+                files = yield self.getFiles(
+                    change=event_change["number"],
+                    patchset=event["patchSet"]["number"]
+                )
+
+            yield self.addChange({
                 'author': _gerrit_user_to_author(event_change["owner"]),
                 'project': util.bytes2unicode(event_change["project"]),
                 'repository': "{}/{}".format(
@@ -167,7 +179,7 @@ class GerritChangeSourceBase(base.ChangeSource):
                 'revision': event["patchSet"]["revision"],
                 'revlink': event_change["url"],
                 'comments': event_change["subject"],
-                'files': ["unknown"],
+                'files': files,
                 'category': event["type"],
                 'properties': properties})
 
@@ -322,6 +334,25 @@ class GerritChangeSource(GerritChangeSourceBase):
         cmd = self._buildGerritCommand("stream-events")
         self.lastStreamProcessStart = util.now()
         self.process = reactor.spawnProcess(self.LocalPP(self), "ssh", cmd, env=None)
+
+    @defer.inlineCallbacks
+    def getFiles(self, change, patchset):
+        cmd = self._buildGerritCommand("query", str(change), "--format", "JSON",
+                                       "--files", "--patch-sets")
+
+        if self.debug:
+            log.msg("querying gerrit for changed files in change %s/%s: %s" %
+                    (change, patchset, cmd))
+
+        out = yield utils.getProcessOutput(cmd[0], cmd[1:], env=None)
+        out = out.splitlines()[0]
+        res = json.loads(bytes2unicode(out))
+
+        if res.get("rowCount") == 0:
+            return ["unknown"]
+
+        patchsets = {i["number"]: i["files"] for i in res["patchSets"]}
+        return [i["file"] for i in patchsets[int(patchset)]]
 
     def activate(self):
         self.wantProcess = True
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -431,6 +431,14 @@ class GerritEventLogPoller(GerritChangeSourceBase):
             yield self.master.db.state.setState(self._oid, 'last_event_ts', event['eventCreatedOn'])
         return res
 
+    @defer.inlineCallbacks
+    def getFiles(self, change, patchset):
+        res = yield self._http.get("/changes/%s/revisions/%s/files/" % (change, patchset))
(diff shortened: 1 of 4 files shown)


master/buildbot/changes/gerritchangesource.py
```diff
@@ -74,6 +75,7 @@ class GerritChangeSourceBase(base.ChangeSource):
     def checkConfig(self,
                     gitBaseURL=None,
                     handled_events=("patchset-created", "ref-updated"),
+                    get_files=False,
```
sa2ajj: it's a good idea to add new parameters *after* the existing ones: this method signature says that it accepts these parameters as keyword ones as well as positional ones...  i'd even suggest to add `*` before that to ensure that in future this parameter is always a keyword one.

master/buildbot/changes/gerritchangesource.py
```diff
@@ -82,33 +84,37 @@ def checkConfig(self,
     def reconfigService(self,
                         gitBaseURL=None,
                         handled_events=("patchset-created", "ref-updated"),
+                        get_files=False,
```
sa2ajj: ditto

master/buildbot/changes/gerritchangesource.py
```diff
@@ -82,33 +84,37 @@ def checkConfig(self,
...
                         debug=False):
         self.gitBaseURL = gitBaseURL
         self.handled_events = list(handled_events)
+        self.get_files = get_files
```
sa2ajj: while we have plenty of "let's have it public data members", i do not think it's a good idea to introduce new ones: please use `self._get_files`

master/buildbot/changes/gerritchangesource.py
```diff
@@ -82,33 +84,37 @@ def checkConfig(self,
...
         self.debug = debug
 
+    @defer.inlineCallbacks
```
sa2ajj: this change (along with the relevant bits) should belong to a separate commit with explanation what it achieves

master/buildbot/changes/gerritchangesource.py
```diff
@@ -82,33 +84,37 @@ def checkConfig(self,
...
             if self.debug:
                 msg = "no type in event %s"
                 log.msg(msg % line)
-            return defer.succeed(None)
+            return
 
-        return self.eventReceived(event)
+        yield self.eventReceived(event)
 
+    @defer.inlineCallbacks
```
sa2ajj: ditto

master/buildbot/changes/gerritchangesource.py
```diff
@@ -298,17 +298,30 @@ def streamProcessStopped(self):
...
+        remote host'''
+
+        cmd = [
+            "ssh",
+            "%s@%s" % (self.username, self.gerritserver),
+            "-p", str(self.gerritport)
+        ]
+
+        if self.identity_file is not None:
+            cmd += ["-i", self.identity_file]
```
sa2ajj: you probably want to use `cmd.extend(...)`

master/buildbot/changes/gerritchangesource.py
```diff
@@ -298,17 +312,49 @@ def streamProcessStopped(self):
...
+
+        if self.debug:
+            log.msg("querying gerrit for changed files in change %s/%s: %s" %
+                    (change, patchset, cmd))
+
+        out = yield utils.getProcessOutput(cmd[0], cmd[1:], env=None)
+        out = out.splitlines()[0]
+        res = json.loads(bytes2unicode(out))
+
+        if res.get("rowCount") == 0:
```
sa2ajj:
```
if not res.get(...):
```

unless, of course, you want to explicitly catch the case when `rowCount` is present and is equal to 0.

master/buildbot/changes/gerritchangesource.py
```diff
@@ -387,6 +433,14 @@ def eventReceived(self, event):
             yield self.master.db.state.setState(self._oid, 'last_event_ts', event['eventCreatedOn'])
         return res
 
+    @defer.inlineCallbacks
+    def getFiles(self, change, patchset):
+        res = yield self._http.get("/changes/%s/revisions/%s/files/" % (change, patchset))
+        res = yield res.content()
+
+        files = json.loads(bytes2unicode(res.splitlines()[1]))
```
sa2ajj: do you need this variable here?  why not just `return list(...)`?

master/buildbot/changes/gerritchangesource.py
```diff
@@ -298,17 +312,49 @@ def streamProcessStopped(self):
             self.startStreamProcess()
             self.streamProcessTimeout = self.STREAM_BACKOFF_MIN
 
+    def buildGerritCommand(self, *gerrit_args):
```
sa2ajj: is this a public method?  if not. use `_buildGerritCommand`

master/buildbot/changes/gerritchangesource.py
```diff
@@ -298,17 +312,49 @@ def streamProcessStopped(self):
...
+    def getFiles(self, change, patchset):
+        cmd = self.buildGerritCommand("query", str(change), "--format", "JSON",
+                                      "--files", "--patch-sets")
+
+        if self.debug:
+            log.msg("querying gerrit for changed files in change %s/%s: %s" %
+                    (change, patchset, cmd))
+
+        out = yield utils.getProcessOutput(cmd[0], cmd[1:], env=None)
+        out = out.splitlines()[0]
```
sa2ajj: you probably could combine this and the next line

master/buildbot/changes/gerritchangesource.py
gvalkov: Agreed - will fix.

master/buildbot/changes/gerritchangesource.py
gvalkov: Roger that. I was trying to be consistent with the surrounding code (which is a bit of a challenge given that the buildbot sources feature every possible formatting and naming convention out there :)

master/buildbot/changes/gerritchangesource.py
gvalkov: The commits as they are are atomic - I wouldn't be able to split them any further. It's all about making `lineReceived()` preemptable so that `getFiles()` can be called asynchronously. The other option would be to run the query or send the GET request in a way that would block the reactor. I'd be happy to learn if this can be done in another way.

master/buildbot/changes/gerritchangesource.py
gvalkov: Done, though they're equivalent.

master/buildbot/changes/gerritchangesource.py
gvalkov: That is unlikely. If the change exists `gerrit query` will always return two lines of json - one with the result and one with `{"type": "stats", "rowCount": 1, ...}`. If it didn'tt exist then the response would be a single `{"type": "stats", "rowCount": 0}`. I think it's sufficient.

master/buildbot/changes/gerritchangesource.py
gvalkov: Personal style I guess. I often add breakpoints between lines in which case it helps to have a variable ready for inspection. There's a bit too much happening in a line like `list(json.loads(bytes2unicode(res.splitlines()[1])))` - I'll break it down further.

master/buildbot/changes/gerritchangesource.py
sa2ajj: they are indeed, though the new code is consistent with what follows

master/buildbot/changes/gerritchangesource.py
sa2ajj: i am not sure i understand, could you please elaborate?

master/buildbot/changes/gerritchangesource.py
sa2ajj: i'm sorry what is unlikely?

master/buildbot/changes/gerritchangesource.py
gvalkov: Before this change, once a line from `gerrit stream-events` was received by `LocalPP.outReceived` it was passed to `GerritChangeSourceBase.lineReceived` and execution proceeded sequentially/synchronously all the way to `addChangeFromEvent`. 

However, we now want to perform a costly blocking operation (i.e `getFiles`) in the middle of all this. By making `addChangeFromEvent` an async function, we allow the buildbot master to do something else useful while `getFiles()` is waiting for its result to come back (I've seen cold queries for changes with many patchsets take up to half a second to return). For this to work, we have to make the whole call chain up to `lineReceived` async for this to work - i.e. `addChangeFromEvent, eventReceived, lineReceived` must all return deferreds.

master/buildbot/changes/gerritchangesource.py
gvalkov: It's just that if the first line of output is `{"type": "stats", "rowCount": 0}` then no results were found, i.e. the change doesn't exist, which is unlikely given that we're reacting to an event about an existing change.

master/buildbot/changes/gerritchangesource.py
gvalkov: Done

master/buildbot/changes/gerritchangesource.py
tardyp: It is not really needed to turn this code into inlineCallbacks. The code is already asynchronous as it return defer.succeed, and returns the  value of self.eventReceived (which could be a deferred, generated by an inlineCallback generator)

master/buildbot/changes/gerritchangesource.py
```diff
@@ -387,6 +433,14 @@ def eventReceived(self, event):
             yield self.master.db.state.setState(self._oid, 'last_event_ts', event['eventCreatedOn'])
         return res
 
+    @defer.inlineCallbacks
+    def getFiles(self, change, patchset):
+        res = yield self._http.get("/changes/%s/revisions/%s/files/" % (change, patchset))
+        res = yield res.content()
+
+        res = bytes2unicode(res.splitlines()[1])  # the first line of every response is `)]}'`
```
tardyp: you can just call .decode() now that we don't have to deal with python2.6

gvalkov: Good news. I would have removed it in the other places where it's called, but several of the tests  depend on the "passthrough" behaviour of `bytes2unicode`. Can fix in a follow up PR.

master/buildbot/changes/gerritchangesource.py
gvalkov: You're right. I initially got into a big mess that I could resolve only by using `inlineCallbacks` on everything down the call chain. It should be simpler now (i.e. `inlineCallback` only on `addChangeFromEvent` since that calls `getFiles`).

gvalkov:
Hi @tardyp . I've simplified the change (no unneeded inlineCallbacks-ification). More over I've been running the change cherry-picked on top of the 2.3.1 release in our production buildbot instance since I opened the PR. Is there anything else I can do to help with merging this change?

Issue #4945, opened 2019-08-06
BitBucket Server Push event uses different event key


andrewspiers:
The method `process` of BitbucketServerEventHandler in master/buildbot/www/hooks/bitbucketserver.py uses the value of the `X-Event-Key` header to call a handler method depending on  the value of the key.

For the push event, bitbucket server uses a value of `repo:refs_changed`.
see https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html
This value seems to be consistent across bitbucket server versions.

I think the BitbucketServerEventHandler method handle_repo_push should be renamed to handle_repo_refs_changed so it is triggered on push events.

PR #4947, opened 2019-08-06
bitbucket server push event handler

andrewspiers:
This is for issue https://github.com/buildbot/buildbot/issues/4945

The Bitbucket server push event handler
uses the x-event-key value 'repo:refs_changed' to signify
push events.

bitbucket server: update header event

replace header event repo:push with repo:refs_changed

https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html


tardyp:
Looks like a very old change, as I can't find in documentation history when it used push event. Probably from the time where the product was called Stash.

PR #4949, opened 2019-08-07
Convert the buildbot.www.auth module into a package

gvalkov:
I've began work on an `LdapAuth` authentication provider. It's based on [ldapauth.py](https://github.com/nextgis/buildbot/blob/master/ldapauth.py), but makes use of the already existing `LdapUserInfo`. This feature can be implemented more cleanly if the `buildbot.www.auth` module was split into a package (similar to how `authz` is structured). 

The change moves a lot of code around, so I'd like to know if it's a go/no-go before I continue working on the authentication provider.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/auth/__init__.py
+++ b/master/buildbot/www/auth/__init__.py
@@ -1,0 +1,29 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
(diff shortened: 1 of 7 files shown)


master/buildbot/www/auth/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+
+from buildbot.www.auth.base import *
```
tardyp: import * is forbidden in guidelines. please use explicit import

gvalkov: Done. I moved the `__all__` of each module into the `__init__`.

PR #4950, opened 2019-08-07
PULL_REQUEST_TEMPLATE: add a note about usage of infinitive instead of past sense

sa2ajj:
## Contributor Checklist:

* [x] I have updated the unit tests -- *no need*
* [x] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory) -- *no need*
* [x] I have updated the appropriate documentation -- *no need*

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,8 +9,13 @@ http://docs.buildbot.net/latest/developer/quickstart.html
 And especially:
 http://docs.buildbot.net/latest/developer/pull-request.html
 
-
 ## Contributor Checklist:
+
+*Important*: please use infinitive when describing changes- not "Added" but "Add": you are submitting a request to change, you have not change anything yet (in your branch, yes, in the target branch, no).
+
+Related to review comments:
+
+* `nitpick` means that it _might_ be a nice change, but it's up to the submitter.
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,9 +9,14 @@ http://docs.buildbot.net/latest/developer/quickstart.html
 And especially:
 http://docs.buildbot.net/latest/developer/pull-request.html
 
-
 ## Contributor Checklist:
 
+*Important*: please use infinitive when describing changes- not "Added" but "Add": you are submitting a request to change, you have not change anything yet (in your branch, yes, in the target branch, no).
```
p12tic: I think it's best to move this to a separate section above that is to be removed. Usually most PRs leave the checkboxes in the PR description.

sa2ajj: would something like "## Commit Message Guidelines" make sense?

tardyp: We already have a documentation for pull-request, with lot of important guidelines in it.

I'm not saying that this particular rule is less important than the others, but it should imho be managed as part of the others.

sa2ajj: i overlooked that one :(  there's no point of repeating the same rules in different places...

Issue #3995, opened 2018-03-14
Incorrect grouping of builds in console view while using ForceScheduler


adydychk:
In my Buildbot`s configuration I start one "dummy" builder which triggers set of real builders.  
Here  is an example:   
![image](https://user-images.githubusercontent.com/31437274/37397796-93959e08-278d-11e8-9ce1-850356e68d43.png)

My problem is when I use ForceScheduler to start this "dummy" builder and I see that my builds are not grouped (NOT in one row). Here is an example:
![image](https://user-images.githubusercontent.com/31437274/37397840-ab3a702e-278d-11e8-96bf-1990be8e1f71.png)
I expect that these builders should be in one row as in the screenshot number 1.

What can I do to group all these builders as in the first screenshot?

To implement ForceScheduler I used example from this page: https://docs.buildbot.net/current/manual/cfg-schedulers.html#forcescheduler-scheduler
To trigger my real builders from "dummy" trigger I used this trigger: https://docs.buildbot.net/current/manual/cfg-buildsteps.html#triggering-schedulers

My Buildbot version: 0.9.15post1

adydychk:
Any comments on this?  
This is very critical feature for my team.

adydychk:
@tardyp Any comments on this?
This is very critical feature for my team.

In addition This feature request is also critical: https://github.com/buildbot/buildbot/issues/4006 :)

tardyp:
I am not sure exactly what are those revisions 13-1104, etc. which scm are you using?

In order to be grouped, you need to make sure your builds are triggered with a source stamp, or that they get the same got_revision property.

Please understand that buildbot is managed by community. If you have critical feature, don't hesitate to work on it, and we will gladly help on making sure it is merged when it is of good enough quality.

eponomar:
Hi @tardyp, I am from the same team as @adydychk.
We are kindly glad for your answer. I want to clarify some aspects:

13-1104 is not revision. 13 is builderid and 1104 is buildrequest number.
What is scm? 

I checked that all builds have the same source stamp, but I don`t set got_revision property.
How can I set this property for all builds? Via UI form?

tardyp:
got_revision is generated by the source control managment step. It is the backup for console if it cannot find a source stamp change.

Here is the code which is involve. Perhaps it will help you to find why it cannot find the source stamp.

https://github.com/buildbot/buildbot/blob/master/www/console_view/src/module/main.module.coffee#L272

Regard

[...]

PR #4951, opened 2019-08-07
Add a pipeline_id column to buildsets

rajdeepbharati:
# Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_db_migrate_versions_056_add_buildsets_pipeline_id.py
+++ b/master/buildbot/test/unit/test_db_migrate_versions_056_add_buildsets_pipeline_id.py
@@ -1,0 +1,85 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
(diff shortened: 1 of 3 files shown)


sa2ajj:
Could you please add some docs/release notes that'd describe what the goal is?  The PR description nor the code tell much...

rajdeepbharati:
@sa2ajj The end goal is to fix this issue https://github.com/buildbot/buildbot/issues/3995
This is still in progress. I'll update the description and add docs once I finish the work.

tardyp:
I agree with @sa2ajj that this needs to come with very strong architecture design on how this will work eventually.
I am afraid  we are out of time for gsoc2019 to start such large refactor task.

I am not telling that it is a bad idea, but before commiting into db change, we should really be sure this is the right way to go (or we should do it in a different branch)

rajdeepbharati:
@tardyp Should I try having the pipelineid same as buildid, or different?

tardyp:
> @tardyp Should I try having the pipelineid same as buildid, or different?

looks like a good idea

rajdeepbharati:
@tardyp Ok, thanks. But now I'm facing a different issue which seems to be due to the previous migration that made committer not null (055): https://paste.macports.org/c6305c433160

master/buildbot/test/unit/test_db_migrate_versions_056_add_buildsets_pipeline_id.py
```diff
@@ -0,0 +1,93 @@
...
+
+            # q = sa.select([buildsets.c.submitted_at, buildsets.c.pipeline_id])
+            # num_rows = 0
+            # for row in conn.execute(q):
+            #     # verify that the default value was set correctly
+            #     self.assertEqual(row.pipeline_id, 100)
+            #     num_rows += 1
+            # self.assertEqual(num_rows, 1)
+
+        return self.do_test_migration(54, 55, setup_thd, verify_thd)
```
tardyp: you do a migration here from version 54 to version 55. So it runs the 055 migration. You should write:
```suggestion
        return self.do_test_migration(55, 56, setup_thd, verify_thd)
```

rajdeepbharati: Oops. Sorry for doing such a trivial mistake

PR #4952, opened 2019-08-08
Enable autoescape in jinja2

kravietz:
Per [Bandit
findings](https://bandit.readthedocs.io/en/latest/plugins/b701_jinja2_autoescape_false.html).
By default `jinja2` disables autoescaping which enables a number of XSS
vectors.

Also marked several other findings in test scripts as false positives
(`# nosec`).

## Remove this paragraph

And especially:

## Contributor Checklist:


tardyp:
not sure why we want to do autoescaping, as this is targeted at xss on html files, and there, we are generating .tac, and the jinja template is not controlled by attackers. (generated at buildbot create-master)

p12tic:
I agree with @tardyp, the attacker does not control input to jinja, which means autoescaping is not necessary. Please reopen the pull request if we misunderstood. Thank you!

Issue #4885, opened 2019-07-09
virtual builders are shown as old builders


rajdeepbharati:
In the builders page, I am having to enable `show old builders` option to display virtual builders.

PR #4956, opened 2019-08-09
Consider virtual builders as active

rajdeepbharati:
Fixes https://github.com/buildbot/buildbot/issues/4885

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -267,6 +267,8 @@ class Build(properties.PropertiesMixin):
                 tags = self.getProperty(
                     self.VIRTUAL_BUILDERTAGS_PROP,
                     self.builder.config.tags)
+                if '_virtual_' not in tags:
+                    tags.append('_virtual_')
 
                 self.master.data.updates.updateBuilderInfo(self._builderid,
                                                            description,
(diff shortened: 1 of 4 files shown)


www/base/src/app/builders/builders.controller.js
```diff
@@ -28,6 +28,11 @@ class Builders {
                     active = true;
                 }
             }
+            // virtual builders need to be supplied 'virtual' tag
+            // to consider them as active builders
+            if (builder.tags.includes('virtual')) {
```
tardyp: I would reserve that keyword, using '_virtual_'.
Also, we need to make sure '_virtual_' tag is set for virtual builders
https://github.com/buildbot/buildbot/blob/e8b3696a23335236b69e5ebaa7d7976c5d994f2a/master/buildbot/process/build.py#L271

rajdeepbharati: Done

master/buildbot/process/build.py
```diff
@@ -267,6 +267,8 @@ def getBuilderId(self):
                 tags = self.getProperty(
                     self.VIRTUAL_BUILDERTAGS_PROP,
                     self.builder.config.tags)
+                if '_virtual_' not in tags:
```
tardyp: you wll need to update the unit tests for this code

rajdeepbharati: Updated

Issue #4761, opened 2019-05-08
Allow hiding of inactive builders in Waterfall view


rajdeepbharati:
Add an option to hide old builders from waterfall view.

rajdeepbharati:
@tardyp Virtual builders don't have any master (hence no active master). Should I make them have some `builder.masterids`?

rajdeepbharati:
Or we can simply do:
```
if (builder.tags.includes('virtual')) {
  active = true;
}
```
in `hasActiveMaster` function (this wouldn't work unless 'virtual' tag has been assigned to virtual builders.

tardyp:
I like the virtual tag idea

PR #4957, opened 2019-08-09
Allow hiding of inactive builders in Waterfall

rajdeepbharati:
Closes https://github.com/buildbot/buildbot/issues/4761

## Contributor Checklist:

### DIFF
--- a/www/waterfall_view/src/module/main.module.js
+++ b/www/waterfall_view/src/module/main.module.js
@@ -118,6 +118,7 @@ var WaterfallController = (function() {
             this.$scope.builders = (this.builders = this.all_builders);
             this.buildLimit = this.c.limit;
             this.$scope.builds = (this.builds = this.dataAccessor.getBuilds({limit: this.buildLimit, order: '-started_at'}));
+            this.$scope.masters = this.dataAccessor.getMasters();
 
             d3Service.get().then(d3 => {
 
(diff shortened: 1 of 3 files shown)


www/waterfall_view/src/module/main.module.js
```diff
@@ -199,6 +209,23 @@ var WaterfallController = (function() {
...
+            if ((builder.masterids == null)) {
+                return false;
+            }
+            for (let mid of Array.from(builder.masterids)) {
+                const m = this.$scope.masters.get(mid);
+                if ((m != null) && m.active) {
+                    active = true;
+                }
+            }
+            if (builder.tags.includes('virtual')) {
```
tardyp: ditto

tardyp:
Shall we pursue this approach?

rajdeepbharati:
@tardyp Did you have other ideas? IMO this is the easy way to handle it.

tardyp:
It is easy, but confusing. I think we should think of a better approach. I think the best would be to show builders that are visible in the current vertical scroll timeframe.
This makes the algorithm more complex, but this would be an interresting algorithm to work-on.
This concept is used for most web text editors, and the logviewer is also implementing it.

rajdeepbharati:
> show builders that are visible in the current vertical scroll timeframe

Doesn't that already happen? I'm a bit confused here (how are these related?)

Issue #4869, opened 2019-06-27
builders page is slow, need search-ish alternative or a filter


dankegel:
The builders page takes about five seconds to load, and even after it loads, finding the builder in question amongst the thousands of builders is hard.  Users have taken to doing things like
```
#!/bin/sh
# wmb - Dude, where's my builder?
curl -sS http://buildhost5.example.com:8010/api/v2/builders | jq '
[.builders[]
 | select(.name | match($search))
 | {name:.name, link: @uri "http://buildhost5.oblong.com:8010/#/builders/\(.builderid)"}]
 | sort_by(.name)
' --arg search "$1"
```

That lets users quickly find the URLs for the builders of interest... but it'd be nice if the
front page had a search box that did the same thing, so they didn't have to resort to
the commandline.

p12tic:
How many builders do you have? Are you using HTTP2?

dankegel:
About 4000 builders.

I'm not using a proxy, just straight up buildbot and chrome, with no special networking tweaks.

mojca:
This is one of the pages that needs filtering by tags, other than console view (#3470) and waterfall (#3471).

In case virtual builders will be implemented in the scope of @rajdeepbharati's GSOC projects, this will definitely be needed for us as well.

Search field could make sense. @rajdeepbharati implemented a special view for MacPorts, but again, if we end up using 20k+ virtual builders, search field on builders page (and probably some property to allow hiding some virtual builders by default) could be used instead of a special view).

I'm not talking about addressing the performance yet, but at least to start with the way of filtering tags, and then continue from there later.

PR #4959, opened 2019-08-10
Add pagination and search bar to builders page

rajdeepbharati:
Closes https://github.com/buildbot/buildbot/issues/4869

## Contributor Checklist:

### DIFF
--- a/www/base/src/app/builders/builders.tpl.jade
+++ b/www/base/src/app/builders/builders.tpl.jade
@@ -37,7 +37,10 @@
             span(ng-show="tags_filter.length > 0")
                 span.label.clickable.label-danger(ng-click="tags_filter = []") x
           th(style="width:20%px;") Workers
-      tr(ng-repeat='builder in builders | filter: isBuilderFiltered | orderBy: "name"')
+      ul(uib-pagination, total-items="builders.length",
+        ng-model="currentPage", class="pagination-sm", boundary-link-numbers="true",
+        max-size=20, items-per-page="page_size")
+      tr(ng-repeat='builder in builders | limitTo: page_size :(currentPage-1)*page_size | filter: isBuilderFiltered | orderBy: "name"')
           td(style="width:200px")
             a(ui-sref='builder({builder: builder.builderid})')
               | {{ builder.name }}
--- a/www/base/src/app/builders/builders.tpl.jade
+++ b/www/base/src/app/builders/builders.tpl.jade
@@ -1,5 +1,7 @@
 .container
   .row
+    form(role="search" style="width:150px")
+      input(type="text" ng-model="searchQuery" class="form-control" placeholder="Search for builders")
     table.table.table-hover.table-striped.table-condensed
       tr
           th Builder Name
@@ -37,10 +39,10 @@
             span(ng-show="tags_filter.length > 0")
                 span.label.clickable.label-danger(ng-click="tags_filter = []") x
           th(style="width:20%px;") Workers
-      ul(uib-pagination, total-items="builders.length",
+      ul(uib-pagination, total-items="(builders | filter: {name: searchQuery} | filter: isBuilderFiltered).length",
         ng-model="currentPage", class="pagination-sm", boundary-link-numbers="true",
-        max-size=20, items-per-page="page_size")
-      tr(ng-repeat='builder in builders | limitTo: page_size :(currentPage-1)*page_size | filter: isBuilderFiltered | orderBy: "name"')
+        max-size=10, items-per-page="page_size")
+      tr(ng-repeat='builder in builders | filter: {name: searchQuery} | filter: isBuilderFiltered | orderBy: "name" | limitTo: page_size :(currentPage-1)*page_size')
           td(style="width:200px")
             a(ui-sref='builder({builder: builder.builderid})')
               | {{ builder.name }}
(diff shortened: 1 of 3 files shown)


rajdeepbharati:
@tardyp

rajdeepbharati:
@p12tic there's one more problem in this: if we supply a tag to the url and the builder having that tag is not on the first page, then it doesn't get displayed on the first page (we are having to navigate)

mojca:
I would say that this is related to the other observation I had. If all builders are hidden (as is the case with virtual builders, for example), the site still shows 10 pages, even though just a single builder is shown. If we end up with 200k+ (virtual) builders, it's not an option to fetch them all. The filter to show or hide old & virtual builders, as well as the tags, should probably be used/applied to filter the selection before you actually fetch, display & paginate the builder names. If only 10 builders match the selection (old, virtual, consistent with the chosen tags), you should not even show the pagination. The database query should directly request just the number of builders matching these criteria.

rajdeepbharati:
> The filter to show or hide old & virtual builders, as well as the tags, should probably be used/applied to filter the selection before you actually fetch, display & paginate the builder names. If only 10 builders match the selection (old, virtual, consistent with the chosen tags), you should not even show the pagination. The database query should directly request just the number of builders matching these criteria.

That filtering is already done before. And virtual/old builders issue is fixed in https://github.com/buildbot/buildbot/pull/4956. So, it wouldn't show empty pages. And you can set number of builders per page in settings.

rajdeepbharati:
> The filter to show or hide old & virtual builders, as well as the tags, should probably be used/applied to filter the selection before you actually fetch, display & paginate the builder names. If only 10 builders match the selection (old, virtual, consistent with the chosen tags), you should not even show the pagination. The database query should directly request just the number of builders matching these criteria.

That filtering is already done before. And virtual/old builders issue is fixed in https://github.com/buildbot/buildbot/pull/4956. So, it wouldn't show empty pages. And you can set number of builders per page in settings.

www/base/src/app/builders/builders.tpl.jade
```diff
@@ -37,7 +39,10 @@
             span(ng-show="tags_filter.length > 0")
                 span.label.clickable.label-danger(ng-click="tags_filter = []") x
           th(style="width:20%px;") Workers
-      tr(ng-repeat='builder in builders | filter: isBuilderFiltered | orderBy: "name"')
+      ul(uib-pagination, total-items="builders.length",
+        ng-model="currentPage", class="pagination-sm", boundary-link-numbers="true",
+        max-size=10, items-per-page="page_size")
+      tr(ng-repeat='builder in builders | orderBy: "name" | limitTo: page_size :(currentPage-1)*page_size | filter: {name: searchQuery} | filter: isBuilderFiltered')
```
tardyp: put the filters first, and you will see the paging work much better
```suggestion
      tr(ng-repeat='builder in builders  | filter: {name: searchQuery} | filter: isBuilderFiltered' | orderBy: "name" | limitTo: page_size :(currentPage-1)*page_size)
```

rajdeepbharati: The search doesn't work on pages>=2 when I put the filter(s) first

mojca: Does this happen because you try to print table of results as they would appear on the second/third page, which is empty anyway when you just have one single builder match the tag?

rajdeepbharati: Oh got it @tardyp . Now it works perfectly. Thank you 😄

rajdeepbharati: Angularjs filter is a bit magical

rajdeepbharati:
This can be merged now

www/base/src/app/builders/builders.tpl.jade
```diff
@@ -37,7 +39,10 @@
             span(ng-show="tags_filter.length > 0")
                 span.label.clickable.label-danger(ng-click="tags_filter = []") x
           th(style="width:20%px;") Workers
-      tr(ng-repeat='builder in builders | filter: isBuilderFiltered | orderBy: "name"')
+      ul(uib-pagination, total-items="builders.length",
```
tardyp: total-items is not builders.length, it has to be the number of builders filtered. Else, you will have empty pages in the end, like mojca suggested

www/base/src/app/builders/builders.tpl.jade
```diff
@@ -37,7 +39,10 @@
...
+        ng-model="currentPage", class="pagination-sm", boundary-link-numbers="true",
+        max-size=10, items-per-page="page_size")
+      tr(ng-repeat='builder in builders | filter: {name: searchQuery} | filter: isBuilderFiltered | orderBy: "name" | limitTo: page_size :(currentPage-1)*page_size')
```
tardyp: filters are exactly like unix pipe. They are run one by one, and each produce a new array. the last array is passed to ng-repeat. in order to properly implement paging, you will have to use ng-init
```suggestion
      ul(ng-init="filtered_builders = (builders | filter: {name: searchQuery} | filter: isBuilderFiltered)" uib-pagination, total-items="filtered_builders.length", [...])
      tr(ng-repeat='builder in filtered_builders| orderBy: "name" | limitTo: page_size :(currentPage-1)*page_size')
```

rajdeepbharati: ng-init didn't do the trick (it was giving many weird errors), so I used filter in total-items. Now it works (doesn't show empty pages)

www/base/src/app/builders/builders.tpl.jade
rajdeepbharati: Ok, fixed that now

PR #4963, opened 2019-08-12
Add navigation buttons to changebuilds page

rajdeepbharati:
## Contributor Checklist:

### DIFF
--- a/www/base/src/app/changes/changebuilds/changebuilds.controller.js
+++ b/www/base/src/app/changes/changebuilds/changebuilds.controller.js
@@ -16,6 +16,12 @@ class ChangeBuildsController {
             $scope.change = change;
         }
 
+        const allChanges = dataAccessor.getChanges();
+
+        allChanges.onChange = (allChanges) => {
+            $scope.last_changeid = allChanges.length;
+        }
+
         const getBuildsData = function() {
             let requestUrl = `changes/${changeId}/builds`;
             if (!buildsFetchLimit == '') {
--- a/www/base/src/app/changes/changebuilds/changebuilds.tpl.jade
+++ b/www/base/src/app/changes/changebuilds/changebuilds.tpl.jade
@@ -1,4 +1,19 @@
 .container
+  nav
+    ul.pager
+      li.previous(ng-class="{'disabled': change.changeid == 1}")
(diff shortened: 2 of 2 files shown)


www/base/src/app/changes/changebuilds/changebuilds.controller.js
```diff
@@ -16,6 +16,12 @@ class ChangeBuildsController {
             $scope.change = change;
         }
 
+        const allChanges = dataAccessor.getChanges();
+
+        allChanges.onChange = (allChanges) => {
+            $scope.last_changeid = allChanges.length;
```
tardyp: this is not the last_changeid, in case there has been some deleted changes.
It is not a good idea to get the list of all changes, can be huge.

tardyp: Needs to use the paging API of the rest API to do this. I am not sure if it is implemented in data-module. You should first implement it there

rajdeepbharati: Ok

rajdeepbharati: BTW, how can individual changes be deleted? I guess they aren't automatically deleted from bb when removed from the VCS

rajdeepbharati: Does paging API mean using `limit`, `offset`, `order` query parameters? If that is the case, then that is already implemented

rajdeepbharati: @tardyp

tardyp: What would need to be implemented is a getMore() in the collection class. This would automatically get more data without requerying data that we already have (like it is done now in waterfall's loadMore() function)

Another option is to add the page functionality in the collection class, which would use  limit and offset to change the collection's data to switch to another page (without always having the whole data in memory)
You still need to listen to new data events, because as new data arrive, the page will scroll (some objects will go to previous page, and some objects will come from next page)

rajdeepbharati: Ok, thank you

PR #4967, opened 2019-08-14
Search tags in waterfall

rajdeepbharati:
## Contributor Checklist:

### DIFF
--- a/www/waterfall_view/src/module/waterfall.tpl.jade
+++ b/www/waterfall_view/src/module/waterfall.tpl.jade
@@ -3,12 +3,16 @@
         .spinner
             i.fa.fa-circle-o-notch.fa-spin.fa-2x
             p loading
-    p Tags: 
-        span(ng-show="w.all_tags.length > 0" ng-repeat="tag in w.all_tags")
+    p
+        form(role="search" style="width:150px")
+            input(type="text" ng-model="searchedTag" class="form-control" placeholder="Search for tags")
+    p
+        span.label.clickable.label-danger(ng-click="w.clearTags()") x
+        | &nbsp;
+        span(ng-repeat="tag in w.all_tags | filter: searchedTag | limitTo: tags_limit")
             span.builder-tag.label.clickable.label-default(ng-click="w.toggleTag(tag)" ng-class="w.isTagFiltered(tag) ? 'label-success': 'label-default'")
                 | {{ tag }}
             | &nbsp;
-        span.label.clickable.label-danger(ng-click="w.clearTags()") x
     .header
         .header-content
     .content
--- a/www/waterfall_view/src/module/main.module.js
+++ b/www/waterfall_view/src/module/main.module.js
@@ -48,6 +48,7 @@ var WaterfallController = (function() {
             this.$location = $location;
             this.$rootScope = $rootScope;
             this.$scope.tags_filter = this.tags_filter = [];
+            this.$scope.searchedTag = '';
             this.dataProcessorService = dataProcessorService;
             this.bbSettingsService = bbSettingsService;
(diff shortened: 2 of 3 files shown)


tardyp:
I would limit the number of displayed tags to 20. and add suspension points to indicate there is more, and one should use search

www/waterfall_view/src/module/waterfall.tpl.jade
```diff
@@ -3,12 +3,16 @@
...
             p loading
-    p Tags: 
-        span(ng-show="w.all_tags.length > 0" ng-repeat="tag in w.all_tags")
+    p
+        form(role="search" style="width:150px")
+            input(type="text" ng-model="searchedTag" class="form-control" placeholder="Search for tags")
+    p
+        span.label.clickable.label-danger(ng-click="w.clearTags()") x
+        | &nbsp;
+        span(ng-show="w.all_tags.length > 0" ng-repeat="tag in w.all_tags | filter: searchedTag")
```
tardyp: ng-show is redundant with ng-repeat.

tardyp:
I think we should rather use a ready to use tag selector like this one:
http://mbenford.github.io/ngTagsInput/demos

I use it in buildbot travis already

rajdeepbharati:
@tardyp Implementing the + - (and/or/none) filtering might be difficult if we use an external library. I'll try to do it and let you know

tardyp:
I think we can have several tag inputs that you can create using a + button.
each tag within a tag input is "and" and each tag input groups are "or"

image

The tag-input is from the lib, and then you have a small directive to manage tag groups

rajdeepbharati:
Okat, that sounds good (having multiple input fields).
> The tag-input is from the lib, and then you have a small directive to manage tag groups

Does directive mean custom directive?

tardyp:
> Does directive mean custom directive?

Yes. those are quite simple. there are example of such object list editor in buildbot travis

https://github.com/buildbot/buildbot_travis/blob/master/buildbot_travis/ui/src/app/configEditors/projects.tpl.jade#L2
https://github.com/buildbot/buildbot_travis/blob/master/buildbot_travis/ui/src/app/configEditors/projects.tpl.jade#L78

mojca:
Unrelated to the search itself, but I believe that when a tag selection is changed, waterfall should be redrawn from scratch. After applying a tag and while still having two builds available it seems like detective work to dig them out. (I can open a separate issue for that.)

Regarding tags themselves, even with this extra layer I'm still not sure if we could express the full boolean algebra, and some other expressions might still be relatively complicated to write (say: show any of those 10 ports built on any of three platforms; probably 30 queries).

rajdeepbharati:
@mojca redrawing would probably be a bad thing to do.
> After applying a tag and while still having two builds available

Could you show an example of this?

tardyp:
> Regarding tags themselves, even with this extra layer I'm still not sure if we could express the full boolean algebra, and some other expressions might still be relatively complicated to write (say: show any of those 10 ports built on any of three platforms; probably 30 queries).


Sounds like a relevent use case. 2 options:

- if we click on the "and" button, then the logic is inverted, and it becomes an "or"
- We implementing arbitrary logic with regex (i.e tag you specify can be regex)

mojca:
> > After applying a tag and while still having two builds available
> 
> Could you show an example of this?

Here's an example of the lack of redraw:
* https://build.contextgarden.net/#/builders?tags=%2Bc%2Fluatex (works fine)
image
* https://build.contextgarden.net/#/waterfall?tags=c%2Fluatex (empty, some times on the axis are offset by one pixel)
image

But please note that this is something that probably deserves a separate issue to be opened.

PR #4973, opened 2019-08-18
collection.service: bind context

rajdeepbharati:
Functions `put, add, recomputeQuery, sendEvents` need to be in the context of `CollectionInstance`. Otherwise, when they are invoked, it leads to error. For example:
```
angular.js:15567 TypeError: this.put is not a function
    at Array.listener (collection.service.js:85)
    at stream.service.js:42
    at Array.map (<anonymous>)
    at StreamInstance.push (stream.service.js:42)
    at socket.service.js:55
    at l.$eval (angular.js:19393)
    at angular.js:19535
    at $ (angular.js:19793)
    at l.$eval (angular.js:19393)
    at l.$apply (angular.js:19492)
```

## Contributor Checklist:

### DIFF
--- a/www/data_module/src/services/data/collection/collection.service.js
+++ b/www/data_module/src/services/data/collection/collection.service.js
@@ -24,6 +24,10 @@ class Collection {
             constructorImpl(restPath, query, accessor) {
                 let className;
                 this.listener = this.listener.bind(this);
+                this.put = this.put.bind(this);
+                this.add = this.add.bind(this);
+                this.recomputeQuery = this.recomputeQuery.bind(this);
+                this.sendEvents = this.sendEvents.bind(this);
                 this.restPath = restPath;
                 if (query == null) { query = {}; }
                 this.query = query;


rajdeepbharati:
@tardyp It was a pretty simple fix.

tardyp:
Can you give a bit more explaination on why bind is needed?

rajdeepbharati:
Because those functions were not in the context `this`, and had been invoked.

tardyp:
Please give some more details. I think the fix is rather the invocation is wrong somewhere, and the functions should be called with correct context. Here you are adding a level of indirection for all calls, including the correct ones. This uses more memory, and more cpu.

The stack trace you have shows that listener the caller. But listener is already a bind function. So the problem should be else where.

rajdeepbharati:
Oh okay, let me check

rajdeepbharati:
@tardyp I'm not sure why this isn't working. But the bug seems to be unrelated to Vue (IMO vue is compatible with data module)

tardyp:
Well, this happens only with our vue boilerplate. Might not be related to vue, but with how we did the boilerplate.

rajdeepbharati:
workaround in https://github.com/macports/buildbot-macports-custom-views/pull/21

PR #4984, opened 2019-08-20
Fix custom templates

### LINKED ISSUE #4980: custom_templates are not working anymore
custom_template uses $templateCache to hijack the templates already included in the core app.
https://github.com/buildbot/buildbot/blob/e8b3696a23335236b69e5ebaa7d7976c5d994f2a/www/base/src/app/layout.jade#L30

With new webpack, we now use require to load templates

https://github.com/buildbot/buildbot/blob/e8b3696a23335236b69e5ebaa7d7976c5d994f2a/www/base/src/app/builders/builder/builder.route.js#L16

tardyp:
We remove pug-loader and replace by a custom loader to better control our templates.
pug-loader was actually building dynamic templates, and not directly html, so now, we build jade to html at build time, and thus spare a few bytes.

Also we fix the dual load of angularjs (once from base, and once from guanlecoja-ui)

FIxes #4980 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/custom_templates.bugfix
+++ b/master/buildbot/newsfragments/custom_templates.bugfix
@@ -1,0 +1,1 @@
+custom_templates are now working again.
--- a/master/docs/manual/configuration/www.rst
+++ b/master/docs/manual/configuration/www.rst
@@ -98,23 +98,28 @@ This server is configured with the ``www`` configuration key, which specifies a 
 
 ``custom_templates_dir``
     This directory will be parsed for custom angularJS templates to replace the one of the original website templates.
+    You can use this to slightly customize buildbot look for your project, but to add any logic, you will need to create a full-blown plugin.
     if the directory string is relative, it will be joined to the master's basedir.
-    Either ``*.jade`` files or ``*.html`` files can be used, and will be used to override ``views/<filename>.html`` templates in the angularjs templateCache.
-    Unlike with the regular nodejs based angularjs build system, Python only jade interpreter is used to parse the jade templates.
+    Buildbot uses the jade file format natively (which has been renamed 'pug' in the nodejs ecosystem), but you can also use html format if you prefer.
+    
+    Either ``*.jade`` files or ``*.html`` files can be used, and will be used to override templates with the same name in the UI.
+    On the regular nodejs UI build system, we use nodejs's pug module to compile jade into html.
+    For custom_templates, we use the pyjade interpreter to parse the jade templates, before sending them to the UI.
     ``pip install pyjade`` is be required to use jade templates.
-    You can also override plugin's directives, but they have to be in another directory.
+    You can also override plugin's directives, but they have to be in another directory, corresponding to the plugin's name in its ``package.json``.
+    For example:
 
     .. code-block:: none
 
         # replace the template whose source is in:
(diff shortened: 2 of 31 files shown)


master/docs/manual/configuration/www.rst
```diff
@@ -98,23 +98,28 @@ This server is configured with the ``www`` configuration key, which specifies a
 
 ``custom_templates_dir``
     This directory will be parsed for custom angularJS templates to replace the one of the original website templates.
+    You can use this to slightly customize buildbot look for your project, but to add any logic, you will need to create a full-blown plugin.
     if the directory string is relative, it will be joined to the master's basedir.
-    Either ``*.jade`` files or ``*.html`` files can be used, and will be used to override ``views/<filename>.html`` templates in the angularjs templateCache.
-    Unlike with the regular nodejs based angularjs build system, Python only jade interpreter is used to parse the jade templates.
+    Buildbot uses the jade file format natively (which has been renamed 'pug' in the nodejs ecosystem), but you can also use html format if you prefer.
```
p12tic: renamed *to* pug

www/build_common/src/webpack.js
```diff
@@ -58,7 +58,7 @@ module.exports.createTemplateWebpackConfig = function(options) {
     };
 
     config.optimization = {
-        minimize: isProd,
+        minimize: true,
```
p12tic: Won't this make debugging harder?

www/build_common/src/ng-template-loader.js
```diff
@@ -0,0 +1,24 @@
...
+const path = require('path')
+var loaderUtils = require("loader-utils");
+
+/* ultra simple loader that only support our simple usecase 
+  pug-loader is much more complicated, but do support lot of features, 
+  including require inside template, which we don't really need.
+*/
+
+module.exports = function() {
+  var fn = this.resourcePath;
```
p12tic: Code convention does not match rest of Buildbot.

www/codeparameter/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "buildbot-codeparameter",
```
p12tic: Won't this increase possibility of accidental name clashes? E.g. what happens if there's an indirect dependency in node_modules of the same name? I think it's worth to prefix our plugins with Buildbot even if they aren't published to npm.

master/buildbot/newsfragments/custom_templates.bugfix
```diff
@@ -0,0 +1 @@
+custom_templates are now working again.
```
p12tic: Wondering if it's worth having 2.4.1 with this fix. I guess it's dealbreaker for some people isn't it?

www/codeparameter/package.json
tardyp: The reason I changed it is that it should match the name of the plugin in buildbot (as per legacy way of identifying custom_templates).
I can use another identifier.

tardyp:
@p12tic @rajdeepbharati should be fixed now

master/buildbot/newsfragments/custom_templates.bugfix
tardyp: we had one report over IRC. not sure if it is worth urgent fix. Probably wait a bit more for more issues

PR #4990, opened 2019-08-26
set committer as nullable

### LINKED ISSUE #4987: hgpoller fails with "null value in column committer"
I'm running a Buildbot master inside a Docker container, and I have an HgPoller set up to poll a Kallithea repository. When the poller detects changes, it fails with the following:

```
buildbot_1  | 2019-08-23 19:58:52+0000 [-] hgpoller: repo poll failed
{... unneeded stack trace ...}
buildbot_1  |   sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "committer" violates not-null constraint
buildbot_1  |   DETAIL:  Failing row contains (9, marisa <marisa@mokou.zone>, null, Add lint script for CI execution, default, 8c43f72068449130372da25f6583531c9e4f88a8, , 1566588893, null, https://callie.marisa.cloud/tometo/tometo, , tometo, 1, null).
```

Reading through the hgpoller source, this seems to stem from here:

https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/hgpoller.py#L285

Am I correct in assuming that `committer=None` is directly opposed to `committer` being a non-null constraint in the database? If so, wouldn't that be a clear bug?

tardyp:
Fix #4987

The migration table already has is nullable.
so #4987 is only about new setups

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/model.py
+++ b/master/buildbot/db/model.py
@@ -370,7 +370,7 @@ class Model(base.DBConnectorComponent):
         sa.Column('author', sa.String(255), nullable=False),
 
         # committer's name
-        sa.Column('committer', sa.String(255), nullable=False),
+        sa.Column('committer', sa.String(255), nullable=True),
 
         # commit comment
         sa.Column('comments', sa.Text, nullable=False),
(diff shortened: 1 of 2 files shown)


mojca:
Don't you need to make a database transaction to modify the tables also for those users who made their first setup with 2.4.0?

tardyp:
I agree with mojca that we need to fix the 2.4.0 instances.
I don't like migration scripts, but I think there is no choice. We need a migration script that detects if the column is non nullable. I'll do it soon in a next PR

PR #4992, opened 2019-08-27
add migration script for non nullable committer column

tardyp:
2.4.0 new instances have their committer column non nullable
so we need to fix them

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_db_migrate_versions_055_add_changes_committer.py
+++ b/master/buildbot/test/unit/test_db_migrate_versions_055_add_changes_committer.py
@@ -1,0 +1,85 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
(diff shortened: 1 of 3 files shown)


master/buildbot/test/unit/test_db_migrate_versions_055_add_changes_committer.py
```diff
@@ -0,0 +1,84 @@
...
+import sqlalchemy as sa
+
+from twisted.trial import unittest
+
+from buildbot.test.util import migration
+from buildbot.util import sautils
+
+
+class Migration(migration.MigrateTestMixin, unittest.TestCase):
+    PREVIOUS_NULLABLE = True
```
rajdeepbharati: Shouldn't this be inside `__init__`?

tardyp: Why? This is a class constant. This is a common pattern for creating similar  parametrised tests

rajdeepbharati: Okay 👍

Issue #4994, opened 2019-08-29
`__contains` filter doesn't allow `.` (dot) character


rajdeepbharati:
For example, I want to query something like:
`http://localhost:8010/api/v2/builders?name__contains=llvm-6.0`
It shows error:
```
{"error": "invalid filter value for name__contains"}
```

rajdeepbharati:
@tardyp @p12tic Is it possible?
I guess it is coming from here https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/rest.py#L325

rajdeepbharati:
I changed [this line](https://github.com/buildbot/buildbot/blob/82e70bc62598fad0737c0b814013be79ee1247b9/master/buildbot/www/rest.py#L321) to `values = [bytes2unicode(v)`
It worked, but a few tests are failing

tardyp:
It is because BuilderName is identifier in the data api.

The regular expression there do not contain dot
https://github.com/buildbot/buildbot/blob/747279acfbfde3a9d3b9b3b78c72329f39cefa29/master/buildbot/data/types.py#L159

Can you fix it?

rajdeepbharati:
OK thanks, I will try to fix this.

mojca:
I started with #4861, but had no time to finish it, so I closed it until I would find time to fix all the unit tests etc. Having this fixed would be awesome also from the perspective of regular builder names, not just the virtual ones.

rajdeepbharati:
@tardyp This one seems to be a really simple fix (just by changing the regex) and surprisingly the CI isn't breaking. But in the one by @mojca (dots in workernames) I do get a ton of failed tests.

PR #4997, opened 2019-09-02
Allow dots in identRe

rajdeepbharati:
Fixes https://github.com/buildbot/buildbot/issues/4994

## Contributor Checklist:

### DIFF
--- a/master/buildbot/data/types.py
+++ b/master/buildbot/data/types.py
@@ -156,7 +156,7 @@ class Boolean(Instance):
 class Identifier(Type):
 
     name = "identifier"
-    identRe = re.compile('^[a-zA-Z_-][a-zA-Z0-9_-]*$')
+    identRe = re.compile('^[a-zA-Z_-][a-zA-Z0-9._-]*$')
     ramlType = "string"
 
     def __init__(self, len=None, **kwargs):


master/buildbot/data/types.py
```diff
@@ -156,7 +156,7 @@ def valueFromString(self, arg):
 class Identifier(Type):
 
     name = "identifier"
-    identRe = re.compile('^[a-zA-Z_-][a-zA-Z0-9_-]*$')
+    identRe = re.compile('^[a-zA-Z._-][a-zA-Z0-9._-]*$')
```
tardyp: start with '.' looks weird for a buildername

mojca: I initially included this one, but I could not figure out where it was used. Thanks for re-discovering it :)
(Maybe it would help to at least comment on where this was used, or is it too broad?)

Isn't `__Janitor` as a builder name, or a builder name starting with a dash (`-`) equally weird as a builder name starting with a dot? I do imagine that some users might for whatever reason use a leading dot (maybe also to distinguish it from other regular builders), and unless there are technical difficulties with that, I don't see a good reason to prevent it (just because it looks weird).

tardyp: > Isn't __Janitor as a builder name, or a builder name starting with a dash (-) equally weird as a builder name starting with a dot?
True.

It is called identifier, and it is starting to become less and less an identifier.

At this point, we could also just change all those identifier types to unrestricted strings, as I am not sure there is really any technical issue beyond that

rajdeepbharati: Hmm, starting with '.' does look weird. Variable names are usually allowed to start with letters or '_'. Same concept can go for buildernames in data api query. I had one doubt: how are we able to use dots in buildernames although it isn't present in the identifiers at https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/identifiers.py#L20 ?

Issue #2957, opened 2017-03-09
irc bot can create or stop builds without any authz


tardyp:
This ticket is a migrated Trac ticket [3377](http://trac.buildbot.net/ticket/3377)

People contributed to the original ticket: @tardyp
Ticket created on: `Oct 28 2015`
Ticket last modified on: `Oct 28 2015`

---

A simple authz would be to only allow channel ops or voices to use commands

---

macdems:
I am right now strongly refactoring IRC bot code. The main idea is to create a bot for a Telegram messaging app (and possibly for other webhook based chats, like Messenger or Slack).

I am addressing this issue in the work. The idea is very simple: in `master.cfg` you can specify which commands can be run by which users (for IRC identified by nicks) or allow or deny access to everybody.

You can take preview on https://github.com/macdems/buildbot (I will submit PR when I finish) or talk to `testbuildbot` on `#testbuildbot` channel on freenode. You can also talk to `@testbuildbot` on Telegram (I will keep them running for a while, but I cannot guarantee that they will be online all the time).

PR #5003, opened 2019-09-07
Refactored IRC bot

macdems:
This PR contains a refactor of an IRC bot, addressing #2957. This is a first part of the split PR #4524 

The code coverage is little below 90%, but I have covered most of my code.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -13,15 +13,24 @@
 #
 # Copyright Buildbot Team Members
 
-
 from twisted.application import internet
+from twisted.internet import defer
 from twisted.internet import task
 from twisted.python import log
 from twisted.words.protocols import irc
 
 from buildbot import config
+from buildbot.process.results import CANCELLED
+from buildbot.process.results import EXCEPTION
+from buildbot.process.results import FAILURE
+from buildbot.process.results import RETRY
+from buildbot.process.results import SUCCESS
+from buildbot.process.results import WARNINGS
+from buildbot.reporters.words import Channel
+from buildbot.reporters.words import Contact
 from buildbot.reporters.words import StatusBot
 from buildbot.reporters.words import ThrottledClientFactory
+from buildbot.reporters.words import dangerousCommand
 from buildbot.util import service
 from buildbot.util import ssl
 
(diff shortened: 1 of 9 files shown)


macdems:
Actually the biggest change is a separation between `Contact` and `Channel` classes, so user on the same channel can share channel state (and private chats are considered channels with the same name as the user). Consequently, the notification state is associated with the channel, which makes more sense: every authorized user can change if for a channel. Previously, if one user set notifications for the channel, another one (even channel operator) could not change their state. For public channels only channel ops and the users explicitly specified in the ACL for `notify`, `mute` and `unmute` commands can change the notification state.

#2957 is addressed by a new `authz` configuration parameter, which replaces `allowForce` and `allowShutdown`. These two params were removed on purpose (probably breaking some existing config), as I consider them the security issue of #2957. Hence, every config that is using them right now is unsafe and should be reviewed. `authz` is a simple ACL for every command (with sane defaults) and is nick-based, so one must rely on the nick registration to the IRC server.

master/buildbot/reporters/words.py
```diff
@@ -121,207 +96,91 @@ def parseArgs(self, *args):
...
+
+def dangerousCommand(method):
+    command = method.__name__
+    if not command.startswith('command_'):
+        raise ValueError('@dangerousCommand can be used only for commands')
+    dangerous_commands.append(command[8:])
+    return method
+
+
+def convertTime(seconds):
```
p12tic: This could be moved to the `buildbot.util` module as it's likely to be useful in other places of Buildbot. Maybe for a future PR.

master/buildbot/reporters/irc.py
```diff
@@ -135,31 +321,80 @@ def kickedFrom(self, channel, kicker, message):
...
 
     shuttingDown = False
     p = None
 
-    def __init__(self, nickname, password, channels, pm_to_nicks, tags, notify_events,
+    def __init__(self, nickname, password, join_channels, pm_to_nicks, authz, tags, notify_events,
                  noticeOnChannel=False,
                  useRevisions=False, showBlameList=False,
                  parent=None,
-                 lostDelay=None, failedDelay=None, useColors=True, allowShutdown=False):
```
p12tic: I think we need a migration path for this feature as according to the Buildbot versioning polity we would have to do a major version bump. Perhaps it makes sense to print a deprecation warning when allowShutdown is passed and them modify authz to have similar effect as previous `allowShutdown` functionality.

master/buildbot/reporters/words.py
macdems: I think it is specific to `words` and commands defined here (and below).

I would rather move `overrideCommand` to `buildbot.util`, as it can have more uses (it copies docstring and `usage` attr of the base method).

master/buildbot/reporters/irc.py
macdems: I consider this a security fix. However, adding decprecated params with a warning is easy, so what is a decision?

macdems: We actually have three options:

1. Totally remove `allowForce` and `allowShutdown`. This is most secure and fully fixes #2957, however, it may break existing configurations.

2. Log a warning when `allowForce` or `allowShutdown` is used and **ignore it**. This way, the existing configurations will still work, but the bot behavior will change (with serious insecurity removed). This is an option I vote for now, until the next major version bump.

3. Log a warning and honor `allowForce` or `allowShutdown` by adding corresponding entries to `authz`. This way any insecure config will remain insecure (I guess there are not many od them, because of explicit warning agains  #2957 in the current documentation).

p12tic: @tardyp ^^

p12tic: Personally I think (1) is the best option, as silent breakage of Buildbot configuration is bad. Many people don't monitor logs for warnings.

master/buildbot/reporters/words.py
macdems: Ok, I did not realize you talked about `convertTime`. I moved it to `buildbot.util`.

master/buildbot/reporters/irc.py
macdems: So after a discussion, I implemented option 3 in 3a83bf8.

master/buildbot/reporters/words.py
```diff
@@ -678,369 +773,473 @@ def command_FORCE(self, args):
...
+    def _get_object_id(self):
+        return self.master.db.state.getObjectId(
+            self.nickname, '{0.__module__}.{0.__name__}'.format(self.__class__))
+
+    @defer.inlineCallbacks
+    def _save_channels_state(self, attr, json_type=None):
+        if json_type is None:
+            json_type = lambda x: x
+        data = [(channel.id, json_type(getattr(channel, attr)))
+                for channel in self.channels.values()]
```
tardyp: If I understand correctly, this will grow without limit even if notification is disabled, right?

PR #5011, opened 2019-09-14
DockerLatentWorker: Allow same volume twice

jangmarker:
Docker supports to mount the same named volume or host directory twice,
but the way that Buildbot implemented volumes that was not possible.

Previously buildbot would parse volume mounts (`<volume name/host path>:<container path>[:<rw/ro>]`) by itself and provide a dictionary with keys of `<volume name/host path>` to the docker-py API. Obviously a dictionary like that does not allow to contain the same volume name/host path twice so a volume could only be mounted once for a container.

## Contributor Checklist:

   * nothing to do here as far as I could see

### DIFF
--- a/master/buildbot/newsfragments/better-docker-volumes.bugfix
+++ b/master/buildbot/newsfragments/better-docker-volumes.bugfix
@@ -1,0 +1,1 @@
+DockerLatentWorker: Allow to bind the same volume twice into a worker's container, Buildbot now requires 'docker-py' (nowadays 'docker') version 1.2.3+ from 2015.
--- a/master/buildbot/test/unit/test_worker_docker.py
+++ b/master/buildbot/test/unit/test_worker_docker.py
@@ -143,7 +143,7 @@ class TestDockerLatentWorker(unittest.SynchronousTestCase, TestReactorMixin):
         self.assertEqual(client.call_args_create_container[0]['volumes'],
                          ['/opt/webapp'])
         self.assertEqual(client.call_args_create_host_config[0]['binds'],
-                         {'/src/webapp': {'bind': '/opt/webapp', 'ro': False}})
+                         ["/src/webapp:/opt/webapp"])
 
     def test_volume_ro_rw(self):
         bs = self.setupWorker('bot', 'pass', 'tcp://1234:2375', 'worker', ['bin/bash'],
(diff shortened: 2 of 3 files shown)


tardyp:
Probably this means that we losse compat with ancient version of docker api. did you looks at it?

jangmarker:
I did just now, thanks for thinking of that.

My research:

As far as I can tell, there was a refactoring in Nov 2016 that introduced documentation which says this volume syntax is possible:
https://github.com/docker/docker-py/pull/1186/files?file-filters%5B%5D=.py#diff-3793699cca9b8cffcf995fca8d9318b4R334

The pull request was included in release 2.3 of the docker-py API from May 2017:
https://github.com/docker/docker-py/commit/8478491cf8cef62d210e477befc02efc6f241ba3
https://github.com/docker/docker-py/releases/tag/2.3.0

So since docker-py 2.3 this behavior was documented.

The feature was originally introduced in June 2015 by:
https://github.com/docker/docker-py/commit/1446b8c5eef57bb5f7392f73fada9ba4dc305de9

which was included in release 1.2.3 of June 2015
https://github.com/docker/docker-py/releases/tag/1.2.3

So my commit breaks docker-py versions before 1.2.3 of June 2015. 
With version 2.0.0 the Python package dropped support for Python 2 and was renamed from `docker-py` to `docker`. Buildbot dropped support for Python 2 with the 2.0.0 release a while ago, so newer Buildbot versions would have to use at least `docker 2.0.0` which supports the docker volumes as strings syntax.

Although the `workers/docker.py` includes a version number of the `docker-py` library there is no check for a minimum version.

So I think were ok here but I included as much info as I could find for your judgement.

tardyp:
Thanks for the research. I think this make sense to add a summary of this in the release note.
"This new API requires docker-py 2.3+, released in 2015."

jangmarker:
@tardyp do you want me to mention `1.2.3` in the message? Because that should technically still be supported except for the Python 2 problem.

Otherwise I'll just add the message that you suggested. Should I add it as a breaking change or just append it to the change note that I added already?

jangmarker:
I added a note with version `1.2.3` of 2015.

p12tic:
A note on the release notes is OK as we already require Python 3 dependency. Thanks!

PR #5013, opened 2019-09-15
www: fix data_module tests with Chromium 77.0.3865

yan12125:
Accept patterns before and after 52e7fb4.
Apparently Chromium has changed back how Regexp is handled.

Inspired by https://stackoverflow.com/a/42317843/3786245

## Contributor Checklist:

### DIFF
--- a/www/data_module/src/services/dataUtils/dataUtils.service.spec.js
+++ b/www/data_module/src/services/dataUtils/dataUtils.service.spec.js
@@ -64,10 +64,16 @@ describe('Data utils service', function() {
             expect(result.test("asd/1/new")).toBeTruthy();
 
             result = dataUtilsService.socketPathRE('asd/1/bnm/*/*').source;
-            expect(result).toBe('^asd\\/1\\/bnm\\/[^\\/]+\\/[^\\/]+$');
+            expect([
+                '^asd\\/1\\/bnm\\/[^\\/]+\\/[^\\/]+$',
+                '^asd\\/1\\/bnm\\/[^/]+\\/[^/]+$'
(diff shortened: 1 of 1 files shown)


yan12125:
Tests in BBTravis fail as it uses older Chrome (71.0.3578). How can I specify a newer version?

tardyp:
We should support both version, probably..

yan12125:
> We should support both version, probably..

Is there a good way to do it? Looks like buildbot tests use Jasmine, which provides `toMatch()`, but I feel matching regular expressions with regular expressions kind of messy.

yan12125:
I updated this PR to include both old and new patterns. Tests pass on 77.0.3865 (my Arch Linux box) and 71.0.3578 (buildbot).

www/data_module/src/services/dataUtils/dataUtils.service.spec.js
```diff
@@ -64,10 +64,16 @@ describe('Data utils service', function() {
...
+                '^asd\\/1\\/bnm\\/[^\\/]+\\/[^\\/]+$',
+                '^asd\\/1\\/bnm\\/[^/]+\\/[^/]+$'
+            ]).toContain(result);
 
             result = dataUtilsService.socketPathRE('asd/1/*').source;
-            expect(result).toBe('^asd\\/1\\/[^\\/]+$');
+            expect([
+                '^asd\\/1\\/[^\\/]+$',
+                '^asd\\/1\\/[^/]+$'
+            ]).toContain(result);
```
tardyp: 👍

PR #5016, opened 2019-09-19
Added Telegram bot

macdems:
This is a Telegram bot built on top of the refactored `words.py`. It can use both long polling and webhooks, so similar concept can be used for easy making of other webhook based bots (e.g. Messenger and Slack).

### DIFF
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -88,7 +88,9 @@ class IRCChannel(Channel):
 
 class IRCContact(Contact):
 
-    def __init__(self, bot, user=None, channel=None):
+    def __init__(self, bot, user, channel=None):
+        if channel is None:
+            channel = user
         super().__init__(bot, user, channel)
 
     def send(self, message):
@@ -109,18 +111,17 @@ class IRCContact(Contact):
         if verb == "kicks":
             response = "%s back" % verb
         elif verb == "threatens":
-            response = "hosts a red wedding for %s" % self.user
+            response = "hosts a red wedding for %s" % self.user_id
         else:
-            response = "%s %s too" % (verb, self.user)
+            response = "%s %s too" % (verb, self.user_id)
         self.act(response)
 
     @defer.inlineCallbacks
     def op_required(self, command):
-        if self.channel.id != self.user and \
-                self.user not in self.bot.authz.get(command.upper(), ()):
-            ops = yield self.bot.getChannelOps(self.channel.id)
-            return self.user not in ops
-        return False
+        if self.is_private_chat or self.user_id in self.bot.authz.get(command.upper(), ()):
+            return False
+        ops = yield self.bot.getChannelOps(self.channel.id)
+        return self.user_id not in ops
 
     # IRC only commands
 
@@ -144,7 +145,7 @@ class IRCContact(Contact):
     def command_MUTE(self, args, **kwargs):
         if (yield self.op_required('mute')):
             self.send("Only channel operators or explicitly allowed users "
-                      "can mute me here, {}... Blah, blah, blah...".format(self.user))
+                      "can mute me here, {}... Blah, blah, blah...".format(self.user_id))
             return
         # The order of these is important! ;)
         self.send("Shutting up for now.")
@@ -167,21 +168,29 @@ class IRCContact(Contact):
     @defer.inlineCallbacks
     @Contact.overrideCommand
     def command_NOTIFY(self, args, **kwargs):
-        if self.channel.id != self.user:
+        if not self.is_private_chat:
             argv = self.splitArgs(args)
             if argv and argv[0] in ('on', 'off') and \
                     (yield self.op_required('notify')):
                 self.send("Only channel operators can change notified events for this channel. "
                           "And you, {}, are neither!"
-                          .format(self.user))
+                          .format(self.user_id))
                 return
         super().command_NOTIFY(args, **kwargs)
+
+    def command_DANCE(self, args, **kwargs):
+        """dance, dance academy..."""
+        self.bot.reactor.callLater(1.0, self.send, "<(^.^<)")
+        self.bot.reactor.callLater(2.0, self.send, "<(^.^)>")
+        self.bot.reactor.callLater(3.0, self.send, "(>^.^)>")
+        self.bot.reactor.callLater(3.5, self.send, "(7^.^)7")
+        self.bot.reactor.callLater(5.0, self.send, "(>^.^<)")
 
     def command_DESTROY(self, args):
         if self.bot.nickname not in args:
             self.act("readies phasers")
         else:
-            self.send("Better destroy yourself, {}!".format(self.user))
+            self.send("Better destroy yourself, {}!".format(self.user_id))
 
     def command_HUSTLE(self, args):
         self.act("does the hustle")
@@ -231,10 +240,10 @@ class IrcStatusBot(StatusBot, irc.IRCClient):
 
     def getContact(self, user, channel=None):
         # nicknames and channel names are case insensitive
-        if user:
-            user = user.lower()
-        if channel:
-            channel = channel.lower()
+        user = user.lower()
+        if channel is None:
+            channel = user
+        channel = channel.lower()
         return super().getContact(user, channel)
 
     # the following irc.IRCClient methods are called when we have input
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -15,6 +15,7 @@
 
 from twisted.application import internet
 from twisted.internet import defer
+from twisted.internet import reactor
 from twisted.internet import task
 from twisted.python import log
 from twisted.words.protocols import irc
@@ -177,11 +178,11 @@ class IRCContact(Contact):
 
     def command_DANCE(self, args, **kwargs):
         """dance, dance academy..."""
-        self.bot.reactor.callLater(1.0, self.send, "<(^.^<)")
-        self.bot.reactor.callLater(2.0, self.send, "<(^.^)>")
-        self.bot.reactor.callLater(3.0, self.send, "(>^.^)>")
-        self.bot.reactor.callLater(3.5, self.send, "(7^.^)7")
-        self.bot.reactor.callLater(5.0, self.send, "(>^.^<)")
+        reactor.callLater(1.0, self.send, "<(^.^<)")
+        reactor.callLater(2.0, self.send, "<(^.^)>")
+        reactor.callLater(3.0, self.send, "(>^.^)>")
+        reactor.callLater(3.5, self.send, "(7^.^)7")
+        reactor.callLater(5.0, self.send, "(>^.^<)")
 
     def command_DESTROY(self, args):
         if self.bot.nickname not in args:
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -88,13 +88,10 @@ class IRCChannel(Channel):
 
 class IRCContact(Contact):
 
-    def __init__(self, bot, user, channel=None):
+    def __init__(self, user, channel=None):
         if channel is None:
             channel = user
-        super().__init__(bot, user, channel)
-
-    def send(self, message):
-        return self.channel.send(message)
+        super().__init__(user, channel)
 
     def act(self, action):
         return self.channel.act(action)
--- a/master/buildbot/reporters/irc.py
+++ b/master/buildbot/reporters/irc.py
@@ -314,13 +314,13 @@ class IrcStatusBot(StatusBot, irc.IRCClient):
         return [n[1:] for n in names if n[0] in '@&~%']
 
     def joined(self, channel):
-        self.log("I have joined %s" % (channel,))
+        self.log("Joined %s" % (channel,))
         # trigger contact constructor, which in turn subscribes to notify events
         channel = self.getChannel(channel=channel)
         channel.add_notification_events(self.notify_events)
 
     def left(self, channel):
-        self.log("I have left %s" % (channel,))
+        self.log("Left %s" % (channel,))
 
     def kickedFrom(self, channel, kicker, message):
         self.log("I have been kicked from %s by %s: %s" % (channel,
--- a/master/buildbot/reporters/telegram.py
+++ b/master/buildbot/reporters/telegram.py
@@ -350,9 +350,11 @@ class TelegramContact(Contact):
                 prompt = "{}, now {}...".format(greeting, prompt)
             else:
                 prompt = prompt[0].upper() + prompt[1:] + "..."
-            kwargs['reply_markup'] = {
-                'force_reply': True
-            }
+            # Telegram seems to have a bug, which causes reply request to pop sometimes again.
+            # So we do not force reply to avoid it...
+            # kwargs['reply_markup'] = {
+            #     'force_reply': True
+            # }
         self.send(prompt, **kwargs)
 
     @defer.inlineCallbacks
--- a/master/buildbot/reporters/telegram.py
+++ b/master/buildbot/reporters/telegram.py
@@ -19,13 +19,9 @@ import random
 import shlex
 
 from twisted.internet import defer
-from twisted.internet import task
-from twisted.python import log
-from twisted.web import resource
-from twisted.web import server
+from twisted.internet import reactor
 
 from buildbot import config
-from buildbot.plugins.db import get_plugins
 from buildbot.process.results import CANCELLED
 from buildbot.process.results import EXCEPTION
 from buildbot.process.results import FAILURE
@@ -36,8 +32,10 @@ from buildbot.reporters.words import Channel
 from buildbot.reporters.words import Contact
 from buildbot.reporters.words import StatusBot
 from buildbot.reporters.words import UsageError
+from buildbot.reporters.words import WebhookBotMixin
 from buildbot.schedulers.forcesched import CollectedValidationError
 from buildbot.schedulers.forcesched import ForceScheduler
+from buildbot.util import asyncSleep
 from buildbot.util import bytes2unicode
 from buildbot.util import httpclientservice
 from buildbot.util import service
@@ -134,7 +132,7 @@ class TelegramContact(Contact):
     @defer.inlineCallbacks
     def command_START(self, args, **kwargs):
         yield self.command_HELLO(args)
-        self.bot.reactor.callLater(0.2, self.command_HELP, '')
+        reactor.callLater(0.2, self.command_HELP, '')
 
     def command_NAY(self, args, tmessage, **kwargs):
         """forget the current command"""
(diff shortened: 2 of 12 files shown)


master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+            del args[0]
+        except ValueError:
+            if args[0] == 'all':
+                all = True
+                del args[0]
+        except IndexError:
+            pass
+
+        if all:
+            num = 20
```
p12tic: Why are we limiting the amount of returned data if the user explicitly requested `all`?

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+            self.ask_for_reply("give me the reason to stop build on `{}`".format(argv[0]))
+
+    @Contact.overrideCommand
+    def command_SHUTDOWN(self, args, **kwargs):
+        if args:
+            return super().command_SHUTDOWN(args)
+        shuttingDown = self.master.botmaster.shuttingDown
+        keyboard = [[
+             self.query_button("🔙 Stop Shutdown", '/shutdown stop')
+             if shuttingDown else
```
p12tic: I think this is not intuitive formatting. Could we move the conditional out of the list definition?

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+        msg = yield self.send("**<(^.^<)**")
+        if msg is not None:
+            mid = msg['message_id']
+            self.bot.reactor.callLater(1.0, self.bot.edit_message, chat, mid, "**<(^.^)>**")
+            self.bot.reactor.callLater(2.0, self.bot.edit_message, chat, mid, "**(>^.^)>**")
+            self.bot.reactor.callLater(2.5, self.bot.edit_message, chat, mid, "**(7^.^)7**")
+            self.bot.reactor.callLater(4.0, self.bot.edit_message, chat, mid, "**(>^.^<)**")
+            self.bot.reactor.callLater(5.0, self.bot.delete_message, chat, mid)
+            self.bot.reactor.callLater(5.5, self.bot.send_sticker, chat, random.choice((
+                'CAADAgAD9wEAAsoDBgtCnbBFfI8M_BYE',
```
p12tic: Needs docs of what these stickers are.

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+                yield self.send(
+                    "{}".format(text),
+                    reply_to_message_id=tmessage['message_id'])
+                yield self.bot.send_sticker(uid, sticker)
+                return
+
+        self.send(
+            "⛔  ACCESS DENIED  ⛔\n\n" +
+            random.choice((
+                "You are simply not allowed to to this!",
```
p12tic: I think it's worth refraining from adding new jokes to the codebase. In this case we are adding a sizeable amount of code for no real benefit for the end user.

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+
+    @defer.inlineCallbacks
+    def answer_query(self, query_id, notify=None):
+        params = dict(callback_query_id=query_id)
+        if notify is not None:
+            params.update(dict(text=notify))
+        return (yield self._post('/answerCallbackQuery', json=params))
+
+    @defer.inlineCallbacks
+    def send_message(self, channel, message, parse_mode='Markdown', **kwargs):
```
p12tic: Could we require `channel` to be `TelegramChannel`?

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+from buildbot.schedulers.forcesched import ForceScheduler
+from buildbot.util import bytes2unicode
+from buildbot.util import httpclientservice
+from buildbot.util import service
+from buildbot.util import unicode2bytes
+
+
+class TelegramChannel(Channel):
+
+    def __init__(self, bot, channel):
```
p12tic: Could we pass channel info and id as separate arguments? The current approach will make code harder to understand and maintain in the future.

What I would propose is having `channel_info` and `channel_id`. Both are `None` by default and exactly one is accepted. `channel_id` would be inferred from `channel_info` dict if it is specified. `channel_id` would be sent to `super()`. `channel_info` would be stored as `self.chat`.

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+        if field['fullName']:
+            yield field
+        if 'fields' in field:
+            yield from collect_fields(field['fields'])
+
+
+class TelegramContact(Contact):
+
+    def __init__(self, bot, user=None, channel=None):
+        super().__init__(bot, user, channel)
```
p12tic: I think it makes sense to have separate arguments for user info and user id data just like in `TelegramChannel`. The info part could be stored as separate instance variable. This way we would not need to override `userid` property and code as a whole would be easier to follow.

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1055 @@
...
+    @property
+    def userid(self):
+        if isinstance(self.user, dict):
+            return self.user['id']
+        else:
+            return self.user
+
+    @property
+    def user_full_name(self):
+        fullname = " ".join((self.user['first_name'],
```
p12tic: Could `first_name` be absent?

master/buildbot/reporters/telegram.py
macdems: Because of the Telegram message length limit. Actually I had to limit it to 10. I have added a comment, so the user knows that the number of shown changes has been limited.

master/buildbot/reporters/telegram.py
macdems: Ok, done.

master/buildbot/reporters/telegram.py
macdems: It is a Telegram stickers hash ID, not base64 of anything special. I have removed the stickers anyway.

master/buildbot/reporters/telegram.py
macdems: I have changed the name of the argument to `chat`, which is numeric chat id. This is consistent with other methods of this kind.

master/buildbot/reporters/telegram.py
macdems: I think you are right. Actually the IRC bot has way too much of such stuff (I have already moved most of it from `words.py` to `irc.py`.

I have removed the complicated code, however, I have left several _access denied_ messages, so there is more variety in the chat. I prefer that the bot mimics human conversation rather than be just a command line. This is why I tried to make it answer in complete sentences where applicable.

master/buildbot/reporters/telegram.py
macdems: I strongly prefer not to change the signature of the overridden method (including constructor) if not absolutely necessary.

I have changed the object attribute names (to `chat_info`) and refactored the code so `TelegramChannel` constructor **always** gets a dict.

master/buildbot/reporters/telegram.py
macdems: The same as above. Now `user` and `channel` are compulsory and for `TelegramContact` they must be a dict. Again, I renamed object attributes to `user_id` and `user_info` to make them clear.

master/buildbot/reporters/telegram.py
macdems: No, it cannot. In [Telegram API](https://core.telegram.org/bots/api#user) it is not optional.

macdems:
I have made the specific revisions. As for coverage, it will be hard to increase it more. Not covered paths are mostly error handling fragments, some of them depend on time.

tardyp:
I have just tried to talk to the testbot, but it didn't answer anything.

macdems:
> I have just tried to talk to the testbot, but it didn't answer anything.

I was down (the whole buildbot instance). I run it on my desktop PC and I had to restart it today morning.

master/buildbot/reporters/telegram.py
p12tic: Would it be possible to show multiple messages if a single message is overflown?

master/buildbot/reporters/telegram.py
p12tic: Thanks, sounds reasonable.

master/buildbot/reporters/telegram.py
p12tic: Sounds good!

p12tic:
Looks good, except for coverage. I think we need to cover most of the error and edge-case handling too, because if the code is not covered by tests it may as well not exist from the perspective of the project maintainer. Such code may easily bitrot e.g. after telegram makes an API change and the code is updated and reviewed by people who don't know the codebase well. This situation is likely in the case of `BuildBot`.

For having precise control of the time you can use `TestReactor`. Search for `TestReactorMixin` in the test sources.

master/buildbot/reporters/telegram.py
macdems: Done. However, listing all changes can be problematic as there can be thousands of them. I have added additional confirmation.

p12tic: Sounds great, thanks!

p12tic:
@tardyp: Ok to merge?

master/buildbot/reporters/irc.py
```diff
@@ -231,10 +237,10 @@ def groupDescribe(self, channel, action):
 
     def getContact(self, user, channel=None):
         # nicknames and channel names are case insensitive
```
tardyp: Is this the case also for telegram? could this be an attack to be able to fool the bot for greater rights?

macdems: `getContact` in telegram accepts dict and retrieves numeric id for user and chat (both cannot be faked as they are set inernally by telegram). So this is non-issue here, as one gives numbers to `authz` for telegram. You need either to give our bot the command `/getid` or talk to https://t.me/myidbot to get the user id number. This is even more secure than in case or IRC, as one cannot get the numeric id of another user easily.

The only important thing is to keep the webhook resource path secret, as this is the only way of ensuring that the updates really come from telegram. I believe that Twisted does not provide any way to reveal the resource paths (I would be stupid if it did, but I am not a Twisted expert).

macdems:
Some screenshots, maybe for use on the webpage...

telegram1

telegram2

master/buildbot/reporters/irc.py
tardyp: ah I though this was in the command code. I indeed saw the /getid command.

Resource path as a secret is okay as long as there is https frontend.

master/docs/manual/configuration/reporters.rst
```diff
@@ -785,6 +787,236 @@ Two additional arguments can be set to control how fast the IRC bot tries to rec
...
+    A content of the SSL certificate of a certificate file open in binary mode. This is necessary if the access to the Buildbot web interface is through HTTPS protocol with self-signed certificate and ``userPolling`` is not set to ``True``.
+
+
+``usePolling``
+    (optional, disabled by default)
+    By default this bot receives messages from Telegram through web-hooks. However, this requires the web frontend of the Buildbot to be configured and accessible through HTTPS (not HTTP) on a public IP and port number 443, 80, 88, or 8443. Furthermore, the Buildbot configuration option :bb:cfg:`buildbotURL` must be correctly set. If this is not possible, :bb:reporter:`TelegramBot` can use polling to receive messages from Telegram. Set ``usePolling`` to ``True`` to do so.
+
+``pollTimeout``
+    (optional)
+    The default time the bot should wait for Telegram to respond to polling using `long polling <https://en.wikipedia.org/wiki/Push_technology#Long_polling>`.
```
tardyp: Franckly, long polling is good tech and works well for this kind of stuff. 
For me It is better than webhook, as simpler to setup.
I have lots of experience with Amazon SQS, which works that way, and never had issue with msg lost despite actively looking for them.
Polling tends to have bad reputation, but in this case, I think it is better.

Having two way of connecting sounds like overkill for maintainance, we should choose one.
I would choose long polling for simplicity and efficiency.

I do realize there is quite amount of quality code for web hook mode that I ask you to throw away :-(

master/docs/manual/configuration/reporters.rst
```diff
@@ -785,6 +787,236 @@ Two additional arguments can be set to control how fast the IRC bot tries to rec
...
+
+2. Type ``/newbot``.
+
+3. Enter a display name for your bot. It can be any string.
+
+4. Enter a unique username for your bot. Usernames are 5-32 characters long and are case insensitive, but may only include Latin characters, numbers, and underscores. Your bot's username must end in `bot`, e.g. `MyBuildBot` or `MyBuildbotBot`.
+
+5. You will be presented with a token for your bot. Save it, as you will need it for :bb:reporter:`TelegramBot` configuration.
+
+6. Optionally, you may type ``/setcommands``, select the username of your new bot and paste the following text:
```
tardyp: Couldn't this be done more automatically?

I would prefer this list to be only generated by the bot rather than hardcoded in the doc.

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1045 @@
...
+                    return
+            except AssertionError as err:
+                # just for tests
+                raise err
+            except Exception as err:
+                msg = "ERROR: cannot send Telegram request {} (will try again): {}".format(path, err)
+                if logme:
+                    self.log(msg)
+                    logme = False
+                yield task.deferLater(self.reactor, self.retry_delay, lambda: None)
```
tardyp: There is asyncsleep in buildbot.util

http://docs.buildbot.net/current/developer/utils.html#buildbot.util.asyncSleep

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1045 @@
...
+                n = message[:4096].rfind('\n')
+                n = n + 1 if n != -1 else 4096
+                params['text'], message = message[:n].rstrip(), message[n:].lstrip()
+
+            if not message and reply_markup is not None:
+                params['reply_markup'] = reply_markup
+
+            params.update(kwargs)
+
+            result = yield self._post('/sendMessage', json=params)
```
tardyp: why not until success for that one?

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1045 @@
...
+                    raise ValueError("[{}] {}".format(res.code, ans.get('description')))
+                updates = ans.get('result')
+            except AssertionError as err:
+                raise err
+            except Exception as err:
+                msg = "ERROR: cannot send Telegram request /getUpdates (will try again): {}".format(err)
+                if logme:
+                    self.log(msg)
+                    logme = False
+                yield task.deferLater(self.reactor, self.retry_delay, lambda: None)
```
tardyp: asyncsleep

master/docs/manual/configuration/reporters.rst
macdems: Can we add the result of some code to docs? Right now, the bot command `/commands botfather` can print the list to copy-paste to Bot Father. Telegram provides no way to do it automatically by bot.

master/buildbot/reporters/telegram.py
```diff
@@ -0,0 +1,1045 @@
...
+                    certificate=None, usePolling=False,
+                    pollTimeout=120, retryDelay=30):
+        super().checkConfig(self.name)
+
+        if authz is not None:
+            for acl in authz.values():
+                if not isinstance(acl, (list, tuple, bool)):
+                    config.error("authz values must be bool or a list of user ids")
+
+        if isinstance(certificate, io.TextIOBase):
```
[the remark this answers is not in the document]
macdems: I will limit it to the certificate text. One can open an file and read it in `master.cfg`...

master/docs/manual/configuration/reporters.rst
p12tic: It does not seem that there's that much additional code to maintain. I think that because the code is already written, has few chance to impact the codebase negatively and has tests, we should just accept it.

tardyp: Well this also adds a precedent, people could think it is a good thing to implement both.
 
We have other support which have both hook or poller. but in all those case, the poller is very inefficient. Here, this is not the case

macdems: Frankly, I would rather keep the code for web-hook along long polling. There is a number of reasons for this:

1. If you take a look at telegram bot frameworks in the web, there is a tendency to use web-hooks in productions servers: more and more bots are using this approach and it is always better go go with the stream. Also official telegram docs specify that there are advantages of using web-hooks instead of polling.

2. The interface code for both polling and web-hook is quite thin, so it should not be too hard to maintain.

3. It is good to keep the web-hook code (maybe encapsulated and better separated from the telegram-specific stuff) as it will make straightforward to add more bots. The two most obvious that come to my mind are Messenger and Slack.

tardyp: ok

master/docs/manual/configuration/reporters.rst
tardyp: We are already generating some of the doc:

https://github.com/buildbot/buildbot/blob/cc304b5a4e656200853a41ba8abc6e3b3a58cb57/master/docs/developer/rest.rst#L188-L200

https://pypi.org/project/sphinx-jinja/

https://github.com/buildbot/buildbot/blob/cc304b5a4e656200853a41ba8abc6e3b3a58cb57/master/docs/conf.py#L323-L326

master/buildbot/reporters/telegram.py
macdems: Actually there is no reason `_post` and `_post_until_success` differ...

master/buildbot/reporters/telegram.py
macdems: I don't know how to easily use `asyncSleep` with `TestReactor`... for unit test.

master/docs/manual/configuration/reporters.rst
macdems: I didn't know, you can use Jinja in Sphinx. That's neat.

master/buildbot/reporters/telegram.py
macdems: Ok, I think I have managed to test it...

master/docs/manual/configuration/reporters.rst
macdems: I have refactored web-hook code a bit and moved it to words. It should be easier to re-use it in the future...

master/buildbot/reporters/words.py
```diff
@@ -1257,3 +1254,60 @@ def clientConnectionLost(self, connector, reason):
...
+
+    def __init__(self, path):
+        resource.Resource.__init__(self)
+        www = get_plugins('www', None, load_now=True)
+        if 'base' not in www:
+            raise RuntimeError("could not find buildbot-www; is it installed?")
+        self._root = www.get('base').resource
+        self.path = path
+
+    def startService(self):
```
tardyp: I think this will be cleaner to implement that as a child service. Mixins are supposed to not override any method of the parent class (and not have an __init__) to avoid super() ordering hell.
Mixin can only inherit from object(). You can create a WebhookBotMixin, but he has to maintain its resource as a child service.

If we want to refactor this for reuse, we need to make it right.

PR #5023, opened 2019-09-30
GitCommit: allow to define behavior when there's nothing to be committed

ffdybuster:
This adds a new option `emptyCommits` to the `GitCommit` build step, which allows to determine the behavior when there's nothing to commit. Default behavior is `disallow`, which fails (current behavior). One can also specify `create-empty-commit` to create empty commits, and `ignore` to not create a commit.

I've also tested both `disallow` and `ignore` in a real environment, both worked as expected (both with and without commits).

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -689,7 +689,7 @@ class GitCommit(buildstep.BuildStep, GitStepMixin, CompositeStepMixin):
     renderables = ['paths', 'messages']
 
     def __init__(self, workdir=None, paths=None, messages=None, env=None,
-                 timeout=20 * 60, logEnviron=True,
+                 timeout=20 * 60, logEnviron=True, emptyCommits='disallow',
                  config=None, **kwargs):
 
         self.workdir = workdir
@@ -699,6 +699,7 @@ class GitCommit(buildstep.BuildStep, GitStepMixin, CompositeStepMixin):
         self.timeout = timeout
         self.logEnviron = logEnviron
         self.config = config
+        self.emptyCommits = emptyCommits
         # The repourl, sshPrivateKey and sshHostKey attributes are required by
         # GitStepMixin, but aren't needed by git add and commit operations
         self.repourl = " "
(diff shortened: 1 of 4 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -752,8 +758,19 @@ def _doCommit(self):
...
 
-        ret = yield self._dovccmd(cmd)
-        return ret
+        if self.emptyCommits == 'create-empty-commit':
+            cmd.extend(['--allow-empty'])
+
+        if self.emptyCommits == 'ignore':
+            stdout, rc = yield self._dovccmd(cmd, collectStdoutAndRC=True, decodeRC={0: SUCCESS, 1: SUCCESS})
+            if rc == 0:
+                return 0
```
p12tic: Looks like there's no unit test covering this code path.

master/buildbot/steps/source/git.py
```diff
@@ -752,8 +758,19 @@ def _doCommit(self):
...
+            if rc == 1 and ("nothing added to commit" in stdout or "nothing to commit" in stdout):
+                return 0
```
p12tic: Looks like there's no unit test covering this code path.

master/buildbot/steps/source/git.py
ffdybuster: Thanks! Added a test for this one.

master/buildbot/steps/source/git.py
ffdybuster: Also for this one.

master/buildbot/steps/source/git.py
```diff
@@ -752,6 +758,17 @@ def _doCommit(self):
...
             cmd.extend(['-m', message])
 
+        if self.emptyCommits == 'create-empty-commit':
+            cmd.extend(['--allow-empty'])
+
+        if self.emptyCommits == 'ignore':
+            stdout, rc = yield self._dovccmd(cmd, collectStdoutAndRC=True, decodeRC={0: SUCCESS, 1: SUCCESS})
+            if rc == 0:
+                return 0
+            if rc == 1 and ("nothing added to commit" in stdout or "nothing to commit" in stdout):
```
tardyp: what if lang is french? I would prefer to test  first with a git status --porcelain

ffdybuster: Hmmm. Bad luck? :) I've tried to implement this in 3e46d7814c8aa73a775a67d4b2460cbe7f142fa2.

ffdybuster:
Looks like the tests pass. I've also tested `ignore` in a real environment (once with no changes, and once with a file modification), and it worked. If the current approach is OK, I'll rebase to remove the changes in `util/git.py` (no longer needed) and clean up the commit list a bit if that's ok.

ffdybuster:
@tardyp thanks! I removed the two commits which modify `util/git.py` and squashed the remaining ones (with a new shiny commit message).

felixfontein:
@p12tic @tardyp thanks a lot for reviewing and merging! :)

PR #5026, opened 2019-10-01
Docs: remove invisible unicode characters

ffdybuster:
I noticed that in the first "Note" in https://docs.buildbot.net/latest/developer/pull-request.html#how-to-create-a-pull-request, the link is broken. When trying to fix it, I noticed that this is caused by some invisible Unicode character before the first backtick. I removed it, and then found some more of these and removed them as well. (I also checked the other buildbot and worker docs, but couldn't find it anywhere else.)

### DIFF
--- a/master/docs/developer/pull-request.rst
+++ b/master/docs/developer/pull-request.rst
@@ -3,7 +3,7 @@ Submitting Pull Requests
 
 As Buildbot is used by software developers, it tends to receive a significant number of patches.
 The most effective way to make sure your patch gets noticed and merged is to submit it via GitHub.
-This assumes some familiarity with git, but not too much. Note that GitHub has some great `​Git guides <http://github.com/guides>`_ to get you started.
+This assumes some familiarity with git, but not too much. Note that GitHub has some great `Git guides <http://github.com/guides>`_ to get you started.
 
 Guidelines
 ----------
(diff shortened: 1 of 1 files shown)


tardyp:
Sounds scary that you can fit in invisible characters. I can't even review them!

tardyp:
Looks like this is http://unicode.scarfboy.com/?s=E2808B

tardyp:
There are also instances of http://unicode.scarfboy.com/?s=E2809D in this file..

tardyp:
Looks like git blame shows them.

image

and @p12tic is the culprit :)

tardyp:
vim and less show them but not vscode, not github..

ffdybuster:
> There are also instances of http://unicode.scarfboy.com/?s=E2809D in this file..

Yes, but these are not invisible :)

ffdybuster:
> Sounds scary that you can fit in invisible characters. I can't even review them!

Probably the best way is to inspect the patch with a hex editor or some other program which shows such special characters.

tardyp:
> Probably the best way is to inspect the patch with a hex editor or some other program which shows such special characters.

That's what I did, but I can't do that for every patch!

ffdybuster:
@tardyp thanks for reviewing and merging!

For avoiding such problems in the future: how about adding a step to the CI which checks the doc files for unwanted characters (like Zero-Width Space) and fails if they are found?

p12tic:
I copied the original content from the BuildBot wiki on trac. So it's likely that the invisible characters come from there.

Issue #5027, opened 2019-10-02
Assertion error "assert not isinstance(result, Deferred)" when I add a new builder and reconfigure


evpo:
I added a new build and called `buildbot reconfig master`.

My configuration ran successfully. Then, immediately after that Buildbot reconfigures workers. Since I added a new Builder it probably tries to contact the corresponding Worker. This is when the assertion fails, I believe. Buildbot continues working and all other builds work fine. However, all build requests for the new build get stuck in the Pending Buildrequests queue forever. Even though the worker is connected and functions.

If I restart the worker in question, it is attached to the Builder and all build requests start to build.

Versions:
```
[["Python","3.7.3"],["Buildbot","2.3.1"],["Twisted","19.7.0"]]
```

The exception extract from the log:
```
2019-10-02 16:50:40+1000 [Broker,2,172.17.0.1] during reconfig:

Traceback (most recent call last):
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
current.result = callback(current.result, *args, **kw)
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/twisted/internet/defer.py", line 1475, in gotResult
_inlineCallbacks(r, g, status)
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
result = result.throwExceptionIntoGenerator(g)
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/twisted/python/failure.py", line 512, in throwExceptionIntoGenerator
return g.throw(self.type, self.value, self.tb)
---  ---
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/buildbot/master.py", line 397, in doReconfig
yield self.reconfigServiceWithBuildbotConfig(new_config)
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
result = result.throwExceptionIntoGenerator(g)
File "/home/mk7/ve/bb-master/sandbox/lib64/python3.7/site-packages/twisted/python/failure.py", line 512, in throwExceptionIntoGenerator
return g.throw(self.type, self.value, self.tb)
[...]

Issue #4757, opened 2019-05-07
Exceptions when adding new workers to builder using sighup; -> worker does not get jobs from builder


YngveNPettersen:
Buildbot 2.2.0

I added new workers to some of our builders. The builders were already active in the system and used for other builders. They were added to the new builders due a short term need for more workers for those builders, and doing a full restart for the master is out of the question in this scenario.

When running sighup, I frequently got the exception below for various builders

When the sighup operation did succeed, only one of the three workers accepted scheduled builds from the builder they had been added to; and that was AFAICT a build on the builder that did not trigger any exceptions. I tried removing one of the workers from the builder, sighup, readding it, and sighup; got exceptions, then succeeded, but builds does not start.

Please note that this is not the first case we have had of exceptions breaking stuff during sighup, see https://github.com/buildbot/buildbot/issues/4673 and https://github.com/buildbot/buildbot/issues/4715 for the others.

```
2019-05-07 10:51:03+0000 [Broker (TLSMemoryBIOProtocol),19,1.2.3.4] during reconfig:
        Traceback (most recent call last):
          File "sandbox/lib/python3.6/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
            current.result = callback(current.result, *args, **kw)
          File "sandbox/lib/python3.6/site-packages/twisted/internet/defer.py", line 1475, in gotResult
            _inlineCallbacks(r, g, status)
          File "sandbox/lib/python3.6/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
            result = result.throwExceptionIntoGenerator(g)
          File "sandbox/lib/python3.6/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
        --- <exception caught here> ---
          File "sandbox/lib/python3.6/site-packages/buildbot/master.py", line 393, in doReconfig
            yield self.reconfigServiceWithBuildbotConfig(new_config)
          File "sandbox/lib/python3.6/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
            result = result.throwExceptionIntoGenerator(g)
          File "sandbox/lib/python3.6/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
            return g.throw(self.type, self.value, self.tb)
          File "sandbox/lib/python3.6/site-packages/buildbot/util/service.py", line 49, in reconfigServiceWithBuildbotConfig
[...]

PR #5029, opened 2019-10-03
buider.attached returns Deferred and causes assert failure

evpo:
Functions decorated with inlineCallbacks cause a failed assert when they
return Deferred: "assert not isinstance(result,Deferred)".


evpo:
This PR is to address #5027

evpo:
and #4757

evpo:
Agreed. A unit test would be nice. I am on tight schedule at the moment but I will see if I can find some time for it.

p12tic:
I can write the test as in this case the test involves complex internals of Buildbot and I'm responsible for the bug anyway.

PR #5032, opened 2019-10-08
Fix git progress

rajdeepbharati:
See https://github.com/buildbot/buildbot/pull/4857#discussion_r296477466
## Contributor Checklist:


master/buildbot/steps/source/git.py
```diff
@@ -388,7 +386,10 @@ def _clone(self, shallowClone):
         command += [self.repourl, '.']
 
         if self.prog:
-            command.append('--progress')
+            if self.supportsProgress:
+                command.append('--progress')
+            else:
+                print("Git versions < 1.7.2 don't support progress")
```
p12tic: @rajdeepbharati: I think we should use `log.msg` here.

rajdeepbharati: Oops... I was wondering what to use

Issue #4801, opened 2019-05-30
Grid view tag filtering bug


rajdeepbharati:
Builders with selected tags are highlighted, but all the builders are displayed on the grid.

```
Error: "branch" is read-only
refresh@http://localhost:8010/grid_view/scripts.js:676:5
toggleTag@http://localhost:8010/grid_view/scripts.js:667:17
fn@http://localhost:8010/scripts.js?_1559194180589 line 40425 > Function:4:346
expensiveCheckFn@http://localhost:8010/scripts.js?_1559194180589:41438:18
callback@http://localhost:8010/scripts.js?_1559194180589:50965:17
$eval@http://localhost:8010/scripts.js?_1559194180589:43182:28
$apply@http://localhost:8010/scripts.js?_1559194180589:43282:25
ngEventHandler/<@http://localhost:8010/scripts.js?_1559194180589:50970:23
dispatch@http://localhost:8010/scripts.js?_1559194180589:149296:27
add/elemData.handle@http://localhost:8010/scripts.js?_1559194180589:149103:28
```

rajdeepbharati:
@mojca are you also facing this issue in master?

rjarry:
I have the same problem, I have a more exploitable backtrace:

```
angular.js:15567 TypeError: Cannot assign to read only property 'branch' of object '#<Object>'
    at Grid.refresh (grid.controller.js:258)
    at Grid.changeResult (grid.controller.js:239)
    at fn (eval at compile (angular.js:16418), <anonymous>:4:255)
    at l.$eval (angular.js:19393)
    at angular.js:27827
    at Object.<anonymous> (angular.js:30812)
    at k (angular.js:387)
    at Object.$$writeModelToScope (angular.js:30810)
    at o (angular.js:30803)
    at angular.js:30797
```

The problem occurs here: https://github.com/buildbot/buildbot/blob/v2.4.1/www/grid_view/src/module/grid.controller.js#L258

PR #5039, opened 2019-10-17
grid-view: fix branch/status filtering error

rjarry:
Fix the following error when filtering by branch/results and/or builder tags:

```
angular.js:15567 TypeError: Cannot assign to read only property 'branch' of object '#<Object>'
    at Grid.refresh (grid.controller.js:258)
    at Grid.changeResult (grid.controller.js:239)
    at fn (eval at compile (angular.js:16418), <anonymous>:4:255)
    at l.$eval (angular.js:19393)
    at angular.js:27827
    at Object.<anonymous> (angular.js:30812)
    at k (angular.js:387)
    at Object.$$writeModelToScope (angular.js:30810)
    at o (angular.js:30803)
    at angular.js:30797
```

There is no need to modify the `$stateParams` object anyway. Affect the values in the local `params` object directly.

Fixes: #4801
Fixes: f098bdd5d7fb ("www: add grid_view plugin")

### DIFF
--- a/www/grid_view/src/module/grid.controller.js
+++ b/www/grid_view/src/module/grid.controller.js
@@ -255,18 +255,10 @@ class Grid {
     }
 
     refresh() {
-        this.$stateParams.branch = this.branch;
-        if (this.tags.length === 0) {
-            this.$stateParams.tag = undefined;
-        } else {
-            this.$stateParams.tag = this.tags;
-        }
-        this.$stateParams.result = this.result;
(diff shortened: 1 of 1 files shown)


rjarry:
I forgot the news fragment but this is only a draft.

If people have some insight about angular internals and why I $stateParams has become read-only that would be great.

tardyp:
thank you so much!

PR #5041, opened 2019-10-17
Custom templates: loading fixed #5035

### LINKED ISSUE #5035: Custom templates still not working
Hi, this is the original issue with broken custom templates #4980

But it doesn't work even after the fix (2.4.1).
The web part of Buildbot is far to complicated for me. But I was able to find lines like this in scripts.js?_1568233606304
```
    , function(e, t) {
        e.exports = window.T["undefined/properties.html"] || '<table class="table table-hover...
    }
```
And I presume there is something wrong if there is "**undefined**/properties.html".

vibbo:
- master/buildbot/www/config.py:parseCustomTemplateDir
    Returns pure HTML. Conversion to JSON is done in
    www/base/src/app/layout.jade
- www/base/package.json
    Missing plugin_name fixed
- www/build_common/src/ng-template-loader.js
    Fixed missing "views/" part of template name

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_www_config.py
+++ b/master/buildbot/test/unit/test_www_config.py
@@ -78,14 +78,14 @@ class IndexResource(TestReactorMixin, www.WwwTestMixin, unittest.TestCase):
         self.assertEqual(res, exp)
 
     def test_parseCustomTemplateDir(self):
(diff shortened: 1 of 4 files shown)


p12tic:
Thanks a lot for looking into this!

Issue #5042, opened 2019-10-17
p4 issues


bschatt:
We had a lot of issues with p4poller recently. Most of them appeared after we enabled ticket-based auth on our p4 server.

1. (moved to #5148)

2. We are currently running in a random issue. Sometimes we get a weird error forever. I think this is due to the password parameter but not sure. I will let you know if I find anything.

> 2019-10-17 17:49:41+0200 [-] P4 poll failed on XXX.XXX.XXX.XXX:1666, //XXXXXXXXX/
> 	Traceback (most recent call last):
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 151, in maybeDeferred
> 	    result = f(*args, **kw)
> 	  File "/home/buildbot/.local/lib/python3.7/site-packages/buildbot/changes/p4poller.py", line 175, in poll
> 	    d = self._poll()
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 1613, in unwindGenerator
> 	    return _cancellableInlineCallbacks(gen)
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
> 	    _inlineCallbacks(None, g, status)
> 	--- <exception caught here> ---
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
> 	    result = g.send(result)
> 	  File "/home/buildbot/.local/lib/python3.7/site-packages/buildbot/changes/p4poller.py", line 245, in _poll
> 	    result = yield self._get_process_output(args)
> 	  File "/home/buildbot/.local/lib/python3.7/site-packages/buildbot/changes/p4poller.py", line 183, in _get_process_output
> 	    d = utils.getProcessOutput(self.p4bin, args, env)
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/utils.py", line 128, in getProcessOutput
> 	    reactor)
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/utils.py", line 28, in _callProtocolWithDeferred
> 	    reactor.spawnProcess(p, executable, (executable,)+tuple(args), env, path)
> 	  File "/usr/lib/python3/dist-packages/twisted/internet/posixbase.py", line 331, in spawnProcess
> 	    args, env = self._checkProcessArgs(args, env)
[...]

PR #5045, opened 2019-10-17
Fix string encoding issue in p4poller

### LINKED ISSUE #5148: P4: bytes encoding error in TicketLoginProtocol
Extracted out of issue #5042 by @bschatt:

I don't recall the exact error message but I had some string encoding issues with the 3 variables (stdin, stderr and stdout) in the TicketLoginProtocol class. I made an easy fix and I can submit a PR if needed:

```
self.stdin = stdin.encode('ascii')
self.stdout = b''
self.stderr = b''
```

bschatt:
~Fixes the first issue reported in #5042 .~

Fixes #5148 (extracted part of #5042 to separate issue).


p12tic:
@bschatt: Thanks for your contribution and sorry for the long delay. No one has reviewed the PR because the tests were broken and there were no active developers using P4.

Regarding the PR, it looks like that the `TicketLoginProtocol` shouldn't have ever used `str`, because according to the documentation of Twisted, the process protocol can only get bytes. I'm not sure how this ever worked, maybe the behavior is more forgiving on python2, but I've not checked this carefully.

PR #5047, opened 2019-10-18
www: build: Make URLs in steps always visible

vibbo:
I think the following change started to hide (other) URLs by mistake.
https://github.com/buildbot/buildbot/commit/056301f21b67c700c2cf501eb849d3809700f1e6#diff-2d366bdd85ca12602b61b8a11ab99251R47-R58

In any case, would you consider using this fix, so (other) URLs are always visible?

## Contributor Checklist:

### DIFF
--- a/www/base/src/app/builders/builds/build.route.js
+++ b/www/base/src/app/builders/builds/build.route.js
@@ -39,13 +39,19 @@ class BuildState {
             }
             ]});
         bbSettingsServiceProvider.addSettingsGroup({
-            name:'TriggerStep',
-            caption: 'Builds per page',
+            name:'Build',
+            caption: 'Build page related settings',
             items:[{
                 type:'integer',
-                name:'page_size',
+                name:'trigger_step_page_size',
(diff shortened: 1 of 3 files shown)


tardyp:
Well, this was actually on purpose to have a cleaner step UI. but if people agree that the prefer to always see the urls..

vibbo:
> Well, this was actually on purpose to have a cleaner step UI. but if people agree that the prefer to always see the urls..

I agree with hiding triggered builds. But URLs added by step are quite useful to see all the time. At least for us.

p12tic:
@vibbo: Could you show a screenshots of how UI looked before and after the PR?

vibbo:
Before change:
Step closed
1-5047-before-closed_step
Step opened
2-5047-before-open_step

After change:
Step closed
3-5047-after-closed_step
Step open
4-5047-after-open_step

Should I make it configurable via `c['www']['ui_default_config']` ?

tardyp:
Indeed, having a setting should close the debate..

PR #5054, opened 2019-10-22
Support Ubuntu Chromium

ffdybuster:
Currently, when using Ubuntu's Chromium (latest version) to visit a Buildbot instance, one gets a big fat "YOUR BROWSER IS OUT-OF-DATE!" warning. According to mikemaccana/outdated-browser-rework#49 this is caused by Ubuntu tweaking the User-Agent string, and so the detected browser name is "Chromium" and not "Chrome". Adding "Chromium" to the supported browser list with the same version as for Chrome should fix this problem.

### DIFF
--- a/master/buildbot/newsfragments/chromium-supported.bugfix
+++ b/master/buildbot/newsfragments/chromium-supported.bugfix
@@ -1,0 +1,1 @@
+Updates supported browser list so that Ubuntu Chromium will not always be flagged as out of date.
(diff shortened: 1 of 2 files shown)


ffdybuster:
I verified that this works by patching `/usr/local/lib/python3.7/dist-packages/buildbot_www/static/browser-warning-list.js` directly for our local Buildbot instance.

ffdybuster:
@tardyp thanks for merging!

PR #5057, opened 2019-10-23
schedulers: More robust Timed scheduler

vibbo:
Handling startBuild and _scheduleNextBuild_locked separately during
actuation. This prevents errors in startBuild function to derail
timed scheduler because _scheduleNextBuild_locked is now called
even in this case.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/schedulers/timed.py
+++ b/master/buildbot/schedulers/timed.py
@@ -206,6 +206,7 @@ class Timed(AbsoluteSourceStampsMixin, base.BaseScheduler):
         "Similar to util.now, but patchable by tests"
         return util.now(self._reactor)
 
+    @defer.inlineCallbacks
     def _scheduleNextBuild_locked(self):
         # clear out the existing timer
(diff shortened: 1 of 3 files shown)


vibbo:
Thanks for the advice. I have updated the PR.

master/buildbot/schedulers/timed.py
```diff
@@ -206,30 +206,31 @@ def now(self):
...
 
-        # set up the new timer
-        @d.addCallback
-        def set_timer(actuateAt):
+        if actuateAt is None:
+            self.actuateAt = None
+        else:
+            # set up the new timer
             now = self.now()
             self.actuateAt = max(actuateAt, now)
```
vibbo: Note: I have also modified this part, where getNextBuildTime can return None (according to it's docs) and it would be passed to max fce. This would result in error.

vibbo:
@tardyp Could you review this changes?

PR #5065, opened 2019-11-08
worker: add --useTls option

aparcar:
The options sets the connection type within the `connection_string` to
`tls` instead of `tcp`. This allows to use an encrypted connection via
the `buildbot-worker --create-worker` command, instead of requiring to
edit the `buildbot.tac` afterwards.

Signed-off-by: Paul Spooren <mail@aparcar.org>

I'll create a unit test once this approach is approved.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1543,6 +1543,7 @@ usepty
 usererrors
 username
 usetestcasenames
+usetls
 usr
 utc
 utf
--- a/master/buildbot/newsfragments/add-useTLS-arg.feature
+++ b/master/buildbot/newsfragments/add-useTLS-arg.feature
@@ -1,0 +1,1 @@
+Add the parameter --useTLS to `buildbot-worker create-worker` to automatically enable TLS in the connection string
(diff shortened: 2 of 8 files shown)


aparcar:
Please somebody look into this. It's a super simple path but highly useful.

tardyp:
Hi,
The code looks okay, but the test do not pass, could you please fix the tests?

aparcar:
@tardyp Hi, are all these tests supposed to be green? I can hardly imagine that my change breaks so much stuff...

tardyp:
Hi @aparcar,
When we change internal settings, the unit tests might indeed break and need to be adapted.
Also, we require any contributor adding a feature to integrate tests for its feature so that we remember that it exists and make sure it doesn't break on further contributions.
This is the key to have Buildbot working in long term.

tardyp:
Please have a look at the failures. Some are due to unknown new words in the white list 
please add in:
common/code_spelling_ignore_words.txt
master/docs/spelling_wordlist.txt

Other are unit tests that need to be updated

p12tic:
Fixed the worker template and updated the tests. @aparcar: Thanks for your contribution!

PR #5079, opened 2019-11-21
www-badges: Attempt to improve rendering.

StevenAWhite:
added Parameters for left and right side pad.
added parameter for border radius
updated documentation


StevenAWhite:
Might need some help understanding what is wrong with the appveyor python35 environment.

PR #5092, opened 2019-11-26
Buildstep should release all _acquiringLocks when cancelled to avoid deadlock

MichaelOultram:
## Contributor Checklist:

Found in a live buildbot deployment, and I set out to write a unit test to prove it. It looks like the `buildstep` code could cause a deadlock if it requires multiple locks, and it gets cancelled at just the wrong time. My change keeps track of all the `_acquiringLocks` and makes sure to either `claim` and `release` (if the lock is available to us), or `stopWaitingUntilAvailable` if the lock is not available to us.

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -355,7 +355,7 @@ class BuildStep(results.ResultComputingConfigMixin,
             config.error("BuildStep updateBuildSummaryPolicy must be "
                          "a list of result ids or boolean but it is %r" %
                          (self.updateBuildSummaryPolicy,))
-        self._acquiringLock = None
+        self._acquiringLocks = []
         self.stopped = False
         self.master = None
         self.statistics = {}
@@ -637,7 +637,6 @@ class BuildStep(results.ResultComputingConfigMixin,
         return ok
 
     def acquireLocks(self, res=None):
-        self._acquiringLock = None
         if not self.locks:
             return defer.succeed(None)
         if self.stopped:
@@ -648,12 +647,13 @@ class BuildStep(results.ResultComputingConfigMixin,
                 self._waitingForLocks = True
                 log.msg("step %s waiting for lock %s" % (self, lock))
                 d = lock.waitUntilMaybeAvailable(self, access)
+                self._acquiringLocks.append((lock, access, d))
                 d.addCallback(self.acquireLocks)
-                self._acquiringLock = (lock, access, d)
                 return d
         # all locks are available, claim them all
         for lock, access in self.locks:
             lock.claim(self, access)
+        self._acquiringLocks = []
         self._waitingForLocks = False
         return defer.succeed(None)
 
@@ -744,9 +744,10 @@ class BuildStep(results.ResultComputingConfigMixin,
 
     def interrupt(self, reason):
         self.stopped = True
-        if self._acquiringLock:
-            lock, access, d = self._acquiringLock
-            lock.stopWaitingUntilAvailable(self, access, d)
+        if self._acquiringLocks:
+            for (lock, access, d) in self._acquiringLocks:
+                lock.stopWaitingUntilAvailable(self, access, d)
+            self._acquiringLocks = []
 
         if self._waitingForLocks:
             self.addCompleteLog(
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -643,6 +643,10 @@ class BuildStep(results.ResultComputingConfigMixin,
             return defer.succeed(None)
         log.msg("acquireLocks(step %s, locks %s)" % (self, self.locks))
         for lock, access in self.locks:
+            for waited_lock, _, _ in self._acquiringLocks:
+                if lock is waited_lock:
+                    continue
+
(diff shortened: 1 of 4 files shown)


master/buildbot/process/buildstep.py
```diff
@@ -649,11 +648,12 @@ def acquireLocks(self, res=None):
                 log.msg("step %s waiting for lock %s" % (self, lock))
                 d = lock.waitUntilMaybeAvailable(self, access)
                 d.addCallback(self.acquireLocks)
-                self._acquiringLock = (lock, access, d)
+                self._acquiringLocks.append((lock, access, d))
                 return d
```
tardyp: As we return just after, I would expect this list to always say at maximum one lock.

We could have a race if waitUnrilMaybeAvailable returns succeed synchronously, then addCallback would recall acquirelocks synchronously, so _acquiringLock will be reset and restarted to lock, access, d just after, so I would switch line 651 and 652

MichaelOultram: > As we return just after, I would expect this list to always say at maximum one lock.

Example:
The build step has two locks: `lock1` and `lock2` and neither lock is available yet. The `acquireLocks` code will `waitUntilMaybeAvailable` on `lock1`. When `lock1` becomes available, it will call the `acquireLocks` method again, and then it will have to `waitUntilMaybeAvailable` on `lock2`. At this point, we would have two items in the `self._acquiringLocks` list.

If we were to cancel this build at this point, we must clean up both `lock1` and `lock2`. The existing master code would only clean up `lock2`, so our (cancelled) build will stay in the waiters of `lock1` forever.

>We could have a race if waitUnrilMaybeAvailable returns succeed synchronously, then addCallback would recall acquirelocks synchronously, so _acquiringLock will be reset and restarted to lock, access, d just after, so I would switch line 651 and 652

Good spot, I will switch those two lines

master/buildbot/test/unit/test_process_buildstep.py
```diff
@@ -200,6 +200,87 @@ def test_regularLocks(self):
         self.expectOutcome(result=SUCCESS)
         yield self.runStep()
 
+    @defer.inlineCallbacks
```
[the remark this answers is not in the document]
MichaelOultram: Yes. Without my changes to buildstep.py, this assert will trigger:
https://github.com/buildbot/buildbot/blob/3e483d80468eee39012cc1381305b3edd18f16bd/master/buildbot/test/unit/test_process_buildstep.py#L263
At this point we have cancelled stepc, so why is lock1 available for this step? To prove further that it is something actually wrong here, I commented out all the asserts for my test. My test will still not pass, as it will become stuck at https://github.com/buildbot/buildbot/blob/3e483d80468eee39012cc1381305b3edd18f16bd/master/buildbot/test/unit/test_process_buildstep.py#L278
meaning that `stepd` never manages to `acquireLocks` (even though both locks are now available). With the changes to buildstep.py, it obviously passes.

master/buildbot/process/buildstep.py
```diff
@@ -744,10 +744,16 @@ def start(self):
...
         self.stopped = True
-        if self._acquiringLock:
-            lock, access, d = self._acquiringLock
-            lock.stopWaitingUntilAvailable(self, access, d)
-            d.callback(None)
+        if self._acquiringLocks:
+            for (lock, access, d) in self._acquiringLocks:
+                if lock.isAvailable(self, access):
+                    # If lock is available to us, we HAVE to claim and release
+                    lock.claim(self, access)
```
tardyp: why? this sounds like a workaround of a more subtle problem

master/buildbot/process/buildstep.py
tardyp: I'm sorry, but this explanation still does not match. At the begining of acquireLocks, we just reset the _acquiringLock variable, and we do nothing with it
acquireLocks() shall only be called once in start step, and then will be called again by waitUntilMaybeAvailable() deferred.

waitUntilMaybeAvailable() just waits, but does not actually lock the lock. That is done later with lock.claim()

This method is designed so that we check that all locks are available in a synchronous manner, and then we claim them all in a row.

master/buildbot/process/buildstep.py
p12tic: It's because we need to remove ourselves from the waiting list. Perhaps the comment could be clarified.

tardyp: Then we should rather add an API to remove ourselves from the waitingList in case of cancel.

tardyp:
superseeded by #5098

p12tic:
I didn't investigate this bug in #5098, so some part of this PR should land.

p12tic:
@MichaelOultram: Thanks a lot for your contribution, this looks like a really nasty bug.

@tardyp: I looked into ways to make the bugfix simplier, turns out it's the optimal solution. We have to track all locks we're waiting on and not remove ourselves out of the wait queues until we can acquire all locks. Otherwise we could end up in a dynamic deadlock situation, where multiple lock requesters form a dependency cycle with regards to their priorities in lock waiting lists.

p12tic:
Merging as my modifications didn't materially change the pull request.

PR #5094, opened 2019-11-27
Home Page Configurability for Front End

uglycoyote:
* [ n/a ? ] I have updated the unit tests
* [ X ] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)
* [ no ] I have updated the appropriate documentation

Please suggest a file where the documentation change should be made, if applicable.

PR #5095, opened 2019-12-02
Changes for home page configurability

uglycoyote:
* [ n/a? ] I have updated the unit tests
* [ X  ] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)
* [ X ] I have updated the appropriate documentation


uglycoyote:
@tardyp Same changes as #5094 but cleaner and including an update to the docs.

tardyp:
BW, this commit is done with your nl email address. As you never revealed it, maybe this is was mistake..

uglycoyote:
@tardyp please go ahead with the merge.  But thanks for bringing it to my attention.  We are in a bit of a grey area but looking back through the `git log` I see that several merged commits from the past have the same information, so it's not unprecedented.

PR #5098, opened 2019-12-08
locks: allow stopWaitingUntilAvailable() in more scenarios

p12tic:
This is an internal improvement needed for to remove a hack from #5092. 

The PR essentially implements two things:

 - no longer allows two concurrent waiters from the same object via `waitUntilMaybeAvailable()`. If that happens, the previous code would just lose the original Deferred and never execute its callback.

 - allow unwaiting a Deferred via `stopWaitingUntilAvailable()` regardless of when this is done as long as the lock is not claimed. Previously there was a race condition which would cause `stopWaitingUntilAvailable()` to fail if the Deferred has already been called, but the lock was not claimed by the called code yet.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/locks.py
+++ b/master/buildbot/locks.py
@@ -186,6 +186,9 @@ class BaseLock:
 
         A single requester must not have more than one pending waitUntilMaybeAvailable() on a
         single lock.
+
+        If the lock is available, the caller must claim it. When the caller is no longer interested
+        into the lock it must call stopWaitingUntilAvailable().
         """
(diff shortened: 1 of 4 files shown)


tardyp:
looks good. I think we should include the test from #5092 ?

p12tic:
I think issue in #5092 should be a separate PR. This PR improves only the low-level interface, no investigation of the deadlock in #5092 has been done. There's small chance that this PR may fix #5092 though.

PR #5101, opened 2019-12-12
MailNotifier: allow to dump formatted mails for debugging

ffdybuster:
Adds a new option `dumpMailsToLog` which allows to dump formatted mails to the logs before they are sent by `MailNotifier`. This is very useful to debug problems with your mail provider (e.g. to find out why they are rejecting your emails as spam).

## Contributor Checklist:

* [ ] I have updated the unit tests
* [x] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)
* [x] I have updated the appropriate documentation

PR #5102, opened 2019-12-12
MailNotifier: fix email To: and CC: header escaping

ffdybuster:
Currently, recipients are added to To: and CC: headers by concatenating the addresses with `", "`. If one of the addresses contains a non-ASCII character, this causes the whole header to be UTF8-escaped. Unfortunately, for To: and CC:, this seems not to be RFC compliant (according to our mail provider, who classified the mails as spam because of this).

This change makes sure that the addresses are parsed, and the comment for every address (i.e. the human-readable name) will be separately escaped. This solves the problem for me: if I change my git name locally to something with umlauts/accents and use it to add a broken commit, the tests run by our Buildbot instance fail and the resulting email's To: header is correctly escaped (checked with #5101) and the mail is not rejected by our mail provider.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/mail-recipients-formatting.bugfix
+++ b/master/buildbot/newsfragments/mail-recipients-formatting.bugfix
@@ -1,0 +1,1 @@
+Fix mail recipient formatting to make sure address comments are separately escaped instead of escaping the whole To: or CC: header, which is not RFC compliant.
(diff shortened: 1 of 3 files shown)


p12tic:
Rebased on master.

ffdybuster:
I've also rebased now that #5101 has been merged.

ffdybuster:
@p12tic also thanks a lot for handling this one!

PR #5105, opened 2019-12-13
www: Fix instability in forceandstop e2e tests

### LINKED ISSUE #5104: www e2e forceandstop test is flaky
The forceandstop www e2e test is flaky. Error message is:

```
  forceandstop
    ✗ should create a build with a dedicated reason and stop it during execution
      - Expected 'http://localhost:8011/#/buildrequests/3?redirect_to_build=true' to match '/builders/[1-9]/builds/[1-9]'.
```

See e.g. https://buildbot.buildbot.net/#/builders/53/builds/57/steps/8/logs/stdio.

The reason for the failure is obvious: the master is too slow to create a build from a buildrequest and the test does not wait for the build to be created, but checks the URL immediately.

p12tic:
This PR fixes an issue in the forceandstop e2e test which does not wait for the forced build to go to the build page. Previously, if the master is slow to create the build, the test might check the buildrequest page and break. Now we always wait to enter the build page before proceeding with subsequent tests.

Additionally, this PR fixes a couple of other issues that were uncovered by these changes.

Fixes #5104.

### DIFF
--- a/smokes/e2e/pages/force.ts
+++ b/smokes/e2e/pages/force.ts
@@ -49,6 +49,17 @@ export class ForcePage extends BasePage {
         await button.click();
     }
 
+    async clickStartButtonAndWaitRedirectToBuild() {
+        let previousUrl = await browser.getCurrentUrl();
+        await this.clickStartButton();
+        await browser.wait(EC.not(EC.urlIs(previousUrl)),
+                           5000,
+                           "failed to create a buildrequest");
+        await browser.wait(EC.not(EC.urlContains('redirect_to_build=true')),
+                           5000,
+                           "failed to create a build");
+    }
+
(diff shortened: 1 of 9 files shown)


smokes/e2e/pages/force.ts
```diff
@@ -49,6 +49,17 @@ export class ForcePage extends BasePage {
         await button.click();
     }
 
+    async clickStartButtonAndGoToBuild() {
```
tardyp: this suggest we force a go to the build.

```suggestion
    async clickStartButtonAndWaitRedirectToBuild() {
```

p12tic: Fixed.

p12tic:
Fixed the nitpick and merging as it's trivial issue.

PR #5109, opened 2019-12-15
worker: Ensure that worker does not leave pending keepalive requests after shutdown

### LINKED ISSUE #4986: buildbot.test.integration.test_worker_workerside.TestWorkerConnection.test_pb_keepalive is flaky
The test is still flaky even after @4969 has been fixed. See e.g. https://buildbot.buildbot.net/#/builders/5/builds/2287. Error message is not very useful:

```[ERROR]
  File "/buildbot/buildbot-job/build/worker/buildbot_worker/pb.py", line 112, in doKeepalive
    details = yield self.perspective.callRemote("keepalive")
twisted.spread.pb.PBConnectionLost: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectionDone'>: Connection was closed cleanly.
]
buildbot.test.integration.test_worker_workerside.TestWorkerConnection.test_pb_keepalive```

p12tic:
Previously the worker would not wait for pending keepalives to finish during shutdown which presented a race condition. The worker could potentially be shut down before the pending keepalives finish. In normal circumstances this is not a problem, but this made buildbot.test.integration.test_worker_workerside.TestWorkerConnection.test_pb_keepalive test flaky, because the concurrent connection shutdown could spew various errors. This PR ignores these errors and also makes sure that we don't leave pending keepalives which could break a test that follows.

Fixes #4986.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/fakeprotocol.py
+++ b/master/buildbot/test/fake/fakeprotocol.py
@@ -16,7 +16,31 @@
 
 from twisted.internet import defer
 
+from buildbot.util import subscription
 from buildbot.worker.protocols import base
+
+
+class FakeTrivialConnection:
+
+    info = {}
+
+    def __init__(self):
+        self._disconnectSubs = subscription.SubscriptionPoint("disconnections from Fake")
+
+    def notifyOnDisconnect(self, cb):
+        return self._disconnectSubs.subscribe(cb)
+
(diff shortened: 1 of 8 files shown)


p12tic:
Looks like tests broke after rebasing, will fix.

PR #5115, opened 2019-12-18
Parse /etc/os-release and expose it as worker info

### LINKED ISSUE #5114: Feature request: Import /etc/os-release into worker properties on startup
When buildbot worker starts up, it should see if /etc/os-release exists and import the key-values specified there into build properties. These can be used to make OS-agnostic workflows that can leverage from knowing what OS is running in the worker.

cmouse:
## Contributor Checklist:

Closes  #5114

This patch makes Worker read /etc/os-release. It also exposes all Worker info fields via new interpolation `Interpolate("worker:field")s`

### DIFF
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -461,6 +461,16 @@ _thePropertyDict = _PropertyDict()
 
 
 @implementer(IRenderable)
+class _WorkerPropertyDict:
+
+    def getRenderingFor(self, build):
+        return build.getBuild().getWorkerInfo()
+
+
+_theWorkerPropertyDict = _WorkerPropertyDict()
+
+
+@implementer(IRenderable)
 class _SecretRenderer:
 
     def __init__(self, secret_name):
@@ -602,6 +612,13 @@ class Interpolate(util.ComparableMixin):
                 "Attribute must be alphanumeric for src Interpolation '%s'" % arg)
             codebase = attr = repl = None
         return _SourceStampDict(codebase), attr, repl
+
+    def _parse_worker(self, arg):
+        try:
+            prop, repl = arg.split(":", 1)
+        except ValueError:
+            prop, repl = arg, None
+        return _theWorkerPropertyDict, prop, repl
 
     def _parse_kw(self, arg):
         try:
--- a/master/buildbot/status/worker.py
+++ b/master/buildbot/status/worker.py
@@ -19,6 +19,7 @@ import time
 from zope.interface import implementer
 
 from buildbot import interfaces
+from buildbot.process.properties import Properties
 from buildbot.util import bytes2unicode
 from buildbot.util.eventual import eventually
 
@@ -41,6 +42,7 @@ class WorkerStatus:
         self.graceful_callbacks = []
         self.pause_callbacks = []
         self.connect_times = []
+        self.info = Properties()
 
     def getName(self):
         return self.name
--- a/master/buildbot/test/integration/test_worker_comm.py
+++ b/master/buildbot/test/integration/test_worker_comm.py
@@ -94,7 +94,12 @@ class FakeWorkerWorker(pb.Referenceable):
             'info': 'here',
             'worker_commands': {
                 'x': 1,
-            }
+            },
+            'numcpus': 1,
+            'none': None,
+            'os_release': b'\xe3\x83\x86\xe3\x82\xb9\xe3\x83\x88'.decode(),
+            b'\xe3\x83\xaa\xe3\x83\xaa\xe3\x83\xbc\xe3\x82\xb9\xe3'
+            b'\x83\x86\xe3\x82\xb9\xe3\x83\x88'.decode(): b'\xe3\x83\x86\xe3\x82\xb9\xe3\x83\x88'.decode(),
         }
 
     def remote_getVersion(self):
@@ -333,6 +338,26 @@ class TestWorkerComm(unittest.TestCase, TestReactorMixin):
         self.workerSideDisconnect(worker)
 
         # wait for the resulting detach
+        yield worker.waitForDetach()
+
+    @defer.inlineCallbacks
+    def test_worker_info(self):
+        yield self.addWorker()
+        worker = yield self.connectWorker()
+        props = self.buildworker.worker_status.info
+        # check worker info passing
+        self.assertEqual(props.getProperty("info"),
+                         "here")
+        # check worker info passing with UTF-8
+        self.assertEqual(props.getProperty("os_release"),
+                         b'\xe3\x83\x86\xe3\x82\xb9\xe3\x83\x88'.decode())
+        self.assertEqual(props.getProperty(b'\xe3\x83\xaa\xe3\x83\xaa\xe3\x83\xbc\xe3\x82'
+                                           b'\xb9\xe3\x83\x86\xe3\x82\xb9\xe3\x83\x88'.decode()),
+                         b'\xe3\x83\x86\xe3\x82\xb9\xe3\x83\x88'.decode())
+        self.assertEqual(props.getProperty("none"), None)
+        self.assertEqual(props.getProperty("numcpus"), 1)
+
+        self.workerSideDisconnect(worker)
         yield worker.waitForDetach()
 
     @defer.inlineCallbacks
--- a/master/buildbot/worker/base.py
+++ b/master/buildbot/worker/base.py
@@ -229,6 +229,12 @@ class AbstractWorker(service.BuildbotService):
         self.worker_status.setAccessURI(info.get("access_uri", None))
         self.worker_status.setVersion(info.get("version", "(unknown)"))
 
+        # store everything as Properties
+        for k, v in info.items():
+            if k in ('environ', 'worker_commands'):
+                continue
+            self.worker_status.info.setProperty(k, v, "Worker")
+
     @defer.inlineCallbacks
     def _getWorkerInfo(self):
         worker = yield self.master.data.get(
--- a/worker/buildbot_worker/base.py
+++ b/worker/buildbot_worker/base.py
@@ -311,6 +311,23 @@ class BotBase(service.MultiService):
     def remote_print(self, message):
         log.msg("message from master:", message)
 
+    @staticmethod
+    def _read_os_release(os_release_file, props):
+        if not os.path.exists(os_release_file):
+            return
+
+        with open(os_release_file, "r") as fin:
+            for line in fin:
+                line = line.strip("\r\n")
+                # as per man page: Lines beginning with "#" shall be ignored as comments.
+                if len(line) == 0 or line.startswith('#'):
+                    continue
+                # parse key-values
+                key, value = line.split("=", 1)
+                if value:
+                    key = 'os_%s' % key.lower()
+                    props[key] = value.strip('"')
+
     def remote_getWorkerInfo(self):
         """This command retrieves data from the files in WORKERDIR/info/* and
         sends the contents to the buildmaster. These are used to describe
@@ -328,18 +345,7 @@ class BotBase(service.MultiService):
                     with open(filename, "r") as fin:
                         files[f] = fin.read()
 
-        if os.path.exists(self.os_release_file):
-            with open(self.os_release_file, "r") as fin:
-                for line in fin:
-                    line = line.strip("\r\n")
-                    # as per man page: Lines beginning with "#" shall be ignored as comments.
-                    if len(line) == 0 or line.startswith('#'):
-                        continue
-                    # parse key-values
-                    key, value = line.split("=", 1)
-                    if value:
-                        key = 'os_%s' % key.lower()
-                        files[key] = value.strip('"')
+        self._read_os_release(self.os_release_file, files)
 
         if not self.numcpus:
             try:
--- a/worker/buildbot_worker/base.py
+++ b/worker/buildbot_worker/base.py
@@ -241,6 +241,8 @@ class BotBase(service.MultiService):
     name = "bot"
     WorkerForBuilder = WorkerForBuilderBase
 
+    os_release_file = "/etc/os-release"
+
     def __init__(self, basedir, unicode_encoding=None):
         service.MultiService.__init__(self)
         self.basedir = basedir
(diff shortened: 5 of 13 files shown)


cmouse:
Do we want to import all fields from /etc/os-release as os_<whatever> and import anything starting with os_ into properties?

master/buildbot/worker/base.py
```diff
@@ -229,6 +229,19 @@ def _applyWorkerInfo(self, info):
         self.worker_status.setAccessURI(info.get("access_uri", None))
         self.worker_status.setVersion(info.get("version", "(unknown)"))
 
+        # import any os_<key>
```
tardyp: I think we should remove that part for this PR.

Ideally, we should create a renderable which retrive these from worker info dynamically, instead of adding a lot of useless properties in every build

worker/buildbot_worker/base.py
```diff
@@ -320,6 +320,16 @@ def remote_getWorkerInfo(self):
                 if os.path.isfile(filename):
                     with open(filename, "r") as fin:
                         files[f] = fin.read()
+
+        if os.path.exists('/etc/os-release'):
+            with open('/etc/os-release', "r") as fin:
+                for line in fin.readlines():
```
tardyp:
```suggestion
                for line in fin:
```
readlines() return a list, and thus will load the whole file before the loop, while the suggested way, we load during the loop

worker/buildbot_worker/base.py
```diff
@@ -320,6 +320,16 @@ def remote_getWorkerInfo(self):
...
+                    # parse key-values
```
tardyp:
```suggestion
                    # as per man page: Lines beginning with "#" shall be ignored as comments.
                    if line.startswith("#"):
                          continue
                    # parse key-values
```

worker/buildbot_worker/test/unit/test_bot.py
```diff
@@ -86,6 +86,9 @@ def test_getWorkerInfo(self):
 
         info = yield self.bot.callRemote("getWorkerInfo")
 
+        # remove any os_ fields as they will mix this up
```
tardyp:
```suggestion
        # remove any os_ fields as they will are dependent on the test environment
```

master/buildbot/worker/base.py
tardyp:
```
Interpolate("%(worker:os_id)s")
```

cmouse: wonder if it would make more sense to just store the info variable as self.info for renderable access?

cmouse: I am not sure how to implement this, is there some example somewhere how this could be done?

tardyp: we indeed will need to cache the info in self.worker.info.

Then Interpolate implementation is a bit tricky:

https://github.com/buildbot/buildbot/blob/b0875d0d017d25665511ad14f5d64887d4865cd8/master/buildbot/process/properties.py#L606

the parser is defined automatically given the keyword, so you need to implement a method named _parse_worker, to support Interpolate("%(worker:os_id)s")

This method should return an IRenderable which will lazily evaluate the info.

phillxnet:
@cmouse and @tardyp this would be a great additional feature. We currently hand code a worker 'os' property ourselves.

I know it might be overkill, and an additional dependency, but the well established python library 'distro' pretty much does the initial parsing of this file, and others I believe, so it might be worth using this instead of re-doing the parsing that it already specialises in. Especially given it can resource/fail through to other means of identifying the os / release info.

https://pypi.org/project/distro/

This same library has recently been adopted also by dnf, via python3-dnf-plugins-core on tumbleweed:
```
zypper info --requires python3-dnf-plugins-core | grep distro
    python3-distro
```
and we have used it to great success in our main projects code and it, currently, also supports legacy and current Python versions.

Apologies if this is noise as I'm unaware of the complexity, from buildbot's internal perspective, of reading a file rather than using distro's existing capabilities; or for that matter of the etiquette of this suggestion.

p12tic:
I think one important thing to consider is not to directly couple the returned data to Python. Current plan is good in this regard, because we're just reading a file and effectively returning its contents. But if we started using some python code that generates values, that will prevent easy reimplementation of worker in another language or platform in the future.

tardyp:
@phillxnet I agree with @p12tic . What do python-distro do support that is not in/etc/os-release?

master/buildbot/worker/base.py
cmouse: Ok. Now it allows actually accessing any worker info field. I am not entirely sure if I am storing the info fields correctly into the Properties object, as in, do I need to do the bytes2unicode before calling it or could I just pass the values as-is into it?

master/buildbot/process/properties.py
```diff
@@ -603,6 +613,17 @@ def _parse_src(arg):
             codebase = attr = repl = None
         return _SourceStampDict(codebase), attr, repl
 
+    def _parse_worker(self, arg):
+        try:
+            prop, repl = arg.split(":", 1)
+        except ValueError:
+            prop, repl = arg, None
+        if not Interpolate.identifier_re.match(prop):
```
tardyp: not sure if we really need that check. If somebody decides to name an info file with chineese characters, this would not be a techical problem, but this check will make it unavailable..

master/buildbot/process/properties.py
```diff
@@ -460,6 +460,16 @@ def getRenderingFor(self, build):
 _thePropertyDict = _PropertyDict()
 
 
+@implementer(IRenderable)
```
tardyp: Need a test! Unit test, and integration test (template example is in https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/integration/test_integration_template.py)

cmouse: Yeah once the actual code stabilizes.

master/buildbot/process/properties.py
cmouse: why is the check done for properties and keywords?

master/buildbot/status/worker.py
```diff
@@ -41,6 +42,30 @@ def __init__(self, name):
...
+    def setInfo(self, info):
+        self.admin = bytes2unicode(info.get("admin"))
+        self.host = bytes2unicode(info.get("host"))
+        self.access_uri = info.get("access_uri", None)
+        if self.access_uri:
+            self.access_uri = bytes2unicode(self.access_uri)
+        self.version = bytes2unicode(info.get("version", b"(unknown)"))
+
+        # store everything as Properties
+        for k, v in info.items():
```
cmouse: @tardyp do i need to decode keyvalues here?

tardyp: I'd prefer we work with unicode everywhere. So we should decode it in the worker.

master/buildbot/process/properties.py
tardyp: Having some running tests will make details like unicode more obvious, so it is good to work on the tests at the same time (aka TDD)

phillxnet:
@tardyp Re:

> @phillxnet I agree with @p12tic . What do python-distro do support that is not in/etc/os-release?

Nothing, currently, other than convenience and proven parsing across many distros, assuming /etc/os-release exists. And given @cmouse has literally titled this pr referencing the surfacing of this file I agree with @p12tic to keep dependencies down, especially if they provide little or no benefit. However my comments were motivated by the purpose of /etc/os-release, i.e. to provide linux distro info and in that context the absence of the os-release file would lead to a failure with this approach. Where as the fail through multi file/command resourcing within the distro library would hopefully prove more robust.

From their: https://distro.readthedocs.io/en/latest/#data-sources
we have:

> The distro package implements a robust and inclusive way of retrieving the information about a Linux distribution based on new standards and old methods, namely from these data sources:
> 
> - The os-release file, if present.
> - The lsb_release command output, if the lsb_release command is available.
> - The distro release file, if present.
> - The `uname command output`_, if present.

In my main project I simply do the following:
```
distro_id = distro.id()
```
which is then "opensuse-tumbleweed" or "opensuse-leap" with version number by some other means but we rarely require that. But will within our Buildbot instance.

A caveat to distro, currently, providing no advantage if os-release exists is the following stated on there GitHub README.md: https://github.com/nir0s/distro

> It is the recommended replacement for Python's original platform.linux_distribution function (which will be removed in Python 3.8). It also provides much more functionality which isn't necessarily Python bound, like a command-line interface.
> 
> Distro currently supports Linux and BSD based systems but Windows and OS X support is also planned.

So that might be important to some, especially if these additional platforms were just inherited via pull requests to distro.

Copying in @nir0s who appears to be project maintainer.

I don't mean to derail or side track this pr review but when I saw mention of /etc/os-release I remembered my selection of distro when I was looking for a python built-in, or established means, to trivially get this info that was also already widely available in linux repos.

master/buildbot/process/properties.py
```diff
@@ -603,6 +613,17 @@ def _parse_src(arg):
             codebase = attr = repl = None
         return _SourceStampDict(codebase), attr, repl
 
+    def _parse_worker(self, arg):
+        try:
+            prop, repl = arg.split(":", 1)
+        except ValueError:
+            prop, repl = arg, None
+        if not Interpolate.identifier_re.match(prop):
+            config.error(
```
p12tic: I think we need a test that covers this error.

worker/buildbot_worker/base.py
```diff
@@ -320,6 +320,19 @@ def remote_getWorkerInfo(self):
                 if os.path.isfile(filename):
                     with open(filename, "r") as fin:
                         files[f] = fin.read()
+
+        if os.path.exists('/etc/os-release'):
+            with open('/etc/os-release', "r") as fin:
+                for line in fin:
+                    # as per man page: Lines beginning with "#" shall be ignored as comments.
+                    if line.startswith('#'):
+                        continue
```
p12tic: I think we need a test that covers this line.

master/buildbot/status/worker.py
```diff
@@ -41,6 +42,30 @@ def __init__(self, name):
...
+            # be careful about few items.
+            if k == 'numcpus':
+                self.info.setProperty(bytes2unicode(k), v, "Worker")
+            elif k in ('environ', 'worker_commands'):
+                continue
+            elif v is None:
+                self.info.setProperty(bytes2unicode(k), None, "Worker")
```
p12tic: I think we need a test that covers this case.

master/buildbot/process/properties.py
cmouse: asch, was supposed to remove the check altogether

worker/buildbot_worker/base.py
cmouse: that might be bit difficult. any ideas how?

master/buildbot/status/worker.py
cmouse: i'll add these

worker/buildbot_worker/base.py
```diff
@@ -320,6 +326,20 @@ def remote_getWorkerInfo(self):
                 if os.path.isfile(filename):
                     with open(filename, "r") as fin:
                         files[f] = fin.read()
+
+        if os.path.exists(self.os_release_file):
```
p12tic: I think a little better way would have been to extract the actual parsing function which gets the contents of the read file or maybe a list of lines. This way we could unit test it directly. Anyway, this is also an acceptable solution.

cmouse: I think this is possibly better way, as @tardyp was concerned about using readlines() earlier.

master/buildbot/status/worker.py
```diff
@@ -41,6 +42,24 @@ def __init__(self, name):
         self.graceful_callbacks = []
         self.pause_callbacks = []
         self.connect_times = []
+        self.info = Properties()
+
+    def setInfo(self, info):
```
tardyp: Don't add new API to status classes. They are meant to disappear.
Please move this to worker class.

master/buildbot/status/worker.py
```diff
@@ -41,6 +42,24 @@ def __init__(self, name):
...
+        self.access_uri = info.get("access_uri", None)
+        if self.access_uri:
+            self.access_uri = bytes2unicode(self.access_uri)
+        self.version = bytes2unicode(info.get("version", b"(unknown)"))
+
+        # store everything as Properties
+        for k, v in info.items():
+            if k in ('environ', 'worker_commands'):
+                continue
+            self.info.setProperty(k, v, "Worker")
```
tardyp: Why storing as a Property, why not just store the info dict as is?

master/buildbot/test/integration/test_worker_comm.py
```diff
@@ -94,7 +94,12 @@ def remote_getWorkerInfo(self):
             'info': 'here',
             'worker_commands': {
                 'x': 1,
-            }
+            },
+            'numcpus': 1,
+            'none': None,
+            'os_release': b'\xe3\x83\x86\xe3\x82\xb9\xe3\x83\x88',
```
tardyp: didn't we say we should decode in the worker?

worker/buildbot_worker/base.py
```diff
@@ -320,6 +326,20 @@ def remote_getWorkerInfo(self):
...
+            with open(self.os_release_file, "r") as fin:
```
tardyp: you should open with 'rb' if you want to manage bytes

master/buildbot/status/worker.py
cmouse: I am 100% convinced that the master side will only get bytes and not string no matter how much I decode on worker.

master/buildbot/test/integration/test_worker_comm.py
cmouse: I am 100% convinced that the master side will only get bytes and not string no matter how much I decode on worker.

master/buildbot/status/worker.py
cmouse: I tried with dict and it did not work well. I can try again if it is important.

worker/buildbot_worker/base.py
cmouse: I don't think I want to manage bytes.

cmouse:
some last minute fixes to nits by tardyp.

p12tic:
Added a full-round-trip integration test to make sure encoding is correct. I get a feeling that the existing integration test would have probably been enough, but since we will add a second worker implementation, we will be able to reuse this integration test on the C++ worker too.

PR #5119, opened 2019-12-20
Fix displaying of radio options (choices) in settings page

vibbo:
## Contributor Checklist:


vibbo:
I have stumbled upon this while trying to understand how settings from master work, so here is a small fix.

Current (caption is ignored):
old
Fixed (when caption is defined):
new_with_caption
Fixed (when caption is not defined):
new_no_caption

tardyp:
cool! thanks

PR #5122, opened 2019-12-20
mq: Wait until delivered callbacks are completed during shutdown

p12tic:
Currently message queues are fire the callbacks without any regard to when they are completed. This has issues during shutdown, because some of the callbacks may not be completed. If message queues don't wait for their completion and proceed to stop the service, sometimes the resources needed for the callbacks to succeed are pulled from under their feet. This fortunately only affects tests.

This PR improves the queues to manage a set of all pending deferreds that have resulted from callbacks and only proceed to return from `stopService()` when all these deferreds fire.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/mq/wamp.py
+++ b/master/buildbot/mq/wamp.py
@@ -59,23 +59,24 @@ class WampMQ(service.ReconfigurableServiceMixin, base.MQBase):
             log.err('wampmq: persistent queues are not persisted: %s %s' %
                     (persistent_name, _filter))
 
-        qr = QueueRef(callback)
+        qr = QueueRef(self, callback)
 
         self._startConsuming(qr, callback, _filter)
         return defer.succeed(qr)
 
     def _startConsuming(self, qr, callback, _filter, persistent_name=None):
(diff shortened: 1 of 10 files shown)


master/buildbot/mq/wamp.py
```diff
@@ -59,42 +59,43 @@ def startConsuming(self, callback, _filter, persistent_name=None):
...
-    def invoke(self, msg, details):
+    def wampInvoke(self, msg, details):
         if details.topic is not None:
             # in the case of a wildcard, wamp router sends the topic
             topic = WampMQ.routingKeyFromMessageTopic(details.topic)
         else:
             # in the case of an exact match, then we can use our own topic
             topic = self.filter
-        return super().invoke(topic, msg)
+        self.mq.invokeQref(self, topic, msg)
```
tardyp: so you don't wait for that one anymore? why?

p12tic: The invoke function didn't return a deferred anyway, so we effectively didn't wait.

PR #5126, opened 2019-12-20
worker: Add newsfragment for PR #5109

p12tic:
Given that newsfragments have been added for similar changes in other PRs, it makes sense to do the same for #5109 too.


tardyp:
Why did this test suddenly break?

p12tic:
That keepalive test was unstable for as long as I can remember.

PR #5128, opened 2019-12-21
Make sure any Deferreds returned by setServiceParent() are waited upon

p12tic:
In many cases we don't wait for the result of `setServiceParent()`. This may result in test unreliability, because some actions performed by `setServiceParent()` may end up being being done after the specific unit test ends. This PR improves the situation.


tardyp:
I resolved conflicts.

Issue #4328, opened 2018-09-15
JanitorConfigurator for specific builders


aj062:
JanitorConfigurator should have an option to delete logs for specific builders (or tags). Users might want to delete logs for only specific builders, or might want to keep logs for specific builders for longer duration.

http://docs.buildbot.net/latest/manual/cfg-configurators.html#janitorconfigurator currently doesn't seems to contain any such option.

JohnLapis:
I'll be working on this.

JohnLapis:
I had a look and found the following. The function (`deleteOldLogsChunks`) which deletes the logs only takes `logHorizon` as parameter, generalizing everything. So, it came to mind:
1.  writing another function in `buildbot/db/logs.py`  which specifies the builders and using it in a loop in `janitor.py`
2.  filtering out the specified builders from this general function.

I think I could do (1) based on the original function, although I don't have practice with code SQL in Python. But (2) looks difficult for me to do since I don't understand the function very well. What do you think?

shawngonsalves:
I would like to contribute to this project as well, but I am new to open-source-contribution. It'd be really helpful if you can share some of your ideas

tardyp:
Hi @shawngonsalves welcome to open-source .

You can get the doc on how to contribute to buildbot in this page:

http://docs.buildbot.net/latest/developer/pull-request.html

while the guide to developing buildbot is there:
https://docs.buildbot.net/latest/developer/quickstart.html

Don't hesitate to create a new issue or comment on an existing issue if you want to take it, and explain how you want to tackle the problem, so that we can advice and avoid you to spend time on some non ideal ideas.

mrmammadov:
Hi @tardyp 
This issue is still open. Can I take it over? I wonder if there is a Jira Board to avoid the duplicate work?

tardyp:
@mrmammadov I assigned this bug to you. you can safely work on it :)

BlacPharaoh:
Is this issue still open, I can work on it

PR #5134, opened 2019-12-26
JanitorConfigurator can specify builders for deleting logs

JohnLapis:
This pull request adds the option requested in issue #4328 adding a new option to the `JanitorConfigurator` and creating a new function in `db/logs.py`

I would like some feedback on certain things:
1. I don't know if it's better to add a new function in `janitory.py`
2. I re-utilized some code from the `deleteOldLogChunks` in the `deleteBuilderLogs` and don't know if it should be abstracted so as to not have almost the same code twice
3. I don't know if I should have personalized the exception filtration again for sqlite since it doesn't support multiple-table criteria
4. I don't know how I should add a documentation


master/docs/relnotes/index.rst
```diff
@@ -10,6 +10,84 @@ Release Notes
 
 .. towncrier release notes start
 
+Buildbot ``2.5.2-dev85`` ( ``2019-12-26`` )
```
p12tic: Don't commit updated release notes, they will be generated on release. We should perhaps remove this functionality from `make docs` to prevent confusion.

master/buildbot/test/unit/test_db_logs.py
```diff
@@ -573,6 +573,80 @@ def test_deleteOldLogChunks_basic(self):
             lines = yield self.db.logs.getLogLines(logid, 0, logdict['num_lines'])
             self.assertEqual(lines, '')
 
+    @defer.inlineCallbacks
+    def test_deleteOldLogChunks_exceptions(self):
+        yield self.insertTestData(
+            self.backgroundData + [
```
p12tic: Would it be possible to define a separate `self.deleteBuildLogsData` and with full data set (i.e. inline the `self.backgroundData` there)? This way it would be easier to understand what data test operates on as it would all be in a single place.

master/buildbot/test/unit/test_db_logs.py
```diff
@@ -143,7 +143,7 @@ def compressLog(self, logid, force=False):
 
     def test_signature_deleteOldLogChunks(self):
         @self.assertArgSpecMatches(self.db.logs.deleteOldLogChunks)
-        def deleteOldLogChunks(self, older_than_timestamp):
+        def deleteOldLogChunks(self, older_than_timestamp, exceptions=[]):
```
p12tic: Shouldn't we use empty map here?

master/buildbot/newsfragments/issue4328.feature
```diff
@@ -0,0 +1,3 @@
+The JanitorConfigurator now has the option to specify the log horizons for builders.
+The deleteOldLogChunks function can now filter builders.
```
p12tic: No need to explain what has been done for internal functions.

master/buildbot/db/logs.py
```diff
@@ -346,7 +346,7 @@ def thdcompressLog(conn):
         return saved
 
     # returns a Deferred that returns a value
-    def deleteOldLogChunks(self, older_than_timestamp):
+    def deleteOldLogChunks(self, older_than_timestamp, exceptions=[]):
```
p12tic: Shouldn't default value of `exceptions` be an empty dict?

master/buildbot/configurators/janitor.py
```diff
@@ -40,32 +40,53 @@ def now():
...
+                self.descriptionDone += ['\n']
+            self.descriptionDone += ["deleted", str(deleted), "logchunks"]
+
+        if self.logHorizon is not None or self.buildersLogHorizon != []:
+            return SUCCESS
 
 
 class JanitorConfigurator(ConfiguratorBase):
-    def __init__(self, logHorizon=None, hour=0, **kwargs):
+    def __init__(self, logHorizon=None, hour=0, buildersLogHorizon=[], **kwargs):
```
p12tic: Shouldn't this be an empty dict? Even better, I would suggest to use `None`, because all instances of JanitorConfigurator will use the same instance of the dict and if anyone modifies it, the default value will change.

master/buildbot/configurators/janitor.py
```diff
@@ -40,32 +40,53 @@ def now():
...
+            older_than_timestamp = datetime2epoch(now() - self.logHorizon)
+            builders = []
+            if self.buildersLogHorizon != []:
+                builders = list(self.buildersLogHorizon.keys())
+            deleted = yield self.master.db.logs.deleteOldLogChunks(older_than_timestamp, exceptions=builders)
+            if self.descriptionDone != []:
+                self.descriptionDone += ['\n']
+            self.descriptionDone += ["deleted", str(deleted), "logchunks"]
+
+        if self.logHorizon is not None or self.buildersLogHorizon != []:
```
p12tic: If this condition does not match, we will return None. Is this correct return value?

master/buildbot/test/unit/test_db_logs.py
JohnLapis: There's no need to since I just filtering by name. The function doesn't need to know the specific logs; that's handled in the other function.

master/docs/relnotes/index.rst
JohnLapis: Thank you for the tip

master/buildbot/db/logs.py
JohnLapis: Same as explained above. In the LogChunksJanitor class, only the names are passed. Maybe passing a dict and just using the names is better. What do you think?

master/buildbot/configurators/janitor.py
JohnLapis: That was a super slip! I must have gotten confused. I did thought of using None but I thought that, in some cases, having an empty dict would be better than handling None.

master/buildbot/configurators/janitor.py
JohnLapis: The JanitorConfigurator class already have a condition to the case neither of logHorizon nor buildersLogHorizon is given that return None. But, since I couldn't think of all the possibilities that the function might return SUCCESS when nothing happened, I chose to code defensively.

master/buildbot/test/unit/test_db_logs.py
JohnLapis: Got it.

master/buildbot/test/unit/test_db_logs.py
p12tic: Okay, I got confused.

master/buildbot/configurators/janitor.py
p12tic: Alright, I will investigate and see if we can have a better return option.

p12tic: Alright, I will investigate and see if we can have a better return option.

master/docs/relnotes/index.rst
tardyp: it is still in the PR. please remove it!

master/buildbot/db/logs.py
```diff
@@ -373,11 +373,107 @@ def thddeleteOldLogs(conn):
...
+                    # WHERE steps.buildid = builds.id
+                    # AND builds.builderid = builders.id
+                    # AND builders.name IN exceptions
+                    res = conn.execute(
+                        sa.select([model.steps.c.id])
+                        .where(sa.and_(model.steps.c.buildid == model.builds.c.id,
+                                       model.builds.c.builderid == model.builders.c.id,
+                                       model.builders.c.name.in_(exceptions)))
+                    )
+                    stepid_exceptions = [i[0] for i in res.fetchall()]
```
tardyp: this will store in memory all steps for all those builders from the beginning of time. This is not acceptable. I can't believe there is no way in sqlite to do this in one request. did you try subqueries?

master/buildbot/db/logs.py
```diff
@@ -373,11 +373,107 @@ def thddeleteOldLogs(conn):
...
+                    # AND logs.stepid <= stepid_max AND type != 'd';
+                    res = conn.execute(
+                        model.logs.update()
+                        .where(sa.and_(model.logs.c.stepid.notin_(stepid_exceptions),
+                                       model.logs.c.stepid <= stepid_max,
+                                       model.logs.c.type != 'd'))
+                        .values(type='d')
+                    )
+                    res.close()
+                else:
```
tardyp: this function is too complex, we need to split it into several subproblems

master/buildbot/db/logs.py
```diff
@@ -373,11 +373,107 @@ def thddeleteOldLogs(conn):
...
+                    # UPDATE logs SET logs.type = 'd' WHERE logs.stepid <= stepid_max AND type != 'd';
+                    res = conn.execute(
+                        model.logs.update()
+                        .where(sa.and_(model.logs.c.stepid <= stepid_max,
+                                       model.logs.c.type != 'd'))
+                        .values(type='d')
+                    )
+                    res.close()
+
+            # query all logs with type 'd' and delete their chunks.
```
tardyp: this needs to be factorized as this is the same code as deleteBuilderLogs

master/buildbot/db/logs.py
```diff
@@ -373,11 +373,107 @@ def thddeleteOldLogs(conn):
...
+                # AND builders.c.name = builder
+                # AND steps.started_at < builder_timestamps[builder]
+                res = conn.execute(
+                    sa.select([model.steps.c.id])
+                    .where(sa.and_(model.steps.c.buildid == model.builds.c.id,
+                                   model.builds.c.builderid == model.builders.c.id,
+                                   model.builders.c.name == builder,
+                                   model.steps.c.started_at < builder_timestamps[builder]))
+                )
+                builder_stepids = [i[0] for i in res.fetchall()]
```
tardyp: same as before. this is not memory bound!

master/buildbot/db/logs.py
tardyp: Initializing parameters with mutable objects is strongly discouraged. https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Issue #3630, opened 2017-09-18
pb.Connection.doKeepalive() is not called in loop


MikeAlexeev:
Hello. I'm using 0.9.11
I'm working on case when latent worker is killed. For example vm with worker stopped.
And I have found that in this case master continue to think that worker is ok (build is not
failed).

There is a doKeepalive procedure in pb.Connection class, but
1) It is called once with reactor.callLater(). I think that it is better to call it in loop
to check periodically if worker is online.
2) I haven't found what happens if there is no answer from worker
3) keepalive_timer value is hardcoded. I think that there could be two places for this value -
global in config (as mentioned in comments) and in worker parameters

doKeepalive() returns Deffered. I added timeout to it with loseConnection callback like this:
```
def doKeepalive(self):
        d = self.mind.callRemote('print', message="keepalive")
        d.addTimeout(30, reactor, lambda x, y: self.loseConnection())
        return d

```

Is it ok or maybe there is another solution for this case? I can prepare PR.
Please correct me if I'm wrong

seankelly:
Seems related to #2914.

PR #5139, opened 2019-12-29
Repeat master-side keepalive requests

p12tic:
Fixes issue #3630.

This PR improves the master-side keepalive behavior to send keepalive requests continuously instead of a single time like it was before. Additionally, one direct use of real `reactor` has been removed which could have been a source of test instability (we should always use `master.reactor`).

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/master-side-keepalive-repeat.bugfix
+++ b/master/buildbot/newsfragments/master-side-keepalive-repeat.bugfix
@@ -1,0 +1,1 @@
+Master side keep-alive requests are now repeated instead of being single-shot (:issue:`3630`).
--- a/master/buildbot/test/fake/fakeprotocol.py
+++ b/master/buildbot/test/fake/fakeprotocol.py
@@ -27,6 +27,9 @@ class FakeTrivialConnection:
     def __init__(self):
         self._disconnectSubs = subscription.SubscriptionPoint("disconnections from Fake")
 
+    def waitShutdown(self):
+        return defer.succeed(None)
+
     def notifyOnDisconnect(self, cb):
         return self._disconnectSubs.subscribe(cb)
 
--- a/master/buildbot/test/unit/test_util_deferwaiter.py
+++ b/master/buildbot/test/unit/test_util_deferwaiter.py
@@ -13,17 +13,22 @@
 #
 # Copyright Buildbot Team Members
 
+from parameterized import parameterized
+
 from twisted.internet import defer
 from twisted.trial import unittest
 
+from buildbot.test.util.misc import TestReactorMixin
+from buildbot.util import asyncSleep
 from buildbot.util.deferwaiter import DeferWaiter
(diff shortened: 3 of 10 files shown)


tardyp:
Looking at the git history of this code, it looks like there has been some loss in translation in between refactorisation of the protocol at 04a8a065776a6b5813082d5154a7b35f61bc7826
refacto over refacto, I think we all got confused by ```callLater(xx_interval, ..)``` assuming that callLater was retriggered when the answer was retrieved.

Stepping backward, I think this is not very well engineered, and should be treated as a protocol bug rather than a implementation detail bug. What are we trying to achieve here?

1) We want to keep the tcp connection up in case there is some router or NAT in between which keep tracks of connection and detects stalled ones.
   In this case, probably the worker keep-alive is probably better as it can be configured by worker, and those are the ones who have weird NAT setups, and this is per worker dependent.

2) We want to make sure the worker is not crashed, and still responding.
 This is the role of the ping -> pong part of the protocol (ping is sent at the begining of a build)

3) related: There is also a missingTimer, which is triggered when a worker is disconnected, and will fire if the worker fails to re-connect (and send an email)

I think this are too many timers, and in the perspective of rewriting a protocol with a more modern RPC, I would remove that one alltogether (maybe keep the nice test infra introduced here, though)

Anyway if we keep this timer for a reason I couldn't think of, we should probably replace it with a ping/pong. print will actually print something in the worker log, which is useless, and we will get complain about it from people actually looking at those logs..

p12tic:
I think it makes sense to have a proper keepalive from the master side. The current approaches only let the master know about a crashed worker only if never shows up. If it crashes in some way without resetting the tcp connection, the master will only know when the tcp connection times out, which can take a while if no builds are started on that worker. Given that someone did create a bug report, I think this use case is important.

Having said that, I agree that printing something to the logs in the worker is not the most appropriate solution, but my changes don't change the situation too much since we are already printing stuff on keepalive.

tardyp:
Right, but as master side keep alive has never actually printed more than
one line it never bothered anybody..

p12tic:
I think the amount of extra logging is not too excessive, because the workers already log their own keepalive with 10 minute period, so by default master-level keepalives will be 6 times rarer. Additionally, a single build will produce more logs than either of these per day, so end users will not see too many messages.

p12tic:
However, we can add a separate ping command. There's already a TODO for this in workerforbuilder's Ping class.

tardyp:
Indeed, I did not realize/remember ping was actually implemented as print. silly indeed.

p12tic:
So is this PR good with the condition that a separate ping command is implemented later?

p12tic:
Thanks!

tardyp:
Problem of implementing a ping command is that it will require upgrade of all workers. So effectively not very useful imho. Its however a must for protocol V2.

PR #5149, opened 2020-01-04
P4: add streams support

bschatt:
Simple patch needed to support Streams depot.

I will update the unit tests and documentation later.

### DIFF
--- a/master/buildbot/steps/source/p4.py
+++ b/master/buildbot/steps/source/p4.py
@@ -59,6 +59,7 @@ class P4(Source):
                  p4extra_args=None,
                  p4bin='p4',
                  use_tickets=False,
+                 stream=False,
                  **kwargs):
         self.method = method
         self.mode = mode
@@ -76,6 +77,7 @@ class P4(Source):
         self.p4client_spec_options = p4client_spec_options
         self.p4extra_args = p4extra_args
         self.use_tickets = use_tickets
+        self.stream = stream
 
         super().__init__(**kwargs)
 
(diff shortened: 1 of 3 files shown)


master/buildbot/steps/source/p4.py
```diff
@@ -307,7 +309,11 @@ def _createClientSpec(self):
         def has_whitespace(*args):
             return any([re.search(r'\s', i) for i in args if i is not None])
 
-        if self.p4viewspec:
+        if self.stream:
+            q = '"' if has_whitespace(self.p4base, self.p4branch) else ''
+            client_spec += "Stream:\t%s%s/%s%s\n\n" % (q, self.p4base, self.p4branch, q)
```
tardyp: why not quoting it unconditionally?

I see that the rest of the code does that as well, but sounds weird anyway

p12tic: Agreed. Also, can we use python3's `.format()`?

bschatt: My code is similar to the code that can be found a few lines below it (and was known to be working).

https://github.com/buildbot/buildbot/blob/20b7713d18501e4b41235d87e7a5f4cb1a531728/master/buildbot/steps/source/p4.py#L327

https://github.com/buildbot/buildbot/blob/20b7713d18501e4b41235d87e7a5f4cb1a531728/master/buildbot/steps/source/p4.py#L349

If I update my code, should I update the rest of the code too ?

tardyp: > If I update my code, should I update the rest of the code too ?

yes please. you probably will need to also update the tests

bschatt:
I think I will be able to start working on this task again soon (probably next week).

master/buildbot/steps/source/p4.py
bschatt: String formatting was fixed in another PR.

I did a few test with perforce and it appears that you can't have whitespace in a Stream path (example: in **//depot/stream** both **depot** and **stream** can't have whitespace). I added a check and removed the quote handling.

I don't think I should change the way the quoting is implemented in the rest of code for now. Maybe in a future PR.

PR #5150, opened 2020-01-04
p4poller: don't parse login ticket

bschatt:
The current login flow uses the "-p" option, which, according to [the perforce documentation](https://www.perforce.com/manuals/v17.1/cmdref/Content/CmdRef/p4_login.html) :

> Display the ticket, rather than storing it in the local ticket file.

This PR simplifies the login flow by using the default perforce login behavior : the ticket is automatically stored in a file. Its location can be changed using the [P4TICKETS environment variable](https://www.perforce.com/manuals/v17.1/cmdref/Content/CmdRef/P4TICKETS.html)). There is no need to parse and store the ticket anymore. This will prevent any issue in the future if the output changes again (as in #4574).

This PR is used in production since October. It fixes the (frequent) random issue I detailed in #5042. Everything is working fine since this patch was applied.

Let me know what you think about it (I can squash the commits once they are reviewed).


p12tic:
Shouldn't we remove our use of `-p` option if we're storing the ticket?

bschatt:
I think we should, yes. But it is currently working as is. I will try to remove it later and test it.

https://github.com/buildbot/buildbot/blob/d2596c79d05d9cbae0197cd087408d55794599cd/master/buildbot/changes/p4poller.py#L191

p12tic:
@bschatt 

> I think we should, yes. But it is currently working as is. I will try to remove it later and test it.

I wonder if it could be the case that the tickets are generated by something else in your setup. Unless the documentation is wrong, the setup that only uses `-p` for `p4 login` shouldn't work.

bschatt:
The server is only running buildbot so it's not possible that the ticket is generated by something else.
 
I will do some tests to check how p4 behaves with the -p option with different versions and I will let you know (soon hopefully).

bschatt:
@p12tic you were right. I was running out of time when I made this patch a few months ago, so I decided to disable ticket timeout for my buildbot user. Which means that it is still using the ticket issued at that time.

I tested the -p behavior and the documentation is correct (it doesn't store the ticket with the -p option).

I will update my patch.

p12tic:
@bschatt I was wondering, what would happen if two or more `p4 login` were racing to refresh the same outdated ticket. Do you have any idea if that wouldn't result in a corrupt ticket file?

bschatt:
@p12tic I don't think so. I just checked using strace and it seems that the client is writing in a temporary file. They are also creation a lock file (.p4tickets.lck).

> lstat("/home/bschatt/.p4tickets", {st_mode=S_IFREG|0400, st_size=170, ...}) = 0
> lstat("/home/bschatt/tmp.8119.1", 0x7fff8bd6c140) = -1 ENOENT (Aucun fichier ou dossier de ce type)
> openat(AT_FDCWD, "/home/bschatt/tmp.8119.1", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 5
> write(5, "xxxxxxxxxxxxxxx"..., 230) = 230
> close(5)                                = 0
> chmod("/home/bschatt/tmp.8119.1", 0644) = 0
> rename("/home/bschatt/tmp.8119.1", "/home/bschatt/.p4tickets") = 0
> chmod("/home/bschatt/.p4tickets", 0400) = 0
> close(4)                                = 0
> chmod("/home/bschatt/.p4tickets.lck", 0444) = 0
> unlink("/home/bschatt/.p4tickets.lck")  = 0

p12tic:
Cool, looks like they put some effort to make this reliable.

For the unit tests, could you try to run locally and update broken ones and remove no longer needed ones? The BuildBot manual has a section about running unit tests: http://docs.buildbot.net/latest/developer/quickstart.html#create-a-buildbot-python-environment

bschatt:
@tardyp the whole point of "p4 login" is to generate a ticket to avoid sending the password over and over.

I found the [2009 documentation](https://www.perforce.com/perforce/doc.091/manuals/cmdref/login.html) which says : 
> Log in to a Perforce server by obtaining a ticket.

So I don't think there is any compatibility issue here : it will work as long as "login" is available (same behavior as the current code).

magestik:
Tests are now passing. I don't think updating the doc is necessary since nothing really change from a user perspective. No compatibility issue either.

bschatt:
Tests are now passing. I don't think updating the doc is necessary since nothing really change from a user perspective. No compatibility issue either.

p12tic:
Looks great, thank you! I've added a release note for this fix.

bschatt:
Ok thanks. Is there anything else I should do ?

p12tic:
@bschatt: There's nothing more needed from your side, thanks a lot!

PR #5156, opened 2020-01-13
worker: Allow ipv6 address for master connection on buildbot-worker

### LINKED ISSUE #5155: buildbot-worker doesn't support ipv6 address for master connection
I'm trying the following command line:
```shell
/usr/bin/buildbot-worker create-worker . "[2001\:\:324]:9989" test toto
```
But it gives the following error:
```
  File "/usr/bin/buildbot-worker", line 11, in <module>
    load_entry_point('buildbot-worker==2.0.1.post2', 'console_scripts', 'buildbot-worker')()
  File "/usr/lib/python3/dist-packages/buildbot_worker/scripts/runner.py", line 238, in run
    config.parseOptions()
  File "/usr/lib/python3/dist-packages/twisted/python/usage.py", line 267, in parseOptions
    self.subOptions.parseOptions(rest)
  File "/usr/lib/python3/dist-packages/twisted/python/usage.py", line 273, in parseOptions
    self.parseArgs(*args)
  File "/usr/lib/python3/dist-packages/buildbot_worker/scripts/runner.py", line 183, in parseArgs
    self['host'], self['port'] = self.validateMasterArgument(master)
  File "/usr/lib/python3/dist-packages/buildbot_worker/scripts/runner.py", line 161, in validateMasterArgument
    master, port = master_arg.split(":")
ValueError: too many values to unpack (expected 2)
```

Is Ipv6 address supposed to be supported?

jocelynj:
This fixes #5155.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/buildbot-worker-ipv6.bugfix
+++ b/master/buildbot/newsfragments/buildbot-worker-ipv6.bugfix
@@ -1,0 +1,1 @@
+Command `buildbot-worker create-worker` now supports ipv6 address for buildmaster connection.
--- a/master/docs/spelling_wordlist.txt
+++ b/master/docs/spelling_wordlist.txt
@@ -408,6 +408,7 @@ iPhone
 IProperties
 iptables
 IPv
+ipv
 irc
 Irc
 IRenderable
--- a/worker/buildbot_worker/scripts/runner.py
+++ b/worker/buildbot_worker/scripts/runner.py
@@ -158,11 +158,24 @@ class CreateWorkerOptions(MakerBase):
         if master_arg[:5] == "http:":
             raise usage.UsageError("<master> is not a URL - do not use URL")
 
-        if ":" not in master_arg:
+        if master_arg.startswith("[") and "]" in master_arg:
+            # detect ipv6 address with format [2001:1:2:3:4::1]:4321
+            master, port_tmp = master_arg.split("]")
+            master = master[1:]
+            if ":" not in port_tmp:
+                port = 9989
+            else:
+                port = port_tmp.split(":")[1]
+
+        elif ":" not in master_arg:
             master = master_arg
             port = 9989
         else:
-            master, port = master_arg.split(":")
+            try:
+                master, port = master_arg.split(":")
+            except ValueError:
+                raise usage.UsageError("invalid <master> argument '{}', "
+                                       "if it is an ipv6 address, it must be enclosed by []".format(master_arg))
 
         if not master:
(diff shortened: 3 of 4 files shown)


jocelynj:
Hum, after thinking about this fix - it won't work if port is not specified, like this:
```buildbot-worker create-worker . 2001:1:2:3:4::1  toto titi```

tardyp:
> Hum, after thinking about this fix - it won't work if port is not specified, like this:
buildbot-worker create-worker . 2001:1:2:3:4::1 toto titi

Fine by me. We just have to document that if IPv6 addresses are to be used, you must supply the port.

I am not against fixing this issue, even if I don't see a lot of reason not to use properly setup hostnames and dns to identify the master address.

jocelynj:
What about supporting syntax [2001:1:2:3:4::1] and [2001:1:2:3:4::1]:9999 instead?

The original issue I had is that buildbot-worker couldn't connect to a hostname which is configured with an ipv6 only, so I had to switch to using an explicit ipv6 address.

p12tic:
@jocelynj Actually the fact that we can't properly support port or absence of it was a reservation that I had. Putting the ipv6 address between square brackets would make it consistent to how it's specified in URIs and solve the ambiguity with the port. Please go ahead.

Thanks!

jocelynj:
I've updated the code and the tests to support [ipv6]:port syntax.

I've also added some text in newsfragment, but I'm not sure if I need to create one, and what to put inside.

p12tic:
> I've also added some text in newsfragment, but I'm not sure if I need to create one, and what to put inside.

All good. The newsfragment is a sentence or two for the release notes. All newsfragments are combined automatically at release time.

p12tic:
@jocelynj: 

The following are remaining issues:

 - docs test fails due to spellcheck errors. You need to add `ipv` to master/docs/spelling_wordlist.txt

 - there are pylint and flake8 errors

 - the unit tests fail with errors that seem to be related to the pull request.

For the last 2 points would be easiest to fix if you setup a python environment according to the instructions here: https://docs.buildbot.net/2.6.0/developer/quickstart.html#create-a-buildbot-python-environment

jocelynj:
I see - looks like it is failing because a string containing `\[` must be tagged as a regexp with `r"..."`

jocelynj:
It looks like that the remaining fails are on master too :(

Should I rebase onto an earlier master?

p12tic:
Whoops, looks like I broke the tests with one of my recent PRs. Interestingly, I can't reproduce the failure locally. This PR looks fine, I'll take care of it. Thanks!

PR #5160, opened 2020-01-17
worker: add option to delete leftover dirs

ThomasFaivre:
Removing these leftover directories can be forgotten resulting in
increasing disk usage.

Add an option to create-workers to remove these directories when
detected (disabled by default to keep current behaviour).

> $ buildbot-worker create-worker --delete-leftover-dirs <dir> <server>
<name> <password>
> mkdir -p <dir>/TFA-test
> 2020-01-17 14:38:55+0100 [Broker,client] Deleting directory 'TFA-test' that is not being used by the buildmaster

Signed-off-by: Robin Jarry <robin@jarry.cc>
Signed-off-by: Thomas Faivre <thomas.faivre@6wind.com>

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/worker_deleteleftoverdirs.feature
+++ b/master/buildbot/newsfragments/worker_deleteleftoverdirs.feature
@@ -1,0 +1,1 @@
+New ``buildbot-worker create-worker --delete-leftover-dirs`` option to automatically remove obsolete builder directories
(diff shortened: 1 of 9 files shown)


tardyp:
could you please rebase?

ThomasFaivre:
> could you please rebase?

Done ;-)

mojca:
I didn't yet try to apply this change, but I wanted to warn about something. Today I noticed that my worker's log is saying `I have a leftover directory 'foo' that is not being used by the buildmaster: you can delete it now`, while in fact the worker uses `foo/darwin-x86_64` to build stuff. I fear that this would delete the top level dir.

p12tic:
Ah, yet another time when having at least an automated test is worthwhile.

PR #5167, opened 2020-01-24
latent: Fix issues in stopService()

### LINKED ISSUE #4935: Insubstantiation notifier left after the refactor
Which blocks the `BuildMaster.stopService`, see https://github.com/buildbot/buildbot/blob/master/master/buildbot/worker/latent.py#L495

p12tic:
Fixes #4935.

Additionally, there's a potential race condition between `stopService()` and `substantiate()`. This PR fixes that too.

Finally, currently workers will wait for current builds to finish during reconfiguration or master shutdown. I believe we don't want to do this during reconfiguration, because in many setups builds take tens of minutes and maybe even hours, it does not make sense for the operator to wait as terminated builds will be retried anyway. During master shutdown the master will wait for builds to finish if the master is shutting down cleanly anyway. Thus I removed this feature.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -680,6 +680,7 @@ insubstantiate
 insubstantiated
 insubstantiating
 insubstantiation
+insubstantiations
 integrations
 internet
 interoperability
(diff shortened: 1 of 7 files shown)


tardyp:
cool work!

PR #5171, opened 2020-01-27
doc: fixes for debian packaging

rjarry:
These are patches that are required to build in debian.

https://salsa.debian.org/python-team/applications/buildbot/blob/debian/master/debian/patches/0002-docs-bundle-sphinxcontrib.jinja-extension.patch
https://salsa.debian.org/python-team/applications/buildbot/blob/debian/master/debian/patches/0005-docs-convert-blockdiag-diagrams-to-ascii.patch

Since I am having a hard time rebasing them for each release, I figured it would be a good thing to apply them upstream.

The sphinxcontrib-jinja package could be integrated into debian, but it would take me quite some time. Also, I am not sure the package is used outside buildbot.

The sphinxcontrib-blockdiag project does not seem maintained anymore. Maybe it is a sane thing to get rid of the dependency. ASCII diagrams should be enough for our use.

### DIFF
--- a/master/docs/conf.py
+++ b/master/docs/conf.py
@@ -33,14 +33,6 @@ except ImportError:
 
 # -- General configuration -----------------------------------------------
 try:
-    import sphinxcontrib.blockdiag
-    assert sphinxcontrib.blockdiag
-except ImportError as e:
-    raise RuntimeError("sphinxcontrib.blockdiag is not installed. "
-                       "Please install documentation dependencies with "
-                       "`pip install buildbot[docs]`") from e
-
-try:
     pkg_resources.require('docutils>=0.8')
 except pkg_resources.ResolutionError as e:
     raise RuntimeError("docutils is not installed or has incompatible version. "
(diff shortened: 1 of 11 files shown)


tardyp:
sphinxcontrib-jinja  has just been integrated into conda. So it is actually used by people outside buildbot.

As per blockdiag, at work, we are using plantuml a lot (there are java dependency, which might not be very much better for you..). I really think we should have more block diagrams in the doc, and making them in asciart is not fun, and the result is not so nice. So I would really like to keep a generated diagram method. Another option would be to store the blockdiag result in the git, and to have a custom sphinx plugin which only regenerate if the blockdiag binary is present.

tardyp:
@p12tic, thoughs?

p12tic:
@rjarry: 

> The sphinxcontrib-jinja package could be integrated into debian, but it would take me quite some time. Also, I am not sure the package is used outside buildbot.

Wouldn't this fail the 4.13 of the Debian policy? https://www.debian.org/doc/debian-policy/ch-source.html#convenience-copies-of-code .

I agree with @tardyp regarding the asciidoc. Saving images as PNG in the repository is a better option and we could use whatever tool we want.

rjarry:
OK so,

this means I have to package sphinxcontrib-jinja into debian :)

Also, I am more comfortable with storing .svg files in git. They can be used for HTML docs as-is. I'll see what I can do.

p12tic:
Agreed with regards to svg.

p12tic:
Fixed in https://github.com/buildbot/buildbot/pull/6223.

PR #5175, opened 2020-01-29
docs: Point to a working lib for lazy module reloading

jocelynj:
lazy-reload doesn't work with python3, so replace by importlib from official
library. It can be used by putting following code in master.cfg:
```python
  import importlib
  import traceback
  import master_workers

  try:
    importlib.reload(master_workers)
  except:
    traceback.print_exc()

  # you can then call any function in master_workers
```


jocelynj:
Should the code I've added on the commit message be in the documentation too?

I wasn't sure how to write this, as the lazy-reload sentence is inside a big warning block.

p12tic:
> Should the code I've added on the commit message be in the documentation too?

I think it makes sense. Maybe we could move the text out of the warning block to a separate section, name it something like "Caveats of reloading the configuration" and then add the example there.

PR #5177, opened 2020-01-30
docs: Implement parallel building

p12tic:
The user can now set SPHINXOPTS="-W -j8" to build documentation several times faster. For some reason, increasing the number of cores past 8 there's no improvement. Perhaps we have pages that take long to process and which block all the others.

### DIFF
--- a/master/docs/bbdocs/ext.py
+++ b/master/docs/bbdocs/ext.py
@@ -196,6 +196,7 @@ class BBDomain(Domain):
         'event': ObjType('event', 'event'),
         'rtype': ObjType('rtype', 'rtype'),
         'rpath': ObjType('rpath', 'rpath'),
+        'raction': ObjType('raction', 'raction'),
     }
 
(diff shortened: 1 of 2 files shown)


tardyp:
shall we use it in the CI as well? to make sure this does not regress

I would even switch the makefile to use that sphinxopt by default.

p12tic:
I think this should not be enabled by default just yet. I saw one of the indices being not generated sometimes. It's not a problem if the developer just wants to regenerate the documentation quickly during development, but for any release artifacts it makes sense to keep single-thread building for now.

Also, if we think to merge #5180, then we need to wait for https://github.com/readthedocs/sphinx_rtd_theme/pull/874.

p12tic:
If we want to merge #5180, then we need to wait for https://github.com/readthedocs/sphinx_rtd_theme/pull/874 and it being released before we can enable parallel building by default.

PR #5182, opened 2020-01-31
buildbot-www: remove unneeded dependencies from setup.py

yan12125:
They were added in https://github.com/buildbot/buildbot/pull/1209.

### Motivation

I maintain buildbot-related packages for Arch Linux. In buildbot [1], some tests need buildbot-www. I'd like to get rid of circular dependencies (buildbot -> buildbot-www -> buildbot), which makes packaging complex.

If I understand the situation correctly, those Python packages are used only during testing, not building. I propose to remove those dependencies to make packaging easier.

[1] https://www.archlinux.org/packages/community/any/buildbot/

## Contributor Checklist:

(I consider this is a minor change that does not need an announcement)

### DIFF
--- a/www/base/setup.py
+++ b/www/base/setup.py
@@ -17,13 +17,11 @@
 
 try:
     from buildbot_pkg import setup_www_plugin
-    import mock  # noqa
-    import buildbot  # noqa
 except ImportError:
     import sys
-    print('Please install buildbot, buildbot_pkg, and mock modules in order '
-          'to install that package, or use the pre-build .whl modules '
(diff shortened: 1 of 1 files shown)


yan12125:
Indeed. How about keeping buildbot but moving it to install_requires?

p12tic:
Makes sense to me. @tardyp what do you think?

p12tic:
Couldn't find a dependency on regular buildbot when building buildbot-www. Pierre currently is busy, so let's just merge the PR.

Thanks @yan12125

Issue #5164, opened 2020-01-21
Feature discussion: Store and view the results of various tests


p12tic:
Having test results being shown somewhere easily accessible from a build page would be a great productivity boost. Currently people need to dig into the logs. If we tracked what tests fail on what builds, we would have a great deal of useful information that would save people a lot of time.

The feature should implement the following use cases:

 - Code issues:
    - Analyze logs produced by  e.g. pylint or flake8
    - Submit review comments on third-party code browser such as GitHub, Gitlab, BitBucket, etc.
    - Store the test results in the database for analytics purposes

  - Test pass/fail results
    - Analyze test pass/fail logs produced by various testing frameworks
    - Track which tests fail since what commit. That would allow people to immediately know which commits are to blame. Further tooling on the BuildBot side could automatically start tests to bisect the failure.
    - Store the test results in the database for analytics purposes. E.g. this would allow to detect an area of the most unstable tests and give information about the root cause.

  - Test quantitative results
    - This use case covers any interesting numeric information coming from a test or some analysis tool. For example, performance metrics, binary size or memory usage metrics categorized by file or code area and so on.
    - Analyze logs produced by various testing frameworks or other tooling
    - Track the results of tests across time which allows detection of performance regressions.

p12tic:
### Database schema

(will be edited as the discussion goes on)

The use cases that need to be supported by the feature have a lot in common, but at the same time there are significant differences of what the users of BuildBot could reasonably expect. This requires the design to be generic enough.

The following is proposed pseudo-schema of the new tables. The schema is slightly denormalized so that performing queries does not introduce too much table joins. In the schema below, `pk` is primary key, `fk` is foreign key.

```
TestResultSet table:
 - (maybe) project_id (int, fk)
 - builder_id (int, fk)
 - build_id (int, fk)
 - step_id (int, fk)
 - testresultset_id (int, pk)
 - testresultset_type (str)
 - testresultset_value_unit (str)
```

[...]

PR #5183, opened 2020-01-31
WIP: Implement APIs for test results functionality

p12tic:
This is a WIP PR with the addition of database schema and REST API specification for the test results functionality described in #5164.

Depends on #5178 for documentation building to work.

### DIFF
--- a/master/buildbot/db/migrate/versions/056_add_test_result_tables.py
+++ b/master/buildbot/db/migrate/versions/056_add_test_result_tables.py
@@ -1,0 +1,163 @@
+# This file is part of Buildbot. Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+
+import sqlalchemy as sa
+
+from buildbot.util import sautils
+
+
+def upgrade(migrate_engine):
+    metadata = sa.MetaData()
+    metadata.bind = migrate_engine
+
+    sautils.Table(
+        'builds', metadata,
+        sa.Column('id', sa.Integer, primary_key=True),
+        # ...
+    )
+
+    sautils.Table(
+        'builders', metadata,
+        sa.Column('id', sa.Integer, primary_key=True),
+        # ...
+    )
+
+    steps = sautils.Table(
+        'steps', metadata,
+        sa.Column('id', sa.Integer, primary_key=True),
+        # ...
+    )
+
+    test_result_sets = sautils.Table(
+        'test_result_sets', metadata,
+
+        sa.Column('id', sa.Integer, primary_key=True),
+
+        sa.Column('builderid', sa.Integer,
+                  sa.ForeignKey('builders.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('buildid', sa.Integer,
+                  sa.ForeignKey('builds.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('stepid', sa.Integer,
+                  sa.ForeignKey('steps.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('category', sa.Text, nullable=False),
+
+        sa.Column('value_unit', sa.Text, nullable=False),
+
+        sa.Column('complete', sa.SmallInteger, nullable=False),
+    )
+
+    test_result_unparsed_sets = sautils.Table(
+        'test_result_unparsed_sets', metadata,
+
+        sa.Column('test_result_setid', sa.Integer,
+                  sa.ForeignKey('test_result_sets.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('status', sa.SmallInteger, nullable=False),
+    )
+
+    test_results = sautils.Table(
+        'test_results', metadata,
+
+        sa.Column('id', sa.Integer, primary_key=True),
+
+        sa.Column('builderid', sa.Integer,
+                  sa.ForeignKey('builders.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('test_result_setid', sa.Integer,
+                  sa.ForeignKey('test_result_sets.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('test_nameid', sa.Integer,
+                  sa.ForeignKey('test_names.id', ondelete='CASCADE'),
+                  nullable=True),
+
+        sa.Column('test_code_pathid', sa.Integer,
+                  sa.ForeignKey('test_code_paths.id', ondelete='CASCADE'),
+                  nullable=True),
+
+        sa.Column('line', sa.Integer, nullable=True),
+
+        sa.Column('value', sa.Text, nullable=False),
+    )
+
+    test_names = sautils.Table(
+        'test_names', metadata,
+
+        sa.Column('id', sa.Integer, primary_key=True),
+
+        sa.Column('builderid', sa.Integer,
+                  sa.ForeignKey('builders.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('name', sa.Text, nullable=False),
+    )
+
+    test_code_paths = sautils.Table(
+        'test_code_paths', metadata,
+
+        sa.Column('id', sa.Integer, primary_key=True),
+
+        sa.Column('builderid', sa.Integer,
+                  sa.ForeignKey('builders.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('path', sa.Text, nullable=False),
+    )
+
+    test_raw_results = sautils.Table(
+        'test_raw_results', metadata,
+
+        sa.Column('id', sa.Integer, primary_key=True),
+
+        sa.Column('test_result_setid', sa.Integer,
+                  sa.ForeignKey('test_result_sets.id', ondelete='CASCADE'),
+                  nullable=False),
+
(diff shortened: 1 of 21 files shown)


master/buildbot/db/migrate/versions/056_add_test_result_tables.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+    test_result_unparsed_sets = sautils.Table(
+        'test_result_unparsed_sets', metadata,
+
+        sa.Column('test_result_setid', sa.Integer,
+                  sa.ForeignKey('test_result_sets.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('status', sa.SmallInteger, nullable=False),
```
tardyp: I am not sure if this is really worth to store the unparsed results in the DB.

I think it makes more sense to first concentrate on the API and user experience, and then focus on the performance if needed.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
```diff
@@ -0,0 +1,163 @@
...
+    test_result_sets = sautils.Table(
+        'test_result_sets', metadata,
+
+        sa.Column('id', sa.Integer, primary_key=True),
+
+        sa.Column('builderid', sa.Integer,
+                  sa.ForeignKey('builders.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('buildid', sa.Integer,
```
tardyp: lets document here the reason why we want to denormalize the data model.

master/buildbot/spec/types/test_raw_result_chunk.raml
```diff
@@ -0,0 +1,31 @@
+#%RAML 1.0 DataType
+displayName: Test raw result chunk.
+description: |
+    This resource represents a part of test raw result.
+    All chunks for a sequence with the ``sequence`` value increasing sequentially starting from 0.
```
tardyp: not sure what this sequence number is about.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: I strongly believe we should look into performance part from the start in case of this feature, because even relatively small projects may generate huge number of test results. Poor performance will quickly translate to abysmal user experience. I've seen this first hard - large number of test results times large number of tests in the database means minutes of waiting for the web frontend to respond.

Even in the case of Buildbot - if we want to parse all test results, each builder will generate 5000 results. We have around 5 such builders - which means 25000 test results per complete test run. Each test result takes around 100 bytes of space. This means that if there's 3 complete test runs per day, we will use almost 3GB of database space per year. Once this does not properly fit into memory the Buildbot will see bad user experience for certain queries.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: It's documented, see https://github.com/buildbot/buildbot/pull/5183/files#diff-58b3908894b0546af761475a0243ffd1R622

master/buildbot/spec/types/test_raw_result_chunk.raml
p12tic: This mimics the logchunk. In that case it contains the first line that the log chunk corresponds to. In the case of raw result chunk, we still want some way to identify specific result chunks.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: Recarding this comment:

> I am not sure if this is really worth to store the unparsed results in the DB.

The reason for that is that I don't want to constrain Buildbot with the need to parse test results immediately in the master. If we store the test results in the DB, we can parse them whenever we want and using whatever tools we want. My idea is to do this outside the master process. Since all the needed data is in the DB, this will be really easy to do.

tardyp: I was more talking about parsing performance than db performance. We should experiment a bit with the junit parser, and see how much time is needed to parse 5k tests results.

Here we focus on parsing, but I agree with you that the worth case could indeed be reading/querying this huge db.

BTW, gitlab is also introducing this feature, and are struggling with perf.

https://docs.gitlab.com/ee/ci/junit_test_reports.html#enabling-the-feature

Their approach seems to do all the parsing in the web client, which is quite interesting..
https://gitlab.com/gitlab-org/gitlab/issues/37725
This suggests that a 27k tests junit takes 14s to parse in the browser. I think this is mixing up some rending, 14s seems huge (2ms per test)

> Each test result takes around 100 bytes of space
current is 7 fields, that is 48 int bytes + value text.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
```diff
@@ -0,0 +1,163 @@
...
+
+        sa.Column('test_result_setid', sa.Integer,
+                  sa.ForeignKey('test_result_sets.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('test_nameid', sa.Integer,
+                  sa.ForeignKey('test_names.id', ondelete='CASCADE'),
+                  nullable=True),
+
+        sa.Column('test_code_pathid', sa.Integer,
```
tardyp: we may want to separate the test cases, which have no code associated, but have huge number for large test suites.

 and the static analysis test errors which are bound to code path, and have much less results per run.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
```diff
@@ -0,0 +1,163 @@
...
+
+        sa.Column('stepid', sa.Integer,
+                  sa.ForeignKey('steps.id', ondelete='CASCADE'),
+                  nullable=False),
+
+        sa.Column('category', sa.Text, nullable=False),
+
+        sa.Column('value_unit', sa.Text, nullable=False),
+
+        sa.Column('complete', sa.SmallInteger, nullable=False),
```
tardyp: we may want to denormalize the pass/fail rate there, or some global result (that needs to be plot over time)

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: My main worry with parsing performance was that if the raw testing result comes faster than the master can parse it, then we can severely reduce the performance of the master for these hundreds of milliseconds or even seconds during which the result is parsed. We then need some way to have an internal queue in which the raw result is stored and which the master drains bit by bit. Storing this into the database is the easiest and then we have the ability to do parsing whenever and wherever we want.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: Agreed. We can just have `tests_passed` and `tests_failed` columns for the `value_unit` which supports this. Performance tests will not fill these columns, but any pass/fail tests will.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: > Each test result takes around 100 bytes of space

> current is 7 fields, that is 48 int bytes + value text.

I was thinking more in terms of total row size. On postgresql per-row overhead is 28 bytes, so adding all of this we'll have around 80-100 bytes per row. On some cases this will be less, for example, if there's some fields are NULL then they won't occupy any space at all.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: I was thinking about doing exactly this when doing the initial design but arrived to the conclusion that a common table is best. The database engine can optimize stuff quite a lot. For example, on postgresql NULL fields will be completely free due to the NULL bitmap. Having a second table could potentially just increase the code complexity without scalability benefits.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
tardyp: > My main worry with parsing performance was that if the raw testing result comes faster than the master can parse it

What I would expect is that we should have specialised steps to parse test results. Those tests will stream test output from the worker, and parse it on the flight (like logobserver). You might not even have to defer the parsing to another thread if you work like this. If the master is starving for cpu, then this will just delay the whole reactor, and then the user will just have to setup multimaster to gain more speed.

If we write it like this I don't think we should have exploding behaviour you describe. The builds will just slow down a bit.

quick rule of thumbs:
https://stackoverflow.com/questions/4922867/what-is-the-junit-xml-format-specification-that-hudson-supports

this 3 test example is ~70 bytes per test to parse

http://elektito.com/2017/08/25/benchmarking-python-xml-parsers/

order of idea of best parser is 10MB/s

That is ~147000 tests parsed per second per master.

now I am worried if the pg database will be able to take that load.

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
tardyp: alright

master/buildbot/db/migrate/versions/056_add_test_result_tables.py
p12tic: 70 bytes per test is I think the best case. BenchmarkDotNet for example uses more than 2kB per test case. Only part that information is useful for us, but we will still need to parse it. A thousand of such tests will take 200ms just to parse on the master side. In larger projects that focus on performance there will be many thousands of performance tests.

At the moment I don't think multimaster is the solution that's acceptable for the user because it's not production ready and we don't have any plans to make it so. There are plenty of rough corners in multimaster. More complex lock setups won't work for example. It also requires significant user-side time investment in understanding internals of Buildbot and configuring everything in a right way. Worse, I don't think we can even make multimaster easy to configure due to its design. The user will always need a separate configuration for each additional master.

For the end user it's just easier setting up a competing CI vendor for any larger CI installation. Therefore I think that at the current situation it's best if we make fewer reasons to need to use the multimaster, not more. 

I think a better plan would be to utilize the fact that nowadays you can just buy a machine with 128 threads for reasonable amount of money. We don't need the complexity of multimaster to run processes at the same machine. The master can easily manage the lifetime of these processes, so as long as we have a common data access layer we can run any CPU-intensive task as a separate process. This way we could potentially scale the performance of the current master design at least 100 times before needing real multi-master. Maybe we could even run stuff like www API workers as separate processes, that needs more thought, but for jobs that don't really need to be executed immediately running them as separate processes is the best solution I think.

p12tic:
Closed in favor of #5254.

PR #5184, opened 2020-02-01
Add `count` to locks

PeterSurda:
- allows lock access to use up a variable amount of units
- replaces the `_skipChecks` argument from `LockAccess` as that's obsolete
- defaults to `1` for backwards compatibility
- count `0` has the same effect as not using the lock at all
- old tests have been made compatible and pass, no new tests have been written
  yet

## Contributor Checklist:

### DIFF
--- a/master/buildbot/locks.py
+++ b/master/buildbot/locks.py
@@ -88,6 +88,9 @@ class BaseLock:
                                                                  repr(self.owners)))
         num_excl, num_counting = self._claimed_excl, self._claimed_counting
 
+        if not access.count:
+            return True
+
         w_index = self._find_waiting(requester)
         if w_index is None:
             w_index = len(self.waiting)
@@ -96,20 +99,21 @@ class BaseLock:
 
         if access.mode == 'counting':
             # Wants counting access
-            return num_excl == 0 and num_counting + len(ahead) < self.maxCount \
+            return not num_excl \
+                and num_counting + len(ahead) + access.count <= self.maxCount \
                 and all([w[1].mode == 'counting' for w in ahead])
         # else Wants exclusive access
-        return num_excl == 0 and num_counting == 0 and not ahead
+        return not num_excl and not num_counting and not ahead
 
     def _addOwner(self, owner, access):
         self.owners.append((owner, access))
         if access.mode == 'counting':
-            self._claimed_counting += 1
+            self._claimed_counting += access.count
         else:
             self._claimed_excl += 1
 
-        assert (self._claimed_excl == 1 and self._claimed_counting == 0) \
-            or (self._claimed_excl == 0 and self._claimed_excl <= self.maxCount)
+        assert (self._claimed_excl and not self._claimed_counting) \
+            or (not self._claimed_excl and self._claimed_excl <= self.maxCount)
 
     def _removeOwner(self, owner, access):
         # returns True if owner removed, False if the lock has been already
@@ -120,7 +124,7 @@ class BaseLock:
 
         self.owners.remove(entry)
         if access.mode == 'counting':
-            self._claimed_counting -= 1
+            self._claimed_counting -= access.count
         else:
             self._claimed_excl -= 1
         return True
@@ -133,10 +137,19 @@ class BaseLock:
 
         assert isinstance(access, LockAccess)
         assert access.mode in ['counting', 'exclusive']
+        assert isinstance(access.count, int)
+        if access.mode == 'exclusive':
+            assert access.count == 1
+        else:
+            assert access.count >= 0
+        if not access.count:
+            return
+
         self.waiting = [w for w in self.waiting if w[0] is not owner]
         self._addOwner(owner, access)
 
-        debuglog(" {} is claimed '{}'".format(self, access.mode))
+        debuglog(" {} is claimed '{}', {} units".format(self, access.mode,
+                  access.count))
 
     def subscribeToReleases(self, callback):
         """Schedule C{callback} to be invoked every time this lock is
@@ -147,7 +160,11 @@ class BaseLock:
         """ Release the lock """
         assert isinstance(access, LockAccess)
 
-        debuglog("{} release({}, {})".format(self, owner, access.mode))
+        if not access.count:
+            return
+
+        debuglog("{} release({}, {}, {})".format(self, owner, access.mode,
+                  access.count))
         if not self._removeOwner(owner, access):
             debuglog("{} already released".format(self))
             return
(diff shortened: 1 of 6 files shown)


master/buildbot/locks.py
```diff
@@ -96,17 +96,19 @@ def isAvailable(self, requester, access):
 
         if access.mode == 'counting':
             # Wants counting access
-            return num_excl == 0 and num_counting + len(ahead) < self.maxCount \
+            return num_excl <= 1 - access.count \
```
p12tic: Here we should count the total of the lock counts in the queue ahead of us. We'll also need to have a unit test to ensure it works correctly.

master/buildbot/locks.py
```diff
@@ -96,17 +96,19 @@ def isAvailable(self, requester, access):
...
+            and not (ahead or not access.count)
 
     def _addOwner(self, owner, access):
         self.owners.append((owner, access))
         if access.mode == 'counting':
-            self._claimed_counting += 1
+            self._claimed_counting += access.count
         else:
-            self._claimed_excl += 1
+            self._claimed_excl += access.count
```
p12tic: I think it makes sense to not allow any other lock count than 1 for exclusive access. During construction of lock access, we should verify that correct access count has been passed.

master/buildbot/locks.py
```diff
@@ -320,19 +328,23 @@ class LockAccess(util.ComparableMixin):
...
+
+    @param count: How many units does the access occupy
+    @type  count: Integer, not negative, default is 1 for backwards
+                  compatibility
     """
 
-    compare_attrs = ('lockid', 'mode')
+    compare_attrs = ('lockid', 'mode', 'count')
 
-    def __init__(self, lockid, mode, _skipChecks=False):
```
p12tic: I'd like to see _skipChecks removal to be moved to a separate commit. If it causes random test failures in the future, it will be possible to bisect the problem to that commit.

master/buildbot/locks.py
```diff
@@ -320,19 +328,23 @@ class LockAccess(util.ComparableMixin):
...
 
-        if not _skipChecks:
-            # these checks fail with mock < 0.8.0 when lockid is a Mock
-            # TODO: remove this in Buildbot-0.9.0+
-            assert isinstance(lockid, (MasterLock, WorkerLock))
-            assert mode in ['counting', 'exclusive']
+        assert isinstance(lockid, (MasterLock, WorkerLock))
+        assert mode in ['counting', 'exclusive']
+        assert isinstance(count, int)
+        assert count >= 0
```
p12tic: I think it makes sense to not allow any other lock count than 1 for exclusive access. Exclusive access means that a single lock access can acquire the lock, so the lock count does not change the behavior. Best to throw exception if count != 1 in exclusive mode.

master/buildbot/locks.py
```diff
@@ -164,14 +170,16 @@ def _tryWakeUp(self):
         num_excl, num_counting = self._claimed_excl, self._claimed_counting
         for i, (w_owner, w_access, d) in enumerate(self.waiting):
             if w_access.mode == 'counting':
-                if num_excl > 0 or num_counting >= self.maxCount:
+                if (num_excl > 0 or num_counting >= self.maxCount) \
+                   and w_access.count > 0:
```
p12tic: I think it makes sense to completely short-circuit whole lock logic when dealing with lock accesses with 0 lock count. Just make isAvailable() return true, and `claim()` and `release()` return at the beginning.

PeterSurda:
Thanks, these are all excellent suggestions, I'll keep working on it incrementally and let you know when it's ready for a next round of reviews.

master/buildbot/locks.py
```diff
@@ -88,6 +88,9 @@ def isAvailable(self, requester, access):
                  % (self, requester, access, self.owners))
         num_excl, num_counting = self._claimed_excl, self._claimed_counting
 
+        if access.count == 0:
```
sa2ajj:
```!python
if not access.count:
```
?

master/buildbot/locks.py
```diff
@@ -96,15 +99,16 @@ def isAvailable(self, requester, access):
 
         if access.mode == 'counting':
             # Wants counting access
-            return num_excl == 0 and num_counting + len(ahead) < self.maxCount \
+            return num_excl == 0 \
```
sa2ajj:
```!python
return not num_excl \
...
```
?

master/buildbot/locks.py
```diff
@@ -96,15 +99,16 @@ def isAvailable(self, requester, access):
...
+                and num_counting + len(ahead) + access.count <= self.maxCount \
                 and all([w[1].mode == 'counting' for w in ahead])
         # else Wants exclusive access
         return num_excl == 0 and num_counting == 0 and not ahead
```
sa2ajj: that's already existing code, but the comment still applies

sa2ajj: and here...

master/buildbot/locks.py
```diff
@@ -133,10 +137,19 @@ def claim(self, owner, access):
 
         assert isinstance(access, LockAccess)
         assert access.mode in ['counting', 'exclusive']
+        assert isinstance(access.count, int)
+        if access.mode == 'exclusive':
+            assert access.count == 1
+        else:
+            assert access.count >= 0
+        if access.count == 0:
```
sa2ajj: looks like there's a lot of similar patterns in this change

master/buildbot/locks.py
PeterSurda: you're right

master/buildbot/locks.py
PeterSurda: yep

master/buildbot/locks.py
PeterSurda: ok

PeterSurda: ok

master/buildbot/locks.py
PeterSurda: ok

master/buildbot/locks.py
PeterSurda: I hope it's ok like the latest change (apart from tests).

master/buildbot/locks.py
PeterSurda: I changed it accordingly, let me know if that's ok.

master/buildbot/locks.py
PeterSurda: done

master/buildbot/locks.py
PeterSurda: done

master/buildbot/locks.py
PeterSurda: done

master/buildbot/locks.py
PeterSurda: done

PeterSurda:
I amended the tests to work with `count` greater than `1`. I will now add new tests that behave differently.

PeterSurda:
Should be ready to review now. I also updated documentation, added new tests and newsfragments.

PR #5187, opened 2020-02-03
Rewrite BitbucketPullRequestPoller

chapipo:
- Use httpclient
- Support Oauth2 authentification

I removed some arguments of the poller (project, useTimestamp) because I was not sure about how they have to work. If needed I will work to reintroduce them.

## Contributor Checklist:


p12tic:
Could we split the changes into separate commits so that it's possible to review it? E.g. the httpclient rewrite could go first and potentially as a separate PR.

chapipo:
Will try to do this but the httpclient rewrite is the biggest :)

master/buildbot/changes/bitbucket.py
```diff
@@ -14,184 +14,282 @@
...
-                 pollInterval=10 * 60,
-                 useTimestamps=True,
-                 category=None,
-                 project='',
-                 pullrequest_filter=True,
-                 encoding=_UNSPECIFIED,
-                 pollAtLaunch=False
-                 ):
+    def __init__(self, owner, slug, **kwargs):
+        name = kwargs.get("name")
```
sa2ajj: two comments about these lines:
* could we use `kwargs.setdefault("name", ...)` instead?
* could you please replace string concatenations with a string interpolation (`"xxx {} xxx {} xxx`.format(...)`)? interpolations are easier to read and easier to review when changes are made

master/buildbot/changes/bitbucket.py
```diff
@@ -14,184 +14,282 @@
...
+                    pullrequest_filter=True,
+                    codebase='',
+                    oauth_key=None,
+                    oauth_secret=None,
+                    repository_type="https",
+                    debug=False,
+                    **kwargs):
+        if repository_type not in ["https", "ssh"]:
+            config.error(
+                "repository_type must be one of {https, svn, git, ssh}")
```
sa2ajj: i am not sure how the error message corresponds to the condition... please add a comment clarifying the relationship or change the condition and/or the message

master/buildbot/changes/bitbucket.py
```diff
@@ -14,184 +14,282 @@
...
+                        codebase='',
+                        oauth_key=None,
+                        oauth_secret=None,
+                        debug=False,
+                        repository_type="https",
+                        **kwargs):
+
+        yield super().reconfigService(name=self.name, **kwargs)
+
+        http_headers = {'User-Agent': 'Buildbot'}
```
sa2ajj: are you sure that the set of headers is not going to increase?  if not, please re-format as
```
http_headers = {
    'User-Agent': 'Buildbot'.
}
```

PR #5189, opened 2020-02-04
GitHubAuth: uses Authentication header instead of query params

tardyp:
query param is deprecated way of authenticating api with github.

## Contributor Checklist:

* [NA] I have updated the appropriate documentation


master/buildbot/newsfragments/github_oauth.bugfix
```diff
@@ -0,0 +1 @@
+:class:`GitHubAuth` is not using `Authorization` headers instead of `access_token` query parameter, as the later was deprecated by Github. (:issue:`5188`)
```
djmitche:
```suggestion
:class:`GitHubAuth` is now using `Authorization` headers instead of `access_token` query parameter, as the latter was deprecated by Github. (:issue:`5188`)
```

PR #5194, opened 2020-02-11
www-badges: Add badgeio styled SVG

jbenden:
This patch introduces a SVG template to match the look-and-feel of
badge.io badges.

In order to effectively mimic the look, some of the parameters
required tweaks, and the character font width calculation needed to be
corrected.

A running example of this change is located on the `README.md` of the [Aircrack-ng project](https://github.com/aircrack-ng/aircrack-ng/).

PS: I'm not sure if I need to add a file in `master/buildbot/newsfragments`. Nor am I sure if I need to add docs.


www/badges/buildbot_badges/__init__.py
```diff
@@ -31,12 +31,12 @@ class Api:
     app = Klein()
 
     default = {  # note that these defaults are documented in configuration/www.rst
-        "left_pad": 5,
+        "left_pad": 0,
         "left_text": "Build Status",
         "left_color": "#555",
-        "right_pad": 5,
-        "border_radius": 5,
-        "style": "plastic",
```
tardyp: why change the default? please document in www.rst example configuration with required default changes.

jbenden: You're probably right, I shouldn't have changed the defaults. I've reverted those changes, while adding a documentation blurb with image.

jbenden: PS: The font style change is really required, as all templates have incorrect character widths calculated; due to the mismatch in font weights used (bold for calculation and normal weight for display.)

jbenden:
I'm not sure how to interpret the remaining CI failures. Please advise.

Thanks,
-Joe

p12tic:
@jocelynj Looks great. Could you add a newsfragment describing this PR? See master/buildbot/newsfragments directory.

Issue #4247, opened 2018-08-03
`buildbot/buildbot-master:latest` is out of date


terencehonles:
I looked into automated builds on docker hub and it appears this setting wouldn't be something I could create a PR for (it's not like the "official images"), but can we add an additional build hook which is the same as the `v.*` tags, but updates the `latest` tag.

I'd like to pull from the most recent released version rather than master, (My current setup is `FROM buildbot/buildbot-master:master`) and I don't really want to have to be updating the Dockerfile.

This would also help with confusion when initially I had `FROM buildbot/buildbot-master` and docker defaulted to the `latest` tag which was not actually what I wanted.

p12tic:
tardyp: Would it make sense to alter the release scripts to point `latest` to the latest release tag, e.g. `1.7.0` currently?

jared-hardy:
@tardyp shouldn't this :latest tag always keep up with the *latest stable* version of the docker container, currently tagged `v2.3.0`, which I believe is standard practice with all docker containers?

terencehonles:
I think it varies from package to package. Since we already have a `master` tag, it would probably make sense to have `latest` point to the most recent stable version as you both are suggesting.

I mainly wanted to point out that if you were to run `docker run buildbot/buildbot-master` it will default to `latest` which is very out of date rather than `master` which also would probably work, but obviously isn't a stable release.

terencehonles:
@p12tic was there any more thoughts on this? It's been awhile, and I would like to use a `latest` tag so I can auto-build and verify a new buildbot setup rather than manually check for a new stable tag or manually verify I'm ok with `master`.

Not sure how much work this would be on your end, but I would hope it would just be that when a new release is pushed you also push to `latest` (so wherever that change would belong).

cc: @tardyp since both of you pinged them on your messages.

terencehonles:
Looking at the buildbot/buildbot-worker tag list it seems like that might be unmaintained. I'm not sure what latest is pointing to and the more recent versioned tag I see is v2.1.0. I was kinda expecting to see [v2.4.0](https://github.com/buildbot/buildbot/releases/tag/v2.4.0) which calls out bumping the worker to a new Ubuntu version.

Do you guys want help getting this configured? I can try to help, but I don't know where to look.

[...]

PR #5195, opened 2020-02-11
master: Fix GitHub "User Agent" required

terencehonles:
When updating buildbot to include #5189 (31c2caae10f9) GitHub returns and error that a user agent is required. This change updates the requests session to include the User-Agent: buildbot/$buildbot_version

See: https://developer.github.com/v3/#user-agent-required

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/oauth2.py
+++ b/master/buildbot/www/oauth2.py
@@ -26,6 +26,7 @@ import requests
 from twisted.internet import defer
 from twisted.internet import threads
 
+import buildbot
 from buildbot import config
 from buildbot.process.properties import Properties
 from buildbot.util import bytes2unicode
@@ -286,7 +287,10 @@ class GitHubAuth(OAuth2Auth):
 
     def createSessionFromToken(self, token):
         s = requests.Session()
-        s.headers = {'Authorization': 'token ' + token['access_token']}
+        s.headers = {
+            'Authorization': 'token ' + token['access_token'],
+            'User-Agent': 'buildbot/%s' % buildbot.version,
+        }
         s.verify = self.sslVerify
         return s
 
--- a/master/buildbot/test/unit/test_www_oauth.py
+++ b/master/buildbot/test/unit/test_www_oauth.py
@@ -28,6 +28,7 @@ from twisted.trial import unittest
 from twisted.web.resource import Resource
 from twisted.web.server import Site
 
+import buildbot
 from buildbot.process.properties import Secret
 from buildbot.secrets.manager import SecretManager
 from buildbot.test.fake.secrets import FakeSecretStorage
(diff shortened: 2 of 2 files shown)


terencehonles:
I don't know if any news fragment is needed because there hasn't been a release since #5189 and this PR should be considered an update to it.

master/buildbot/www/oauth2.py
```diff
@@ -26,7 +26,7 @@
 from twisted.internet import defer
 from twisted.internet import threads
 
-from buildbot import config
+from buildbot import config, version as buildbot_version
```
p12tic: We use one import per line.

terencehonles:
@p12tic do you mind bumping the `bb` runner? Not sure what happened, but looking at my history I see https://buildbot.buildbot.net/#/builders/2/builds/1444 doesn't have the right revision, and I think someone with the correct perms can just hit "rebuild"

p12tic:
@terencehonles Looks like the Buildbot-Github integration has broken yesterday for some reason, I can't even login.

p12tic:
Unit tests pass locally, so we should just merge this PR to get out of a chicken-and-egg situation.

tardyp:
patch applied in prod, and bot restarted..

tardyp:
shame on ui @github for making such a backward incompatible change..

p12tic:
I guess it makes sense to expedite the release of v2.7. What was the information in the emails from Github, when do they disable the old API?

tardyp:
I agree we should release soon.

The email from Github was not for User-Agent, just for Authorization header versus query param.

p12tic:
I understand, but was there any timeline about the deprecation of the Authorization query param?

tardyp:
> Hello there!
> 
> On February 4th, 2020 at 09:12 (UTC) your application (metabuildbot) used an access token (with the User-Agent python-requests/2.21.0) as part of a query parameter to access an endpoint through the GitHub API.
> 
> https://api.github.com/user
> 
> Please use the Authorization HTTP header instead as using the `access_token` query parameter is deprecated.
> 
> Depending on your API usage, we'll be sending you this email reminder once every 3 days for each token and User-Agent used in API calls made on your behalf.
> Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.
> 
> Visit https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters for more information.
> 
> Thanks,
> The GitHub Team
>

tardyp:
so no timeline..

p12tic:
According to https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/ there will be brownouts starting by May 15, 2020 and the API will be disabled on July 1, 2020.

terencehonles:
Thanks guys, I'll rebuild my docker image now :)

And if I might point you to my issue #4247, not sure if you want help maintaining the docker images.

PR #5199, opened 2020-02-15
Add LineProcessProtocol and use it

p12tic:
This slightly simplifies the codebase by no longer needing to implement line handling in various places.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/util/protocol.py
+++ b/master/buildbot/util/protocol.py
@@ -1,0 +1,79 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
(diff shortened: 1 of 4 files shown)


master/buildbot/util/protocol.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class LineBuffer:
+    def __init__(self, delimiter):
+        self._buffer = b''
+        self._delimiter = delimiter
+
+    def add_data(self, data):
+        # returns lines that have been processed, if any
+        lines = (self._buffer + data).split(self._delimiter)
```
tardyp: you should rather use a generator for this. this is more scalable and pythonic

master/buildbot/util/protocol.py
```diff
@@ -0,0 +1,71 @@
...
+        self._buffer = lines.pop(-1)
+        return [l.rstrip(b'\r') for l in lines]
+
+    def get_trailing_lines(self):
```
tardyp:
```suggestion
    def get_trailing_line(self):
```
there is only one line..

master/buildbot/util/protocol.py
p12tic: Fixed.

master/buildbot/util/protocol.py
p12tic: Fixed.

p12tic:
@tardyp: This code is specific to Twisted ProcessProtocol. LogLineObserver has a little different requirements, so it does for example work only with decoded strings and does not handle lines that arrive split in the middle.

Issue #5200, opened 2020-02-17
Traceback from worstStatus


rpurdie:
We keep seeing a traceback when we reconfig a running controller. This is usually when a build is sitting waiting on triggered jobs, the triggering job then throws an exception with a traceback:

2020-02-17 12:52:11+0000 [-] BuildStep.failed; traceback follows
	Traceback (most recent call last):
	  File "/usr/local/lib/python3.5/dist-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	    current.result = callback(current.result, *args, **kw)
	  File "/usr/local/lib/python3.5/dist-packages/twisted/internet/defer.py", line 1475, in gotResult
	    _inlineCallbacks(r, g, status)
	  File "/usr/local/lib/python3.5/dist-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
	    result = result.throwExceptionIntoGenerator(g)
	  File "/usr/local/lib/python3.5/dist-packages/twisted/python/failure.py", line 512, in throwExceptionIntoGenerator
	    return g.throw(self.type, self.value, self.tb)
	--- <exception caught here> ---
	  File "/usr/local/lib/python3.5/dist-packages/buildbot/process/buildstep.py", line 566, in startStep
	    self.results = yield self.run()
	  File "/usr/local/lib/python3.5/dist-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
	    result = result.throwExceptionIntoGenerator(g)
	  File "/usr/local/lib/python3.5/dist-packages/twisted/python/failure.py", line 512, in throwExceptionIntoGenerator
	    return g.throw(self.type, self.value, self.tb)
	  File "/usr/local/lib/python3.5/dist-packages/buildbot/steps/trigger.py", line 306, in run
	    results = yield self.worstStatus(results, rclist, unimportant_brids)
	  File "/usr/local/lib/python3.5/dist-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
	    result = g.send(result)
	  File "/usr/local/lib/python3.5/dist-packages/buildbot/steps/trigger.py", line 200, in worstStatus
	    if set(unimportant_brids).issuperset(set(brids_dict.values())):
	builtins.UnboundLocalError: local variable 'brids_dict' referenced before assignment

Looking at the code, its right, it can reference that variable before assignment:

https://github.com/buildbot/buildbot/blob/master/master/buildbot/steps/trigger.py#L200

[...]

PR #5205, opened 2020-02-21
steps/trigger: Avoid traceback for UnboundLocalError

rpurdie:
Avoid:

File "buildbot/steps/trigger.py", line 200, in worstStatus
    if set(unimportant_brids).issuperset(set(brids_dict.values())):
builtins.UnboundLocalError: local variable 'brids_dict' referenced before assignment

by rearranging the code so the variable is set.

Fixes issue #5200.


tardyp:
thanks!

PR #5226, opened 2020-03-06
test: Split fakedb module into separate files

p12tic:
The buildbot/test/fake/fakedb.py file is second-largest file within the BuildBot project at 2.6k lines. The size of the file makes it harder to work with it, because the developer needs to understand the structure of the file to make changes without introducing a mess. This PR splits the file into separate files each for one DB component, just like non-fake database code is implemented.


tardyp:
good work!

PR #5231, opened 2020-03-09
requirements: Pin mock because newer version requires Python 3.6

p12tic:
We still support Python 3.5.


tardyp:
yeah python 3.5 support is starting to byte us at work as well. There are a number of projects that start to drop support for 3.5

e.g:
```
mock = "==3.0.5"  # last version to support python 3.5
zipp = "== 1.2.0" # last version to support python 3.5
python-gitlab = "==1.15.0"  # >=2 => python >=3.6
```

problem is that python3.5 is the default python3 for ubuntu 16.03 which is still quite big.

Issue #5207, opened 2020-02-23
REST API wrong behaviour


vladbogo:
Hi,

I have a question regrading the REST API. We are trying to get a list of the last changes for a particular branch. 

However, using https://buildbot.mariadb.org/api/v2/changes?branch=10.1&order=-changeid&limit=10 does not return the last 10 changes for the 10.1 branch. Is this the expected behaviour? Based on the documentation I would assume that this would return the last 10 changes for branch 10.1. If the limit argument is increased, more changes start to appear.

Things get a lot more confusing when deleting the "order" argument https://buildbot.mariadb.org/api/v2/changes?branch=10.1&limit=10 when the request times out and the buildbot is unresponsive for several minutes. Any thoughts on this? Is there any sort of cache used?

Thanks!

vladbogo:
Any updates?

tardyp:
Indeed, this is a bug, we do not support limit and filtering with the current code.

as per this code, the limit is done here, but only when order is -changeid

https://github.com/buildbot/buildbot/blob/81d693a8e973b51552cd23bfa695c4b2b0a30f2c/master/buildbot/data/changes.py#L85

This should rather be implemented like in the builds endpoint:
https://github.com/buildbot/buildbot/blob/81d693a8e973b51552cd23bfa695c4b2b0a30f2c/master/buildbot/data/builds.py#L161
where resultSpec is passed to the db api

and then used with a generic database query generation
https://github.com/buildbot/buildbot/blob/81d693a8e973b51552cd23bfa695c4b2b0a30f2c/master/buildbot/db/builds.py#L147

would you like to fix this code?

vladbogo:
Thanks for your answer. I will look into it and try to fix it.

PR #5246, opened 2020-03-23
Add limit and filtering support for the changes data API

vladbogo:
Correctly support limit and filtering for the changes data API as described in  #5207 .


master/buildbot/db/changes.py
```diff
@@ -242,33 +242,20 @@ def thd(conn):
             return row_uids
         return self.db.pool.do(thd)
 
-    def getRecentChanges(self, count):
```
tardyp: there is developer doc related to that method. please remove.

master/buildbot/reporters/telegram.py
```diff
@@ -262,7 +263,8 @@ def command_LIST(self, args, **kwargs):
                     return
 
             else:
-                changes = yield self.master.db.changes.getRecentChanges(num)
+                rs = resultspec.ResultSpec(order=['-changeid'], limit=num)
```
tardyp: we should use data api instead of db api in reporters.

vladbogo: I assume you mean something like
```suggestion
                changes = yield self.master.data.get(('changes',), order=['-changeid'], limit=num)
```

However, if changing like I suggested, the test_reporters_telegram fails since change['when_timestamp'] is an integer and cannot be converted to a pretty date at line274, change['date'] = change['when_timestamp'].strftime('%Y-%m-%d %H:%M').

This happens due to the conversion from data/changes.py line 37.

What would be the best approach here?

vladbogo: @tardyp any thoughts?

tardyp: hi @vladbogo you can use ```from buildbot.util import epoch2datetime``` in order to convert it back to timestamp.

vladbogo:
@tardyp any thoughts?

PR #5249, opened 2020-04-02
deferwaiter: Add a way to check whether there are waited deferreds

p12tic:
This change allows to check whether calling `yield waiter.wait()` would block.

## Contributor Checklist:


p12tic:
Merging as the change is trivial.

PR #5254, opened 2020-04-05
Implement APIs for test results functionality (v2)

p12tic:
This PR implements the API for the test results functionality described in #5164.

This PR contains:
 - API spec
 - database model
 - data layer implementation
 - integration with BuildStep
 - enhancement of the PyLint step to submit test results for the issues it finds

The test result submission performance is around ~5000 results per second to a postgres database on my desktop. The bottleneck is on the Python side, the database server stays at ~10% CPU usage on my desktop.

## Contributor Checklist:

* [not yet] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1376,6 +1376,7 @@ subdirectory
 subfields
 sublicense
 submittedat
+submitters
 submodule
 submodules
 suboption
--- a/master/buildbot/data/connector.py
+++ b/master/buildbot/data/connector.py
@@ -54,6 +54,8 @@ class DataConnector(service.AsyncService):
         'buildbot.data.forceschedulers',
         'buildbot.data.root',
         'buildbot.data.properties',
+        'buildbot.data.test_results',
+        'buildbot.data.test_result_sets',
     ]
     name = "data"
 
--- a/master/buildbot/data/test_result_sets.py
+++ b/master/buildbot/data/test_result_sets.py
@@ -1,0 +1,142 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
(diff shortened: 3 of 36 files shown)


tardyp:
So what's new with this v2? 
Its hard for me to review as I would need to switch between the two PRs to remember the context and discussion.

p12tic:
Previous PR only had broken API spec and database model, this PR fixes issues there and adds all the upper layers. I agree that perhaps it would have been better idea to continue the previous PR.

tardyp:
This looks great. I am not sure I can really do a thorough review, and I checked a few important point (to me).

- There is APi for batching to the db, which allows to optimize the access.
- The step API is nice and simple, which makes it easy to write parsers
- The data model is pretty flexible, and allows to send KPIs and pass/fail

There is a few api incoherency. There are newTestResults, addTestResults, add_test_results at different point of the stack I would have used addTestResults everywhere. normally, buildbot is compliant to twisted's camelCase role. I personally dislike this choice, and sometime am culprit of neglecting this rule in my buildbot contrib, but as a reviewer, I must point it.

I think we should merge this as (after rebase) is and work on that nitpick in a followup PR.

As a followup PR, I would also suggest to write import and export of xunit2 tests results.
xunit2 is the defacto standard for test report interoperability, and we should make sure that our model is a superset of that.

https://github.com/jenkinsci/xunit-plugin/blob/xunit-2.3.2/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd

https://docs.pytest.org/en/latest/reference.html#confval-junit_family

p12tic:
@tardyp Thanks a lot for the review.

> There is a few api incoherency. There are newTestResults, addTestResults,

This is good catch. I don't remember anymore why I thought the `new*` naming was more prevalent, we use only `add*` in the data layer. Fixed now.

> normally, buildbot is compliant to twisted's camelCase role.
A lot of the BuildBot oode now uses camelCase only for public, documented APIs and snake_case for private ones. I'll create a proposal to gather opinions on using snake_case for private APIs as that would match PEP8 more.

> As a followup PR, I would also suggest to write import and export of xunit2 tests results.
Agreed.

I'll release the next version of BuildBot before merging this PR, so that we have time to spot any issues e.g. with DB migration.

PR #5270, opened 2020-04-28
Add owner/repo_name args to BitbucketStatusPush

dunderpate:
Allow users to specify a custom owner and repo_name for a BitbucketStatusPush
instance. This is useful when we want to push status to a bitbucket instance
even though we are pulling from a different repository. In cases like this,
extracting the owner and/or repo_name from the sourcestamp may not be possible.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/add-owner-repo-to-bitbucketstatuspush.feature
+++ b/master/buildbot/newsfragments/add-owner-repo-to-bitbucketstatuspush.feature
@@ -1,0 +1,1 @@
+Added arguments `owner` and `repo_name` to BitbucketStatusPush for cases the clone URL isn't bitbucket
--- a/master/buildbot/reporters/bitbucket.py
+++ b/master/buildbot/reporters/bitbucket.py
@@ -43,7 +43,13 @@ class BitbucketStatusPush(http.HttpStatusPushBase):
     def reconfigService(self, oauth_key, oauth_secret,
                         base_url=_BASE_URL,
                         oauth_url=_OAUTH_URL,
+                        owner=None,
+                        repo_name=None,
                         **kwargs):
+
+        self.owner = owner
+        self.repo_name = repo_name
+
         oauth_key, oauth_secret = yield self.renderSecrets(oauth_key, oauth_secret)
         yield super().reconfigService(**kwargs)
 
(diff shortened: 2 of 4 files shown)


dunderpate:
Just a heads up, I did patch this into our existing CI pipeline and I can confirm it works as expected.

tardyp:
Code looks good. I am not sure of the necessity of the feature though.

It looks like the only reason you need this is to implement local mirrors.

Did you try to use the git instead of feature in order to access your mirror on the CI machines, rather than the upstream?

```
[url "https://localmirror.example.com/"]
    insteadOf = https://bitbucket.com/myorg/
```

dunderpate:
Hi @tardyp,

You're absolutely right. It's not a very common use-case. I can also see that it sort of defeats the plumbing around codebases and owner/repo extraction. Good points! I can certainly look into and try the `insteadOf` option in git and see if it works for us. That's actually really interesting and I've never heard of it. git never ceases to amaze when it comes to the number of features and edge cases they've considered.

Thanks for your thoughtful reply! I'm totally good with closing this PR if you see fit as you bring up some good reasons why it wouldn't make sense upstream. At least I got some good discussion out of it and learned something new 😄

Thanks again,

Dunderpate

tardyp:
Hi @dunderpate ,
Thanks for understanding. For a first PR, this was very good quality code. That is rarely the case, and it is a shame we cannot merge it.
I am looking forward saying more PR from you! 
Don't hesitate to send an issue first so that we can discuss the options before doing the hard work.

dunderpate:
Hi @tardyp,

Sounds good, I'm a huge fan of the project and would love to contribute more if possible. I'll definitely hit the team up before putting in the effort next time to see if there might be a better approach 👍 Thanks again for the info and long live buildbot 🥳

dunderpate:
Hi @tardyp,

Just an FYI, I did look into that option and it worked perfectly! Would it be helpful if I added a blurb about this in the docs?

Thanks again man 👍

Dunderpate

tardyp:
hi @dunderpate About git mirrors...

I am currently working on open-sourcing the git mirror project that I have been working on for the last 2 years.

https://gitlab.com/grouperenault/git_cdn

Before going wider on the announcement, I would like to make it reviewed by a smaller community. Would you like to take a look and tell whether it would solve your mirroring usecases, and if it is properly usable/documented?

Pierre

Issue #3714, opened 2017-10-24
[ui] Running build show wrong time


pohmelie:
When build is active (yellow) and you click on pulsing bubble, then window below appear. First 500-1000 ms its building time string looks like at picture (and this is exactly negative time since epoch). Looks easy to fix for those who is familiar with codebase.
![screenshot_2017-10-24_18-02-22](https://user-images.githubusercontent.com/2385765/31951103-da95aa58-b8e5-11e7-9237-5f1be76b9471.png)

tardyp:
It is indeed probably not a complicated fix. 
Anybody wanting to take a look, the problem probably is easily fixable here: 

https://github.com/buildbot/buildbot/blob/master/www/base/src/app/common/filters/moment/moment.filter.coffee

nasirhemed:
Can I work on this ?

tardyp:
@nasirhemed yes please! don't hesitate to send a PR.

instruction to start developing JS in buildbot can be found here: 
https://docs.buildbot.net/latest/developer/quickstart.html#hacking-the-buildbot-javascript

nasirhemed:
@tardyp 
This issue has been fixed. I wasn't able to reference this issue.

pohmelie:
You can reference commit message here.

nasirhemed:
Merge pull request #5272 from nasirhemed/bug/ui-negative-epoch

pohmelie:
I can't check this fix, but will close the issue. Feel free to reopen if need.

PR #5272, opened 2020-04-28
[ui] Do not display building time once a build is started

nasirhemed:
Issue: [Running build show wrong time](https://github.com/buildbot/buildbot/issues/3714)  

When build is active (yellow) and you click on pulsing bubble, then window below appear. First 500-1000 ms its building time string looks like at picture (and this is exactly negative time since epoch).

Fixed this by not displaying the building time when the timestamp is negative. 
chrome_GaOPo5rRrD


tardyp:
cool! thanks

mojca:
Thank you!

PR #5283, opened 2020-05-05
Base64 encoding logs/patches in mail reporter

dunderpate:
* This should make emails comformant with RFC 5322 2.1.1
* This should also fix errors that arise when email servers check for
  line lengths > 998.

    * Wasn't quite sure what the implication would be of adding an assertion making sure
        the attachments were base64-encoded given the mixed Python2/3 support.
    * Another test that might be interesting here is to create an email with a line length
        longer than the RFC allows, and then verify that the resulting as_string() call does
        not contain lines longer than the spec. A simple regex should be able to do that.
    * I didn't see a good reason to update the documentation for this patch, feel free to let
       me know if there is an appropriate place for it.

### DIFF
--- a/master/buildbot/newsfragments/base64-encode-logs-and-attachments.bugfix
+++ b/master/buildbot/newsfragments/base64-encode-logs-and-attachments.bugfix
@@ -1,0 +1,1 @@
+Base64 encoding logs and attachments sent via email so emails conform to RFC 5322 2.1.1
--- a/master/buildbot/reporters/mail.py
+++ b/master/buildbot/reporters/mail.py
@@ -15,6 +15,7 @@
 
 import re
 from email import charset
+from email import encoders
 from email.header import Header
 from email.message import Message
 from email.mime.multipart import MIMEMultipart
@@ -170,6 +171,9 @@ class MailNotifier(NotifierBase):
     def patch_to_attachment(self, patch, index):
         # patches are specifically converted to unicode before entering the db
         a = MIMEText(patch['body'].encode(ENCODING), _charset=ENCODING)
+        # convert to base64 to conform with RFC 5322 2.1.1
+        del a['Content-Transfer-Encoding']
+        encoders.encode_base64(a)
         a.add_header('Content-Disposition', "attachment",
                      filename="source patch " + str(index))
         return a
@@ -227,6 +231,9 @@ class MailNotifier(NotifierBase):
                     text = log['content']['content']
                     a = MIMEText(text.encode(ENCODING),
                                  _charset=ENCODING)
+                    # convert to base64 to conform with RFC 5322 2.1.1
+                    del a['Content-Transfer-Encoding']
+                    encoders.encode_base64(a)
                     a.add_header('Content-Disposition', "attachment",
                                  filename=filename)
                     m.attach(a)


p12tic:
@dunderpate: Thanks for the PR. It looks good to me from the first sight.

One think that is not clear and which I could not figure out by cursory looking into RFC 5322: how does the receiving party know that the data is encoded in base64? We only changed the encoding but didn't add / change any headers that could specify the encoding. Could you maybe explain this to me?

dunderpate:
Hi @p12tic,

You bring up a good point! I just double checked a generated email and I think I'm doubling up the Content-Transfer-Encoding headers for each attachment this way. For example:

```
        --===============2846112211774884597==
        Content-Type: text/plain; charset="utf-8"
        MIME-Version: 1.0
        Content-Transfer-Encoding: 7bit
        Content-Transfer-Encoding: base64
        Content-Disposition: attachment; filename="archive.stdio"

        aHB5dGhvbiB0b29scy9jaS9hcmNoaXZlLnB5IC0tdW5pdC10ZXN0LXJlc3VsdHMgLS1jb3ZlcmFn
        ZQpoIGluIGRpciBjOlxiXGlxLWhtaVxkZWJ1Z1xidWlsZCAodGltZW91dCAxMjAwIHNlY3MpCmgg
        d2F0Y2hpbmcgbG9nZmlsZXMge30KaCBhcmd2OiBbYidweXRob24nLCBiJ3Rvb2xzL2NpL2FyY2hp
```

Which, AFAIK isn't allowed. I simply tested by opening an attachment in my client and figured we were good. My bad. I'll give this another crack and see if I can trim off that leading 7bit Content-Transfer-Encoding. I'm assuming this might possibly mess with the tests as well. I'll update this PR and then we'll see how the tests react 😄 

But, to answer your question, the receiving party's client would see the Content-Transfer-Encoding is set to base64 for a particular part of the multi-part and then know that they should base64-decode the attachment. I believe the header (Content-Transfer-Encoding: base64) is added automatically by the encoders.encode_base64 lines that were added in the patch.

Thanks for the reply and the review!

Dunderpate

dunderpate:
Okay, I think we have a winner here @p12tic 😄

This is the output with the updated code:

```
        --===============3169172229356118166==
        Content-Type: text/plain; charset="utf-8"
        MIME-Version: 1.0
        Content-Transfer-Encoding: base64
        Content-Disposition: attachment; filename="unlock.stdio"

        aHB5dGhvbiB0b29scy9jaS91bmxvY2tfd29ya2VyLnB5CmggaW4gZGlyIGM6XGJcaXEtaG1pXGRl
        YnVnXGJ1aWxkICh0aW1lb3V0IDEyMDAgc2VjcykKaCB3YXRjaGluZyBsb2dmaWxlcyB7fQpoIGFy
```

Much more reasonable. I'm not sure if the tests are going to like this, so I'll keep an eye on 'em. If they fail, I'll do what I can to update them to work with this patch. Thanks again for your reply, otherwise I would have totally missed that!

Also, as a reference, this is the approach that is recommended by the Python documentation around removing a header:

* https://docs.python.org/3/library/email.compat32-message.html#email.message.Message.__setitem__

Thanks @p12tic!

Dunderpate

dunderpate:
Cool, looks like the unit tests passed. Ready for another look whenever you have time @p12tic. Thanks again for checking it out.

p12tic:
Thank you!

Issue #5285, opened 2020-05-05
Proposal: Improving flexibility of status reporting


p12tic:
The current design of all BuildBot reporters is centered on reporting the pass/fail status of a build to different data sinks. This unfortunately is not flexible enough if we want to add additional data formats, such as those allowed by the test results PR (#5254).

For example, we may want to support the following use cases:

 - there is a PyLint step. It submits test results to the database after run. A reporter detects when the build has been run as a result of a PR on Github, in which case it would gather the test results on master branch, test results on PR branch and create a comment for each new issue as a comment on the github PR near specific code lines.

 - there is a performance testing step. It submits test results to the database after run. A reporter that detects when the build has been run as a result of a PR on Github, in which case it would gather test results on master branch, test results on PR branch, compare them and create a github comment on the PR listing all significant differences.

 - there is a performance testing step. It submits test results to the database after run. A reporter that detects when the build has been run on a master branch on Github in which case it would gather test results, compare them to last 10 test results on the master branch, create a nice HTML report and send it via email to some developer list.

As can be seen, the data processing requirements for the test results are completely different compared to the current status reporters, where pretty much the only information to be reported is the status of the build and information related to it.

However, it's very likely that it will be possible to send a certain type of report via multiple medium types. Thus it makes sense to separate the concepts of content generation and sending of the report via some medium. The content is not unrelated to the medium as the capabilities of the medium vary wildly. For example, there's a night and day difference between formatting possibilities allowed by a Github comment and what is supported by a HTML page that is uploaded somewhere.

I propose that we extract the content part out of the reporters into a separate concept, called `ReportGenerator` (better suggestions are welcome). This would be responsible for generating a report of a particular format. The report would be communicated to the reporter as a `dict` that can potentially be converted to json and back without information loss. The structure of that dictionary would define what data would be available to the reporters.

The reporters would be responsible for merging the reports and converting the result to a format that is appropriate for presentation.

What report generators are enabled on a reporter would be controlled by a new parameter. The existing way to control a reporter would be deprecated, but preserved for backwards compatibility until BuildBot 3.0.

[...]

PR #5286, opened 2020-05-05
Extract report generation to classes separate to actual reporters

p12tic:
This is a proof of concept implementing #5285 proposal to `NotifierBase` class.

It only implements the data movement part, not the low level parts of the integration of report generators and the reporters.

### DIFF
--- a/master/buildbot/reporters/notifier.py
+++ b/master/buildbot/reporters/notifier.py
@@ -18,14 +18,10 @@ import abc
 from twisted.internet import defer
 
 from buildbot import config
-from buildbot.process.results import CANCELLED
-from buildbot.process.results import EXCEPTION
-from buildbot.process.results import FAILURE
-from buildbot.process.results import SUCCESS
-from buildbot.process.results import WARNINGS
 from buildbot.reporters import utils
-from buildbot.reporters.message import MessageFormatter as DefaultMessageFormatter
-from buildbot.reporters.message import MessageFormatterMissingWorker
+from buildbot.reporters.generators.build import BuildStatusGenerator
+from buildbot.reporters.generators.buildset import BuildSetStatusGenerator
+from buildbot.reporters.generators.worker import WorkerMissingGenerator
 from buildbot.util import service
 
 ENCODING = 'utf-8'
@@ -34,9 +30,6 @@ ENCODING = 'utf-8'
 class NotifierBase(service.BuildbotService):
     name = None
     __meta__ = abc.ABCMeta
-
-    possible_modes = ("change", "failing", "passing", "problem", "warnings",
-                      "exception", "cancelled")
 
     def computeShortcutModes(self, mode):
         if isinstance(mode, str):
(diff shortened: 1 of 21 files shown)


master/buildbot/reporters/notifier.py
```diff
@@ -146,160 +158,37 @@ def stopService(self):
...
-        res = yield utils.getDetailsForBuildset(
-            self.master, bsid,
-            wantProperties=self.messageFormatter.wantProperties,
-            wantSteps=self.messageFormatter.wantSteps,
-            wantPreviousBuild=self.wantPreviousBuild(),
-            wantLogs=self.messageFormatter.wantLogs)
-
-        builds = res['builds']
-        buildset = res['buildset']
+    def _send_reports_for_type(self, type, key, msg):
```
tardyp, line 162: I don't think the switch based on type of generator is optimal API. I see a bit of dependency inversion here.

I think it would be better to let the generator advertise what kind of event it is interested in, and then just have generic reconfigureService which would register to the right events. 

This makes the whole thing a bit more complex and would make you have the generator being child services.
We have to take care of unregistering for events when a generator is not configured anymore (via buildbot reconfig)

master/buildbot/reporters/notifier.py
```diff
@@ -55,85 +48,104 @@ def checkConfig(self, mode=("failing", "passing", "warnings"),
...
+        self._needs_buildsets_complete = \
+            any(isinstance(g, BuildSetStatusGenerator) for g in generators)
+        self._needs_build_finished = \
+            any(isinstance(g, BuildStatusGenerator) for g in generators)
+        self._needs_worker_missing = \
+            any(isinstance(g, WorkerMissingGenerator) for g in generators)
 
     @defer.inlineCallbacks
     def startService(self):
         yield super().startService()
```
tardyp, line 107: this calls reconfigureService() eventually

master/buildbot/reporters/notifier.py
```diff
@@ -55,85 +48,104 @@ def checkConfig(self, mode=("failing", "passing", "warnings"),
...
-        self._workerMissingConsumer = yield startConsuming(
-            self.workerMissing,
-            ('workers', None, 'missing'))
 
-    @defer.inlineCallbacks
+        self._buildsetCompleteConsumer = None
+        self._buildCompleteConsumer = None
+        self._workerMissingConsumer = None
+
+        if self._needs_buildsets_complete:
```
tardyp, line 114: This cannot work in the reconfig use case. The flag would be updated but the consumer will not be stopped or created according to config change.

master/buildbot/reporters/notifier.py
tardyp: making generators child services would be the best in term of separation of concerns, but I admit this is a bit complicated to handle the life cycle of objects.
So for this PR, lets manage the event register in the notifier. 
Note that this effectively makes the notifier another event router.

master/buildbot/reporters/notifier.py
p12tic: Fixed now.

master/buildbot/reporters/notifier.py
p12tic: Fixed now.

master/buildbot/reporters/notifier.py
p12tic: I agree that it would be better to request events in the generators. This PR was just the first bite into the problem, I was focusing into handling of generator output and wanted to do as few other changes as possible. We still ended up with a PR with more than a thousand lines changed.

p12tic:
@tardyp: Is this PR good to go?

tardyp:
looks good. shall we fix all remaining #5312 reports and release 0.8.1 before merging it?

tardyp:
we can also make 0.8.1 as a cherry-pick branch. Lets prioritize this dev which is already hard enough

p12tic:
I think cherry-picking bug fixes is better as all feature work is risky. We probably don't want to introduce other regressions into 2.8.1.

PR #5288, opened 2020-05-06
Handling BitbucketStatusPush response code of 200

dunderpate:
* Before, we were only considering the request failed if it's return
  code wasn't 201. We will now consider the request failed if the return
  code is not 200 and 201.
* This patch resolves an issue where false positive errors were being
  thrown when BitBucket was responding with 200, which the code
  considered an error.

    * I took a look and I don't think it's necessary to describe this change in behavior in the documentation.

### DIFF
--- a/master/buildbot/newsfragments/bitbucketstatuspush-response-code-200.bugfix
+++ b/master/buildbot/newsfragments/bitbucketstatuspush-response-code-200.bugfix
@@ -1,0 +1,1 @@
+Handling the case where the BitbucketStatusPush return code is not 200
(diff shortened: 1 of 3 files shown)


dunderpate:
Hi @p12tic, I think this one's ready for review. The one `bb` failure seems unrelated to this patch. Just wanted to ping you and let you know 👍

p12tic:
@dunderpate Thanks! I reran another failing test yesterday and it's green today, so we're good to go.

dunderpate:
Sweet, thanks a ton @p12tic!

Issue #4918, opened 2019-07-25
DirectoryUpload hangs if directory does not exist.


pepsiman:
I've recently upgraded from buildbot 0.8.8 to buildbot 2.3.1.

In 0.8.8, if a DirectoryUpload step tries to upload a directory that does not exist, the step is marked as exception and the build completes.

In 2.3.1, if a DirectoryUpload step tries to upload a directory that does not exist, the step does not finish and cannot be cancelled.

An exception is logged in the worker's log:
```
2019-07-24 17:09:26+0100 [Broker,client]  startCommand:uploadDirectory [id 5705]
2019-07-24 17:09:27+0100 [Broker,client] WorkerForBuilder.commandFailed <buildbot_worker.commands.transfer.WorkerDirectoryUploadCommand object at 0x7faf129f24a8>
2019-07-24 17:09:27+0100 [Broker,client] Unhandled Error
        Traceback (most recent call last):
          File "lib/python3.4/site-packages/twisted/spread/pb.py", line 1054, in _recvMessage
            netResult = object.remoteMessageReceived(self, message, netArgs, netKw)
          File "lib/python3.4/site-packages/twisted/spread/flavors.py", line 125, in remoteMessageReceived
            state = method(*args, **kw)
          File "lib/python3.4/site-packages/buildbot_worker/base.py", line 153, in remote_startCommand
            d = self.command.doStart()
          File "lib/python3.4/site-packages/buildbot_worker/commands/base.py", line 160, in doStart
            d = defer.maybeDeferred(self.start)
        --- <exception caught here> ---
          File "lib/python3.4/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
            result = f(*args, **kw)
          File "lib/python3.4/site-packages/buildbot_worker/commands/transfer.py", line 227, in start
            archive.add(self.path, '')
          File "lib64/python3.4/tarfile.py", line 1913, in add
            tarinfo = self.gettarinfo(name, arcname)
          File "lib64/python3.4/tarfile.py", line 1785, in gettarinfo
            statres = os.lstat(name)
        builtins.FileNotFoundError: [Errno 2] No such file or directory: '/export/buildbot/worker/foo/results'
```

Nothing is logged in the master's log.

Expected behaviour:
The step is marked as failed.

99rgosse:
I've got the same behavior with RemoteUnpack, when the directory exists and no filename is provided:

[...]

PR #5293, opened 2020-05-08
Try and Except to DirectoryUpload and RemoteTransfer :issue:`#4918`

99rgosse:
## Contributor Checklist:
Issue #4918 

* [x] I have ran the unit tests existing and they passed
* [x] I have read the `README.txt` in that `master/buildbot/newsfragments` directory (but did not fully understood the need)

I have added a test for the transfer.py
I am not able to find where to add another test for remotetransfer.py but with some help I would gladly add one

### DIFF
--- a/master/buildbot/process/remotetransfer.py
+++ b/master/buildbot/process/remotetransfer.py
@@ -128,7 +128,11 @@ class DirectoryWriter(FileWriter):
 
         # Unpack archive and clean up after self
         archive = tarfile.open(name=self.tarname, mode=mode)
-        archive.extractall(path=self.destroot)
+        try:
+            archive.extractall(path=self.destroot)
+        except OSError:
+            print("Error:{}is not a valid directory".format(self.destroot))
+            raise FileNotFoundError
         archive.close()
         os.remove(self.tarname)
 
--- a/master/buildbot/process/remotetransfer.py
+++ b/master/buildbot/process/remotetransfer.py
@@ -130,8 +130,8 @@ class DirectoryWriter(FileWriter):
         archive = tarfile.open(name=self.tarname, mode=mode)
         try:
             archive.extractall(path=self.destroot)
-        except OSError:
-            print("Error:{}is not a valid directory".format(self.destroot))
+        except OSError as e:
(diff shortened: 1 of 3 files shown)


99rgosse:
And I have done a typo when redoing my repository on github... sorry for the inconvenience

master/buildbot/process/remotetransfer.py
```diff
@@ -128,7 +128,11 @@ def remote_unpack(self):
 
         # Unpack archive and clean up after self
         archive = tarfile.open(name=self.tarname, mode=mode)
-        archive.extractall(path=self.destroot)
+        try:
+            archive.extractall(path=self.destroot)
+        except OSError:
+            print("Error:{}is not a valid directory".format(self.destroot))
```
tardyp, line 134: print is going to be burried inside twisted.log.
We use them for debug only, so that we can grep for print, and quicly see leftover debug prints .
In that case, it would make more sense to display the error in a step's logs.
You can use
```suggestion
     self.step.addLogWithException("not a valid directory", e)
```

worker/buildbot_worker/commands/transfer.py
```diff
@@ -224,7 +224,11 @@ def start(self):
         # Not possible with older versions:
         # exceptions.AttributeError: 'TarFile' object has no attribute '__exit__'
         archive = tarfile.open(mode=mode, fileobj=self.fp)
-        archive.add(self.path, '')
+        try:
+            archive.add(self.path, '')
+        except OSError:
+            print("Error:{} is not a valid directory".format(self.path))
```
tardyp, line 230: same, but in that case, you would use

```suggestion
           self.sendStatus({'header': "Error:{} is not a valid directory".format(self.path)})
```

99rgosse:
Thank you very much \o/
I searched something like this addlog and sendstatus but was surely not looking at the correct places !
And by the way thank you for the tips on rewriting history, I hope it is now more acceptable like this :)

tardyp:
The codecov checks are not passed. We need to have our code covered by unit tests so that we can guarantee that this contribution will continue to work in the future.

99rgosse:
I will do that !

PR #5296, opened 2020-05-13
When using Haiku, the "tail" utility is in /bin/tail, not /usr/bin/tail.

rcgordon:
Spending some time to get buildbot-worker up and running on Haiku ( https://haiku-os.org ). The 0.8 series ran fine, once upon a time, so I suspect there are a handful of extremely small issues to clean out, such as this one: Haiku has its "tail" program in a different place than most POSIX systems.

I fixed this on the worker and master, but I'm only trying to get the worker to run on Haiku.

I didn't think this change was worthy of a newsfragment; let me know if I should push one, though!

Thanks,
--ryan.

### DIFF
--- a/master/buildbot/newsfragments/haiku-bin-tail.bugfix
+++ b/master/buildbot/newsfragments/haiku-bin-tail.bugfix
@@ -1,0 +1,1 @@
+Log watcher looks for the "tail" utility in the right location on Haiku OS.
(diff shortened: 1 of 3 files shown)


p12tic:
Looks great. Could you add a newsfragment? It's still a bug fix :-)

rcgordon:
> Looks great. Could you add a newsfragment? It's still a bug fix :-)

Done!  :)

PR #5297, opened 2020-05-13
test_buildbot_net_usage.py: Add "gentoo" to the VERSION_ID check excl…

dol-sen:
…usion

Make the exclusions test into a list (slight speed improvement, easier to extend) and add "gentoo".

Signed-off-by: Brian Dolbec <dolsen@gentoo.org>


tardyp:
thanks!

PR #5298, opened 2020-05-13
Initial pyjade port to pypugjs

dol-sen:
Signed-off-by: Brian Dolbec <dolsen@gentoo.org>

This just swaps out the use of pyjade code for a maintained fork of the original pyjade code.
I have a paired down version of this patch which applies to the 2.7.0 release for our Gentoo package repository.  I have tested buildbot both with the test suite and running the example buildbot "Hello world"  master.cfg

### DIFF
--- a/master/docs/manual/configuration/www.rst
+++ b/master/docs/manual/configuration/www.rst
@@ -108,11 +108,11 @@ This server is configured with the ``www`` configuration key, which specifies a 
     You can use this to slightly customize buildbot look for your project, but to add any logic, you will need to create a full-blown plugin.
     if the directory string is relative, it will be joined to the master's basedir.
     Buildbot uses the jade file format natively (which has been renamed to 'pug' in the nodejs ecosystem), but you can also use html format if you prefer.
-    
+
     Either ``*.jade`` files or ``*.html`` files can be used, and will be used to override templates with the same name in the UI.
     On the regular nodejs UI build system, we use nodejs's pug module to compile jade into html.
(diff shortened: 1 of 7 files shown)


master/docs/manual/configuration/www.rst
```diff
@@ -131,6 +131,8 @@ This server is configured with the ``www`` configuration key, which specifies a
         * quotes in attributes are not quoted. https://github.com/syrusakbary/pyjade/issues/132
```
tardyp, line 131: we should probably make a quick test and see if this is still the case.

tardyp:
Looks good. If we can verify that pyjade/132 is fixed that would actually make a nice improvement. We will need a news fragment as the upgrade will not be seemless as an optional dependency is changing.

master/docs/manual/configuration/www.rst
dol-sen: OK, the new code is identical aside from some minor reformatting. So, no it has not been fixed.

The new code is at lines 145 to 155 of nodes.py

dol-sen:
I've opened this issue with pypugjs:
Double quotes in attribute (carry-over bug from pyjade issue #132) #60
at https://github.com/kakulukia/pypugjs/issues/60

Hopefully it can be fixed quickly.  Have you got an up-to-date link to a file that this fails for?  I can take a look at fixing the code to submit a PR.

p12tic:
@dol-sen Could you add a newsfragment? I would then merge the PR. Thanks a lot!

dol-sen:
Did tardyp get that bug fixed?  We did confirm the bug still exists in pyjade.  He was working on a fix.

PR #5299, opened 2020-05-14
OpenStackLatentWorker: Support passing full auth parameters

cmouse:
Allow passing in any openstack supported keystone options. This is needed to support more than username/password authencation e.g. token based authentication or using application credentials.

## Contributor Checklist:


master/docs/manual/configuration/workers-openstack.rst
```diff
@@ -58,19 +58,12 @@ These are the same details set in either environment variables or passed as opti
     A callable may instead be passed.
     It will be passed the list of available images and must return the image to use.
 
-``os_username``
```
p12tic, line 61: Can we keep these for compatibility? If `os_auth_args` is not supplied, we would contruct it ourselves within `__init__`.

PR #5300, opened 2020-05-15
Implement files for GitHub PRs in github web hook

rvanlaar:
The github webhook now populates Changes with filenames for
PullRequests. This was already implemented for Github Push events.

The PullRequest webhook doesn't have a files changed field.
It's only possible to get the changed files for a pull request by
querying:

/repos/{organisation}/{repo}/pulls/{number}/files

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -251,7 +251,7 @@ class GitHubEventHandler(PullRequestMixin):
         Get Files that belong to the Pull Request
         :param repo: the repo full name, ``{owner}/{project}``.
             e.g. ``buildbot/buildbot``
-        :param numer: the pullrequest number.
+        :param number: the pull request number.
         """
         headers = {"User-Agent": "Buildbot"}
         if self._token:
--- a/master/buildbot/www/hooks/github.py
+++ b/master/buildbot/www/hooks/github.py
@@ -192,6 +192,8 @@ class GitHubEventHandler(PullRequestMixin):
             log.msg("GitHub PR #{} {}, ignoring".format(number, action))
             return (changes, 'git')
 
+        files = yield self._get_pr_files(repo_full_name, number)
+
         properties = self.extractProperties(payload['pull_request'])
         properties.update({'event': event})
         properties.update({'basename': basename})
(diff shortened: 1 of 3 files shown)


master/buildbot/www/hooks/github.py
```diff
@@ -242,6 +245,30 @@ def _get_commit_msg(self, repo, sha):
...
+    @defer.inlineCallbacks
+    def _get_pr_files(self, repo, number):
+        """
+        Get Files that belong to the Pull Request
+        :param repo: the repo full name, ``{owner}/{project}``.
+            e.g. ``buildbot/buildbot``
+        :param number: the pull request number.
+        """
+        headers = {"User-Agent": "Buildbot"}
+        if self._token:
```
tardyp, line 257: I guess if the token is not there, we should return [] directly.

I mean people must provide a token to get pr files.

authentication-less api is heavily rate limited on GH. I don't think it is suitable for CI.

master/buildbot/www/hooks/github.py
```diff
@@ -242,6 +245,30 @@ def _get_commit_msg(self, repo, sha):
...
+
+        url = "/repos/{}/pulls/{}/files".format(repo, number)
+        http = yield httpclientservice.HTTPClientService.getService(
+            self.master,
+            self.github_api_endpoint,
+            headers=headers,
+            debug=self.debug,
+            verify=self.verify,
+        )
+        res = yield http.get(url)
```
tardyp, line 268: what happen in case of error (rate limiting for example)

rvanlaar: The same thing that would happen with the `_get_commit_msg`.

master/buildbot/www/hooks/github.py
rvanlaar: This code directly mirrors the function `_get_commit_msg` https://github.com/buildbot/buildbot/blob/master/master/buildbot/www/hooks/github.py#L225-L243

Then that would be a problem as well.

tardyp: This must have been an overlook before. can we fix the two offending places?

rvanlaar: Github's unauthenticated rate limit is 60/hour. Authenticated it's 5000. [1]
For us at ScummVM, the 60/hour limit is more than enough.

Having the option to not have need a token sped up my development iteration.
This would also need a documentation update, reflection this change [2].

How are we going to resolve this?

[1] https://developer.github.com/v3/#rate-limiting
[2] https://docs.buildbot.net/latest/manual/configuration/wwwhooks.html#github-hook

tardyp: fair enough. We can add a simple warning in the doc.

PR #5308, opened 2020-05-22
DockerLatentWorker: Handle exceptions better when stopping

cmouse:
## Contributor Checklist:

If there is e.g. connectivity problem to the docker host, ensure we handle exceptions when stopping worker, otherwise the substantiation gets stuck on preparing worker.


master/buildbot/worker/docker.py
```diff
@@ -338,12 +340,21 @@ def stop_instance(self, fast=False):
     def _thd_stop_instance(self, instance, fast):
         docker_client = self._getDockerClient()
         log.msg('Stopping container {}...'.format(instance['Id'][:6]))
-        docker_client.stop(instance['Id'])
+        try:
+            docker_client.stop(instance['Id'])
+        except docker.errors.APIError as api_err:
+            log.msg('Cannot stop container {}: {}'.format(instance['Id'][:6], api_err))
+        except RequestException as req_err:
```
tardyp, line 347: I would prefer a little bit more explanation on the different use case you are trying to manage.

If there is connectivity issues to the docker, this means that we will leave running workers on that docker environment. I would prefer we do a exponential backoff restart in that case.
leaving running workers is problematic as it might cost $$ to the buildbot owner.

cmouse: will buildbot automatically retry in case of failure or does this need some special handling?

p12tic: Buildbot does not retry, it requires `stop_instance` to succeed. The retry logic should be placed in the worker implementation because only it knows whether the error is transient or not.

cmouse:
I'll look into this in few weeks.

cmouse:
Updated to catch the exception that's actually causing the problems.

cmouse:
@tardyp  @p12tic something I should do this to this?

p12tic:
@cmouse Sorry for missing this PR. It looks good to me in principle. Could we add tests to cover this new behaviour? There's already a number of tests in `test_docker.py`, it shouldn't be complex to extend that.

mokibit:
Fixed in https://github.com/buildbot/buildbot/pull/8303.

PR #5313, opened 2020-05-28
Update ubuntu worker docker image to 20.04

PhilippSelenium:
#5041  Contributor Checklist:


p12tic:
I wonder if it makes sense to wait until 20.04.1 is out. Ubuntu policy is to wait until the first patch release until they start notify existing installations to upgrade.

PhilippSelenium:
(y)

p12tic:
Hm, I think we can merge this PR after all. Ubuntu 20.04.1 will likely be released in July, so even if we release Buildbot a little earlier, the 20.04 tag will already contain most of the updates.

PR #5316, opened 2020-05-29
GitHub status push

steven-johnson:
If you attempt to use GitHubStatusPush with a builder that specifies multiple codebases, you can currently get runtime failures:
- the existing code assumes that the [0]'th entry of sourcestamps always has valid 'branch' property (it might not, if the codebase is one that is pulled without a branch specification)
- the existing code assumes that *all* entries of sourcestamps have a valid 'revision' property (same reason as above)

This revises the code to be more defensive, and skip the update attempts for sourcestamps that have no revision data (logging the skip in verbose mode).

Note that this is a modestly updated version of https://github.com/buildbot/buildbot/pull/4347, which was offered ~two years ago but eventually rejected; I am resubmitting the PR because this issue seems to have become vastly more problematic for my project after we finally upgraded to more recent versions of buildbot, plus a Py2 -> Py3 conversion.

I have *NOT* updated the unit tests -- it's not at all clear to me what is necessary to adequately provide a unit test for this. If someone can provide concrete guidance I'll attempt to do so.

### DIFF
--- a/master/buildbot/newsfragments/Improve_resiliency_of_GitHubStatusPush.bugfix
+++ b/master/buildbot/newsfragments/Improve_resiliency_of_GitHubStatusPush.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug in GitHubStatusPush that would cause silent failures for builders that specified multiple codebases.
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -129,7 +129,6 @@ class GitHubStatusPush(http.HttpStatusPushBase):
         for sourcestamp in sourcestamps:
             project = sourcestamp['project']
 
-            # If branch is None, so is issue
             issue = None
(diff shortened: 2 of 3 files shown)


p12tic:
Looks like we step on each other toes :) I've updated your old pull request and added tests. I think this pull request can be closed and the old one reopened.

steven-johnson:
> Looks like we step on each other toes :) I've updated your old pull request and added tests. I think this pull request can be closed and the old one reopened.

OMG, thanks so much!

p12tic:
Looks like I can't reopen the old pull request as there's a new one from the same branch.

steven-johnson:
> Looks like I can't reopen the old pull request as there's a new one from the same branch.

Feel free to just clone the changes as you see fit (eg opening another branch or whatnot) -- whatever is easiest to get this issue improved :-)

PR #5322, opened 2020-06-01
Include buildbot.test.fakedb in the source distribution

pampersrocker:
## Contributor Checklist:

 * Not sure if this is possible
 * AFAIK there is none for this

During fixup of my travis CI scripts for my buildbot plugin I've noticed that an import inside buildbot was not working and upon further inspection, that the new fakedb module is missing from source builds of buildbot.

I cannot really test this, as a custom `sdist` does not install any of the `test` modules, so I am not sure if any other module is missing.
 
See the following travis CI log for further info: https://travis-ci.org/github/lab132/buildbot-gitea/builds/693579780

Regards,
Marvin

### DIFF
--- a/master/buildbot/newsfragments/sdist_fakedb.bugfix
+++ b/master/buildbot/newsfragments/sdist_fakedb.bugfix
@@ -1,0 +1,1 @@
+Fixed source distribution missing required buildbot.test.fakedb module for unit tests.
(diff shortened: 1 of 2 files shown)


p12tic:
Thanks!

flokli:
This currently breaks unit tests in the 2.8.0 release. Do you plan releasing a 2.8.1 fixing this?

PR #5323, opened 2020-06-05
Fix plugin import documentation

blag:
This PR updates the plugin import path in the documentation in the plugin import example, and I also changed the `kind` placeholder to `{kind}` to explicitly indicate that it's a placeholder (because actually trying to run it will generate a syntax error). Previously, it was valid Python code that would hit a runtime error when it tried to import from literally `buildbot.kind`.

Also, the documentation refers to `status` plugins, but I think it should actually be referring to `reporters`s plugins, so I fixed that.

Since this is only a documentation fix, I did not add any unit tests, but I did add a newsfragment.
 
## Contributor Checklist:

* [x] I have updated the unit tests - N/A

### DIFF
--- a/master/buildbot/newsfragments/plugin-imports.doc
+++ b/master/buildbot/newsfragments/plugin-imports.doc
@@ -1,0 +1,1 @@
+Fix import path for plugins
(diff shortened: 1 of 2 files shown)


tardyp:
thanks!

PR #5342, opened 2020-06-11
build: Simplify step startup

p12tic:
This PR refactors build startup slightly to make the code flow more apparent by using @defer.inlineCallbacks.

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -555,12 +555,8 @@ class Build(properties.PropertiesMixin):
 
             yield self.master.data.updates.setBuildProperties(self.buildid, self)
 
-            # try to finish the build, but since we've already faced an exception,
-            # this may not work well.
-            try:
-                yield self.buildFinished(["build", "exception"], EXCEPTION)
-            except Exception:
-                log.err(Failure(), 'while finishing a build with an exception')
+            # Note that buildFinished can't throw exception
+            yield self.buildFinished(["build", "exception"], EXCEPTION)
 
     @defer.inlineCallbacks
     def stepDone(self, results, step):
@@ -660,7 +656,9 @@ class Build(properties.PropertiesMixin):
 
         If 'results' is SUCCESS or WARNINGS, we will permit any dependent
         builds to start. If it is 'FAILURE', those builds will be
-        abandoned."""
+        abandoned.
+
+        This method never throws."""
         try:
             self.stopBuildConsumer.stopConsuming()
             self.finished = True
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -335,7 +335,7 @@ class Build(properties.PropertiesMixin):
             self.setupBuild()  # create .steps
         except Exception:
             yield self.buildPreparationFailure(Failure(), "setupBuild")
-            self.buildFinished(['Build.setupBuild', 'failed'], EXCEPTION)
+            yield self.buildFinished(['Build.setupBuild', 'failed'], EXCEPTION)
             return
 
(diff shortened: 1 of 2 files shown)


master/buildbot/process/build.py
```diff
@@ -530,28 +529,38 @@ def startNextStep(self):
...
-        if self.finished:
-            return  # build was interrupted, don't keep building
-        terminate = yield self.stepDone(results, step)  # interpret/merge results
-        if terminate:
-            self.terminate = True
-        yield self.startNextStep()
+            terminate = yield self.stepDone(results, step)  # interpret/merge results
+            if terminate:
+                self.terminate = True
+            yield self.startNextStep()
```
tardyp, line 550: I think this makes the whole thing recursive. For a huge number of step, we will get out of stack space. We should probably break the stack with eventually()

master/buildbot/process/build.py
```diff
@@ -530,28 +529,38 @@ def startNextStep(self):
...
-        d.addBoth(self._flushProperties)
-        d.addCallback(self._stepDone, s)
-        d.addErrback(self.buildException)
-        return None
+
+        # the following function returns a deferred, but we don't wait for it
+        self._start_next_step_impl(s)
+        return defer.succeed(None)
 
     @defer.inlineCallbacks
```
tardyp, line 537: I think we should keep this shortcut (perhaps skip the inlinecallback and results forwarding)

master/buildbot/process/build.py
p12tic: I think it's not recursive, `startNextStep()` will exit without waiting for the step to complete. Effectively the behaviour is the same as before.

master/buildbot/process/build.py
p12tic: Why? I think it makes code harder to read, because it's an indirection to a bit of logic that's almost impossible to get wrong. Also, we don't do this for other `self.master.data.updates` methods, so having a separate method introduces a little bit of confusion too.

Looking into history, previously it wasn't a single-line method, so it made sense to have it this way at the time. I don't think it's worthwhile now.

tardyp: I think the intention of flushProperties is important. This means that at those point, the properties are flushed into the db and available in the UI via events.

We shouldn't call this too often because there is performance impacts.

master/buildbot/process/build.py
tardyp: ah I missed the comment about not waiting for the deferred.

master/buildbot/process/build.py
```diff
@@ -530,28 +529,38 @@ def startNextStep(self):
...
-        if terminate:
-            self.terminate = True
-        yield self.startNextStep()
+            terminate = yield self.stepDone(results, step)  # interpret/merge results
+            if terminate:
+                self.terminate = True
+            yield self.startNextStep()
+
+        except Exception as e:
+            log.msg("{} build got exception when running step {}".format(self, step))
```
tardyp, line 553: the coverage report shows this is not covered. We should probably have something to cover it.

p12tic: Agreed, I'll write a test for that.

master/buildbot/process/build.py
p12tic: I agree, but the same about availability of data to the UI can be said about other `setBuild*` functions. I guess if we call e.g. `setBuildStateString` too often we would see performance impact too :)

master/buildbot/process/build.py
p12tic: Fixed now.

PR #5343, opened 2020-06-11
Make generator support in NotifierBase generic

p12tic:
This PR is a followup of #5286. In that PR a defect was noticed that NotifierBase refers child report generators by type. Additionally, NotifierBase effectively became an event router.

I was able to solve the first problem easily. Now the reporters advertise their wanted event keys that the notifier uses to build a list of events it listens to and when an event comes, the notifier uses its key to retrieve the reports from the generators that were interested into that event.

I've investigated making report generators proper child services of NotifierBase and unfortunately it seems we can't do this without removing functionality. One of the main ideas behind splitting reporters and report generator concepts was that it would be possible to generate multiple types of reports for a single event and then the reporter would take useful information out of all of them and present it as a single report specific to the reporter. If the generators consume the events themselves it becomes almost impossible to merge the reports produced by the same event.

I think that in this case we should consider report generators not as a independent services, but as reusable components of e.g. NotifierBase. It is the reporter which reacts to the event and produces a report to it, generators only define what data will be placed into the report. Thus it should be the reporter which consumes the events, not the generators.

As a bonus, the current implementation is much simplier and easier to understand than the one that is based on independent services. After coding both approaches I think that while upgrading generators to services looks nicer on architectural level, there aren't practical benefits of doing so.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/generators/build.py
+++ b/master/buildbot/reporters/generators/build.py
@@ -22,6 +22,10 @@ from .utils import BuildStatusGeneratorMixin
 
 class BuildStatusGenerator(BuildStatusGeneratorMixin):
 
+    wanted_event_keys = [
+        ('builds', None, 'finished'),
+    ]
+
     def __init__(self, mode=("failing", "passing", "warnings"),
                  tags=None, builders=None, schedulers=None, branches=None,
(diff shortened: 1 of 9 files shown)


tardyp:
nice! 

reasoning looks good, and implementation is straigforward. keep up the good work!

PR #5345, opened 2020-06-11
scripts: Fix crash in windows service startup code

### LINKED ISSUE #5344: Windows service fails after upgrade from 2.7.0 to 2.8.1
Hello,
Windows 7
Buildbot 2.8.1
Twisted 20.3.0
pywin32 227

I upgraded my Buildbot master from version 2.7.0 to version 2.8.1.
Buildbot is configured to run as a Windows service. After the upgrade, the service starts then quickly stops. 
The twistd.log does not get written to.
The Windows Event log(Application) shows:
```
The instance's SvcRun() method failed 
<Error getting traceback - traceback.print_exception() failed 
%2: %3
```
The Windows Event log(System) shows:
```
The BuildBot service failed to start due to the following error: 
The service did not respond to the start or control request in a timely fashion.
```

buildbot/master/buildbot/scripts/windows_service.py (line 233)
`cmd = '{} --spawn {} start --nodaemon {}'.format(self.runner_prefix, hstop, bbdir)`
needs to be changed to:
`cmd = '{} --spawn {} start --nodaemon {}'.format(self.runner_prefix, int(hstop), bbdir)`

Notice the call to int(hstop)
The first version produces something like:
`python.exe --spawn <PyHANDLE:456> start --nodaemon c:/master`
instead of:
`python.exe" --spawn 456 start --nodaemon c:/master`
Thanks,
Joe

p12tic:
Fixes #5344. Thanks @BeagleJoe for the proposed patch in the GitHub issue.

Issue has been caused by 6bafe09d93fcb40389ff4a1d736347032ad98d96.

This is serious enough issue that it warrants Buildbot 2.8.2, but in order to release we need testing by at least 2 people.

## Contributor Checklist:

* [not available] I have updated the unit tests

### DIFF
--- a/master/buildbot/scripts/windows_service.py
+++ b/master/buildbot/scripts/windows_service.py
@@ -230,7 +230,7 @@ class BBService(win32serviceutil.ServiceFramework):
             self.info("Starting BuildBot in directory '{}'".format(bbdir))
             hstop = self.hWaitStop
 
-            cmd = '{} --spawn {} start --nodaemon {}'.format(self.runner_prefix, hstop, bbdir)
+            cmd = '{} --spawn {} start --nodaemon {}'.format(self.runner_prefix, int(hstop), bbdir)
             # print "cmd is", cmd
             h, t, output = self.createProcess(cmd)
             child_infos.append((bbdir, h, t, output))
--- a/master/buildbot/newsfragments/windows-service-crash.bugfix
+++ b/master/buildbot/newsfragments/windows-service-crash.bugfix
@@ -1,0 +1,1 @@
+Fix crash in Buildbot Windows service startup code (:issue:`5344`)


BeagleJoe:
@p12tic 
The reason that 2.7.0 worked is that it used older (Python 2 style formatting). I tried using {:d} style, but got an error unknown format specifier passed to PyHANDLE. I think that the Pywin32 code is older. The object does have __int__() function though.

p12tic:
Yes, I got initially confused as I thought the file has not had any changes during 2.7.x timeframe and thought the problem must be in pywin32. My assumption was incorrect :)

master/buildbot/scripts/windows_service.py
```diff
@@ -230,7 +230,7 @@ def SvcDoRun(self):
             self.info("Starting BuildBot in directory '{}'".format(bbdir))
             hstop = self.hWaitStop
 
-            cmd = '{} --spawn {} start --nodaemon {}'.format(self.runner_prefix, hstop, bbdir)
+            cmd = '{} --spawn {} start --nodaemon {}'.format(self.runner_prefix, int(hstop), bbdir)
```
tardyp, line 233: I think this int() deserves a comment as it is really not obvious, and people will clean that up with a {:d}. The bug explain why, but it is too many indirection.

PR #5355, opened 2020-06-14
Add more deprecation warnings for buildbot.status removal

p12tic:
`buildbot.status` has been deprecated since 0.9.0. It makes sense to remove it as that code is not covered by tests well and is probably already broken. Before that, we must ensure that warnings are shown in the logs to all users for at least one release.

This PR depends on #5353, so it will be hard to review until that PR is merged.

### DIFF
--- a/master/buildbot/test/__init__.py
+++ b/master/buildbot/test/__init__.py
@@ -19,6 +19,9 @@ import warnings
 from distutils.version import LooseVersion
 
 from buildbot import monkeypatches
+from buildbot.test.util.warnings import assertProducesWarning  # noqa pylint: disable=wrong-import-position
+from buildbot.test.util.warnings import assertProducesWarnings  # noqa pylint: disable=wrong-import-position
+from buildbot.warnings import DeprecatedApiWarning  # noqa pylint: disable=wrong-import-position
 
 # import mock so we bail out early if it's not installed
 try:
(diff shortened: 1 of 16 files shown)


master/buildbot/config.py
```diff
@@ -389,7 +378,7 @@ def copy_str_param(name, alt_key=None):
             # would hide the title.
             warnings.warn('WARNING: Title is too long to be displayed. ' +
                           '"Buildbot" will be used instead.',
-                          category=DeprecatedConfigWarning)
+                          category=ConfigWarning)
```
tardyp, line 381: I agree we should have a ConfigWarning class that for warnings that are not deprecations. but it seems you omitted to add it.

master/buildbot/test/__init__.py
```diff
@@ -38,6 +41,28 @@
     raise ImportError("\nBuildbot tests require mock version 0.8.0 or "
                       "higher; try 'pip install -U mock'")
 
+
+# This is where we load deprecated module-level APIs to ignore warning produced by importing them.
+# After the deprecated API has been removed, leave at least one instance of the import in a
+# commented state as reference.
+
+
+with assertProducesWarnings(DeprecatedApiWarning,
```
tardyp, line 50: I would have put that in a separate test suite, to make sure that all other parts of buildbot are not importing status.
Here we will also be ignoring the import of status in other modules

master/docs/developer/master-overview.rst
```diff
@@ -71,7 +71,7 @@ The master has a number of useful attributes:
     A :py:class:`buildbot.process.debug.DebugServices` instance that manages
     debugging-related access -- the manhole, in particular.
 
-``master.status``
+``master.status`` (deprecated)
```
tardyp, line 74: its not only deprecated, it is not working anymore since 0.9.

p12tic:
@tardyp: Are you sure all of these status APIs aren't working anymore since 0.9.x? Looking into the code it seemed that they implement many parts on top of the data API and there are still hooks into the rest of the buildmaster code, so at least parts of the status API should work.

Also, there are e.g. buildbot.status.builder module that imports a bunch of stuff which the users may still be using. I think we can't just remove it without putting a warning for at least a single release.

tardyp:
indeed, I realize there are still a bunch of thing. I don't think the status is implementing things on top of data API. There are just there a bunch of helpers that could indeed work more or less.

So I agree with you on 3.0

p12tic:
> I don't think the status is implementing things on top of data API. 
There were some uses, but yes, most of stuff comes directly from internals of the buildmaster.

master/buildbot/test/__init__.py
p12tic: In this case it's impossible, because many parts of the master import status everywhere. This can be solved only by removing it :)

master/buildbot/config.py
p12tic: I added it back in #5353.

PR #5357, opened 2020-06-14
remove uses of status

tardyp:
So I have spent a couple hour removing the status API, and fix the test.

eventually, the tests are all passing, when additional status code removal commit is applied.
This one PR does everything but delete the status API, so it can be safely merged.

I have split into several reviewable commits, as there are several non trivial changes.

### DIFF
--- a/master/buildbot/master.py
+++ b/master/buildbot/master.py
@@ -43,7 +43,6 @@ from buildbot.process.botmaster import BotMaster
 from buildbot.process.users.manager import UserManagerManager
 from buildbot.schedulers.manager import SchedulerManager
 from buildbot.secrets.manager import SecretManager
-from buildbot.status.master import Status
 from buildbot.util import check_functional_environment
 from buildbot.util import service
 from buildbot.util.eventual import eventually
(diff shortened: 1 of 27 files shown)


p12tic:
Looks great. I guess it makes sense to merge the part of the PR that removes unnecessary uses of `buildbot.status` within the rest of the Buildbot codebase. The actual removal then can wait until v3.0.

tardyp:
@p12tic I have removed the status API removal part and submitted another PR which will wait.

The biggest change is the move of worker info from status API to worker.

The rest is only tests fixes, and core code not calling status updates api.

another non-trivial issue is the test_latent code, which used synchronous build stop API.
Using the data api makes it asynchronous via the event api, so there is a need for a bit more complex synchronization code. (wait that the build is visible in the db before stopping it)

master/buildbot/master.py
```diff
@@ -440,12 +436,6 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
     def allSchedulers(self):
         return list(self.scheduler_manager)
 
-    def getStatus(self):
```
p12tic, line 443: I think we're removing public API here, no?

PR #5371, opened 2020-06-17
Add a way for builds to store transient data

p12tic:
This PR adds a new database table and related APIs for steps to store master-side transient data that should be fetched by subsequent steps or reporters. It is effectively a per-build key-value store. We expect to store json data there. A separate table is used because the size of the data may be large and it is no longer needed after the build.

Currently there are no higher level APIs. 

There's one defect: currently steps are expected to serialize and deserialize the data themselves. We should do serialization in the DB layer. There's no data API, so there's no concern of repeatedly serializing/deserializing json, which was the primary reason why this design decision was done initially.

Also I don't think `build_data` is the best possible name for the APIs, but I haven't been able to think of anything better. Suggestions are welcome :-)

### DIFF
--- a/master/buildbot/db/build_data.py
+++ b/master/buildbot/db/build_data.py
@@ -1,0 +1,130 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+import sqlalchemy as sa
+
+from twisted.internet import defer
+
+from buildbot.db import base
+
+
+class BuildDataDict(dict):
+    pass
+
+
+class BuildDataConnectorComponent(base.DBConnectorComponent):
+
+    def _insert_race_hook(self, conn):
+        # called so tests can simulate a race condition during insertion
+        pass
+
+    @defer.inlineCallbacks
+    def setBuildData(self, buildid, name, value, source):
+        def thd(conn):
+            build_data_table = self.db.model.build_data
+
+            update_values = {
+                'value': value,
+                'source': source
+            }
+
+            insert_values = {
+                'buildid': buildid,
+                'name': name,
+                'value': value,
+                'source': source,
+            }
+
+            while True:
+                q = build_data_table.update()
+                q = q.where((build_data_table.c.buildid == buildid) &
+                            (build_data_table.c.name == name))
+                q = q.values(update_values)
+                r = conn.execute(q)
+                if r.rowcount > 0:
+                    return
+                r.close()
+
+                self._insert_race_hook(conn)
+
+                try:
+                    q = build_data_table.insert().values(insert_values)
+                    r = conn.execute(q)
+                    return
+                except (sa.exc.IntegrityError, sa.exc.ProgrammingError):
+                    # there's been a competing insert, retry
+                    pass
+
+        yield self.db.pool.do(thd)
+
+    @defer.inlineCallbacks
+    def getBuildData(self, buildid, name):
+        def thd(conn):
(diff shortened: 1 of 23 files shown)


master/buildbot/db/migrate/versions/057_add_build_data_tables.py
```diff
@@ -0,0 +1,49 @@
...
+
+    build_data = sautils.Table(
+        'build_data', metadata,
+        sa.Column('id', sa.Integer, primary_key=True),
+        sa.Column('buildid', sa.Integer,
+                  sa.ForeignKey('builds.id', ondelete='CASCADE'),
+                  nullable=False),
+        sa.Column('name', sa.String(256), nullable=False),
+        # JSON encoded value
+        sa.Column('value_json', sa.Text(1000000), nullable=False),
```
tardyp, line 40: I would use Blob for storage as it is supposed to be stored outside of the table.

master/buildbot/db/build_data.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class BuildDataConnectorComponent(base.DBConnectorComponent):
+
+    def _insert_race_hook(self, conn):
+        # called so tests can simulate a race condition during insertion
+        pass
+
+    @defer.inlineCallbacks
+    def setBuildData(self, buildid, name, value_json, source):
```
tardyp, line 35: I think buildData should be binary blob, and not only json. Encoding/decoding can be wherever is convenient. If this is json, responsible of the encode and decode could be the worker. transfer text between work may mean some encode decode in between, which can be avoided if we just use bytes everywhere until use.

master/buildbot/db/migrate/versions/057_add_build_data_tables.py
tardyp: also text is often stored in utf16 so it is double the needed size.

p12tic: Changed to Blob.

master/buildbot/db/build_data.py
p12tic: Agreed, changed to Blob/bytes

p12tic:
@tardyp Data API has been added now.

master/buildbot/spec/types/build_data.raml
```diff
@@ -0,0 +1,47 @@
...
+
+    All update methods are available as attributes of ``master.data.updates``.
+
+    .. py:class:: buildbot.data.build_data.BuildData
+
+        .. py:method:: setBuildData(buildid, name, value, source)
+
+            :param integer buildid: build id to attach data to
+            :param unicode name: the name of the data
+            :param bytestr value: the value of the data as ``bytes``.
```
tardyp, line 22: how does that transform in json? I think json strings are utf-8, so this will clash if we encode binary data. I see that we also have ss patches that are binary, but I am not sure we have thorough test coverage on this.

p12tic: Why should they clash? If we store binary then we assume that the reader/writer knows the encoding and uses the same each time.

p12tic: Were you referring to use BLOB in the DB just as a workaround for some users using databases with non-UTF-8 default encoding for Text column?

tardyp: I was rather thinking of the case where a build_data generator would generator other data format than json (e.g zip containing several files). In this case, the byte string will contain arbitrary bytes.

I realize I wasn't clear in my first comment. I was there thinking of the REST api, which will convert data api's dicts into json. There is a toJson helper in buildbot.util, which converts datatimes into timestamps, but nothing to convert bytes. It will be needed.

```
json.dumps({'a': b'a'})
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```


https://stackoverflow.com/questions/1443158/binary-data-in-json-string-something-better-than-base64

PR #5372, opened 2020-06-17
Build documentation in parallel 

p12tic:
After sphinx-rtd-theme 5.0 has been released, we can now build documentation in parallel mode.

### DIFF
--- a/master/docs/Makefile
+++ b/master/docs/Makefile
@@ -6,6 +6,9 @@ dev: clean html
 
 VERSION := $(shell if [ -n "$$VERSION" ]; then echo $$VERSION; else PYTHONPATH=..:$${PYTHONPATH} python -c 'from buildbot import version; print(version)'; fi)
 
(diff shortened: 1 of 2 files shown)


master/docs/Makefile
```diff
@@ -6,6 +6,9 @@ dev: clean html
 
 VERSION := $(shell if [ -n "$$VERSION" ]; then echo $$VERSION; else PYTHONPATH=..:$${PYTHONPATH} python -c 'from buildbot import version; print(version)'; fi)
 
+# sphinx is fastest on Buildbot docs when using around 8 cores. Anything more does not improve the build speed
```
tardyp, line 9: then we want to use max, no?

p12tic: It is exactly this, min/max are not intuitive in this context, function like clamp is better.

Say, someone has 12 cores, min(8, 12) == 8. If someone has 1 core, min(8, 1) == 1.

tardyp: obviously :}

PR #5385, opened 2020-06-24
reporters/gerrit: Allow using HTTP auth

flokli:
This adds support to report back gerrit status via HTTP, if a URL is
specified, and credentials are passed via auth=.

And especially:

## Contributor Checklist:


master/buildbot/reporters/gerrit.py
```diff
@@ -401,42 +407,68 @@ def getProperty(build, name):
...
-        if gerrit_version is None:
-            self.callWithVersion(
-                lambda: self.sendCodeReview(project, revision, result))
-            return
+    @defer.inlineCallbacks
+    def sendCodeReview(self, project, revision, result, change_id=None):
+        if self._http == None:
+            gerrit_version = self.getCachedVersion()
+            if gerrit_version is None:
+                self.callWithVersion(
```
tardyp, line 418: callWithVersion do not make much sense now that sendCodeReview is inlineCallback. Please remove the function and call the getVersion directly from here.

master/buildbot/reporters/gerrit.py
```diff
@@ -401,42 +407,68 @@ def getProperty(build, name):
...
                 return
 
-    def sendCodeReview(self, project, revision, result):
-        gerrit_version = self.getCachedVersion()
-        if gerrit_version is None:
-            self.callWithVersion(
-                lambda: self.sendCodeReview(project, revision, result))
-            return
+    @defer.inlineCallbacks
+    def sendCodeReview(self, project, revision, result, change_id=None):
```
tardyp, line 414: I think it would just be simpler to create a subclass, and override the relevent method. In anycase, we need two method here rather than a big one splitted with a large if else block

flokli: What should be subclasses? The idea was that `GerritStatusPush` now supports both http and ssh API.

tardyp: It would be more explicit to have a (unchanged) GerritStatusPush and a new GerritHttpStatusPush

- API wise, the only difference is whether there is an auth or not. It does not seem very intuitive.
- coding wise, the main method is a big if else. This would be much better in a overridden method.

flokli:
I still intend to address this, just haven't gotten the time recently. Will try to get to this at the end of the week!

flokli:
Thanks @tardyp for your feedback, but I'm sorry, but I can't pursue this any further. I'll leave the branch around if someone wants to pick this up.

Issue #5416, opened 2020-07-22
Show GitHub Avatar in Buildbot


ryandesign:
When I log into my Buildbot with GitHub, I see an avatar in the upper right of the Buildbot screen and when I click it I see my name and email address from my GitHub account. But the avatar image is unfamiliar to me. Presumably it is from Gravatar, which I understand is your default avatar provider. How do I change this to show my GitHub user icon?

I found [documentation](https://docs.buildbot.net/latest/manual/configuration/www.html?highlight=avatar_methods#web-server) which says

> `avatar_methods`
>
> List of methods that can be used to get avatar pictures to use for the web server. By default, buildbot uses Gravatar to get images associated with each users

but it does not say what other values of that option are possible.

I see that something called AvatarGitHub was submitted over 4 years ago in #1894, and @tardyp provided an improved implementation of it in a code review, but the PR was closed without being merged. I don't find any occurrences of "AvatarGitHub" in the Buildbot 2.8.2 source code. Did this feature simply never get included? Or did it get included a different way?

tardyp:
hi #1894 was awesome, but unfortunately nobody stepped up to finish it.
Anybody interrested can pick up stalled request and rebase them to continue work, no problem..

ryandesign:
For anyone else wanting to look into this, I updated your code as follows. (This diff is against buildbot-2.8.2.tar.gz from pypi.)
```
--- buildbot/www/avatar.py.orig	2020-06-14 09:25:22.000000000 -0500
+++ buildbot/www/avatar.py	2020-07-23 05:46:53.000000000 -0500
@@ -14,6 +14,7 @@
 # Copyright Buildbot Team Members
 
 import hashlib
+import txrequests
 from urllib.parse import urlencode
 from urllib.parse import urljoin
 
@@ -31,6 +32,37 @@
         raise NotImplementedError()
 
 
+class AvatarGitHub(AvatarBase):
+    name = "github"
+    cache = {}
+    BASE_URL = 'https://api.github.com'  # can be overridden for GHE
+    SEARCH_PATH = '/search/users'
+
+    def getSession(self):
+        return txrequests.Session()
+
+    @defer.inlineCallbacks
+    def getUserAvatar(self, email, size, defaultAvatarUrl):
+        if not email in self.cache:
+            return None
+        url = self.cache.get(email)
+        if not url:
+            search_url = self.BASE_URL + self.SEARCH_PATH
+            with self.getSession() as session:
[...]

PR #5420, opened 2020-07-26
Show GitHub user avatar

rajdeepbharati:
Closes https://github.com/buildbot/buildbot/issues/5416

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/avatar.py
+++ b/master/buildbot/www/avatar.py
@@ -27,8 +27,18 @@ from buildbot.www import resource
 class AvatarBase(config.ConfiguredMixin):
     name = "noavatar"
 
-    def getUserAvatar(self, email, size, defaultAvatarUrl):
+    def getUserAvatar(self, email, username, size, defaultAvatarUrl):
         raise NotImplementedError()
+
+
+class AvatarGitHub(AvatarBase):
+    name = "github"
+
+    AVATAR_URL = 'https://avatars.githubusercontent.com/'
+
+    def getUserAvatar(self, email, username, size, defaultAvatarUrl):
+        github_url = self.AVATAR_URL + username.decode("utf-8")
+        raise resource.Redirect(github_url)
 
 
 class AvatarGravatar(AvatarBase):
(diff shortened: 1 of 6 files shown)


master/buildbot/www/avatar.py
```diff
@@ -72,12 +82,15 @@ def render_GET(self, request):
     def renderAvatar(self, request):
         email = request.args.get(b"email", [b""])[0]
         size = request.args.get(b"size", 32)
+        username = request.args.get(b"username")
```
tardyp, line 85: maybe we should use the same as line 83...

```suggestion
        username = request.args.get(b"username", [None])[0]
```

www/base/src/app/common/directives/loginbar/loginbar.tpl.jade
```diff
@@ -15,7 +15,9 @@ ul.nav.navbar-nav.navbar-right(ng-show="config.auth.name != 'NoAuth'")
 
     li.dropdown(uib-dropdown, ng-hide="anonymous")
         a.dropdown-toggle(uib-dropdown-toggle)
-            .avatar(ng-if="email")
+            .avatar(ng-if="username")
```
tardyp, line 18: I would pass the two arguments and let the avatar plugin choose what is suitable.

In this case, your implementation would break the GravatarAvatar that buildbot is using.

tardyp:
@rajdeepbharati ping :)

tardyp:
cool. please add the newsfragment, and documentation

rajdeepbharati:
@tardyp added

mojca:
Thanks a lot!

master/buildbot/www/avatar.py
```diff
@@ -27,17 +27,27 @@
...
 
 
+class AvatarGitHub(AvatarBase):
+    name = "github"
+
+    AVATAR_URL = 'https://avatars.githubusercontent.com/'
+
+    def getUserAvatar(self, email, username, size, defaultAvatarUrl):
+        github_url = self.AVATAR_URL + username.decode("utf-8")
+        raise resource.Redirect(github_url)
```
ryandesign, line 41: Thanks. Looks like this returns the full-size image. Is there a way to use the size parameter / should we be doing so?

PR #5441, opened 2020-08-03
Fix ci

tardyp:
Here are some steps I did to upgrade our CI environment.

- rebuilt https://hub.docker.com/repository/docker/buildbot/metabbotcfg/builds
The build was broken for a while since the base image moved to ubuntu 18 (python-software-properties is not needed anymore)
I encountered various build issues, which made me upgrade a few other thing.
- updated node environment to node LTS (12 instead of 8), because chromedriver didn't want to install with node8.
- remove python 2.6, because we don't use it anymore

The next day, obviously all CI broke :(

- The new virtualenv do not want 'no-site-package- anymore. They way it was the default for years, this is why I removed it.
- circle-ci failed with a weird error. This took me a while to figure out, but this was due to cached .venv, which was for python 3.5 and now we use 3.7. So I did invalidate the cache.
- Then moving to 3.7 revealed a few issues in our requirement.txt (enum34 and typings shall not be installed there)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ docker-buildbot-master:
 	$(DOCKERBUILD) -t buildbot/buildbot-master:master master
 
 $(VENV_NAME):
-	virtualenv -p $(VENV_PY_VERSION) --no-site-packages $(VENV_NAME)
+	virtualenv -p $(VENV_PY_VERSION) $(VENV_NAME)
 	$(PIP) install -U pip setuptools
 
 # helper for virtualenv creation
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,15 +16,15 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-            - 3-dependencies-{{ checksum "requirements-ci.txt" }}-{{ checksum "requirements-cidocs.txt" }}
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -107,7 +107,7 @@ docker-buildbot-master:
 	$(DOCKERBUILD) -t buildbot/buildbot-master:master master
 
 $(VENV_NAME):
-	virtualenv -p $(VENV_PY_VERSION) --no-site-packages $(VENV_NAME)
```
p12tic, line 110: Do you know the reason why this does not work by chance? In my experience this leads to various issues depending on what packages and what versions of them are installed on the system.

tardyp: I documented what I did in the PR description.

for your specific question, according to:

https://stackoverflow.com/questions/60783788/virtualenv-virtualenvwrapper-virtualenv-error-unrecognized-arguments-no-sit

> --no-site-packages is the default for virtualenv (and has been for like 5 years?) you can remove export VIRTUALENVWRAPPER_VIRTUALENV_ARGS='--no-site-packages' from your .bashrc
> it appears in virtualenv>=20 that this option was removed

was news to me as well :-/

p12tic: Thanks, it all makes sense now.

PR #5447, opened 2020-08-07
test: Specify AWS_DEFAULT_REGION in ec2 tests

### LINKED ISSUE #5370: EC2 latent worker tests try to connect to AWS if default credentials are setup
The common pattern for tests that need to connect to third-party services is to enable the tests via an environment variable. However, in the EC2 latent worker tests, they seem to try to connect to real AWS if AWS credentials are setup.

p12tic:
boto3 will read ~/.aws/config to get the default region. If this is different than what is expected in tests, then the tests would be broken.

Fixes #5370.


tardyp:
❤️

PR #5449, opened 2020-08-10
Fix transient data api REST endpoint

p12tic:
There was an unresolved issue in #5371 of what we do with actually binary data. It looks like that the most sensible solution is to only offer a raw endpoint to actually download it. This way both the REST endpoint and internal data APIs can use the same representation of the data.

To make the other APIs that don't return the values more useful, a length field has been added. This way there's at least some information about what's in the data. The migration script is incomplete and just sets length values of all existing rows to zero.

## Contributor Checklist:

* [n/a] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)

### DIFF
--- a/master/buildbot/db/migrate/versions/058_add_build_data_length.py
+++ b/master/buildbot/db/migrate/versions/058_add_build_data_length.py
@@ -1,0 +1,29 @@
+# This file is part of Buildbot. Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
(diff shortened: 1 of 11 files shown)


master/buildbot/db/migrate/versions/058_add_build_data_length.py
```diff
@@ -0,0 +1,29 @@
...
+# Copyright Buildbot Team Members
+
+
+import sqlalchemy as sa
+
+from buildbot.util import sautils
+
+
+def upgrade(migrate_engine):
+    # build_data without the 'length' column has never been released, so we don't care about
```
tardyp, line 23: 👍

PR #5454, opened 2020-08-16
log: Decode bytes before line processing, not after

p12tic:
Turns out that we don't really support bytes in new style steps log processing. This PR fixes that. Nothing seems to cover this use case right now, but I did some old-style -> new-style step conversion work to understand how logs are handled and that code will cover it.

## Contributor Checklist:

* [not needed] I have updated the unit tests

### DIFF
--- a/master/buildbot/process/log.py
+++ b/master/buildbot/process/log.py
@@ -119,13 +119,13 @@ class PlainLog(Log):
         super(PlainLog, self).__init__(master, name, type, logid, decoder)
 
         def wholeLines(lines):
-            if not isinstance(lines, str):
-                lines = self.decoder(lines)
             self.subPoint.deliver(None, lines)
             return self.addRawLines(lines)
         self.lbf = lineboundaries.LineBoundaryFinder(wholeLines)
 
     def addContent(self, text):
+        if not isinstance(text, str):
+            text = self.decoder(text)
         # add some text in the log's default stream
         return self.lbf.append(text)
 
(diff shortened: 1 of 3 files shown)


tardyp:
I guess we need some unit tests to check this feature, no?

p12tic:
@tardyp  I've converted some master old style steps to new style which covers this new code. This is how I found the issue. I'll submit the PR after this PR lands as the next PR is dependent on this one.

master/buildbot/process/log.py
```diff
@@ -176,12 +174,18 @@ def wholeLines(lines):
             return lbf
 
     def addStdout(self, text):
+        if not isinstance(text, str):
```
tardyp, line 177: at some point, I had in mind to do opportunistic decoding, so that we can have a fast path which send log data directly from worker to db when no logobserver is setup. I think that would help greatly in performance.

I think it is easier to have the decoding asap, though

tardyp (approved):

master/buildbot/process/log.py
```diff
@@ -176,12 +174,18 @@ def wholeLines(lines):
...
+            text = self.decoder(text)
         return self._getLbf('o').append(text)
 
     def addStderr(self, text):
+        if not isinstance(text, str):
+            text = self.decoder(text)
```
tardyp, line 183: the coverage says this is never executed. Do we really need it? don't we have some autodecode before?

I am pretty sure bytes are used on the network link, so maybe it is already converted before.

master/buildbot/process/log.py
p12tic: I haven't considered this, thanks for explaining.

master/buildbot/process/log.py
p12tic: This will be covered by a subsequent PR where some master steps were changed to new style. This is how I noticed the problem at first. I just thought it makes sense to have a separate PR to not mix everything.

PR #5466, opened 2020-08-17
Convert transfer steps to new style

p12tic:
This PR depends on #5458 for the shared newsfragment for the new style conversion changes.

## Contributor Checklist:


tardyp (approved):

tardyp:
There is a bit of missing coverage. I would suggest to add some tests for the error cases.

interrupt case is probably a bit too hard to test.
as for the asyncSleep, I would put that as an coverage exception, as I think this is what keep changing the coverage

p12tic:
Looks like adding tests needs some enhancements in the test infrastructure, so I'll submit them in a next PR

PR #5467, opened 2020-08-17
Convert source steps to new style

p12tic:
This PR depends on #5458 for the shared newsfragment for the new style conversion changes.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/source/base.py
+++ b/master/buildbot/steps/source/base.py
@@ -182,13 +182,6 @@ class Source(LoggingBuildStep, CompositeStepMixin):
                 "Sourcestep {} does not have a codebase, other sourcesteps do".format(self.name)
             super().setProperty(name, value, source)
 
-    def describe(self, done=False):
-        desc = self.descriptionDone if done else self.description
-        if self.descriptionSuffix:
-            desc = desc[:]
-            desc.extend(self.descriptionSuffix)
-        return desc
-
     def computeSourceRevision(self, changes):
         """Each subclass must implement this method to do something more
         precise than -rHEAD every time. For version control systems that use
--- a/master/buildbot/steps/source/base.py
+++ b/master/buildbot/steps/source/base.py
@@ -257,6 +257,13 @@ class Source(LoggingBuildStep, CompositeStepMixin):
                                       "Faked {} checkout/update 'successful'\n".format(self.name))
             return SKIPPED
 
+        if getattr(self, 'startVC', None) is not None:
+            msg = 'Old-style source steps are no longer supported. Please convert your custom ' \
+                  'source step to new style (replace startVC with run_vc and convert all used ' \
+                  'old style APIs to new style). Please consider contributing the source step to ' \
+                  'upstream BuildBot so that such migrations can be avoided in the future.'
+            raise NotImplementedError(msg)
+
         if not self.alwaysUseLatest:
             # what source stamp would this step like to use?
             s = self.build.getSourceStamp(self.codebase)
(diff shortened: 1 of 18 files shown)


master/buildbot/steps/source/base.py
```diff
@@ -279,22 +280,20 @@ def start(self):
...
-                    self.addCompleteLog("patch", bytes2unicode(patch[1]))
+                    yield self.addCompleteLog("patch", bytes2unicode(patch[1]))
             else:
                 log.msg("No sourcestamp found in build for codebase '{}'".format(self.codebase))
-                self.step_status.setText(
-                    ["Codebase", '{}'.format(self.codebase), "not", "in", "build"])
-                self.addCompleteLog("log",
-                                    "No sourcestamp found in build for codebase '{}'".format(
-                                            self.codebase))
+                self.descriptionDone = ["Codebase", '{}'.format(self.codebase),
```
tardyp, line 286: the '{}'.format looks suspicious. probably str(self.codebase) is better..

master/buildbot/steps/source/base.py
```diff
@@ -279,22 +280,20 @@ def start(self):
...
         else:
             revision = None
             branch = self.branch
             patch = None
 
-        d = self.startVC(branch, revision, patch)
-        d.addCallback(self.finished)
-        d.addErrback(self.failed)
-        return None
+        res = yield self.run_vc(branch, revision, patch)
```
tardyp, line 298: this means that any custom VC will stop working. is this what we want?

master/buildbot/steps/worker.py
```diff
@@ -232,12 +234,13 @@ class CompositeStepMixin():
     def workerPathToMasterPath(self, path):
         return os.path.join(*self.worker.path_module.split(path))
 
+    @defer.inlineCallbacks
     def addLogForRemoteCommands(self, logname):
         """This method must be called by user classes
         composite steps could create several logs, this mixin functions will write
         to the last one.
         """
-        self.rc_log = self.addLog(logname)
```
tardyp, line 240: ouch.. I guess this could never work. thus it is not used by anybody..

tardyp (changes requested):
I think we should call startVC if it exists, and send a deprecation warning..

master/buildbot/steps/source/base.py
p12tic: Hmm. So the problem is that the base class is being changed to new style step and any sub-classes must be changed in lock-step. Just calling startVC would just hide bugs.

The same problem is present in other steps. The source steps are probably unique in that the base class packs a lot of functionality.

I wonder if we could work around this by implementing a base run_vc that would check for existence of startVC and if such exists, would setup the step for the old style much like BuildStep.run does. This temporary solution would be removed on v3.0

tardyp: I would rather avoid another pile of hacks like I did in the base step. As the migration is pretty straighforward (just changing method name and yielding addLogForRemoteCommands, maybe we should just warn and create a small migration doc.
I would not expect so many complicated source steps anyway.

p12tic: We also need to return the result as a return value from `startVC` instead of calling `finished` or `failed`. This makes it much more difficult to support the old source steps.

I think the best would be to check whether `startVC` exists and throw an exception without even trying to run the step.

We can recommend to include the source step into the buildbot proper in the exception message :-)

p12tic: @tardyp: Fixed in 7570d9c.

master/buildbot/steps/source/base.py
p12tic: Fixed in e4d76550e1.

tardyp (approved):

PR #5474, opened 2020-08-20
Transfer step improvements

p12tic:
This PR adds several improvements which were discovered while enhancing transfer steps:

 - Transfer step result was wrong in the case of step interruption.

 - urlText argument was not universally supported.

Also, the test framework enhanced to support step interruption and a number of new unit tests were written.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/shell.py
+++ b/master/buildbot/steps/shell.py
@@ -13,7 +13,6 @@
 #
 # Copyright Buildbot Team Members
 
-import inspect
 import re
 
 from twisted.python import failure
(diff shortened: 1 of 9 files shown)


master/buildbot/steps/shell.py
```diff
@@ -128,9 +127,23 @@ def __init__(self, workdir=None,
 
         # check validity of arguments being passed to RemoteShellCommand
         invalid_args = []
-        signature = inspect.signature(
-            remotecommand.RemoteShellCommand.__init__)
-        valid_rsc_args = signature.parameters.keys()
+        valid_rsc_args = [
```
tardyp, line 130: why?

tardyp (approved):
cool

master/buildbot/steps/shell.py
p12tic: We now replace RemoteShellCommand in tests with a wrapper function instead of FakeRemoteShellCommand. inspect can't therefore see the arguments.

ShellMixin also hardcodes the list of arguments, so I think this is actually an improvement.

PR #5475, opened 2020-08-20
Convert master steps to new style

p12tic:
This PR converts the MasterShellCommand and LogRenderable steps to new style.

## Contributor Checklist:


p12tic:
The coverage decrease is because the old-style step description methods are no longer covered. I think it does not make sense to spend time adding tests for them because these methods will soon be removed.

PR #5487, opened 2020-08-22
Docs: Add GNU make information on Windows

vx-i:
GNU make is not part of the standard Windows installation. So the documentation should explain how to install it before starting with the Buildbot development environment.

## Contributor Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ master/docs/buildbot.info-2
 .stgit-edit.txt
 .project
 .pydevproject
+.vscode
 .settings
 */build/
 master/docs/buildbot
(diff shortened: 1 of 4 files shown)


p12tic:
> but how to recognize that make is executed in pwsh.exe?

Aren't there any environment variables that would indicate this?

I was thinking about something like this which is used to select tar options (master/docs/Makefile):

```
TAR_TRANSFORM := $(if $(filter bsdtar,$(TAR_VERSION)),-s /^html/$(VERSION)/,--transform s/^html/$(VERSION)/)
```

vx-i:
I'm very sorry, but something went badly wrong. Now I have 33 commits. Can somebody help me with rebase and squash? Probably I need to revert this change, right?

p12tic:
I think you somehow forgot to pull Buildbot master on your local repository.

Try this:
```
git fetch -p origin
git rebase -i origin/master
```

Usually it's best to create PRs from a different branch than master, this way one can operate on the master branch as just like any other branch and e.g. do `git pull origin master` to update to new code on upstream repository.

p12tic:
If the above does not work, I can rebase it myself.

vx-i:
> If the above does not work, I can rebase it myself.

Rebase tells me 'noop', please try to rebase for me :-)

p12tic:
Fixed now.

PR #5498, opened 2020-08-23
Group related unit tests into directories

p12tic:
When running unit tests I often want to run tests related to particular area. E.g. all DB tests, or all data layer tests and so on. Unfortunately it's not possible to do this using trial, one needs to list all test modules on the command line, which is very inconvenient.

This PR solves this issue by grouping unit tests into separate directories, which are then made modules. This way it's possible to e.g. run all DB tests via `trial buildbot.test.unit.db`.


tardyp (approved):
cool!

PR #5501, opened 2020-08-24
Add implementation of ReconnectingWebSocket

vlasov-evgeny:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/add-implementation-of-reconnecting-web-socket.feature
+++ b/master/buildbot/newsfragments/add-implementation-of-reconnecting-web-socket.feature
@@ -1,0 +1,1 @@
+Add implementation of ``ReconnectingWebSocket`` to front-end.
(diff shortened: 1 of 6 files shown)


tardyp:
The problem with reconnectingWebsocket is that during the time of the disconnection, any events happening in between has been lost. So the client will not be able to reconstruct the state of the master.

This will lead to strange bug reports where the user sees different thing from two different browser windows.

Another less annoying for the user way is to reload every collection transparently from the rest API, when the reconnectingwebsocket is detecting a reconnection. This will ensure that the displayed data is always up to date.

vlasov-evgeny:
I'll take a look at how the lists should be updated in this case.

tardyp:
@vlasov-evgeny do you need any help for this?

p12tic:
Please reopen if/when this PR is worked on again. Thanks a lot!

PR #5507, opened 2020-08-29
fix github status breaking on Nightly scheduler sourcestamps

terencehonles:
The Nightly scheduler will create sourcestamps with mostly None values, and this change will handle None better.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/fix-github-status-errors.bugfix
+++ b/master/buildbot/newsfragments/fix-github-status-errors.bugfix
@@ -1,0 +1,1 @@
+Fixed :py:class:`~GitHubStatusPush` logging an error when triggered by the NightlyScheduler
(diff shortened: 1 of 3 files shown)


tardyp (changes requested):
change looks good. We require unit test to be updated for each change of code to make sure the code coverage is good, and the code is well tested. thanks for understanding

terencehonles:
> change looks good. We require unit test to be updated for each change of code to make sure the code coverage is good, and the code is well tested. thanks for understanding

Done, I was mainly trying to trigger the CI since I didn't have tests running locally. I think it might be worthwhile to refactor the tests into a set of known scenarios and and then try them across all the reporters rather than having each reporter have its own tests for the scenarios that are added in an ad hoc manner.

I don't have time to do that right now, but It might be a good thing to try to track.

p12tic:
> I think it might be worthwhile to refactor the tests into a set of known scenarios and and then try them across all the reporters rather than having each reporter have its own tests for the scenarios that are added in an ad hoc manner.

Agreed

p12tic (approved):

PR #5508, opened 2020-08-29
fix github webhook expecting data from unsuccessful API calls

terencehonles:
The change #5300 does not check for an error code from GitHub and will break when trying to read the filenames from what it assumes is a successful API request. I don't know if these APIs work if there is no token provided (how I am running buildbot), and instead of optionally including the authorization it probably should bail immediately. This change will at least respect the status code from GitHub. However I don't know if there are any implications of the commit message being the replacement string or the files being an empty list.

The APIs in question are: 
- happens to work, but changed to check the response code: https://docs.github.com/en/rest/reference/repos#get-a-commit
- broken (expects an array, but gets a dict which iterates its keys): https://docs.github.com/rest/reference/pulls#list-pull-requests-files

HTTP tracing logs:

```
http https://api.github.com/repos/nexleaf/NexleafDP/commits/ac8a90c90b26cef97a9ee3e80214d24e1d5bd83a {'headers': {'User-Agent': 'Buildbot'}}
==> 404: b'{"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/repos#get-a-commit"}'
http https://api.github.com/repos/nexleaf/NexleafDP/pulls/1761/files {'headers': {'User-Agent': 'Buildbot'}}
==> 404: b'{"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/pulls#list-pull-requests-files"}'
```

The error I recieved:

```
...
   File "/var/lib/buildbot/master.cfg", line 1045, in _get_pr_files
     return [f["filename"] for f in data]
   File "/var/lib/buildbot/master.cfg", line 1045, in <listcomp>
     return [f["filename"] for f in data]
 builtins.TypeError: string indices must be integers

```

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/www/test_hooks_github.py
+++ b/master/buildbot/test/unit/www/test_hooks_github.py
@@ -564,6 +564,9 @@ gitJsonPayloadCreateTag = b"""
 
   ]
 }"""
+
+gitJsonPayloadNotFound = b"""{"message":"Not Found"}"""
+
 _HEADER_CT = b'Content-Type'
 _CT_ENCODED = b'application/x-www-form-urlencoded'
 _CT_JSON = b'application/json'
@@ -863,24 +866,16 @@ class TestChangeHookConfiguredWithGitChange(unittest.TestCase,
                          "Update the README with new information\n"
                          "This is a pretty simple change that we need to pull into master.")
         self.assertEqual(change["branch"], "refs/pull/50/merge")
-        self.assertEqual(change['files'], ['README.md'])
+        self.assertEqual(change['files'], [])
         self.assertEqual(change["revlink"],
                          "https://github.com/defunkt/github/pull/50")
         self.assertEqual(change['properties']['basename'], "master")
         self.assertDictSubset(gitPRproperties, change["properties"])
 
     def test_git_with_pull_encoded(self):
-        commit_endpoint = '/repos/defunkt/github/commits/05c588ba8cd510ecbe112d020f215facb17817a7'
-        files_endpoint = '/repos/defunkt/github/pulls/50/files'
-        self._http.expect('get', commit_endpoint, content_json=gitJsonPayloadCommit)
-        self._http.expect('get', files_endpoint, content_json=gitJsonPayloadFiles)
         self._check_git_with_pull([gitJsonPayloadPullRequest])
 
     def test_git_with_pull_json(self):
-        commit_endpoint = '/repos/defunkt/github/commits/05c588ba8cd510ecbe112d020f215facb17817a7'
-        files_endpoint = '/repos/defunkt/github/pulls/50/files'
-        self._http.expect('get', commit_endpoint, content_json=gitJsonPayloadCommit)
-        self._http.expect('get', files_endpoint, content_json=gitJsonPayloadFiles)
         self._check_git_with_pull(gitJsonPayloadPullRequest)
 
     @defer.inlineCallbacks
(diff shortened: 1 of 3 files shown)


terencehonles:
It looks like @tardyp called it: https://github.com/buildbot/buildbot/pull/5300#discussion_r426153937

terencehonles:
> It looks like @tardyp called it: [#5300 (comment)](https://github.com/buildbot/buildbot/pull/5300#discussion_r426153937)

I updated the code to incorporate the requested changes, next time I might be more insistent that a contributor heeds your suggestion :confused:

master/buildbot/www/hooks/github.py
```diff
@@ -230,20 +230,24 @@ def _get_commit_msg(self, repo, sha):
...
 
         url = '/repos/{}/commits/{}'.format(repo, sha)
         http = yield httpclientservice.HTTPClientService.getService(
             self.master, self.github_api_endpoint, headers=headers,
             debug=self.debug, verify=self.verify)
         res = yield http.get(url)
-        data = yield res.json()
-        msg = data.get('commit', {'message': 'No message field'})['message']
-        return msg
+        if 200 <= res.code < 300:
```
tardyp, line 246: we should probably emit some kind of warning in this case, to let the bot owner know that github is failing some of the requests.

tardyp (changes requested):
again, I would like to get some  tests for this change, and also avoid silent error swallowing. thanks for you contribs!

master/buildbot/www/hooks/github.py
terencehonles: Done

master/buildbot/www/hooks/github.py
```diff
@@ -230,20 +231,27 @@ def _get_commit_msg(self, repo, sha):
...
+            data = yield res.json()
+            return data['commit']['message']
+        else:
```
p12tic, line 250: I think this else is not needed, we can just fall through o the end of the function.

master/buildbot/www/hooks/github.py
```diff
@@ -266,8 +278,14 @@ def _get_pr_files(self, repo, number):
             verify=self.verify,
         )
         res = yield http.get(url)
-        data = yield res.json()
-        return [f["filename"] for f in data]
+        if 200 <= res.code < 300:
+            data = yield res.json()
+            return [f["filename"] for f in data]
+        else:
```
p12tic, line 284: I think this else is not needed, we can just fall through o the end of the function.

master/buildbot/www/hooks/github.py
p12tic: Also, coverage results show that this error path is not covered by any tests. Can we add a test that covers it?

You may see "unhandled exception in logs" or similar errors after writing a test. To fix this, you'll need to add an expectation of failures in the logs like this: `self.assertEqual(len(self.flushLoggedErrors()), 1)`.

master/buildbot/www/hooks/github.py
p12tic: Also, coverage results show that this error path is not covered by any tests. Can we add a test that covers it?

p12tic (changes requested):
Your contribution looks great, thank you! The only thing which is left is covering a couple of error code paths with tests and the PR will be good to go.

master/buildbot/www/hooks/github.py
terencehonles: Done, not sure why I ended up adding the `else:` but it's been adjusted.

master/buildbot/www/hooks/github.py
terencehonles: Same

master/buildbot/test/unit/www/test_hooks_github.py
```diff
@@ -564,6 +564,9 @@
 
   ]
 }"""
+
+gitJsonPayloadNotFound = b"""{"message":"Not Found"}"""
```
p12tic, line 568: I would inline such small test data. Duplication is even beneficial in tests as it makes tests self-contained and thus easier to understand and change.

p12tic (approved):

master/buildbot/test/unit/www/test_hooks_github.py
terencehonles: Cool, I wasn't sure which you would prefer so I went with making it shared. Since you merged I'll just keep that in mind for my next PR (I'm assuming I'll have one at some point).

PR #5450, opened 2020-08-12
Fix missing VS2017 entry_points

j-carl:
Fix missing VS2017 entry_points in 'buildbot.steps'

## Contributor Checklist:

*  [ ] I have updated the unit tests
*  [x] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)
*  [ ] I have updated the appropriate documentation

PR #5510, opened 2020-08-29
2.8: Fix missing VS2017 entry_points

p12tic:
This backports PR #5450 to 2.8.x branch.


mojca (commented):
I don't know what this does, but wouldn't support for 2019 be in place as well?

Issue #4488, opened 2018-12-21
buildbot master bytes issue with python3 when triggered via buildbot try


albertfong:
OS: ubuntu:bionic (docker base image)
Python: Python 3.6.7
Buildbot version: 1.7.0
Twisted version: 18.9.0

When submitting a build via buildbot try, the buildbot try command succeeds, and the build master receives the job, adds the buildset, and starts the build. However, the master errors out, leaving the builds in a bad state (UI shows started and running, but no progress is made).

build master logs suggest a bytes/string issue with the 'branch' property:

```master_1_29527f36cb6f | 2018-12-21 21:56:54+0000 [Broker,0,172.21.0.5] user ci requesting build on builders None
master_1_29527f36cb6f | 2018-12-21 21:56:54+0000 [-] added buildset 30 to database
master_1_29527f36cb6f | 2018-12-21 21:56:55+0000 [-] starting build <Build projects-linux-clang number:None results:success> using worker <WorkerForBuilder builder='projects-linux-clang' worker='clang' state=AVAILABLE>
master_1_29527f36cb6f | 2018-12-21 21:56:55+0000 [-] <Build projects-linux-clang number:None results:success>.startBuild
master_1_29527f36cb6f | 2018-12-21 21:56:55+0000 [-] from a running build; this is a serious error - please file a bug at http://buildbot.net
master_1_29527f36cb6f |         Traceback (most recent call last):
master_1_29527f36cb6f |           File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 460, in callback
master_1_29527f36cb6f |             self._startRunCallbacks(result)
master_1_29527f36cb6f |           File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 568, in _startRunCallbacks
master_1_29527f36cb6f |             self._runCallbacks()
master_1_29527f36cb6f |           File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 654, in _runCallbacks
master_1_29527f36cb6f |             current.result = callback(current.result, *args, **kw)
master_1_29527f36cb6f |           File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 1475, in gotResult
master_1_29527f36cb6f |             _inlineCallbacks(r, g, status)
master_1_29527f36cb6f |         --- <exception caught here> ---
[...]

PR #5534, opened 2020-09-07
Fixes buildbot issues with buildbot try on python3, see issue #4488

sab24:
Hi, 

First pull request here, I have tried to commit the code that fixes buildbot try for me. I ran the unit tests and 4 failed with the current master code and 4 failed after, so my code does not cause issues. I see no reason to update the unit tests or the documentation.

### DIFF
--- a/master/buildbot/newsfragments/try_python3.bugfix
+++ b/master/buildbot/newsfragments/try_python3.bugfix
@@ -1,0 +1,1 @@
+Fixes deserialisation bug #4488 that occurs in buildbot try since python3
--- a/master/buildbot/process/properties.py
+++ b/master/buildbot/process/properties.py
@@ -190,6 +190,8 @@ class Properties(util.ComparableMixin):
 
     def setProperty(self, name, value, source, runtime=False):
         name = util.bytes2unicode(name)
+        if isinstance(value, bytes):
+            value = util.bytes2unicode(value)
         if not IRenderable.providedBy(value):
             json.dumps(value)  # Let the exception propagate ...
         source = util.bytes2unicode(source)


tardyp (changes requested):
We always required additional tests when a bug is fixed, so that we are sure this use case is taken in account later. In this case, you can just add a simple case in this test, which makes sure there is no crash when sending a byte into property.


https://github.com/buildbot/buildbot/blob/c8b9757aba98ec58e75e58c303b85b3b60371d2f/master/buildbot/test/unit/process/test_properties.py#L945

this will also help coverage report to be positive

p12tic:
I think I'll look into this myself. I'm not sure the bug fix is correct, also we would need to add tests to buildbot try. All of this is I think too much for a first time contributor.

sab24:
> I think I'll look into this myself. I'm not sure the bug fix is correct, also we would need to add tests to buildbot try. All of this is I think too much for a first time contributor.

The error is triggered in a production environment running python3 but test coverage shows that my change is not covered by unit testing. Does this mean that unit testing of buildbot try is still using python2?

p12tic:
It's just that the Buildbot try uses a different code path, that's why it's showed as not covered. I need to figure out why we are submitting bytes in the first place.

sab24:
> It's just that the Buildbot try uses a different code path, that's why it's showed as not covered. I need to figure out why we are submitting bytes in the first place.

I would appreciate it if you would take a look at this. 

Running biuldbot try I now get errors like  
```
fatal: Invalid refspec '\x6d6173746572'
```

From the git command
```
argv: ['git', 'fetch', '-t', 'git://github.com/buildbot/pyflakes.git', '\\x6d6173746572', '--progress']
```

when trying to run 
```
buildbot try --connect=pb --master=172.28.2.2:5555 --username=<user> --passwd=<pass> --vc=git --wait
```

This looks like a binary issue as well to me

mokibit:
I fixed the root cause in #5623.

p12tic:
Indeed, the root cause has been fixed in #5623, so I'm closing this PR. Thanks @sab24 for making the contribution.

PR #5544, opened 2020-09-11
worker: Don't reuse the started latent worker when it's incompatible

lephilousophe:
When a LatentWorker is substanciated, it is reused even when the build properties are not compatible.
For example, if the Docker image name is set using build properties. When one worker is started, even if the newer build uses a different Docker image, the old one will get reused and obviously fail.

This is due to an error when specifying inheritance in various LatentWorker.
For example with DockerLatentWorker, the MRO used by Python 3 will be : DockerLatentWorker, DockerBaseWorker, AbstractLatentWorker, AbstractWorker, CompatibleLatentWorkerMixin.
With this fix it becomes: CompatibleLatentWorkerMixin, DockerLatentWorker, DockerBaseWorker, AbstractLatentWorker, AbstractWorker.
This lets builds_may_be_incompatible and isCompatibleWithBuild to be taken from CompatibleLatentWorkerMixin as it's intended.

I don't see which part of the documentation should be updated.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/incompatible-worker-reused.bugfix
+++ b/master/buildbot/newsfragments/incompatible-worker-reused.bugfix
@@ -1,0 +1,1 @@
+Latent workers don't reuse the started worker when it's incompatible with the requested build.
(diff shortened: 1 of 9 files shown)


p12tic (approved):

p12tic:
Good catch, thank you!

PR #5546, opened 2020-09-14
Fix latent workers reject builds for prop mismatch

jangmarker:
This is a request of comments, because the logic of LatentWorker is quite complicated and I honestly cannot claim to understand it all in its depth. The issue this is trying to fix is as follows:

When a LatentWorker is taken down it does not clear its property list.
Then, when a new build tries to be scheduled on that worker it might be
rejected for a property mismatch, even though it would match because the
properties would be recreated when the latent worker is prepared for the
new build.

Note, that this is only an issue when the latent worker has been stopped
and the underlying instance, e.g. docker container, has been destroyed.
Otherwise, the worker would be correct to reject the not-matching build.

This has fixed an issue that we have been seeing with DockerLatentWorker, were workers had been scheduled with builds and then suddenly they wouldn't accept any new builds until Buildbot had been restarted.

## Contributor Checklist:

I will do these once I know that the general idea of this patch is valid:

### DIFF
--- a/master/buildbot/util/latent.py
+++ b/master/buildbot/util/latent.py
@@ -40,6 +40,9 @@ class CompatibleLatentWorkerMixin:
         self._actual_build_props = copy.deepcopy(props)
         defer.returnValue(props)
 
+    def resetWorkerPropsOnStop(self):
+        self._actual_build_props = None
+
     @defer.inlineCallbacks
     def isCompatibleWithBuild(self, build):
         if self._actual_build_props is None:
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -19,6 +19,7 @@ version = "1.10.6"
 
 class Client:
     latest = None
+    containerCreated = False
 
     def __init__(self, base_url):
         Client.latest = self
@@ -30,6 +31,9 @@ class Client:
         self._pullable = ['alpine:latest', 'tester:latest']
         self._pullCount = 0
         self._containers = {}
+
+        if Client.containerCreated:
+            self.create_container("some-default-image")
 
     def images(self):
         return self._images
--- a/master/buildbot/test/unit/worker/test_docker.py
+++ b/master/buildbot/test/unit/worker/test_docker.py
@@ -46,7 +46,10 @@ class TestDockerLatentWorker(unittest.TestCase, TestReactorMixin):
 
         self.build = Properties(
             image='busybox:latest', builder='docker_worker', distro='wheezy')
+        self.build2 = Properties(
+            image='busybox:latest', builder='docker_worker2', distro='wheezy')
         self.patch(dockerworker, 'client', docker)
+        docker.Client.containerCreated = False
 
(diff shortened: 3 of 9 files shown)


tardyp:
Is this a followup to https://github.com/buildbot/buildbot/pull/5544?

master/buildbot/util/latent.py
```diff
@@ -40,6 +40,9 @@ def renderWorkerPropsOnStart(self, build):
         self._actual_build_props = copy.deepcopy(props)
         defer.returnValue(props)
 
+    def resetWorkerPropsOnStop(self):
+        self._actual_build_props = None
```
tardyp, line 44: I guess https://github.com/buildbot/buildbot/pull/5544 was actually trying to avoid rescheduling on the same worker if the properties were incompatible. Is this an overlook in #5544 ?

I would avoid calling this in all subclass, but rather generically reset the _actual_build_props in the generic code.

tardyp (changes requested):

jangmarker:
Thanks your comments!

Yes, I believe it is a follow-up to #5544, but if I understand correctly it's fixing a case that #5544 does not cover:

Imagine a DockerLatentWorker that uses image `my-image` and has volumes which use `Interpolate` based on some builder property. Now assume there are two builders `builder1` and `builder2` that want to use this worker - each builder needs the worker to instantiate a different container because `builder1` requires different volumes than `builder2`. Buildbot has supported this setup so far if the DockerLatentWorker was configured to always only allow one build at the same time and would destroy the container as soon as the build was done - so it would create a new container for every build.

Now #5544 makes sure that no builds of `builder2` are scheduled on the worker as long as the container was instantiated for `builder1`. But when `builder1` is not building anymore the worker can in theory create a suitable container for `builder2`.
That was not possible with the bug that his patch is trying to fix.

I'm very happy to be shown that my understanding here is wrong, but this is what I could make of it so far.

master/buildbot/util/latent.py
jangmarker: I'm not sure where in the generic class `CompatibleLatentWorkerMixin` this could be done? There is no hook whatsoever that calls back into the `CompatibleLatentWorkerMixin` when the instance is destroyed.

Or do you mean that `AbstractLatentWorker` should call `CompatibleLatentWorkerMixin.resetWorkerPropsOnStop()`?
That would violate that `AbstractLatentWorker` isn't necessarily a `CompatibleLatentWorkerMixin`, wouldn't it?

tardyp (approved):
@p12tic I also would like your feedback on this

p12tic (approved):
Looks good to me, thanks!

p12tic:
Right, so what's left is a unit test to ensure the behavior does not regress in the future.

jangmarker:
I've now added a set of tests. I verified the tests against the previous code (where three of them fail as they are supposed to) and against the code with this patch (where they all pass).

p12tic:
Looks great, thank you!

PR #5548, opened 2020-09-16
Fixes reading HashiCorp Vault secrets

pavelgk:
## Contributor Checklist:

* [x]  I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)

### DIFF
--- a/master/buildbot/test/unit/test_secret_in_vault.py
+++ b/master/buildbot/test/unit/test_secret_in_vault.py
@@ -57,11 +57,18 @@ class TestSecretInVaultV1(TestSecretInVaultHttpFakeBase):
         self.assertEqual(value, "value1")
 
     @defer.inlineCallbacks
-    def testGetAnyKey(self):
+    def test_get_any_key_without_value_name(self):
         self._http.expect(method='get', ep='/v1/secret/any_key', params=None,
                           data=None, json=None, code=200,
-                          content_json={"data": {"any_key": "value1"}})
-        value = yield self.srvcVault.get("any_key")
+                          content_json={"data": {"any_value": "value1"}})
+        yield self.assertFailure(self.srvcVault.get("any_key"), KeyError)
+
+    @defer.inlineCallbacks
+    def test_get_any_key_with_value_name(self):
+        self._http.expect(method='get', ep='/v1/secret/any_key', params=None,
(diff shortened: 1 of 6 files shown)


master/buildbot/test/unit/test_secret_in_vault.py
```diff
@@ -110,23 +117,30 @@ def setUp(self):
...
         value = yield self.srvcVault.get("value")
         self.assertEqual(value, "value1")
 
+    @defer.inlineCallbacks
+    def testGetAnyKey(self):
+        self._http.expect(method='get', ep='/v1/secret/data/', params=None,
+                          data=None, json=None, code=200,
+                          content_json={"data": {"data": {"any_key": "value1"}}})
+        value = yield self.srvcVault.get("any_key")
+        self.assertEqual(value, "value1")
```
tardyp, line 132: How would I get this value from a Renderable object? This is not very clear for me, we need a test for it, and then documentation

tardyp (changes requested):
Thanks for this fix!

p12tic (approved):
Looks good to me, thanks a lot!

p12tic:
@pavelgk: Could you squash the commits to 1-2 commits so that the history is not polluted. Thanks!

pavelgk:
@p12tic : Sure. Should I update documentation? It looks fine to me.

p12tic:
Well, if the current documentation is not buggy it can stay as is :-)

master/buildbot/test/unit/test_secret_in_vault.py
p12tic: Fixed now.

p12tic:
So I've looked into adding a simple test covering the single line not covered in this PR and it turned out that the implementation is not fully correct as it breaks access by `key` identifier when the attribute name is not specified. I've fixed this and updated the PR.

Documentation on secret identifiers has been also added.

pavelgk:
@p12tic I've prepared documentation with screenshots about how to use vault. Is it possible to add it?

p12tic:
Oh, that's great news. Could you maybe create a separate PR for that?

pavelgk:
@p12tic could you review my PR #5569?

PR #5551, opened 2020-09-17
www: add the ability to easily copy build properties to the clipboard

terencehonles:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/copy-build-properties-to-clipboard.feature
+++ b/master/buildbot/newsfragments/copy-build-properties-to-clipboard.feature
@@ -1,0 +1,1 @@
+Add the ability to copy build properties to the clipboard.
--- a/www/base/src/app/common/directives/properties/properties.directive.js
+++ b/www/base/src/app/common/directives/properties/properties.directive.js
@@ -5,10 +5,30 @@ class Properties {
             restrict: 'E',
             scope: {properties: '='},
             template: require('./properties.tpl.jade'),
+            controller: '_propertiesController',
         };
+    }
+}
+
+function _properties($scope) {
+    $scope.copy = function(value) {
+        value = JSON.stringify(value);
+
+        if (navigator.clipboard && navigator.clipboard.writeText) {
+            navigator.clipboard.writeText(value);
+        } else {
+            var element = document.createElement('textarea');
+            element.style = 'position:absolute; width:1px; height:1px; top:-10000px; left:-10000px';
+            element.value = value;
+            document.body.appendChild(element);
+            element.select();
+            document.execCommand('copy');
+            document.body.removeChild(element);
+        }
     }
 }
 
 
 angular.module('common')
-.directive('properties', [Properties]);
+.directive('properties', [Properties])
+.controller('_propertiesController', ['$scope', _properties]);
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -78,6 +80,11 @@ frontend_tests_headless: frontend_deps
 frontend: frontend_deps
 	for i in pkg $(WWW_PKGS); do $(PIP) install -e $$i || exit 1; done
 
+# build frontend wheels for installation elsewhere
+frontend_wheels: frontend_deps
```
terencehonles, line 84: I didn't find any good testing documentation for the frontend so I used this and deployed the buildbot_www package on my buildbot master.

Makefile
```diff
@@ -143,7 +150,9 @@ tarballs: $(ALL_PKGS_TARGETS)
 
 
 clean:
+	@if [ -t 0 ]; then read -p 'You are about to delete all untracked files! (use [Ctrl+C] to abort) '; fi
```
terencehonles, line 153: This is a rather dangerous form of clean and will possibly cause someone to loose their work. I added a prompt if they are running this with a TTY so they can abort and this probably shouldn't block any existing uses of this target.

Makefile
```diff
@@ -1,12 +1,14 @@
...
 
 .PHONY: docs pylint flake8 virtualenv
 
 
-VENV_NAME:=.venv$(VENV_PY_VERSION)
-PIP?=$(VENV_NAME)/bin/pip
-VENV_PY_VERSION?=python3
+VENV_NAME := .venv$(VENV_PY_VERSION)
+PIP ?= $(ROOT_DIR)/$(VENV_NAME)/bin/pip
+PYTHON ?= $(ROOT_DIR)/$(VENV_NAME)/bin/python
```
terencehonles, line 10: Might it be better to activate the `venv` instead of have `pip` defined by a variable? Since this was an existing pattern I copied it.

terencehonles:
While I'm here, I had thought I could restart the buildbot master while the workers were still running (or at least there was an option to do so, I was quickly scanning the docs but didn't see it and figured it's not a high priority thing for me to find, but if you know the option I'd be happy to make sure it's configured)

Makefile
tardyp: I agree. I actually never use this make command

Makefile
```diff
@@ -1,12 +1,14 @@
 # developer utilities
 DOCKERBUILD := docker build --build-arg http_proxy=$$http_proxy --build-arg https_proxy=$$https_proxy
+ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
```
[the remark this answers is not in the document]
terencehonles: Done, the Makefile changes are in #5554

tardyp:
> While I'm here, I had thought I could restart the buildbot master while the workers were still running (or at least there was an option to do so, I was quickly scanning the docs but didn't see it and figured it's not a high priority thing for me to find, but if you know the option I'd be happy to make sure it's configured)

As for the makefile, not a very good idea to mix such issues. I missed it.
 you may want to use `buildbot restart --clean` so that the master waits for running builds to finish before restarting.

terencehonles:
> > While I'm here, I had thought I could restart the buildbot master while the workers were still running (or at least there was an option to do so, I was quickly scanning the docs but didn't see it and figured it's not a high priority thing for me to find, but if you know the option I'd be happy to make sure it's configured)
> 
> As for the makefile, not a very good idea to mix such issues. I missed it.
> you may want to use `buildbot restart --clean` so that the master waits for running builds to finish before restarting.

Ah ok, I had thought it was something in the config and that it would allow reconnecting to workers after restarting rather than waiting for them to cleanly finish. I have it configured with systemd, so I'll have to see about adding that as part of its shutdown procedure.

terencehonles:
It looks like I should be using `SIGUSR1` instead of `SIGTERM`

PR #5554, opened 2020-09-21
update Makefile

terencehonles:
- add target "frontend_wheels" to build frontend wheels for testing somewhere "pip install -e ." is not appropriate.
- remove target "clean" as it is more dangerous than most would expect (it does not just clean built artifacts).

## Contributor Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,14 @@
 # developer utilities
 DOCKERBUILD := docker build --build-arg http_proxy=$$http_proxy --build-arg https_proxy=$$https_proxy
+ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 .PHONY: docs pylint flake8 virtualenv
 
 
-VENV_NAME:=.venv$(VENV_PY_VERSION)
-PIP?=$(VENV_NAME)/bin/pip
-VENV_PY_VERSION?=python3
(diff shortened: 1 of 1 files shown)


terencehonles:
See pre existing comments:

- https://github.com/buildbot/buildbot/pull/5551#discussion_r490487897
- https://github.com/buildbot/buildbot/pull/5551#discussion_r490488484
- https://github.com/buildbot/buildbot/pull/5551#discussion_r490489480

Makefile
```diff
@@ -143,7 +150,9 @@ tarballs: $(ALL_PKGS_TARGETS)
 
 
 clean:
+	@if [ -t 0 ]; then read -p 'You are about to delete all untracked files! (use [Ctrl+C] to abort) '; fi
 	git clean -xdf
```
p12tic, line 154: Maybe it makes sense to remove this rule altogether? It seems that the user is better off just calling `git clean` manually.

tardyp: agreed.

tardyp (changes requested):
lets remove make clean. +1 for the rest

Makefile
terencehonles: sounds good to me, wasn't sure if it was used anywhere

tardyp (approved):

PR #5565, opened 2020-10-06
Allow user to set docker init conf default

jsonar-cpapke:
I am trying to use docker latent workers to test a program that runs on systemd. But it seems that if init=True in the hostconfig, the docker container will terminate itself right after starting.

This change allows the setting of the init flag if the user wants. But it shouldn't affect anyone who doesn't explicitly set the flag, as the default is still the same.

There don't seem to be any documentation changes required related to this, as the only documentation to refer to hostconfig points to the docker-py documentation.

## Contributor Checklist:


tardyp (approved):
great work thanks!

PR #5567, opened 2020-10-07
bug fix for api v4 in enterprise github

xeniorac:
Hi Github api fails for enterprise version. I found a fix by trying the graphQL query using a REST API and tested if it works with BuildBot as well... 
This is my first contribution to Buildbot Open Source Library

## Contributor Checklist:

* [ N/A ] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/newsfragments/githuboauth2apiv4.bugfix
+++ b/master/buildbot/newsfragments/githuboauth2apiv4.bugfix
@@ -1,0 +1,1 @@
+Fixed Github v4 API URL
--- a/master/buildbot/test/unit/www/test_oauth.py
+++ b/master/buildbot/test/unit/www/test_oauth.py
@@ -78,12 +78,16 @@ class OAuth2Auth(TestReactorMixin, www.WwwTestMixin, ConfigErrorsMixin,
             "ghclientID", "clientSECRET", apiVersion=4, getTeamsMembership=True)
         self.githubAuthEnt = oauth2.GitHubAuth(
             "ghclientID", "clientSECRET", serverURL="https://git.corp.fakecorp.com")
+        self.githubAuthEnt_v4 = oauth2.GitHubAuth(
+            "ghclientID", "clientSECRET", apiVersion=4, getTeamsMembership=True,
+            serverURL="https://git.corp.fakecorp.com")
         self.gitlabAuth = oauth2.GitLabAuth(
             "https://gitlab.test/", "glclientID", "clientSECRET")
         self.bitbucketAuth = oauth2.BitbucketAuth("bbclientID", "clientSECRET")
 
         for auth in [self.googleAuth, self.githubAuth, self.githubAuth_v4, self.githubAuth_v4_teams,
-                     self.githubAuthEnt, self.gitlabAuth, self.bitbucketAuth]:
+                     self.githubAuthEnt, self.gitlabAuth, self.bitbucketAuth,
+                     self.githubAuthEnt_v4]:
             self._master = master = self.make_master(url='h:/a/b/', auth=auth)
             auth.reconfigAuth(master, master.config)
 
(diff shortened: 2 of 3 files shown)


xeniorac:
@tardyp

master/buildbot/www/oauth2.py
```diff
@@ -248,7 +248,14 @@ def __init__(self,
                     'Retrieving team membership information using GitHubAuth is only '
                     'possible using GitHub api v4.')
         else:
-            self.apiResourceEndpoint = self.serverURL + '/graphql'
+            if serverURL is not None:
+                # setup for enterprise github
+                if serverURL.endswith("/"):
```
tardyp, line 253:
```suggestion
               serverURL = serverURL.lstrip("/"):
```

tardyp: Please put all the onprem vs saas logic near line 231 including computing of the self.graphqlEndpoint

tardyp (changes requested):
we also need tests for this code.

tardyp:
why do you say you have updated the tests while you have not?

master/buildbot/test/unit/www/test_oauth.py
```diff
@@ -157,6 +159,20 @@ def test_getGithubELoginURL(self):
                "scope=user%3Aemail+read%3Aorg")
         self.assertEqual(res, exp)
 
+    @defer.inlineCallbacks
+    def test_getGithubELoginURL_v4(self):
```
tardyp, line 163:
```suggestion
    def test_getGithubLoginURL_v4(self):
```

master/buildbot/test/unit/www/test_oauth.py
```diff
@@ -157,6 +159,20 @@ def test_getGithubELoginURL(self):
...
+        res = yield self.githubAuthEnt_v4.getLoginURL('http://redir')
+        exp = ("https://git.corp.fakecorp.com/login/oauth/authorize?client_id=ghclientID&"
```
tardyp, line 165: usually we use example.com for example urls

master/buildbot/www/oauth2.py
```diff
@@ -237,6 +237,9 @@ def __init__(self,
             self.tokenUri = '{0}/login/oauth/access_token'.format(serverURL)
         self.serverURL = serverURL or self.resourceEndpoint
 
+        enterpriseGraphqlEndpoint = None
```
tardyp, line 240: Please do the lstrip() line 231 (overlook of a previous PR)

Please simplify the code to avoid two `if serverURL is not None` lines.

master/buildbot/www/oauth2.py
```diff
@@ -248,7 +251,8 @@ def __init__(self,
                     'Retrieving team membership information using GitHubAuth is only '
                     'possible using GitHub api v4.')
         else:
-            self.apiResourceEndpoint = self.serverURL + '/graphql'
+            defaultGraphqlEndpoint = self.serverURL + '/graphql'
+            self.apiResourceEndpoint = enterpriseGraphqlEndpoint or defaultGraphqlEndpoint
```
tardyp, line 255: you can define default self.apiResourceEndpoint line 228 and then overwrite it line 237.

tardyp (changes requested):
we still need some cleanup on this code.

master/buildbot/www/oauth2.py
xeniorac: oh okay, did an lstrip but probably misunderstood the ask. sorry about that, on it

master/buildbot/test/unit/www/test_oauth.py
xeniorac: So does that mean I need to use git.example.com instead of git.corp.fakecorp.com?

master/buildbot/www/oauth2.py
xeniorac: got it

master/buildbot/www/oauth2.py
```diff
@@ -226,12 +226,14 @@ def __init__(self,
                  **kwargs):
 
         super().__init__(clientId, clientSecret, autologin, **kwargs)
+        self.apiResourceEndpoint = None
         if serverURL is not None:
             # setup for enterprise github
-            if serverURL.endswith("/"):
-                serverURL = serverURL[:-1]
+            serverURL = serverURL.rstrip("/")
```
xeniorac, line 232: Also another question. The if here before removed the '/' at the end of the string.
Shouldn't it be an rstrip that does the same?

master/buildbot/test/unit/www/test_oauth.py
p12tic: Yes.

master/buildbot/www/oauth2.py
p12tic: Agreed, this looks better.

master/buildbot/newsfragments/githuboauth2apiv4.bugfix
```diff
@@ -0,0 +1 @@
+Appended server URL with `/api/graphql` as a standard instead of `/graphql`
```
p12tic, line 1: I think the impact of the bugfix would be clearer if we said something like "Fixed Github v4 API URL"

master/buildbot/test/unit/www/test_oauth.py
p12tic: Actually, in this case I see that the test was already using fakecorp domain before this PR, so I think that for consistency using it in subsequent test cases makes sense. Cleanup could happen in a separate unrelated PR, so you don't need to do anything.

p12tic:
Looks good to me, just fix the newsfragment and it's good to merge.

xeniorac:
@p12tic @tardyp 
Fixed the newsfragments.

p12tic (approved):

p12tic:
Thanks a lot!

PR #5568, opened 2020-10-08
Fix known hosts for wrapper script

jsonar-cpapke:
I noticed an issue where if the version of git is less than 2.3.0, the known hosts were not being used in the git call as they were not being added to the wrapper script.

I believe that the change I made shouldn't affect anyone in any other way than making this work as expected.

I don't think any documentation needs to be updated for this change, as the documentation implies that this should work already.

## Contributor Checklist:


master/buildbot/test/unit/steps/test_source_git.py
```diff
@@ -526,7 +541,11 @@ def test_mode_full_clean_ssh_host_key_1_7(self):
         self.expectOutcome(result=SUCCESS)
         self.expectProperty(
             'got_revision', 'f6ad368298bd941e934a41f3babc827b2aa95a1d', self.sourceName)
-        return self.runStep()
+        yield self.runStep()
+
+        expected = b'#!/bin/sh\nssh -i "%s" -o "UserKnownHostsFile=%s" "$@"\n' \
```
p12tic, line 546: Can we use `.format()` for string formatting?

master/buildbot/test/unit/steps/test_source_git.py
```diff
@@ -309,7 +316,10 @@ def test_mode_full_clean_ssh_key_1_7(self):
         self.expectOutcome(result=SUCCESS)
         self.expectProperty(
             'got_revision', 'f6ad368298bd941e934a41f3babc827b2aa95a1d', self.sourceName)
-        return self.runStep()
+        yield self.runStep()
+
+        expected = b'#!/bin/sh\nssh -i "%s" "$@"\n' % (unicode2bytes(ssh_key_path),)
```
p12tic, line 321: Can we use `.format()` for string formatting?

p12tic (changes requested):
Overall looks great, thank you!

The only nitpick that I had was to use python3 string formatting functions. I'll merge the PR once this gets fixed.

p12tic:
Also, could you squash some of the commits so that they don't pollute history. E.g. commits that fix breakage created in the same PR could be squashed with the commit that introduce the breakage. Also, please use a prefix for the area in the commit messages, e.g. `git:` or `steps:` in this case.

jsonar-cpapke:
Sorry for the delay on updating this pull request.
I have squashed some of the commits into one. I wasn't sure exactly which commits you wanted prefixed, so I only prefixed the one that affected the code, but not the tests. If you want me to prefix the test commits as well, let me know and I will hopefully be able to get that done! (quicker this time)

master/buildbot/test/unit/steps/test_source_git.py
jsonar-cpapke: Changed to using `.format()`!

master/buildbot/test/unit/steps/test_source_git.py
jsonar-cpapke: Changed to using `.format()`!

p12tic (approved):
LGTM, thanks!

PR #5577, opened 2020-10-12
Allow explicit obfuscation of RemoteCommand arguments.

### LINKED ISSUE #3954: Can't use secret interpolation in p4poller.P4Source and steps.P4
I tried to use secrets for perforce passwords (using buildbot 1.0.0).

First I tried to use it in the P4 change source:
p4ChangeSource = p4poller.P4Source(p4port=p4Port,
                                                           p4user=p4User,
                                                           p4passwd=util.Secret('p4passwd'),
                                                           ...)
which fails with:
... [-] P4 poll failed ..
...
exceptions.TypeError: Arguments contain a non-string value

Then I tried to use the secret in the P4 source checkout step:
factory.addStep(steps.P4(p4port=p4Port,
                         ...,
                         p4passwd=util.Secret('p4passwd'),
                         ....)
This fails with:
Upon execvpe p4 ['p4', ..., '-P', Unpersistable('Unpersistable data: instance of class buildbot.process.properties.Secret deemed insecure'), ...] in environment id 56699632
:Traceback (most recent call last):
  File "/.../internet/process.py", line 445, in _fork environment)
  File "/.../internet/process.py", line 523, in _execChild os.execvpe(executable, args, environment)
  File "/.../lib/python2.7/os.py", line 353, in execvpe _execvpe(file, args, env)
  File "/.../lib/python2.7/os.py", line 380, in _execvpe func(fullname, *argrest)
TypeError: coercing to Unicode: need string or buffer, instance found

I suppose, the secrets are still not implemented for perforce functions.
Or am I doing something wrong?

### LINKED ISSUE #5566: steps.SVN: support username as Secret
Hi! In `steps.SVN` I can set password as `util.Secret`, but not username. I'd like to hide username in the logs and store secrets together. I have a workaround, but I really want this code to just work:

```
factory.addStep(steps.SVN(
    repourl='https://...',
    username=util.Secret('svn_username'),  # only `str` instance will work
    password=util.Secret('svn_password'),
    codebase='trunk',
    haltOnFailure=True,
    preferLastChangedRev=True,
    alwaysUseLatest=True,
))
```

senyai:
Plus marked username and password renderable in SVN and P4.

Closes #3954 and closes #5566.

Related documentation will be in a separate PR.

### DIFF
--- a/master/buildbot/process/remotecommand.py
+++ b/master/buildbot/process/remotecommand.py
@@ -33,6 +33,11 @@ class RemoteException(Exception):
     pass
 
 
+class ObfuscatedArgument(tuple):
+    def __new__(cls, real, fake='XXXXXX'):
+        return tuple.__new__(ObfuscatedArgument, ('obfuscated', real, fake))
+
+
 class RemoteCommand(base.RemoteCommandImpl):
 
     # class-level unique identifier generator for command ids
@@ -359,6 +364,7 @@ class RemoteShellCommand(RemoteCommand):
         if decodeRC is None:
             decodeRC = {0: SUCCESS}
         self.command = command  # stash .command, set it later
+        self.obfuscated = False
         if isinstance(self.command, (str, bytes)):
             # Single string command doesn't support obfuscation.
             self.fake_command = command
@@ -366,6 +372,7 @@ class RemoteShellCommand(RemoteCommand):
             # Try to obfuscate command.
             def obfuscate(arg):
                 if isinstance(arg, tuple) and len(arg) == 3 and arg[0] == 'obfuscated':
+                    self.obfuscated = True
                     return arg[2]
                 return arg
             self.fake_command = [obfuscate(c) for c in self.command]
@@ -404,8 +411,14 @@ class RemoteShellCommand(RemoteCommand):
                 # buildbot-worker doesn't support worker-configured usePTY,
                 # and usePTY defaults to False.
                 self.args['usePTY'] = False
-
-        self.args['command'] = self.command
+        if self.obfuscated and self.step.workerVersionIsOlderThan('shell', '2.16'):
+            log.msg("WARNING! Worker does not understand obfuscation; "
+                    "real values will be logged")
+            command = [c[1] if isinstance(c, tuple) and len(c) == 3 and c[0] == 'obfuscated'
+                       else c for c in self.command]
+        else:
+            command = self.command
+        self.args['command'] = command
         if self.remote_command == "shell":
             # non-ShellCommand worker commands are responsible for doing this
             # fixup themselves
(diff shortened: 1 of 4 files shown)


p12tic:
Thanks a lot for the PR.

The idea makes sense in principle, but I think it should be better integrated with secrets somehow. Right now I think what could make sense is a separate argument to `Secret` to tell whether the logs or only the program arguments should be cleaned up. Additionally we could have SecretValue class which would work exactly like a regular secret except that the value would be provided directly as an argument (i.e. this would effectively be equivalent to ObfuscatedArgument suggested in this PR).

Let me think about this more. I'll get back once I have something concrete to suggest.

senyai:
That's exactly my thoughts. But the worker has no renderables, and this len(3) tuple is the only way to hide something on the worker. Even if we can come up with a better solution, that will be as general as possible and will be able to do something like util.Interpolate('-r %(obfuscate:key)s'), my PR is a first step. /me trying to understand how to run codecov for a single module.

p12tic:
> But the worker has no renderables, and this len(3) tuple is the only way to hide something on the worker. 

Yes, the worker will get the old len(3) tuples. But the translation to these tuples will happen on the master and we have design freedom for the inputs of that translation.

> /me trying to understand how to run codecov for a single module.

Do you see codecov results when clicking on a PR number in the codecov report? That's what I'm using myself.

senyai:
I was looking for this command `trial --coverage buildbot.test.unit.steps.test_source_p4.TestP4`

p12tic:
@senyai I've looked into details of how full solution of obfuscated arguments could be implemented. I think obfuscating just the command-line arguments, but not the logs would be relatively complex task, as whole property rendering infrastructure would need to be changed. Additionally, it would be error-prone, as some commands could echo part of the arguments to the logs which would not be obfuscated.

I suggest that for the beginning a hardcoded `SecretValue` could be implemented. I suggest to look into the implementation of the `Secret` class for inspiration, in this case we would need to pass the `secret_value` to the constructor instead of its name.

In order for support `secret_value:value` in Interpolate, you'd need to implement a `_parse_secret_value` method there.

By the way, the support for secrets in SVN and P4 arguments could go separately earlier.

Once the above is done, we can look into implementing secrets that are only obfuscated from command line if you'd still be interested in that.

senyai:
@p12tic I think I understand what you mean by `SecretValue`, but I don't see how it will allow me to pass a simple username as an obfuscated argument. I was thinking more about adding `UnsafeSecret` class, that will act like `Secret`, unless it's marked safe (in svn or p4 step, for example), then `UnsafeSecret` will be rendered as tuple. Maybe I'm just imagining it. Let's close this one.

PR #5578, opened 2020-10-12
fix renderable issue when using P4 stream

bschatt:
Fixes this error when using a renderable for p4base or p4branch in stream mode.
```
  File "/home/buildbot/.local/lib/python3.7/site-packages/buildbot/steps/source/p4.py", line 119, in __init__
    if " " in p4base or " " in p4branch:
builtins.TypeError: argument of type '_Renderer' is not iterable
```

### DIFF
--- a/master/buildbot/newsfragments/p4-renderable.bugfix
+++ b/master/buildbot/newsfragments/p4-renderable.bugfix
@@ -1,0 +1,1 @@
+Fix support of renderables for `p4base`` and ``p4branch`` arguments of the P4 step.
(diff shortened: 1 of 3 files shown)


p12tic (changes requested):
Looks good, but can we add tests? A newsfragment is probably also useful because this bug is very easy to trigger. Thanks!

senyai:
I think these checks should be done in `_createClientSpec` (`raise buildstep.BuildStepFailed()` in case of an error), because there's no reason to ignore IRenderable.

bschatt:
The same thing is done a few lines above. https://github.com/buildbot/buildbot/blob/a2dc9a66a9cb38efa5a13f8361733c97f8246e29/master/buildbot/steps/source/p4.py#L101

I believe the reason is that function check for errors at config time, not run time. The perforce command will fail if the path is incorrect anyway, right?

p12tic (approved):
LGTM. I've fixed a typo in a test name and added newsfragment.

PR #5585, opened 2020-10-18
Merge HttpStatusPushBase and NotifierBase

p12tic:
This PR makes HttpStatusPushBase to derive from NotifierBase and reuse the report generator functionality.

Documentation on newly exposed APIs will be updated in a subsequent PR.

## Contributor Checklist:

* [n/a] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)

### DIFF
--- a/master/buildbot/reporters/http.py
+++ b/master/buildbot/reporters/http.py
@@ -13,29 +13,56 @@
 #
 # Copyright Buildbot Team Members
 
-import abc
-
 from twisted.internet import defer
 from twisted.python import log
 
 from buildbot import config
-from buildbot.reporters import utils
+from buildbot.reporters import notifier
+from buildbot.reporters.generators.build import BuildStatusGenerator
+from buildbot.reporters.message import DeprecatedMessageFormatterBuildJson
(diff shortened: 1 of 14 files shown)


master/buildbot/reporters/http.py
```diff
@@ -45,58 +72,27 @@ def reconfigService(self, builders=None, debug=None, verify=None,
...
-                                       wantSteps=self.wantSteps,
-                                       wantPreviousBuild=self.wantPreviousBuild,
-                                       wantLogs=self.wantLogs)
-        if self.filterBuilds(build):
-            yield self.send(build)
-
-    @abc.abstractmethod
-    def send(self, build):
-        pass
+    # Deprecated overridden method, will be removed in Buildbot 3.0
```
tardyp, line 93: looks like an overlook

tardyp (approved):

master/buildbot/reporters/http.py
p12tic: I've added this as a documentation that subclasses may implement this method and we need to handle it.

PR #5587, opened 2020-10-18
p4 step: use proper exception

senyai:


p12tic (changes requested):
I think in this case we need to have a separate exception type. BuildStepFailed currently reports all issues that are encountered when running the actions of the step, whereas WorkerTooOldError repors all issues related to the worker setup.

I think we should rename `WorkerTooOldError` to `WorkerSetupError` instead.

p12tic:
Check out deprecatedModuleAttribute for a way to provide the old name with a deprecation warning.

senyai:
Agree. Will make different PR when I have time.

PR #5590, opened 2020-10-18
properties: more accurate way to cleanup secrets

senyai:
Do as comment says ;)
Fixes case when one secret is a substring of another.


p12tic:
Thanks for finding this. Could we also have a unit test that covers this case? This area will likely be refactored in the future for performance and we better not undo this commit.

PR #5591, opened 2020-10-19
[DO NOT MERGE] test for twisted trunk

tardyp:
for twisted release management, we test buildbot with twisted trunk from time to time


p12tic:
I assume all testing that was necessary has been completed.

rodrigc:
I'm reopening this, because this looks like a bug on the Twisted side.  I filed this bug: https://twistedmatrix.com/trac/ticket/10070

Please leave this open for a bit as I try to track this down and push a fix into Twisted, which has gone through a lot of changes in the past year.

p12tic:
https://github.com/twisted/twisted/pull/1502 is now merged, I think we can close this PR too.

PR #5598, opened 2020-10-24
Renamed WorkerTooOldError to WorkerSetupError

senyai:
Because `WorkerTooOldError` is never about an old worker.

# Contributor Checklist:


master/buildbot/interfaces.py
```diff
@@ -33,10 +35,19 @@ class BuilderInUseError(Exception):
...
 
 
-class WorkerTooOldError(Exception):
+class WorkerSetupError(Exception):
     pass
 
 
+WorkerTooOldError = WorkerSetupError
+deprecatedModuleAttribute(
+    Version("buildbot", 2, 8, 5),
```
p12tic, line 44:
```suggestion
    Version("buildbot", 2, 9, 0),
```

We only deprecate APIs on minor version bumps.

p12tic (changes requested):
Overall looks good, thank you!

Could you add a newsfragment?

p12tic:
Also, it seems that the `isort` check now fails, meaning that the import order needs to be changed due to renamed import.

p12tic:
Thanks!

Issue #5136, opened 2019-12-26
latent: Can't stop build when it's preparing worker


p12tic:
I've seen this on buildbot.buildbot.net and didn't debug it so I'm not sure if it's caused by Kubernetes or by a general bug in latent worker implementation.

tardyp:
Indeed, the way of stopping is to interrupt the current step. This is not supported by all the steps. 
For such steps, we will just wait until the step is finished, and stop before starting the next.
Basically, only ShellCommand and TriggerStep's interrupt are implemented AFAIR, the others are so quick it is not worth to implement.

For preparation, the is a fake step.
https://github.com/tardyp/buildbot/blob/6cf74f2b1876222fa585b79669f969b958ab9929/master/buildbot/process/build.py#L315

We could hack and change the interrupt callback at each intermediate step of the preparation phase, but I think that would be a bit dangerous.

benzcash:
I'm seeing this too on a kubernetes install. How I I reset a `preparing worker` job?

iusyk:
> I'm seeing this too on a kubernetes install. How I I reset a `preparing worker` job?

Looks like no way, I have a test without the steps at all - preparing is going ........

p12tic:
FYI @benzcash @lusyk: I believe this bug has been fixed in #5599.

p12tic:
Please reopen the bug if you still see any issues like this.

jpflori:
Hi all,

I have as slightly related issue with libvirt latent workers.

The connection got lost but the build went on, or rather kept waiting for worker feedback.
I tried to stop the build by clicking the stop button but it does not work and stays in yellow/building.
It only adds new lines with:
`no reason`
or
`[Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectionLost'>: Connection to the other side was lost in a non-clean fashion.
]`

I'm using 2.10.0 from pip which must not include  #5599 
I'll report once this gets released.

If there is any way to kill the build without restarting the master that would be nice to know.

p12tic:
@jpflori: Could you please file a separate bug with more information? A fuller traceback would help a lot. #5599 is included into 2.10.0, so this is a different issue, maybe related to our libvirt implementation of latent workers.

Issue #3459, opened 2017-07-21
latentworker cannot escape "preparing worker" if the worker created but never connects


jonnyyu:
in some edge case, the created worker did created, so the code returns from substantiate method and starts waiting for attached event.

but for some various reasons, the worker didn't launch successfully, thus it didn't connect back to buildbot master.  In this situation, this build seems stuck waiting for the worker to connect forever, without any timeout mechanism.

tardyp:
There is a missing_timeout that the latent worker are using to detect if the latent fail to connect. which latent worker are you using?

seankelly:
And what version of Buildbot?

jerrykan:
I'm seeing this problem with the `DockerLatentWorker` in buildbot v0.9.10.

The build seems to try and continue forever and even clicking on cancel doesn't stop it.

jstacoder:
👍🏼

vdsbenoit:
I have the same kind of issue with Buildbot Travis. I use Docker for my latent workers. 

I have a spawner job that runs 4 build jobs. Systematically at least 1 of the 4 latent workers fails at creating a container. The job is stuck on "preparing worker" and there is no way to manually stop the build (stop button has no effect). The build is cancelled after 1h timeout (very long).

![image](https://user-images.githubusercontent.com/44053930/60273910-6c21a880-98f7-11e9-9c3b-7cff517209a4.png)

This happens only with Windows Docker containers. When I perform a `docker ps` I can see that only 3 containers exist while the 4th job is still stuck on "preparing worker". 

Maybe Buildbot could try to create another container if it receives no responses from Docker. Or at least, there could be a way to force a build stop.

Here is my configuration:
Buildbot version: 2.0.1
Twisted version: 18.9.0
Buildbot-travis version: 0.6.4 
Docker Windows Server version: 18.09.2, build 6247962

vdsbenoit:
The hard fix on my side was to set the DEFAULT_MISSING_TIMEOUT in AbstractWorker to 60. 

After 60 seconds, Buildbot try to create the container again, and it works.

tardyp:
This can happen if the docker container does not start correctly, cannot
contact the master, etc.
The missing timeout is used for that kind of issue as well.

So this is usually an infra issue that you need to try and debug.
You can look at the container list on your docker machine and look at the
log to see any issue
docker ps
docker logs
etc.

Pierre

vdsbenoit:
Hi Pierre,

[...]

Issue #2948, opened 2017-03-09
*LatentBuildSlave doesn't get properly reconfigured


tardyp:
This ticket is a migrated Trac ticket [3336](http://trac.buildbot.net/ticket/3336)

People contributed to the original ticket: @djmitche, `benoit.allard@...`, @tomprince, @unknown_contributor, @tardyp, @sa2ajj, @unknown_contributor, @unknown_contributor
Ticket created on: `Aug 14 2015`
Ticket last modified on: `Jun 21 2016`

---

all the !Latent[[BuildSlave]] hierarchy redefine the `__init__`, and don't call parent, so all the reconfigure stuff does not work for latent slaves

-- original report:

I had modified the dockerfile parameter, and the log was still showing the execution of a very old version of that file.

Note: the content of the file is given as string to the constructor.

---

_Comment from_: @tardyp
_Date_: `Sep 13 2015`

Looking at the code, the slave is reconfigured, but the image is not recreated.

The image is actually cached between builds, so we have to invalidate that cache if at reconfigure we found out that the image code changed.

---
_Comment from_: @tardyp
_Date_: `Sep 13 2015`

wait a minute.. Indeed, all the [[LatentBuildSlave]] hierarchy redefine the __init__, and dont call parent, so all the reconfigure stuff does not work.

---
_Comment from_: @dustin
_Date_: `Sep 14 2015`

Maybe I'm missing something, but I see all of these constructors called..

     56 class [[DockerLatentBuildSlave]]([[AbstractLatentBuildSlave]]):
     57     instance = None
     58 
     59     def __init__(self, name, password, docker_host, image=None, command=None,
     60                  volumes=None, dockerfile=None, version=None, tls=None, follow[[StartupLogs]]=False, 
     61                  **kwargs):
    ....
     91         [[AbstractLatentBuildSlave]].__init__(self, name, password, **kwargs)

    623 class [[AbstractLatentBuildSlave]]([[AbstractBuildSlave]]):
    ....
    642     def __init__(self, name, password,
    643                  build_wait_timeout=60 * 10,
    644                  **kwargs):
    645         [[AbstractBuildSlave]].__init__(self, name, password, **kwargs)

(`[[AbstractBuildSlave]]` has no `__init__` so this calls back to `[[BuildbotService]].__init__`)

---
_Comment from_: @tardyp
_Date_: `Sep 14 2015`

[...]

PR #5599, opened 2020-10-25
Fix docker latent worker reconfig issues

jangmarker:
These commit attempts to fix two issues that we have been seeing with docker workers and reconfigurations:
1. reconfigurations can make buildbot unresponsive, i.e. it wouldn't pick up builds and it wouldn't gracefully restart
2. builds that are stuck in the preparation state (of the docker latent worker) cannot be stopped

I introduced these changes to the best of my understanding and I've tested them (details in the commit messages).
However, I'd appreciate a careful review because I am not very familiar with this complex piece of code.

This could help with fixing #3459 and seems related to #2948 (which is pretty old).

## Contributor Checklist:

  - not yet, but I'd be happy to write tests if the fixes are deemed correct
* [ ] ~I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)~
* [ ] ~I have updated the appropriate documentation~

### DIFF
--- a/master/buildbot/master.py
+++ b/master/buildbot/master.py
@@ -268,11 +268,13 @@ class BuildMaster(service.ReconfigurableServiceMixin, service.MasterService):
 
             yield self.mq.setup()
 
+            # the buildbot scripts send the SIGHUP signal to reconfig master
             if hasattr(signal, "SIGHUP"):
                 def sighup(*args):
                     eventually(self.reconfig)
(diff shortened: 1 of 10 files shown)


p12tic (approved):
Looks good, thank you. I'll try to write a unit test before merging.

p12tic:
@jangmarker: Thanks again for this pull request. I've cleaned the code a little bit, but the basic idea remains.

I think this PR fixes #5136.

PR #5606, opened 2020-11-01
buildbot.data.types.DateTime is an integer

### LINKED ISSUE #5529: Cannot filter on started_at in REST API
I wanted to list all builds in the last week using the REST API.  `.../api/v2/builds?order=-started` gets the list of all builds ever sorted in date order, but adding for example `started_at__ge=1598375285` to filter to just builds started after a specific timestamp results in an `invalid filter value for started_at__ge`.

senyai:
fixes #5529 
The issue is that DateTime class has default `valueFromString` that throws `TypeError`.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/data/types.py
+++ b/master/buildbot/data/types.py
@@ -116,8 +116,25 @@ class Integer(Instance):
 
 
 class DateTime(Instance):
+
     name = "datetime"
-    types = (datetime.datetime)
+    types = (datetime.datetime,)
+
+    def valueFromString(self, arg):
+        return int(arg)
+
+    def validate(self, name, object):
+        if isinstance(object, datetime.datetime):
+            return
+        if isinstance(object, int):
+            try:
+                datetime.datetime.fromtimestamp(object)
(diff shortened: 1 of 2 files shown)


senyai:
I will fix the tests

p12tic:
I think there was a reason why this is an integer. Maybe @tardyp knows why? Don't hurry with tests just yet.

tardyp:
I think there was no particular reason. Just an overlook on the usecase.

I think that it makes sense to also allow 2010-10-25 like strings. If there are dash in the screen, then try to parse a date with iso8601 derivatives

maybe something like this:

YYYY-MM-dd
YYYY-MM-ddTHH:mm
YYYY-MM-ddTHH:mm:ss
YYYY-MM-ddTHH:mm:ss.ss
YYYY-MM-ddTHH:mm:ss.ss±hh:mm

my2c

tardyp (approved):

tardyp:
I merge it like this, as this already adds lots of value imho.

PR #5616, opened 2020-11-05
Fix the git for windows feature detection and initial branch checkout issues in #5601

### LINKED ISSUE #5601: Git step on Windows does not checkout branch on initial clone
I am currently running Buildbot master 2.8.1-dev721 (89f092e2c1891b28ca84d0acc21a9068dd16a0ad).

While setting up a new Builder I noticed that the Git step did not checkout the specified branch during the initial clone; it was properly checked out in subsequent builds on the same worker though.

I think there might be two related issues there.

First my windows git client returns `git version 2.27.0.windows.1`  for `--version`.
`pkg_resources.parse_version()` apparently considers `2.27.0.windows.1` to be lower than any of the feature test versions.
Specifically `supportsBranch` stays `False` which is why `git clone` is run without the `--branch` option (even though it should be supported).

Secondly `switchToBranch` defaults to `False` and is only set to `True` when `supportsBranch` is `True` but not usable in context.
Hence `git clone` is not followed by a dedicated `git checkout` either.

ngladitz:
Fixes #5601

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -362,14 +362,14 @@ class Git(Source, GitStepMixin):
         """Retry if clone failed"""
 
         command = ['clone']
-        switchToBranch = False
+        switchToBranch = self.branch != 'HEAD'
         if self.supportsBranch and self.branch != 'HEAD':
             if self.branch.startswith('refs/'):
                 # we can't choose this branch from 'git clone' directly; we
                 # must do so after the clone
-                switchToBranch = True
                 command += ['--no-checkout']
             else:
+                switchToBranch = False
                 command += ['--branch', self.branch]
         if shallowClone:
             command += ['--depth', str(int(shallowClone))]
--- a/master/buildbot/util/git.py
+++ b/master/buildbot/util/git.py
@@ -75,14 +75,11 @@ class GitMixin:
         self.supportsSshPrivateKeyAsConfigOption = False
 
     def parseGitFeatures(self, version_stdout):
-
-        if 'git' not in version_stdout:
+        match = re.match(r"^git version (\d+(\.\d+)*)", version_stdout)
+        if not match:
             return
 
-        try:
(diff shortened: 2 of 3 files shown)


master/buildbot/steps/source/git.py
```diff
@@ -362,14 +362,14 @@ def _clone(self, shallowClone):
         """Retry if clone failed"""
 
         command = ['clone']
-        switchToBranch = False
+        switchToBranch = True
```
senyai, line 365: Shouldn't be `True`, when `self.supportsBranch` is `False`. Wouldn't it be right to set it to `self.supportsBranch`?

master/buildbot/util/git.py
```diff
@@ -75,14 +75,11 @@ def setupGit(self, logname=None):
         self.supportsSshPrivateKeyAsConfigOption = False
 
     def parseGitFeatures(self, version_stdout):
-
-        if 'git' not in version_stdout:
+        match = re.match("^git version (\\d+(\\.\\d+)*)", version_stdout)
```
senyai, line 78: raw string would be better: `r"^git version (\d+(\.\d+)*)"`

master/buildbot/steps/source/git.py
ngladitz: `switchToBranch` must be `True` when `self.supportsBranch` is `False`.
Otherwise no post-clone branch switching happens when `supoortsBranch` is `False`; which is one of the issues I was having.

master/buildbot/util/git.py
ngladitz: thanks, I've updated it accordingly

master/buildbot/steps/source/git.py
ngladitz: ... though `switchToBranch = True` is redundant for 'HEAD' which should be fixed now

tardyp:
thx!

PR #5619, opened 2020-11-14
Fix tests with Python 3.9

yan12125:
Ref: https://github.com/pypa/setuptools/issues/2086

## Contributor Checklist:

(I'm not sure if I should state Python 3.9 support in this pull request)


tardyp:
Do we really need those imports. Rather than shutting up the warning, I would prefer to fix the root cause.

yan12125:
The root cause is outside buildbot - setuptools imports the deprecated package. Or maybe avoiding importing setuptools from buildbot_pkg?

tardyp:
sorry about misunderstanding. this was very clear.

yan12125:
No problem and thanks for the review!

PR #5627, opened 2020-11-17
ci: Remove travis configuration

p12tic:
Travis has [added](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) a hard cap of around 1000 build minutes on open-source repositories per month. A single build takes around 20 total build minutes on Travis, so we have a limit of around 50 builds per month. This is a tiny amount. Each PR corresponds to around 4 builds on average (1 build after the PR has been merged, 1 for initial attempt and 2 for subsequent revisions). So if we continued using Travis CI free plan we would have the CI for only around 12 PRs per month.


mojca:
From my understanding it's not even 1000 minutes per month, but 1000 minutes in total, and that's it. After you run out of those, you can write them a friendly email to request more.

You can in principle add Azure or GitHub actions if you want a replacement, but removing Travis support totally makes sense.

tardyp:
I kept travis because it was much more performant with realdb tests.

If we do that we should uncomment the relevent lines in bbtravis in order to reactivate those tests

p12tic:
Addressed in a better way in #5642.

PR #5630, opened 2020-11-19
OpenStackLatentWorker: Renderable fixes

cmouse:
#5041  Contributor Checklist:

### DIFF
--- a/master/buildbot/worker/openstack.py
+++ b/master/buildbot/worker/openstack.py
@@ -237,8 +237,19 @@ class OpenStackLatentWorker(CompatibleLatentWorkerMixin,
         return image_uuid
 
     @defer.inlineCallbacks
+    def _getFlavor(self, build):
+        flavor_uuid = yield build.render(self.flavor)
+        # check if we got name instead of uuid
+        for flavor in self.novaclient.flavors.list():
+            if flavor.name == flavor_uuid:
+                flavor_uuid = flavor.id
+        return flavor_uuid
+
+    @defer.inlineCallbacks
     def renderWorkerProps(self, build):
         image = yield self._getImage(build)
+        flavor = yield self._getFlavor(build)
+
         if self.block_devices is not None:
             block_devices = []
             for bd in self.block_devices:
@@ -246,20 +257,20 @@ class OpenStackLatentWorker(CompatibleLatentWorkerMixin,
                 block_devices.append(rendered_block_device)
         else:
             block_devices = None
-        return (image, block_devices)
+        return (image, flavor, block_devices)
 
     @defer.inlineCallbacks
     def start_instance(self, build):
         if self.instance is not None:
             raise ValueError('instance active')
 
-        image, block_devices = yield self.renderWorkerPropsOnStart(build)
-        res = yield threads.deferToThread(self._start_instance, image,
+        image, flavor, block_devices = yield self.renderWorkerPropsOnStart(build)
+        res = yield threads.deferToThread(self._start_instance, image, flavor,
                                           block_devices)
         return res
 
-    def _start_instance(self, image_uuid, block_devices):
-        boot_args = [self.workername, image_uuid, self.flavor]
+    def _start_instance(self, image_uuid, flavor_uuid, block_devices):
+        boot_args = [self.workername, image_uuid, flavor_uuid]
         boot_kwargs = dict(
             meta=self.meta,
             block_device_mapping_v2=block_devices,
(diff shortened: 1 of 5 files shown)


master/buildbot/worker/openstack.py
```diff
@@ -231,39 +234,58 @@ def _getImage(self, build):
...
+            if hasattr(self.novaclient, "glance"):
+                image_uuid = image(self.novaclient.glance.list())
+            else:
+                image_uuid = image(self.novaclient.images.list())
         else:
             image_uuid = yield build.render(image)
         return image_uuid
 
+    @defer.inlineCallbacks
+    def _getFlavor(self, build):
```
p12tic, line 247: Since this is private, let's name it as `_get_flavor`.

master/docs/manual/configuration/workers-openstack.rst
```diff
@@ -52,6 +52,8 @@ These are the same details set in either environment variables or passed as opti
 
 ``flavor``
     The flavor ID to use for the instance.
+    A callable may instead be passed.
+    It will be passed the list of available flavors and must return the flavor to use.
```
p12tic, line 56: It's unclear what type of the flavor is.

`It will be passed the list of available flavors as a list of strings and must return the flavor to use as string.`

master/buildbot/worker/openstack.py
cmouse: This is not consistent with the rest of the class, I can rename it but it will be different from others.

master/buildbot/worker/openstack.py
```diff
@@ -207,7 +207,10 @@ def _determineVolumeSize(self, source_type, source_uuid):
         if source_type == 'image':
             # The size returned for an image is in bytes. Round up to the next
             # integer GiB.
-            image = nova.images.get(source_uuid)
+            if hasattr(nova, "glance"):
+                image = nova.glance.get(source_uuid)
+            else:
```
seankelly, line 212: Probably safe to delete this. Nova removed this part a long time ago (novaclient 8.0 I think?)

master/buildbot/worker/openstack.py
```diff
@@ -231,39 +234,58 @@ def _getImage(self, build):
...
         else:
             image_uuid = yield build.render(image)
         return image_uuid
 
+    @defer.inlineCallbacks
+    def _getFlavor(self, build):
+        # If flavor is callable, then pass it the list of flavors. The
+        # function should return the flavor's UUID to use.
+        flavor = self.flavor
+        if callable(flavor):
```
seankelly, line 251: I know I used `callable` for the `images` method but I think only using a renderer fits better with the rest of Buildbot.

master/buildbot/worker/openstack.py
seankelly: My recollection is Buildbot follows Twisted's style and that would be `_getFlavor`.

master/buildbot/worker/openstack.py
cmouse: it's actually quite useful, because otherwise you cannot easily resolve `m1.medium` into UUID.

cmouse: Example code:
```.py
# Used by openstack client - return matching flavor uuid
def os_get_flavor(flavors):
    want_flavor = util.Interpolate('%(prop:os_flavos)s')
    for flavor in flavors:
        if flavor.name == want_flavor:
            return flavor.id
    return want_flavor

# Used by openstack client - return matching image uuid
def os_get_image(images):
    want_image = util.Interpolate('%(prop:os_image)s')
    for image in images:
        if image.name == want_image:
            return image.id
    return want_image
```

master/buildbot/worker/openstack.py
cmouse: I feel bit bad about removing it. I know it's ages ago, but then again, it's very small tweak to maintain compability with ancient openstack.

master/buildbot/worker/openstack.py
cmouse: Actually, maybe we could just move this code inside the worker? Would that be ok?

seankelly: I would definitely prefer to be able to use the name instead of the UUID so I think it belongs in Buildbot so others can also benefit.

master/buildbot/worker/openstack.py
p12tic: We've changed this recently, see http://docs.buildbot.net/latest/developer/style.html

cmouse: @p12tic so what should be done? Fix the entire file, just this instance or?

p12tic (approved):

p12tic:
I'll wait for @seankelly approval as there were further comments.

master/buildbot/worker/openstack.py
p12tic: I think we can use the new approach on the newly added code only. This does introduce inconsistency, but as the time goes, more and more of Buildbot will migrate and this will become less of an issue.

seankelly (approved):

PR #5632, opened 2020-11-20
process/build: Show workername in worker prepare step

rpurdie:
It's useful for some workflows to see which worker was used in the build summary
screen rather than having to click further into the UI. The easiest way to do this
is to show the worker name in the worker_preparation step description.


tardyp:
looks good, but we need to update a bunch of tests, which verify the output of this state string

tardyp:
ex: https://buildbot.buildbot.net/#/builders/59/builds/1428

rpurdie:
Sorry, I've updated the pull request with a fix for the test that failed

PR #5633, opened 2020-11-20
fix: use `git fetch -f -t` to clobber existing tags instead of failing

terencehonles:
I believe this change should be acceptable, but before writing any tests or updating further I wanted to understand if this makes sense.

I'm not sure why `-t` (`--tags`) was originally provided since I believe `self.branch`, if a tag, would have worked as expected. However, when using `--tags` without `-f` (`--force`) I am getting a build error. I nuked the directory manually, and now see there are two retry options `retryFetch` and `clobberOnFailure`. The latter would do what I did manually and this **may** be the only reason that flag exists.

I'm not sure if a CI maintainer should have to run into this issue, and then fix it. So I would either suggest the `clobberOnFailure` default to `True` or that the `-f` flag be passed to `fetch`. I'm not sure if there is a good reason / likely case where a user would use `fetch` to detect tags changing so I don't think there is a strong reason not to accept this change.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/fix-git-fetch-breaking-on-changed-tags.bugfix
+++ b/master/buildbot/newsfragments/fix-git-fetch-breaking-on-changed-tags.bugfix
@@ -1,0 +1,1 @@
+Fixed an issue where ``git fetch`` would break on tag changes by adding the ``-f`` option. This could previously be handled by manually specifying ``clobberOnFailure``, but as that is rather heavy handed and off by default, this new default functionality will keep Buildbot in sync with the repository it is fetching from.
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -312,7 +312,7 @@ class Git(Source, GitStepMixin):
                 fetch_required = False
 
         if fetch_required:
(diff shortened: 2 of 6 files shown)


p12tic:
Looks good in principle. Now we need to update tests and write a newsfragment.

terencehonles:
> Looks good in principle. Now tests need to be updated and newsfragment needs to be written.

@p12tic done, It looked like Travis was stuck in "queued". I looked over the tests and I'm not sure I need to add additional tests since this is just changing the default behavior so I just added the news fragment and re-pushed to hopefully kick travis (it might need an authorized user's attention though since I don't know if that stalled(?) job will block this next build).

p12tic:
We can ignore the Travis test runner for the moment. Travis won't dedicate resources to open source repositories anymore. We'll turn this test runner off soon.

You can easily run tests locally, see http://docs.buildbot.net/latest/developer/quickstart.html#create-a-buildbot-python-environment . If tests pass on your machine there's high chance that they'll pass on CI. You probably don't need to add new tests, only update existing ones, which now fail because additional command line option is present.

p12tic (approved):
Thanks!

PR #5635, opened 2020-11-21
Rename getFakeService() of the fake HTTP service to getService()

p12tic:
Having `buildbot.test.util.fakehttpclientservice.HttpClientService.getFakeService()` named differently than `getService()` makes it easy to create non-working HTTP tests which don't actually check whether the requests have been executed. This PR fixes on such case.

After investigation, it does not seem that there's a reason for the difference. `super()` works as expected within classmethods, so it does not seem there's a way for the user of the API to shoot one's foot. This PR thus renames `getFakeService` to just `getService`.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/httpclientservice.py
+++ b/master/buildbot/test/fake/httpclientservice.py
@@ -27,6 +27,7 @@ from buildbot.util import service
 from buildbot.util import toJson
 from buildbot.util import unicode2bytes
 from buildbot.util.logger import Logger
+from buildbot.warnings import warn_deprecated
 
 log = Logger()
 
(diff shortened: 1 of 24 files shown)


master/buildbot/test/fake/httpclientservice.py
```diff
@@ -78,20 +79,23 @@ def updateHeaders(self, headers):
 
     @classmethod
     def getFakeService(cls, master, case, *args, **kwargs):
-        ret = cls.getService(master, *args, **kwargs)
+        warn_deprecated('2.9.0', 'getFakeService() has been deprecated, use getService()')
+        return cls.getService(master, case, *args, **kwargs)
 
+    @classmethod
+    @defer.inlineCallbacks
+    def getService(cls, master, case, *args, **kwargs):
```
tardyp, line 87: I think the difference is that getFakeService does not have the same signature, as it has the case argument.

tardyp (approved):

master/buildbot/test/fake/httpclientservice.py
p12tic: Yes. On the other hand, the caller will need to know the initialization arguments of the service class one wants to instantiate/retrieve, so the real signature of each `getService` is different anyway.

PR #5639, opened 2020-11-23
Libvirt refactor

PeterSurda:
# This is an attempt to solicit feedback. It's not ready for merge but I want to know how much work is left to do before it is.

## The rationale for this refactor are my annoyances with the libvirt worker:
- all the connections are established **before** the master can initialize. If there is an issue with just one out of huge amount of workers that you may have defined, master will fail to start
- one connection per worker is opened even for workers that share the same host. If you have a huge number of workers, this will open a huge amount of sockets. I was running out of filehandles during master reconfigure.
- even if there are no problems with connections, they are done **sequentially**. If you have a huge number of workers, it takes a long time, and the startup may timeout anyway
- if a connection fails during the runtime of master, the worker becomes permanently unavailable, until you restart the master

## Background:
I have a bunch of servers for workers. The servers use WoL to turn on when a new worker is needed so I don't have to keep them on 24/7. In order to utilise the hardware efficiently, any server can run any number of any type of worker as long as it has capacity (I use cloud-image for automation both for servers and the workers, and distribute the images through NFS, and `MasterLock` for capacity restrictions). This needs a huge number of definitions of workers (which is done through a function in `master.cfg` that generates them).

## Proposed solution in this PR:
- connection attempts are started **after** buildbot initialization finishes
- connection attempts are done in **separate threads** as not to block
- will reconnect automatically on failures
- the worker definition now uses the `uri` argument instead of `connection`
- workers on the same uri share a connection
- ~locks~ queues are used to make sure a connection is only used for one operation at a time
- I would appreciate if someone could update the tests

PS. based on my experience with other projects, the reason why libvirt connections aren't thread safe is that they are basically stream sockets. Pure stream sockets aren't thread safe. The same problem you'll have with anything using TCP connections, unless the library provides an interface to work around this. Presumably libvirt doesn't.

## Contributor Checklist:


master/buildbot/worker/libvirt.py
```diff
@@ -33,175 +34,112 @@
...
-
-    def execute(self, cb, *args, **kwargs):
-        kickstart_processing = not self.queue
-        d = defer.Deferred()
-        self.queue.append((d, cb, args, kwargs))
-        if kickstart_processing:
-            self._process()
-        return d
-
-    def executeInThread(self, cb, *args, **kwargs):
```
tardyp, line 89: I wasn't aware of this code. This is not very useful, If we want serialisaion of calls to libvirt, because it is not threadsafe, the correct solution is to use a threadpool of 1 thread, then twisted will de-facto serialize it.

https://twistedmatrix.com/documents/current/api/twisted.python.threadpool.ThreadPool.html

master/buildbot/worker/libvirt.py
```diff
@@ -33,175 +34,112 @@
...
+
+    def closeConnect(self):
+        with self.lock:
+            conn = self.conn
+            self.conn = None
+            self.connecting = False
+        conn.close()
+
+
+class ConnectionPool():
```
tardyp, line 78: I would use threadpool api for this.

master/buildbot/worker/libvirt.py
```diff
@@ -281,33 +222,31 @@ def stop_instance(self, fast=False):
...
-
-        domain = self.domain
-        self.domain = None
+            return None
 
         if self.graceful_shutdown and not fast:
             log.msg("Graceful shutdown chosen for {}".format(self.workername))
-            d = domain.shutdown()
-            d.addErrback(_destroy_domain, domain)
+            with self.getLock():
```
tardyp, line 234: you should look at other latent worker for idea of implementation.

Idea for blocking code, is to use thread pool, and to call threaded method.

I would associate one connection per thread so that you know that it will not be shared.

Then you just have to build a _start_instance_thd() method, which is alone in its thread and you don't have to 
manage ressource lock.
If your connection is bound to a thread it is then not shared apart from threaded calls
The threaded call is effectively a barrier/lock

tardyp (changes requested):
Hello,
Indeed, this is a good idea to get some feedback through PR, thanks.

I realize this code do not handle correctly the shared ressource.

I advise you to look at other recently worked latent worker (kubernetes/openstack) for best practices

PeterSurda:
@tardyp Thank you for your feedback. The openstack worker, for example, makes much more sense to me than the original libvirt worker and is more similar to how I handle establishing a connection in my PR.

I'm not sure how to tell the thread pool to do a 1-1 pairing of threads and uri in twisted though.

PeterSurda:
@tardyp or is what you're saying, I should have a separate pool for each uri, as opposed to one pool with per-uri threads? I think I see how to do that.

tardyp:
I think you can use thread local data:

https://docs.python.org/3/library/threading.html#thread-local-data


Pierre


Le lun. 23 nov. 2020 à 17:43, Peter Šurda <notifications@github.com> a
écrit :

tardyp:
openstack is http, and can easily share its connection per API.
If libvirt is socket based, then indeed, we need to take care of it.

Pierre

PeterSurda:
> I think you can use thread local data:

I am familiar with this. I'm just not sure how to influence `ThreadPool`'s decisions about which thread to use. Should I make a new class that's a child of `ThreadPool` that will have this functionality? It doesn't look like twisted has an existing API to do what I want.

PeterSurda:
> openstack is http, and can easily share its connection per API. If libvirt is socket based, then indeed, we need to take care of it.

You could theoretically use a new connection for every libvirt method call, but I expect this to have a noticeable performance penalty.

p12tic:
@PeterSurda 

Regarding "The servers use WoL to turn on when a new worker is needed so I don't have to keep them on 24/7."

There is partial native support for this feature in Buildbot. It's called `machines` which abstract any kind of hardware that workers could be running on. I have it working on my setup, but haven't pushed the functionality to upstream Buildbot because it requires making sure the API is properly future-proof which is a lot of work. But given that we now know that there will be external users it makes sense for me to go back and finish the feature. This is of course after I complete the github comment reporter work :-)

p12tic:
Regarding the libvirt worker the changes look good to me in principle, but I haven't looked into the details yet.

PeterSurda:
> There is partial native support for this feature in Buildbot. It's called machines which abstract any kind of hardware that workers could be running on. I have it working on my setup,

That's great. I would need to look at it in more detail to find out how it can help me. If it can reduce the number of necessary worker definitions, that would unclutter the web interface somewhat, but isn't a big deal because I define them calling a function in a loop. For all the other things I need, the libvirt changes in this PR are sufficient for me. So it's not a high priority to finish. I'm assigning some other developer to finish this PR so that it can be merged eventually.

One feature I'd like to add, but perhaps in a separate PR, is to customise what server is selected. Now it looks like random, and due to the way I rewrote the `canStartBuild`, it always ends up waking up all the servers when a new job is created. So I'd like there to be a configuration option to, for example, use a server that is already up as long as it has capacity, or loadbalance, or some other selection algorithm. Another one would be support for ephemeral worker passwords. But compared to the problems I had before, these are just minor annoyances.

PeterSurda:
So here's an update. I implemented my own threadpool because I couldn't figure out how to add the functionality I needed with twisted. It now automatically creates a thread per each URI, and the thread has a processing queue for starting and stopping instances. I disabled callbacks for `start_instance` and `stop_instance`, since both may take a long time to return anything, I don't see the point. I also added a `nextWorker` implementation to prioritise servers that are connected (it doesn't seem to work at the moment though).

Still needs work on fixing, it's just so that you know the direction I'm taking now.

p12tic:
@PeterSurda:

> I implemented my own threadpool because I couldn't figure out how to add the functionality I needed with twisted.

I _think_ Pierre was talking about a separate Twisted thread pool per connection. The thread pool would only have a single thread. This way there's an implicit queue of actions to be done in the thread pool and everything is serialized. Would this match your requirements or you had other problems with Twisted?

PeterSurda:
@p12tic That was one of the problems. The second one was that I wasn't able to find a way to use a custom queue. The way I do it now, as you can see, is a blocking get from the queue, but if there is no connection to the libvirt server, the element is inserted back into the front of the queue. Maybe it can be done by referring to private attributes but I don't like that approach, among other things it may break randomly if twisted's internals change.

tardyp:
So I didn't mean to use one threadpool per connection. I really meant to use one threadpool for the whole class, and then let the thread do the serialization.

The threadpool is actually a queue.
When you say
```
threadPool.callInThread(self.start_instance_thd)
```
It will actually put the method call in a queue and wake up the threads. The first thread to wake up will look at the queue, and take the first element in the queue.

Then you just have to make sure you start_instance_thd do whatever is needed to start the instance in a single call. During all that time, you are sure that you are alone in your thread without your connection handle (provided the connection handle is thread local). No need to worry more about synchronization.

I don't think there is a need to do a queue of action for a single worker. the latent worker mechanism is already ensuring that you don't call stop worker twice in a row for the same worker. If it does not, I think that serialization part should be done at that stage, because it is the same problem for all worker drivers.

For the serialization/synchronization *between* workers, then the queue of the threadPool should so the job.

>  the element is inserted back into the front of the queue

Which queue are you talking about? normally, a queue is always FIFO. If its LIFO than its called a stack.

PeterSurda:
> So I didn't mean to use one threadpool per connection. I really meant to use one threadpool for the whole class, and then let the thread do the serialization.

I haven't figured out how to do a per-URI serialization then, even after looking at twisted's sources.

> Then you just have to make sure you start_instance_thd do whatever is needed to start the instance in a single call. During all that time, you are sure that you are alone in your thread without your connection handle (provided the connection handle is thread local). No need to worry more about synchronization.

I still don't understand how to prevent multiple threads being assigned to the same URI. Maybe I didn't explain sufficiently that now we're using a single connection per URI, rather than per worker.

> I don't think there is a need to do a queue of action for a single worker. the latent worker mechanism is already ensuring that you don't call stop worker twice in a row for the same worker. If it does not, I think that serialization part should be done at that stage, because it is the same problem for all worker drivers.

The issue is, at least in my case, which I don't think is unique, multiple workers sharing the same URI (one physical server can run multiple VMs). In my case there are several dozens of VM per server defined. The old mechanism required a connection per VM, i.e. several dozens of connections, which is a total overkill and causing problems. Now we only use one connection per server, but we need to prevent two workers from using the connection at the same time.

> For the serialization/synchronization _between_ workers, then the queue of the threadPool should so the job.

But I don't see a way for the pool to do a 1-1 assignment between threads and URIs. I found nothing in the API of twisted, nor after reviewed the source of twisted. It looks to me it would require much more complicated code than if I do a specailised implementation.

> Which queue are you talking about? normally, a queue is always FIFO. If its LIFO than its called a stack.

The queue in my PR acts as FIFO if buildbot is queueing a job, but as a LIFO while it's trying to establish a connection. This allows a smooth reconnect, and the connection to remain off as long as there is nothing in the queue.

PeterSurda:
Added support for ephemeral passwords through libvirt's metadata functionality. Requires a libvirt hook to actually work, will add an example one once I'm done cleaning it up. The hook supports cloud-init and I'm working on modifying the config file directly through mounting the image for VMs that don't support cloud-init.

p12tic:
I've extracted the queue into a separate, reusable component as the basic concept is not related to libvirt. The rest of code has been rebased on top of it.

PeterSurda:
@p12tic thanks

PeterSurda:
@p12tic maybe I can put the queue into a separate PR, then this one will become smaller?

p12tic:
I have some reservations about the current PR and need to investigate whether they are true and how to address them. This may change how the queue works. So for now let's keep everything in single PR.

PeterSurda:
@p12tic ok let me know what's wrong

p12tic:
I've been looking into this PR lately and the primary issue is that `start_instance` must only finish when all startup actions are done, otherwise we introduce a race condition. I'm planning to bring back some of the original design where there were wrappers to execute method calls in a thread.

PeterSurda:
@p12tic That makes sense. The question I have though is that for some of my servers, it may take a couple of minutes for a to become available after the ssh connection is initiated. The motherboard doesn't support suspend, so it has to perform a full boot after being woken up, and since I use cloud-init for configuration, it takes a while to update and install packages. In between, the server is kind of in a limbo, there is no clear way to say if the VM can be started or not. I just want the libvirt plugin to be able to handle such a situation sensibly, and I wasn't sure if using `deferred` doesn't simply result in buildbot being frozen while waiting for the connection.

One option, that I originally tired, was to put the server in a short term quarantine and move to another one. But then we should also be able to avoid a situation where this results in **all** the servers being woken up in sequence just because ssh doesn't immediately connect to them.

The current code mostly handles such a situation, but sometimes it erroneously thinks the VM was started and it's just waiting for the worker to report back. This then takes an hour to timeout while jobs are stuck. I now work around it by giving priority to those servers that do support suspend (using `nextWorker`), and there ssh connects without a retry, so the problem only appears if there are too many jobs pending.

Also, perhaps I should split this into more than one PR for easier incremental merging, and keep the old `connection` argument to keep backwards compatibility and make it optional to use either `connection` and `uri`. I don't want to unnecessarily break old configs other people may have.

p12tic:
>  I wasn't sure if using deferred doesn't simply result in buildbot being frozen while waiting for the connection.

I can assure you that Buildbot won't be frozen. It will happily process all other events and will continue code only after the connection is established or timeout occurs. The only part which would be effectively frozen would be the worker startup function, but that's fine.

> One option, that I originally tired, was to put the server in a short term quarantine and move to another one. But then we should also be able to avoid a situation where this results in all the servers being woken up in sequence just because ssh doesn't immediately connect to them.

Right, we need some way to solve this kind of long startup problem. Short term quarantine looks like a sensible workaround for now.

> Also, perhaps I should split this into more than one PR for easier incremental merging, and keep the old connection argument to keep backwards compatibility and make it optional to use either connection and uri. I don't want to unnecessarily break old configs other people may have.

Don't worry, I have already written some code that solves the compatibility problem.

One thing that I will ask you to do later is to test the final PRs on your Buildbot environment, as the testing that I'm doing will be pretty simple.

PeterSurda:
Ok, I only have two more issues:

- the alphabetical sorting in `nextWorker` (well, technically in `__lt__`) is reversed. Do you want me to fix it or will you do it?
- I also added support for ephemeral passwords, using cloud-init in the libvirt qemu hook. It will use cloud-localds to create an ephemeral CD image from a base file parsed through a jinja template (to replace the master hostname, username and password). For systems that don't support cloud-init, you can create a boot service that will treat the cloud init config as whatever it wants, for example for a mac VM I treat it like a shell script and execute it on boot. I think I should put that into a separate PR. It doesn't depend on the code in this PR. That PR can probably be merged faster as it causes smaller changes.

p12tic:
@PeterSurda 

I pushed a cleaned-up version of the code. I've removed next worker handling for now and also the stuff related to quarantines. These will come a bit later. Could you maybe test this PR in your setup? I've only checked with a single local libvirt VM.

> I also added support for ephemeral passwords,

These would be great to have in a separate PR.

p12tic:
Based on some additional testing this PR works. So I don't think it's worth holding it up, we can add the missing functionality later.

PeterSurda:
I'm sorry but I don't think I'll be able to test it in the near future. Don't wait for my feedback, in case I have problems with it later, I'll send my comments separately.

PR #5641, opened 2020-11-23
steps/trigger: Add status counts to triggered build summaries

rpurdie:
Its useful for the build summary to state how may of the builds in a trigger
have which statuses. Add callbacks to the triggered builds so that counts
of the results can be shown to the user in the summary.

In order to be more readable, statusToString gains a count parameter so
that plural values can be shown more correctly.

The trigger unit test is updated to contain mixed success/failure counts
which also tests the plural status code.

### DIFF
--- a/master/buildbot/steps/trigger.py
+++ b/master/buildbot/steps/trigger.py
@@ -25,6 +25,7 @@ from buildbot.process.buildstep import SUCCESS
 from buildbot.process.buildstep import BuildStep
 from buildbot.process.properties import Properties
 from buildbot.process.properties import Property
+from buildbot.process.results import ALL_RESULTS
 from buildbot.process.results import statusToString
 from buildbot.process.results import worst_status
 
@@ -107,6 +108,7 @@ class Trigger(BuildStep):
         self.brids = []
         self.triggeredNames = None
         self.waitForFinishDeferred = None
+        self._result_list = []
         super().__init__(**kwargs)
 
     def interrupt(self, reason):
@@ -236,6 +238,20 @@ class Trigger(BuildStep):
                                           url)
 
     @defer.inlineCallbacks
+    def _add_results(self, brid):
+        @defer.inlineCallbacks
+        def _is_buildrequest_complete(brid):
+            buildrequest = yield self.master.db.buildrequests.getBuildRequest(brid)
+            return buildrequest['complete']
+
+        event = ('buildrequests', str(brid), 'complete')
+        yield self.master.mq.waitUntilEvent(event, lambda: _is_buildrequest_complete(brid))
+        builds = yield self.master.db.builds.getBuilds(buildrequestid=brid)
+        for build in builds:
+            self._result_list.append(build["results"])
+        self.updateSummary()
+
+    @defer.inlineCallbacks
     def run(self):
         schedulers_and_props = yield self.getSchedulersAndProperties()
 
@@ -297,6 +313,9 @@ class Trigger(BuildStep):
                 # the beginning
                 url = self.master.status.getURLForBuildrequest(brid)
                 yield self.addURL("{} #{}".format(sch.name, brid), url)
+                # No yield since we let this happen as the builds complete
+                self._add_results(brid)
+
             dl.append(resultsDeferred)
             triggeredNames.append(sch.name)
             if self.ended:
@@ -330,4 +349,11 @@ class Trigger(BuildStep):
     def getCurrentSummary(self):
         if not self.triggeredNames:
             return {'step': 'running'}
-        return {'step': 'triggered {}'.format(', '.join(self.triggeredNames))}
+        summary = ""
+        if self._result_list:
+            for status in ALL_RESULTS:
+                count = self._result_list.count(status)
+                if count:
+                    summary = summary + ", {} {}".format(self._result_list.count(status),
+                                                     statusToString(status, count))
+        return {'step': 'triggered {}{}'.format(', '.join(self.triggeredNames), summary)}
--- a/master/buildbot/test/integration/test_trigger.py
+++ b/master/buildbot/test/integration/test_trigger.py
@@ -28,7 +28,7 @@ expectedOutputRegex = \
     \*\*\* STEP worker_preparation \*\*\* ==> worker ready \(success\)
     \*\*\* STEP shell \*\*\* ==> 'echo hello' \(success\)
         log:stdio \({loglines}\)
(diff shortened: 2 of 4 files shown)


master/buildbot/steps/trigger.py
```diff
@@ -107,6 +108,7 @@ def hasRenderable(l):
         self.brids = []
         self.triggeredNames = None
         self.waitForFinishDeferred = None
+        self.reslist = []
```
p12tic, line 111: I think something like `_result_list` would be more readable.

master/buildbot/steps/trigger.py
```diff
@@ -330,4 +343,12 @@ def getResultSummary(self):
...
         if not self.triggeredNames:
             return {'step': 'running'}
-        return {'step': 'triggered {}'.format(', '.join(self.triggeredNames))}
+        summary = ""
+        log.err("trigger current summary %s" % str(self.reslist))
+        if self.reslist:
+            for status in ALL_RESULTS:
+                count = self.reslist.count(status)
+                if count:
+                    summary = summary + ", %s %s" % (self.reslist.count(status),
```
p12tic, line 352: Let's use python3 `.format()` here.

master/buildbot/steps/trigger.py
```diff
@@ -330,4 +343,12 @@ def getResultSummary(self):
     def getCurrentSummary(self):
         if not self.triggeredNames:
             return {'step': 'running'}
-        return {'step': 'triggered {}'.format(', '.join(self.triggeredNames))}
+        summary = ""
+        log.err("trigger current summary %s" % str(self.reslist))
```
p12tic, line 347: Debug code?

master/buildbot/test/integration/test_trigger.py
```diff
@@ -100,4 +111,18 @@ def masterConfig():
         BuilderConfig(name="build",
                       workernames=["local1"],
                       factory=f2)]
+    if addFailure:
+        f3 = BuildFactory()
+        f3.addStep(steps.ShellCommand(command='false'))
+        c['builders'].append(
```
p12tic, line 117: I think this could be reformatted to fill 100 lines better.

p12tic (changes requested):
Looks good in principle, thanks! I had a couple of nitpicks.

p12tic:
@tardyp What do you think about `statusToString` API change?

master/buildbot/steps/trigger.py
```diff
@@ -235,6 +237,13 @@ def addBuildUrls(self, rclist):
                                                               builderNames[builderid], num),
                                           url)
 
+    @defer.inlineCallbacks
+    def addResults(self, br, brid):
```
p12tic, line 241: This is private function, should be named `_add_results`.

master/buildbot/steps/trigger.py
```diff
@@ -297,6 +306,10 @@ def run(self):
                 # the beginning
                 url = self.master.status.getURLForBuildrequest(brid)
                 yield self.addURL("{} #{}".format(sch.name, brid), url)
+                deferredRes = self.master.mq.waitUntilEvent(('buildrequests', str(brid),
```
p12tic, line 309: I think it would be more readable if this was extracted into a separate `@defer.inlineCallbacks function` which is not awaited via `yield` here (add a comment explaining this). I guess that function could be merged with `addResults()`.

master/buildbot/steps/trigger.py
rpurdie: Makes sense, changed

master/buildbot/steps/trigger.py
rpurdie: Yes, sorry. Been staring at it for too long now!

master/buildbot/steps/trigger.py
rpurdie: I still struggle to get twisted code right so if I'm changing this I'll need to look at it more carefully to double check I understand what you mean. I've made the other tweaks though.

tardyp: You can have a look at basic asyncio tutorial to get a grasp about it.

for example: https://stackabuse.com/python-async-await-tutorial/

twisted has the same concepts but 10 years before, so the API is not as nice.

but basically you can say

```python
@defer.inlineCallbacks
def method(...)
```

is equivalent to

```python
async def method(...)
```
and

```python
     yield self.returningDeferredFunction()
```
is equivalent to
```python
    await self.returningFutureOrCoroutineFunction()
```

There are some subtle differences, but most of the time, this is not relevent.

The key learning is: don't use callbacks, there are barely any case where it is useful.
There is only a slight performance gain, and it can be easier to reason on for hardcore network protocol stuff. But for 99% of buildbot code, we want to switch to inlineCallbacks (the coroutines of twisted)

rpurdie: Thanks! I think I understand the basics but what I was struggling with here was how to have a callback function which didn't block the rest of the code (so no yield) yet used the inline callbacks. I'm sure if I stare at it further I can work it out, I just ran out of time yesterday...

tardyp: ah. Indeed, this is the 1% case where you want to create a callback.

As it is not obvious, I guess we'll just need a simple comment.

I think what @p12tic was suggesting is to just use fire and forget

```
    @defer.inlineCallbacks
    def _wait_for_results(self, brid):
        br, brid = yield self.master.mq.waitUntilEvent(('buildrequests', str(brid), 'complete', is_buildrequest_complete(brid)
        builds = yield self.master.db.builds.getBuilds(buildrequestid=brid)
        for build in builds:
            self._result_list.append(build["results"])
        self.updateSummary()
```
and then in run() method, you just call this method and ignore the returning deferred.

Note, that you will need to implement is_buildrequest_complete, by just calling the data api, in order to avoid race condition between actual completion and event registering

rpurdie: The is_buildrequest_complete(brid) here threw me for a while, it needs to be is_buildrequest_complete, brid :)

tardyp: just edited my previous comment to add the needed lambda:

rpurdie: Thanks, I got there as well and have updated the pull request. Might be a style issue left on where the function is located but hopefully getting there!

tardyp (approved):
looks good!

p12tic (approved):

PR #5643, opened 2020-11-23
docs: Document migration to new style in the build step pages

p12tic:
This concludes the preparation for build step migration to new-style.


p12tic:
@tardyp : Ping just in case you're not watching all PRs.

PR #5644, opened 2020-11-23
steps: Deprecate the MTR step due to lack of tests

p12tic:
The MTR step does not have proper unit tests, so it'll be too time consuming to migrate this step to new style as a test suite needs to be written in addition to the actual migration. The step has been written in 2009 and didn't have substantial changes ever since, so going extra mile ourselves is probably not a good use of resources.


p12tic:
@tardyp : Ping just in case you're not watching all PRs.

PR #5652, opened 2020-11-27
Fix for ldap3.core.exceptions.LDAPInvalidFilterError: malformed filter

vx-i:
Fix ldap3.core.exceptions.LDAPInvalidFilterError if LDAP filter string contains characters that needs to be escaped. For example:

displayName: Lastname, Firstname

master.cfg

# Active Directory user information
userInfoProvider = util.LdapUserInfo(
    uri='ldap://ldap.local:3268',
    bindUser='user@ldap.local',
    bindPw='pass',
    accountBase='OU=Companies,DC=ldap,DC=local',
    groupBase='OU=Companies,DC=ldap,DC=local',
    accountPattern='(&(objectClass=user)(sAMAccountName=%(username)s))',
    accountFullName='displayName',
    accountEmail='mail',
    groupMemberPattern='(&(objectClass=group)(member=%(dn)s))',
    groupName='Name',
    avatarPattern='(&(objectClass=user)(mail=%(email)s))',
    avatarData='thumbnailPhoto',
)

## Contributor Checklist:

* [x] I have updated the appropriate documentation (not applicable)

### DIFF
--- a/master/buildbot/test/unit/www/test_ldapuserinfo.py
+++ b/master/buildbot/test/unit/www/test_ldapuserinfo.py
@@ -14,13 +14,14 @@
 #
 # Copyright Buildbot Team Members
 
-import sys
 import types
 
 import mock
 
 from twisted.internet import defer
 from twisted.trial import unittest
+
+from buildbot.www import ldapuserinfo
 
 
 def get_config_parameter(p):
@@ -31,9 +32,6 @@ def get_config_parameter(p):
 fake_ldap = types.ModuleType('ldap3')
 fake_ldap.SEARCH_SCOPE_WHOLE_SUBTREE = 2
 fake_ldap.get_config_parameter = get_config_parameter
-
-with mock.patch.dict(sys.modules, {'ldap3': fake_ldap}):
-    from buildbot.www import ldapuserinfo
 
 
 class FakeLdap:
@@ -67,7 +65,7 @@ class CommonTestCase(unittest.TestCase):
 
     def _makeSearchSideEffect(self, attribute_type, ret):
         ret = [[{'dn': i[0], attribute_type: i[1]} for i in r]
-             for r in ret]
+               for r in ret]
         self.userInfoProvider.search.side_effect = ret
 
     def makeSearchSideEffect(self, ret):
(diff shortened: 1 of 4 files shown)


p12tic (changes requested):
Tests need to be updated.

Looks good otherwise.

vx-i:
Good evening all, I think I need some help with tests. Because ldap3 is faked within tests, I get this error:

builtins.AttributeError: module 'ldap3' has no attribute 'utils'

How can I get rid of this?

p12tic:
@vx-i Indeed. Looking into history it seems that the reason for this weird whole-module mocking in tests was that in the past ldap library was C-based, which I assume would have resulted in a hard dependency of that library in tests.

ldap3 is pure-python library, so this is no longer needed.

I think you could just mock get_config_parameter using the usual mock.patch in the test and remove the current whole module patching. Then the tests should work.

Thanks!

vx-i:
@p12tic The tests still failing cause ldap3 is not in the requirements-ci.txt. Is there a reason why it is not in the list?

p12tic:
@vx-i You're right, could you add it to requirements-ci.txt? Thanks.

master/buildbot/test/unit/www/test_ldapuserinfo.py
```diff
@@ -91,7 +89,8 @@ def makeUserInfoProvider(self):
         self.userInfoProvider = ldapuserinfo.LdapUserInfo(
             uri="ldap://uri", bindUser="user", bindPw="pass",
             accountBase="accbase", groupBase="groupbase",
-            accountPattern="accpattern", groupMemberPattern="groupMemberPattern",
+            accountPattern="accpattern",
```
p12tic, line 92: I think this wasn't needed for the fix you're doing was it? Buildbot uses 100 column line length limit so the current code was fine. Could you revert this change here and in a couple of places below?

p12tic (changes requested):
Just a single nitpick, but otherwise looks great. Thank you!

master/buildbot/test/unit/www/test_ldapuserinfo.py
vx-i: Sorry, I didn't know. Fixed.

p12tic (approved):

p12tic:
I've rebased on top of newest master and fixed a couple of nitpicks that I still had.

PR #5673, opened 2020-12-08
fix: fix profile menu wrapping

terencehonles:
This change fixes the user profile name visibility which was not kept in (negated) sync with the avatar and causes the menu to wrap funny.

Changes:
Screenshot from 2020-12-07 19-01-51

"Back" to the following (expanded the menu to show the name is already there):
Screenshot from 2020-12-08 08-52-34

## Contributor Checklist:

### DIFF
--- a/www/base/src/app/common/directives/loginbar/loginbar.tpl.jade
+++ b/www/base/src/app/common/directives/loginbar/loginbar.tpl.jade
@@ -15,9 +15,8 @@ ul.nav.navbar-nav.navbar-right(ng-show="config.auth.name != 'NoAuth'")
 
     li.dropdown(uib-dropdown, ng-hide="anonymous")
         a.dropdown-toggle(uib-dropdown-toggle)
-            .avatar()
-                img(ng-src="avatar?username={{username}}&email={{email}}")
-            span(ng-if="!email")
+            img.avatar(ng-if="config.avatar_methods.length" ng-src="avatar?username={{username}}&email={{email}}")
+            span(ng-if="!config.avatar_methods.length")
                 | {{full_name || username}}
                 b.caret
         ul.dropdown-menu(uib-dropdown-menu)
--- a/www/base/src/styles/styles.less
+++ b/www/base/src/styles/styles.less
@@ -121,19 +121,12 @@ Firefox will insert newlines if the selection go over a no-select span, so we ne
   padding-right: 10px;
 }
 
-.avatar {
-  background-size: 32px 32px;
+img.avatar {
+  background-color: #ccc;
   border-radius: 50%;
-  display: block;
-  margin: -10px;
   height: 40px;
+  margin: -10px;
   width: 40px;
-  overflow: hidden;
-  img {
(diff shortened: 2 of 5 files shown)


www/base/src/app/common/directives/loginbar/loginbar.tpl.jade
```diff
@@ -15,9 +15,8 @@ ul.nav.navbar-nav.navbar-right(ng-show="config.auth.name != 'NoAuth'")
 
     li.dropdown(uib-dropdown, ng-hide="anonymous")
         a.dropdown-toggle(uib-dropdown-toggle)
-            .avatar()
-                img(ng-src="avatar?username={{username}}&email={{email}}")
-            span(ng-if="!email")
+            img.avatar(ng-if="config.avatar_methods.length" ng-src="avatar?username={{username}}&email={{email}}")
+            span(ng-if="!config.avatar_methods.length")
```
terencehonles, line 19: I don't know if there's a better way to show if there's an avatar or not. This previously was looking at if there was an email, but with #5420 the avatar shows more often.

www/guanlecoja-ui/src/module/topbar/topbar.less
```diff
@@ -63,17 +63,10 @@
         padding-left: 0px;
     }
 }
-.avatar {
```
terencehonles, line 66: Not sure why this is declared twice. It gets output twice too.

I'm not sure which would be the better one to remove so I kept them in sync.

www/base/src/styles/styles.less
```diff
@@ -121,19 +121,12 @@ Firefox will insert newlines if the selection go over a no-select span, so we ne
   padding-right: 10px;
 }
 
-.avatar {
-  background-size: 32px 32px;
+img.avatar {
```
terencehonles, line 124: Is `div > img` for IE? I'm not sure why the border radius wasn't placed directly on the image and I'm assuming this is just to support older IE, but I'm assuming I didn't notice any prefixes or they aren't being generated anymore.

p12tic (approved):

p12tic:
Thanks a lot!

www/base/src/app/common/directives/loginbar/loginbar.tpl.jade
p12tic: The current method sounds reasonable.

www/guanlecoja-ui/src/module/topbar/topbar.less
p12tic: I think this is from the time when guanlecoja-ui was a separate project. After it got merged to Buildbot, only minimal cleanup has been done.

www/base/src/styles/styles.less
p12tic: We no longer support IE at all. Edge 13 is the minimum (according to https://github.com/buildbot/buildbot/blob/master/www/base/src/app/app.browserwarning.notranspile.js)

Issue #5669, opened 2020-12-06
db upgrade broken with sqlalchemy.exc.NoSuchTableError: migration_tmp


landryb:
exact same issue as #4978 :
```
upgrading database (sqlite:/state.sqlite)         
Warning: Stopping this process might cause data loss                                                                                                                                                                 
'A serious error has been encountered during the upgrade. The previous upgrade has been likely interrupted. The database has been damaged and automatic recovery is impossible.'
'If you believe this is an error, please submit a bug to the Buildbot project.'                                                                                                                                      
Got fatal Exception on DB                      
Traceback (most recent call last):                                                                                                                                                                                   
  File "/usr/local/lib/python3.8/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)                                                                                                                                                 
  File "/usr/local/lib/python3.8/site-packages/twisted/python/context.py", line 122, in callWithContext                                                                                                              
    return self.currentContext().callWithContext(ctx, func, *args, **kw)                                                                                                                                             
  File "/usr/local/lib/python3.8/site-packages/twisted/python/context.py", line 85, in callWithContext                                                                                                               
    return func(*args,**kw)                                                                               
  File "/usr/local/lib/python3.8/site-packages/buildbot/db/pool.py", line 217, in __thd                                                                                                                              
    log.err(e, 'Got fatal Exception on DB')                                                                                                                                                                          
--- <exception caught here> ---                                                                                                                                                                                      
  File "/usr/local/lib/python3.8/site-packages/buildbot/db/pool.py", line 190, in __thd                   
    rv = callable(arg, *args, **kwargs)                                                                                                                                                                              
  File "/usr/local/lib/python3.8/site-packages/buildbot/db/model.py", line 1089, in thd                                                                                                                              
    upgrade(engine)                                                                                                                                                                                                  
  File "/usr/local/lib/python3.8/site-packages/buildbot/db/model.py", line 1053, in upgrade                                                                                                                          
    schema.runchange(version, change, 1)                                                                                                                                                                             
  File "/usr/local/lib/python3.8/site-packages/migrate/versioning/schema.py", line 93, in runchange       
    change.run(self.engine, step)                                                                         
  File "/usr/local/lib/python3.8/site-packages/migrate/versioning/script/py.py", line 154, in run                                                                                                                    
    script_func(engine)                        
[...]

PR #5680, opened 2020-12-10
require latest sqlalchemy-migrate

tardyp:
waiting for alembic, we should require recent version of sqlalchemy
because 0.11 can corrupt the db

Fix: #5669


p12tic (changes requested):
I think we need a release note.

PR #5695, opened 2020-12-15
2.9: Fix warnings of depretation of buildbot.status even when it's not used

p12tic:
Backport of #5694 to 2.9.x.


tardyp:
Not sure why we still need a 2.9.x branch

p12tic:
The reason for separate branch is that master already contains changes that shouldn't be in 2.9.3.

PR #5708, opened 2020-12-23
websocket: remove setup of webstatus feature of autobahn which is not…

tardyp:
… used

back port for 2.9.x


p12tic (changes requested):
Spelling check fails.

PR #5722, opened 2020-12-30
reporters: Deprecate template file support in message formatters

p12tic:
Passing template paths to message formatters introduces an implicit dependency on the state of the filesystem to the configuration. This is error-prone in configurations such as when multimaster is used. It is easy enough to read a file master.cfg if needed, so Buildbot should only expose a string-based API.

## Contributor Checklist:


p12tic:
cc @tardyp

PR #5729, opened 2021-01-02
Allow ReconfigurablePollingChangeSource in change_hook/poller.

### LINKED ISSUE #5727: change_hook/poller not working for ReconfigurablePollingChangeSource
In [poller.py](https://github.com/buildbot/buildbot/blob/a0e1d8840e8856ead136a1ad6e2021931355af15/master/buildbot/www/hooks/poller.py#L40), the change sources are filtered like this:

```python
        for source in change_svc:
            if not isinstance(source, PollingChangeSource):
                continue
```

This means that any pollers derived from the super-class `ReconfigurablePollingChangeSource` will not be found. Since [new code is supposed to use `ReconfigurablePollingChangeSource`](https://docs.buildbot.net/current/developer/cls-changesources.html?highlight=reconfigurablepollingchangesource#pollingchangesource), the code should probably read:

```python
        for source in change_svc:
            if not isinstance(source, ReconfigurablePollingChangeSource):
                continue
```

stoklund:
New change pollers are supposed to derive from `ReconfigurablePollingChangeSource`
which is now a superclass of the `PollingChangeSource` class used by old code.

Make sure that `/change_hook/poller` supports both kinds of poller classes.

Fixes #5727

### DIFF
--- a/master/buildbot/newsfragments/reconfigurable-poller.bugfix
+++ b/master/buildbot/newsfragments/reconfigurable-poller.bugfix
@@ -1,0 +1,3 @@
+Fixed :issue:`5727` which prevented polling change sources derived from
+:py:class:`~buildbot.changes.base.ReconfigurablePollingChangeSource` from
+working correctly with `/change_hook/poller`.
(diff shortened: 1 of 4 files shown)


p12tic (approved):
Looks great, thank you!

p12tic:
I've squashed the commits into one.

PR #5730, opened 2021-01-02
Use full signal names in scripts

p12tic:
This allows to run search for SIGXXX through the codebase and easily see where it's used. It was not clear e.g. what could send SIGUSR1 to the master.


p12tic:
Turns out that twisted.runProcess uses 'KILL' or 'TERM' as signal names so the downside of breaking this consistency will result high chance of real bugs, rather than the codebase being just harder to understand.

PR #5732, opened 2021-01-02
gerrit: Deduplicate only source-related gerrit events

### LINKED ISSUE #5596: GerritChangeSource marks comment-added events as duplicated
Hi,

We recently upgraded our master from 2.0.0 to 2.8.4.  After the upgrade, all the comment-added gerrit events seem to be classified as duplicated.  This message is being printed when a comment is added (we use gerrit comments to signal CI)

    "gerrit: duplicate change event {} by {}"

Looks like the changes to GerritChangeSourceBase.addChange broke this.  In particular, the stampdict passed to findOrCreateId looks to be missing a few fields for the duplication detection to work correctly.  I suspect that change-merged events have a similar problem, but i didnt check for sure.

As a workaround I reverted the entire gerritchangesource.py to 2.0.0 level, which works for us.

p12tic:
Fixes #5596.

Current GerritChangePoller will deduplicate all gerrit change events and only react to the first event related to a single revision. This makes it impossible to start builds using comments.

The PR limits the deduplication only to gerrit events related to source code changes.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -163,7 +163,7 @@ class GerritChangeSourceBase(base.ChangeSource):
         return func(properties, event)
 
     @defer.inlineCallbacks
-    def addChange(self, chdict):
+    def addChange(self, event_type, chdict):
         stampdict = {
             "branch": chdict["branch"],
             "revision": chdict["revision"],
@@ -177,7 +177,7 @@ class GerritChangeSourceBase(base.ChangeSource):
         stampid, found_existing = yield(
              self.master.db.sourcestamps.findOrCreateId(**stampdict))
 
-        if found_existing:
+        if found_existing and event_type in ("patchset-created", "ref-updated"):
             if self.debug or True:
                 eventstr = "{}/{} -- {}:{}".format(
                     self.gitBaseURL, chdict["project"], chdict["branch"],
@@ -236,7 +236,7 @@ class GerritChangeSourceBase(base.ChangeSource):
                 patchset=event["patchSet"]["number"]
             )
 
-        yield self.addChange({
+        yield self.addChange(event['type'], {
             'author': _gerrit_user_to_author(event_change["owner"]),
             'project': util.bytes2unicode(event_change["project"]),
             'repository': "{}/{}".format(
--- a/master/buildbot/newsfragments/gerrit-deduplicate-only-source-events.bugfix
+++ b/master/buildbot/newsfragments/gerrit-deduplicate-only-source-events.bugfix
@@ -1,0 +1,2 @@
+Fixed a regression in ``GerritChangeSource`` since v2.6.0 that caused only the first event related to a Gerrit change to be reporter as a change to Buildbot (:issue:`5596`).
+Now such deduplication will be applied only to ``patchset-created`` and ``ref-updated`` events.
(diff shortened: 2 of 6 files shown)


p12tic:
@jangmarker Given https://github.com/KDAB/buildbot/commit/7c1272d0b79be5f011024d6501faf66fa87c1ba4 could you please tell us whether this PR would make GerritChangeSource more useful to your use case?

sknebel:
(I work with Jan, answering for him/KDAB) 

Without having tested it, it would improve our situation, since it would allow us to enable a polling change source for `patchset-created` and `event-updated`. :+1:

Ideally the deduplication would also happen for other event types, so we could have both a polling and a streaming change source for `comment-added` events too - we do add changes for all comments and changefilters on the schedulers parse the comments to decide if builds are to be executed, which we thus only want to happen exactly-once per comment. But comments would need to be deduplicated based on other fields (e.g. the comment timestamp maybe, if that's available?), so that's not as straightforward to add to the existing structure.

p12tic:
@sknebel Thanks for the answer.

Based on the comments in Buildbot issue so far and here, it seems to me that Buildbot is supposed to effectively rebuild after every comment. On a cursory view this would rebuild the same changeset over and over after each comment. Is this the intended behavior? If yes, could you maybe explain why? I'm trying to understand the Gerrit's workflow, as I haven't used it myself in professional capacity.

sknebel:
Yes, this is with additional code that parses the comment and acts on it. In our setup, simplest case is indeed just a plain rebuild (e.g. to retry after a buildfailure that is the fault of the environment - which developers could also do through buildbot UI, but prefer to be able to do in gerrit), but other variations trigger a rebuild that involves more extensive testing or packaging tasks that we do not want to run on every commit.

p12tic:
I think I now understand this much better, thank you.

gerard:
IMHO, every comment should trigger an event, but not a build, as people will do general code reviews.  Our implementation looks for specific keywords on the gerrit comment (lets say !run-tests).  If nothing gets matched, then the event is ignored.  This is done by using change filters on the scheduler, which then parse the comment content and act accordingly (ie, trigger a specific build or ignore)

 * https://docs.buildbot.net/latest/manual/configuration/schedulers.html#change-filters

The fix seems to disable the dedup capability for comment-added, but that should be ok for us (ie, we dont use Stream and Poller simultaneously).

p12tic:
I'm merging this because it's not clear what the best fix would be given the current constraints. To deduplicate additional event types we would need a more complex implementation that would store additional data in the database. So this PR is a good interim solution.

hi-rodee:
I have a setting in my Gerrit called "start-CI". When this setting is set to +1, it triggers the build as I have enabled it for "comment-added" events. However, the build continues to trigger even if I add subsequent comments. While I have a filter in place to trigger the build only when "start-CI" is set to 1, once this setting is initially enabled, any additional comments I make trigger the build. This happens because the "start-CI" setting remains set for the "comment-added" event, causing my condition to return True for any subsequent comments thereafter.

Is this issue fixed in any later version of BuildBot? I am currently on 2.7.0 and I pulled gerritchangesource.py on top to get the comment-added working.

hi-rodee:
> I have a approval bit in my Gerrit called "start-CI". When this setting is set to +1, it triggers the build as I have enabled it for "comment-added" events. However, the build continues to trigger even if I add subsequent comments. While I have a filter in place to trigger the build only when "start-CI" is set to 1, once this approval bit is initially enabled, any additional comments I make trigger the build. This happens because the "start-CI" apporval remains set for the "comment-added" event, causing my condition to return True for any subsequent comments thereafter.
> 
> Is this issue fixed in any later version of BuildBot? I am currently on 2.7.0 and I pulled gerritchangesource.py on top to get the comment-added working.

I could address this by checking the oldValue of start-CI approval bit from the Gerrit stream-event output.

PR #5744, opened 2021-01-06
Allow secrets in token of GitHubPullrequestPoller

### LINKED ISSUE #4921: GitHubPullrequestPoller token secret gives builtins.TypeError:
When attempting the following in GitHubPullrequestPoller:
```
token=util.Secret("githubToken"),
```
I receive: builtins.TypeError: must be str, not Secret

and similarly via Interpolate:
```
token=util.Interpolate("%(secret:githubToken)s"),
```
builtins.TypeError: must be str, not Interpolate

Apologies if I'm doing this wrong but given:
https://docs.buildbot.net/latest/manual/secretsmanagement.html#as-argument-to-services
I had expected this to work. I got the impression that the secret manager was as yet not supported across the board and so if this is an example of an unsupported use then it seems like a nice one to have.

Thanks all for a great project.

Incidentally in both cases I can cast via str() and avoid this issue but then run into another issue which I will report separately.

alexreinking:
What it says in the title. I haven't updated the unit tests because I'm not very familiar with the codebase yet. I have some time to try on my own, but would appreciate any pointers.

Fixes #4921

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/githubpullrequestpoller-secrets.bugfix
+++ b/master/buildbot/newsfragments/githubpullrequestpoller-secrets.bugfix
@@ -1,0 +1,1 @@
+``GitHubPullrequestPoller`` now supports secrets in its ``token`` argument (:issue:`4921`)
--- a/master/buildbot/changes/github.py
+++ b/master/buildbot/changes/github.py
@@ -106,6 +106,7 @@ class GitHubPullrequestPoller(base.ReconfigurablePollingChangeSource,
 
         http_headers = {'User-Agent': 'Buildbot'}
         if token is not None:
+            token = yield self.renderSecrets(token)
             http_headers.update({'Authorization': 'token ' + token})
 
         self._http = yield httpclientservice.HTTPClientService.getService(
--- a/master/buildbot/test/unit/changes/test_github.py
+++ b/master/buildbot/test/unit/changes/test_github.py
@@ -20,7 +20,11 @@ from twisted.trial import unittest
 
 from buildbot.changes.github import GitHubPullrequestPoller
 from buildbot.config import ConfigErrors
+from buildbot.process.properties import Properties
+from buildbot.process.properties import Secret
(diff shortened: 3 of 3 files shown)


p12tic:
@alexreinking : Thanks a lot, this looks good.

As for tests, the tests for this class live in master/buildbot/test/unit/changes/test_github.py. Secrets provider can be setup like this (from master/buildbot/test/unit/www/test_oauth.py):

```
        fake_storage_service = FakeSecretStorage()
        fake_storage_service.reconfigService(secretdict={"client-id": "secretClientId",
                                                         "client-secret": "secretClientSecret"})
        secret_service = SecretManager()
        secret_service.services = [fake_storage_service]
        yield secret_service.setServiceParent(self._master)
```

Thanks!

master/buildbot/newsfragments/githubpullrequestpoller-secrets.bugfix
```diff
@@ -0,0 +1 @@
+``GitHubPullrequestPoller`` now supports secrets in its ``token`` argument
```
p12tic, line 1: Nitpick: append ```(:issue:`4921`) ``` to mention the bug number.

alexreinking: Done!

alexreinking:
@p12tic - that snippet didn't work for me unless I moved the `reconfigService` call to the end. Looks like both `secret_service.setServiceParent` and `self.master.startService` try to configure the `FakeSecretStorage` service and wind up overwriting the secretdict.

Either way, I added a test by running the same steps as in SimplePR, but with the token set to `Secret('token')`. Let me know if I need to do anything else! I also squashed this down to one commit since my fiddling isn't interesting history-wise.

p12tic (approved):
Looks great, thanks!

alexreinking:
You're welcome! Thank you for helping me with, reviewing, and ultimately accepting my PR!

Issue #5743, opened 2021-01-05
Use MD5 instead of md5 to create auth credentials


markus-sus:
I ran into a problem in which using urllib request in python3 fails to connect to the buildbot api using basic authentication. The reason is simply that the built in python library cannot handle 'md5' hash algorithm but it can handle 'MD5'.

I think the relevant code is in www/auth.py when calling DigestCredentialFactory in functions HTPasswdAuth and UserPasswordAuth.

Looking at this header in the browser the data seems to be reported as 'MD5' but when python urllib.request library connects it will get the header as md5.

I did a local test in which I changed the lines to use MD5 instead of md5 in the auth.py and it seems to solve the problem and still works in the browser.

Is there any specific reason why md5 is used? If not then I could go through the process of supplying a patch.

I will file a bug report about making python urllib.request library handle different cases just to see what they think.

markus-sus:
Ok. I found some kind of specification for the www-authenticate header and it seems like 'md5' is not listed as an alternative and it should be MD5 so I think this should change in the buildbot code.

tardyp:
Hello,

As per twisted doc, it is supposed to be case insensitive, and the default is lowercase:
https://github.com/twisted/twisted/blob/trunk/src/twisted/cred/credentials.py#L200

In the RFC, is is also mentionned that this is case insensitive.
https://tools.ietf.org/html/rfc3230#section-4.1.1

So I would say this is a problem in python library.

https://github.com/python/cpython/blob/3.9/Lib/urllib/request.py#L1206
these checks should be case insensitive.

tardyp:
https://github.com/python/cpython/pull/24122

markus-sus:
Oooh, thank you. Then I do not have to do file a bug report.

p12tic:
I think we still need to work around this in Buildbot code as the feature does not work as intended right now. There will be many people during the next following years that will use Python versions without fix for python/cpython#24122, most of them will not know the bug exists and it will appear as if it is Buildbot which is broken, not Python.

@markus-sus: It would be great if you could do a PR to fix this problem.

I think we don't need a test here, but please add a comment with a link to the bug report in cpython explaining why we have the algorithm as uppercase.

markus-sus:
I will schedule a task for making a PR sometime in the next two weeks.
Not very used to git and github so bear with me as I need to closely follow some guide :/

tardyp:
you can see the documentation on how to make PR
[...]

PR #5746, opened 2021-01-11
CHANGE: md5 -> MD5 to handle python bug

markus-sus:
There is a bug in python in which the md5 hash algorithm is checked case sensitive.
Only MD5 is accepted by urllib which is not according to standard.
Buildbot is using md5 which is accepted by standard but due to python having this problem a workaround is needed

Fix: https://github.com/buildbot/buildbot/issues/5743

https://github.com/python/cpython/pull/24122 for more info.

## Contributor Checklist:


tardyp (approved):

p12tic:
Thanks!

PR #5700, opened 2020-12-16
www: Add build summary tooltip to build bubbles in console and grid v…

slim-builder:
Add feature based on request in #4733.

## Contributor Checklist:
* [x] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)

PR #5747, opened 2021-01-12
www: Display fixes for buildsummary tooltip ...

slim-builder:
... and add the tooltip to masters page and builds tables

## Contributor Checklist:

### DIFF
--- a/www/base/src/app/common/directives/builds/buildstable.tpl.jade
+++ b/www/base/src/app/common/directives/builds/buildstable.tpl.jade
@@ -15,10 +15,16 @@
           td(ng-show="builders") {{ builders.get(build.builderid).name }}
           td
             a.bb-buildid-link(ui-sref="build({builder:build.builderid, build:build.number})")
-              span.badge-status(ng-class="results2class(build, 'pulse')")
-                  span.badge-inactive
-                    | {{build.number}}
-                  span.badge-active {{results2text(build)}}
+              script(type="text/ng-template" id="buildsummarytooltip")
+                buildsummary(buildid="build.buildid" type="tooltip")
+              span.badge-status(uib-tooltip-template="'buildsummarytooltip'"
+                                tooltip-class="buildsummarytooltipstyle"
+                                tooltip-placement="auto left-bottom"
+                                tooltip-popup-close-delay="100"
+                                ng-class="results2class(build, 'pulse')")
+                span.badge-inactive
+                  | {{build.number}}
+                span.badge-active {{results2text(build)}}
           td
             span(title="{{build.started_at | dateformat:'LLL'}}")
               | {{build.started_at | timeago }}
--- a/www/base/src/app/common/directives/buildsummary/buildsummarytooltip.tpl.jade
+++ b/www/base/src/app/common/directives/buildsummary/buildsummarytooltip.tpl.jade
@@ -1,37 +1,44 @@
 .panel.panel-default(ng-class="results2class(buildsummary.build)", style="margin-bottom:0px")
     .panel-heading
         .flex-row(ng-if="buildsummary.build.started_at")
-            .flex-grow-1.text-left
+            .flex-grow-1
                 span(ng-if="buildsummary.prefix") {{buildsummary.prefix}}#{' '}
                 span
-                    | {{buildsummary.builder.name}}#{' '}
+                    | {{buildsummary.builder.name | limitStringLengthTo:80}}
                 span.badge-status(ng-class="results2class(buildsummary.build)")
                     | {{buildsummary.build.number}}#{' '}
-                span(ng-if="buildsummary.reason") | {{buildsummary.reason}}
-            .flex-grow-1.text-right
+                span(ng-if="buildsummary.reason")
+                    | {{"&nbsp;"}} | {{buildsummary.reason}}
+        .flex-row(ng-if="buildsummary.build.started_at")
+            .flex-grow-1
                 span(ng-show="buildsummary.build.complete")
                     | {{(buildsummary.build.complete_at - buildsummary.build.started_at)| durationformat:'LLL' }}
                 span(ng-show="!buildsummary.build.complete")
                     | {{(buildsummary.now - buildsummary.build.started_at)| durationformat:'LLL' }}
-                span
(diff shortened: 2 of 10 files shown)


www/base/src/app/common/directives/builds/buildstable.tpl.jade
```diff
@@ -15,10 +15,16 @@
           td(ng-show="builders") {{ builders.get(build.builderid).name }}
           td
             a.bb-buildid-link(ui-sref="build({builder:build.builderid, build:build.number})")
-              span.badge-status(ng-class="results2class(build, 'pulse')")
-                  span.badge-inactive
-                    | {{build.number}}
-                  span.badge-active {{results2text(build)}}
+                script(type="text/ng-template" id="buildsummarytooltip")
```
tardyp, line 18: the indent should be 2 in this file

www/base/src/app/common/directives/builds/buildstable.tpl.jade
```diff
@@ -15,10 +15,16 @@
...
+                    buildsummary(buildid="build.buildid" type="tooltip")
+                span.badge-status(uib-tooltip-template="'buildsummarytooltip'"
```
tardyp, line 20: the double quote sounds weird. why is that?

www/base/src/app/common/directives/buildsummary/buildsummarytooltip.tpl.jade
```diff
@@ -1,37 +1,46 @@
 .panel.panel-default(ng-class="results2class(buildsummary.build)", style="margin-bottom:0px")
     .panel-heading
         .flex-row(ng-if="buildsummary.build.started_at")
-            .flex-grow-1.text-left
+            .flex-grow-1
                 span(ng-if="buildsummary.prefix") {{buildsummary.prefix}}#{' '}
                 span
-                    | {{buildsummary.builder.name}}#{' '}
+                    | {{buildsummary.builder.name.slice(0,80)}}#{' '}
```
tardyp, line 7: we should probably have the ellipsis logic in the controller, or at least factor it

tardyp (changes requested):
a cool feature! I suggested some code improvement

www/base/src/app/common/directives/builds/buildstable.tpl.jade
slim-builder: I fixed the indents.

www/base/src/app/common/directives/builds/buildstable.tpl.jade
slim-builder: The single quotes inside of the double quotes are necessary for the uib-tooltip-template to display correctly.  I followed the tooltip code sample at https://angular-ui.github.io/bootstrap/

www/base/src/app/common/directives/buildsummary/buildsummarytooltip.tpl.jade
slim-builder: I modified the string truncation and ellipsis logic in the template to be a little bit more readable.  I hope this change is acceptable.

floriandorre:
https://github.com/buildbot/buildbot/pull/5700#issuecomment-759421256

p12tic:
@slim-builder: Could you please add a some spacing between the buildstep bubble and the step name? A single space is enough. Right now they are right next to each to each other.

www/base/src/app/common/directives/buildsummary/buildsummarytooltip.tpl.jade
```diff
@@ -1,37 +1,38 @@
 .panel.panel-default(ng-class="results2class(buildsummary.build)", style="margin-bottom:0px")
     .panel-heading
         .flex-row(ng-if="buildsummary.build.started_at")
-            .flex-grow-1.text-left
+            .flex-grow-1
                 span(ng-if="buildsummary.prefix") {{buildsummary.prefix}}#{' '}
                 span
-                    | {{buildsummary.builder.name}}#{' '}
+                    | {{buildsummary.builder.name | limitTo: 80}}{{buildsummary.builder.name.length > 80 ? " &hellip;" : ""}}#{' '}
```
tardyp, line 7: The limitTo filter is a good idea, but I still think this is too much repetitive boilerplate.
I would suggest to factorize this into a angularJs filter. something like this: https://stackoverflow.com/a/53004210

tardyp (changes requested):

www/base/src/app/common/directives/buildsummary/buildsummarytooltip.tpl.jade
p12tic: @tardyp I would prefer simplier and more verbose code in this case. AngularJs is a legacy codebase which fewer and fewer people will know and understand. I would argue that introducing any kind of additional complexity is technical debt by itself, because this will need to be unwound in the future when someone needs to rewrite this to a maintained framework.

slim-builder: I added a custom angularJs filter called limitStringLengthTo.  Please re-review this PR.

www/base/src/app/common/filters/limitStringLength.filter.js
```diff
@@ -0,0 +1,11 @@
+var limitStringLength = function ($filter) {
```
tardyp, line 1: I think this is indeed simpler, and does not add that much complexity.

@p12tic as you disagreed with this approach, I'd like your approval on this. I would understand your desire to  revert to previous design. (@slim-builder you would use 'git reflog' in order to find and go back to prev commit, with git reset)

Eventually, from a UX perspective, we should actually have a directive for that. And have that directive show the full-lenght string as a tooltip.

I think this goes outside of this PR.

p12tic (approved):

www/base/src/app/common/filters/limitStringLength.filter.js
p12tic: LGTM.

floriandorre:
Hi, would it be possible to only show the build step that are not hidden by hideStepIf ?

p12tic:
@floriandorre This is good idea. The hidden status is exposed as the  `step.hidden` attribute.

PR #5753, opened 2021-01-16
Improve AvatarGithub to handle emails

lephilousophe:
This PR adds two fixes in avatar handling and enhance AvatarGitHub.
One fix is for cache which currently doesn't work and the other is for size handling which is broken when specified on URL.

AvatarGitHub gets email support. It queries GitHub REST API to find the correct avatar URL and doesn't try to build a URL from undocumented information.
For this, httpclientservice is used. It needs a BuildbotMaster object so it is added by AvatarResource on all AvatarBase object when configuring.
Fake httpclientservice didn't support headers in requests for testing. It's now checked from expect.
The size argument is now supported by appending s parameter to avatar URL. It doesn't seem documented but works.

AvatarGitHub documentation has been updated to add information on class constructor.

Migrating to GitHub GraphQL API may improve performance but it hasn't been tried.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/www/test_avatar.py
+++ b/master/buildbot/test/unit/www/test_avatar.py
@@ -22,6 +22,13 @@ from buildbot.www import auth
 from buildbot.www import avatar
 
 
+class TestAvatar(avatar.AvatarBase):
+
+    def getUserAvatar(self, email, username, size, defaultAvatarUrl):
+        return defer.succeed((b"image/png", '{!r} {!r} {!r}'.format(
+                                email, size, defaultAvatarUrl).encode('utf-8')))
+
+
 class AvatarResource(TestReactorMixin, www.WwwTestMixin, unittest.TestCase):
 
     def setUp(self):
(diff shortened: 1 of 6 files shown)


master/buildbot/www/avatar.py
```diff
@@ -34,12 +37,143 @@ def getUserAvatar(self, email, username, size, defaultAvatarUrl):
...
+        if not avatar and email:
+            # Try searching a user with said mail
+            avatar = yield self._search_avatar_by_user_email(email)
+        if not avatar and email:
+            # No luck, try to find a commit with this email
+            avatar = yield self._search_avatar_by_commit(email)
+
+        if not avatar:
+            # No luck
+            return
```
p12tic, line 171: I think we should do `return None` to make code more explicit.

p12tic (changes requested):
This is great PR, thanks!

I would like to have tests for all the code paths that are currently not handled as they will be very useful when refactoring code in the future.

Also, it seems that pylint tests will fail. To run these the same way our CI does, do the following:

pip install -r requirements-ci.txt
make pylint flake8

You can change the number of cores that are being used to run pylint in master/Makefile

master/buildbot/test/unit/www/test_avatar.py
```diff
@@ -92,3 +102,286 @@ def getUserAvatar(self, email, username, size, defaultAvatarUrl):
         res = yield self.render_resource(rsrc, b'/?email=foo')
         self.assertEqual(res, dict(redirected=b'//www.gravatar.com/avatar/acbd18db4cc2f85ce'
                          b'def654fccc4a4d8?d=retro&s=32'))
+
+githubUsernameSearchReply = {
```
p12tic, line 106: Let's prefer snake_case for any new code that is not a public API.

master/buildbot/www/avatar.py
```diff
@@ -34,12 +37,143 @@ def getUserAvatar(self, email, username, size, defaultAvatarUrl):
...
+            self.github_api_endpoint = self.DEFAULT_GITHUB_API_URL
+        self.token = token
+        self.debug = debug
+        self.verify = verify
+
+        self.master = None
+        self.client = None
+
+    @defer.inlineCallbacks
+    def _getHTTPClient(self):
```
p12tic, line 60: Let's prefer snake_case for any new code that is not a public API.

lephilousophe: Done

master/buildbot/test/unit/www/test_avatar.py
lephilousophe: Done

master/buildbot/www/avatar.py
lephilousophe: Done

lephilousophe:
flake8 and pylint are now clear from errors.
Let's see how codecov behaves now.

p12tic (approved):
Looks great, thanks!

PR #5756, opened 2021-01-17
Use username when email is empty in REST API

lephilousophe:
When using GithubAuth with API version 4 (to get teams membership) you only get email when profile has one public.
This is IMO a good behavior but in this case, using a force scheduler for example lose the information of the owner because email field is empty.
This fix makes use of username when email is not set.

I don't think documentation should be updated.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/rest.py
+++ b/master/buildbot/www/rest.py
@@ -231,7 +231,10 @@ class V2RootResource(resource.Resource):
             if 'anonymous' in userinfos and userinfos['anonymous']:
                 owner = "anonymous"
             else:
-                owner = userinfos['email']
+                for field in ('email', 'username', 'full_name'):
+                    owner = userinfos.get(field, None)
+                    if owner:
+                        break
             params['owner'] = owner
 
             result = yield ep.control(method, params, kwargs)
--- a/master/buildbot/newsfragments/handle-empty-email.bugfix
+++ b/master/buildbot/newsfragments/handle-empty-email.bugfix
@@ -1,0 +1,1 @@
+If logged in user has an empty email, use its username instead or its full name ultimately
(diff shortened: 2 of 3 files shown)


master/buildbot/www/rest.py
```diff
@@ -230,8 +230,10 @@ def writeError(msg, errcode=399,
             userinfos = self.master.www.getUserInfos(request)
             if 'anonymous' in userinfos and userinfos['anonymous']:
                 owner = "anonymous"
-            else:
+            elif userinfos['email']:
                 owner = userinfos['email']
+            else:
+                owner = userinfos['username']
```
tardyp, line 236: username is not specified as being part of the userinfos, so we should probably use .get('username')

https://docs.buildbot.net/latest/developer/cls-auth.html#buildbot.www.auth.UserInfoProviderBase

tardyp (changes requested):

master/buildbot/www/rest.py
lephilousophe: What is best? full_name or username?
We could use email then username, then full_name or something like that.

tardyp: I think username and email are more stable and should be preferred.
looking at the implementation, it look like most userInfo providers are actually filling the username.
Perhaps best approach is to just document that we should fill the username.

lephilousophe: There is another source where username is mentioned: https://github.com/buildbot/buildbot/blob/master/master/docs/developer/www-server.rst#server-side-session

lephilousophe: I used get and looking at email, username, full_name in this order.
UserInfoProviderBase only sets email so I don't think updating alone is a good idea.

lephilousophe:
I can't look at code coverage anymore. Is this intended?
Anyway, I added tests which should cover this piece of code.

p12tic:
> I can't look at code coverage anymore. Is this intended?

I think some time must pass after a push until coverage results are available again.

p12tic (approved):

p12tic:
Looks great, thank you!

PR #5765, opened 2021-01-21
Fix reference to tuplematch in base reporters

umar456:
The tuplematch object was not referenced in the ReporterBase class when referenced from the GitHubStatusPush class.

It seems this is not an issue when it is executed from the test but only occures when called from GitHubStatusPush. I don't know how I can test this.

Fixes #5765 

## Contributor Checklist:


p12tic:
Updated to match our coding standard. Thanks a lot!

PR #5768, opened 2021-01-23
master/www: Optionally enable WS pings to avoid disconnects

### LINKED ISSUE #4078: Send keepalive messages on websocket to prevent cloudflare disconnect
buildbot behind cloudflare is currently reloading for me every 100 seconds. This is due to a cloudflare limitation that is resetting websockets with 100 seconds of inactivity. It seems the only ways around that is either an upgrade to their enterprise plan or adding keepalive messages to the application.

https://community.cloudflare.com/t/cloudflare-websocket-timeout/5865

I'm on buildbot 1.0.0 due to openbsd -stable, I've grepped through the source on master and it seems there's already a ping command implemented on the server, but it's not automatically used by the client yet. The client should invoke it every 60 seconds to avoid a disconnect due to timeouts.

Thanks!

leoetlino:
This adds a configuration option for the websocket ping interval.

Pings are useful to keep the connection alive and avoid disconnects
when going through reverse web proxies. By default, nginx closes
websocket connections after 60 seconds of inactivity and CloudFlare
enforces a 100 second timeout for non-enterprise users.

Considering the web UI reloads everything when the websocket connection
is lost, adding this option would improve usability as it prevents
full page reloads from occurring every minute.

The option defaults to 0 (no ping) to match the existing behavior.

Fixes #4078.

## Contributor Checklist:

* [x] I have updated the unit tests *(not applicable)*

### DIFF
--- a/master/docs/manual/configuration/www.rst
+++ b/master/docs/manual/configuration/www.rst
@@ -184,6 +184,14 @@ This server is configured with the ``www`` configuration key, which specifies a 
             'Workers.showWorkerBuilders': True,
         }
 
+``ws_ping_interval``
+
+    Send websocket pings every ``ws_ping_interval`` seconds.
+    This is useful to avoid websocket timeouts when using reverse proxies or CDNs.
+    If the value is 0, pings are disabled.
+
+    The default is 0.
+
 .. note::
 
     The :bb:cfg:`buildbotURL` configuration value gives the base URL that all masters will use to generate links.
--- a/master/buildbot/newsfragments/ws-ping-interval.feature
+++ b/master/buildbot/newsfragments/ws-ping-interval.feature
@@ -1,0 +1,1 @@
+A new ``www.ws_ping_interval`` configuration option was added to avoid websocket timeouts when using reverse proxies and CDNs. Fixes :issue:`4078`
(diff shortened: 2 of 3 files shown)


tardyp (approved):
looks good! thanks!

master/docs/manual/configuration/www.rst
```diff
@@ -184,6 +184,14 @@ This server is configured with the ``www`` configuration key, which specifies a
             'Workers.showWorkerBuilders': True,
         }
 
+``ws_ping_interval``
+
+    Send websocket pings every ws_ping_interval seconds.
```
p12tic, line 189: @leoetlino : Spellcheck fails on ws_ping_interval, could you put it into backquotes so that it's interpreted as code?

p12tic (changes requested):
Just a simple nitpick to fix build. Otherwise looks good, thanks!

master/docs/manual/configuration/www.rst
leoetlino: Sure. Also fixed the newsfragment

p12tic (approved):
Thanks!

pablogsal:
@p12tic I think this should be updated in https://github.com/buildbot/buildbot/blob/c1c4812cc67265798dd597b06f113943dc154112/master/buildbot/config.py#L717

PR #5777, opened 2021-01-26
2.10: Fix tour config file error in IRC setup.

p12tic:
This PR backports #5762 to 2.10.


p12tic:
Test failure seems to be unrelated.

PR #5778, opened 2021-01-27
BitbucketPullrequestPoller: PR properties saving, auth possibility

obaltian:
This PR adds 2 important features for **BitbucketPullRequestPoller**:
1. An interface to provide Bitbucket authorization data (many bitbucket storages require this).
2. Possibility to save PR API response entries as build properties (just as **GithubPullRequestPoller** can)

This is done to unify the opportunities of 2 remote service pollers.
For backwards compatibility and to adhere naming rules, public class interface name is  `bitbucket_property_whitelist`.

### Contributor checklist

### PS
Dear maintainers, please, clarify the question below:
Since **BitbucketServerEventHandler** and **BitbucketCloudEventHandler** already provide a `pullrequesturl` property, I'm not sure how we can unify their properties interface with the pollers ones.
Maybe we should enhance them with the same functionality (`PullRequestMixin`), and save `pullrequesturl` property for backwards compatibility?
If you have any thoughts, it can be done in this PR.

Thanks for your opinions!


p12tic:
This PR looks good on a cursory glance, thanks! I will return to do a thorough review and will answer your question.

p12tic:
@obaltian I've looked into the PR again with more focus and it looks really nice.

To answer your question:

> Since BitbucketServerEventHandler and BitbucketCloudEventHandler already provide a pullrequesturl property, I'm not sure how we can unify their properties interface with the pollers ones.
Maybe we should enhance them with the same functionality (PullRequestMixin), and save pullrequesturl property for backwards compatibility?

I haven't written the code in question, but it looks to me that the parsed properties will expose the pull request URL under different paths. E.g. `bitbucket.links.html.href` here, but it could be something like `github.pull.request.url` or something like that. Since pull request URL is a property which all web version control systems will expose one way or another, I think it makes sense to have it as a "special" common property that the pollers will fill in when that information is available. The same can be said about several other properties, but let's limit the current discussion to pullrequesturl for now.

Does the above sound good to you?

p12tic:
Could you rebase the PR on top of newest master branch to remove the merge commits? Also, I'm not completely sure, but there might be a couple of commits that essentially fix a problem introduced in a previous commit. These should be merged into commits introducing the problem as this kind of commit history often just confuses whoever is investigating it.

obaltian:
> Does the above sound good to you?
Yes, thanks. This means I will update my PR with 2 things, just as you said:
1. I would add PullRequestMixin functionality into bitbucket event handlers, mentioned above
2. I would add `pullrequesturl` property to the github event handler and bitbucket/github pullrequest pollers.

Is it okay?

p12tic:
> Is it okay?

Your plan sounds great. Thanks!

obaltian:
@p12tic Done! Please, check the latest changes.
I left the usual **BitBucketHandler** class unchanged, since I didn't found any pullrequest-related functionality.

However, I'm very confused with codecov fail above, since I didn't changed any files or functionality it reported too...

obaltian:
Hello, @p12tic. Could you kindly review asked changes in this PR?
Just a reminder. Thanks!

p12tic:
@obaltian  Looks good, thanks. With regard to commits, I didn't mean to merge everything into a single commit, just that each commit should do a single thing and if there are two commits doing the same thing (e.g. create feature X, fix bug in feature X), then they should be in a single commit. I've ended up splitting the commit into multiple ones so that if there are bugs it's easier to bisect in the future.

p12tic (approved):

p12tic:
Test failure is an instability that is unrelated to this PR

PR #5789, opened 2021-02-05
Add support for addChange(..., src='fossil').

stoklund:
I just published a [buildbot-fossil](https://pypi.org/project/buildbot-fossil/) plugin on PyPI which enables Buildbot to be used with projects hosted in a [Fossil](https://fossil-scm.org/) repository.

Fossil uses usernames in the same style as Subversion. I ran into a problem when my Fossil change source calls `master.addChange()` with `src='fossil'`. No user object is created because `createUserObject()` doesn't know about Fossil users, and an error message is logged.

This PR crudely adds Fossil to the list of allowed user name sources. Do you have any suggestions for a better way of handling this problem in a third party plugin?

### DIFF
--- a/master/buildbot/newsfragments/fossil-users.feature
+++ b/master/buildbot/newsfragments/fossil-users.feature
@@ -1,0 +1,1 @@
+Added support for Fossil user objects for use by the buildbot-fossil plugin.
--- a/master/buildbot/process/users/users.py
+++ b/master/buildbot/process/users/users.py
@@ -25,7 +25,7 @@ from buildbot.util import bytes2unicode
 from buildbot.util import flatten
 from buildbot.util import unicode2bytes
 
-srcs = ['git', 'svn', 'hg', 'cvs', 'darcs', 'bzr']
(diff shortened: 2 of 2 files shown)


veniaventis:
Hello 
I'm beginner Python developer, and need practice with projects.
Can I help you ? 

I apologize for writing here, this is just the first time I participate in an open source project:(

p12tic (approved):

p12tic:
@stoklund Thanks. We actually need an API that a change source could use to register allowed source types (or just remove the check altogether).

By the way, would you want to contribute your plugin to Buildbot proper? I've looked into the code and aside from a couple of nitpicks it would be good to go.

stoklund:
> By the way, would you want to contribute your plugin to Buildbot proper? I've looked into the code and aside from a couple of nitpicks it would be good to go.

Thanks for reviewing my code! At the moment, the plugin is marked as beta because I only just started using it. After I test the different code paths more thoroughly, I think it would be a good idea to contribute it to Buildbot proper.

I'll make a PR for that at some point.

PR #5848, opened 2021-03-05
fix secret replacement for whitespace only / empty secrets

terencehonles:
I have a configuration for a testing environment which stubs out a secret with the empty string. This causes the status strings and logs to be replaced with a ton of `<SECRET_NAME> <SECRET_NAME>...` i.e.:

```python
>>> 'finished'.replace('', '<SECRET>')`
'<SECRET>f<SECRET>i<SECRET>n<SECRET>i<SECRET>s<SECRET>h<SECRET>e<SECRET>d<SECRET>'
```

## Contributor Checklist:

* [x] I have updated the unit tests
* [x] I have created a file in the `master/buildbot/newsfragments` directory (and read the `README.txt` in that directory)
* [x] I have updated the appropriate documentation

PR #5849, opened 2021-03-06
use `ssh -o BatchMode=yes` when making SSH connections

terencehonles:
SSH connections are now run with `-o BatchMode=yes` to prevent interactive prompts which may tie up a step until it times out. From the SSH docs:

>  If set to yes, user interaction such as password prompts and host key confirmation requests will be disabled. This option is useful in scripts and other batch jobs where no user is present to interact with ssh.

This will allow a step to immediately fail and the error can now be inspected. Steps which make their own SSH connections are encouraged to also use `-o BatchMode=yes` in order to not have surprising behaviors if the host configuration is different.

## Contributor Checklist:


terencehonles:
I'm not sure if there are any implementations that do not support this flag, but I ran into this by my own SSH connection that didn't have a host key set up in the configuration I was running, and it wasn't obvious what was happening since there were other things acting strange (#5848). Once I got to the bottom of it I decided to look at the Git step to see how it worked, and it just happened that I had already accepted GitHub's host key and removing it caused the step to hang.

master/buildbot/www/ldapuserinfo.py
```diff
@@ -26,15 +26,19 @@
...
-
 from twisted.internet import threads
 
 from buildbot.util import bytes2unicode
 from buildbot.util import flatten
 from buildbot.www import auth
 from buildbot.www import avatar
 
+try:
+    import ldap3
```
p12tic, line 37: This change needs to be split off to a separate PR.

p12tic (changes requested):
Looks good, but please split off the ldap3 import change to a separate commit/PR.

master/buildbot/www/ldapuserinfo.py
terencehonles: Ok, I only included it because I noticed `make trial` was failing on this test because just importing this file would cause the test suite to fail.

terencehonles: Moved to #5857

p12tic (approved):

master/buildbot/newsfragments/ssh-batch-mode.feature
```diff
@@ -0,0 +1,11 @@
+SSH connections are now run with ``-o BatchMode=yes`` to prevent interactive
```
p12tic, line 1: For the future, it's best to keep the release notes short, one-two sentences unless it's a change that affects majority of users in a way that requires their attention (e.g. deprecation/replacement of frequently used feature). It's best to explain this in the documentation of affected steps.

PR #5857, opened 2021-03-06
fix failing tests if dependencies are not available

terencehonles:
The tests `master/buildbot/test/unit/www/test_ldapuserinfo.py` require `ldap3` or they will fail with an import error. This change allows the import to happen and will throw the error only if `LdapUserInfo` is used and `ldap3` is not available.

## Contributor Checklist:

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -83,6 +83,9 @@ matrix:
     - python: "3.9"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=trial
 
+    - python: "3.9"
+      env: TWISTED=latest SQLALCHEMY=latest TESTS=minimal_install
+
     - python: "3.8"
       env: TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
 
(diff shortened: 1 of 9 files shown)


p12tic:
@terencehonles Now that I need to actually review the PR, it got me thinking. Is there any downside of asking to install `ldab3` in order to run tests? We already have "tests" that require additional software to be installed such as isort, flake8, pylint. So we could just ask to do `pip install -r requirements-ci.txt` in order to run tests. The development quickstart documentation (http://docs.buildbot.net/2.10.1/developer/quickstart.html) already tells to run from within `.venv` virtualenv, so this wouldn't be too large change. What do you think?

I should have raised these concerns in the #5849 PR before you spent time creating a separate PR. Sorry about that.

terencehonles:
It's fine, it only took me a couple seconds to pull out the patch and rebase. I'll try return to the question in the week. It might make sense to add the CI requirements to the make step.

I do think the install and the test should be optional though because ldap3 is not a core dependency. All the other tests will skip themselves rather than fail if the dependencies are not met. I think it might also make sense to do a test of a clean pip install without the ci requirements and see if there are any import errors. Otherwise the CI requirements may be effectively unnamed requirements.

Not sure what you think about that.

p12tic:
I think you bring valid point regarding effectively unnamed dependencies, we need a separate test for that. For that test to work we'll need something like what's in this PR.

terencehonles:
:+1: I'll try to dedicate some time to think about how that would need to be structured and tested in the current CI setup. I don't think that _needs_ to be tested locally, so did you want me to also add an install of the CI requirements in the `make trial` flow so that the tester will have optional dependencies installed?

terencehonles:
@p12tic this had mostly fallen off my radar, but I believe the following CI changes will run the tests with a more restricted installation and I also fixed some other tests that were failing between when I created the original PR and now.

p12tic (approved):
Looks great, thanks!

PR #5915, opened 2021-03-16
process: builder: fix getNewestCompleteTime

ynezz:
Currently getNewestCompleteTime() returns no times as it filters out
records with complete=0(false) which has obviously complete_at=None as
complete_at value is set with epoch and complete=1(true) when the build
finishes. Test case was wrong since beginning because new BuildRequest
in fakedb is created with complete=0.

## Contributor Checklist:


p12tic:
@ynezz This looks great, thanks. Could you also add a release note in master/buildbot/newsfragments directory?

ynezz:
@p12tic Done.

p12tic (approved):
Thanks!

PR #5919, opened 2021-03-18
git: Don't reset last active branch to wrong commit

p12tic:
When we change the checked out code using `git reset --hard <rev>` we also change the revision that the currently active branch points to. This leads to that branch pointing to a wrong commit as we're actually checking out a different branch (we switch to it just a little bit later). Instead of modifying the current branch we should just modify the working copy, which can be done by `git checkout -f <rev>`.

We need branches to point to correct commits in order for `GitDiffInfo` step to work.

[git checkout -f documentation](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt--f):

> When switching branches, proceed even if the index or the working tree differs from HEAD. This is used to throw away local changes. When checking out paths from the index, do not fail upon unmerged entries; instead, unmerged entries are ignored.

I've tested how `git checkout -f` works just to be sure and it's equivalent to `git reset --hard` except that active branch is not affected. I've also tested a real Buildbot installation with the change.

### DIFF
--- a/master/buildbot/newsfragments/git-step-correct-branch-commits.bugfix
+++ b/master/buildbot/newsfragments/git-step-correct-branch-commits.bugfix
@@ -1,0 +1,2 @@
+Fixed the ``Git`` source step causing last active branch to point to wrong commits.
+This only affected the branch state in the repository, the checked out code was correct.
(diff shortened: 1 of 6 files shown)


p12tic:
cc @tardyp Could you take a look? This is one of the more risky changes that may have a significant impact if we get it wrong.

tardyp (approved):

tardyp:
good finding!

tardyp:
merging, to ease review of #5929

PR #5929, opened 2021-03-19
WIP: Implement reporter for code issues in a PR

p12tic:
This is a WIP PR that demonstrates a reporter that takes a list of issues reported by e.g. pylint, determines if they are made in code that has been changed in a PR and for each such issue submits a comment on specific line on a github PR.

See an example screenshot below.

20210319_github_pr_comments

Further work: we probably need to automatically remove all comments that have not been commented on when a new test run completes. Otherwise it introduces additional effort needed by the developers to close no longer relevant issues when a PR is updated.


master/buildbot/reporters/generators/comment.py
```diff
@@ -0,0 +1,154 @@
...
+        return result.get('test_code_path') is not None and result.get('line') is not None
+
+    def _is_result_in_target_diff(self, target_changes_by_path, result):
+        changes_in_path = target_changes_by_path.get(result['test_code_path'])
+        if not changes_in_path:
+            return False
+
+        result_line = result['line']
+
+        preceding_change_i = bisect.bisect(changes_in_path, (result_line + 1, -1))
```
tardyp, line 88: Not sure if bisect is very well known module of python.

I wasn't aware of it.

Also, the fact that we are comparing tuple like this looks suspicious.

I think we need a proper helper function to compare code line intervals (with unit tests)

master/buildbot/reporters/generators/comment.py
```diff
@@ -0,0 +1,154 @@
...
+        if preceding_change_i == 0:
+            return False  # there's no change with start position earlier than the test result line
+        preceding_change_start, preceding_change_length = changes_in_path[preceding_change_i - 1]
+
+        return result_line >= preceding_change_start and \
+            result_line < preceding_change_start + preceding_change_length
+
+    @defer.inlineCallbacks
+    def _get_usable_results_in_changed_lines(self, master, target_changes_by_path, result_sets):
```
tardyp, line 97: an error could be in an untouched line (removal of used import for example).

gitlab allows you to comment on untouched lines. 

So for this generic code, I believe we shouldn't filterout

maybe we could split two sets of results something like:
```
touched_line_issues
untouched_lines_issues
```

tardyp (commented):
I think this feature is really cool!

link2xt:
GitHub has "annotations" which can be seen in https://github.com/actions-rs/clippy-check screenshot and described in https://github.community/t/what-are-annotations/16173

Submitting these comments as a review has a disadvantage that if you push additional commits closing these problems, review comments will not go away and have to be resolved manually.

PR #5931, opened 2021-03-19
users: Remove unused code

p12tic:
This code is completely unused. There's more unused code in the file, but it also has unit tests, so it may useful to still leave it there if we want eventually to add more users-related functionality.


tardyp (approved):
rip

PR #5936, opened 2021-03-21
Fix Bitbucket status push

nlamprian:
Not sure if this is the expected behavior, but the Bitbucket Cloud hook provides sourcestamps with API URLs. API URLs were not parsed successfully by the Bitbucket status push.

I've added two commits. The first one extends the method to support API URLs. The second one refactors the method by the use of a regex. If you don't like the regex, I can remove the commit.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/bitbucket.py
+++ b/master/buildbot/reporters/bitbucket.py
@@ -54,6 +54,7 @@ class BitbucketStatusPush(ReporterBase):
     def reconfigService(self, oauth_key, oauth_secret, base_url=_BASE_URL, oauth_url=_OAUTH_URL,
                         debug=None, verify=None, generators=None, **kwargs):
         oauth_key, oauth_secret = yield self.renderSecrets(oauth_key, oauth_secret)
+        self.base_url = base_url
         self.debug = debug
         self.verify = verify
 
@@ -62,8 +63,7 @@ class BitbucketStatusPush(ReporterBase):
 
         yield super().reconfigService(generators=generators, **kwargs)
 
-        if base_url.endswith('/'):
-            base_url = base_url[:-1]
+        base_url = base_url.rstrip('/')
 
         self._http = yield httpclientservice.HTTPClientService.getService(
             self.master, base_url,
@@ -80,8 +80,7 @@ class BitbucketStatusPush(ReporterBase):
     def sendMessage(self, reports):
         build = reports[0]['builds'][0]
         results = build['results']
-        oauth_request = yield self.oauthhttp.post("",
-                                                  data=_GET_TOKEN_DATA)
+        oauth_request = yield self.oauthhttp.post("", data=_GET_TOKEN_DATA)
         if oauth_request.code == 200:
             content_json = yield oauth_request.json()
             token = content_json['access_token']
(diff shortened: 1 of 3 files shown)


master/buildbot/reporters/bitbucket.py
```diff
@@ -124,14 +124,18 @@ def get_owner_and_repo(repourl):
...
+                    git@bitbucket.org:OWNER/REPONAME.git
+                    https://bitbucket.org/OWNER/REPONAME.git
+                    ssh://git@bitbucket.org/OWNER/REPONAME.git
+                    https://api.bitbucket.org/2.0/repositories/OWNER/REPONAME
         :return: owner, repo: The owner of the repository and the repository name
         """
         parsed = urlparse(repourl)
 
-        if parsed.scheme:
+        base_parsed = urlparse(_BASE_URL)
```
p12tic: I think we should use self.base_url here.

master/buildbot/reporters/bitbucket.py
```diff
@@ -124,26 +124,15 @@ def get_owner_and_repo(repourl):
...
-                    https://bitbucket.com/OWNER/REPONAME.git
-                    ssh://git@bitbucket.com/OWNER/REPONAME.git
+                    git@bitbucket.org:OWNER/REPONAME.git
+                    https://bitbucket.org/OWNER/REPONAME.git
+                    ssh://git@bitbucket.org/OWNER/REPONAME.git
+                    https://api.bitbucket.org/2.0/repositories/OWNER/REPONAME
         :return: owner, repo: The owner of the repository and the repository name
         """
-        parsed = urlparse(repourl)
+        m = re.search(r"^(?:git@|ssh://git@|https://(?:api\.)?)bitbucket\.\w+"
```
p12tic, line 133: I think let's keep this using urlparse. The regex hardcodes the domain and thus no longer supports custom base_url for self-hosted installations.

master/buildbot/reporters/bitbucket.py
```diff
@@ -124,26 +124,15 @@ def get_owner_and_repo(repourl):
...
-        if parsed.scheme:
-            path = parsed.path[1:]
-        else:
-            # we assume git@host:owner/repo.git here
-            path = parsed.path.split(':', 1)[-1]
-
-        if path.endswith('.git'):
-            path = path[:-4]
-        while path.endswith('/'):
-            path = path[:-1]
```
p12tic, line 143: I know it is not a problem in your PR, but could you replace this loop with use of `lstrip('/')`? There's also a case above where `rstrip('/')` would be useful.

p12tic (changes requested):
Thanks a lot! In principle the PR is good, I've added several comments.

master/buildbot/reporters/bitbucket.py
nlamprian: Done

master/buildbot/reporters/bitbucket.py
nlamprian: The method was static. I had to do some more changes. I'm not sure about the test. I hope I did it right.

nlamprian:
On a parallel note, since we are on the topic, how would you feel about adding the buildid to the status, `build['builder']['name']-build['buildid']`. Assuming a dynamic build sequence, this would allow pushing multiple independent statuses for the same commit.

p12tic (approved):

p12tic:
> On a parallel note, since we are on the topic, how would you feel about adding the buildid to the status, build['builder']['name']-build['buildid']. Assuming a dynamic build sequence, this would allow pushing multiple independent statuses for the same commit.

I don't have an answer yet, will comment a bit later.

Issue #5930, opened 2021-03-19
[Feature] Build summary tooltip improvements


PhilippSelenium:
The build summary tooltip box jumps like crazy (To a degree where it almost causes seizures ;)) when hovering over the build numbers.

Solutions:
* Possibility to disable build summary tooltip 
* rendered more relaxed and smoothly
* ?

See: https://github.com/buildbot/buildbot/pull/5700

slim-builder:
@p12tic @PhilippSelenium 
I believe that the tooltip "jumps" because of a performance issue when loading the build summary steps.  The tooltip will display the build summary header first, then later on add the displayed build steps, which causes the tooltip size to increase abruptly.  The abrupt change in the tooltip size causes the "jumps".  I'm not sure how to fix this issue, but as a workaround I propose hiding the build steps and having the tooltip display the build summary header only (example screenshot is below)
![Screenshot from 2021-03-19 11-30-20](https://user-images.githubusercontent.com/12524918/111827944-c6bb1f00-88a7-11eb-8946-1f70d726daff.png)

tardyp:
I think if we force the minimum width to be a bit larger, we can avoid those artifacts.

I think the current tooltip is looking great when everything is finally loaded, so I would prefer we manage to make it work.

I believe some css tweak should fix the issue

PhilippSelenium:
Looks better, Thanks!

PR #5937, opened 2021-03-21
www: Reduce flickering of buildsummary tooltip during mouseover

slim-builder:
Fix for #5930 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/buildsummary-tooltip-flicker-fix.bugfix
+++ b/master/buildbot/newsfragments/buildsummary-tooltip-flicker-fix.bugfix
@@ -1,0 +1,1 @@
+Fix for :issue:`5930`, reduce flickering of build summary tooltip during mouseover of build numbers.
--- a/master/docs/spelling_wordlist.txt
+++ b/master/docs/spelling_wordlist.txt
@@ -520,6 +520,7 @@ mistyped
 mixin
 Mixins
 mkdir
+mouseover
 mq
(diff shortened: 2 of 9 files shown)


tardyp:
hey thanks for the fix!
Note sure the delay will change much of the problem.

can you send a gif/mp4 of the result?

You can simulate slow network in the chrome developer mode..
image

tardyp:
Looks good to me, thanks!

Did you try to look at animating the height property? 
I think this could make even smoother

p12tic:
I think we could look into the height property as part of another PR. The present PR fixes the bug well enough.

tardyp:
Hello, I am afraid we'll have to revert it as it does not fix the flickering issue to me.

I have deployed it to buildbot.buildbot.net and still see the flickering, just less present because of the delay.

https://user-images.githubusercontent.com/109859/112010560-b4e7af00-8b27-11eb-97d4-7d4ebed068c5.mp4


We can see frame by frame what happens. The height raise rapidly when the step data arrives during one frame, and the next frame the popup re-aligns on the popup anchor.

image
image
image

p12tic:
Indeed, that's true. It's barely visible though.

What this PR did fix though (even if by working around the problem) was extremely annoying flickering when the active build bubble is changing size (e.g. in the workers page). So I think this is solving 90% of the problem and we don't really need to revert, a next iteration is enough.

Issue #4016, opened 2018-03-27
Add pending build status for build requests


chrwen-omicron:
This issue is meant to start a discussion.

We're using Buildbot in combination with GitLab where the status of builds get pushed to. Our problem is that GitLab doesn't know what builds might get started by a commit.
For example there might be no worker available or dependent builders. Because of that you can't tell in GitLab if a commit or merge request is fine. Additionally you can't use the automatic merge feature in merge request when the pipeline is green, because there might still be a build which hasn't started.
So I was thinking about something like adding/reporting a pending build status for build requests. That way they would be immediately available in GitLab.
I know there are a few things which have to be thought through, but I would like to hear some other opinions/experiences.

staticfloat:
I'd like to second the request that status reporters be able to respond to a buildrequest.  In the Julia language project, we have a large number of builders, and it's not always easy to realize that a test hasn't even started yet because there were many other builds in the queue for one particular architecture (we have e.g. x86_64, i686, aarch64, across Mac, Windows, Linux, FreeBSD, etc...).  It would be helpful to be able to see a "pending" for all of these immediately so that a maintainer can quickly see the status of the build.

PR #5938, opened 2021-03-21
implement report generator for pending build requests

g-arjones:
## Contributor Checklist:

Closes https://github.com/buildbot/buildbot/issues/4016

This is the first attempt to implement a report generator for pending build requests.

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -201,7 +201,7 @@ class Build(properties.PropertiesMixin):
 
     @staticmethod
     def setupPropertiesKnownBeforeBuildStarts(props, requests, builder,
-                                              workerforbuilder):
+                                              workerforbuilder=None):
         # Note that this function does not setup the 'builddir' worker property
         # It's not possible to know it until before the actual worker has
         # attached.
@@ -225,17 +225,21 @@ class Build(properties.PropertiesMixin):
         # get worker properties
         # navigate our way back to the L{buildbot.worker.Worker}
         # object that came from the config, and get its properties
-        workerforbuilder.worker.setupProperties(props)
+        if workerforbuilder is not None:
+            workerforbuilder.worker.setupProperties(props)
 
-    def setupOwnProperties(self):
+    @staticmethod
+    def setupBuildProperties(props, requests, sources=None, number=None):
         # now set some properties of our own, corresponding to the
         # build itself
-        props = self.getProperties()
-        props.setProperty("buildnumber", self.number, "Build")
+        props.setProperty("buildnumber", number, "Build")
 
-        if self.sources and len(self.sources) == 1:
+        if sources is None:
+            sources = requests[0].mergeSourceStampsWith(requests[1:])
+
+        if sources and len(sources) == 1:
             # old interface for backwards compatibility
-            source = self.sources[0]
+            source = sources[0]
             props.setProperty("branch", source.branch, "Build")
             props.setProperty("revision", source.revision, "Build")
             props.setProperty("repository", source.repository, "Build")
@@ -314,7 +318,8 @@ class Build(properties.PropertiesMixin):
         self.preparation_step = buildstep.BuildStep(name="worker_preparation")
         self.preparation_step.setBuild(self)
         yield self.preparation_step.addStep()
-        self.setupOwnProperties()
+        Build.setupBuildProperties(self.getProperties(), self.requests,
+                                   self.sources, self.number)
 
         # then narrow WorkerLocks down to the right worker
         self.locks = [(l.getLockForWorker(workerforbuilder.worker.workername),
--- a/master/buildbot/reporters/generators/buildrequest.py
+++ b/master/buildbot/reporters/generators/buildrequest.py
@@ -1,0 +1,88 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+from twisted.internet import defer
+from zope.interface import implementer
+
+from buildbot import interfaces
+from buildbot.process.build import Build
+from buildbot.process.buildrequest import BuildRequest
+from buildbot.process.properties import Properties
+from buildbot.reporters import utils
+from buildbot.reporters.message import MessageFormatterRenderable
+
+from .utils import BuildStatusGeneratorMixin
+
+
+@implementer(interfaces.IReportGenerator)
+class BuildRequestGenerator(BuildStatusGeneratorMixin):
+
+    wanted_event_keys = [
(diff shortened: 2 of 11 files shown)


g-arjones:
My initial goal was to reuse code from [process.build.Build](https://github.com/g-arjones/buildbot/blob/6c396acaae821e4d4f259d9a6c89464a7887de26/master/buildbot/process/build.py#L44) in [reporters.utils.getDetailsForBuildRequest](https://github.com/g-arjones/buildbot/blob/6c396acaae821e4d4f259d9a6c89464a7887de26/master/buildbot/reporters/utils.py#L92) to create a partial `build` dict (something like [process.build.Build.setupPropertiesKnownBeforeBuildStart](https://github.com/g-arjones/buildbot/blob/6c396acaae821e4d4f259d9a6c89464a7887de26/master/buildbot/process/build.py#L203)) but I had no luck. Please let me know if there's something I am missing.

master/buildbot/reporters/generators/buildrequest.py
```diff
@@ -0,0 +1,86 @@
...
+    @defer.inlineCallbacks
+    def buildrequest_message(self, master, build):
+        patches = self._get_patches_for_build(build)
+        users = []
+        buildmsg = yield self.formatter.format_message_for_build(master, build, mode=self.mode,
+                                                                 users=users)
+
+        results = build['results']
+        subject = buildmsg['subject']
+        if subject is None and self.subject is not None:
```
p12tic, line 71: self.subject will always be None with the current implementation.

master/buildbot/test/unit/reporters/test_generators_buildrequest.py
```diff
@@ -0,0 +1,156 @@
...
+            'type': 'plain',
+            'results': None,
+            'builds': [build],
+            'users': [],
+            'patches': [],
+            'logs': []
+        })
+
+    @defer.inlineCallbacks
+    def test_is_message_needed_ignores_unspecified_tags(self):
```
p12tic, line 96: This test is no longer needed as the is_message_needed_by_props is tested in test_generators_utils. I see that TestBuildStartEndGenerator has it too so existing code has the same problem.

I think we could change this test to call `generate` like the test_generate_new below. This way we'll cover the case when generate does not generate a report. The test would need to be moved to the end of the file in that case.

master/buildbot/test/unit/reporters/test_generators_buildrequest.py
```diff
@@ -0,0 +1,156 @@
...
+        g = BuildRequestGenerator(tags=['not_existing_tag'])
+        buildrequest = yield self.insert_buildrequest_new()
+        build = yield g.partial_build_dict(self.master, buildrequest)
+
+        # force tags
+        build['builder']['tags'] = ['tag']
+        self.assertFalse(g.is_message_needed_by_props(build))
+
+    @defer.inlineCallbacks
+    def test_is_message_needed_tags(self):
```
p12tic, line 106: This test is not needed.

master/buildbot/reporters/utils.py
```diff
@@ -87,6 +87,32 @@ def getDetailsForBuild(master, build, wantProperties=False, wantSteps=False,
     return ret
 
 
+@defer.inlineCallbacks
+def getDetailsForBuildRequest(master, buildrequest):
```
p12tic, line 91: Let's use snake_case for new code. Eventually all of buildbot will use snake_case like normal Python code.

master/buildbot/reporters/utils.py
```diff
@@ -87,6 +87,32 @@ def getDetailsForBuild(master, build, wantProperties=False, wantSteps=False,
...
+    buildset = yield master.data.get(("buildsets", buildrequest['buildsetid']))
+    builder = yield master.data.get(("builders", buildrequest['builderid']))
+
+    build = dict()
+    build['buildrequest'], build['buildset'] = buildrequest, buildset
```
p12tic, line 96: I would prefer separate lines

p12tic: I think it makes sense to use the dict literal syntax ( `{key: 'value', ...}`) here.

p12tic:
> My initial goal was to reuse code from process.build.Build in reporters.utils.getDetailsForBuildRequest to create a partial build dict (something like process.build.Build.setupPropertiesKnownBeforeBuildStart, without a worker) but I had no luck. Please let me know if there's something I am missing.

What was the problem that you were seeing? I think it makes sense to adapt `setupPropertiesKnownBeforeBuildStart` to our requirements. The obvious one is to make workerforbuilder to be optional, but maybe there are other things you saw that aren't obvious to me?

p12tic (changes requested):
The PR looks good in principle. It's great to see the report generators to be expanded beyond the functionality that was present before the recent reporter API change.

I saw a couple of things to be improved, but otherwise good job!

g-arjones:
> What was the problem that you were seeing?

It was a misinterpretation of the code. Actually, `setupPropertiesKnownBeforeBuildStart` alone won't do what we really need. I was able to make it all work (and to simplify the generator a bit more) by creating a "temporary" [`Build`](https://github.com/buildbot/buildbot/blob/cac10fd1bb9c978fe576c640957a146fb895e4e0/master/buildbot/process/build.py#L44), calling `setupOwnProperties` on it and then converting its properties to a dictionary. I am not sure whether `Build` was intended to be used like that but I don't see a problem with this. What do you think?

> adapt setupPropertiesKnownBeforeBuildStart to our requirements. The obvious one is to make workerforbuilder to be optional

Agreed. I will do that if you believe the "temporary" `Build` is not a problem.

g-arjones:
Updated. Thank you for the feedback.

master/buildbot/process/build.py
```diff
@@ -225,23 +225,32 @@ def setupPropertiesKnownBeforeBuildStarts(props, requests, builder,
...
-            source = self.sources[0]
+            source = sources[0]
             props.setProperty("branch", source.branch, "Build")
             props.setProperty("revision", source.revision, "Build")
             props.setProperty("repository", source.repository, "Build")
             props.setProperty("codebase", source.codebase, "Build")
             props.setProperty("project", source.project, "Build")
 
+    def setupOwnProperties(self):
+        props = self.getProperties()
```
p12tic, line 250: I guess we no longer need this setupOwnProperties() function as it's used in a single place.

master/buildbot/process/build.py
```diff
@@ -225,23 +225,32 @@ def setupPropertiesKnownBeforeBuildStarts(props, requests, builder,
...
+        Build.setupBuildProperties(props, self.requests, self.sources)
+        props.setProperty("buildnumber", self.number, "Build")
```
p12tic, line 252: I think it makes sense to pass the buildnumber as a parameter to setupBuildProperties.

p12tic (changes requested):
Just a couple of nitpicks more and the PR is ready.

One thing I would like to ask is to split the implementation to 3 commits: one for the getBuilderId extraction, another for reusability of build property setup and the third containing the rest of the code.

master/buildbot/process/build.py
g-arjones: Just to make sure I got this one. You want to remove the method completely (thus breaking the API) right?

p12tic: Yes. Buildbot is not exposing its internals as a real API, it's considered an implementation detail. I don't see this function documented and Buildbot has a policy of non-documented == internal detail, thus we can do whatever we want with it.

p12tic: There are a couple of tests that you'll need changing, but it's just a simple function inlining. Though now that the functions themselves are static it may make sense to greatly simplify the test by avoiding a reference to a real build altogether (but you don't need to do that in this PR).

g-arjones:
All done.

p12tic (approved):
Thanks!

PR #5939, opened 2021-03-22
Revert "www: Reduce flickering of buildsummary tooltip during mouseover"

tardyp:
Reverts buildbot/buildbot#5937


p12tic:
I posted this on the merged PR, but I think it's worth duplicating here.

It's true that the flickering is not completely fixed, but It's barely visible and thus a large improvement.

What this PR did fix though (even if by working around the problem) was extremely annoying flickering when the active build bubble is changing size (e.g. in the workers page). So I think this is solving 90% of the problem and we don't really need to revert, a next iteration would be better..

PR #5941, opened 2021-03-22
buildrequest owners do not always exist

### LINKED ISSUE #5940: Regression: Owners column empty for pending build-requests
Owners column is empty for pending build-requests in Buildbot 2.10.2. This used to work properly in Buildbot 1.8.2

This issue can also be noticed at https://buildbot.buildbot.net/#/builders/3
empty_owners_column_pending_buildrequests

tardyp:
use owner property if it is not present

fix #5940

### DIFF
--- a/www/base/src/app/builders/builder/builder.tpl.jade
+++ b/www/base/src/app/builders/builder/builder.tpl.jade
@@ -20,6 +20,7 @@
                 span(title="{{br.submitted_at | dateformat:'LLL'}}")
                   | {{br.submitted_at | timeago }}
               td
+                span(ng-if="br.properties.owners === undefined") {{br.properties.owner[0]}}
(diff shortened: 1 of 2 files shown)


aj062:
Verified that this fixes the issue.

p12tic (approved):

www/base/src/app/builders/builder/builder.tpl.jade
```diff
@@ -20,6 +20,7 @@
                 span(title="{{br.submitted_at | dateformat:'LLL'}}")
                   | {{br.submitted_at | timeago }}
               td
+                span(ng-if="br.properties.owners === undefined") {{br.properties.owner[0]}}
```
p12tic, line 23: I wonder if it makes sense to use "owners" by default and migrate build request to also use that naming long term.

PR #5944, opened 2021-03-24
Make GerritEventLogPoller adhere to debug argument

muks:
It appears that an {{or True}} was left behind in the source code that causes the logs to flood with duplicate change events from the Gerrit event log.

A minor doc tweak was also made.

## Contributor Checklist:


p12tic:
Thanks!

Issue #5906, opened 2021-03-15
Message queue doesn't provide correct message for step finished


aj062:
Message queue doesn't provide correct message for step finished in Buildbot 2.10.1. It used to work in Buildbot 1.8.2.

Below is a code snippet which used to properly print step state_string in Buildbot 1.8.2 and earlier. In Buildbot 2.10.1 (and also in 2.10.2), it returns messages like: "pending", "shell" etc.. For some steps, it prints step "description" instead of "descriptionDone". It seems like this api is invoked by buildbot little too early, before descriptionDone is set.

```class Events(service.BuildbotService):
    @defer.inlineCallbacks
    def startService(self):
        yield service.BuildbotService.startService(self)
        startConsuming = self.master.mq.startConsuming

        self._stepStartedConsumer = yield startConsuming(self.stepStarted, ('steps', None, 'started'))
        self._stepFinishedConsumer = yield startConsuming(self.stepFinished, ('steps', None, 'finished'))

    def stepFinished(self, key, step):
        print('step state_string: {}'.format(step.get('state_string')))
```

aj062:
This was working fine in Buildbot v2.8.4, but stopped working in v2.9.0

cc @tardyp

aj062:
This is a regression from https://github.com/buildbot/buildbot/commit/e2fb0fb3514b99c7e42137a6b97eb6b876a6a97a

Verified that this feature was working fine until this commit.

aj062:
Verified that reverting https://github.com/buildbot/buildbot/commit/e2fb0fb3514b99c7e42137a6b97eb6b876a6a97a fixes this issue.

p12tic:
This is now fixed and released as 2.10.5, 3.0.3 and 3.1.0.

PR #5952, opened 2021-03-26
make sure finished event is sent when the step is truly finished

tardyp:
fix for #5906

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -523,8 +523,6 @@ class BuildStep(results.ResultComputingConfigMixin,
                 self.results = EXCEPTION
                 hidden = False
 
-        yield self.master.data.updates.finishStep(self.stepid, self.results,
-                                                  hidden)
         # perform final clean ups
         success = yield self._cleanup_logs()
         if not success:
@@ -539,6 +537,9 @@ class BuildStep(results.ResultComputingConfigMixin,
             yield sub.finish()
 
         self.releaseLocks()
+
+        yield self.master.data.updates.finishStep(self.stepid, self.results,
+                                                  hidden)
 
         return self.results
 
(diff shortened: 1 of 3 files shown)


master/buildbot/process/buildstep.py
```diff
@@ -511,6 +511,9 @@ def setRenderable(res, attr):
             if self.results != CANCELLED:
                 self.results = EXCEPTION
 
+        self.updateSummary()
```
p12tic, line 514: I think the problem here is that we call finishStep too early. We don't even know the final step results which are set in https://github.com/buildbot/buildbot/pull/5952/files#diff-e77ab9c0beb5448f34c444ea388a12754b60ec7ba820cc6a0550989f37f3bde4R534. So I think it would be better to move finishStep to a time when everything is finished, i.e. after the self.updateSummary.stop().

master/buildbot/process/buildstep.py
```diff
@@ -511,6 +511,9 @@ def setRenderable(res, attr):
...
+        yield self.updateSummary.flush()
```
aj062, line 515: This fails with error: `builtins.AttributeError: 'function' object has no attribute 'flush'`

But this works: `yield self.realUpdateSummary.flush() `

master/buildbot/util/debounce.py
```diff
@@ -77,6 +77,11 @@ def retry(_):
             if queued:
                 self.__call__()
 
+    def flush():
```
aj062, line 80: missing 'self' here

p12tic:
LGTM, I guess we need to update tests.

aj062 (commented):
Look good to me as well. Verified that this change fixes the issue.

tardyp:
@p12tic tests fixed. I think this uncovered a flaky test

p12tic (approved):

aj062 (approved):

PR #5956, opened 2021-03-28
Make GerritStatusPush also work with GitPoller source

muks:
It would previously work with GerritChangeSource and
GerritEventLogPoller only. But using events from these Gerrit specific
changesources can be problematic for some users, and so we support
GitPoller as well.

Reviews are sent back to the Gerrit server for builds with refspecs
matching: refs/changes/n/n/n

## Contributor Checklist:


p12tic (changes requested):
Looks good in principle. This will need a unit test and a news fragment.

PR #5959, opened 2021-03-31
Fix exception in FileDownload when file doesn't exist on master

pocek:
Raising an exception breaks use cases such as FileDownload(haltOnFailure=False).
Also, the behavior deviated unnecessarily from FileUpload. This regression was
made in 5c03c53a8cb245d9b288d2f371e829c545d1a1e3. The test didn't catch the
behavior change because it wasn't there yet.

The test, introduced in 86dfd8269b84ad0beb57dfeefbcd9b4adf18818e, was partly
broken due to misplaced and wrong expectLogfile call. This is now fixed too.


p12tic (changes requested):
Looks good, could you add a news fragment? Thanks!

p12tic (approved):

PR #5960, opened 2021-04-04
WWW: Add Basic authentication support for AvatarGitHub

lephilousophe:
On GitHub, organizations cannot create personal access tokens to increase the rate limits.
This commit adds ability to login to GitHub using an OAuth client ID/client secret pair.
This one can be created by organizations (and it is to authenticate users).

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/www/test_avatar.py
+++ b/master/buildbot/test/unit/www/test_avatar.py
@@ -16,6 +16,7 @@
 from twisted.internet import defer
 from twisted.trial import unittest
 
+from buildbot import config
 from buildbot.test.fake import httpclientservice as fakehttpclientservice
 from buildbot.test.util import www
 from buildbot.test.util.misc import TestReactorMixin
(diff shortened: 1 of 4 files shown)


master/buildbot/www/avatar.py
```diff
@@ -51,6 +54,13 @@ def __init__(self,
         if github_api_endpoint is None:
             self.github_api_endpoint = self.DEFAULT_GITHUB_API_URL
         self.token = token
+        self.client_creds = None
```
p12tic, line 57: I think we should throw if only one of client_id and client_secret is specified and also if they are used together with token.

master/buildbot/test/unit/www/test_avatar.py
```diff
@@ -663,3 +663,44 @@ def test_email_error(self):
...
+        yield self.master.stopService()
+
+    @defer.inlineCallbacks
+    def test_username(self):
+        username_search_endpoint = '/users/defunkt'
+        self._http.expect('get', username_search_endpoint,
+            content_json=github_username_search_reply,
+            headers={'Accept': 'application/vnd.github.v3+json'})
+        res = yield self.render_resource(self.rsrc, b'/?username=defunkt')
+        self.assertEqual(res, dict(redirected=b'https://avatars3.githubusercontent.com/'
```
p12tic, line 705: Let's use dict literal here.

master/buildbot/test/unit/www/test_avatar.py
```diff
@@ -663,3 +663,44 @@ def test_email_error(self):
...
+
+    @defer.inlineCallbacks
+    def tearDown(self):
+        yield self.master.stopService()
+
+    @defer.inlineCallbacks
+    def test_username(self):
+        username_search_endpoint = '/users/defunkt'
+        self._http.expect('get', username_search_endpoint,
+            content_json=github_username_search_reply,
```
p12tic, line 702: Could we align code to opening parenthesis?

master/buildbot/www/avatar.py
lephilousophe: Okay, I can do it.
But I don't really know how to add this to tests in a proper way.
Could you give me some guidance on this?

master/buildbot/test/unit/www/test_avatar.py
```diff
@@ -663,3 +663,44 @@ def test_email_error(self):
...
+
+
+class GitHubAvatarBasicAuth(TestReactorMixin, www.WwwTestMixin, unittest.TestCase):
+
+    @defer.inlineCallbacks
+    def setUp(self):
+        self.setUpTestReactor()
+
+        master = self.make_master(
+            url='http://a/b/', auth=auth.NoAuth(),
```
p12tic, line 675: Can we align to the oening parenthesis? If avatar.AvatarGithub does not fit we can extract it to a separate variable.

p12tic (changes requested):
Minor comments, looks good in principle. Thanks!

master/buildbot/test/unit/www/test_avatar.py
lephilousophe: That's a copy paste from the classes above (test_gravatar, and GitHubAvatar).
Should I change these as well?

p12tic: No, let's move to the literal syntax gradually. Changing for the sake of changing will make git blame much less useful.

master/buildbot/www/avatar.py
p12tic: The error is in configuration, so it's best to use `config.error` to raise it. As for testing, check out `with self.assertRaises(ConfigErrors):` in other tests. You will most likely just need to call the constructor in the `with` body.

master/buildbot/test/unit/www/test_avatar.py
lephilousophe: Done

master/buildbot/www/avatar.py
lephilousophe: I tried something. Let me know if it fits.

master/buildbot/test/unit/www/test_avatar.py
lephilousophe: Done.

master/buildbot/test/unit/www/test_avatar.py
lephilousophe: Done.

master/buildbot/www/avatar.py
p12tic: LGTM, thanks

master/buildbot/www/avatar.py
```diff
@@ -51,6 +55,17 @@ def __init__(self,
...
+        if bool(client_id) != bool(client_secret):
+            config.error('client_id and client_secret must be both provided or none')
+        if client_id:
+            if token:
+                config.error('client_id and client_secret must not be provided when token is')
+            self.client_creds = base64.b64encode(b':'.join(
+                cred.encode('utf-8') for cred in (client_id, client_secret)
+            )).decode('ascii')
+        self.client_id = client_id
```
p12tic, line 67: I think we don't need to store the client_id and client_secret as members because we don't use them.

Sorry for asking for another iteration of the PR.

lephilousophe: Indeed, it was a leftover of a previous version of my change.

master/buildbot/www/avatar.py
```diff
@@ -22,13 +23,14 @@
 from twisted.internet import defer
 from twisted.python import log
 
-from buildbot.util import config
+from buildbot import config
+from buildbot.util.config import ConfiguredMixin
```
p12tic, line 27: The isort test now fails.

p12tic (approved):
LGTM, just the isort test failure needs to be fixed.

lephilousophe:
Meh. I didn't tested for it.
You should really have some make pr-test rule which runs trial, flake8, pylint, isort and everything needed. :D
What I used until now is `trial && make flake8 && make pylint`.

p12tic:
>  You should really have some make pr-test rule which runs trial, flake8, pylint, isort and everything needed. :D

Agreed. There was one once upon the time, but it bitrot over the years.

Issue #5948, opened 2021-03-25
worker version is shown incorrectly in the UI 


aj062:
worker version is shown incorrectly in the UI, e.g.: on workers page, or individual worker pages.
e.g.: 
http://buildbot.buildbot.net/#/workers
http://buildbot.buildbot.net/#/workers/5
<img width="1187" alt="worker_version_incorrect" src="https://user-images.githubusercontent.com/3098702/112486433-3de22e80-8d52-11eb-83c4-3046a9e42c58.png">

aj062:
Seems like this was working fine in buildbot-worker v1.5.0, but stopped working in v1.6.0

p12tic:
@tardyp Do you remember if this was an intentional change? I always assumed that Buildbot switched to date-based version for the worker and didn't investigate :))

pmisik:
This looks like duplicate of #4464
The problem is missing VERSION file in whl package.

p12tic:
Fixed in #5977, will be released as v3.1.1.

Issue #4464, opened 2018-12-04
Wrong worker version in workers page


BishopGIS:
Workers with version **1.6** report start date instead version. In screenshot Windows worker with version 1.3 reports correct version and fresh workers - start date.

<img width="201" alt="2018-12-04 22 00 29" src="https://user-images.githubusercontent.com/1708122/49466164-61df7600-f810-11e8-8e74-d127cf7d4823.png">

seankelly:
Nitpick, it reports the modified time of the `__init__.py` file. This only happens if all other version detection methods fail.

seankelly:
The `VERSION` file was present in 1.5.0 and is missing in 1.6.0.

vibbo:
Yes, this can be quite annoying.
The culprit commit is this one: https://github.com/buildbot/buildbot/commit/b339c0ede7876d301415b7b875e33cfc496448c1
There is no longer a "VERSION" file in *.whl package in https://pypi.org/project/buildbot-worker/1.6.0/#files and it was present in version 1.5.0. (Note: It is still present in *.tar.gz though).

As I am unfamiliar with the release process, would someone from maintainers fix this problem? Reverting the deleted code was enough to fix this problem, but I am not entirely sure what was the purpose of deleting it in the first place...

p12tic:
Fixed in #5977, will be released as v3.1.1.

PR #5977, opened 2021-04-05
Fix missing VERSION file in buildbot_worker wheel package.

pmisik:
Bug #5948, #4464

## Contributor Checklist:

### DIFF
--- a/worker/setup.py
+++ b/worker/setup.py
@@ -33,10 +33,28 @@ try:
     import setuptools
     from setuptools import setup
     from setuptools.command.sdist import sdist
+    from distutils.command.install_data import install_data
 except ImportError:
     setuptools = None
     from distutils.command.sdist import sdist
     from distutils.core import setup
+
+
+class our_install_data(install_data):
+
+    def finalize_options(self):
+        self.set_undefined_options('install',
+                                   ('install_lib', 'install_dir'),
+                                   )
+        install_data.finalize_options(self)
+
+    def run(self):
+        install_data.run(self)
+        # ensure there's a buildbot_worker/VERSION file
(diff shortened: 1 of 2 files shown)


tardyp:
I am not sure why this worked, and do not work anymore.

Should we include the VERSION file in the MANIFEST.in?

pmisik:
> 
> 
> I am not sure why this worked, and do not work anymore.
> 
> Should we include the VERSION file in the MANIFEST.in?

I think problem is that VERSION file is generated during packing.
`our_sdist` is called while sdist is packed but not while `bdist_wheel` command is running.

It seems revert of b339c0ede7876d301415b7b875e33cfc496448c1 solves the problem.

tardyp:
what worries me is that this solution is different from what is in the other packages.

We should have similar logic everywhere.

I think that revert  b339c0e leads to that actually as I see that master package has same package tweak that wasn't removed
install_data method, is probably the method that works the best. I was wrong in trying to remove it.

@p12tic  any opinion?

worker/setup.py
```diff
@@ -33,12 +33,30 @@
...
     from setuptools.command.sdist import sdist
+    from distutils.command.install_data import install_data
 except ImportError:
     setuptools = None
     from distutils.command.sdist import sdist
     from distutils.core import setup
 
 
+class our_install_data(install_data):
+
```
tardyp, line 44: lgtm

tardyp (approved):

p12tic:
I think that it makes sense to keep what master is doing. So this PR is in good direction.

p12tic (approved):

PR #5978, opened 2021-04-08
graphql basic request parsing and error management

tardyp:
Some basic infrastructure.

we had api/v2

graphql will be at api/v3

Goal is to let the user enable it via `c['www']['graphql'] = {'debug'=True}`

we make extra effort to not require graphql-core unless enabled

### DIFF
--- a/master/buildbot/test/unit/www/test_graphql.py
+++ b/master/buildbot/test/unit/www/test_graphql.py
@@ -61,6 +61,7 @@ class V3RootResource(TestReactorMixin, www.WwwTestMixin, unittest.TestCase):
         yield self.render_resource(self.rsrc, b"/", method=b"PATCH")
         self.assertSimpleError("invalid HTTP method", 400)
 
+    # https://graphql.org/learn/serving-over-http/#get-request
     @defer.inlineCallbacks
     def test_get_query(self):
         yield self.render_resource(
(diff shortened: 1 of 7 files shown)


master/buildbot/test/unit/www/test_graphql.py
```diff
@@ -0,0 +1,108 @@
...
+        yield self.render_resource(self.rsrc, b"/", method=b"PATCH")
+        self.assertSimpleError("invalid HTTP method", 400)
+
+    @defer.inlineCallbacks
+    def test_get_query(self):
+        yield self.render_resource(
+            self.rsrc,
+            b"/?query={tests{id}}",
+        )
+        # TODO verify results
```
p12tic, line 70: WIP? Should be easy to capture what the current call returns and just assert that this does not change.

master/buildbot/www/graphql.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def reconfigResource(self, new_config):
+        # @todo v2 has cross origin support, which might need to be factorized
+        self.gql_config = new_config.www.get("graphql")
+        self.debug = True
+        self.graphql = None
+        if self.gql_config is not None:
+            try:
+                import graphql
+            except ImportError:
```
p12tic, line 43: Needs `pragma: no cover`.

master/buildbot/www/graphql.py
```diff
@@ -0,0 +1,93 @@
...
+            raise Error(400, [e.formatted for e in errors])
+
+    @defer.inlineCallbacks
+    def asyncRender(self, request):
+        if self.graphql is None:
+            raise Error(501, "graphql not enabled")
+
+        # graphql accepts its query either in post data or get's query
+        if request.method == b"POST":
+            query = request.content.read()
```
p12tic, line 82: Could we add a test for this?

master/buildbot/www/graphql.py
```diff
@@ -0,0 +1,93 @@
...
+        elif request.method in (b"GET"):
+            if b"query" not in request.args:
+                raise Error(400, b"GET request must contain a 'query' parameter")
```
p12tic, line 85: Could we add a test for this?

p12tic (changes requested):
Overall good, I just wrote a couple of nitpicks and it would be great to have a couple of missing trivial tests in the PR.

master/buildbot/test/unit/www/test_graphql.py
tardyp: yes. for the moment, there is no output, just error management.

tardyp: the next PR will be the plumbing with data api. I guess it is easier for review to just land that part.

p12tic (approved):

PR #5980, opened 2021-04-10
Graphql plumbing with data api

tardyp:
early demo of the graphql plumbery based on #5978 which should be good to go now

One of the surprise is that we need to use the asyncioreactor to integrate with graphql-core...

### DIFF
--- a/master/buildbot/test/util/misc.py
+++ b/master/buildbot/test/util/misc.py
@@ -24,7 +24,7 @@ from twisted.python import threadpool
 from twisted.trial.unittest import TestCase
 
 import buildbot
-from buildbot.asyncio import TwistedLoop
+from buildbot.asyncio import AsyncIOLoopWithTwisted
 from buildbot.process.buildstep import BuildStep
 from buildbot.test.fake.reactor import NonThreadPool
 from buildbot.test.fake.reactor import TestReactor
@@ -93,7 +93,7 @@ class TestReactorMixin:
         self.addCleanup(self.reactor.stop)
 
         if use_asyncio:
-            self.asyncio_loop = TwistedLoop(self.reactor)
+            self.asyncio_loop = AsyncIOLoopWithTwisted(self.reactor)
             asyncio.set_event_loop(self.asyncio_loop)
             self.asyncio_loop.start()
 
--- a/master/buildbot/www/graphql.py
+++ b/master/buildbot/www/graphql.py
@@ -33,19 +33,11 @@ class V3RootResource(resource.Resource):
 
     def reconfigResource(self, new_config):
         # @todo v2 has cross origin support, which might need to be factorized
-        self.gql_config = new_config.www.get("graphql")
+        graphql_config = new_config.www.get("graphql")
         self.debug = True
         self.graphql = None
-        if self.gql_config is not None:
-            try:
-                import graphql
-            except ImportError:  # pragma: no cover
-                raise ImportError(
-                    "graphql is enabled but 'graphql-core' is not installed"
-                )
-            self.graphql = graphql
-            self.debug = self.gql_config.get("debug")
-            self.schema = graphql.build_schema(self.master.data.get_graphql_schema())
+        if graphql_config is not None:
+            self.graphql = True
 
     def render(self, request):
         def writeError(msg, errcode=400):
@@ -64,12 +56,6 @@ class V3RootResource(resource.Resource):
             request.finish()
 
         return self.asyncRenderHelper(request, self.asyncRender, writeError)
-
-    def renderQuery(self, query):
-        query = self.graphql.parse(query)
-        errors = self.graphql.validate(self.schema, query)
-        if errors:
-            raise Error(400, [e.formatted for e in errors])
 
     @defer.inlineCallbacks
     def asyncRender(self, request):
@@ -99,8 +85,13 @@ class V3RootResource(resource.Resource):
         else:
             raise Error(400, b"invalid HTTP method")
 
-        res = yield self.renderQuery(query)
-        return res
+        res = yield self.master.graphql.query(query)
+        errors = None
+        if res.errors:
+            errors = [e.formatted for e in res.errors]
+        request.setHeader(b"content-type", b"application/json; charset=utf-8")
+        data = json.dumps({"data": res.data, "errors": errors}).encode()
+        request.write(data)
 
 
 RestRootResource.addApiVersion(3, V3RootResource)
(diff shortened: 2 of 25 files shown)


master/buildbot/test/util/misc.py
```diff
@@ -73,7 +73,13 @@ class TestReactorMixin:
     Mix this in to get TestReactor as self.reactor which is correctly cleaned up
     at the end
     """
-    def setUpTestReactor(self):
+    def setUpTestReactor(self, use_asyncio=False):
+        if use_asyncio:
+            import asyncio
+            import twisted.internet
+            from twisted.internet import asyncioreactor
+            twisted.internet.reactor = asyncioreactor.AsyncioSelectorReactor(asyncio.get_event_loop())
```
p12tic, line 81: Shouldn't the code in question use `self.reactor` and thus we should patch that one appropriately? Would it be possible to create fake asyncio reactor?

tardyp: I tried a few options to do that, but the asyncio event loop is much more complex that the twisted one. There are tons of stuff to implement, especially the task manager which schedules the coroutine looks very complex to do.

this post talks a bit about that, and suggests that nobody has yet written an event loop using twisted semantics. 
https://meejah.ca/blog/python3-twisted-and-asyncio

We only have the opposite, i.e twisted loop written in asyncio

I also tried to convert graphql to twisted, but it is also too much hack.

The main drawback of twisted here is that the coroutine support is not equivalent.
you cannot await another coroutine you have to wrap it with ensure-defered.

tardyp: @meejah Do you have some advice on this point? In Buildbot, our test suite is quite fast because we fake the reactor on most unit tests. 

As we start to integrate with asyncio libraries, we must install the asyncioreactor, which takes 60ms per test to initialise( nearly 60x decrease in perf) 

graphql-core merely only uses coroutines and asyncio.gather. Do you think it is worth to try and implement a fake asyncio loop with only support for those?

tardyp: So this wasn't very obvious at first glance, but the asyncio task scheduler only requires loop.call_soon to be implemented. That was quite easy to translate into twisted callLater(0, ...)

meejah: Okay so graphql-core is written against asyncio .. so you need "something" that lets you inter-op. I think if you want Twisted as well, the best case is still `asyncioreactor`. I don't know what the state-of-the-art for asyncio testing is .. I guess you mean implementing something like `MemoryReactor` / `iosim.IOPump` (but only with the APIs required for graphql to function)...?

Or are you "faking out the reactor" some other way for the Twisted tests?

tardyp: actually graphql-core does not handle the IO part of graphql. It just handle the parsing and resolving. For the latter part there is a little bit of coroutines and asyncio.gather.

The problem of asyncioreactor is that I can see that there seems to be some limitation on windows, which requires to use the non default loop. Windows users have enough problems, I would like to avoid more.

I eventually managed to make the thing work with this solution:
https://github.com/buildbot/buildbot/pull/5980/commits/60c6fd7de28d58986d032eb0e7aeb4cb24140226

This involves implementing a custom asyncio event loop which only implements the call_soon api, which is actually the only thing needed for coroutine scheduling to work.

meejah: Ahhh, right .. and `asyncio` ties all futures/tasks/etc to an event-loop unlike `Deferred` .. this solution looks decent to me (although I can't claim to have looked in detail).

master/buildbot/test/util/misc.py
```diff
@@ -89,6 +92,11 @@ def deferToThread(f, *args, **kwargs):
         self.addCleanup(_setReactor, None)
         self.addCleanup(self.reactor.stop)
 
+        if use_asyncio:
+            self.loop = TwistedLoop(self.reactor)
```
p12tic, line 96: I would prefer if we used a less generic name. In the case of reactor it's obvious, whole Buildbot is created using Twisted. `loop` can mean anything on the other hand. Anything like `asyncio_loop`, `event_loop` or something like that would be fine.

p12tic (approved):

p12tic:
Looks good, just needs a rebase on top of latest master

p12tic:
I rebased and pushed myself to make sure I didn't miss anything in the review.

master/buildbot/www/graphql.py
```diff
@@ -65,12 +68,36 @@ def writeError(msg, errcode=400):
...
         return self.asyncRenderHelper(request, self.asyncRender, writeError)
 
-    def renderQuery(self, query):
+    async def renderQuery(self, query):
         query = self.graphql.parse(query)
         errors = self.graphql.validate(self.schema, query)
         if errors:
             raise Error(400, [e.formatted for e in errors])
 
+        async def field_resolver(parent, resolve_info, **args):
```
tardyp, line 77: I am still not very happy with this implementation.
I think it is still a bit too hacky. I have not yet found where to manage this, but probably this needs to be implemented in an independent module

sphet:
I wonder if someone can articulate what graphql offers, specifically with respect toe api/v3

tardyp:
graphql is supposed to be an api on top of the rest api, so we will keep api/v2 and api/v3 indefinitely.

This is the case for lots of project who add some graphql feature to their app (github gitlab are examples)

The main interest for me is for the UI latency. graphql allows to define queries that would request the whole data needed to display a page in our request. while in rest api you need dozens. This is known has N+1 problem.

We will probably still have N+1 problem at the db side, but there is much less latency at this position of the stack, and we can later optimize the resolvers to optimize up to the db.

www/rebase/src/pages/home.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { useQuery, useSubscription } from "../gqless";
+
+
+export function Home() {
```
tardyp, line 5: @p12tic all this to eventually go to this demo...

note how the api is generated automatically from schema.
and the subscribe is juste like queries. You can decide to either get automatic updates or not. just by using query or sub hooks.

tardyp:
@p12tic this starts to be a little bit too large.
Can we get a final review + merge?

I'll resubmit the UI demo in another PR.

p12tic (approved):
I would prefer to replace `gql` to `graphql` in the PR as especially for new people seeing `gql` will be very confusing and disorienting. Otherwise no complaints, thanks.

tardyp:
Yeah I don't like the prefix and Mixin design that much because it is not
really a Mixin, just a way of splitting code. I may refactor it later
if/when I find a better idea...

Issue #5991, opened 2021-04-20
ws_ping_interval is not added in the set of allowed keys for ww


pablogsal:
Checking the configuration after adding this parameter throws:

```
Configuration Errors:
  unknown www configuration parameter(s) ws_ping_interval
make: *** [Makefile:34: check] Error 1
```

The parameter should be updated in https://github.com/buildbot/buildbot/blob/c1c4812cc67265798dd597b06f113943dc154112/master/buildbot/config.py#L717

vstinner:
ws_ping_interval parameter was added to Buildbot 3.0.0 ( 2021-03-08 ) in the issue #4078.

vstinner:
The buildbot server of the Python project has timeout issues: https://bugs.python.org/issue41701 Using ws_ping_interval is an attempt to workaround this issue.

pablogsal:
@p12tic Would it be possible to make a bugfix release or similar so we can leverage the fix in the CPython buildbot server?

p12tic:
@pablogsal Yes, there will a bugfix release, maybe today.

p12tic:
@pablogsal 3.1.1 has been just released.

pablogsal:
> @pablogsal 3.1.1 has been just released.

Fantastic, thanks a lot! ❤️

PR #5993, opened 2021-04-24
config: Allow ws_ping_interval in the configuration.

p12tic:
Fixes https://github.com/buildbot/buildbot/issues/5991.


vstinner:
Thanks for the fix :-)

PR #5994, opened 2021-04-25
Asyncioloop

tardyp:
asyncio_loop extracted from the graphql WIP.

I think the graphql wip is not yet ready so I put this as separate PR as I think that one is clean enough


p12tic (approved):

master/buildbot/asyncio.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def as_deferred(f):
+    return asyncio.get_event_loop().as_deferred(f)
+
+
+if sys.version_info[:2] >= (3, 7):
+    def make_handle(callback, args, loop, context):
+        return events.Handle(callback, args, loop, context)
+else:
```
p12tic, line 45: Since there's a flake8 failure anyway, I think it makes sense to mark this else with `pragma: no cover`

p12tic: (though I see that we should run coverage on 3.6 too, it's interesting why this wasn't covered.

tardyp: its because I didn't put the ci/ prefix before coverage. It is required so that codecov wait for the end of the test before sending results

PR #5995, opened 2021-04-25
Refactor BitbucketStatusPush.sendMessage

nlamprian:
- Move status and request into their own methods
- Add check for empty sourcestamp repository

I'm extending the class and I thought I would move the status and request out of sendMessage, so that the module variables don't have to be imported.

## Contributor Checklist:


master/buildbot/reporters/bitbucket.py
```diff
@@ -76,11 +76,20 @@ def reconfigService(self, oauth_key, oauth_secret, base_url=_BASE_URL, oauth_url
     def _create_default_generators(self):
         return [BuildStartEndStatusGenerator()]
 
+    def _get_status(self, is_complete, results):
+        if is_complete:
```
tardyp, line 80: If you want to use _get_status and _get_request in your subclass, then you should put them as public members (getStatus() getRequest(). 

Then the maintainer will avoid changing them between major releases, which will help you upgrade without fear

p12tic: Exposing such methods ends up making more work when we refactor, at least when it's private it's on the user to keep up with our internal API changes.

I would be interested in what the use case of exposing these methods is. Maybe we could extend the current class with additional functionality?

nlamprian: I would very much be interested in adding more flexibility to the class. Actually, I don't need to overwrite sendMessage. I only seek control over the key and the name of the build status. Right now, the key is based solely on the builder name, which means rebuilds overwrite old statuses. As a first step, we want to change `build['builder']['name']}` to `build['builder']['name']/build['buildid']`, and then, I want to have finer control over the key, and particularly, I want to set it based on a property.

So, I would propose the following:
```
    def getStatusKey(self, build):
        return build['builder']['name']

    def getStatusName(self, build):
        return build['builder']['name']
```

p12tic: For the key, would it make sense to add `status_key` configuration argument to the reporter and support renderables for it? Both `buildername and `buildnumber` are available as properties, so you would be able to construct your wanted value using `Interpolate` or even a completely custom function.

 For status name we could do the same.
 
 For the description, would it make sense to reuse the `subject` of the build report? (see how return value of `BuildStatusGeneratorMixin.build_message` is constructed). This way you could pass the description via the report generators with custom subject in their message formatters (see `subject` argument of constructor of e.g. `MessageFormatterRenderable`) and it would work like the rest of reporters will work eventually.

nlamprian: I had exactly the same idea about the description, but as I was configuring the reporter instance, I noticed the result looked bloated, so I went with subclassing in my own implementation. It's a lot simpler. But, if it is to be done right, I agree with everything you said above. I'd be happy to attempt to implement these changes.

nlamprian: Done. It was easier than I expected. If you are pleased with the changes, I'll update the tests and docs next.

For the description, although I don't understand how the formatter gets access to more values than what it's returned by create_context_for_build, the use of the report subject works fine.

For the key and name, the use of the properties still does not meet the requirements for my use case. More specifically, I have events that trigger one build, and then that build triggers a second one. I want both builds to push to the same build status. The way I have it set up in my custom reporter is by checking for a parent build, and if that exists, I set the status key based on the parent build instead of the current one. This means I need access to the whole build dictionary, not just the properties. Can this be accommodated?

p12tic (approved):

p12tic (changes requested):
Looks good in principle, please fix the tests.

p12tic (approved):

PR #5996, opened 2021-04-26
Add check for double call to Log.finish

nlamprian:
I got the following exception:

```
2021-04-26 22:55:04+0200 [-] when trying to finish a log
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 501, in errback
	    self._startRunCallbacks(fail)
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 568, in _startRunCallbacks
	    self._runCallbacks()
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 1475, in gotResult
	    _inlineCallbacks(r, g, status)
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
	  File "/usr/lib/python3/dist-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
	  File "/usr/lib/python3/dist-packages/buildbot/process/log.py", line 203, in finish
	    yield super().finish()
	  File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
	    result = g.send(result)
	  File "/usr/lib/python3/dist-packages/buildbot/process/log.py", line 115, in finish
	    self._had_errors = len(self.subPoint.pop_exceptions()) > 0
	builtins.TypeError: object of type 'NoneType' has no len()
```

while forgetting to yield log.finish() in a step

```
log = yield self.getLog("log")
log.finish()
output = cmd.stdout
```

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/log.py
+++ b/master/buildbot/process/log.py
@@ -33,6 +33,7 @@ class Log:
 
         self.subPoint = util.subscription.SubscriptionPoint("%r log" % (name,))
         self.subscriptions = {}
+        self._finishing = False
         self.finished = False
         self.finishWaiters = []
         self._had_errors = False
@@ -97,7 +98,9 @@ class Log:
 
     @defer.inlineCallbacks
     def finish(self):
+        assert not self._finishing, "Did you maybe forget to yield the method?"
         assert not self.finished
+        self._finishing = True
 
         def fToRun():
             self.finished = True
@@ -119,6 +122,7 @@ class Log:
         d = self.master.data.updates.compressLog(self.logid)
         d.addErrback(
             log.err, "while compressing log %d (ignored)" % self.logid)
+        self._finishing = False
 
 
 class PlainLog(Log):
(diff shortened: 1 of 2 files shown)


p12tic:
The code is correct, we don't want to call `pop_exceptions()` twice. The root cause here is probably that by not yielding `log.finish()`, the function may block in `yield self.lock.run(fToRun)` and `self.finished` is still False. As a result when step goes to close all non-finished logs in `_cleanup_logs()` it will call the `finished()` function a second time on the same log, which will eventually call `pop_exceptions()` twice.

I think not waiting for `finish()` to complete is a user error because it may result in part of the log data missing. So a better solution would be to add something like `self._finishing` member, set it on start of `finish()` to True and assert that it's `False` at the very beginning of the `finish()` function.

nlamprian:
You are absolutely right. I made the changes you suggested.

p12tic (approved):

p12tic:
Thanks a lot!

PR #5999, opened 2021-04-28
Implement worker HTTP tunneling

kulikjak:
We recently needed our worker to connect through an HTTP proxy, and thus I wrote the following patch that implements that. Since I don't know twisted much, I based it largely on twisteds' own `_WrappingFactory` and `_WrappingProtocol`.

I am not expecting it to be accepted in the current state; I am making this PR to find out whether you are interested in such functionality. If so, I can improve it such that it follows all the guidelines and your suggestions.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/integration/test_worker_proxy.py
+++ b/master/buildbot/test/integration/test_worker_proxy.py
@@ -88,7 +88,14 @@ def run_proxy(queue):
     write_to_log("run_proxy\n")
 
     try:
-        loop = asyncio.get_event_loop()
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            # We can get RuntimeError due to current thread being not main thread on Python 3.8.
+            # It's not clear why that happens, so work around it.
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+
         coro = asyncio.start_server(handle_client, host="127.0.0.1")
         server = loop.run_until_complete(coro)
 
--- a/master/buildbot/test/integration/test_worker_proxy.py
+++ b/master/buildbot/test/integration/test_worker_proxy.py
@@ -23,13 +23,13 @@ from twisted.internet import defer
 
 from buildbot.test.util.integration import RunMasterBase
 
-from .interop.test_commandmixin import CommandMixinMasterPB
-from .interop.test_compositestepmixin import CompositeStepMixinMasterPb
-from .interop.test_integration_secrets import SecretsConfigPB
-from .interop.test_interruptcommand import InterruptCommandPb
-from .interop.test_setpropertyfromcommand import SetPropertyFromCommandPB
-from .interop.test_transfer import TransferStepsMasterPb
-from .interop.test_worker_reconnect import WorkerReconnect
+from .interop import test_commandmixin
+from .interop import test_compositestepmixin
+from .interop import test_integration_secrets
+from .interop import test_interruptcommand
+from .interop import test_setpropertyfromcommand
+from .interop import test_transfer
+from .interop import test_worker_reconnect
 
 # This integration test puts HTTP proxy in between the master and worker.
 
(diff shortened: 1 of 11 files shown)


tardyp:
Hello,

I like this feature a lot. We had the plan to switch to a http/ws protocol at some point, but nobody got some opportunity to work on it yet.
I think this is a good way to accomodate those security principle of proxying everything.

As this will probably be a niche feature we need it to have it tested properly.

I would suggest to make end to end integration tests, maybe by reusing those we already have in:

https://github.com/buildbot/buildbot/blob/graphql2/master/buildbot/test/integration/interop/__init__.py#L19

kulikjak:
Nice! I will add some tests then and other things necessary. One thing I am not sure about - should I leave the proxy settings as environment variables (considering that, as you said, it's a pretty niche feature), or should it be set via the config file?

tardyp:
I think the cleanest way is to have that configuration done in the buildbot.tac.

You can add an extra argument to Worker class that can be configured straight away in the buildbot.tac.

Rather than configuring the 3 parameters, I would directly ask the connection string.
if it is not None, then it will setup the proxy, if it is None, then nothing change from before.

Then in your buildbot.tac, as it is just a python script, you can build the connection string from any number of environment variable.

kulikjak:
I reworked the PR based on your suggestions and added more docstrings. I will dive into the test suite now.

tardyp:
Hello,
Even if now, the tests are fixed, we need *new* tests for this feature.
So that we can keep the coverage good, and we don't mistakingly brake the feature later.

I don't think we need many unit tests, but what we need are e2e tests (like I suggested earlier)

https://github.com/buildbot/buildbot/blob/graphql2/master/buildbot/test/integration/interop/__init__.py#L19

kulikjak:
Sorry, I wasn't able to finish this yet. I wanted to rework the patch a bit as it currently doesn't work with HTTPS tunnels, but I didn't have time for that yet. But it's still on my list.

kulikjak:
I managed to add some working tests for the tunneling functionality.

I hope I understand your suggestions correctly; I am happy to change/rewrite it to better conform to the project. Now, all tests from the _interop_ directory are imported and again executed with the tunnel in use. I tested this with locally running [tinyproxy](http://tinyproxy.github.io/), and all tests pass.

I guess it would be best to leverage native twisted Proxy, but I wasn't yet able to set it such that it works...

tardyp:
Hello, thanks for updating this!

Now we need to make those tests run in the CI or it is useless.
I think the easiest approach is to leverage github actions service and to add a tinyproxy docker service there (or squid or whatever)

We do it already for databases, you can see it here:
https://github.com/buildbot/buildbot/blob/master/.github/workflows/ci.yml#L52

kulikjak:
Thanks, I could have thought of that. I am not very familiar with Github Actions, but I tested this a little in a private repo and it should work.

kulikjak:
FYI: I tried adding the proxy (in my forked repo so that I don't spam it here), and I was not successful yet. The main problem is that running a proxy server in a docker container means that it's in a separate network, and thus **localhost** used in addresses doesn't work anymore (I can get to the proxy server thanks to mapped port, but it cannot connect back to the master).

I was trying to force the use of hosts' network (with `--network=host`), but it seems that GitHub Actions cannot do that yet...

tardyp:
Hello,

Indeed the service container cannot go back to build container. So we are back at spawning a process.

I did prepare an asyncio http proxy
https://gist.github.com/tardyp/f95c67de3f54614a453a745b78e4b116
 
I think you can spawn that from the test and it should work.

kulikjak:
Hi, thanks for the asyncio proxy.

I managed to wire it directly into the test script and start it in a thread, but I can split it into separate subprocess. Also, while I managed to stop the proxy with a signal and raised `KeyboardInterrupt`, there might very well be a better way to stop it.

master/buildbot/test/integration/test_worker_proxy.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class RunMasterBehindProxy(RunMasterBase):
+
+    def setUp(self):
+        self.proxy_process = multiprocessing.Process(target=run_proxy)
+        self.proxy_process.start()
+
+    def tearDown(self):
+        self.proxy_process.terminate()
```
tardyp, line 99: killing in tearDown looks good for me

master/buildbot/test/integration/test_worker_proxy.py
```diff
@@ -0,0 +1,136 @@
...
+
+from .interop.test_commandmixin import CommandMixinMasterPB
+from .interop.test_compositestepmixin import CompositeStepMixinMasterPb
+from .interop.test_integration_secrets import SecretsConfigPB
+from .interop.test_interruptcommand import InterruptCommandPb
+from .interop.test_setpropertyfromcommand import SetPropertyFromCommandPB
+from .interop.test_transfer import TransferStepsMasterPb
+from .interop.test_worker_reconnect import WorkerReconnect
+
+# This integration test puts HTTP proxy in between the master and worker. You
```
tardyp, line 33: well, this comment is obsolete

tardyp (approved):
cool!

master/buildbot/test/integration/test_worker_proxy.py
kulikjak: Ah yes. I removed part of this obsolete comment and left the other one... It's fixed now.

kulikjak:
For some yet unknown reason, Python 3.7 is failing on AppVeyor (I presume that 3.8 failure is not related to this change). I am trying to dump test prints, but I cannot do so on Windows yet - I guess I will try it on windows...

SaintPatricksCoin (approved):
worker/buildbot_worker/tunnel.py
worker/buildbot_worker/test/unit/test_scripts_create_worker.py
worker/buildbot_worker/scripts/runner.py
worker/buildbot_worker/scripts/create_worker.py
master/buildbot/test/util/integration.py

kulikjak:
Thank you both @tardyp and @p12tic for all the comments and for improving/fixing the test suite!

Issue #5933, opened 2021-03-19
buildbot try fails for files containing non utf-8 characters


aj062:
buildbot try fails for files containing non utf-8 characters, i.e. different encoding.

e.g.:
```
[~]$buildbot try --connect=pb --master=master-name:1234 --username=buildbot_try --passwd=test --diff=testfile.patch --repository=''
using 'pb' connect method
Unhandled Error
Traceback (most recent call last):
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/base.py", line 913, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/usr/local/lib/python3.6/site-packages/buildbot/util/eventual.py", line 48, in _turn
    cb(*args, **kwargs)
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/defer.py", line 460, in callback
    self._startRunCallbacks(result)
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/defer.py", line 568, in _startRunCallbacks
    self._runCallbacks()
--- <exception caught here> ---
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/usr/local/lib/python3.6/site-packages/buildbot/clients/tryclient.py", line 937, in trapSystemExit
    why.trap(SystemExit)
  File "/usr/local/lib64/python3.6/site-packages/twisted/python/failure.py", line 460, in trap
    self.raiseException()
  File "/usr/local/lib64/python3.6/site-packages/twisted/python/failure.py", line 488, in raiseException
    raise self.value.with_traceback(self.tb)
  File "/usr/local/lib64/python3.6/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/usr/local/lib/python3.6/site-packages/buildbot/clients/tryclient.py", line 917, in <lambda>
    d.addCallback(lambda res: self.createJob())
  File "/usr/local/lib/python3.6/site-packages/buildbot/clients/tryclient.py", line 582, in createJob
    diff = f.read()
  File "/usr/lib64/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
builtins.UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcf in position 177: invalid continuation byte

[~]$cat testfile.patch
[...]

PR #6001, opened 2021-04-28
Read and store patches as binary instead of unicode.

aj062:
Read and store patches as binary instead of unicode.

Fix  'buildbot try fails for files containing non utf-8 characters #5933'

### DIFF
--- a/master/buildbot/clients/tryclient.py
+++ b/master/buildbot/clients/tryclient.py
@@ -578,7 +578,7 @@ class Try(pb.Referenceable):
             if difffile == "-":
                 diff = sys.stdin.read()
             else:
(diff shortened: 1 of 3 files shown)


tardyp (approved):
looks good to me.

p12tic:
We pass patch body to the DB layer as is, so this needs more careful look.

aj062:
This change has been running on our production buildbot instance (ews-build.webkit.org) for almost a month, without any noticeable issue.

p12tic:
This PR has been replaced by https://github.com/buildbot/buildbot/pull/6103. Thanks a lot @aj062.

PR #6046, opened 2021-05-11
Fix timeouts on master-side shell commands

p12tic:
This PR adds an reimplementation of Twisted's util.getProcessOutput family of functions with timeout functionality. This prevents a hang of master-side shell function from causing damage to the master itself. An example of hanging master-side shell command is a git pull via ssh during which the target computer is suspended. In practice this may be caused by any network interruption during ssh operations that causes the connection to not being closed or reset appropriately.

### DIFF
--- a/master/buildbot/test/unit/changes/test_svnpoller.py
+++ b/master/buildbot/test/unit/changes/test_svnpoller.py
@@ -23,8 +23,9 @@ from twisted.trial import unittest
 from buildbot.changes import svnpoller
 from buildbot.process.properties import Interpolate
 from buildbot.test.util import changesource
-from buildbot.test.util import gpo
 from buildbot.test.util.misc import TestReactorMixin
+from buildbot.test.util.runprocess import ExpectMaster
+from buildbot.test.util.runprocess import MasterRunProcessMixin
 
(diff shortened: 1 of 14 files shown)


master/buildbot/test/unit/changes/test_svnpoller.py
```diff
@@ -23,8 +23,10 @@
 from buildbot.changes import svnpoller
 from buildbot.process.properties import Interpolate
 from buildbot.test.util import changesource
-from buildbot.test.util import gpo
```
tardyp, line 26: should we eventually remove gpo?

tardyp (approved):
This looks like a lot of new code. did you base on something?

master/buildbot/test/unit/changes/test_svnpoller.py
p12tic: Yes, there are a couple of remaining users, these will go in another PR.

p12tic:
> This looks like a lot of new code. did you base on something?

Some code is based on our worker-side process runner, but of course I needed to rewrite almost everything to make it compatible with master's constraints.

PR #6055, opened 2021-05-15
Fix master dockerfile to build full code, switch to Debian and virtualenv

### LINKED ISSUE #4998: Docker image for 2.4.0 contains packages from 2.3.1
The Docker image `buildbot/buildbot-master:v2.4.0` contains Python packages for buildbot `2.3.1`:

```sh
buildbot-console-view==2.3.1
buildbot-grid-view==2.3.1
buildbot-waterfall-view==2.3.1
buildbot-worker==2.3.1
buildbot-www==2.3.1
```

The `buildbot` package itself has version `2019.8.18`, which is not available from PyPI.

Even if `2019.8.18` turns out to be identical to `2.4.0`, this is somewhat problematic: Installing packages into the image with a dependency on buildbot will always uninstall the buildbot version that comes with the official image.

It would be nice if the Dockerfile could install buildbot from PyPI to ensure deterministic builds for those who need to derive their own image.

<details>
  <summary>Output of pip freeze</summary>

```
$ docker run --rm buildbot/buildbot-master:v2.4.0 pip freeze
asn1crypto==0.24.0
attrs==19.1.0
autobahn==19.8.1
Automat==0.7.0
buildbot==2019.8.18
buildbot-console-view==2.3.1
buildbot-grid-view==2.3.1
buildbot-waterfall-view==2.3.1
buildbot-worker==2.3.1
buildbot-www==2.3.1
certifi==2019.6.16
cffi==1.12.3
chardet==3.0.4
constantly==15.1.0
cryptography==2.7
decorator==4.4.0
future==0.17.1
hyperlink==19.0.0
idna==2.8
incremental==17.5.0
Jinja2==2.10.1
MarkupSafe==1.1.1
pbr==5.4.2
psycopg2==2.8.3
pyasn1==0.4.6
pyasn1-modules==0.2.6
pycparser==2.19
PyHamcrest==1.9.0
PyJWT==1.7.1
pyOpenSSL==19.0.0
python-dateutil==2.8.0
PyYAML==5.1.2
requests==2.22.0
service-identity==18.1.0
six==1.12.0
SQLAlchemy==1.3.7
sqlalchemy-migrate==0.12.0
sqlparse==0.3.0
Tempita==0.5.2
Twisted==19.7.0
txaio==18.8.1
txrequests==0.9.6
urllib3==1.25.3
zope.interface==4.6.0
```
</details>

p12tic:
This PR fixes #4998.

Previously we were installing buildbot-www from whatever is available on Pypi. This leads to wrong versions being taken for docker containers built from tags. E.g. if 2.10.5 is built, one expects buildbot-www to correspond to 2.10.5 code, not what's latest on Pypi. Since older releases are sometimes built, it makes sense to avoid using Pypi as this may download completely incompatible buildbot-www. Fortunately we can just reuse the existing release build code to build www packages for the container and solve the problem this way.

Current Dockerfile uses Alpine, which is great for reducing the container footprint, but for master this is less important because by design there will be few instances running. Additionally, most of the users of Buildbot (myself included) are much more likely to have more experience with Debian. Given that I need to make significant changes to the Dockerfile anyway, I'm switching it to Debian.

Finally, virtualenv is used to avoid fighting to whatever python packages may be already installed on the system.

cc @tardyp


Conan-Kudo:
Would you consider using Fedora or [RHEL UBI](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image) instead? In my experience, the Debian Python runtime tends to be somewhat broken compared to the Fedora one, and Fedora's Python runtime is fully compatible and functional with wheels. Not to mention, with fresher versions of Python, that can lead to performance improvements across the board over time as new Python versions become available.

Alternatively, the main case I would make for the RHEL UBI container base is that a lot of the benefits of a Fedora base exists there too, but it's also long-term supported and would make it easier for folks to run in Red Hat OpenShift in a manner where RHEL/OpenShift customers can benefit while still allowing everyone to use it outside of RHEL (I'd be interested in helping on making this more of a reality too if we went this route).

cc: @fatherlinux

p12tic:
> In my experience, the Debian Python runtime tends to be somewhat broken compared to the Fedora one

What issues in particular you are referring to

Conan-Kudo:
I've had issues using venvs and wheels with Debian's Python runtime. The most recent issue I've had is with trying to install Python modules that use PEP-517/PEP-518 compliant `pyproject.toml` on how to build and install things. The altered paths in the Python interpreter also basically broke the ability to use system modules with virtualenvs somehow (I'm not sure how this happened for me, but it was pretty aggravating).

p12tic:
I've just looked into Buildbot metrics and during the last year we've collected events from the following systems where Buildbot master was run:
 - Debian and derivatives: ~48%
 - Alpine: 44%
 - Fedora/RedHat/Centos: ~1%
 - Other ~7%

Given these results we just can't switch to rpm-based distributions as that would just make additional work for the users. If one does not want to use the default, it's very easy to create one's own docker/buildah image with Buildbot packages from pypi.

tardyp (approved):
great!

PR #6058, opened 2021-05-16
Fix #5903 - new SecretInVault implementation

### LINKED ISSUE #5903: flawed HashiCorpVaultSecretProvider implementation
After unsuccessfuly trying to get vault working with buildbot I realized, that current naive implementation has several major flaws preventing it to be usable in real world scenarios. To give more context, we are using hashicorp vault in production at my work for about 5 years now, having several thousand secrets stored in generic kv store. We use typically 5 level deep paths allowing us for fine grained access control to secrets.

### Current issues:
1. separating optional key with `/` is ambiguos, as vault path components are separated by `/` as well,
2. only flat (single component) path can be used now, which is not viable real world use (it would severely hinder management of access control in vault),
3. only token authentication is supported, but even when I write some python code in config to obtain it, tokens are short lived with forced TTL. So in real world scenarios buildbot uptime will typically exceed token TTL, breaking pipelines after some time after config/restart,
4. vault paths in general may contain any characters so it is good practice to URL escape path components.

### To resolve this I suggest following changes:
1. use [HVAC](https://github.com/hvac/hvac) module for vault communication (it will help with points 3. and 4.),
2. use colon `:` for separating path from key (this convention is used in [Ansible hashi_vault lookup plugin](https://docs.ansible.com/ansible/latest/collections/community/hashi_vault/hashi_vault_lookup.html)), ideally with escaping extra colons in path or key name using `\`,
3. support additional login methods, at least *approle* which is probably most typical for CI/CD pipelines,
4. drop support for default `value` key, it's only confusing with no real world benefit; mentioned Ansible plugin returns dictionary with all keys when no key is specified, but I'm not sure if buildbot could handle this (if not, we can throw exception to enforce key specification).

I would welcome any comments or objections regarding this topic. I'll make PR with new HashiCorpVaultSecretProvider once we agree on this.

mixicz:
Added new HashiCorp Vault secret provider `HvacKvSecretProvider` in order to replace old bugged `HashiCorpVaultSecretProvider` implementation (https://github.com/buildbot/buildbot/issues/5903).

## Contributor Checklist:

### DIFF
--- a/master/buildbot/secrets/providers/vault.py
+++ b/master/buildbot/secrets/providers/vault.py
@@ -16,12 +16,12 @@
 vault based providers
 """
 
-
 from twisted.internet import defer
 
 from buildbot import config
 from buildbot.secrets.providers.base import SecretProviderBase
 from buildbot.util import httpclientservice
+from buildbot.warnings import warn_deprecated
 
 
 class HashiCorpVaultSecretProvider(SecretProviderBase):
@@ -33,6 +33,8 @@ class HashiCorpVaultSecretProvider(SecretProviderBase):
 
     def checkConfig(self, vaultServer=None, vaultToken=None, secretsmount=None,
                     apiVersion=1):
+        warn_deprecated("3.4.0", "Use of HashiCorpVaultSecretProvider is deprecated and will be "
+                        "removed in future releases. Use HashiCorpVaultKvSecretProvider instead")
         if not isinstance(vaultServer, str):
             config.error("vaultServer must be a string while it is {}".format(type(vaultServer)))
         if not isinstance(vaultToken, str):
--- a/master/buildbot/secrets/providers/vault_hvac.py
+++ b/master/buildbot/secrets/providers/vault_hvac.py
@@ -1,0 +1,182 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+"""
+HVAC based providers
+"""
+
+from twisted.internet import defer
+from twisted.internet import threads
+
+from buildbot import config
+from buildbot.secrets.providers.base import SecretProviderBase
+
+
+class VaultAuthenticator:
+    """
+    base HVAC authenticator class
+    """
+
+    def authenticate(self, client):
+        pass
+
+
+class VaultAuthenticatorToken(VaultAuthenticator):
+    """
+    HVAC authenticator for static token
+    """
+
+    def __init__(self, token):
+        self.token = token
+
+    def authenticate(self, client):
+        client.token = self.token
+
+
+class VaultAuthenticatorApprole(VaultAuthenticator):
+    """
+    HVAC authenticator for Approle login method
+    """
+
+    def __init__(self, roleId, secretId):
+        self.roleId = roleId
+        self.secretId = secretId
+
+    def authenticate(self, client):
+        client.auth.approle.login(role_id=self.roleId, secret_id=self.secretId)
+
+
+class HashiCorpVaultKvSecretProvider(SecretProviderBase):
+    """
+    Basic provider where each secret is stored in Vault KV secret engine.
+    In case more secret engines are going to be supported, each engine should have it's own class.
+    """
+
+    name = 'SecretInVaultKv'
+
+    def checkConfig(self, vault_server=None, authenticator=None, secrets_mount=None,
+                    api_version=2, path_delimiter='|', path_escape='\\'):
+        if not isinstance(vault_server, str):
+            config.error("vault_server must be a string while it is {}".format(type(vault_server)))
+        if not isinstance(path_delimiter, str) or len(path_delimiter) > 1:
+            config.error("path_delimiter must be a single character")
+        if not isinstance(path_escape, str) or len(path_escape) > 1:
+            config.error("path_escape must be a single character")
+        if not isinstance(authenticator, VaultAuthenticator):
+            config.error("authenticator must be instance of VaultAuthenticator while it is {}"
+                         .format(type(authenticator)))
+        if api_version not in [1, 2]:
+            config.error("api_version {} is not supported".format(api_version))
+
+    def reconfigService(self, vault_server=None, authenticator=None, secrets_mount=None,
+                        api_version=2, path_delimiter='|', path_escape='\\'):
+        try:
+            import hvac
+        except ImportError:
+            config.error("SecretInVaultKv needs the hvac package installed (pip install hvac)")
+
+        if secrets_mount is None:
+            secrets_mount = "secret"
+
+        self.secrets_mount = secrets_mount
+        self.path_delimiter = path_delimiter
+        self.path_escape = path_escape
+        self.authenticator = authenticator
+        self.api_version = api_version
+        if vault_server.endswith('/'):
+            vault_server = vault_server[:-1]
+        self.client = hvac.Client(vault_server)
+        self.client.secrets.kv.default_kv_version = api_version
+        return self
+
+    def escaped_split(self, s):
+        """
+        parse and split string, respecting escape characters
+        """
+        ret = []
+        current = []
+        itr = iter(s)
+        for ch in itr:
+            if ch == self.path_escape:
+                try:
+                    # skip the next character; it has been escaped and remove
+                    # escape character
+                    current.append(next(itr))
+                except StopIteration:
+                    # escape character on end of the string is safest to ignore, as buildbot for
+                    # each secret identifier tries all secret providers until value is found,
+                    # meaning we may end up parsing identifiers for other secret providers, where
+                    # our escape character may be valid on end of string
+                    pass
+            elif ch == self.path_delimiter:
+                # split! (add current to the list and reset it)
+                ret.append(''.join(current))
+                current = []
+            else:
+                current.append(ch)
+        ret.append(''.join(current))
+        return ret
+
+    def thd_hvac_wrap_read(self, path):
+        if self.api_version == 1:
+            return self.client.secrets.kv.v1.read_secret(path=path, mount_point=self.secrets_mount)
+        else:
+            return self.client.secrets.kv.v2.read_secret_version(path=path,
+                                                                 mount_point=self.secrets_mount)
+
+    def thd_hvac_get(self, path):
+        """
+        query secret from Vault and try to re-authenticate in case Unauthorized
+        exception when active token reaches its TTL
+        """
+
+        # no need to "try" import, it was already handled by reconfigService()
+        import hvac
+
+        try:
+            response = self.thd_hvac_wrap_read(path=path)
+        except (hvac.exceptions.Unauthorized, hvac.exceptions.InvalidRequest):
+            self.authenticator.authenticate(self.client)
+            response = self.thd_hvac_wrap_read(path=path)
+
+        return response
+
+    @defer.inlineCallbacks
+    def get(self, entry):
+        """
+        get the value from vault secret backend
+        """
+
+        parts = self.escaped_split(entry)
+        if len(parts) == 1:
+            raise KeyError("Vault secret specification must contain attribute name separated from "
+                           "path by '{}'".format(self.path_delimiter))
+        if len(parts) > 2:
+            raise KeyError("Multiple separators ('{0}') found in vault path '{1}'. All occurences "
+                           "of '{0}' in path or attribute name must be escaped using '{2}'"
+                           .format(self.path_delimiter, entry, self.path_escape))
+        name = parts[0]
+        key = parts[1]
+
+        response = yield threads.deferToThread(self.thd_hvac_get, path=name)
+
+        # in KVv2 we have extra "data" dictionary, as vault provides metadata as well
+        if self.api_version == 2:
+            response = response['data']
+
+        try:
+            return response['data'][key]
+        except KeyError as e:
+            raise KeyError(
+                "The secret {} does not exist in Vault provider: {}".format(entry, e)) from e
--- a/master/buildbot/secrets/providers/vault_hvac.py
+++ b/master/buildbot/secrets/providers/vault_hvac.py
@@ -101,7 +101,7 @@ class HashiCorpVaultKvSecretProvider(SecretProviderBase):
         self.path_escape = path_escape
         self.authenticator = authenticator
         self.api_version = api_version
-        if vault_server.endswith('/'):
+        if vault_server.endswith('/'):  # pragma: no cover
             vault_server = vault_server[:-1]
         self.client = hvac.Client(vault_server)
         self.client.secrets.kv.default_kv_version = api_version
--- a/master/buildbot/secrets/providers/vault_hvac.py
+++ b/master/buildbot/secrets/providers/vault_hvac.py
@@ -67,6 +67,13 @@ class HashiCorpVaultKvSecretProvider(SecretProviderBase):
 
     def checkConfig(self, vault_server=None, authenticator=None, secrets_mount=None,
                     api_version=2, path_delimiter='|', path_escape='\\'):
+        try:
+            import hvac
+            [hvac]
+        except ImportError:  # pragma: no cover
+            config.error(f"{self.__class__.__name__} needs the hvac package installed " +
+                         "(pip install hvac)")
+
         if not isinstance(vault_server, str):
             config.error("vault_server must be a string while it is {}".format(type(vault_server)))
         if not isinstance(path_delimiter, str) or len(path_delimiter) > 1:
@@ -83,12 +90,12 @@ class HashiCorpVaultKvSecretProvider(SecretProviderBase):
                         api_version=2, path_delimiter='|', path_escape='\\'):
         try:
             import hvac
-        except ImportError:
-            config.error("SecretInVaultKv needs the hvac package installed (pip install hvac)")
+        except ImportError:  # pragma: no cover
+            config.error(f"{self.__class__.__name__} needs the hvac package installed " +
+                         "(pip install hvac)")
 
         if secrets_mount is None:
             secrets_mount = "secret"
-
         self.secrets_mount = secrets_mount
         self.path_delimiter = path_delimiter
         self.path_escape = path_escape
--- a/master/buildbot/test/unit/test_secret_in_hvac.py
+++ b/master/buildbot/test/unit/test_secret_in_hvac.py
@@ -1,0 +1,201 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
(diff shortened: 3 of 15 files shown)


master/buildbot/secrets/providers/vault.py
```diff
@@ -33,6 +34,8 @@ class HashiCorpVaultSecretProvider(SecretProviderBase):
 
     def checkConfig(self, vaultServer=None, vaultToken=None, secretsmount=None,
                     apiVersion=1):
+        warnings.warn("Use of HashiCorpVaultSecretProvider is deprecated and will be removed in "
```
p12tic, line 37: We have warn_deprecated

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class HvacKvSecretProvider(SecretProviderBase):
+    """
+    basic provider where each secret is stored in Vault KV secret engine
+    """
+
+    name = 'SecretInVaultKv'
+
+    def checkConfig(self, vaultServer=None, authenticator=None, secretsmount=None,
```
p12tic, line 67: Prefer snake_case for new code.

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+        get the value from vault secret backend
+        """
+
+        parts = self.escapedSplit(entry)
+        if len(parts) == 1:
+            raise KeyError("Vault secret specification must contain attribute name separated from "
+                "path by '{}'".format(self.pathDelimiter))
+        if len(parts) > 2:
+            raise KeyError("Multiple separators ('{0}') found in vault path '{1}'. All occurences "
+                           "of '{0}' in path or attrbute name must be escaped using '{2}'"
```
p12tic, line 152: typo (attrbute).

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+                ret.append(''.join(current))
+                current = []
+            else:
+                current.append(ch)
+        ret.append(''.join(current))
+        return ret
+
+    def hvacWrapRead(self, path):
+        if self.apiVersion == 1:
+            return self.client.secrets.kv.v1.read_secret(path=path, mount_point=self.secretsmount)
```
p12tic, line 123: This will involve network operations, won't it? We'll need to run this code from withing a thread pool thread so that whole Buildbot is not blocked while this function is running. See how `threads.deferToThread()` is used throughout the codebase. `def get()` will need to become `@defer.inlineCallbacks`, which is a sort of a Twisted's version of async/await.

master/buildbot/test/unit/test_secret_in_hvac.py
```diff
@@ -0,0 +1,192 @@
...
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+from buildbot.test.util import interfaces
+from twisted.trial import unittest
+
+from unittest.mock import patch
+
+from buildbot.secrets.providers.vault_hvac import (HvacAuthenticatorToken, HvacAuthenticatorApprole,
```
p12tic, line 21: Buildbot uses one import per line.

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+            secretsmount = "secret"
+
+        self.secretsmount = secretsmount
+        self.pathDelimiter = pathDelimiter
+        self.pathEscape = pathEscape
+        self.authenticator = authenticator
+        self.apiVersion = apiVersion
+        if vaultServer.endswith('/'):
+            vaultServer = vaultServer[:-1]
+        self.client = hvac.Client(vaultServer)
```
p12tic, line 93: If the client class is stored in a class variable, then we could easily override it in tests and no mocking would be needed.

p12tic:
Seems like a relatively well done first implementation. I had a number of comments, most of which talk about things you wouldn't have possibly known, so good job! Thanks a lot for your work.

master/buildbot/secrets/providers/vault_hvac.py
mixicz: snake_case for all of: method names, method parametres, variables and class attributes?

master/buildbot/secrets/providers/vault_hvac.py
mixicz: Yes, that's correct, this is network operation. I'll refactor it per your suggestion.

master/buildbot/secrets/providers/vault_hvac.py
p12tic: Yes, except of course when you're overriding existing method, as is `checkConfig` in this case :)

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+        ret = []
+        current = []
+        itr = iter(s)
+        for ch in itr:
+            if ch == self.pathEscape:
+                try:
+                    # skip the next character; it has been escaped and remove
+                    # escape character
+                    current.append(next(itr))
+                except StopIteration:
```
tardyp, line 110: is it really normal to have escape character in the end of string?

master/buildbot/secrets/providers/vault_hvac.py
tardyp: you can see other code in buildbot, the code that is supposed to block shall be prefixed by thd_

So you can keep your code as is and make sure you use deferToThread and inlineCallbacks in the get() method.

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __init__(self, roleId, secretId):
+        self.roleId = roleId
+        self.secretId = secretId
+
+    def authenticate(self, client):
+        client.auth.approle.login(role_id=self.roleId, secret_id=self.secretId)
+
+
+class HvacKvSecretProvider(SecretProviderBase):
```
tardyp, line 60: I am not too sure about the name. I would prefer it has Vault in it.  hvac is implementation detail

master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -0,0 +1,164 @@
...
+# Copyright Buildbot Team Members
+"""
+HVAC based providers
+"""
+
+
+from buildbot import config
+from buildbot.secrets.providers.base import SecretProviderBase
+
+import hvac
```
tardyp, line 23: hvac should be optional dependency.  For those, we usually do a try import and provide a more meaningfull message, e.g. what to pip install (so that the user don't guess and be vulnerable to typo squatting)

master/docs/manual/secretsmanagement.rst
```diff
@@ -115,11 +115,58 @@ Arguments:
...
+
+The secret identifiers that need to be passed to, e.g. :ref:`Interpolate`, have format: ``"path/to/secret:key"``.
+In case path or key name does contain colon chraracter, it is possible to escape it using "\" or specify different separator character using ``pathDelimiter`` parameter when initializing secret provider.
+
+Example use:
+
+.. code-block:: python
+
+    passwd = util.Secret('path/to/secret:password')
+
```
tardyp, line 162: state here a comment about optional dep

master/setup.py
```diff
@@ -250,7 +250,8 @@ def define_plugin_entries(groups):
         ('buildbot.secrets', [
             ('buildbot.secrets.providers.file', ['SecretInAFile']),
             ('buildbot.secrets.providers.passwordstore', ['SecretInPass']),
-            ('buildbot.secrets.providers.vault', ['HashiCorpVaultSecretProvider'])
+            ('buildbot.secrets.providers.vault', [
+                'HashiCorpVaultSecretProvider', 'HvacKvSecretProvider'])
```
tardyp, line 254: This class is in vault_hvac.py, not vault.py
```suggestion
            ('buildbot.secrets.providers.vault', [
                'HashiCorpVaultSecretProvider'],
             'buildbot.secrets.providers.vault_hvac', 
               ['HvacKvSecretProvider'])
```

tardyp (changes requested):
agree that this is good work, thanks!

some more comments to perfect it..

master/buildbot/secrets/providers/vault_hvac.py
mixicz: Not normal, but possible (most probably typo or some mistake, as exploitation is not likely in this scenario). 
I prefer to have such edge cases covered, as this exception would be quite confusing for most users. But when retrospectively looking at the code, I should display warning instead of just ignoring exception in this case. So thanks for pointing this out.

master/buildbot/secrets/providers/vault_hvac.py
mixicz: How about `VaultKvSecretProvider`? I have used Hvac in name to distinguish it from old implementation, but you're right this is just implementation detail and should not be presented to user.

master/buildbot/secrets/providers/vault_hvac.py
tardyp: Could we just support the old API as well? I think it would make more sense to just keep the old name :)

You can support the old API via kwargs, and just adapt it to the new api. Those are not to far away, you just have to create an authenticator based on the token (and warn_deprecated)

mixicz: There already was discussion regarding this topic in https://github.com/buildbot/buildbot/issues/5903.

Honestly, old implementation has so many flaws, that it is hardly usable in real world scenarios. Keeping backwards compatibility would bloat the code with heap of confusing conditionals and bug simulations to support just few experimental installations.

In this particular case, I really do prefer keeping the code clean and just deprecate old implementation.

p12tic: @tardyp There are edge cases which make migration harder than needed (a mandatory extra argument for all users). So I agree that a separate implementation makes sense here.

master/buildbot/secrets/providers/vault_hvac.py
mixicz: On second thought, buildbot is handling secrets in way, where it asks all secret providers for each secret identifier until it finds one able to handle the secret. As result, we can end up parsing foreign secret identifiers for which it may be valid to have *our* escape character on end of the string. In such case, throwing warnings would be very confusing for user and unhandled exceptions would even break things. So `pass` after `except StopIteration:` seems to be best option.

master/buildbot/secrets/providers/vault_hvac.py
tardyp: I agree with your arguments.

What about calling the new one HashiCorpVaultSecretProvider and the older one LegacyHashiCorpVaultSecretProvider?

We can provide detection of use of the older api, and the very few users can either  use LegacyHashiCorpVaultSecretProvider or upgrade the few secrets they have to use new api.

It annoys me to need to forever choose a bad name to keep upgrade seemless for very few user.

https://martinfowler.com/bliki/TwoHardThings.html

master/buildbot/secrets/providers/vault_hvac.py
p12tic: Add a comment explaining this situation a little bit.

master/buildbot/secrets/providers/vault_hvac.py
p12tic: The problem is that we can't detect the use of the old API reliably. Though maybe we can just use differently named arguments just for the sake of upgrade process.

tardyp: I think we can just add (vaultToken=None) argument, and if this is not None, then this is a usage of the old version of the API. 
Then the user would have the responsability to update its secret renders

mixicz: Well I suggest `HashiCorpVaultKvSecretProvider` then, with added "Kv" part (basically substitute `s/Hvac/HashiCorpVault/` in my original name). 

Why I did add "Kv" to name? When I dug deeper in hvac implementation I realized it would be very complicated to have support for multiple secret engines in single class, as they can be very different and on both hvac and in vault they have completely separate independent API. So I decided to follow vault and hvac pattern in this regard and implement kv secret engine as separate class. In case more secret engines will be implemented in future, each will have it's own class (authentication and common functionalities can be moved to some base class when necessary).

tardyp:
OK this plan makes sense to me.

master/buildbot/secrets/providers/vault_hvac.py
mixicz: > I think we can just add (vaultToken=None) argument, and if this is not None, then this is a usage of the old version of the API.
> Then the user would have the responsability to update its secret renders

Well, for API to be really compatible I would also have to somehow handle default value for `apiVersion`, which is different between the two. While it is possible, it would also bloat the code (and even duplicated thanks to `reconfigService()` vs. `checkConfig()`). And all it would help would be slightly simpler change in secretProvider configuration? User still have to rewrtite all secret identifiers, so he has to invest significantly more effort to migrate to new vault provider anyway. And not to mention that token login method is not suitable for production use (because of enforced TTL on vault tokens), so in the end all it would help are some experimental installations. Is it really worth it?

p12tic: @tardyp My vote would be to just have a separate class. The current implementation is quite broken and the user will need to put quite a lot to migrate anyway. Adding a requirement to rename a class would be a minor inconvenience.

tardyp: I am okay with the HashiCorpVaultKvSecretProvider.

It makes sense as it only supports KV indeed and not other secrets. We need to be very clear in the doc about the difference as it might be a bit too subtle.

mixicz: OK, I'll update docs accordingly.

master/buildbot/secrets/providers/vault_hvac.py
mixicz: I'm somehow stuck at making hvac optional dependency. While it's ok to make it optional in vault_hvac.py, I still need it for unit tests to simulate expired token (`raise hvac.exceptions.Unauthorized` in `FakeHvacKvV*`). If I understand it correctly, when I remove hvac from requirements-ci.txt, unit tests will fail in clean environment. Is there some way to specify dependency for unit tests only? Or any way how to overcome this?

p12tic: Having hvac in requirements-ci.txt is okay. What we are trying to achieve is that tests ran when hvac is not installed would just skip themselves. See e.g. how master/buildbot/test/unit/steps/test_http.py is setup.

master/setup.py
```diff
@@ -250,7 +250,10 @@ def define_plugin_entries(groups):
         ('buildbot.secrets', [
             ('buildbot.secrets.providers.file', ['SecretInAFile']),
             ('buildbot.secrets.providers.passwordstore', ['SecretInPass']),
-            ('buildbot.secrets.providers.vault', ['HashiCorpVaultSecretProvider'])
+            ('buildbot.secrets.providers.vault', [
```
p12tic, line 253: We use 100 column line limit, so this and the following line can stay not wrapped.

master/buildbot/secrets/providers/vault_hvac.py
mixicz: refactored in suggested way

master/buildbot/secrets/providers/vault_hvac.py
mixicz: I have refactored tests as sugested

master/buildbot/test/unit/test_secret_in_hvac.py
```diff
@@ -0,0 +1,198 @@
...
+
+class FakeHvacKvV1:
+    token = None
+
+    def read_secret(self, path, mount_point):
+        if self.token is None:
+            raise hvac.exceptions.Unauthorized
+        if path == "wrong/path":
+            raise hvac.exceptions.InvalidPath(message="Fake InvalidPath exception")
+        return dict(data=dict(key="value"))
```
p12tic, line 51: I would prefer dict literals due to readability

master/docs/developer/secrets.rst
```diff
@@ -50,14 +50,13 @@ Vault provider
 
 .. code-block:: python
 
-    c['secretsProviders'] = [secrets.SecretInVault(vaultToken=open('VAULT_TOKEN').read(),
-                                                vaultServer="http://localhost:8200",
-                                                apiVersion=2)]
+    c['secretsProviders'] = [secrets.SecretInVaultKv(authenticator=secrets.VaultAuthenticatorApprole(roleId="xxx", secretId="yyy"),
```
p12tic, line 53: I think this should be HashiCorpVaultKvSecretProvider. The previous value is an error.

master/docs/manual/secretsmanagement.rst
```diff
@@ -115,11 +115,60 @@ Arguments:
...
 .. _HashiCorpVaultSecretProvider:
 
 HashiCorpVaultSecretProvider
 ````````````````````````````
 
+**Warning: use of ``HashiCorpVaultSecretProvider`` is deprecated in favor of newer :ref:`HashiCorpVaultKvSecretProvider` and will be removed in future releases.**
```
p12tic, line 170: Use `.. note::`

p12tic (changes requested):
Just a couple of nits, otherwise looks great, thanks a lot.

It seems that this whole PR could be a single commit - could you rebase on latest master and squash the commits into one?

master/docs/developer/secrets.rst
mixicz: You are right, I tested it and buildbot throws an error.

mixicz:
I looked into CI errors and I'm quite confused with exception: 
``twisted.trial.unittest.FailTest: Items in the first set but not the second:
'buildbot.secrets.providers.vault_hvac.VaultAuthenticator'``

Manually run tests are fine and circleci didn't complain as well. I didn't manage to find any useful advice regarding this. Do you have any idea what's going on?

mixicz:
> Just a couple of nits, otherwise looks great, thanks a lot.
> 
> It seems that this whole PR could be a single commit - could you rebase on latest master and squash the commits into one?

OK, I'll try to. 
But first I want to fix CI errors, so expect some more commits before I will squash them.

yan12125:
> Manually run tests are fine and circleci didn't complain as well

FWIW, those integration tests require extra steps before running them. Here are my steps:
```
$ python setup.py egg_info
$ PYTHONPATH=$PWD trial buildbot/test/integration/test_setup_entrypoints.py
```

mixicz:
> FWIW, those integration tests require extra steps before running them. Here are my steps:
> 
> ```
> $ python setup.py egg_info
> $ PYTHONPATH=$PWD trial buildbot/test/integration/test_setup_entrypoints.py
> ```

Thanks, this helped.

p12tic:
In the existing test which is now failing with `buildbot.warnings.DeprecatedApiWarning: [3.2.1 and later] Use of HashiCorpVaultSecretProvider is deprecated and will be removed in future releases. Use HashiCorpVaultKvSecretProvider instead` you can ignore the warning by using this:

```
with assertProducesWarnings(DeprecatedApiWarning,
                            message_pattern='have been deprecated'):
```

Also I think the version should be specified as 3.3.0 since we're not going to add this feature to a point bugfix release.

p12tic:
Aside from the above, this PR looks great and is ready to go in. Only some trivial errors are still left to fix - correct import sorting (the isort test) and spelling. pylint spelling exceptions are stored in common/code_spelling_ignore_words.txt, documentation spelling exceptions are stored in master/docs/spelling_wordlist.txt.

p12tic:
@mixicz Are you planning to continue working on this? Only a couple of trivial problems are left and this PR could go in.

mixicz:
> @mixicz Are you planning to continue working on this? Only a couple of trivial problems are left and this PR could go in.

Yes, I do plan to finish this, but I had different priorities recently. There is small chance I'll have time to finish this next week, but most probably it'll have to wait 2 more weeks until after my vacation.

I'm sorry it takes this long, but all those trivial "5 minute" problems usually take me several hours, as literally only thing in this project I have some knowledge about is vault and python language.

p12tic:
@mixicz @tardyp  Turns out we can't use a colon to separate key and value because it breaks usage in `Interpolate`. I've replaced the default separator with pipe character (`|`). Do you have any better suggestions?

mixicz:
> @mixicz @tardyp Turns out we can't use a colon to separate key and value because it breaks usage in `Interpolate`. I've replaced the default separator with pipe character (`|`). Do you have any better suggestions?

@p12tic from vault point of view, pipe will work fine, it has very low probability of using inside vault paths or keys, so it won't have to be escaped often. Colon was chosen to maintain same convention as Ansible uses, but if we can't use it, pipe is good choice.

p12tic:
I've just noticed that I've lost an email (but not author name) during rebase. Sorry about that.

mixicz:
> I've just noticed that I've lost an email (but not author name) during rebase. Sorry about that.

No problem, really. Thanks a lot for finishing this for me.

PR #6063, opened 2021-05-21
steps.pbuilder: Add support for --othermirror

Zash:
This is useful for adding additional repositories with local build-dependencies etc.

## Contributor Checklist:


master/buildbot/steps/package/deb/pbuilder.py
```diff
@@ -77,6 +79,10 @@ def __init__(self,
             self.distribution = distribution
         if mirror:
             self.mirror = mirror
+        if othermirror:
+            if type(othermirror) is list:
```
p12tic, line 83: Let's require it to be a list. This would be one less special case to support and also lead to a slightly more consistent interface.

p12tic (changes requested):
This is great PR, thanks. I had just a single comment.

master/buildbot/steps/package/deb/pbuilder.py
Zash: Sounds good, will do. And now I understand why the coverage went down.

p12tic:
@Zash: Could you please add `othermirror` to docs spelling word list at `master/docs/spelling_wordlist.txt`?

PR #6079, opened 2021-05-26
SASL basic support

tardyp:
This is a basic SASL support for libera users which are sasl enforced.

this PR is not finished and is only here for users to test

I think the SASL enablement is early enough in the protocol but cannot verify.


master/buildbot/reporters/irc.py
```diff
@@ -328,6 +351,9 @@ def left(self, channel):
     def kickedFrom(self, channel, kicker, message):
         self.log("I have been kicked from {} by {}: {}".format(channel, kicker, message))
 
+    def noticed(self, user, channel, message):
+        self.log("I have been noticed by by {}: {}".format(user or channel, message))
```
p12tic, line 355: Typo `by by`

p12tic (changes requested):
LGTM, but needs tests.

PR #6097, opened 2021-06-08
invalid property selector name

senyai:
it's singular form there https://github.com/buildbot/buildbot/blob/master/master/buildbot/process/properties.py#L649-L655


tardyp (approved):
nice catch!

PR #6099, opened 2021-06-11
Separate worker paused state and worker quarantine state

### LINKED ISSUE #5611: Quarantine timer unpausing manually paused worker
When I pause a worker during a build and an exception happens on that build, the worker becomes unpaused. I suspect it is the quarantine timer causing this to happen. 

If this is the case, is there a way to prevent the quarantine timer unpausing a manually paused worker or a way to increase the initial delay before the worker is unpaused?

MichaelOultram:
Fixes #5611 

I think this changeset is a step in the right direction. It separates "paused" (which is a manual operation) and "quarantined" (which is an automatic process). Thought process behind my changeset:
- If I manually pause a worker, then I do not want it to be unpaused by the quarantine timer.
- If I unpause a worker, I would like it to leave quarantine immediately so I can see if it is now working. The quarantine timeout should not be reset as if the worker is still broken, then it should go back into quarantine for some time and not eat the build queue.

With this changeset, it's now more difficult to see when a worker is in quarantine from the UI (as it won't be paused). The worker would have a streak of red so you'd know it's borked. Ideally the quarantine state would go into the db and be visible from the UI/API. Quarantine state could include:
- is worker in quarantine
- how long quarantine was set for
- when the worker will leave quarantine

Plumbing the UI/DB parts would increase the changeset size, and I'm not sure I'm that comfortable with the full buildbot stack to do it. This changeset does stand on it's own as an improvement on what already exists.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1527,6 +1527,9 @@ unlinked
 unparsable
 unparseable
 unpatch
+unpause
+unpaused
+unpausing
 unregister
 unregisters
 unserialized
--- a/master/buildbot/newsfragments/separate_worker_paused_and_quarantine_concepts.feature
+++ b/master/buildbot/newsfragments/separate_worker_paused_and_quarantine_concepts.feature
@@ -1,0 +1,1 @@
+Pausing a worker is a manual operation which the quarantine timer was overrwriting. Worker paused state and quarantine state are now independent. (:issue:`5611`)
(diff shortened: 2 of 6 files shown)


p12tic (approved):

p12tic:
This looks great! Could you address the pylint and docs spelling issues?

MichaelOultram:
@p12tic spellings are now fixed

PR #6121, opened 2021-06-20
Git poller: replace lastRev state instead of updating

### LINKED ISSUE #6100: Git poller state exceeds database field size after some time
When the Git poller covers all branches, it can happen that eventually the `lastRev` state gets too large so that it cannot be stored in the DB anymore (resulting in `Data too long for column 'value_json' at row 1` exceptions). This is because it never forgets branches that used to exist in the past (https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L239), so its `lastRev` state never shrinks.

I'm wondering whether the `lastRev` state really needs to be updated (and not replaced) by the current set of references. The only reason why update is used (instead of replacing) I can think of is that if a branch is deleted and then restored (with the same HEAD), it will not trigger a change.

I'd like to create a PR which removes the updating, or at least makes it configurable. What is the preferred solution here?

felixfontein:
Fixes #6100.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/gitpoller.py
+++ b/master/buildbot/changes/gitpoller.py
@@ -236,7 +236,7 @@ class GitPoller(base.PollingChangeSource, StateMixin, GitMixin):
                 log.err(_why="trying to poll branch {} of {}".format(
                         branch, self.repourl))
 
-        self.lastRev.update(revs)
+        self.lastRev = revs
         yield self.setState('lastRev', self.lastRev)
 
     def _get_commit_comments(self, rev):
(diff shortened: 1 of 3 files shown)


master/buildbot/changes/gitpoller.py
```diff
@@ -236,6 +236,8 @@ def poll(self):
                 log.err(_why="trying to poll branch {} of {}".format(
                         branch, self.repourl))
 
+        # replace self.lastRev contents with revs
+        self.lastRev.clear()
         self.lastRev.update(revs)
```
p12tic, line 241: Is there a reason why we can't just set lastRev to revs?

felixfontein: Probably not. I've changed it in 01bb6f5269e503bb46ea4288dceadc42e04eff68.

p12tic (approved):

p12tic:
Rebased, will merge soon.

felixfontein:
@p12tic thanks a lot for reviewing and merging!

PR #6128, opened 2021-06-27
process: Remove deprecated process.subunitlogobserver

p12tic:
This has been deprecated for a very long time.


p12tic:
I though maybe let's deprecate all such functionality first, who knows how many old buildbots are using them.

It's not the only property with this problem.

Issue #6132, opened 2021-07-07
WorkerLock's maxCountForWorker ineffective when acquiring lock for a buildstep


slawek-es:
I have been trying to define a `WorkerLock` with the number of maximum concurrent builds specifically set for one of my workers, and add it to a single build step, like this:

```
MY_LOCK = util.WorkerLock("mylock",
                          maxCount=10,
                          maxCountForWorker={"myworker": 1})
```
```
class MyCustomBuild(Compile):
    def __init__(
        self,
        **kwargs: Any,
    ) -> None:
        kwargs.setdefault("locks", [MY_LOCK.access("counting")])
		super().__init__(**kwargs)
        ...
```

It looks like there is an issue which prevents the worker-specific maximum count from being used, as I can see that the `MyCustomBuild` step can be executed twice concurrently on the same worker.

After I added instrumentation to Buildbot's code, it seemed to me that a `Worker` object gets passed to `RealWorkerLock.getLockForWorker()`: the value passed as `workername` was stringified as `<Worker 'myworker'>` rather than just `myworker`.

p12tic:
Fixed in #6133.

PR #6133, opened 2021-07-07
Fix WorkerLock's maxCountForWorker when used with buildsteps

slawek-es:
There has been an issue where the Worker object, rather than just the
worker's name, has been passed to 'RealWorkerLock.getLockForWorker'
when the lock was acquired for a single build step. As a result, the
erroneous 'workername' value was not found in the 'maxCountForWorker'
dictionary, and so we fell back to the 'maxCount' value.

We fix this by passing the workername rather than the Worker object
from 'BuildStep.startStep' into 'getLockForWorker'.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/workerlock-maxcountforworker.bugfix
+++ b/master/buildbot/newsfragments/workerlock-maxcountforworker.bugfix
@@ -1,0 +1,1 @@
+Fixed worker lock handling that caused max lock count to be ignored (:issue:`6132`).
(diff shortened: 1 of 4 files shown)


slawek-es:
This fixes https://github.com/buildbot/buildbot/issues/6132

p12tic (changes requested):
LGTM, thanks a lot.

We probably need tests. These are currently in master/buildbot/test/unit/process/test_buildstep.py, but I'm not sure how to reproduce the bug in the test environment.

I will get back once I have an idea.

tardyp:
I guess we need a test for this usecase to make sure that this feature will always be working

p12tic (approved):
LGTM. Thanks for submitting this PR and doing full investigation.

p12tic:
I've added tests and rebased. Turns out we actually didn't test whether any locking happened.

PR #6137, opened 2021-07-12
Add support for configuring github auth token for github change dialect via secrets management

anish:
## Contributor Checklist:

I have not actually been able to write the correct unit test for this. There is no existing test for `_get_pr_files` or `_get_commit_msg` but only `handle_pull_request`. All test in `test_hooks_github.py` are aimed at incoming POST requests, while this would have to be for an outgoing GET request in response to those. Still working on it, would appreciate any examples of a similar test that already exists


p12tic (approved):
LGTM. I'll look into how to create tests for this PR.

Issue #6146, opened 2021-07-19
Custom services can't be updated when using reconfig


jsonar-cpapke:
When writing a custom class that is subclassing a BuildbotService class in some way (like a custom `MailNotifier` or `LatentWorker`), any changes to the class code will not be picked up during a reconfig of the buildbot master.

This of course is because of how reconfig works, but it would be nice to have a way of forcing the service to use the new code in these situations so that custom classes can be updated while using reconfig.

PR #6147, opened 2021-07-19
Allow forcing use of new service class code on reconfig

jsonar-cpapke:
Closes https://github.com/buildbot/buildbot/issues/6146

I'm not sure if this is the right way to do something like this, so I haven't added tests or a new fragment yet.

The basic issue I'm trying to solve is being able to use custom services with reconfig.

For example, if I create a custom `MailNotifier`, I would need to stop and start the buildbot master if the class ever changes, as when doing a reconfig the old instance is always used.

With the `canReconfigWithSibling` function, I could have my custom classes always return `False` to force buildbot to use the new class code on a reconfig.

## Contributor Checklist:


master/buildbot/util/service.py
```diff
@@ -470,15 +473,16 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
...
             old = old_by_name[n]
             new = new_by_name[n]
-            # detect changed class name
-            if reflect.qual(old.__class__) != reflect.qual(new.__class__):
-                removed_names.add(n)
-                added_names.add(n)
             # compare using ComparableMixin if they don't support reconfig
-            elif not hasattr(old, 'reconfigServiceWithBuildbotConfig'):
-                if not util.ComparableMixin.isEquivalent(old, new):
+            if not hasattr(old, 'reconfigServiceWithBuildbotConfig'):
```
tardyp, line 477: why do you change the logic order, could'n you just extract the reflect test into a method, and keep the same if order?

tardyp (changes requested):
Hello,
In principle, the reconfig is really for reloading the configuration, and not the code.
reloading the code introduce a number of problems when you scale.
You will soon want to split your code into modules, and then you will want to do ``importlib.reload(module)``.
And that is a thing that a lot of people have tried and that does not work well, because of module ordering, and different version of the code which lives in the python memory (for builds before the reload and after the reload)

So we don't recommend to have a code that is as much as possible reconfigurable by design (with configuration as data files, rather than code)

and when we need to have configuration as code, then you just do a clean restart of the master(s)

So this was the warning part. Then if we really want to include this feature, I think it indeed may make sense.
For the implementation logic I though of a ``forceReloadCode=True`` attribute, or a version attribute, both does not add much, and the ``canReconfigWithSibling`` is probably the most generic way of doing this. The default implementation of canReconfigWithSibling could indeed check for a version attribute or a forceReloadCode attribute for its decision rather than just the class name.

We could also play the game of comparing the actual code by looking at ``__code__`` of inspect object.

What I don't understand too much in your proposal is why you change the logic order.

master/buildbot/util/service.py
jsonar-cpapke: I was going with the assumption that if a class doesn't have the `reconfigServiceWithBuildbotConfig` function, then it wouldn't have the new `canReconfigWithSibling` function.

I don't really understand when a class/instance could be missing `reconfigServiceWithBuildbotConfig`, so I just went with a guess.

tardyp: ah.. you are right... 
This might be some transition code. I think you may just drop the test, and we will see with test failures if there are still some services that are not BuildbotService.

jsonar-cpapke:
As for how the function should check if a reconfig can be done or not, I'm happy to add any of the suggestions.

I started with the simple class name comparison as that is what already existed. I could add a check to compare the previous and current code, but I worry that it might be more confusing as if a parent class were to change it would not pick up those changes. Only changes on the "current" class would be detected. 
But of course this would probably help out for many people who don't want to worry about how reconfig works.

master/buildbot/util/service.py
jsonar-cpapke: 👍  I changed the code to only do the call to `canReconfigWithSibling` now.

master/buildbot/util/service.py
```diff
@@ -462,23 +465,17 @@ def reconfigServiceWithBuildbotConfig(self, new_config):
...
             old = old_by_name[n]
             new = new_by_name[n]
-            # detect changed class name
-            if reflect.qual(old.__class__) != reflect.qual(new.__class__):
+            # check if we are able to reconfig service
+            if not old.canReconfigWithSibling(new):
                 removed_names.add(n)
                 added_names.add(n)
-            # compare using ComparableMixin if they don't support reconfig
-            elif not hasattr(old, 'reconfigServiceWithBuildbotConfig'):
```
p12tic, line 478: Could we move this change to a separate PR?

It would be great to add this to the description of the commit so that any future readers are less confused:

```
Child services of BuildbotServiceManager can't have reconfigServiceWithBuildbotConfig and even if they do it does not do anything because it may only be invoked by the parent service. In this case parent service is BuildbotServiceManager and it does not invoke reconfigServiceWithBuildbotConfig.

In the end this code made reconfiguration less efficient because services were added to removed_names and added_names unnecessarily.
```

p12tic:
@jsonar-cpapke The PR looks good to me. Could you combine the commits about `reconfigServiceWithBuildbotConfig` into single commit and commits related to canReconfigWithSibling function to another?

Thanks a lot!

master/buildbot/util/service.py
jsonar-cpapke: I'm not sure I understand that comment. This line was related to (I believe) old classes that were not subclassing `BuildbotService` and so didn't have the `reconfigServiceWithBuildbotConfig` method and couldn't be reconfigured.

But the assumption from removing this check is all services can be reconfigured, and so we don't need to check the existence of this method anymore.

p12tic: OK, you're right, I missed that `BuildbotService` derives from `ReconfigurableServiceMixin`. So perhaps the old code was using `reconfigServiceWithBuildbotConfig` as a proxy for existence of `reconfigServiceWithSibling` which is what `BuildbotService` is actually calling later.

Then maybe this sentence within the description of the commit would make sense?

"""
All services handled by BuildbotServiceManager are now expected to derive from BuildbotService and thus will have both reconfigServiceWithBuildbotConfig and reconfigServiceWithSibling. Therefore the check is no longer needed.
"""

I just want to have some more information of what's happened because if this commit breaks stuff, the explanation for the breakage could better be in its description and not in some PR comments that are harder to find.

jsonar-cpapke: Yes, I think that comment would make sense, I'll add that to the commit message.

jsonar-cpapke:
@p12tic I have combined the commits as requested. I also added a message to the commit where I removed the check for `reconfigServiceWithBuildbotConfig`.

p12tic (approved):

PR #6152, opened 2021-07-28
Fix several build request collapsing bugs

### LINKED ISSUE #3762: Multiple nightly builds with the same time and builder do not work
When multiple Nightly schedulers fire, I get an error in the logs, and none of them actually work.

In my config:
```
for distribution in [1.0, 2.0, 3.0]:
        Nightly(name="nightly-offline-updater-{}".format(distribution),
                builderNames=["Offline Updater"],
                properties={'update_release': distribution},
                hour=22)
```

In the logs:
```
2017-11-16 22:00:00-0500 [-] added buildset 712 to database
2017-11-16 22:00:00-0500 [-] added buildset 713 to database
2017-11-16 22:00:00-0500 [-] added buildset 714 to database
2017-11-16 22:00:01-0500 [-] tried to complete 2 buildrequests, but only completed 1
2017-11-16 22:00:01-0500 [-] Got fatal Exception on DB
        Failure: buildbot.db.buildrequests.NotClaimedError:
```

I see some other person encountered this last year but I cannot find any bugs which seem like this issue:
https://lists.buildbot.net/pipermail/users/2016-October/000849.html

As a workaround I'll probably switch to a series of Triggerables with one Nightly Trigger, but the configuration is less elegant that way.

### LINKED ISSUE #4642: If two schedulers have same builder and both are triggered, then they cancel each other
Buildbot version: 2.0.1
Twisted version: 18.9.0

I have the following configuration
```.py

def obox2_repo_filter_fn(repo):
  return re.match("git@.*:[^/]*/obox2.git$", repo)

def branch_not_2_x_repo_filter_fn(branch):
  if re.match(".*-2[.][0-9]$", branch):
    return False
  return True

c['schedulers'].append(schedulers.SingleBranchScheduler(
                       name="obox2-master",
                       builderNames=["obox2-plugin-ci","obox2-plugin-scanbuild-ci","obox2-plugin-coverage-ci"],
                       change_filter=util.ChangeFilter(
                         repository='git@git.dovecot.net:dovecot/obox2.git',
                         branch='master'
                       ),
                       treeStableTimer=10,
                       properties = {
                         'docker_image' : 'registry.dovecot.net/dovecot:stretch-worker'
                       }
))
c['schedulers'].append(schedulers.AnyBranchScheduler(
                       name="obox2-ci",
                       builderNames=["obox2-plugin-ci"],
                       change_filter=util.ChangeFilter(
                         repository_fn=obox2_repo_filter_fn,
                         branch_fn=branch_not_2_x_repo_filter_fn
                       ),
                       treeStableTimer=10,
                       properties = {
                         'docker_image' : 'registry.dovecot.net/dovecot:stretch-worker'
                       }
))
```

When I commit changes into dovecot/obox2 master branch, the `obox2-plugin-ci` builder is not triggered, because both schedulers are triggering the same builder. If I change the branch filter to exclude master, then `obox2-plugin-ci` is triggered on master commit.

### LINKED ISSUE #4686: Schedulers does not trigger build if a different scheduler of same type triggered for the same commit
Configuration: v2.0.1
* a GitPoller for the repo URL, and the list of branches
* Each scheduler is initialized with the name of the scheduler, a list of builders, and a ChangeFilter for the branch.
* The schedulers in this case is either a SingleBranchScheduler or a schedulers.Nightly

We have a configuration, among other configurations, with two branches that follow the same commits, with one usually being ahead of the other, or they are on the same commit. The branch names are used to differentiate compile time customization.  Each of these branches have two schedulers,   a SingleBranchScheduler and a schedulers.Nightly.

Expected behavior is that when one of the branches is move to a new commit, new builds are triggered for the relevant branch's new commit.

Actual behavior is that if one of the schedulers of a type have triggered a build for that commit, the other scheduler of the same type, for the other branch, will NOT trigger a build until the branch contains at least one commit that is not yet  part of the first scheduler's branch.

It seems like buildbot is triggering builds based of unique groups of (repository, scheduler_kind, commit) rather than the expected (repository, scheduler_name, branch, commit) group.

This seems to be a regression compared to the previous Chromium-developed 0.8.4-based buildbot we used before.

One possibility that might be useful to investigate is to check which name is used when associating the triggered commits with the scheduler. Is it the master.cfg configured name, or the default name member of the class? Note that the correct name is used in the properties for the triggered build.

p12tic:
This PR is a collection of 3 bug fixes in the default build request collapsing functionality. The collapsed build requests are not seen anywhere in the UI, so to the end users it would look like as if the schedulers were the broken piece.

Fixes #4686. Fixes #3762. Fixes #4642.

## Contributor Checklist:

* n/a I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/db/model.py
+++ b/master/buildbot/db/model.py
@@ -63,6 +63,9 @@ class Model(base.DBConnectorComponent):
     #
     # * sqlalchemy does not handle sa.Boolean very well on MySQL or Postgres;
     #   use sa.SmallInteger instead
+    #
+    # * BuildRequest.canBeCollapsed() depends on buildrequest.id being auto-incremented which is
+    #   sqlalchemy default.
 
     # Tables related to build requests
     # --------------------------------
--- a/master/buildbot/newsfragments/buildrequest-collapse-check-scheduler-props.bugfix
+++ b/master/buildbot/newsfragments/buildrequest-collapse-check-scheduler-props.bugfix
@@ -1,0 +1,1 @@
+Improved default build request collapsing functionality to take into account properties set by the scheduler and not collapse build requests if they differ (:issue:`4686`).
--- a/master/buildbot/newsfragments/buildrequest-collapse-ensure-claim.bugfix
+++ b/master/buildbot/newsfragments/buildrequest-collapse-ensure-claim.bugfix
@@ -1,0 +1,1 @@
+Fixed a race condition that would result in attempts to complete unclaimed buildrequests (:issue:`3762`).
--- a/master/buildbot/newsfragments/buildrequest-collapse-fix-cancel-each-other.bugfix
+++ b/master/buildbot/newsfragments/buildrequest-collapse-fix-cancel-each-other.bugfix
@@ -1,0 +1,1 @@
+Fixed a race condition in default buildrequest collapse function which resulted in two concurrently submitted build requests potentially being able to cancel each other (:issue:`4642`).
--- a/master/buildbot/process/buildrequest.py
+++ b/master/buildbot/process/buildrequest.py
@@ -254,6 +254,11 @@ class BuildRequest:
         return buildrequest
 
     @staticmethod
+    def filter_buildset_props_for_collapsing(bs_props):
+        return {name: value for name, (value, source) in bs_props.items()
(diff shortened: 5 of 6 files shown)


mwild1:
Thanks for this! :heart:

PR #6156, opened 2021-07-29
Fix support for SQLAlchemy 1.4

### LINKED ISSUE #5992: compatibility with sqlalchemy 1.4
Hi,

It would be great if buildbot gets compatibility with 1.4. In PR #5912 the constraints has simply been set to <1.4 but this is unfortunately not a solution for distro packaging of buildbot.

Is there an estimated timeline for compatibility with 1.4 as this issue is blocking distro efforts?

p12tic:
Fixes #5992.

We should plan to migrate off sqlalchemy-migrate to Alembic as it uses more and more deprecated APIs that will be soon removed.

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -56,6 +56,7 @@ env:
 
   # Test different versions of SQLAlchemy
   - TWISTED=17.9.0 SQLALCHEMY=1.2.0 TESTS=trial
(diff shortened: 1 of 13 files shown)


tardyp (approved):
cool! I though this would be much more difficult.

great work! ship it!

PR #6174, opened 2021-08-06
Fix diverse doc quirks in the 'Launching worker as Windows service' section

levitte:


master/docs/manual/installation/misc.rst
```diff
@@ -115,33 +115,55 @@ Upgrade pip, setuptools and wheel
    cd c:\Python39
    python.exe -m pip install --upgrade pip setuptools wheel
 
-Install, create and activate virtualenv
-```````````````````````````````````````
+Install, create and activate virtualenv [optional]
+``````````````````````````````````````````````````
+This step is optional and may depend on your needs.
```
tardyp, line 120: well, we highly recommend to use this and avoid messing up with your python install.
a virtualenv can easily be recreated, while a python install, it is much more annoying

master/docs/manual/installation/misc.rst
```diff
@@ -115,33 +115,55 @@ Upgrade pip, setuptools and wheel
...
 
 .. code-block:: bat
 
-   python.exe -m pip install virtualenv==16.7.10
-   Scripts\virtualenv.exe --no-site-packages c:\bbw\sandbox
-   cd c:\bbw
+   python.exe -m pip install virtualenv>=1.7
```
tardyp, line 124: shouldn't we quote the '>'?

master/docs/manual/installation/misc.rst
```diff
@@ -115,33 +115,55 @@ Upgrade pip, setuptools and wheel
...
+```````````````
 
 .. code-block:: bat
 
   python.exe -m pip install pywin32
+
+If you are in an activated virtualenv, finish up like this:
+
+.. code-block:: bat
+
```
tardyp, line 138: with modern pywin32, this is not needed anymore.

see the CI for best known method:

https://github.com/buildbot/buildbot/blob/master/appveyor.yml#L21

tardyp (changes requested):
a few nitpicks

master/docs/manual/installation/misc.rst
levitte: Hmm, the main installation page expresses it more as an option depending on comfort: https://github.com/buildbot/buildbot/blob/master/master/docs/manual/installation/installation.rst#installation-in-a-virtualenv

I thought this was being more consistent...

master/docs/manual/installation/misc.rst
levitte: Ah, very good point

master/docs/manual/installation/misc.rst
levitte: Ok, then I'll drop this part

tardyp:
Well that main doc part might have been written at a point where virtualenv
was less mainstream. Now, unless you are using docker or building dedicated
VM, virtualenv should be preferred for building/deploying python apps

Le dim. 8 août 2021 à 12:19, Richard Levitte ***@***.***> a
écrit :

levitte:
> Well that main doc part might have been written at a point where virtualenv was less mainstream. Now, unless you are using docker or building dedicated VM, virtualenv should be preferred for building/deploying python apps

I think that the mystery for me is how the service knows that it should run within a virtual environment.  I can't see that the virtual environment is activated within the service, so I assume it would mean that it uses the system installation (whether that's in a system directory or in the buildbot user's AppData doesn't matter)...  what am I missing?

levitte:
> > Well that main doc part might have been written at a point where virtualenv was less mainstream. Now, unless you are using docker or building dedicated VM, virtualenv should be preferred for building/deploying python apps
> 
> I think that the mystery for me is how the service knows that it should run within a virtual environment. I can't see that the virtual environment is activated within the service, so I assume it would mean that it uses the system installation (whether that's in a system directory or in the buildbot user's AppData doesn't matter)... what am I missing?

Never mind, I should have looked first...

https://github.com/buildbot/buildbot/blob/a6ef0219aa51dc6dd4cf28480e8b2f91ad22b018/worker/buildbot_worker/scripts/windows_service.py#L146-L168

levitte:
The more I look at this, the more I find that the "Launching worker as Windows service" section essentially repeats what's already described in [`master/docs/manual/installation/worker.rst`](https://github.com/buildbot/buildbot/blob/master/master/docs/manual/installation/worker.rst) and [`master/docs/manual/installation/installation.rst`](https://github.com/buildbot/buildbot/blob/master/master/docs/manual/installation/installation.rst), it could as well refer to those, since creating the buildbot worker itself shouldn't differ.

All that the "Launching worker as Windows service" section really needs is the instruction how to set up the service itself.
(and install pywin32, but I imagine it would be better to make that a required install module on win32, i.e. a modification of appropriate `setup.py`, no?)

levitte:
#6177 is an alternative PR.  I think I'll simply drop this on in favor of that one.

levitte:
> #6177 is an alternative PR. I think I'll simply drop this on in favor of that one.

Said and done

PR #6200, opened 2021-08-17
ec2: Set AMI for latent worker before making spot request

Conan-Kudo:
When trying to set up latent workers where the AMI is dynamically
discovered as part of instantiating workers, we want to make sure
that AMI is set before we make the spot request so that it is used
for the spot request.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/newsfragments/enable-ami-runtime-determinism.bugfix
+++ b/master/buildbot/newsfragments/enable-ami-runtime-determinism.bugfix
@@ -1,0 +1,1 @@
+Set AMI for latent worker before making spot request to enable dynamically setting AMIs for instantiating workers
--- a/master/buildbot/worker/ec2.py
+++ b/master/buildbot/worker/ec2.py
@@ -448,6 +448,7 @@ class EC2LatentWorker(AbstractLatentWorker):
         log.msg('%s %s requesting spot instance with price %0.4f' %
                 (self.__class__.__name__, self.workername, bid_price))
 
+        image = self.get_image()
         reservations = self.ec2.meta.client.request_spot_instances(
             SpotPrice=str(bid_price),
             LaunchSpecification=self._remove_none_opts(
@@ -473,7 +474,6 @@ class EC2LatentWorker(AbstractLatentWorker):
             raise LatentWorkerFailedToSubstantiate()
         instance_id = request['InstanceId']
         self.instance = self.ec2.Instance(instance_id)
-        image = self.get_image()
         instance_id, start_time = self._wait_for_instance()
         return instance_id, image.id, start_time
 


Conan-Kudo:
cc: @tonyhutter

tonyhutter:
@freqlabs would you mind taking a look at this?  I believe it's your patch from https://github.com/openzfs/zfs-buildbot/pull/233#issuecomment-895355730.  I'm not clear if the patch works on it's own in buildbot without the additional changes in zfs-buildbot.

If it does work on it's own, that's a good thing :+1: since we wouldn't need to carry around the patch in zfs-buildbot.

freqlabs (approved):
It should be harmless. We've only been overriding get_image for the FreeBSD workers, and the Linux workers do work. I've visually checked if anything has drastically changed since the 0.8.13~ish version currently in production and it seems okay, but I don't have anything set up to actually test this myself.

tonyhutter (approved):

Conan-Kudo:
@tardyp Could you please take a look and see if it could be merged?

tardyp:
Hello, I don't think it is harmful to get the image here. 

I am just worried that your method of overriding the self.ami in get_image() is not really documented and will get forgotten in future development. So we cannot garantee your usecase will work forever.

I would prefer if you can provide an official way of choosing image/ami, which is at the moment not supported/documented/tested.

pipecc (approved):

Conan-Kudo:
> Hello, I don't think it is harmful to get the image here.
> 
> I am just worried that your method of overriding the self.ami in get_image() is not really documented and will get forgotten in future development. So we cannot garantee your usecase will work forever.
> 
> I would prefer if you can provide an official way of choosing image/ami, which is at the moment not supported/documented/tested.

I don't know how to design such a thing, but I think we should come up with something to replace the hack mechanism. I would _really_ appreciate this being merged for now because we _need_ this for the OpenZFS Buildbot. Or an alternative mechanism implemented, but that goes back to I don't know how to design this...

tardyp:
The best way is to make some parameters renderables. There are some examples in for example the docker or kubernetes latent worker that you can look at.
As I am not sure exactly what you implemented in your get_image, it is best that the actual users of the latent workers are contributing it.

freqlabs:
Our workers inherit from `EC2LatentWorker` and take a `get_image` parameter to `__init__` that is used as follows:
https://github.com/openzfs/zfs-buildbot/blob/97b22564c60921c5385644ea7d6eea752927513e/master/zfsworkers.py#L238-L244

Here are the implementations of our custom `get_image` functions:
https://github.com/openzfs/zfs-buildbot/blob/97b22564c60921c5385644ea7d6eea752927513e/master/master.cfg#L506-L524

Making `ami` renderable instead would be neat. It's certainly a more invasive change, though.

PR #6205, opened 2021-08-24
fix docker build

tardyp:
@p12tic not sure how this could have ever worked. 

Is there something I missed?

### DIFF
--- a/pkg/buildbot_pkg.py
+++ b/pkg/buildbot_pkg.py
@@ -205,9 +205,13 @@ class BuildJsCommand(distutils.cmd.Command):
 
             yarn_program = None
             for program in ["yarnpkg", "yarn"]:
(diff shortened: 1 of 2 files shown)


pkg/buildbot_pkg.py
```diff
@@ -205,9 +205,13 @@ def run(self):
 
             yarn_program = None
             for program in ["yarnpkg", "yarn"]:
-                yarn_version = check_output([program, "--version"], shell=shell)
-                if yarn_version != "":
-                    yarn_program = program
+                try:
```
tardyp, line 208: checkoutput raise in case of failure, which would be the case for at least one of those two.

p12tic: Right, I have both installed...

p12tic (approved):

PR #6217, opened 2021-09-01
Allow RemoveWorkerFileSecret to handle renderables

hsorby:
## Contributor Checklist:


p12tic (changes requested):
Looks good to me, thanks. Please add a newsfragment. I don't think a test is necessary in this case.

PR #6223, opened 2021-09-07
Remove dependency on blockdiag

p12tic:
funcparserlib requires setuptools older than 58 because it still uses 2to3 functionality which has been removed. This breaks installation of Buildbot on setuptools newer than 58. funcparserlib is required by blockdiag which we use for automatic generation of diagrams in our docs. This PR switches the diagrams to svg and removes the dependency.


master/docs/_images/Makefile
```diff
@@ -1,5 +1,5 @@
 
-SOURCES = overview_src.svg changes_src.svg workers_src.svg master_src.svg
+SOURCES = overview_src.svg changes_src.svg workers_src.svg master_src.svg multimaster_src.svg auth_rules_src.svg
```
rjarry, line 2: I did not understand what are these `_src` files for. Could you explain?

master/docs/developer/www-data-module.rst
```diff
@@ -10,12 +10,6 @@ Its main purpose is to handle the 3 way binding.
 2 way binding is the angular MVVM concept, which seamlessly synchronise the view and the model.
 Here, we introduce an additional way of synchronisation, which is from the server to the model.
 
-.. blockdiag::
-
-    blockdiag {
-      View <- Model <- Server;
-    }
-
```
rjarry, line 18: You did not replace this with an image?

master/docs/developer/www-data-module.rst
```diff
@@ -44,13 +38,6 @@ For example, the Change wrapper decodes the author name and email from the "auth
 
 Each wrapper class also has specific access methods, which allow to access more data from the REST hierarchy.
 
-.. blockdiag::
-
-    blockdiag {
-      data.getBuilds -> Collection -> Builds -> b.getSteps -> Collection -> Steps;
-    }
-
-
```
rjarry, line 53: You did not replace this with an image?

rjarry (changes requested):

master/docs/_images/Makefile
p12tic: src files are for editing with inkscape. It contains additional information needed for e.g. editing of text. It's later exported to svg which contains text outline that can no longer be edited via inkscape.

master/docs/developer/www-data-module.rst
p12tic: Too lazy as explained in the message of the first commit. Let's use svg from your PR here.

master/docs/_images/Makefile
rjarry: ok thanks

master/docs/developer/www-data-module.rst
rjarry: :+1:

rjarry (approved):
looks good to me :+1:

PR #6234, opened 2021-09-19
newsfragments: Move the newsfragments to the root of the project

p12tic:
Navigating to newsfragments directory and creating a file there is necessary for each new feature or bugfix. Being hidden in the directory tree wastes some seconds for everyone who needs to look there. To avoid this let's move the directory to the root of the project.


tardyp:
cool! at the time of towncrier adoption, this option wasn't available.

PR #6236, opened 2021-09-19
Message formatter improvements

p12tic:
This PR adds a way to limit logs presented to reporters to just the metadata, add urls to logs to that metadata and adds documentation for whole context dictionary.

## Contributor Checklist:

### DIFF
--- a/master/docs/manual/configuration/report_generators/formatter.rst
+++ b/master/docs/manual/configuration/report_generators/formatter.rst
@@ -48,12 +48,12 @@ The constructor of the class takes the following arguments:
 
 ``want_logs``
     This parameter (defaults to False) will extend the content of the steps of the given ``build`` object with the log metadata of each steps from the build.
-    This requires ``wantSteps`` to be True.
+    This implies ``wantSteps`` to be `True`.
     Use it only when mandatory, as this greatly increases the overhead in terms of CPU and memory on the master.
 
 ``want_logs_content``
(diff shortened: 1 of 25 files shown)


master/docs/manual/configuration/report_generators/formatter.rst
```diff
@@ -44,17 +44,106 @@ The constructor of the class takes the following arguments:
     Use it only when necessary as this increases the overhead in terms of CPU and memory on the master.
 
 ``wantLogs``
-    This parameter (defaults to False) will extend the content of the steps of the given ``build`` object with the full logs of each steps from the build.
-    This requires ``wantSteps`` to be True.
+    Deprecated, use ``want_logs`` and ``want_logs_content`` set to the same value.
+
+``want_logs``
```
tardyp, line 49: shouldn't we also deprecate the want* in favor of want_*?

p12tic: I thought that this would be an unnecessary change. If I have a need to change wantSteps semantics before Buildbot 4.0 we can do that then without another deprecation.

tardyp: I think it would be best to at least try to be coherent on the parameter name within a single API. I think people can get that during transition, naming can be a bit incoherent, but within the same class, I feel it awkward.

p12tic: Fixed now.

tardyp (approved):

PR #6239, opened 2021-09-21
Adding ability to use `--filter` in `git clone`

jsonar-cpapke:
Adding a new parameter to the Git step for adding filters to the git clone step. The main advantage of this is adding the ability to do either `--filter=tree:0` or `--filter=blob:none` which both reduce the amount of data downloaded during a copy.

The version 2.27 was used as that is the earliest mention I found of `--filter` existing when looking in the docs https://git-scm.com/docs/git-clone/2.27.0

For more information about how the tree or blob filter work, you can look at https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

## Contributor Checklist:


p12tic (approved):
Thanks!

PR #6238, opened 2021-09-21
Allow setting io_timeout on GitPoller subprocesses

Sanqui:
This fixes Buildbot startup for huge repos.

We were having trouble starting a fresh Buildbot instance for our existing repository.  The repository is nearing 20GB and cloning it from scratch takes many minutes.  Since the default `io_timeout` on the `git fetch` subprocess call is 300 seconds, this resulted in Buildbot getting stuck in a loop like the following:

```
2021-09-21 09:37:58+0000 [-] <RunProcess '['git', 'fetch', '[repourl]', ...]'>: killing process because <RunProcess '['git', 'fetch', '[repourl]', ...]>: command timed out: 300 seconds without output
2021-09-21 09:37:58+0000 [-] <RunProcess '['git', 'fetch', '[repourl]', ...]'>: killing process using TERM
2021-09-21 09:38:03+0000 [-] <buildbot.changes.gitpoller.GitPoller object at 0x7ff56cced0a0>: while polling for changes
	Traceback (most recent call last):
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 701, in errback
	    self._startRunCallbacks(fail)
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 764, in _startRunCallbacks
	    self._runCallbacks()
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 858, in _runCallbacks
	    current.result = callback(  # type: ignore[misc]
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1751, in gotResult
	    current_context.run(_inlineCallbacks, r, gen, status)
	--- <exception caught here> ---
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1657, in _inlineCallbacks
	    result = current_context.run(
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/twisted/python/failure.py", line 500, in throwExceptionIntoGenerator
	    return g.throw(self.type, self.value, self.tb)
	  File "/home/build/buildbot-configuration/.venv/lib/python3.9/site-packages/buildbot/changes/gitpoller.py", line 250, in poll
	    yield self._dovccmd('fetch', [self.repourl] + refspecs,
[...]

PR #6247, opened 2021-09-28
Fix GitPoller fetch commands timing out on huge repositories

Sanqui:
This pull request supersedes #6238 and fixes the same problem in a simpler and more general way, please check that issue for details.

Long story, short, git fetch would timeout on huge repositories, making Buildbot not start properly.  By adding the `--progress` to the command, git will continuously output into `stderr` even if no tty is present.  Therefore, Buildbot will not kill the command due to an IO timeout, allowing the fetch to complete.

## Contributor Checklist:

No update needed for this bugfix.


p12tic (approved):

p12tic:
Looks great, thanks!

Issue #6262, opened 2021-10-13
MsBuild141 failure


Gugli:
When using the MsBuild141 buildstep, it always fails with : 

```
" do "%I\\VC\Auxiliary\Build\vcvarsall.bat" x86 && msbuild "VisualStudio\Nadeo.sln" /p:Configuration="Release" /p:Platform="x64" /maxcpucount /t:"Project" was unexpected at this time.
```
An unmatched double quote has been introduced in the command string by 7fb40b52529a52cd19220fd3d40ed6cd2df8eb5b.

PR #6263, opened 2021-10-13
Bugfix : an extra quote made the command line invalid

Gugli:
Fixes https://github.com/buildbot/buildbot/issues/6262

### DIFF
--- a/newsfragments/msbuild141.bugfix
+++ b/newsfragments/msbuild141.bugfix
@@ -1,0 +1,1 @@
+Fix a bug in the MsBuild141 step : an extra double-quote made the worker command fail to parse.


Gugli:
Also : 
I've made some local changes in my deployment to allow MsBuild141 to :
- handle Defines with the /p:DefineConstants msbuild param
- allow the Clean or Rebuild of a specific project with the /t:ProjectName:Rebuild and /t:ProjectName:Clean syntax

Do you think these features have a chance to be merged. If so : should I make one or two Pull Requests ? (It's like 20 lines of code in vstudio.py)

p12tic:
>  I've done my best, but I'm not a native English speaker, so I'm not sure I'm qualified to redact such texts.

No problem, at least we have something so that it's not forgotten during the release.

> Do you think these features have a chance to be merged. If so : should I make one or two Pull Requests ? (It's like 20 lines of code in vstudio.py)

I think these make sense, you can do one PR, but please split the changes into two commits. Thanks!

PR #6270, opened 2021-10-14
Restore old resource names in the data API

### LINKED ISSUE #6264: problem displaying logs on browser with buildbot 3.3.1-dev109
We have problem displaying logs in the browser
https://gkernelci.gentoo.org/#/builders

p12tic:
Looks like we'll need a little bit more complex handling for graphql because old resource names are exposed via the REST API. Currently I just broke the GraphQL API and restored what was in the past. @tardyp

Looking further I think the best would be to have two sets of identifiers: one for v2 API, another for GraphQL. This way everything would be clear and easy to understand, even if a lot of data would be duplicated.

Fixes #6264.

### DIFF
--- a/master/buildbot/data/properties.py
+++ b/master/buildbot/data/properties.py
@@ -94,8 +94,8 @@ class PropertiesListEndpoint(base.Endpoint):
 
 class Property(base.ResourceType):
 
-    name = "property"
-    plural = "properties"
+    name = "_property"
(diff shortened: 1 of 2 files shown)


tardyp:
gaah. I tried to be very careful not to brake old api, but indeed did not think about this.

okay to merge it even if we need to sort out and remove the need for `_properties` in graphl queries.

PR #6273, opened 2021-10-15
Release notes for 3.4.0

p12tic:


p12tic:
> Looks good. I think we should deploy that to metabuildbot for a while to make sure we don't have more regression due to all those changes.

Fair enough.

Issue #6251, opened 2021-10-05
Only a single build request is considered for scheduling if nextWorker returns None


arbor-bdoyle:
We're utilizing a custom nextWorker function in order to assign each build to a unique hardware resource that tests are run against. We have several pools of these with different configurations, and the pool to pull a resource from is selected based on the build request properties. Since a given resource can only be utilized by a single build's worth of tests at a time, sometimes a build request will be submitted and not have one available. In this case, nextWorker returns None in order to kick the request back into the queue for consideration later.

I have observed via watching buildbot's log output that it only seems to consider the first build request in the queue each time it tries to distribute builds. This means that even if there are requests in the queue which use a different pool with resources available, they will be stuck waiting until the pool required by the request at the head of the queue has a resource available.

After some digging, I believe this is caused by the following code: https://github.com/buildbot/buildbot/blob/1cfba0346239d04f1fa45f597dfc21e93039e522/master/buildbot/process/buildrequestdistributor.py#L184-L187

It seems like the intent here was to skip the request if there are no workers which can build it, but instead it is bailing out on checking any of them. Further down in this method it checks over all the workers again in order to run canStartBuild, and that chunk appears to do the correct thing (continuing around the outer loop again) if no worker can be selected. Should the initial check just be eliminated, and combined into the loop below?

It also seems like build requests may be dropped entirely if canStartBuild fails for all workers in this later loop, since the request was removed on line 190, but I haven't verified this theory and I'm not sure if the orphaned request is added back by some process elsewhere.

(Side note on my particular use case: It seems like locks on the resources would be the better solution, but I think there's a separate issue involving a race with latent docker workers checking locks and then starting up which causes all the workers to get jammed in the acquiring locks phase. This results in effectively the same jamming because all the workers are consumed waiting for the same resource. I haven't had time to investigate this race yet. Hence, nextWorker as a workaround.)

tardyp:
I agree that the code looks weird, with a loop that is only meant to be run once.

I can see that even if we would continue, _getNextUnclaimedBuildRequest() will any return the same buildrequest so we would retry forever the same buildrequest, unless nextBuild is smarter that it needs and account that the build was just chosen and couldn't be assigned a worker.

I think the whole API design is flawed in the failure modes indeed.

What would make more sense at first sight is a prioritizeBuild hook, that would sort the buildrequests in whatever order desired by the project, and then the while loop would consider those builds one by one.

[...]

PR #6277, opened 2021-10-20
Reconsider nextWorker for each build request

arbor-bdoyle:
This is a proposed fix for #6251 which maintains the existing nextWorker API. I haven't yet added unit tests to catch the issue or updated docs, but will do so before removing the draft status from this PR if this approach looks good.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/buildrequestdistributor.py
+++ b/master/buildbot/process/buildrequestdistributor.py
@@ -144,21 +144,30 @@ class BuildChooserBase:
 
 
 class BasicBuildChooser(BuildChooserBase):
-    # BasicBuildChooser generates build pairs via the configuration points:
-    #   * config.nextWorker  (or random.choice if not set)
-    #   * config.nextBuild  (or "top" if not set)
-    #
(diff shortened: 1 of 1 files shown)


p12tic:
> Is there an example of doing a perf test somewhere or would this be a one-off thing?

We currently don't have normal performance tests that could be easily ran. I was more expressing an opinion of what needs to be done, not necessarily by you.

I think you can continue working on the other things and we can look into what options are for testing performance after you've finished.

fxddps2:
Hello, folks!

I've stumbled upon the same issue and wonder if this draft is working as intended? I don't care about potential performance penalties if this solves the current issue of having effectively 1 buildrequest in the queue, which is simply horrible.
My current buildbot version is 3.7.0 (upped from 2.5.0) and this behavior is still present.

vilhalmer:
I've just remembered this PR is still sitting here, and wanted to drop by and say it will never get done. 😅 I lost access to my arbor Github account during a mail server migration and have since left the company. It was stable internally for at least a year, but the project using Buildbot was ultimately abandoned so I never had the time to clean up this work.

PR #6283, opened 2021-10-28
Fix for SecretInPass provider with gpgPassphrase

### LINKED ISSUE #4953: SecretInPass: gpg asks for passphrase even if already provided in configuration
When using `buildbot.secrets.providers.passwordstore.SecretInPass`, gpg prompts for a password even when one already given to `SecretInPass`. 

For example, even if the passphrase is given in `master.cfg`:

```python
c['secretsProviders'] = [
    SecretInPass(gpgPassphrase="mypassphrase",
                           dirname="/path/to/password/store"),
]
```
gpg will still prompt for the passphrase.

**Cause**
The problem is caused by:

https://github.com/buildbot/buildbot/blob/v2.3.1/master/buildbot/secrets/providers/passwordstore.py#L50-L55

The problem can be solved using the solutions given in: [gpg asks for password even with --passphrase](https://unix.stackexchange.com/q/60213/217242).

WXbet:
We need this addional parameter to make it working, fixes #4953

## Contributor Checklist:

### DIFF
--- a/master/buildbot/secrets/providers/passwordstore.py
+++ b/master/buildbot/secrets/providers/passwordstore.py
@@ -48,7 +48,7 @@ class SecretInPass(SecretProviderBase):
     def reconfigService(self, gpgPassphrase=None, dirname=None):
         self._env = {**os.environ}
         if gpgPassphrase:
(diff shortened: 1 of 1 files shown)


p12tic (approved):
Thanks! 

Code looks good to me, what remains is updates to tests and a release note in a news fragment.

PR #6292, opened 2021-11-06
Fix an error in passing CMake options and definitions to cmake.

sound-fx:
Fix an error in passing `CMake` options and definitions on the cmake command line.  The cmake usage documentation (see below) dictates that the options shall precede the \<path-to-source> (and also \<path-to-build>).

The new unit test `test_options_path` checks for this.

> cmake --help
> Usage
> 
>   cmake [options] \<path-to-source>
>   cmake [options] \<path-to-existing-build>
>   cmake [options] -S \<path-to-source> -B \<path-to-build>
> 

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/cmake.py
+++ b/master/buildbot/steps/cmake.py
@@ -70,15 +70,15 @@ class CMake(ShellMixin, BuildStep):
                 '-G', self.generator
             ])
 
-        if self.path:
-            command.append(self.path)
-
         if self.definitions is not None:
             for item in self.definitions.items():
                 command.append('-D%s=%s' % item)
 
         if self.options is not None:
             command.extend(self.options)
+
+        if self.path:
+            command.append(self.path)
 
         cmd = yield self.makeRemoteShellCommand(command=command)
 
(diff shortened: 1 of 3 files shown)


p12tic:
@sound-fx 

For unit test check master/buildbot/test/unit/steps/test_cmake.py, it contains tests for path and for options separately, but not at the same time. It shouldn't be hard to use the existing tests as an example and figure out a test that tests both.

p12tic:
The code itself looks good, thanks.

sound-fx:
> For unit test check master/buildbot/test/unit/steps/test_cmake.py, it contains tests for path and for options separately, but not at the same time. It shouldn't be hard to use the existing tests as an example and figure out a test that tests both.

Thank you for the suggestion.  I'm happy to add a test for this, but it will be at least a month, maybe longer, before I can get to this.  (I can't divert attention from some looming deadlines.)

p12tic:
Thanks for the unit test. I've rebased the PR on top of latest master.

p12tic:
Fixed pylint & flake8.

PR #6293, opened 2021-11-09
Discontinue use of deprecated teams API in Bitbucket OAuth client

klhowell:
Atlassian has deprecated the 'teams' APIs in favor of new
'workspaces' APIs and will soon discontinue support for 'teams'
endpoints. This change produces equivalent functionality using
the recommended 'workspaces' endpoints.

See:
https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-teams-deprecation/


p12tic:
@klhowell Thanks, looks good to me.

Could you please add `workspaces` to common/code_spelling_ignore_words.txt so that spell checker doesn't complain? Thanks!

klhowell:
> @klhowell Thanks, looks good to me.
> 
> Could you please add `workspaces` to common/code_spelling_ignore_words.txt so that spell checker doesn't complain? Thanks!

@p12tic  It looks like 'workspaces' is already in that file. Let me know if I'm misunderstanding.

klhowell:
Okay, it should be in there now.

PR #6304, opened 2021-12-04
replace client with APIClient

ligi:
otherwise you get this error: AttributeError: module 'docker.client' has no attribute 'Client'


p12tic (approved):

ligi:
Is the pipeline somehow stuck? On the site I still see the old version: https://docs.buildbot.net/current/manual/configuration/workers-docker.html

PR #6310, opened 2021-12-12
Allow auto-update of base docker images

PeterSurda:
- alwaysPull in docker latent worker will now not only apply when using `image`, but also when using a `dockerfile`. In that case it will try to update the base image. E.g. if your `dockerfile` contains `FROM ubuntu:focal ...`, it will always try to update the `ubuntu:focal` image, instead of just pulling only if it's missing

- I'm soliciting feedback, I tried to use a minimalistic approach and not sure what all else I have to do for such a small change

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -48,8 +48,10 @@ class Client:
     def wait(self, id):
         return 0
 
-    def build(self, fileobj, tag):
+    def build(self, fileobj, pull, tag):
         if fileobj.read() == b'BUG':
+            pass
+        elif pull != bool(pull):
             pass
         else:
             logs = []
--- a/master/docs/manual/configuration/workers-docker.rst
+++ b/master/docs/manual/configuration/workers-docker.rst
@@ -242,7 +242,8 @@ In addition to the arguments available for any :ref:`Latent-Workers`, :class:`Do
 
 ``alwaysPull``
     (optional, defaults to false)
-    Always pulls image if autopull is set to true.
+    Always pulls (update) image if autopull is set to true.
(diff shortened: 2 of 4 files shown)


master/docs/manual/configuration/workers-docker.rst
```diff
@@ -242,7 +242,7 @@ In addition to the arguments available for any :ref:`Latent-Workers`, :class:`Do
 
 ``alwaysPull``
     (optional, defaults to false)
-    Always pulls image if autopull is set to true.
+    Always pulls (update) image if autopull is set to true. Also affects the base image specified by `FROM ....` if using a dockerfile, autopull is not needed then.
```
p12tic, line 245: One line per sentence.

p12tic:
Just one nitpick, otherwise looks good.

master/docs/manual/configuration/workers-docker.rst
PeterSurda: Ok can do.

master/buildbot/test/fake/docker.py
```diff
@@ -48,9 +48,11 @@ def stop(self, id):
     def wait(self, id):
         return 0
 
-    def build(self, fileobj, tag):
+    def build(self, fileobj, pull, tag):
         if fileobj.read() == b'BUG':
             pass
+        elif pull != bool(pull):
```
tardyp, line 54: not sure what this is about

PeterSurda: I think it was complaining that `pull` is unused or some other complaint. This seemed like the cleanest fix, if the value is valid, it will pretend an image is created, otherwise it will pretend it failed to create an image. At least that's how I understand it.

PR #6314, opened 2021-12-25
Add `target` parameter for docker build

PeterSurda:
- allows to select stage when using multistage build
- it's a renderable so can be set dynamically

Similar to my earlier PR, #6310. I followed the documentation to docker API, which states the default value is `""`.

Should I write a separate test for it? Since the unit tests run against a fake docker library, I'm not sure how useful it would be.

PS. at this time, I haven't tested it in production yet, but I tested #6310 and it seems to be working.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -50,10 +50,12 @@ class Client:
     def wait(self, id):
         return 0
 
-    def build(self, fileobj, pull, tag):
+    def build(self, fileobj, tag, pull, target):
         if fileobj.read() == b'BUG':
             pass
         elif pull != bool(pull):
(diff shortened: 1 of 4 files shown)


p12tic (approved):

p12tic:
Thanks for the pull request. It looks good, could you rebase on top of latest master?

We probably need a proper unit test here, but the risk of this particular commit breaking something are low, so let's not require a unit test for now.

PeterSurda:
> Thanks for the pull request. It looks good, could you rebase on top of latest master and prepend `docker:` to the commit message?

Like this?

p12tic:
> Like this?

Yes, thanks.

PR #6319, opened 2022-01-03
tests: Use fake connection for step tests instead of fake remote command

p12tic:
Faking at a lower level will allow us to test the behavior of real RemoteCommand in more cases. Currently we use fake remote command which tries to mimic the real remote command badly. 

Fortunately running remote commands only depends on worker connection for running the commands so we can use
a fake connection and intercept sent commands at that level.

## Contributor Checklist:


p12tic:
Appveyor test fails due to new deprecation warnings which should be fixed by https://github.com/buildbot/buildbot/pull/6320. I will ignore them for this PR.

PR #6323, opened 2022-01-06
Experimental msgpack protocol

mokibit:
This PR introduces experimental MessagePack (msgpack) protocol.
Master-worker communication protocol should better not be dependent on specific language. Currently Perspective Broker protocol uses Twisted networking engine which is written in Python. MessagePack is compact and simple and there is official implementation for variety of languages. 
Using MessagePack protocol will allow implementation of Buildbot worker in different languages.


p12tic:
cc @tardyp. 

The new protocol is essentially a translation of what we have on pb protocol to msgpack encoding. There are many things that can be improved since we are not bound by backwards compatibility, but the amount of work is already large, so I think it's best to merge what we have and then iterate.

Improvements that I've thought about:
 - drop the internal worker-side builder concept and just send commands to the worker that should be executed. The worker shouldn't care what builder a command is for.
 - clean up how updates are sent. There was a dict logical encoding problem that caused one update to be able to contain one type of messages. We can do better now.
 - send logs as lines so that master does not need to split them. We store logs as lines to the database anyway.
 - attach time information to log lines.

Each of the above will require a new revision of the protocol.

master/buildbot/worker/protocols/manager/msgpack.py
```diff
@@ -0,0 +1,413 @@
...
+
+                if username in self.get_dispatcher().users:
+                    password, afactory = self.get_dispatcher().users[username]
+                    h = hash(hash(password) + self.challenge)
+                    self.worker_name = username
+                    if h == msg['hash']:
+                        result = True
+                        try:
+                            self.connection = yield afactory(self, username)
+                        except Exception as e:
```
p12tic, line 122: Was this some debugging code?

mokibit: Fixed.

master/buildbot/util/misc.py
```diff
@@ -51,3 +51,74 @@ def writeLocalFile(path, contents, mode=None):  # pragma: no cover
...
+            dict_kwargs[elements_eq[0]] = elements_eq[1]
+        elif len(elements_eq) <= 1:
+            list_args.append(elements_eq[0])
+        else:
+            raise Exception('unsupported porstr {}'.format(portstr))
+    return (list_args, dict_kwargs)
+
+
+def get_host_port(input):
+    # return a tuple (host, port)
```
tardyp, line 71: why something custom?
twisted already has the concept of connection string. shouldn't we ust use that?

master/docs/developer/master-worker-msgpack.rst
```diff
@@ -0,0 +1,1290 @@
...
+.. note::
+    This is experimental protocol.
+=================================================================
+
+Messages between master and worker are sent using WebSocket protocol in both directions.
+Data to be sent conceptually is a dictionary and is encoded using MessagePack.
+One such encoded dictionary corresponds to one WebSocket message.
+
+A message can be either a request or a response.
+Request message is sent when one counter party wants another one to perform an action.
```
tardyp, line 11: counter party? never heard of this term. should we just use 'side'?

master/docs/developer/master-worker-msgpack.rst
```diff
@@ -0,0 +1,1290 @@
...
+    Otherwise, its value is a boolean ``True`` and the message of exception is specified in the value of ``result``.
+
+
+interrupt_command
+~~~~~~~~~~~~~~~~~
+
+Request
++++++++
+
+This message requests worker to halt the current command.
```
tardyp, line 318:
```suggestion
This message requests worker to halt the specified command.
```

master/docs/developer/master-worker-msgpack.rst
```diff
@@ -0,0 +1,1290 @@
...
+
+    If this dictionary contains "PYTHONPATH" key, path separator and "$PYTHONPATH" is appended to that value.
+
+    Resulting environment dictionary sent to the command is created folowing these rules:
+
+    1) If ``env`` has value for specific key and it is ``None``, resulting dictionary does not have this key.
+
+    2) If ``env`` has value for specific key and it is not ``None``, resulting dictionary contains this value with substitutions applied.
+
+    Any matches of a pattern ``${name}`` in this value, where name is any number of alphanumeric charachters, are substituted with the value of the same key from worker environment.
```
tardyp, line 833:
```suggestion
    Any matches of a pattern ``${name}`` in this value, where name is any number of alphanumeric characters, are substituted with the value of the same key from worker environment.
```

master/docs/developer/master-worker-msgpack.rst
```diff
@@ -0,0 +1,1290 @@
...
+    Maximum number of bytes to transfer from the master.
+    The operation will fail if the file exceeds this size.
+    Worker will request data from master until it notices it exceeded ``maxsize``.
+
+``blocksize``
+    Value is an integer.
+    It represents maximum size for each data block to be sent from master to worker.
+
+``mode``
+    Value is ``None`` or an integer which represents an access mode for the new file.
```
tardyp, line 1014: I guess we should just refer to unix documentation

tardyp (commented):
Hello,
I am very happy to see this work.
This is actually very impressive.

I have put some remarks of what I like a bit less.
- the self made auth makes me worry I would prefer stick on HTTP. I don't understand all the attacks possible on CR based auth, so I am not sure what to look at. I think BasicAuth over https is just the basic state of the art. 
- The connection parsing string. I am not sure why this complexity. as we are over websocket, can't we just use standard http(s) url?
- the documentation could use some factorization. I see a lot of common parameters
- Its very big. Usually I prefer to recommend sending smaller PRs. I haven't read all in detail.
- looks like the tests are failing is it still WIP?

tardyp:
some additional comments to answer @p12tic 

> The new protocol is essentially a translation of what we have on pb protocol to msgpack encoding. There are many things that can be improved since we are not bound by backwards compatibility, but the amount of work is already large, so I think it's best to merge what we have and then iterate.

I agree on this approach. I think we dont need to have incremental version unless someone really need to put this in prod early


> drop the internal worker-side builder concept and just send commands to the worker that should be executed. The worker shouldn't care what builder a command is for.

I think the current implement only allows one command per builder. I don't say this was a good idea, but a thing to keep in mind.

> clean up how updates are sent. There was a dict logical encoding problem that caused one update to be able to contain one type of messages. We can do better now.

right

> send logs as lines so that master does not need to split them. We store logs as lines to the database anyway. attach time information to log lines.

That will help a lot on the scalability. We could even compress it directly on worker to save bandwith and master CPU.
This would require decompression in case of observer, but this is still less work than compression.

> add protocol version and additional metadata such as password hashing algorithms to the first message exchange between worker and master.

We have to be very careful here as the first package is unauthenticated attack surface.

> Each of the above will require a new revision of the protocol.

Is it really worth it?

master/buildbot/worker/protocols/base.py
```diff
@@ -57,19 +57,22 @@ def updateRegistration(self, username, password, portStr):
...
+            # inform the manager, logging any problems in the deferred
+            accepted = yield self.master.workers.newConnection(conn, workerName)
+
+            # return the Connection as the perspective
+            if accepted:
+                return conn
+            else:
+                # TODO: return something more useful
+                raise RuntimeError("rejecting duplicate worker")
+        except Exception as e:
```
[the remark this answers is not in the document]
mokibit: Fixed.

master/buildbot/worker/protocols/manager/msgpack.py
```diff
@@ -0,0 +1,411 @@
...
+
+from buildbot.util import deferwaiter
+from buildbot.util import misc
+from buildbot.util.eventual import eventually
+from buildbot.worker.protocols.manager.base import BaseDispatcher
+from buildbot.worker.protocols.manager.base import BaseManager
+
+
+def challenge():
+    # return: Some random data.
```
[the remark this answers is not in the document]
mokibit: Authentication logic has been copied from Twisted. I wanted to make the authentication as similar as possible to what is already used by perspective broker.
I agree that this is not perfect and HTTP authentication will be implemented in the final version.

master/buildbot/worker/protocols/manager/msgpack.py
```diff
@@ -0,0 +1,411 @@
...
+    challenge = b''
+    for x in range(random.randrange(15, 25)):
+        challenge = challenge + chr(random.randint(65, 90)).encode('utf-8')
+    challenge = str(md5(challenge).digest())
+    return challenge
+
+
+def hash(password):
+    md = md5()
```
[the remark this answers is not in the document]
mokibit: This has been copied from Twisted, full explanation in the comment above.

master/buildbot/worker/protocols/manager/msgpack.py
```diff
@@ -0,0 +1,411 @@
...
+        try:
+            self.contains_msg_key(msg, ('username', 'hash'))
+            username = msg['username']
+            try:
+                dispatcher = self.get_dispatcher()
+                yield dispatcher.master.initLock.acquire()
+
+                if username in dispatcher.users:
+                    password, afactory = dispatcher.users[username]
+                    h = hash(hash(password) + self.challenge)
```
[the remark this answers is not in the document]
mokibit: This has been copied from Twisted, full explanation in the comment above.

master/buildbot/worker/protocols/manager/msgpack.py
```diff
@@ -0,0 +1,411 @@
...
+
+        self.maybe_log_master_to_worker_msg(dict_output)
+        payload = msgpack.packb(dict_output, use_bin_type=True)
+
+        self.sendMessage(payload, isBinary=True)
+
+    def onMessage(self, payload, isBinary):
+        if not isBinary:
+            name = self.worker_name if self.worker_name is not None else '<???>'
+            log.msg('Message type form worker {} unsupported'.format(name))
```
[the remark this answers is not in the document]
mokibit: Fixed.

master/buildbot/util/misc.py
mokibit: In msgpack protocol worker needs host and port for establishing connection. This function parses both from Twisted connection string so that we can build connection string usable for autobahn.

With @p12tic we discussed that it may make sense to change API so that user specifies host, port and other information explicitly. Hence, this function would not be needed.

master/docs/developer/master-worker-msgpack.rst
mokibit: Fixed.

master/docs/developer/master-worker-msgpack.rst
mokibit: Fixed.

master/docs/developer/master-worker-msgpack.rst
mokibit: Fixed.

master/docs/developer/master-worker-msgpack.rst
mokibit: Buildbot only supports a subset of Unix modes and only those are explained in this documentation.
Referring to full Unix documentation would cause confusion for the user because it would be unclear whan exactly Buildbot supports.

mokibit:
Hello, @tardyp,

thanks for your comments and advises, much appreciated!  

> the self made auth makes me worry I would prefer stick on HTTP. I don't understand all the attacks possible on CR based auth, so I am not sure what to look at. I think BasicAuth over https is just the basic state of the art.

Authentication logic has been copied from Twisted. I wanted to make the authentication as similar as possible to what is already used by perspective broker.
I agree that this is not perfect and HTTP authentication will be implemented in the final version.

> The connection parsing string. I am not sure why this complexity. as we are over websocket, can't we just use standard http(s) url?

In msgpack protocol worker needs host and port separately for establishing connection. This function parses both from Twisted connection string so that we can build connection string usable for autobahn.

With @p12tic we discussed that it may make sense to change API so that user specifies host, port and other information explicitly. Hence, this function would not be needed in the future.

> the documentation could use some factorization. I see a lot of common parameters

I kept in mind that factorization and referencing the text makes it more difficult and time-consuming for the users to read the documentation and to find the needed sections. 
For me it is always easier to have the needed information in one block especially if it is new information. This is important in the specification like this, which should prefer clarity even if it causes some duplication.

> Its very big. Usually I prefer to recommend sending smaller PRs. I haven't read all in detail.

In regards to that I made a separate commit for protocol specification file.
I believe it is not possible to break the protocol itself into several commits because it will not be functioning code.

> looks like the tests are failing is it still WIP?

PR is now completed and is ready for proper review.

tardyp:
Quick feedback

> In msgpack protocol worker needs host and port separately for establishing connection. This function parses both from Twisted connection string so that we can build connection string usable for autobahn.

I don't think you answer my main remark, why don't we just https://host:port
and then just convert to whatever autobahn/twisted is needing?

for me this will be just easier for the user to have standard URL

> Authentication logic has been copied from Twisted.

Make sense, but pb's design is very old and predates ssl, so reproducing whatever pb was doing is not for me the way to go. There is a lot of code/doc to support this method. I think it should just go away. If you wish to merge before http auth is implemented, that's fine to me.
Just remove the CR auth and let the worker tell its name without any auth.

mokibit:
> I don't think you answer my main remark, why don't we just https://host:port and then just convert to whatever autobahn/twisted is needing?

> for me this will be just easier for the user to have standard URL

I fixed this, now users can pass URL.


> Make sense, but pb's design is very old and predates ssl, so reproducing whatever pb was doing is not for me the way to go. There is a lot of code/doc to support this method. I think it should just go away. If you wish to merge before http auth is implemented, that's fine to me.
Just remove the CR auth and let the worker tell its name without any auth.

CR auth is removed, now worker only sends its name and password in plain text. HTTP authentication will be implemented later. 

I think PR is ready for a full review.

mokibit:
@tardyp ping.

tardyp (approved):
I think we could improve a bit more the code coverage by adding a bit more tests.  I approve it anyway as it will be easier to improve that gradually.

mokibit:
>I think we could improve a bit more the code coverage by adding a bit more tests. 

Integration tests have been added.

PR #6340, opened 2022-01-20
refactoring code with Assign Multiple Targets

zjzh:
refactoring code with Assign Multiple Targets which is more pythonic, concise, readable and efficient;how do think this change which has practical value？
## Remove this paragraph

And especially:

## Contributor Checklist:


p12tic:
I think the new version of the code is actually less readable, because one now needs to count which value corresponds to which member.

PR #6342, opened 2022-01-24
process: Use yield for log finish method

vibbo:
Just a minor fix. Addresses Unhandled Error:
```
    File "/home/user/bbm/sandbox/lib/python3.7/site-packages/buildbot/process/log.py", line 101, in finish
	    assert not self._finishing, "Did you maybe forget to yield the method?"
	builtins.AssertionError: Did you maybe forget to yield the method?
```


p12tic:
Thanks for the PR, looks good. I added a newsfragment and rebased on top of latest master.

Unit test would be very difficult to write in this case and it would have only limited usefulness.

PR #6343, opened 2022-01-25
Bugfix authz tests and rebuild

vibbo:


p12tic:
Thanks, the PR looks good. I've rebased on top of latest master and fixed an unrelated issue that could cause crashes when a build being accessed does not exist.

Issue #5988, opened 2021-04-20
build_wait_timeout option in Latent worker not working as expected


BishopGIS:
Buildbot version 3.1.0 both master and latent worker (worker.LibVirtWorker).

After upgrade from 2.4 to 3.1 the builds using latent worker begin to stop in about 10 minutes.
It seems to me that build_wait_timeout now starts timer not after the build task finished but from the moment from start of latent worker.
Changing build_wait_timeout to several hours fix the problem (quick fix).

Also in documentation there is no mention to set build_wait_timeout in seconds (http://docs.buildbot.net/current/manual/configuration/workers.html?highlight=build_wait_timeout#common-options).

First I tried to set 20 min (build_wait_timeout=20) and I got restarting latent worker every 20s.

BishopGIS:
The problem with builds was because daylight saving on LibVirtWorker. There was 1 hour difference leads such behaviour.
Disabling this in worker OS fixed problem. 
So bug label can be removed.

p12tic:
Thanks for the bug report, this issue has been fixed in #6344.

PR #6344, opened 2022-01-25
fix race condition in build wait timers for substantiated workers

RyanDraves:
Patched an issue where latent workers that have already been substantiated will start a `build_wait_timer` during substantiation of a new build. This timer starts after the `build_wait_timer` is cleared from a new build arriving, disconnecting the worker `build_wait_timeout` seconds later, mid-build or not.

## Contributor Checklist:

  - added a nice comment

### DIFF
--- a/master/buildbot/worker/latent.py
+++ b/master/buildbot/worker/latent.py
@@ -236,7 +236,6 @@ class AbstractLatentWorker(AbstractWorker):
             return defer.succeed(False)
 
         if self.state == States.SUBSTANTIATED and self.conn is not None:
-            self._setBuildWaitTimer()
             return defer.succeed(True)
 
         if self.state in [States.SUBSTANTIATING,
--- a/newsfragments/latent-worker-build-wait-timer.bugfix
+++ b/newsfragments/latent-worker-build-wait-timer.bugfix
@@ -1,0 +1,1 @@
+Fixed handling of ``build_wait_timeout`` on latent workers which previously could result in latent worker being shut down while a build is running in certain scenarios (:issue:`5988`).


master/buildbot/worker/latent.py
```diff
@@ -236,7 +236,9 @@ def substantiate(self, wfb, build):
             return defer.succeed(False)
 
         if self.state == States.SUBSTANTIATED and self.conn is not None:
-            self._setBuildWaitTimer()
+            # Skip the wait timer if a build is already queued
+            if not self.building:
```
p12tic, line 240: This line is equivalent to deleting `self._setBuildWaitTimer()`. self.building checks whether at least one of WorkerForBuilder instances associated with the worker are busy (via `isBusy`). They will necessarily be busy, because at least one got its `buildStarted` called in the path to `Build.startBuild` which eventually indirectly calls `LatentWorker.substantiate()`.

However, this also means that `_setBuildWaitTimer` is completely unnecessary here. If `buildStarted` is guaranteed to be called when we are here, then regardless of what happens `buildFinished` will be called too for that particular build. That is the place for `_setBuildWaitTimer` which we already do there.

p12tic: All in all, thanks for the PR, it identified an important bug and did 95% of the work fixing it. I'll finish the polishing.

RyanDraves: Works for me, thanks. Do you think `self.building` / `isBusy` should be a part of the state machine? I found that lack of information in the worker's state to be terribly confusing to debug, which led to me exploring way too much of the call tree to find the root issue.

p12tic:
This PR fixes #5988.

master/buildbot/worker/latent.py
p12tic: `self.building` is quite separate from the internal state machine in that the states aren't dependent on each other. If we merge them then the state machine would grow double in size.

I agree that the latent worker itself is quite confusing. It became such once all edge cases got covered. It seems that there should be a way to refactor everything for simplicity, but nobody found any such way yet.

PR #6352, opened 2022-02-02
add note/comment handling to gitlab hook

Try2Code:
hi!

I would like to improve the integration with gitlab. ATM buildbot is not able to react on merge request comments although gitlab can be configured to send out notifications about them (its rest API uses the term 'note' for this). 
The code below does add note handling, but not limited to merge requests. Notes to a normal issue trigger the hook and hence the buildbot. But since there is now context (repo, commit) a normal testing procedure (build,make,make check) is not possible.

What I want to know is: Is this change interesting? Does is make sense for buildbot? I  am aware of the plugin feature, so I could write a plugin and don't have to bother you. Trigger buildbot by merge request comments seems a pretty usable feature though. and there already IS a gitlab hook - so why write a plugin when the code can be improved.

Please keep in mind that this in only a first version. I'd prefer to integrate the changed into the existing merge request handler. But I am not sure if this violates some design decision I am not aware of.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/test_buildbot_net_usage_data.py
+++ b/master/buildbot/test/unit/test_buildbot_net_usage_data.py
@@ -151,5 +151,5 @@ class Tests(unittest.TestCase):
         self.assertEqual(len(distro), 2)
         self.assertNotIn("unknown", distro[0])
         # Rolling distributions like Arch Linux (arch) does not have VERSION_ID
-        if distro[0] not in ["arch", "gentoo"]:
+        if distro[0] not in ["arch", "gentoo", "antergos"]:
             self.assertNotIn("unknown", distro[1])
--- a/master/buildbot/test/unit/www/test_hooks_gitlab.py
+++ b/master/buildbot/test/unit/www/test_hooks_gitlab.py
@@ -796,6 +796,216 @@ gitJsonPayloadMR_open_forked = b"""
 }
 """  # noqa pylint: disable=line-too-long
 
+# == Merge requests from a fork of the project
+# (Captured more accurately than above test data) captured from gitlab/v4 rest api
+gitJsonPayloadMR_commented = b"""
+{
+  "object_kind": "note",
+  "event_type": "note",
+  "user": {
+    "id": 343,
+    "name": "Name Surname",
+    "username": "rollo",
+    "avatar_url": "null",
+    "email": "[REDACTED]"
+  },
+  "project_id": 926,
+  "project": {
+    "id": 926,
+    "name": "awesome_project",
+    "description": "",
+    "web_url": "https://gitlab.example.com/mmusterman/awesome_project",
+    "avatar_url": null,
+    "git_ssh_url": "git@gitlab.example.com:mmusterman/awesome_project.git",
+    "git_http_url": "https://gitlab.example.com/mmusterman/awesome_project.git",
+    "namespace" : "mmusterman",
+    "visibility_level": 0,
+    "path_with_namespace": "awesome_project",
+    "default_branch": "master",
+    "ci_config_path": null,
+    "homepage": "https://gitlab.example.com/mmusterman/awesome_project",
+    "url": "git@gitlab.example.com:mmusterman/awesome_project.git",
+    "ssh_url": "git@gitlab.example.com:mmusterman/awesome_project.git",
+    "http_url": "https://gitlab.example.com/mmusterman/awesome_project.git"
+  },
+  "object_attributes": {
+    "attachment": null,
+    "author_id": 343,
+    "change_position": {
+      "base_sha": null,
+      "start_sha": null,
+      "head_sha": null,
+      "old_path": null,
+      "new_path": null,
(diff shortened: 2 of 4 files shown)


p12tic:
> Trigger buildbot by merge request comments seems a pretty usable feature though. 

It's a welcome feature. There are already setups that can be configured to be triggered by comments, e.g. Gerrit.

tardyp:
make sense to me as well, thanks! you will need to add some tests..

Try2Code:
thx for the positive feedback @tardyp @p12tic !

My plan is to leave as much as possible to the config file. The problem is that note events get sent as a special event type, but I want to allow note changes in merge requests,only. All others should be ignored so far. 

Note events in a wiki or a general issue could be used for other purposes, but I would like to restrict the addition as much as possible for now.

@tardyp yh, I will work on the tests. the testing environment seems to be beast on its own :see_no_evil:  ...

p12tic:
@Try2Code The config side of this will most likely be handled by change filters. So the webhook should send all changes it receives regardless of whether they are important or not and then change filters that are attached to schedulers should pick the changes they want to start builds on.

Try2Code:
@p12tic  change fitler is how I use the feature atm. I would like to avoid the exiting error message in the log

- merge requests sent without given 'action'
- note events without given merge_request element

Try2Code:
@p12tic @tardyp a starting point for adding test would be very helpful :pray:

Try2Code:
hi again! 

I started with the implementation of the tests, but without any added tests, I get failiures. Followed the instructions on https://buildbot.readthedocs.io/en/latest/developer/quickstart.html to setup the environment.


Errors look a lot like this

```
[FAIL]
  File "/home/ram/src/forks/buildbot/master/buildbot/test/unit/test_buildbot_net_usage_data.py", line 155, in test_linux_distro
    self.assertNotIn("unknown", distro[1])
  File "/home/ram/src/forks/buildbot/.venv/lib/python3.10/site-packages/twisted/trial/_synctest.py", line 503, in assertNotIn
    raise self.failureException(msg or f"{containee!r} in {container!r}")
twisted.trial.unittest.FailTest: 'unknown' in 'unknown_version'

buildbot.test.unit.test_buildbot_net_usage_data.Tests.test_linux_distro
[FAIL]
  File "/home/ram/src/forks/buildbot/master/buildbot/test/unit/www/test_ws.py", line 90, in test_too_many_arguments
    self.assert_called_with_json(
  File "/home/ram/src/forks/buildbot/master/buildbot/test/unit/www/test_ws.py", line 47, in assert_called_with_json
    self.assertEqual(actual_json, expected_json)
  File "/home/ram/src/forks/buildbot/.venv/lib/python3.10/site-packages/twisted/trial/_synctest.py", line 422, in assertEqual
    super().assertEqual(first, second, msg)
  File "/usr/lib/python3.10/unittest/case.py", line 845, in assertEqual
  File "/usr/lib/python3.10/unittest/case.py", line 1144, in assertDictEqual
    self.fail(self._formatMessage(msg, standardMsg))
twisted.trial.unittest.FailTest: {'error': "Invalid method argument 'WsProto[76 chars]': 1} != {'_id': 1, 'code': 400, 'error': "Invalid m[65 chars]o''"}
  {'_id': 1,
   'code': 400,
-  'error': "Invalid method argument 'WsProtocol.cmd_ping() got an unexpected "
?                                     -----------

+  'error': "Invalid method argument 'cmd_ping() got an unexpected keyword "
?                                                                  ++++++++

-           "keyword argument 'foo''"}
?            --------

+           "argument 'foo''"}

buildbot.test.unit.www.test_ws.WsResource.test_too_many_arguments
```

Any idea what these errors related to? maybe a different python release? I use 3.10 here

Try2Code:
First error I could solve. I use an Arch-derived distro. so it behaves like it, but the bb test does not catch it. 
will rebase with master first

Try2Code:
@p12tic @tardyp
I use captured json from my gitlab installation (using webhook testing) for test input. Apparently the test fails because the missing change. 

Any idea, what's going on? I think I don't understand enough about the testing system ...

Try2Code:
I'd love to get some help with the testing.  I thing the existing testing is only for _normal_ MR things like push/create/close/merge, but not for related note events. But those events are created by the gitlab REST API. 

From what I can tell, the exiting MR-testing does not cover note events and it would need something new. Any help is welcome :white_flag:

p12tic:
I'm sorry it took so long to look into the PR. I'm looking into it right now.

I've rebased the PR on top of master.

p12tic:
@Try2Code I've fixed the tests. The problem was that the captured json could not be decoded as it contained a newline character within a string value. 

In case you look into diffs, you'll also I've also changed the project name, author and other strings to match to the rest of captured json data. This is just unimportant nitpick, I wrote it to clear any confusion. 

The code itself looks good to me.

Try2Code:
thank you so much @p12tic :pray: - I was kind of lost in the json test input

PR #6355, opened 2022-02-02
www: Fix window title updates

p12tic:
This PR upstreams a couple of bug fixes for www frontend by @jangmarker.


tardyp (approved):
thank you both!

PR #6356, opened 2022-02-02
gerritchangesource: Strip refs/heads from branch names

p12tic:
Changes should have correct branch name and not e.g. refs/heads/master because this is expected by other components of Buildbot. For example before this commit Nightly schedulers didn't work on the changes emitted by Gerrit change sources.


tardyp:
I fear this will break people's change filters..

PR #6359, opened 2022-02-05
[test PR] check whether codecov reporting works properly

p12tic:


p12tic:
Coverage reports seem to work now.

PR #6361, opened 2022-02-06
codecov: Fix after_n_builds config param

p12tic:
Turns out that Appveyor builds don't upload reports.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,10 +1,10 @@
 codecov:
   require_ci_to_pass: yes
   notify:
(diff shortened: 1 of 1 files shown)


tardyp:
I think they did at some point, as there are windows specific tests we want to take in account for coverage

p12tic:
> I think they did at some point, as there are windows specific tests we want to take in account for coverage

I agree, but for now I just wanted codecov to work reliably and Appveyor is a separate problem that needs to be addressed separately.

PR #6367, opened 2022-02-07
bitbucket: fix PullrequestPoller: GitHub Issue 4153

sjalloq:
- BitbucketPullrequestPoller calls twisted.web.client.getPage without encoding
   the url to bytes.  This fix copies adds a BitbucketPullrequestPoller.getPage
   method that optionally encodes the url before calling client.getPage.
 - test_bitbucket.py has been updated with the fake client.getPage also checking
   that the url argument is a bytes object rather than string.
 - fixed URL to Bitbucket API. All external requests should be directoed to
   api.bitbucket.org

## Contributor Checklist:

* [Y] I have updated the unit tests
* [N] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* [N] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -100,13 +100,18 @@ class BitbucketPullrequestPoller(base.ReconfigurablePollingChangeSource, PullReq
         page = yield self._getChanges()
         yield self._processChanges(page)
 
+    def getPage(self, url, **kwargs):
+        if isinstance(url, str):
+            url = url.encode('utf-8')
+        return client.getPage(url, **kwargs)
+
     def _getChanges(self):
         self.lastPoll = time.time()
         log.msg("BitbucketPullrequestPoller: polling "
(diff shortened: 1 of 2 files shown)


sjalloq:
Note that in between creating this branch and pushing this PR, Twisted has outed a new release that drops support for the deprecated client.getPages().  Not sure if that is what is causing all the errors.

sjalloq:
I've now got an update that uses `txrequests.get()` in the `BitbucketPullrequestPoller.getPage()` method.  It drops web.client.getPage completely and may be easier to update for the relevant http client.  I think the Buildbot httpclient uses txrequests?  If so, I guess we also need to revert the change to the tests as the `url` can be `str` or `bytes` now.

p12tic:
@sjalloq: Could you put your txrequests change into the PR? If it works with real bitbucket API I can update the tests to match the behavior.

sjalloq:
Yeah, I've got it on a separate branch so will close this and re-create the PR.

PR #6369, opened 2022-02-07
bitbucket: fix PullrequestPoller: GitHub Issue: Resolves #4153

### LINKED ISSUE #4153: BitbucketPullrequestPoller fails all requests in python3
`_getChanges` and `_processChanges` all perform `client.getPage` with str arguments, which fail because `client.getPage` expect to receive `bytes` for the `url` parameter.

I worked around the issue locally by replacing all the `client.getPage` calls by `self.getPage` which encodes the url to bytes when necessary, and returns `client.getPage`, but there may be a cleaner way to handle this.

sjalloq:
- BitbucketPullrequestPoller called twisted.web.client.getPage without encoding
   the url to bytes.  However, as of Twisted 22.1.0 the getPage method has been
   removed.  This fix updates BitbucketPullrequestPoller.getPage to use txrequests.
 - fixed URL to Bitbucket API. All external requests should be directoed to
   api.bitbucket.org

## Contributor Checklist:

* [N] I have updated the unit tests
* [N] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* [N] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -14,14 +14,14 @@
 # Copyright Buildbot Team Members
 
 
-import base64
 import json
 import time
 from datetime import datetime
 
+from txrequests import Session
+
 from twisted.internet import defer
 from twisted.python import log
-from twisted.web import client
 
 from buildbot.changes import base
 from buildbot.util import bytes2unicode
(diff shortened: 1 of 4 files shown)


sjalloq:
This commit seems to work with the API from within our cluster so the proxy support in txrequets works out of the box.  Happy for you to knock it into shape to be merged or ask me to sort things out that don't meet expectations.  Thanks.

PR #6404, opened 2022-02-16
master: Add an option to janitor to delete logs for specific builders.

rufinio:
## Contributor Checklist:

There is a need for data from different builders to be kept for different lengths of time before it can be deleted.
This change adds the ability to specify the duration before data is deleted per buildername.


master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
     name = 'LogChunksJanitor'
     renderables = ["logHorizon"]
 
-    def __init__(self, logHorizon):
+    def __init__(self, logHorizon=None, horizonPerBuilder=None):
```
p12tic, line 43: Buildbot wants to gradually move to `snake_case` naming style like the rest of Python ecosystem. Replace all `horizonPerBuilder` with `horizon_per_builder`.

master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
...
         self.logHorizon = logHorizon
+        self.horizonPerBuilder = horizonPerBuilder
 
     @defer.inlineCallbacks
     def run(self):
-        older_than_timestamp = datetime2epoch(now() - self.logHorizon)
-        deleted = yield self.master.db.logs.deleteOldLogChunks(older_than_timestamp)
-        self.descriptionDone = ["deleted", str(deleted), "logchunks"]
+        if self.logHorizon is not None:
+            older_than_timestamp = datetime2epoch(
```
p12tic, line 51: Nit: should fit in a single line I think (Buildbot uses 100 chars per line).

master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
...
+                now() - self.logHorizon)
+            deleted = yield self.master.db.logs.deleteOldLogChunks(
+                older_than_timestamp=older_than_timestamp)
+            self.descriptionDone = ["deleted", str(deleted), "logchunks"]
```
p12tic, line 55: This should be just before return I think, because if both if conditions are true then we would set descriptionDone before step is actually finished.

master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
...
     @defer.inlineCallbacks
     def run(self):
-        older_than_timestamp = datetime2epoch(now() - self.build_data_horizon)
-        deleted = yield self.master.db.build_data.deleteOldBuildData(older_than_timestamp)
-        self.descriptionDone = ["deleted", str(deleted), "build data key-value pairs"]
+        if self.build_data_horizon is not None:
+            older_than_timestamp = datetime2epoch(now() - self.build_data_horizon)
+            deleted = yield self.master.db.build_data.deleteOldBuildData(
+                older_than_timestamp=older_than_timestamp)
+            self.descriptionDone = ["deleted", str(deleted), "build data key-value pairs"]
```
p12tic, line 80: This should be just before return I think, because if both if conditions are true then we would set descriptionDone before step is actually finished.

master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
...
         return SUCCESS
 
 
 class JanitorConfigurator(ConfiguratorBase):
     """ Janitor is a configurator which create a Janitor Builder with all needed Janitor steps"""
 
-    def __init__(self, logHorizon=None, hour=0, build_data_horizon=None, **kwargs):
+    def __init__(self, logHorizon=None, hour=0, build_data_horizon=None, horizonPerBuilder=None,
+    **kwargs):
```
p12tic, line 93: weird indentation

master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
...
         super().__init__()
         self.logHorizon = logHorizon
         self.build_data_horizon = build_data_horizon
+        self.horizonPerBuilder = horizonPerBuilder
         self.hour = hour
         self.kwargs = kwargs
+        if ((self.logHorizon is not None or self.build_data_horizon is not None)
+        and self.horizonPerBuilder is not None):
```
p12tic, line 101: Weird indentation.

master/buildbot/test/unit/db/test_build_data.py
```diff
@@ -226,7 +289,7 @@ def setUp(self):
 
 class TestRealDB(unittest.TestCase,
                  connector_component.ConnectorComponentMixin,
-                 Tests):
+                 RealTests):
```
p12tic, line 292: Ugh :-( Thanks for noticing.

master/buildbot/configurators/janitor.py
p12tic: OK, now I see that only either logHorizon or horizonPerBuilder is not None, not both.

master/buildbot/configurators/janitor.py
```diff
@@ -39,43 +40,67 @@ class LogChunksJanitor(BuildStep):
...
+            bbconfig.error("JanitorConfigurator: horizonPerBuilder only " +
```
p12tic, line 102: I think it makes sense to support all of logHorizon, build_data_horizon, log_horizon_per_builder and build_data_horizon_per_builder at the same time.

Having both global and per-builder log horizons makes sense because it's then possible to have a longer global horizon and shorter per-builder horizons. It also doesn't increase the complexity of the code.

p12tic (changes requested):
Looks great, thanks! I had several comments, most of them very simple to address.

I think it makes sense to support global horizon and per-builder horizons at the same time (see my comment with longer justification).

rufinio:
per-builder horizon offers the possibility of using wildcards. Thus, I think global horizon can be reached by setting name to wildcard.

master/buildbot/configurators/janitor.py
rufinio:
```
    def __init__(self, logHorizon=None, hour=0, build_data_horizon=None, horizonPerBuilder=None,
        **kwargs):
```
leads to flake8 error E125.
<br />
```
    def __init__(self,
                 logHorizon=None,
                 hour=0,
                 build_data_horizon=None,
                 horizon_per_builder=None,
                 **kwargs):
```
should it be like this?

master/buildbot/configurators/janitor.py
rufinio: flake8 complains about my further solutions. Please show me how it should look like.

PR #6453, opened 2022-03-12
util: Allow full URLs for HTTPClientService methods

p12tic:
HTTPClientService currently only supports URLs that are relative to base URL. This is a serious limitation in case of REST APIs that return paths to other resources in the API responses. If the paths are full URLs then it would be simplest if we could just pass the URLs to HTTPClientService unmodified.

## Contributor Checklist:


p12tic:
@tardyp Do you think this change has any security considerations? I think that if an API wants us to fetch a resource from somewhere else we should just do it. An attacker could just serve us a malicious payload in the original API response with the URL, there's no need for additional steps.

PR #6454, opened 2022-03-12
Use httpclientservice in Bitbucket change source

p12tic:
This PR replaces older direct use of txrequests Sessions by httpclientservice service as used in the rest of Buildbot.


p12tic:
@sjalloq Could you please test if this doesn't break BitBucket change source in your setup?

sjalloq:
@p12tic Will take a look on Monday. Thanks.

PR #6460, opened 2022-03-17
Fixed MsBuild141 using incorrect vswhere syntax to retrieve the install location for VS2017

pampersrocker:
## Contributor Checklist:

 * Please provide some proper documentation for your unit test framework, this is just unusable for outsiders, and I tried to use it extensively on one of my plugins.
 * Not applicable

The old command caused an error something like `( was unexpected at this time`, indicating an error in the syntax for the batch command.

### DIFF
--- a/master/buildbot/steps/vstudio.py
+++ b/master/buildbot/steps/vstudio.py
@@ -557,9 +557,10 @@ class MsBuild141(VisualStudio):
         self.descriptionDone = 'built ' + self.describe_project()
         yield self.updateSummary()
 
-        command = ('FOR /F "tokens=*" %%I in (\'vswhere.exe -property  installationPath\') '
-                   f' do "%%I\\%VCENV_BAT%" x86 && msbuild "{self.projectfile}" '
-                   f'/p:Configuration="{self.config}" /p:Platform="{self.platform}" /maxcpucount')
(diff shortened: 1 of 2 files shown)


lefi7z:
I ran into another issue at that line of code. Could you explain, what has actually changed? It's only the `-version [15.0, 16.0)` property that's been included, no?
cheers

pampersrocker:
> I ran into another issue at that line of code. Could you explain, what has actually changed? It's only the `-version [15.0, 16.0)` property that's been included, no? cheers

I see the syntax issue for the command line was already fixed with some newer buildbot versions. The `-version [15.0, 16.0)` is still necessary though, otherwise `vswhere` returns the newest installed version of vstudio. But this being `MSBuild141` we specifically want VS 2017.

lefi7z:
I see, VS 2017 corresponds to version 15 which is then selected. For VS 2019/22 there should be some updated build steps. Also for MsBuild15/16/17.

I would also like to include `-products *` (see PR #6554) so that this also works if only the *Build Tools* are installed (w/o the whole Visual Studio IDE). Because this is how I would set up a minimal build server.

p12tic:
Thanks for the PR and sorry for taking so long to get to it. The changes are included into this PR https://github.com/buildbot/buildbot/pull/6554 which also adds more steps, so I will merge that one.

p12tic:
Apologies for taking so long for review. I'm closing this PR in favor of https://github.com/buildbot/buildbot/pull/6554.

PR #6490, opened 2022-04-20
waterfall test: fix instability on latest version of chromedriver

tardyp:
upgrading the chrome-driver, it looks like those tests have an instability

  - Failed: Error while waiting for Protractor to sync with the page: "both angularJS testability and angular testability are undefined.  This could be either because this is a non-angular page or because your test involves client-side navigation, which can interfere with Protractor's bootstrapping.  See http://git.io/v4gXM for details"
      at /buildbot/test/buildbot/smokes/node_modules/protractor/built/browser.js:463:23
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at noRunningBuilds (/buildbot/test/buildbot/smokes/e2e/pages/home.ts:49:24)Error:
      at ElementArrayFinder.applyAction_ (/buildbot/test/buildbot/smokes/node_modules/protractor/built/element.js:459:27)
      at ElementArrayFinder.<computed> [as getText] (/buildbot/test/buildbot/smokes/node_modules/protractor/built/element.js:91:29)
      at noRunningBuilds (/buildbot/test/buildbot/smokes/e2e/pages/home.ts:49:56)
      at /buildbot/test/buildbot/smokes/node_modules/selenium-webdriver/lib/webdriver.js:938:14
      at /buildbot/test/buildbot/smokes/node_modules/selenium-webdriver/lib/promise.js:2232:22
      at new Promise (<anonymous>)
      at SimpleScheduler.promise (/buildbot/test/buildbot/smokes/node_modules/selenium-webdriver/lib/promise.js:2242:12)
      at SimpleScheduler.execute (/buildbot/test/buildbot/smokes/node_modules/selenium-webdriver/lib/promise.js:2227:17)
      at pollCondition (/buildbot/test/buildbot/smokes/node_modules/selenium-webdriver/lib/promise.js:2195:19)
      at /buildbot/test/buildbot/smokes/node_modules/selenium-webdriver/lib/promise.js:2191:7

## Remove this paragraph

If you don't remove this paragraph from the pull request description, this means you didn't read our contributor documentation, and your patch will need more back and forth before it can be accepted!

Please have a look at our developer documentation before submitting your Pull Request.

http://docs.buildbot.net/latest/developer/quickstart.html

And especially:
http://docs.buildbot.net/latest/developer/pull-request.html

## Contributor Checklist:

* [ ] I have updated the unit tests
[...]

PR #6493, opened 2022-04-24
Some code cleanups

yan12125:
* Remove import tricks for Python 2.x
* Remove a unittest trick for Python < 3.2. Buildbot master already
  requires Python >= 3.6.
* Remove a trick for hmac.compare_digest(), which is available since
  Python 3.3 [1]
* Remove an unused multiprocessing.Process import
* Remove tricks for older Twisted - buildbot master requires Twisted
  17.9.0 or newer
    * The mentioned bug in buildbot.monkeypatches is fixed in Twisted
      17.1 [2]
    * The compatibility class for twisted.logger.Logger is not needed as
      the latter is available since Twisted 15.2.0 [3]

[1] https://docs.python.org/3/library/hmac.html
[2] https://github.com/twisted/twisted/commit/76ca7457621bdbd58f7220ac038bff84f9ef9a1d
[3] https://github.com/twisted/twisted/commit/30eab9904f289f3d85cfaf7a66dc37379da0ea23

(All of them are trivial changes, so I put them all in one commit. Please tell me if multiple commits are better, thanks!)

## Contributor Checklist:

* [ ] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory) - not needed in my opinion
* [ ] I have updated the appropriate documentation - not needed

### DIFF
--- a/master/buildbot/monkeypatches/__init__.py
+++ b/master/buildbot/monkeypatches/__init__.py
@@ -15,7 +15,6 @@
 
 import os
 import unittest
-from builtins import int
 
 from twisted.python import util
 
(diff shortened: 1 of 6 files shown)


yan12125:
I cherry-picked a commit from https://github.com/buildbot/buildbot/pull/6490 to fix failures on GitHub Actions. The only remaining failure `bb/smokes` seems to be related to that PR as well.

p12tic (commented):
LGTM, but could you split each cleanup into separate commits? Right now it's not completely clear what is what. Additionally, larger commits make bisecting harder.

yan12125:
Sure, will rewrite commits before this weekend.

yan12125:
> could you split each cleanup into separate commits?

Done! For the third commit ("Remove a compatibility trick for twisted.logger.Logger"), how about further cleaning up and replacing all `buildbot.util.logger.Logger` usage with `twisted.logger.Logger`?

p12tic:
> how about further cleaning up and replacing all buildbot.util.logger.Logger usage with twisted.logger.Logger

I think this makes sense.

p12tic (approved):
Great PR, thanks.

yan12125:
> I think this makes sense.

Nice! Will do that in another PR.

PR #6494, opened 2022-04-25
Remove worker side builder

mokibit:
This PR removes the need to use worker side builder.
Now  master is in charge of creating needed builder directories in worker by using msgpack commands "mkdir", "rmdir", "stat" and others.


p12tic (approved):

p12tic:
Also rebased on top of latest master.

PR #6497, opened 2022-04-26
Fix reporters issues

g-arjones:
- Gitlab commit API expects the build state to be one of pending, running, success, failed, canceled (not cancelled) (reference: https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit)
- That means the gitlab reporter must differentiate started builds from pending builds
- Generate build reports for canceled build requests (otherwise builds are reported as pending indefinitely)

## Contributor Checklist:


g-arjones:
Just realized that the build status generator doesn't seem to receive buildrequest cancel events... What am I missing?

p12tic (approved):
This is great, thank you!

PR #6501, opened 2022-05-04
Replace buildbot.util.logger.Logger with twisted.logger.Logger

yan12125:
Follow-up https://github.com/buildbot/buildbot/pull/6493#issuecomment-1113179984

## Contributor Checklist:

* [ ] I have updated the unit tests - not needed IMO
      Not sure if this is necessary. `buildbot.util.logger` seems an internal module, but there are uses in the wild: https://www.google.com/search?q=%22buildbot.util.logger%22
* [ ] I have updated the appropriate documentation - not needed IMO


yan12125:
I force-pushed this branch and drop the workaround for the first autobahn issue, which is fixed in 22.4.2. The second issue (conflicts of `config` field) still persists, and I not sure how to fix it.

p12tic:
LGTM, thanks.

yan12125:
Oops, the second commit `Workaround incompatibility with newer autobahn by pinning` is for verifying my assumption (newer autobahn fails some tests) and not for merging as-is...

PR #6509, opened 2022-05-13
wamp: Remove unused config attribute to fix compat with new autobahn

p12tic:
This fixes compatibility with autobahn 22.4.x.


yan12125:
Thanks for the fix! I think https://github.com/buildbot/buildbot/commit/dcbccd20ab9d92025d2b323929088029c5111ab7 can be reverted?

PR #6511, opened 2022-05-13
.bbtravis.yml cleanup

p12tic:
Previously related tests were placed into two test matrices which made whole test setup harder to understand. Combining test matrices into one makes it obvious what configurations we're testing.

This PR also enables 2 additional tests.


p12tic:
Failures on Windows are not related to this PR.

PR #6513, opened 2022-05-13
RFC: Format the code base using black

p12tic:
The current max line length of 99 was preserved. The command that was used to create this commit is as follows:

```
black -l 99 master worker pkg common smokes
```

cc @tardyp


aliceinwire (approved):
looks good to me

Mic92:
We can actually use `ruff` now instead as it could be also re-used for linting. It's compatible with black.

PR #6518, opened 2022-05-14
Rest: Filter buildsets by sourcestamp attributes

jangmarker:
Before implementing unit tests I'd first like to confirm that this approach can go forward. Open questions that I need feedback on:
- Is it ok to encode the sourcestamp objects as JSON in the REST query?
- Is the SQL query approach reasonable?

## Description
Allow the buildsets REST endpoint to filter by sourcestamp attributes.
We use the 'contains' relation to filter by a set of sourcestamp attributes:
`buildsets?sourcestamps__contains={"ssid":26}`

The list of buildsets is then only listing buildsets that have a
sourcetsamp with that ssid. If the sourcestamps__contains filter is used
multiple times it follows the "or" semantics like the other filters. If
a single filter lists multiple attributes these follow and semantics,
within the filter.

The filters are implemented on SQL-level. Necessary joins are only
performed if this filter is actually used.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/data/sourcestamps.py
+++ b/master/buildbot/data/sourcestamps.py
@@ -14,11 +14,14 @@
 # Copyright Buildbot Team Members
 
 
+import json
+
 from twisted.internet import defer
 
 from buildbot.data import base
 from buildbot.data import patches
 from buildbot.data import types
+from buildbot.util import bytes2unicode
 
 
 def _db2data(ss):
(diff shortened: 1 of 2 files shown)


master/buildbot/data/buildsets.py
```diff
@@ -65,7 +65,8 @@ def getSs(ssid):
         'complete_at': 'buildsets.complete_at',
         'results': 'buildsets.results',
         'parent_buildid': 'buildsets.parent_buildid',
-        'parent_relationship': 'buildsets.parent_relationship'
+        'parent_relationship': 'buildsets.parent_relationship',
+        'ssids': 'buildset_sourcestamps.sourcestampid',
```
jangmarker, line 69: This is left over from an earlier approach. I'll remove it.

PR #6538, opened 2022-06-07
docker: call docker_client.close() to prevent connection leaks

tardyp:
In DockerLatentWorker both _thd_start_instance and _thd_stop_instance
use a new docker_client instance that is never cleaned up. This can
cause (ssh) connections to the docker socket to linger for a long
time. Explicitly call docker_client.close() when done with the client.

Also add a mock close def to test/fake/docker.py.

Original work from  mjw:
https://code.wildebeest.org/git/user/mjw/buildbot/?h=docker-py-close

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/docker.py
+++ b/master/buildbot/test/fake/docker.py
@@ -115,6 +115,10 @@ class Client:
     def remove_container(self, id, **kwargs):
         del self._containers[id]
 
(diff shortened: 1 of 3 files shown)


ameenkhan8081 (commented):
****

p12tic (approved):

PR #6552, opened 2022-06-11
Send more that one update in a single update message on msgpack

mokibit:
This PR introduces a new protocol version where update messages are sent not as dictionaries but as a tuple.
This makes it possible to send several command update values in one update message.
Furthermore, splitting the command output into lines is now moved to the worker side to take the load off from the master.
 
### DIFF
--- a/master/buildbot/process/remotecommand.py
+++ b/master/buildbot/process/remotecommand.py
@@ -25,6 +25,7 @@ from buildbot.process import metrics
 from buildbot.process.results import CANCELLED
 from buildbot.process.results import FAILURE
 from buildbot.process.results import SUCCESS
+from buildbot.util import lineboundaries
 from buildbot.util.eventual import eventually
 from buildbot.worker.protocols import base
 
@@ -74,6 +75,7 @@ class RemoteCommand(base.RemoteCommandImpl):
         # This is really only a problem with old-style steps, which do not
         # wait for the Deferred from one method before invoking the next.
         self.loglock = defer.DeferredLock()
+        self._line_boundary_finders = {}
 
     def __repr__(self):
         return f"<RemoteCommand '{self.remote_command}' at {id(self)}>"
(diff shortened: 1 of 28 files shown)


common/pylintrc
```diff
@@ -79,6 +79,7 @@ disable=
   too-many-branches,
   too-many-arguments,
   too-many-locals,
+  too-many-nested-blocks,
```
tardyp, line 82: sounds like a good test to keep, maybe you can just white list the offending lines

master/buildbot/process/remotecommand.py
```diff
@@ -196,6 +198,34 @@ def interrupt(self, why):
...
+                        logname, data = value
+                        whole_line = self.split_line(logname, data)
+                        value = (logname, whole_line)
+                        if whole_line is not None:
+                            self.remoteUpdate(key, value, False)
+                    else:
+                        self.remoteUpdate(key, value, False)
+            except Exception:
+                # log failure, terminate build, let worker retire the update
+                self._finished(Failure())
```
tardyp, line 220: the for will continue. is that intended?

master/buildbot/process/remotecommand.py
```diff
@@ -212,7 +242,20 @@ def remote_update(self, updates):
             # log.msg("update[%d]:" % num)
             try:
                 if self.active and not self.ignore_updates:
-                    self.remoteUpdate(update)
+                    for key, value in update.items():
```
tardyp, line 245: sounds like the exact same code as earlier. needs refactor

tardyp (changes requested):

common/pylintrc
mokibit: Fixed, good point.

master/buildbot/process/remotecommand.py
mokibit: Fixed.

master/buildbot/process/remotecommand.py
mokibit: It is intended.

A subsequent commit will move splitting the lines on msgpack to worker side, so earlier code, which now seems to be repeating, will be removed. Only in PB protocol splitting the lines will be on master side.

mokibit:
@tardyp, thanks for review, I have addressed all your comments.

mokibit:
@tardyp, thanks for review, I have addressed all your comments.

p12tic (approved):

PR #6554, opened 2022-06-21
Let 'vswhere.exe' search for all products including BuildTools.

lefi7z:
On dedicated servers, one would usually not install the full Visual Studio SDK, but only the command line `BuildTools`. I had the issue that the `steps.MSBuild` would not find the installation path if it wasn't configured to search for all Visual Studio products.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/steps/vstudio.py
+++ b/master/buildbot/steps/vstudio.py
@@ -557,7 +557,8 @@ class MsBuild141(VisualStudio):
         self.descriptionDone = 'built ' + self.describe_project()
         yield self.updateSummary()
 
-        command = ('FOR /F "tokens=*" %%I in (\'vswhere.exe -property  installationPath\') '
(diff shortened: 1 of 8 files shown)


yan12125:
Looks like more places need `-products *`. For example: https://github.com/buildbot/buildbot/blob/d9a9827bafd0815b768ccb8016c25f22b2d9eb6b/master/buildbot/test/unit/steps/test_vstudio.py#L940

p12tic:
Thanks for the PR and sorry for long review. I have rebased it on top of latest master, combined several commits fixing the same thing into one and added a release note. This will be merged once tests pass.

p12tic:
Seems like github does not know about one of the author emails, so it shows my avatar as the author here. This is incorrect, I'm only the committer (which is not avoidable due to rebasing). The git repo itself contains correct author information.

PR #6559, opened 2022-07-04
Active Directory workaround

ulrich-teichert-kk:
I recently had to connect a buildbot instance to an Active Directory and I needed a workaround to get it done, which I'd like to share. The problem is caused by Active Directory returning additional search result references which can not be suppressed by query parameters (e.g. https://superuser.com/questions/1438270/why-does-ad-return-searchresref-ldap-references-for-every-ldap-search). This leads to an exception when doing the user lookup.
I am sure that there are more elegant ways to filter out the unwanted search references than I did ;-) The unit tests ran fine without any changes before and after my change.

* [N/A ] I have updated the unit tests

### DIFF
--- a/master/buildbot/process/remotecommand.py
+++ b/master/buildbot/process/remotecommand.py
@@ -204,15 +204,10 @@ class RemoteCommand(base.RemoteCommandImpl):
             for key, value in updates:
                 if self.active and not self.ignore_updates:
                     if key in ['stdout', 'stderr', 'header']:
-                        whole_line = self.split_line(key, value)
-                        if whole_line is not None:
-                            self.remoteUpdate(key, whole_line, False)
+                        self.remoteUpdate(key, value[0], False)
                     elif key == "log":
                         logname, data = value
-                        whole_line = self.split_line(logname, data)
-                        value = (logname, whole_line)
-                        if whole_line is not None:
-                            self.remoteUpdate(key, value, False)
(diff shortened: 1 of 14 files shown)


master/buildbot/www/ldapuserinfo.py
```diff
@@ -113,6 +113,13 @@ def thd():
                               attributes=[
                                   self.accountEmail, self.accountFullName] +
                               self.accountExtraFields)
+            # active directory delivers references in addition to the search
+            # results - filter them out
+            tmp = []
+            if len(res) > 1:
+                for elem in res:
+                    if elem['type'] == 'searchResEntry':
+                        tmp.append(elem)
```
tardyp, line 122: do you mean to do res = tmp after the loop?

tardyp (changes requested):
Please add a unit test to check your use case does not break in future refactor. (and to check that you code actually works as expected :) )

master/buildbot/www/ldapuserinfo.py
ulrich-teichert-kk: Right <blushes>, seems I was too sloppy when making the patch. Will Fix.

ulrich-teichert-kk: Well, now the missing line is... below our conversation and easy to miss... Sorry for that.

PR #6564, opened 2022-07-11
Fix `retryFetch` and `clobberOnFailure`.

kax-slamcore:
As per http://docs.buildbot.net/latest/manual/configuration/steps/source_git.html#
>  retryFetch (optional, default: False)
>
>    If true, if the git fetch fails, then Buildbot retries to fetch again instead of failing the entire source checkout.

However, this does not currently work.

In `Git._fetchOrFallback`, the call to `self._fetch(None)` will throw a `BuildStepFailed` if the fetch command fails. The fallback code in `_fetchOrFallback` will not be executed and the fetch will not be retried. Same for `clobberOnFailure=True` if it is requested.

This change modifies `_fetch` to accept an `abandonOnFailure` flag to control whether it should throw.

## Test Plan

In the first commit I've added tests to test_source_git.py for the `retryFetch=True` case that fail, demonstrating the issue. 
The second commit fixes the failing tests.

## Contributor Checklist:

* [ ] I have updated the appropriate documentation - :thinking: Not sure this is needed. Please advise.

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -305,7 +305,7 @@ class Git(Source, GitStepMixin):
         return self.workdir
 
     @defer.inlineCallbacks
-    def _fetch(self, _):
+    def _fetch(self, _, abandonOnFailure=True):
         fetch_required = True
 
         # If the revision already exists in the repo, we don't need to fetch.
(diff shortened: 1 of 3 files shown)


p12tic (approved):

p12tic:
Replaced the PR by https://github.com/buildbot/buildbot/pull/6587 so that I could rebase the commits on latest master.

p12tic:
This PR got merged via https://github.com/buildbot/buildbot/pull/6587. Thanks!

PR #6577, opened 2022-07-23
Send master newlines indexes and times of log output in msgpack

mokibit:
This PR introduces new functionality of the worker: now it sends more information about command "update" output messages to the master.
Not only the text of output or file contents is sent, but also additional information about those lines: indexes of newline positions in the text and specific times each output line was received. This will be needed for more informative master logs.
Worker also now takes the load of from master, by splitting the output contents into lines of appropriate size.

### DIFF
--- a/worker/buildbot_worker/util/lineboundaries.py
+++ b/worker/buildbot_worker/util/lineboundaries.py
@@ -23,56 +23,100 @@ log = Logger()
 
 class LineBoundaryFinder:
 
-    __slots__ = ['partial_line', 'warned']
-    # split at reasonable line length.
-    # too big lines will fill master's memory, and slow down the UI too much.
-    MAX_LINELENGTH = 4096
+    __slots__ = ['max_line_length', 'partial_line', 'warned', 'time']
     # the lookahead here (`(?=.)`) ensures that `\r` doesn't match at the end
     # of the buffer
     # we also convert cursor control sequence to newlines
     # and ugly \b+ (use of backspace to implement progress bar)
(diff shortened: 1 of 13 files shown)


p12tic (approved):

worker/buildbot_worker/util/lineboundaries.py
```diff
@@ -23,56 +23,100 @@
...
+
+            line = text[first_position: (position + 1)]
+            ret_lines.append(line)
+            ret_line_count += 1
+            ret_text_length = ret_text_length + len(line)
+            ret_indexes.append(ret_text_length)
+            first_position = position + 1
+
+        position = len(text)
+        while position - first_position >= self.max_line_length:
```
tardyp, line 81: I am worried about the complexification of this algorithm.
I can't see exactly why the algorithm was rewritten.

text management in python must be treated very carefully in python as python is not very performant.

Here the algorithm seems much less performant, and the good reason is not obvious to me. at least it deserves a comment

worker/buildbot_worker/util/lineboundaries.py
```diff
@@ -23,56 +23,100 @@
...
+
+        if ret_text == '':
+            return None
+
+        return (ret_text, ret_indexes, line_times)
+
+    def get_lf_positions(self, text):
+        lf_position = 0
+        lf_positions = []
+        while lf_position != -1:
```
tardyp, line 111: re.find_all might be more performant

tardyp (commented):
log processing must be very performant. even if it is in the worker. badly performing log processing could impact the load on the worker.

PR #6592, opened 2022-08-08
updated path of master.cfg to match the tutorial

alexxroche:
It feels like this tutorial was updated from a previous tutorial, and bb-master is residual from that.

## Contributor Checklist:

* [x] I have not updated the unit tests
* [x] I have not created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* [x] I **have** updated the appropriate documentation


tardyp:
thanks!

PR #6594, opened 2022-08-10
Require Twisted>=18.7.0

p12tic:
Recent treq requires Twisted 18.7.0. Additionally, on python3.9 unit tests are no longer working when using Twisted 17.9. It's simplier to upgrade to a new Twisted than to investigate how to support so old Twisted versions.


p12tic:
Looks like CircleCi is not picking our branches. Merging anyway, as it's unlikely issue with code.

PR #6605, opened 2022-08-20
Initial version of www frontend written in React

p12tic:
This is a work-in-progress PR that shows the initial www frontend written using React and Typescript.

Currently around half of the pages have been rewritten. This is enough to demonstrate the viability of the low-level interfaces to the APIs exposed by Buildbot, which was the bulk of complexity due to how different React and AngularJS frameworks are.

The UI is only accessible in development mode using `npm start` from `www/react-base` directory.


www/react-base/src/components/BuildSticker/__snapshots__/BuildSticker.test.tsx.snap
```diff
@@ -0,0 +1,155 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
tardyp, line 1: Not sure those snapshot achieve anything. I don't remember we had UI regression that this would detect.
Maintaining it  means you have to regen them all the time.

Usually, test fixtures are good for reviewers to make sure the test generate appropriate data, but here reading the html do not show you really how it looks like.

www/react-base/src/data/ReactUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+import {createContext, useContext, useEffect, useRef} from "react";
+import DataClient from "./DataClient";
+import {IDataAccessor} from "./DataAccessor";
+import {IDataCollection} from "./DataCollection";
+
+// The default value is not used as the context is injected
+export const DataClientContext =
+  createContext(new DataClient(undefined as any, undefined as any));
+
+export function useDataAccessor() {
```
tardyp, line 27: looks cool, but I think it deserves some documentation :)

www/react-base/src/components/BuildSummaryTooltip/BuildSummaryTooltip.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const buildRegexes = useMemo(
+    () => baseUrls.map(url => new RegExp(`${url}#builders/([0-9]+)/builds/([0-9]+)$`)),
+    [baseUrls]);
+
+  return {
+    buildrequest: buildrequestRegexes,
+    build: buildRegexes,
+  };
+}
+
```
tardyp, line 82: This file looks a bit complicated. I think it deserves some factorization

tardyp (commented):
Looks like huge work!

I am not sure I totally get how the live update works.

www/react-base/src/data/ReactUtils.ts
p12tic: Agreed, this PR is bare minimum that's needed for the code to work _at all_. Documentation will come later :)

www/react-base/src/components/BuildSummaryTooltip/BuildSummaryTooltip.tsx
p12tic: Agreed. Currently everything is a direct rewrite from AngularJS side. I first tried to make it work and I will clean everything later.

www/react-base/src/components/BuildSticker/__snapshots__/BuildSticker.test.tsx.snap
p12tic: This was effectively a rewrite of the unit tests we had on the AngularJS implementation side. There we checked explicitly if X element has member Y and so on. Snapshots seem to be better to cover this kind of output because one doesn't need to write the output comparison code at all. The check is for the generated HTML output, not how it would look like. Consequently, tests like these are only useful for low-level components. This also means that we wouldn't need to regenerate the snapshots all that often. I fully agree that for higher-level components snapshot tests don't make sense.

p12tic:
Live update works as follows:

 - We subscribe to a REST query with WebSocket updates via a call to `useDataApiQuery` which returns an instance of `DataCollection`.
 - `DataCollection` tracks what data we have received so far. Initially it will be zero, once the REST query returns it will contain some elements and then it will be updated via WebSocket connection.
 - We use `DataCollection` within the React component as a normal data container. For example, take all elements out of the `DataCollection` and draw as many rows in table as needed.
 - Internally `DataCollection` uses `mobx` state management library to tell React when it needs to re-render components due to new data having arrived.

Issue #3535, opened 2017-08-16
buildbot stop is not reliable in buildbot 0.9


aj062:
If have noticed multiple times that buildbot stop never finish. Running 'buildbot stop' again results in Error "twisted.internet.error.ReactorNotRunning: Can't stop reactor that isn't running". Only option left at that time is to "kill -9 pid".

'buildbot stop' should be more reliable.

For example, below are logs from a test instance (single master) which was not running any build at the time of 'buildbot stop'. However, 'buildbot stop' never finished, issuing another 'buildbot stop' didn't help either.

```
2017-08-16 09:42:23-0700 [-] Received SIGTERM, shutting down.
2017-08-16 09:42:23-0700 [-] doing housekeeping for master 1 build.hostname.com:[path_here]
2017-08-16 09:42:24-0700 [-] Initiating clean shutdown
2017-08-16 09:42:24-0700 [-] No running jobs, starting shutdown immediately
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
	Failure: exceptions.RuntimeError: Triggerable scheduler stopped before build was complete
	
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
	Failure: exceptions.RuntimeError: Triggerable scheduler stopped before build was complete
	
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
	Failure: exceptions.RuntimeError: Triggerable scheduler stopped before build was complete
	
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
	Failure: exceptions.RuntimeError: Triggerable scheduler stopped before build was complete
	
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
	Failure: exceptions.RuntimeError: Triggerable scheduler stopped before build was complete
	
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
	Failure: exceptions.RuntimeError: Triggerable scheduler stopped before build was complete
	
2017-08-16 09:42:24-0700 [-] (ignored) while invoking Triggerable schedulers:
	Traceback (most recent call last):
[...]

Issue #4015, opened 2018-03-27
Old build master still exists


lancelodliu:
In the Builds-Build Masters Tab, the old master I used months ago still exists and the workers are connected to it.
![image](https://user-images.githubusercontent.com/19603852/37945205-a9ac7922-31b1-11e8-8b06-c9e06cceecb6.png)
![image](https://user-images.githubusercontent.com/19603852/37945254-de5a020c-31b1-11e8-9935-a0febda06c6e.png)

aj062:
I also notice this issue. I notice it when I have a dev instance and I rename the hostname (and buildbotURL and similar entries in master.cfg), and bring up new buildbot instance.

I think that's because the "masters" table in database has the entry for previous build master. 

@tardyp Can we safely delete that old master entry from "masters" table in database? or is there any other/better way to delete old masters?

tardyp:
The master table is here to be able to support multimaster. If you change hostname or change path of master, then, this will create another master.

Problem with erasing a master is that you will need to cascade delete all the data that is associated with it (or the db engine will complain about foreign key issue).
Notably, all builds record which master they built on.

This is something we could do in the Janitor step.
If after some times there is no build associated with a master anymore, its time to delete it from the db.

scottalthoff:
most of my workers connect to the new master, but two of them connect to both the old and the new one. Is there any way to stop this from happening?

GretaCB:
Also seeing workers connected to an old master that no longer exists. Is there a way to stop the worker from connecting to the non-existent master? Not sure if the potential Janitor step has been implemented or if there is a way to do this manually.

mlabbe:
I ran into this problem because I run buildbot master in a container, and the name of the master is coming from the hostname of the container. This container hostname is a guid that changes whenever I upgrade buildbot, but can happen for many more reasons.  In my case, I was linking to almost 100 non-existent masters. Clicking the link to each othem further resulted in a 404 error.

The result is that upgrading buildbot shows my workers connected to old masters, present when they are not, and showing multi-master.  

The workaround is specifying the hostname inside the container, purging my db, and then never changing it again.  The continued existence of this bug has subtle effects that  are probably impacting people's perceived reliability of buildbot, as this was a relatively tricky one to track down.

PR #6620, opened 2022-08-26
Call reactor.stop() after master finished stopping its services

mokibit:
This PR fixes the problem #3535.

"buildbot stop" command was unstable especially in cases when stopping
of builds takes longer or connection between master and worker is slow
or disappears. This happens due to reactor.stop() being called before
finishing of the services of master had a chance to complete.
Sometimes master process would be stopped successfuly, sometimes this
action would make master process unresponsive.

This change improves reliability of master process shutdown: now after
master gets SIGTERM, it first stops all its services and only then
reactor.stop() is called. In this way reactor.stop() can not interfere
with master shutdown.

### DIFF
--- a/master/buildbot/master.py
+++ b/master/buildbot/master.py
@@ -118,6 +118,8 @@ class BuildMaster(service.ReconfigurableServiceMixin, service.MasterService):
         # local cache for this master's object ID
         self._object_id = None
 
+        self._got_sigterm = False
+
         # Check environment is sensible
         check_functional_environment(self.config)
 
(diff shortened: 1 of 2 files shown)


p12tic (approved):

master/buildbot/master.py
```diff
@@ -325,6 +327,16 @@ def sigusr1(*args):
...
+            @defer.inlineCallbacks
+            def call_after_signal(sig_num, stack):
+                if not self._got_sigterm:
+                    self._got_sigterm = True
+                    yield self.disownServiceParent()
+                    self.reactor.stop()
+                else:
+                    log.msg('Ignoring SIGTERM, master is already shutting down.')
+
+            signal.signal(signal.SIGTERM, call_after_signal)
```
tardyp, line 339: shouldn't we be similar to the code in line 278?

I think the hasattr must be for windows.

p12tic: According to https://docs.python.org/3/library/signal.html#signal.SIGTERM `SIGTERM` is available on Windows unconditionally.

Though your comment got me wondering whether current SIGUSR1 handling is broken on multimaster setups because it is not implemented in terms of `stopService`. In particular, `self.data.updates.masterStopped` will not be called.

p12tic: The issue with masterStopped being not called in clean shutdown may be the cause of https://github.com/buildbot/buildbot/issues/4015.

PR #6644, opened 2022-09-23
Added TreqResponseWrapper

### LINKED ISSUE #6633: builtins.AttributeError: instead of intended log message
Hello,

Because I had set up an incorrect url, or an url that was not possible to access due to limited api token access I got this error message:
```
          File "/usr/local/lib/python3.10/site-packages/buildbot/changes/github.py", line 142, in _getPulls
            f"while loading {result.url}")
        builtins.AttributeError: '_Response' object has no attribute 'url'
```

However, looking at the code the error should probably be something more like:
```
"GitHubPullrequestPoller error {error_code} '{error message}' while loading {some_url}")
```

Is the attribute error the intended behaviour? If not then I guess I would do some kind of patch for it (i.e. checking for existing url attribute).

markus-sus:
This fixes issues with the url attribute missing in the response when treq is used.

* [ ] I have updated the appropriate documentation because
      I didn't think any documentation have to be updated, I use treq with GitHubPullrequestPoller

Fixes #6633

### DIFF
--- a/master/buildbot/test/integration/test_setup_entrypoints.py
+++ b/master/buildbot/test/integration/test_setup_entrypoints.py
@@ -116,6 +116,7 @@ class TestSetupPyEntryPoints(unittest.TestCase):
             'buildbot.util.git.GitStepMixin',
             'buildbot.util.giturlparse.GitUrl',
             'buildbot.util.httpclientservice.HTTPClientService',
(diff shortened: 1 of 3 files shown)


p12tic (approved):

markus-sus:
Holy crap. That was quick.

PR #6654, opened 2022-10-18
fix/noNegativeTime

### LINKED ISSUE #3457: elapsed time for current build is negative
Frequently, while loading webpages, buildbot shows negative time for time elapsed for current build/step. It gets corrected automatically after few seconds. However, if buildbot is not very fast, this is very noticeable. Seems like some script taking long time to load and displaying negative time meanwhile.

Few screenshots below.

screen shot 2017-07-20 at 6 47 57 pm

screen shot 2017-07-20 at 9 20 18 pm

bjorges:
If start time is before end time we set duration to 0 to avoid negative duration, closes #3457

### DIFF
--- a/master/buildbot/statistics/capture.py
+++ b/master/buildbot/statistics/capture.py
@@ -14,6 +14,7 @@
 # Copyright Buildbot Team Members
 
 import abc
+import datetime
 import re
 
 from twisted.internet import defer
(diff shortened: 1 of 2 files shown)


p12tic:
@bjorges  Thanks for the pull request. Does it really fix https://github.com/buildbot/buildbot/issues/3457? By cursory look it does not seem times displayed on the frontend are related to statistics service, but maybe I'm wrong.

This PR is good anyway, as negative build times shouldn't be captured in statistics module either.

PR #6683, opened 2022-11-19
Add notfy_events events

jethornton:
Remove misleading text
## Contributor Checklist:
Document Only Change


jethornton:
I don't understand the failure, some input is needed.

JT

p12tic (approved):

p12tic:
@jethornton Thanks for the PR. The test failure is unrelated, it crept into the testsuite by means that we haven't investigated yet.

PR #6728, opened 2022-12-04
Add release note for React-based www UI

p12tic:


tardyp:
\o/ congrats for the release. shall we deploy that on metabuildbot?

p12tic:
> shall we deploy that on metabuildbot?

Not yet, I would like to convert the rest of web UI plugins to React. 

Also, we should probably touch metabuildbot as little as possible, because Ansible update scripts are currently broken due to FreeBSD server wanting an upgrade. Once we move to Debian server and put metabuildbot to a container we can do all the pending upgrades.

PR #6740, opened 2022-12-05
Expand path before deleting secret

lms-jk:
## Description

File paths were expanded when the worker downloaded the secret, but not when it tried to delete them.

## Contributor Checklist:

I am quite new to Buildbot and still try to find my way around the code, so please feel free to correct me.

It seems to me, that the affected function `protocol_args_setup` is not covered by any unit test. I am willing to write some new unit tests, but I would need some guidance to do so.

No changes in the documentation should be necessary.

### DIFF
--- a/master/buildbot/test/integration/test_download_secret_to_worker.py
+++ b/master/buildbot/test/integration/test_download_secret_to_worker.py
@@ -1,0 +1,104 @@
+# This file is part of Buildbot.  Buildbot is free software: you can
+# redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, version 2.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51
+# Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Copyright Buildbot Team Members
+
+import os
+
+from parameterized import parameterized
+
+from twisted.internet import defer
+from twisted.trial.unittest import SkipTest
+
+from buildbot.config import BuilderConfig
+from buildbot.process.factory import BuildFactory
+from buildbot.schedulers.forcesched import ForceScheduler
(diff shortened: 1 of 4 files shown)


p12tic:
@lms-jk Thanks for the PR.

Some context on why the code is structured the way it is:

PB is a legacy protocol which is not well specified at a low level and depends on how the underlying Twisted library actually implements things. There's a new protocol based on MessagePack that allows to potentially have a worker that is not based on Python. The MessagePack-based protocol is much simpler than PB protocol because as many things as possible have been moved to master side. For example, master knows HOME of the worker and thus is able to expand the path on master side and send a simple and unambiguous instruction to the worker: "remove this absolute path" instead of "remove this path, but also try to interpret it when it's relative and also take into account HOME expansion". This is how the code in `protocol_args_setup()` got created - it does all the path manipulation that is specific to PB protocol and then transfers the data to code that is shared between MessagePack and PB.

I agree that it makes sense to expand HOME in "rmfile" command. We should do the same in MessagePack protocol: see `remoteStartCommand()` in protocols/msgpack.py.

As for the tests, I think unit tests for aren't very useful in this case because we'll need to mock `os.path.expanduser` so the test will be basically repeating code that's being tested. We would need an integration test. One option is to test `CompositeStepMixin.runRmFile()` (see e.g. `master/buildbot/test/integration/interop/test_compositestepmixin.py`). Another option would be to test `DownloadSecretsToWorker` and `RemoveWorkerFileSecret` pair as that's what we're really interested in fixing. The complicated part would be to figure out the location of where the test files would live. I think the test could get relative path from `__file__` to `$HOME` and then put files in some subdirectory. If source checkout if outside of `$HOME` or `$HOME` does not exist, we could raise `SkipTest`.

I think the ideal testing way is quite involved relative to the complexity of the fix so I would understand if you just don't want to do it. Please just update MessagePack code in that case and the PR will be accepted.

lms-jk:
Thanks for the thorough explanation and instructions, I appreciate it! I will try to design a test for the `DownloadSecretsToWorker` and `RemoveWorkerFileSecret` pair.

lms-jk:
I reverted the changes I made to the test-cases and rebased my patch on the current master. I am currently missing the time to properly understand the testing methodology, so  I would gladly take your offer to merge without test-cases.

lms-jk:
Commits rebased on current master. Is anything missing for this PR to be accepted?

p12tic:
@lms-jk Sorry for the delay, I've implemented the tests and will merge the PR soon.

lms-jk:
No worries. Thank you for the work you do on this awesome open source project!

PR #6753, opened 2022-12-07
www/react: Miscellaneous fixes

p12tic:


yan12125:
Two js_unit tests fail as `protocol` values in https://github.com/buildbot/buildbot/blob/master/www/react-base/src/data/WebSocketClient.test.ts are not changed from `https` to `https:`. Other failures on Windows workers seem flaky ones and unrelated to this PR.

Issue #5264, opened 2020-04-20
Error cloning with private SSH key


austenadler:
I am always getting the error:
```
Load key "/home/.../ssh-key": invalid format
```
When using keys generated on linux with ssh-keygen. I believe it is due to the newline being trimmed when the key is sent to the worker.

Steps taken:

I added:
```
import subprocess
print(subprocess.run(["cp", keyPath, "/tmp/copied"])
```
After this line: https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/git.py#L33. The copied file has no trailing newline. I tried a git clone with this keyfile and it failed. If I manually added a newline to `/tmp/copied`, clone worked.

I added a second trailing newline to the keyfile but it still copied with no newlines. Finally, I added a space to the end of the file and it copied with the space and the clone worked after restarting the master.

Possible fix:
Figure out why this key doesn't have a trailing newline.

Version:
```
Buildbot version: 2.7.0
Twisted version: 20.3.0
```

tardyp:
Hello,
The code for downloading the key is here:
https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/git.py#L264

There might be some kind of strip() hidden somewhere.

You could print(repr(self.sshPrivateKey[-3:])) to figure out where it is. actual print will be stored in the twisted.log file

Pierre

austenadler:
Okay I added it:
```python
print("XXXDEBUG: ", [ord(i) for i in private_key[-3:]])
# XXXDEBUG:  [45, 45, 45]
# The dash character from ---OPENSSL PRIVATE KEY---
```

So `p.render(self.sshPrivateKey)` is stripping the newline. I'm having trouble tracking all the renderers in secret. I can look a little but I do not know where the keyfile is converted into a string.

austenadler:
I see. I had configured `strip=True` in my build master. I did not realize this would trim newlines from private keys. This is not an issue

tardyp:
Hey,

For the record, where exactly did you have the strip=True?

austenadler:
Same as in #4681, I put it in SecretInAFile:

```python
c['secretsProviders'] = [
secrets.SecretInAFile(dirname="/var/lib/buildbot/secrets", strip=True),
]
```

Apr 21, 2020 2:26:11 AM Pierre Tardy <notifications@github.com>:

razrdog0:
[...]

PR #6770, opened 2023-01-09
Ssh key fix

razrdog0:
## Contributor Checklist:

I'm unsure if there is a documentation change required.


razrdog0:
This pull request should fix the issue from https://github.com/buildbot/buildbot/issues/5264

p12tic (approved):
Thanks a lot! This looks great.

p12tic:
I've squashed commits into one as commits fixing an issue in a preceding commit are not an useful thing to keep in history.

razrdog0:
Can you provide some more details on the error you're getting? perhaps in a fresh issue. I'd be happy to take a look and copy the fix over for known hosts later this week if that is the issue.

PR #6775, opened 2023-01-19
docs/www: clarify the SSL support

### LINKED ISSUE #6774: Is HTTPS on the web interface still unsupported?
Hi,

the [buildbot documentation](https://docs.buildbot.net/latest/manual/configuration/www.html#web-server) states that HTTPS on the web interface is unsupported, but twisted offers this functionality via [serverFromString](https://docs.twistedmatrix.com/en/stable/api/twisted.internet.endpoints.html#serverFromString). It has been working flawlessly for a few months on my installation by simply replacing the tcp port with this string:
```
'ssl:8010:privateKey=ssl/key.pem:certKey=ssl/cert.pem:extraCertChain=ssl/chain.pem'
```
Is the use of HTTPS discouraged on the twisted web server or should the documentation be updated?

lms-jk:
## Contributor Checklist:

* [] I have updated the unit tests

Is a unit test necessary?

### DIFF
--- a/master/docs/manual/configuration/www.rst
+++ b/master/docs/manual/configuration/www.rst
@@ -12,7 +12,7 @@ This server is configured with the ``www`` configuration key, which specifies a 
 
 ``port``
     The TCP port on which to serve requests.
-    It might be an integer or any string accepted by `serverFromString <https://twistedmatrix.com/documents/current/api/twisted.internet.endpoints.html#serverFromString>`_ (ex: "tcp:8010:interface=127.0.0.1" to listen on another interface).
+    It might be an integer or any string accepted by `serverFromString <https://docs.twistedmatrix.com/en/stable/api/twisted.internet.endpoints.html#serverFromString>`_ (ex: "tcp:8010:interface=127.0.0.1" to listen on another interface).
     Note that SSL is not supported.
     To host Buildbot with SSL, use an HTTP proxy such as lighttpd, nginx, or Apache.
     If this is ``None`` (the default), then the master will not implement a web server.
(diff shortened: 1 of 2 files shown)


lms-jk:
This PR fixes #6774.

pmisik:
You're right, you can run Buildbot on SSL with the current Twisted. 
I'd say it's good for those who are starting to use Buildbot on a single instance.
Buildbot doesn't expose configuration of any SSL related hardening options.

For corporate instances Buildbot running behind a robust http server (nginx, apache, ...) has some advantages that users can take advantage of (e.g.: well-known hardening/security options, IT department can independently take care of the http server (e.g.: X.509 certificates live time management, vulnerability assessment, ...).

Running Buildbot behind a reverse proxy server provides even more added value (load balancing, SSO, ...).

I would suggest keep and rephrase the last sentence you removed to
`It is recommended to host Buildbot with SSL, using an HTTP proxy server such as lighttpd, nginx or Apache.`

tardyp:
I agree with @pmisik 200%. I don't see a good reason to let twisted do the SSL termination.

The best practice for securing buildbot is to put behind a reverse proxy. It is super easy and shall be the 101 for a good devops engineer.

It was written that it is not supported, not because it cannot be made functional, but because it is not a good idea.

lms-jk:
Thanks for clarifying! I adapted @pmisik suggestion a bit. Let me know what you think.

tardyp (approved):

p12tic (approved):

p12tic:
Thanks!

PR #6779, opened 2023-01-26
libvirt: specify backing_fmt for qemu-img

ematsumiya:
When cheap_copy is True (default) in LibVirtWorker, clone_cmd doesn't specify the backing file format (backing_fmt).  qemu >= 6.1 requires that it's explicitly specified for several commands, including qemu-img.

Even though on more recent versions it allows the shorthand '-F' to be used, use the generic '-o' option to set it for backwards compatibility (backing_fmt option existed prior to v6.1).

Adapt unit test for this change.

I couldn't test how much backwards '-o backing_fmt' is supported, but on the oldest VM I have (SLES12-SP5) it has qemu 3.1.1 and that option is there.

References:
https://www.qemu.org/docs/master/about/removed-features.html#qemu-img-backing-file-without-format-removed-in-6-1

https://github.com/qemu/qemu/commit/1899bf47375ad40555dcdff12ba49b4b8b82df38


p12tic:
Thanks!

Issue #4407, opened 2018-11-06
buildbot.test.integration.interop.test_transfer.TransferStepsMasterPb.test_transfer is flaky on Windows


p12tic:
https://ci.appveyor.com/project/djmitche/buildbot/builds/20090290/job/8bnsx9sh621jpnwd
https://ci.appveyor.com/project/djmitche/buildbot/builds/20044022/job/4i1b47hop0ogisy2

p12tic:
Will disable in https://github.com/buildbot/buildbot/pull/4405.

yan12125:
This test is flaky on Linux, too. A test failure from Arch Linux: https://tests.reproducible-builds.org/archlinux/community/buildbot/build2.log

This log file will be overwritten during the next Arch Linux package test, so I paste relevant lines here:
```
[ERROR]
Traceback (most recent call last):
Failure: twisted.internet.defer.TimeoutError: <buildbot.test.integration.interop.test_transfer.TransferStepsMasterPb testMethod=test_transfer> (test_transfer) still running at 5.0 secs

buildbot.test.integration.interop.test_transfer.TransferStepsMasterPb.test_transfer
===============================================================================
[ERROR]
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/twisted/trial/util.py", line 276, in _runSequentially
    thing = yield d
  File "/usr/lib/python3.7/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/tmp/buildbot-archlinuxrb-build-EeIrmqxN/buildbot/trunk/src/tmp_install/usr/lib/python3.7/site-packages/buildbot/test/util/integration.py", line 201, in dump
    raise self.failureException(dump.getvalue())
twisted.trial.unittest.FailTest: FAILED! dumping build db for debug
*** BUILD 1 *** ==> building (not finished)
    *** STEP worker_preparation *** ==> worker ready (success)
    *** STEP string_download *** ==> downloading to file1.txt (success)
    *** STEP string_download_1 *** ==> downloading to file2.txt (success)
    *** STEP download *** ==> running (not finished)


buildbot.test.integration.interop.test_transfer.TransferStepsMasterPb.test_transfer
-------------------------------------------------------------------------------
```

PR #6780, opened 2023-01-26
Use unittest.mock

opoplawski:
## Contributor Checklist:

* [NA] I have updated the unit tests 
* [*] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* [*] I have updated the appropriate documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ flake8:
 
 frontend_deps: $(VENV_NAME)
 	$(PIP) install -e pkg
-	$(PIP) install mock wheel buildbot
+	$(PIP) install wheel buildbot
 	cd www/build_common; $(YARN) install --pure-lockfile
 	for i in $(WWW_DEP_PKGS); \
 		do (cd $$i; $(YARN) install --pure-lockfile; $(YARN) run build); done
--- a/common/smokedist.sh
+++ b/common/smokedist.sh
@@ -17,7 +17,7 @@ do
     fi
     . $VE/bin/activate
     pip install -U pip
-    pip install  mock requests flask
+    pip install requests flask
     pip install dist/buildbot-[0-9]*.$suffix
     pip install dist/buildbot?pkg*.$suffix
     pip install dist/*.$suffix
--- a/common/validate.sh
+++ b/common/validate.sh
@@ -34,7 +34,7 @@ if $help; then
     echo "USAGE: common/validate.sh [oldrev] [--quick] [--no-js] [--help]"
     echo "  This script will test a set of patches (oldrev..HEAD) for basic acceptability as a patch"
     echo "  Run it in an activated virtualenv with the current Buildbot installed, as well as"
-    echo "      sphinx, flake8, mock, and so on"
+    echo "      sphinx, flake8, and so on"
     echo "To use a different directory for tests, pass TRIALTMP=/path as an env variable"
     echo "if --quick is passed validate will skip unit tests and concentrate on coding style"
     echo "if --no-js is passed validate will skip tests that require Node and NPM"
--- a/master/buildbot/test/__init__.py
+++ b/master/buildbot/test/__init__.py
@@ -16,7 +16,7 @@
 import os
 import sys
 import warnings
-from pkg_resources import parse_version
+from unittest import mock
 
 import setuptools  # force import setuptools before any other distutils imports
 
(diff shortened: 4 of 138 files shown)


yan12125:
Thank you very much for these changes! I'm also looking for getting rid of the external mock library from buildbot.

Some tests fail with:
```
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.9/site-packages/twisted/trial/runner.py", line 691, in loadByName
    return self.suiteFactory([self.findByName(name, recurse=recurse)])
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.9/site-packages/twisted/trial/runner.py", line 451, in findByName
    obj = reflect.namedModule(searchName)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.9/site-packages/twisted/python/reflect.py", line 156, in namedModule
    topLevel = __import__(name)
  File "/buildbot/buildbot-job/build/master/buildbot/test/__init__.py", line 42, in <module>
    if parse_version(mock.__version__) < parse_version("0.8"):
builtins.AttributeError: module 'unittest.mock' has no attribute '__version__'
```
Checking if mock is newer than 0.8 fails as `unittest.mock` does not have `__version__`. I assume that check can be removed as [mock 0.8](https://pypi.org/project/mock/0.8.0/) is from 2012 and unlikely installed on current systems.

Some other tests fail as the external `mock` is still imported. For example:
```
[ERROR]
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.7/site-packages/twisted/trial/runner.py", line 596, in loadPackage
    module = modinfo.load()
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.7/site-packages/twisted/python/modules.py", line 389, in load
    return self.pathEntry.pythonPath.moduleLoader(self.name)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.7/site-packages/twisted/python/reflect.py", line 297, in namedAny
    topLevelPackage = _importAndCheckStack(trialname)
  File "/buildbot/buildbot-job/build/sandbox/lib/python3.7/site-packages/twisted/python/reflect.py", line 244, in _importAndCheckStack
    raise excValue.with_traceback(excTraceback)
  File "/buildbot/buildbot-job/build/master/buildbot/test/unit/changes/test_gitpoller.py", line 19, in <module>
    import mock
builtins.ModuleNotFoundError: No module named 'mock'
```
Most likely those files should be updated as well.

On the other hand, files under `worker/` probably still needs the external `mock`, as `unittest.mock` exists [since Python >= 3.3](https://docs.python.org/3/library/unittest.mock.html), while worker still [supports Python 2.7](https://github.com/buildbot/buildbot/blob/v3.7.0/worker/setup.py).

opoplawski:
Hopefully this is better.  Still need to conditionalize the mock dep in worker/setup.py for Python < 3.3.

opoplawski:
This isort stuff is driving me batty, I don't think I can deal with it.

yan12125:
I guess the isort test will pass if you apply diffs shown in logs? For example, https://buildbot.buildbot.net/api/v2/logs/1163246/raw for the latest test.

---

Failures in `bb/trial_worker/2.7/tw:20.3.0/sqla:latest/db:sqlite:///wp:` indicates missing `mock`. As that test uses `requirements-ciworker.txt` instead of `setup.py`, `mock` may be still necessary in `requirements-ciworker.txt`.

---

msgpack-related failures are tricky. The cause is that `unittest.mock` does not support `call_args.args` until Python 3.8 [1]. A possible fix is:
```Diff
diff --git a/master/buildbot/test/unit/worker/test_protocols_manager_msgmanager.py b/master/buildbot/test/unit/worker/test_protocols_manager_msgmanager.py
index f729f8475..b002e7ae3 100644
--- a/master/buildbot/test/unit/worker/test_protocols_manager_msgmanager.py
+++ b/master/buildbot/test/unit/worker/test_protocols_manager_msgmanager.py
@@ -90,7 +90,7 @@ class TestBuildbotWebSocketServerProtocol(unittest.TestCase):
 
         protocol.onMessage(msgpack.packb(msg), True)
         yield protocol._deferwaiter.wait()
-        args_tuple = protocol.sendMessage.call_args.args
+        args_tuple, _ = protocol.sendMessage.call_args
         result = msgpack.unpackb(args_tuple[0], raw=False)
         self.assertEqual(result, expected)
 
@@ -101,7 +101,7 @@ class TestBuildbotWebSocketServerProtocol(unittest.TestCase):
 
         self.protocol.onMessage(msgpack.packb(msg), True)
 
-        args_tuple = self.protocol.sendMessage.call_args.args
+        args_tuple, _ = self.protocol.sendMessage.call_args
         return msgpack.unpackb(args_tuple[0], raw=False)['result']
 
     @defer.inlineCallbacks
```

[1] https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args

opoplawski:
It seems like the checks are okay now and I've rebased down to a single commit.  What else is needed?

yan12125:
Thanks! There are failures in `bb/trial/3.9/tw:18.7.0/sqla:latest/db:sqlite:///wp:` check. More investigation are needed.

yan12125:
I cannot reproduce errors in `bb/trial/3.9/tw:18.7.0/sqla:latest/db:sqlite:///wp:` check. Error logs are similar to those mentioned in https://github.com/buildbot/buildbot/issues/4407, so errors may not be caused by this pull request.

yan12125:
63d87485fc2f39151d16c4f0a682351e64038965 causes an conflict with `master/buildbot/test/unit/test_util.py`.

@p12tic Mind to have a look if conflicts are resolved?

opoplawski:
I've rebased and resolved the conflict.

yan12125:
Thanks. I cherry-picked this PR to Arch Linux for now.

Apparently test failures other than isort are related to https://github.com/pypa/setuptools/pull/3843.

p12tic (approved):

p12tic:
Rebased on top of latest master. 

Sorry for the long delay. After looking into the PR at first I came into the conclusion that `unittest.mock` is not a full replacement for `mock` and didn't notice that incompatibilities were addressed in code later.

p12tic:
Thanks a lot for the PR.

PR #6807, opened 2023-03-03
Github hook: Include "previous filename" key in list of files

DavidSpickett:
Over in LLVM we have a builder that only looks at the llvm/ subfolder of the repo. The change https://github.com/llvm/llvm-project/commit/efae3174f09560353fb0f3d528bcbffe060d5438 moved a file out of that folder, which broke the build.

However because buildbot only collects the "filename" key, it did not think this was a commit the builder cared about so in the UI the commit was filtered out.

Github provides a "previous_filename" key so you can detect moved files, for the change above you would get:
```
{
  "filename": "cmake/Modules/LLVMCheckLinkerFlag.cmake",
   <...>
  "previous_filename": "llvm/cmake/modules/LLVMCheckLinkerFlag.cmake"
},
```

We should include the previous name also, since moving a file can break a build just as well as deleting or changing one.

You might wonder if including moved file names is a good idea since they don't exist anymore. We do the same thing with deleted files already. Since those only have a "filename" key:
```
    {
      "sha": "f61ec0585f9a45a1ee8a0e850a2388d3c67bd856",
      "filename": "cmake/Modules/LLVMCheckCompilerLinkerFlag.cmake",
      "status": "removed",
```

And as far as I can tell this list is for file changes of all kinds a builder might care about. Not "here are the files that now exist".
## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/www/test_hooks_github.py
+++ b/master/buildbot/test/unit/www/test_hooks_github.py
@@ -451,7 +451,8 @@ gitJsonPayloadCommit = {
 
 gitJsonPayloadFiles = [
     {
-        "filename": "README.md"
+        "filename": "README.md",
+        "previous_filename": "old_README.md"
     }
 ]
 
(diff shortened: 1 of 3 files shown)


p12tic:
Thanks!

p12tic (approved):

PR #6820, opened 2023-03-15
Fix env override when calling makeRemoteShellCommand

jsonar-cpapke:
Before this change, setting env when directly calling `makeRemoteShellCommand` was effectively ignored. This change updates that to use the override if it exists.

## Contributor Checklist:


p12tic (approved):

p12tic:
Looks great, thank you!

PR #6863, opened 2023-05-04
Added cron features to nightly scheduler

vx-i:
I would like to add some cron features to nightly scheduler, like last day of month or 3rd monday of month. Maybe I need help with unit tests.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/util/croniter.py
+++ b/master/buildbot/util/croniter.py
@@ -6,6 +6,8 @@
 
 # -*- coding: utf-8 -*-
 
+import calendar
+import copy
 import re
 from datetime import datetime
 from time import mktime
@@ -16,12 +18,32 @@ from dateutil.relativedelta import relativedelta
 search_re = re.compile(r'^([^-]+)-([^-/]+)(/(.*))?$')
 only_int_re = re.compile(r'^\d+$')
 any_int_re = re.compile(r'^\d+')
+
+MONTH_ALPHAS = {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5, 'jun': 6,
+            'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10, 'nov': 11, 'dec': 12}
+DAY_OF_WEEK_ALPHAS = {'sun': 0, 'mon': 1, 'tue': 2, 'wed': 3, 'thu': 4, 'fri': 5, 'sat': 6}
+ALPHAS = {}
+for i in MONTH_ALPHAS, DAY_OF_WEEK_ALPHAS:
+    ALPHAS.update(i)
+del i
+step_search_re = re.compile(r'^([^-]+)-([^-/]+)(/(\d+))?$')
+
+WEEKDAYS_RE_PART = '|'.join(DAY_OF_WEEK_ALPHAS.keys())
+MONTHS_RE_PART = '|'.join(MONTH_ALPHAS.keys())
 star_or_int_re = re.compile(r'^(\d+|\*)$')
+# This regular expression will search for special cases in the day of week
+# and months. Expression like 2#3, l3, jan-feb or mon-tue.
+# There are 3 groups: pre, he and last.
+special_day_of_week_or_month_re = re.compile(
+    (r'^(?P<pre>((?P<he>(({WEEKDAYS})(-({WEEKDAYS}))?)').format(WEEKDAYS=WEEKDAYS_RE_PART) +
+    (r'|(({MONTHS})(-({MONTHS}))?)|\w+)#)|l)(?P<last>\d+)$').format(MONTHS=MONTHS_RE_PART)
+)
 
 __all__ = ('croniter',)
 
 
 class croniter:
+    MONTHS_IN_YEAR = 12
     RANGES = (
         (0, 59),
         (0, 23),
@@ -35,14 +57,19 @@ class croniter:
     )
 
     ALPHACONV = (
-        {},
-        {},
-        {},
-        {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5, 'jun': 6,
-         'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10, 'nov': 11, 'dec': 12},
-        {'sun': 0, 'mon': 1, 'tue': 2, 'wed': 3, 'thu': 4, 'fri': 5, 'sat': 0},
+        {},  # 0: min
+        {},  # 1: hour
+        {'l': 'l'},  # 2: day of month
+        # 3: month
+        copy.deepcopy(MONTH_ALPHAS),
+        # 4: day of week
+        copy.deepcopy(DAY_OF_WEEK_ALPHAS),
+        # command/user
         {}
     )
+
+    DAY_OF_MONTH_FIELD_INDEX = 2
+    DAY_OF_WEEK_FIELD_INDEX = 4
 
     LOWMAP = (
         {},
@@ -51,6 +78,15 @@ class croniter:
         {0: 1},
         {7: 0},
         {},
+    )
+
+    MAX_VALUE_PER_INDEX = (
+        60,
+        24,
+        31,
+        12,
+        7,
+        60
     )
 
     bad_length = 'Exactly 5 or 6 columns has to be specified for iterator' \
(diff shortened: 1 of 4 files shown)


master/buildbot/util/croniter.py
```diff
@@ -67,18 +97,58 @@ def __init__(self, expr_format, start_time=time()):
...
 
         for i, expr in enumerate(self.exprs):
+            expr = expr.lower()
             e_list = expr.split(',')
             res = []
 
             while e_list:
                 e = e_list.pop()
+                nth = None
+                special_dow_rem = special_dow_re.match(str(e))
```
p12tic, line 110: Won't `e` be already a string?

master/buildbot/util/croniter.py
```diff
@@ -16,12 +18,29 @@
 search_re = re.compile(r'^([^-]+)-([^-/]+)(/(.*))?$')
 only_int_re = re.compile(r'^\d+$')
 any_int_re = re.compile(r'^\d+')
+
+M_ALPHAS = {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5, 'jun': 6,
```
p12tic, line 22: MONTH_ALPHAS

master/buildbot/util/croniter.py
```diff
@@ -16,12 +18,29 @@
...
+            'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10, 'nov': 11, 'dec': 12}
+DOW_ALPHAS = {'sun': 0, 'mon': 1, 'tue': 2, 'wed': 3, 'thu': 4, 'fri': 5, 'sat': 6}
```
p12tic, line 24: DAY_OF_WEEK_ALPHAS

master/buildbot/util/croniter.py
```diff
@@ -16,12 +18,29 @@
...
+ALPHAS = {}
+for i in M_ALPHAS, DOW_ALPHAS:
+    ALPHAS.update(i)
+del i
+step_search_re = re.compile(r'^([^-]+)-([^-/]+)(/(\d+))?$')
+
+WEEKDAYS = '|'.join(DOW_ALPHAS.keys())
```
p12tic, line 31: WEEKDAYS_RE_PART

master/buildbot/util/croniter.py
```diff
@@ -16,12 +18,29 @@
...
+MONTHS = '|'.join(M_ALPHAS.keys())
```
p12tic, line 32: MONTHS_RE_PART

master/buildbot/util/croniter.py
```diff
@@ -67,18 +97,58 @@ def __init__(self, expr_format, start_time=time()):
...
             e_list = expr.split(',')
             res = []
 
             while e_list:
                 e = e_list.pop()
+                nth = None
+                special_dow_rem = special_dow_re.match(str(e))
+
+                # Handle special case in the day of week expression: 2#3, l3
+                if i == 4 and special_dow_rem:
```
p12tic, line 113: Let's not use `4` as a special constant, but introduce names for them, e.g. `cron_min_index`, `cron_day_of_month_index`, etc.

master/buildbot/util/croniter.py
```diff
@@ -67,18 +97,58 @@ def __init__(self, expr_format, start_time=time()):
...
+                    g = special_dow_rem.groupdict()
+                    he, last = g.get('he', ''), g.get('last', '')
```
p12tic, line 115: Splitting this into two lines would be easier to understand.

master/buildbot/util/croniter.py
```diff
@@ -35,12 +54,14 @@ class croniter:
...
     ALPHACONV = (
-        {},
-        {},
-        {},
-        {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5, 'jun': 6,
-         'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10, 'nov': 11, 'dec': 12},
-        {'sun': 0, 'mon': 1, 'tue': 2, 'wed': 3, 'thu': 4, 'fri': 5, 'sat': 0},
+        {},  # 0: min
+        {},  # 1: hour
+        {"l": "l"},  # 2: day of month
```
p12tic, line 59: Day of week?

master/buildbot/util/croniter.py
```diff
@@ -16,12 +18,29 @@
...
+    ALPHAS.update(i)
+del i
+step_search_re = re.compile(r'^([^-]+)-([^-/]+)(/(\d+))?$')
+
+WEEKDAYS = '|'.join(DOW_ALPHAS.keys())
+MONTHS = '|'.join(M_ALPHAS.keys())
 star_or_int_re = re.compile(r'^(\d+|\*)$')
+special_dow_re = re.compile(
+    (r'^(?P<pre>((?P<he>(({WEEKDAYS})(-({WEEKDAYS}))?)').format(WEEKDAYS=WEEKDAYS) +
+    (r'|(({MONTHS})(-({MONTHS}))?)|\w+)#)|l)(?P<last>\d+)$').format(MONTHS=MONTHS)
```
p12tic, line 36: We need a comment of what this regex is trying to accomplish (e.g. what would be the accepted inputs, etc.). Otherwise even if I approve and merge the PR, down the line someone else or even me myself will think that the regex is unnecessarily too complex and simplify it breaking some use cases.

master/buildbot/util/croniter.py
p12tic: Also, `he` and `last` need a more complete names to describe what they mean.

master/docs/manual/configuration/schedulers.rst
```diff
@@ -641,7 +641,8 @@ The full list of parameters is:
 ``dayOfMonth`` (optional)
 
     The day of the month to start a build.
-    This defaults to ``*``, meaning every day.
+    This defaults to ``*``, meaning every day or l for last day.
```
p12tic, line 644: Let's use `L` because the matcher is not case sensitive anyway and `L` is more different from `1`, thus easier to read.

p12tic: Also needs quoting.

master/buildbot/util/croniter.py
```diff
@@ -67,18 +97,58 @@ def __init__(self, expr_format, start_time=time()):
...
+                    g = special_dow_rem.groupdict()
+                    he, last = g.get('he', ''), g.get('last', '')
+                    if he:
+                        e = he
+                        try:
+                            nth = int(last)
+                            assert 1 <= nth <= 5
+                        except (KeyError, ValueError, AssertionError) as e:
+                            raise ValueError(
+                                "[{expr_format}] is not acceptable.  Invalid day_of_week "
```
p12tic, line 123: I think we miss `f` here to make it `f-string`.

master/buildbot/util/croniter.py
```diff
@@ -35,12 +54,14 @@ class croniter:
...
-        {},
-        {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5, 'jun': 6,
-         'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10, 'nov': 11, 'dec': 12},
-        {'sun': 0, 'mon': 1, 'tue': 2, 'wed': 3, 'thu': 4, 'fri': 5, 'sat': 0},
+        {},  # 0: min
+        {},  # 1: hour
+        {"l": "l"},  # 2: day of month
+        # 3: month
+        copy.deepcopy(M_ALPHAS),
+        # 4: day of month
```
p12tic, line 62: Day of week?

p12tic:
Thanks for the contribution. I'm now wondering - maybe just replacing the current code with a dependency on the `croniter` python package would get all the features that we want without maintaining the same code base twice?

vx-i:
> Thanks for the contribution. I'm now wondering - maybe just replacing the current code with a dependency on the `croniter` python package would get all the features that we want without maintaining the same code base twice?

I don't know why the decision was made at the time, but I'm sure it was for a reason. Fewer dependencies means more stable code.

p12tic:
> I don't know why the decision was made at the time, but I'm sure it was for a reason. Fewer dependencies means more stable code.

Yes, but in this case we're duplicating the effort by copy-pasting the code. I think a better investment would be to use croniter as a dependency.

vx-i:
> Yes, but in this case we're duplicating the effort by copy-pasting the code. I think a better investment would be to use croniter as a dependency.

I would like to disagree on this topic. I just used the croniter code as a starting point. Extracted only relevant parts, refactored code and made it more readable, tested inside Buildbot, added documentation. If you decide to use croniter as an external dependency you skip all my work, delete old internal croniter code, etc., etc. Additionally you add the risk to break things as external API's may change anytime.

vx-i:
@p12tic What should we do? Any plans on this topic? To accept PR in the short term and switch to croniter in the medium term?

p12tic:
@vx-i Thanks for persisting. Your proposal makes sense as a compromise.

PR #6979, opened 2023-07-21
Reporters: Serialize only related events

vibbo:
This one is for discussion first IMO.

**Problem**:
Serialization of events processing introduced in https://github.com/buildbot/buildbot/pull/6603 causes **significant** performance reduction.
In our case delays can reach hours as even the smallest delays accumulate (we have large Buildbot instance with single `BitbucketServerCoreAPIStatusPush` reporter) .

**Proposed solution**
I would like to introduce these changes as solution for this problem without compromising the original intention of https://github.com/buildbot/buildbot/pull/6603.

* The idea is to serialize only related events. Keys for grouping such related events are generated by  `_get_chain_key_for_event`. IMO only events related to `buildrequests` (buildrequests, builds) need such serialization and other events can be processes asynchronously.

* I also removed the wait for pending events in `reconfigService` as I find is useless. As far as I can tell it only creates potential for missing (not consuming) events while the reconfig is ongoing. And the pending events will be processed anyway.

---

### DIFF
--- a/master/buildbot/reporters/base.py
+++ b/master/buildbot/reporters/base.py
@@ -36,7 +36,7 @@ class ReporterBase(service.BuildbotService):
         super().__init__(*args, **kwargs)
         self.generators = None
         self._event_consumers = []
-        self._pending_got_event_call = None
+        self._pending_got_event_calls = {}
 
     def checkConfig(self, generators):
         if not isinstance(generators, list):
@@ -57,10 +57,6 @@ class ReporterBase(service.BuildbotService):
             yield consumer.stopConsuming()
         self._event_consumers = []
 
-        if self._pending_got_event_call is not None:
-            yield self._pending_got_event_call
-        self._pending_got_event_call = None
-
         self.generators = generators
 
         wanted_event_keys = set()
@@ -76,9 +72,11 @@ class ReporterBase(service.BuildbotService):
         for consumer in self._event_consumers:
             yield consumer.stopConsuming()
         self._event_consumers = []
-        if self._pending_got_event_call is not None:
-            yield self._pending_got_event_call
-        self._pending_got_event_call = None
+
+        for pending_call in list(self._pending_got_event_calls.values()):
+            yield pending_call
+        self._pending_got_event_calls = {}
+
         yield super().stopService()
 
     def _does_generator_want_key(self, generator, key):
@@ -87,17 +85,22 @@ class ReporterBase(service.BuildbotService):
                 return True
         return False
 
+    def _get_chain_key_for_event(self, key, msg):
+        if key[0] in ["builds", "buildrequests"]:
+            return ("buildrequestid", msg["buildrequestid"])
+        return None
+
     @defer.inlineCallbacks
     def _got_event(self, key, msg):
-        pending_got_event_call = self._pending_got_event_call
-
-        # Mark this call as pending.
-        self._pending_got_event_call = d = defer.Deferred()
-
-        # Wait for previously pending call, if any, to ensure
-        # reports are sent out in the order events were queued.
-        if pending_got_event_call is not None:
-            yield pending_got_event_call
+        chain_key = self._get_chain_key_for_event(key, msg)
+        if chain_key is not None:
+            d = defer.Deferred()
+            pending_call = self._pending_got_event_calls.get(chain_key)
+            self._pending_got_event_calls[chain_key] = d
+            # Wait for previously pending call, if any, to ensure
+            # reports are sent out in the order events were queued.
+            if pending_call is not None:
+                yield pending_call
 
         try:
             reports = []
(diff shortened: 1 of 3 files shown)


master/buildbot/reporters/base.py
```diff
@@ -87,17 +86,24 @@ def _does_generator_want_key(self, generator, key):
...
-
-        # Wait for previously pending call, if any, to ensure
-        # reports are sent out in the order events were queued.
-        if pending_got_event_call is not None:
-            yield pending_got_event_call
+        chain_key = self._get_chain_key_for_event(key, msg)
+        if chain_key is not None:
+            d = defer.Deferred()
+            pending_got_event_call = self._pending_got_event_calls.get(
+                chain_key
```
p12tic, line 100: I think this could fit in a single line. `pending_got_event_call` could be named `pending_call` to shorten the line as well.

master/buildbot/reporters/base.py
```diff
@@ -76,9 +72,12 @@ def stopService(self):
         for consumer in self._event_consumers:
             yield consumer.stopConsuming()
         self._event_consumers = []
-        if self._pending_got_event_call is not None:
-            yield self._pending_got_event_call
-        self._pending_got_event_call = None
+
+        for pending_got_event_call in list(
+            self._pending_got_event_calls.values()
```
p12tic, line 77: I think this could fit on a single line and `pending_got_event_call` could be named pending_call to shorten the line as well.

master/buildbot/reporters/base.py
```diff
@@ -57,10 +57,6 @@ def reconfigService(self, generators):
             yield consumer.stopConsuming()
         self._event_consumers = []
 
-        if self._pending_got_event_call is not None:
```
p12tic, line 60: The reason for this wait is that the callback code may depend on the exact configuration, so we should wait until it completes before changing the configuration under the callback's feet.

Unfortunately even this code did not prevent race conditions because the deriving class will run its `reconfigService` only after `startConsuming()`, so new `_got_event` calls could happen before reconfiguration is finished. Thus I think it's OK to remove it and later we will need a proper solution.

p12tic:
@vibbo Looks great. I only had a couple of stylistic comments.

master/buildbot/reporters/base.py
p12tic: I filed #6981 and #6982 so that these problems are not forgotten.

master/buildbot/test/unit/reporters/test_base.py
```diff
@@ -189,19 +189,19 @@ def generate_throw(*args, **kwargs):
 
     @defer.inlineCallbacks
     def test_reports_sent_in_order_despite_slow_generator(self):
-        gen = self.setup_mock_generator([('fake1', None, None)])
+        gen = self.setup_mock_generator([('builds', None, None)])
```
p12tic, line 192: I think we don't need to change this test to not create an illusion that the test depends on real event keys. Was the test broken for some reason?

p12tic: Nevermind, I was typing before thinking.

master/buildbot/test/unit/reporters/test_base.py
```diff
@@ -189,19 +189,19 @@ def generate_throw(*args, **kwargs):
 
     @defer.inlineCallbacks
     def test_reports_sent_in_order_despite_slow_generator(self):
```
p12tic, line 191: I think we also need a similar test where there's a call to `_got_event(..., {'buildrequestid': 2})` in between the calls to `_got_event(..., {'buildrequestid': 1})`. Then give result to the deferred for  `_got_event(..., {'buildrequestid': 2})` before giving result to the other deferreds and verify that processing for buildrequestid==2 was not blocked on processing for buildrequestid==1 (just the order of items in call_args_list is enough)

master/buildbot/test/unit/reporters/test_base.py
```diff
@@ -213,17 +213,102 @@ def test_reports_sent_in_order_despite_slow_generator(self):
...
+        mock_generate_calls['builds'][2]['finished'].callback(22)
+        notifier._got_event(('builds', None, 'finished'), {'buildrequestid': 2})
+
+        # Finish generate call for first event
+        mock_generate_calls['builds'][1]['new'].callback(11)
 
-        gen.generate.return_value.callback(1)
-        self.assertTrue(d.called)
-        self.assertIsNone(notifier._pending_got_event_call)
+        # Now sendMessage should have been called six times in given order
```
p12tic, line 310: six -> four

master/buildbot/reporters/base.py
```diff
@@ -76,9 +72,10 @@ def stopService(self):
         for consumer in self._event_consumers:
             yield consumer.stopConsuming()
         self._event_consumers = []
-        if self._pending_got_event_call is not None:
-            yield self._pending_got_event_call
-        self._pending_got_event_call = None
+
+        for pending_call in list(self._pending_got_event_calls.values()):
+            yield pending_call
+
```
p12tic, line 78: `self._pending_got_event_calls = {}` is missing here

p12tic:
Added two tiny comments

p12tic (approved):

PR #6995, opened 2023-07-28
Support setting the project in GitHubPullrequestPoller

lsgunth:
## Contributor Checklist:

## Summary

Presently the GitHubPullrequestPoller forces the project of the change to be set to the full name of
the GitHub repository as returned by the github API. This behaviour is not always useful if the pull request
puller needs to match the project of existing change sources.

Add the ability to specify the project explicitly like most other change sources. If the project is not specified
it will fall back on the previous behaviour.


GitHubPullrequestPoller-project.feature
```diff
@@ -0,0 +1,2 @@
+Support specifying the project in GitHubPullrequestPoller. Previously
+it was forced to be equal to GitHub's repoistory full name.
```
p12tic, line 2: Typo. repoistory => repository. Tests will fail due to this.

p12tic: Hm, tests work. Anyway, I will fix the typo at release time.

lsgunth:
That was quick, thanks!

PR #6996, opened 2023-07-28
reporter: Prevent possible loss of events during reconfig

### LINKED ISSUE #6982: There's a small window where events will be missed by reporters during reconfig
The code in `ReporterBase.reconfigService()` stops consuming events for all event keys and then starts consuming events for all event keys again. This is not necessary for any wanted event keys which are wanted in both old and new configuration. The code should stop consuming events for no longer wanted event keys, start consuming events for newly wanted event keys and leave the rest untouched.

vibbo:
Reconfig now removes consumers for event keys that are no longer wanted and add consumers for newly appeared event keys. There is no longer possibility of losing events for event keys that were unchanged during reconfig.

Fixes #6982

---
## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/base.py
+++ b/master/buildbot/reporters/base.py
@@ -35,7 +35,7 @@ class ReporterBase(service.BuildbotService):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.generators = None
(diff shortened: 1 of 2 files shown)


p12tic (approved):
Great, thanks!

PR #6997, opened 2023-07-31
Future is now

### LINKED ISSUE #6980: buildbot worker and Python 2 (dependency on future)
I have one simple question – Do you still need to support Python 2 for buildbot worker? I'm asking because this project uses future. future will no longer work with Python 3.12 because imp module has been removed from stdlib and the situation will be even worse with 3.13 because the lib2to3 which is the core of the future will be removed from stdlib as well.

It should be simple to remove the dependency on future, it's a good idea to do it, and I can help you with that if you want to.

frenzymadness:
This PR fixes #6980 - it removes all usages of `__future__` and dependency on `future` library + some other small Python 2/3 fixes.

It's probably not a complete removal of Python 2 code.

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -50,12 +50,10 @@ matrix:
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.7
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.6
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.5
-    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
 
     # Worker tests on python and twisted package combinations.
     # We support worker on Python 2.7, on which Twisted 20.3.0 is the last version that works
     - env: PYTHON=3.9 TWISTED=18.7.0 SQLALCHEMY=latest TESTS=trial_worker
-    - env: PYTHON=2.7 TWISTED=20.3.0 SQLALCHEMY=latest TESTS=trial_worker
 
     # Configuration when SQLite database is persistent between running tests
     # (by default in other tests in-memory SQLite database is used which is
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,6 +1,6 @@
 [settings]
 line_length=110
-known_future_library=__future__,future
+known_future_library=__future__
 known_standard_library=html
 known_twisted=twisted,zope,autobahn,klein,txaio
 known_mock=mock
--- a/master/buildbot/changes/changes.py
+++ b/master/buildbot/changes/changes.py
@@ -13,7 +13,7 @@
 #
 # Copyright Buildbot Team Members
 
-import html  # py2: via future
+import html
 import time
 
 from twisted.internet import defer
--- a/master/buildbot/test/integration/test_upgrade.py
+++ b/master/buildbot/test/integration/test_upgrade.py
@@ -80,8 +80,6 @@ class UpgradeTestMixin(db.RealDatabaseMixin, TestReactorMixin):
                     tf.extract(inf)
                     prefixes.add(inf.name.split('/', 1)[0])
 
-            # (note that tf.extractall isn't available in py2.4)
-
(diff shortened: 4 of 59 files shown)


p12tic (approved):
Looks great, thanks!

p12tic:
Do not merge: this should go after Buildbot 4.0 branches off.

eli-schwartz:
> Do not merge: this should go after Buildbot 4.0 branches off.

Since it appears the future is not "now" after all, I've submitted the much more modest #7267.

a-detiste:
Yes please.

Not like this https://github.com/buildbot/buildbot/pull/7267

eli-schwartz:
@p12tic What's the timeframe for merging this and releasing a new version that incorporates it?

p12tic:
> What's the timeframe for merging this and releasing a new version that incorporates it?

Merging hopefully today, new version of Buildbot will be in couple of weeks.

eli-schwartz:
> > What's the timeframe for merging this and releasing a new version that incorporates it?
> 
> Merging hopefully today, new version of Buildbot will be in couple of weeks.

Disappointingly, the new version of buildbot was released 3 days ago but does not actually incorporate the changes in any way shape or form. It appears that it was never backported.

So, we are back at "buildbot is imminently in danger of being unable to be packaged for python 3.12 and there is no indication of when a version of buildbot that can be installed on systems with python 3.12 will exist".

I would very much appreciate any information you have available regarding when buildbot intends to officially support python 3.12.

p12tic:
We started testing Python 3.12 starting v3.10.1 (ace19ac6f5a465669d302742c80eee920364682f), so I would expect Buildbot to fully work on Python 3.12 starting with that release.

eli-schwartz:
> @eli-schwartz Buildbot test suite is successfully running on Python 3.12 on 3.11.x releases as part of internal Buildbot CI. What exact problems are you seeing?

The "future" package does not work and does not pass tests on python 3.12, which was the original motivating factor for this PR.

This PR would not exist if not for people who are concerned about the dependency on "future" and who reported in #6980 that

> I'm asking because this project uses future. future will no longer work with Python 3.12 because imp module has been removed from stdlib and the situation will be even worse with 3.13 because the lib2to3 which is the core of the future will be removed from stdlib as well.

Frankly, if we cannot even agree on this much then I see no hope for the future of this discussion.

Now, granted, you are not actually using any of the interesting bits of "future", and in particular you aren't using `future.standard_library`, so if you only use one tiny bit of a fatally broken dependency and that one tiny bit works -- then sure, you won't notice the issue in your CI.

However that doesn't help downstream users. Gentoo cannot package "future" for python 3.12 because it's heavily broken and also doesn't pass its own tests (you run the buildbot tests, not the testsuites of all dependencies...) and that's not really negotiable.

Every distro is in this boat, and I seriously doubt anyone is willing to package totally broken software that fails to work just because one tiny utility library that buildbot uses, still works at least for now.

eli-schwartz:
@p12tic,

I was concerned about this exact problem and I offered a less ambitious PR that migrated from "future" to "six" and kept all the python2 compatibility, with the intention that ***my*** PR could be trivially backported to any maintenance branch.

You rejected my PR with the statement that it was not needed because you dropped python2 support and would be backporting it to buildbot 3.x.

Is this true or not?

If it is not true then please re-open my PR and I will retarget it for the maintenance branch.

I would like buildbot to be capable of being packaged in python 3.12, and this ***requires*** stopping the use of "future".

a-detiste:
My ugly patch is here: https://sources.debian.org/src/buildbot/3.10.1-2/debian/patches/remove-future/

Replacing `future` by `six` is the way to go if Py2 compatibility still hascto be retained for a while.

Greetings

p12tic:
@eli-schwartz 

Thanks for explanation.

> The "future" package does not work and does not pass tests on python 3.12, which was the original motivating factor for this PR.
>
> ...
>
> I was concerned about this exact problem and I offered a less ambitious PR that migrated from "future" to "six" and kept all the python2 compatibility, with the intention that my PR could be trivially backported to any maintenance branch.
>
> You rejected my PR with the statement that it was not needed because you dropped python2 support and would be backporting it to buildbot 3.x.

Indeed, there was some confusion on my part. I don't test Buildbot with distribution-provided packages so it didn't occur to me that tests are not sufficient.

I will resurrect your PR and apply to 3.11.x branch and do a release shortly. Thanks!

eli-schwartz:
Thank you! :) It is very much appreciated.

PR #7008, opened 2023-08-02
Allow force scheduler to set request priority

iomartin:
This is the first of a few PRs I'm working on regarding build request priorities.

This PR adds a `priority` parameter to the `ForceScheduler`, so users can tweak their preferences. While the priority is not currently used to choose the next build, users can create their own `nextBuild` if they wish to (I'm working on a PR for this that will come later). The default priority is 0 so this should have no changes in expected behavior.

## Contributor Checklist:


p12tic (approved):
Looks great, thanks!

PR #7010, opened 2023-08-02
Add set_priority method to /buildrequests API

iomartin:
This is the second PR about buildrequests priority, following #7008.

This adds a new option to change the priority for a pending buildrequest, using the /buildrequest API.

I'm having some trouble getting a unit test for this, though. I'm looking at the existing tests that send a `cancel` action via the /buildrequests API but I can't find any that is simple enough that I could tweak for `set_priority`. I would appreciate some help from the maintainers.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/data/buildrequests.py
+++ b/master/buildbot/data/buildrequests.py
@@ -99,11 +99,7 @@ class BuildRequestEndpoint(Db2DataMixin, base.Endpoint):
             return (yield self.db2data(buildrequest))
         return None
 
-    @defer.inlineCallbacks
-    def control(self, action, args, kwargs):
-        if action != "cancel":
-            raise ValueError(f"action: {action} is not supported")
-        brid = kwargs['buildrequestid']
+    def cancel_request(self, brid, args, kwargs):
         # first, try to claim the request; if this fails, then it's too late to
         # cancel the build anyway
         try:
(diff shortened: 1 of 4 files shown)


iomartin:
Fixed the CI error

master/buildbot/data/buildrequests.py
```diff
@@ -99,11 +99,7 @@ def get(self, resultSpec, kwargs):
             return (yield self.db2data(buildrequest))
         return None
 
-    @defer.inlineCallbacks
-    def control(self, action, args, kwargs):
-        if action != "cancel":
-            raise ValueError(f"action: {action} is not supported")
-        brid = kwargs['buildrequestid']
+    def cancelRequest(self, brid, args, kwargs):
```
p12tic, line 102: Let's use `snake_case` for new functions.

p12tic:
I agree about the tests. We would need an integration test for buildrequest priorities. In the integration test the build requests would need to be created before worker is attached (so that it does not pick the first created build request). Currently this is not really supported and would require changes to RunMasterBase class. This is too much to ask from new contributors.

p12tic:
The PR looks good, I only had a small stylistic request. Thanks a lot!

iomartin:
Thanks @p12tic! I fixed the style recommendation

PR #7011, opened 2023-08-03
Use old behavior with no warning after hvac update to v3.0.0

mokibit:
This PR avoids appearing warning after `hvac` is update to v3.0.0 as `raise_on_deleted_version` parameter will change its default value to `False`. The current default of `True` will preserve previous behavior. To use the old behavior with no warning, explicitly set this value to `True`.

### DIFF
--- a/master/buildbot/secrets/providers/vault_hvac.py
+++ b/master/buildbot/secrets/providers/vault_hvac.py
@@ -16,6 +16,9 @@
 HVAC based providers
 """
 
+
+import pkg_resources
+
 from twisted.internet import defer
 from twisted.internet import threads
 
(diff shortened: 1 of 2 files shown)


master/buildbot/secrets/providers/vault_hvac.py
```diff
@@ -140,6 +144,10 @@ def thd_hvac_wrap_read(self, path):
         if self.api_version == 1:
             return self.client.secrets.kv.v1.read_secret(path=path, mount_point=self.secrets_mount)
         else:
+            if self.version >= pkg_resources.parse_version("3.0.0"):
```
p12tic, line 147: elif

PR #7012, opened 2023-08-03
Sort requests and builders by priority

iomartin:
This is another priority PR, after #7008 and #7010.

Now that priorities are no longer always 0, let's use them.

This PR changes the default "nextBuild" and "prioritizeBuilders" to use the priority as the primary sorting key, using the oldest unclaimed build request as the secondary sorting keys.

Note that this does not change default behavior, as by default build requests are created with priority 0 and they are sorted using python's sort() functions, which are [guaranteed to be stable](https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts).

This PR uses higher priority number to mean higher priority (i.e., run before). Some other places prefer lower numbers to mean higher priority. If the maintainers would prefer, I can invert the meaning.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/builder.py
+++ b/master/buildbot/process/builder.py
@@ -197,6 +197,22 @@ class Builder(util_service.ReconfigurableServiceMixin,
         else:
             return None
 
+    @defer.inlineCallbacks
+    def get_highest_priority(self):
+        """Returns the priority of the highest priority unclaimed build request
+        for this builder, or None if there are no build requests.
+
+        @returns: priority or None, via Deferred
+        """
+        bldrid = yield self.getBuilderId()
+        unclaimed = yield self.master.data.get(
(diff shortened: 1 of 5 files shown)


master/buildbot/process/builder.py
```diff
@@ -197,6 +197,22 @@ def getNewestCompleteTime(self):
         else:
             return None
 
+    @defer.inlineCallbacks
+    def getHighestPriority(self):
```
p12tic, line 201: Let's use `snake_case` for new code.

p12tic (approved):
Looks great, thanks. Just had one comment.

iomartin:
Thanks, fixed!

p12tic:
Needs rebase on top of latest master.

p12tic (approved):

PR #7016, opened 2023-08-04
Display buildrequest priority on web UI

iomartin:
Add the priority to the tables on Pending Buildrequets and on each individual builder. Use the  priority for sorting as well.

Unfortunately, I'm having issues building the frontend on my machine, but the changes are simple enough that I felt confident opening this PR. I'm not a node/js developer so not 100% sure about the orderBy syntax but I think it's correct.

## Contributor Checklist:


p12tic (approved):
Also needs changes to the React frontend.

PR #7029, opened 2023-08-08
Scheduler priority

iomartin:
One more buildrequest priority PR (following #7008, #7010, #7012 and #7016).

The goal of this PR is to allow schedulers to set a default priority for buildrequests it creates. It supports either an integer or a callable, to which we pass the builderNames and the codebases + changes related to that buildrequest. The function must return an integer.

The integer path is pretty straightforward, but for the callable option, should I be passing something else? The `changesByCodebase` is a dictionary whose keys are codebase names and the values is a list of changes (as dictionaries).

Feedback welcome on the design.

I would also like to add more robust tests, the ones I added only check that we set the scheduler.priority to the right value, it doesn't check that buildrequests were actually created with the right priority. But I got a bit confused with the framework/fake changes etc, so help is appreciated.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/util/scheduler.py
+++ b/master/buildbot/test/util/scheduler.py
@@ -245,7 +245,7 @@ class SchedulerMixin(interfaces.InterfaceTests):
 
     def fake_addBuildsetForSourceStamps(self, waited_for=False, sourcestamps=None,
                                         reason='', external_idstring=None, properties=None,
-                                        builderNames=None, priority=0, **kw):
+                                        builderNames=None, priority=None, **kw):
         if sourcestamps is None:
             sourcestamps = []
         properties = properties.asDict() if properties is not None else None
--- a/master/buildbot/test/util/scheduler.py
+++ b/master/buildbot/test/util/scheduler.py
@@ -224,7 +224,8 @@ class SchedulerMixin(interfaces.InterfaceTests):
         return self._addBuildsetReturnValue(builderNames)
 
     def fake_addBuildsetForChanges(self, waited_for=False, reason='', external_idstring=None,
-                                   changeids=None, builderNames=None, properties=None, **kw):
+                                   changeids=None, builderNames=None, properties=None,
+                                   priority=None, **kw):
         if changeids is None:
             changeids = []
         properties = properties.asDict() if properties is not None else None
@@ -238,6 +239,7 @@ class SchedulerMixin(interfaces.InterfaceTests):
                     "changeids": changeids,
                     "properties": properties,
                     "builderNames": builderNames,
+                    "priority": priority,
                 }
             )
         )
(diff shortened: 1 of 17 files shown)


iomartin:
Fixed the CI error

master/docs/manual/customization.rst
```diff
@@ -254,6 +254,40 @@ The ``nextBuild`` function is passed as parameter to :class:`BuilderConfig`:
...
+Scheduler Priority Functions
+----------------------------
+When a :class:`Scheduler` is creating a a new :class:`BuildRequest` from a (list of) :class:`Change` (s),it is possible to set the :class:`BuildRequest` priority.
+This can either be an integer or a function, which receives a list of builder names and a dictionary of :class:`Change`, grouped by their codebase.
+
+A simple implementation might look like this:
+
+.. code-block:: python
+
+   def schedulerPriority(builderNames, changesByCodebase):
```
p12tic, line 270: snake_case

p12tic (approved):
Looks great, had just one minor comment.

master/buildbot/test/util/scheduler.py
```diff
@@ -224,7 +224,8 @@ def sourceStampKey(sourceStamp):
         return self._addBuildsetReturnValue(builderNames)
 
     def fake_addBuildsetForChanges(self, waited_for=False, reason='', external_idstring=None,
-                                   changeids=None, builderNames=None, properties=None, **kw):
+                                   changeids=None, builderNames=None, properties=None,
+                                   priority=None, **kw):
```
p12tic, line 228: The purpose of this function is to register how `addBuildsetForChanges` would have been called if it was real. Thus add `priority` to addBuildsetForChanges list of dicts below.

p12tic:
For a more real test you can copy one of basic scheduler test. For example, if you copied `SingleBranchScheduler.test_gotChange_no_treeStableTimer_important`, setup custom priority (either int or callback), then you should be able to check `self.addBuildsetCalls` for priority values.

p12tic:
The design looks sensible. I had more comments about the tests.

iomartin:
The `SingleBranchScheduler` tests do not have `self.addBuildsetCalls`, I'm looking into `test_timed_Nightly.py` which uses it

iomartin:
Added a test with custom priority that checks the buildrequest is created with that priority.
I didn't add one with the ballback because it would need a more significant change in the framework since the tests use the fake_addBuildset* functions.

I also added a couple few more commits to cleanup and reorganize a few things that I'd changed in my previous PRs

p12tic:
> The SingleBranchScheduler tests do not have self.addBuildsetCalls

It's correct that SingleBranchScheduler tests don't currently use `self.addBuildsetCalls`, however they have  `attachScheduler(..., overrideBuildsetMethods=True)` as part of their setup which provides setup for `self.addBuildsetCalls`. Have you found otherwise? I think priority tests aren't related to nightly scheduler, but to base functionality, so they should go to `test_basic.py` file.

p12tic:
Thanks!

PR #7031, opened 2023-08-14
Move newfragments file from root to newfragments directory.

pmisik:
It seems it was accidentally added into root directory by commit 57e834597ff72ad8ac02fb164a16654e0cedabe6

## Contributor Checklist:


pmisik:
If you wish I can squash the commits into single one (this PR is trivial fix not affecting code).

p12tic:
Fixing typo and moving file to correct location is better as two commits. Thanks for the PR.

Issue #7003, opened 2023-08-01
Buildbot is using deprecated pkg_resources


p12tic:
See https://setuptools.pypa.io/en/latest/pkg_resources.html.

PR #7035, opened 2023-08-19
Replace deprecated pkg_resources with importlib-based solutions

mokibit:
This PR solves https://github.com/buildbot/buildbot/issues/7003.

Buildbot is migrating to importlib-based solutions since use of `pkg_resources` is deprecated in favor of [importlib.resources](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata). Some useful APIs are also provided by [packaging](https://pypi.org/project/packaging) (e.g. requirements and version parsing). 

## Contributor Checklist:

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,7 +1,7 @@
 [settings]
 line_length=110
 known_future_library=__future__,future
-known_standard_library=pkg_resources,html
+known_standard_library=html
 known_twisted=twisted,zope,autobahn,klein,txaio
 known_mock=mock
 known_third_party=migrate,sqlalchemy,ldap3,txrequests,requests,MySQLdb,coverage,jinja2,dateutil,sphinx,setuptools,jwt,flask,docutils,aiohttp
--- a/master/buildbot/plugins/db.py
+++ b/master/buildbot/plugins/db.py
@@ -17,7 +17,8 @@
 
 import traceback
 import warnings
-from pkg_resources import iter_entry_points
+from importlib.metadata import distributions
+from importlib.metadata import entry_points
 
 from zope.interface import Invalid
 from zope.interface.verify import verifyClass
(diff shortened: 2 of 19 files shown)


mokibit:
Master's setup.py file code is being run before 'importlib_metadata' is installed from requirements-ci.txt file. At the same time, master's code requires 'importlib_metadata' to be imported as it is already being used in setup.py file.

This PR should be merged after Buildbot master stops supporting Python3.7 and the problem will be fix automatically.

yan12125:
> Master's setup.py file code is being run before 'importlib_metadata' is installed from requirements-ci.txt file. At the same time, master's code requires 'importlib_metadata' to be imported as it is already being used in setup.py file.
> 
> This PR should be merged after Buildbot master stops supporting Python3.7 and the problem will be fix automatically.

Current setup.py uses pkg_resources (or importlib_metadata as proposed here) to make sure nothing unusual happens with pip <1.4 [1]. Buildbot master requires Python >= 3.7, while pip < 1.4 supports up to Python 3.2 [2]. I believe no one is going to use ancient pip to install the latest buildbot, so the pip version check can probably be removed.

[1] https://github.com/buildbot/buildbot/blob/222eeab5e7e8ddb5b7f44cbc968409eb9aea8476/master/setup.py#L473
[2] https://github.com/pypa/pip/blob/1.3.1/setup.py

p12tic:
Now that Python 3.7 has been dropped we can use standard `importlib` module.

yan12125:
I looked into a few test failures, and they are broken by missing 'importlib_resources'

> buildbot.errors.PluginDBError: Unable to load buildbot.www:base: No module named 'importlib_resources'

(https://buildbot.buildbot.net/#/builders/114/builds/1170)

Switching from importlib_metadata and importlib_resources to importlib.metadata and importlib.resources probably fixes them.

yan12125:
Among test failures on buildbot.buildbot.net, some are trivial:

* Missing `packaging` module
* isort issues

But one error is tricky:
```
buildbot.errors.PluginDBError: Unable to load buildbot.www:base: module 'importlib.resources' has no attribute 'files'
```
(https://buildbot.buildbot.net/#/builders/83/builds/3788)

`importlib.resources.files` is added in Python 3.9 [1], so importlib_resources is probably still necessary for older Python.

[1] https://docs.python.org/3/library/importlib.resources.html

PR #7048, opened 2023-08-22
HTTPClientService: Use MAX_THREADS also for txrequests

vibbo:
Minor change


vibbo:
Default of [txrequest.Session](https://github.com/tardyp/txrequests/blob/master/txrequests/sessions.py#L30) is 4 and default of [HTTPClientService](https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/httpclientservice.py#L110) is 5.

So technically yes, but practically it is irrelevant increase.

It's more about enabling (in a [monkey patch way](https://github.com/buildbot/buildbot/blob/master/master/buildbot/util/httpclientservice.py#L106-L108)) to configure HTTPClientService.

PR #7050, opened 2023-08-22
reporter: Prevent Bitbucket status key length exceeding limits

### LINKED ISSUE #7049: Bitbucket cloud status update returns 400 for status_key longer than 40 characters
When using a status_key for reporters.BitbucketStatusPush, there is no check whether it is longer than 40 characters, which appears to be the limit that Bitbucket cloud accepts.
This error appears in the log file in cases where status_key exceeds 40 characters:
`400: unable to upload Bitbucket status b'{"type": "error", "error": {"message": "key: Ensure this value has at most 40 characters (it has 48).", "fields": {"key": ["Ensure this value has at most 40 characters (it has 48)."]}}}'`

Checking the length for this field might be difficult during configuration checking time, since it might exceed 40 characters only after a certain amount of builds have been done ( e.g. with the value Interpolate("%(prop:buildername)s/%(prop:buildnumber)s") )

I would propose hashing the status_key in order to get predictable key lengths. sha1 happens to be 20 bytes long which is 40 characters in hexadecimal notation.

kevle:
This should resolve #7049 .

Since the "key" is not shown in user facing portions of the Bitbucket UI, I think using a hash should be OK.

I do not know whether I have properly used the "yield" keyword here or if I accidentally prevent deferred execution, so it would be great if someone knowing those mechanisms could check whether everything is in order.
`trial buildbot.test.unit.reporters.test_bitbucket` finishes cleanly.

## Contributor Checklist:

* [ ?] I have updated the appropriate documentation

### DIFF
--- a/master/buildbot/reporters/bitbucket.py
+++ b/master/buildbot/reporters/bitbucket.py
@@ -13,6 +13,7 @@
 #
 # Copyright Buildbot Team Members
 
+import hashlib
 from urllib.parse import urlparse
 
 from twisted.internet import defer
@@ -112,9 +113,16 @@ class BitbucketStatusPush(ReporterBase):
         props = Properties.fromDict(build['properties'])
         props.master = self.master
 
+        def key_hash(key):
+            sha_obj = hashlib.sha1()
+            sha_obj.update(key.encode('utf-8'))
+            return sha_obj.hexdigest()
+
+        status_key = yield props.render(self.status_key)
+
         body = {
             'state': status,
-            'key': (yield props.render(self.status_key)),
+            'key': key_hash(status_key),
             'name': (yield props.render(self.status_name)),
             'description': reports[0]['subject'],
             'url': build['url']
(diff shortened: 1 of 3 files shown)


master/buildbot/reporters/bitbucket.py
```diff
@@ -112,9 +113,16 @@ def sendMessage(self, reports):
         props = Properties.fromDict(build['properties'])
         props.master = self.master
 
+        def key_hash(key):
+            sha_obj = hashlib.sha1()
+            sha_obj.update(key.encode('utf-8'))
+            return sha_obj.hexdigest()
+
+        status_key = (yield props.render(self.status_key))
```
p12tic, line 121: There's extra pair of parentheses. Better use `status_key = yield props.render(self.status_key)`. In the line below the parentheses were required because the expression was part of a dict literal.

p12tic (approved):
Looks good, just a small nitpick.

master/buildbot/reporters/bitbucket.py
kevle: Thanks!

Issue #5878, opened 2021-03-08
DirectoryUpload fails with exception when file is not available


aj062:
transfer.FileUpload fails with exception when expected file is not available. It should be gracefully handled, and the step should just fail, instead of causing an exception.

Exception:
```
  File "/Library/Python/2.7/site-packages/twisted/internet/defer.py", line 149, in maybeDeferred
    result = f(*args, **kw)
  File "/Library/Python/2.7/site-packages/buildbot_worker/commands/transfer.py", line 229, in start
    archive.add(self.path, '')
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py", line 2009, in add
    tarinfo = self.gettarinfo(name, arcname)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py", line 1881, in gettarinfo
    statres = os.lstat(name)
exceptions.OSError: [Errno 2] No such file or directory: 'test'
```

mokibit:
After issue reproduction, it is clear that `transfer.FileUpload` fails as expected: with a step failure.
However,  command `directoryUpload` fails with exception, which should be gracefully handled when the directory is not available in worker. Maybe original poster had this command in mind.

The problem with `directoryUpload` has been fixed in https://github.com/buildbot/buildbot/pull/7062.

PR #7062, opened 2023-08-30
Catch UploadDirectory exception when directory is not available in worker

mokibit:
Step transfer.directoryUpload fails with exception when expected directory is not available in worker. This PR handles the exception gracefully, and the step just fails instead of causing an exception.


mokibit:
This PR fixes https://github.com/buildbot/buildbot/issues/5878.

p12tic:
Merged via https://github.com/buildbot/buildbot/pull/7095.

PR #7072, opened 2023-09-01
master: Remove pkg_resources require()

p12tic:
This is a commit from https://github.com/buildbot/buildbot/pull/7035 in an attempt to fix tests. The tests currently fail with errors similar to what's caused by `https://github.com/jazzband/pip-tools/issues/1576`. The errors are unfortunately harder to reproduce than I have time for. Since we're removing the feature anyway, let's save some time by not investigating the root cause and just removing the feature.

The original description of the commit is below:

----

pkg_resources.require() has no substitute in importlib_metadata, so migration is not possible. Removing this function also requests removing the private member '_check_extras' of class _Plugins, as it has no more use cases and is irrelevant.

### DIFF
--- a/master/buildbot/plugins/db.py
+++ b/master/buildbot/plugins/db.py
@@ -194,25 +194,17 @@ class _Plugins:
     represent plugins within a namespace
     """
 
-    def __init__(self, namespace, interface=None, check_extras=True):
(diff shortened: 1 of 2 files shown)


p12tic:
Does not fix tests.

PR #7090, opened 2023-09-06
Retrieve changes by branch

vladbogo:
Add the possibility to retrieve the last changes for each branch. Only retrieve the changes that have an associated buildset.

## Contributor Checklist:


vladbogo:
Hi.

I have created this draft PR in order to add the functionality for being able to retrieve the latest changes for a particular branch. It only retrieves changes that have a buildset associated since if I am not mistaken there might be changes that do not have one.

It would be great to get an initial feedback and I also have a question, mainly related to tests:
* what tests are expected for such a change? For the DB part, should I just create some changes with buildsets and retrieve those?

master/buildbot/db/changes.py
```diff
@@ -250,6 +250,34 @@ def thd(conn):
...
+            else:
+                q = sa.select([changes_tbl.c.changeid.distinct()]).select_from(
+                    from_clause).where(changes_tbl.c.branch == branch)
+            q = q.limit(count).order_by(changes_tbl.c.changeid.desc())
+            rp = conn.execute(q)
+            changeids = [self._getDataFromRow(row) for row in rp]
+            rp.close()
+            return list(changeids)
+
+        d = self.db.pool.do(thd)
```
p12tic, line 272: Use `yield` here and `@defer.inlineCallbacks` on getChangesForBranch.

master/buildbot/db/changes.py
```diff
@@ -250,6 +250,34 @@ def thd(conn):
...
+                q = sa.select([changes_tbl.c.changeid.distinct()]).select_from(
+                    from_clause)
+            else:
+                q = sa.select([changes_tbl.c.changeid.distinct()]).select_from(
+                    from_clause).where(changes_tbl.c.branch == branch)
+            q = q.limit(count).order_by(changes_tbl.c.changeid.desc())
+            rp = conn.execute(q)
+            changeids = [self._getDataFromRow(row) for row in rp]
+            rp.close()
+            return list(changeids)
```
p12tic, line 270: I think changeids is already a list.

master/buildbot/data/changes.py
```diff
@@ -104,8 +105,12 @@ def get(self, resultSpec, kwargs):
                 changes = []
         else:
             if resultSpec is not None:
-                resultSpec.fieldMapping = self.fieldMapping
```
p12tic, line 107: I may be missing something, but why can't we use fieldMapping for `branch is not None` case? Other potential filters may still be relevant.

master/buildbot/data/changes.py
```diff
@@ -42,6 +42,7 @@ def _fixChange(self, change, is_graphql):
                     {'name': k, 'source': v[1], 'value': json.dumps(v[0])}
                     for k, v in props.items()
                 ]
+                change['builds'] = yield self.master.db.builds.getBuildsForChange(change['changeid'])
```
p12tic, line 45: Just out of interest, are you using the graphql API?

p12tic:
For now I've commented on some trivial things that I noticed. For your questions I will come back a bit later.

master/buildbot/data/changes.py
vladbogo: no, not using graphql

master/buildbot/data/changes.py
vladbogo: Yes, I think other fields may be relevant. However, based on the table joins, some of them might have a different result which might be confusing since it will filter all the entries for a particular branch. Also, some of them might be trickier to integrate. I'll look into this in more detail.

master/buildbot/db/changes.py
vladbogo: done

master/buildbot/db/changes.py
vladbogo: not sure if it's correct in the new form since I think it will only return the changeids. Or am I wrong? Any suggestions?

master/buildbot/data/changes.py
```diff
@@ -42,6 +42,9 @@ def _fixChange(self, change, is_graphql):
                     {'name': k, 'source': v[1], 'value': json.dumps(v[0])}
                     for k, v in props.items()
                 ]
+                change['builds'] = yield self.master.db.builds.getBuildsForChange(
```
p12tic, line 45: I think this will be slow when retrieving many changes. Let's skip it for now, especially if you aren't using graphql.

p12tic:
I looked more into the PR and now I wonder whether existing code would support the use case for retrieving changes for branch.

When checking the buildbot API it already supports retrieving changes by branch. For example: https://buildbot.buildbot.net/api/v2/changes?limit=50&branch=refs/pull/XYZ/merge&order=when_timestamp. I see that when_timestamp is incorrect when it's filled by data from Github, so this needs additional investigation. But basic functionality is already there, don't you think?

For retrieving buildsets for branch, there's no such functionality, so that part could go in.

For tests, what you wrote makes sense. For an example, you can check tests in master/buildbot/test/unit/data/test_buildsets.py

master/buildbot/data/changes.py
vladbogo: Based on further looking, I think it should be possible to use the resultSpec, but that would be on top of the join between the changes and buildset tables, so some filters might have a different meaning between the case where 'branch' is specified or not. Is this ok?

master/buildbot/data/changes.py
vladbogo: Ok, I see. Probably it does't make sense to always fetch them. But can it be at least added when `branch` is specified? The reason the builds are needed is to directly show them in the grid view without any additional calls. Maybe they can be merged in a single query. What do you think would be the best approach? Also, if there are two many changes, we can leave it out for now and I'll work on a new PR to add this.

vladbogo:
Yes, you are right. However, if I am not mistaken, this will include changes that don't have a build associated. I think that changes that do not have commits can be added only when a push contains multiple commits. So, in this case, you would get the latest X changes for a branch, but from a grid view perspective, you won't be able to get the latest X changes that actually have a buildbot run.

Basically what the PR adds, is the join with the buildset table, that ensure that each returned change, has a build associated. And it only does this if the `branch` is specified which I agree is not the cleanest way. I am not sure if there is a better/easier way to achieve this.  Do you think a different endpoint would be more suitable?

vladbogo:
@p12tic any thoughts on the above comments?

p12tic:
> any thoughts on the above comments?

@vladbogo Sorry for forgetting about this PR. I thought it makes sense for me to look into adding this feature into the existing API, but didn't have time for that yet..

vladbogo:
I see. Please let me know if I can contribute in any way

vladbogo:
@p12tic are there any updates on this?

PR #7109, opened 2023-09-15
p4: fix got_revision not taking into account changelist only deleting files

tdesveaux:
This is built onto #7108 so I'll leave this in draft until it's resolved.

This PR's goal is to fix the `got_revision` in the following situation.

Let's assume a Perforce changelist 99 which does file modification/addition/etc.
And a Changelist 100 which only remove files.

The current code will correctly sync changelist 100.
However, `p4 changes -m1 #have` will return changelist 99 as there is no files from changelist 100 in the client's havelist.

To circumvent this, I retrieve the current change seen by the client specs with `p4 changes -m1 //{p4client}/...` (with an optionally given target revision), then use this to sync at the given revision which will be used to set `got_revision`.

Since I needed a client for this, I moved the client creation out of the `mode_` functions, which could be considered a breaking change. Let me know if this is an issue.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -1691,3 +1691,4 @@ yieldmetricsvalue
 za
 zadka
 zope
+ztag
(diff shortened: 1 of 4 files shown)


p12tic (approved):

p12tic:
Rebased and fixed pylint error.

PR #7129, opened 2023-10-04
www/react: Fix ChangeDetails for change with empty revlink

tdesveaux:
Change can have a `revlink` which is an empty string.
In this case, `ChangeDetails` would display the change message as a link with a href to `""`.


p12tic (approved):

p12tic:
Rebased on latest master.

PR #7157, opened 2023-10-11
process/buildrequestdistributor: sort unclaimed br cache by requestid

iomartin:
When choosing the next buildrequest to build, we first look at their priority, then break ties by choosing the oldest ("submitted_at" field).

However, if multiple buildrequests were created at the same time, they will all have the same "submitted_at". This makes it unpredictable which will be selected.

To make it predictable, we can break ties using the buildrequestid instead. This preserves the property of choosing the oldest (as a br with a lower id will have a submitted_at <= then a br with a higher id).

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -200,6 +200,7 @@ buildreq
 buildrequest
 buildrequestcompletions
 buildrequestdistributor
+buildrequestid
 buildrequestresults
 buildrequests
 buildrequestsconnectorcomponent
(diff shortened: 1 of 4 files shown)


iomartin:
Any feedback @p12tic?

I believe the errors are unrelated to my changes

p12tic (changes requested):
Thanks for the PR. It looks good, but contains accidental changes to `master/docs/relnotes/index.rst` which should be reverted.

Please also rebase on latest master.

iomartin:
Ok so it seems that my changes are in fact responsible for the failed CI test as the other PR I just opend (#7208) didn't fail that test. I'm looking into it.

iomartin:
Oh nevermind. After running a couple more times, I noticed that it was failing on different tests every time, and now it eventually passed. They were timeout errors, so the machine running them was likely overloaded.

It should be good to merge if you are happy with the changes

p12tic (approved):

PR #7159, opened 2023-10-14
allow secrets to be group-readable

Mic92:
Systemd's LoadCredential feature makes secrets group-readable (0o440) permissions. Services cannot modify secrets or their permission as they are mounted read-only.
By relaxing the check we allow buildbot to use secrets provided by systemd. In pratice group-readable secrets are a common pattern to share secrets between different process of the same group and buildbot shouldn't restrict this.

## Remove this paragraph

And especially:

## Contributor Checklist:


p12tic (approved):
Looks reasonable, thanks!

flokli:
Related: https://github.com/systemd/systemd/issues/29435

PR #7181, opened 2023-10-23
www-react: Fix crash on build ending

tdesveaux:
`RawData.renderDataElement` would receive an `undefined` value and try to call `.toString()` on it, resulting on a blank page.

I took the liberty to make some changes to the function as well as I felt it more legible. Let me know if you'd rather I revert it back to it's previous form.

### DIFF
--- a/www/react-base/src/components/RawData/RawData.tsx
+++ b/www/react-base/src/components/RawData/RawData.tsx
@@ -55,6 +55,9 @@ export const RawData = ({data}: RawDataProps) => {
     if (value === null) {
       return <dd>{"null"}&nbsp;</dd>;
     }
+    if (value === undefined) {
+      return <dd>{"undefined"}&nbsp;</dd>;
+    }
     if (isArrayOfObjectsRaw(value)) {
       return (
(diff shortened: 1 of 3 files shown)


www/react-base/src/components/RawData/RawData.tsx
```diff
@@ -52,25 +52,30 @@ export const RawData = ({data}: RawDataProps) => {
   }
 
   const renderDataElement = (value: any) => {
-    if (!isObjectRaw(value) && !isArrayOfObjectsRaw(value)) {
-      return <dd>{value === null ? "null" : value.toString()}&nbsp;</dd>;
+    if (value === null) {
+      return <dd>{"null"}&nbsp;</dd>;
     }
-    if (isArrayOfObjectsRaw(value)) {
+    else if (value === undefined) {
```
p12tic, line 58: Else is not needed if the preceding if branches returned from function.

p12tic (changes requested):
Looks good, I just had a tiny comment. Thanks a lot!

PR #7198, opened 2023-11-02
Draft: DockerLatentWorker: Explicitly disconnect when stopping instance

### LINKED ISSUE #4678: Worker reconnect fail due to exception while attaching
We have had a few instances of workers losing the network connection, and not get properly attached when they try to reconnect.

As far as I can tell, the connection is re-established, the master accepts it, but then some kind of exception happens while it is handling the registration of the worker.

Afterwards, the connection seems to be OK from the worker's side (the master does not disconnect, and apparently responds to keep-alives), but from the master's side, the connection is not registered as active, and it is not associated with any builders.

Fixing this situation either requires manually stop/restart of the worker instance on each worker, or a stop/restart of the master instance.

Environment: Buildbot 2.0.1, TLS connection between worker and master.

The exception reported is as follows; there is no information about what the assert is about.

2019-03-01 11:24:05+0000 [Broker (TLSMemoryBIOProtocol),236,1.2.3.4] worker 'arbeider' attaching from IPv4Address(type='TCP', host='1.2.3.4', port=51630)
2019-03-01 11:24:05+0000 [Broker (TLSMemoryBIOProtocol),236,1.2.3.4] Got duplication connection from 'arbeider' starting arbitration procedure
2019-03-01 11:24:15+0000 [-] Connected worker 'arbeider' ping timed out after 10 seconds
2019-03-01 11:24:15+0000 [-] Old connection for 'arbeider' was lost, accepting new
2019-03-01 11:24:15+0000 [Broker (TLSMemoryBIOProtocol),236,1.2.3.4] Got workerinfo from 'arbeider'
2019-03-01 11:24:15+0000 [Broker (TLSMemoryBIOProtocol),236,1.2.3.4] worker arbeider cannot attach
          File "sandbox/lib/python3.6/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
          File "sandbox/lib/python3.6/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
          File "sandbox/lib/python3.6/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
          File "sandbox/lib/python3.6/site-packages/buildbot/worker/base.py", line 638, in attached
            log.err(e, "worker %s cannot attach" % (self.name,))
          File "sandbox/lib/python3.6/site-packages/buildbot/worker/base.py", line 636, in attached
            yield AbstractWorker.attached(self, bot)
        builtins.AssertionError:

cmouse:
Seems to fix #4678

## Contributor Checklist:

### DIFF
--- a/master/buildbot/worker/docker.py
+++ b/master/buildbot/worker/docker.py
@@ -379,3 +379,6 @@ class DockerLatentWorker(CompatibleLatentWorkerMixin,
             except docker.errors.APIError as e:
                 log.msg('Error while removing the image: %s', e)
         docker_client.close()
+        conn = self.conn
+        self.conn = None
+        self._disconnect(conn)


YngveNPettersen:
I suspect this patch, as currently implemented, will not fix https://github.com/buildbot/buildbot/issues/4678 , although it may fix a different aspect of it that has been mentioned, the one related to docker/latent workers, so it should not close that report automatically.

The environment for which https://github.com/buildbot/buildbot/issues/4678 was reported does not use latent workers.

I think I've just discovered the real cause of the problem we experienced, and is about to deploy a hotfix for it.

PR #7213, opened 2023-11-12
pretty print argv

Mic92:
Before:

argv: [b'some', b'args', b'with spaces']

After:

argv: some args 'with spaces'

This is nice because we can easily copy'n'paste commands like this into
a shell to locally reproduce them.

## Contributor Checklist:


Mic92:
Mhm, the test code allows to mix both binaries and string? Seems a bit messy.

PR #7219, opened 2023-11-17
A hacky modification to hide the sensitive information from the workers (fake_command)

hydrogencl:
I want to add a hacky method (using only re.sub) to hide the sensitive information (a.k.a USER:Token) in case some repository is not full open and need to be accessed via token protection. 

The `fake_command` and `display` in the buildbot_worker/runprocess.py will reveal the user:token information in the stdout log in the webportal. This `re.sub` can hide the information and can prevent being written in the report. 

## Contributor Checklist:

    - There is no need for the documentation since this might be a secuity issue for repositories.


worker/buildbot_worker/runprocess.py
```diff
@@ -298,8 +298,12 @@ def __init__(self, command_id, command, workdir, unicode_encoding, send_update,
         @param useProcGroup: (default True) use a process group for non-PTY
             process invocations
         """
+<<<<<<< HEAD
+        self.secret_string = "########"
+=======
         self.command_id = command
 
+>>>>>>> 09a54abffc1ea98dae2fe39df058759c2930a118
```
Mic92, line 306: Merge conflict?

hydrogencl: i will fix it soon

PR #7220, opened 2023-11-19
adding a new fragment for cover-up the user:token information from runprocess.py

hydrogencl:
I add a re.sub solution to cover-up the sensitive information (i.e. token) if people are using a close repository but they have to use the token instead of ssh connection. This cover-up will cover the information between `https://` and the `git url`. Only the class RunProcess in `worker/runprocess.py`  is modified. 

## Contributor Checklist:


hydrogencl:
Can someone help for the `bb/e2e_react_tgz/3.9/tw:latest/sqla:latest/db:sqlite:///wp: — Build done`? I don't quite understand the meaning of this build and can't really fix it ...

p12tic:
@hydrogencl Could you please file a bug and tell us what issue you're facing. The PR unfortunately can't go in as is because the code changes are more of a hack than a proper solution.

p12tic:
Closing for now. Please reopen if you plan to improve the PR.

hydrogencl:
Sorry, I just found out that my comment was not successfully submitted. 

The problem is that:

If people want to use non-open repositories, but the ssh connection is not usable (e.g. in some HPC sites the ssh connection is not allowed due to security reasons), we have to use `user:token` to do the connection. However the infomation of token will be revealed in the bb logs due to the nature of `runprocess.py`. 

TL;DR: How do we use `user:token` method without revealing the token information in the buildbot log?

p12tic:
@hydrogencl You can use Secrets: https://docs.buildbot.net/latest/manual/secretsmanagement.html. These would automatically be stripped out of the logs.

PR #7026, opened 2023-08-07
Add support for Python 3.12

pmisik:
## Contributor Checklist:

* [ ] I have updated the unit tests
* [ ] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* [ ] I have updated the appropriate documentation

PR #7228, opened 2023-12-03
add python 3.12 to ci matrix

Mic92:
This version was released in october, so we should also support this in buildbot

## Contributor Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
 
     steps:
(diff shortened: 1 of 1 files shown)


yan12125:
Looks like this pull request and https://github.com/buildbot/buildbot/pull/7229 are already integrated as part of https://github.com/buildbot/buildbot/pull/7026

Mic92:
> Looks like this pull request and #7229 are already integrated as part of #7026

Seems like it, but potentially CI won't be turned on this pull request yet: https://github.com/buildbot/buildbot/pull/7026#issuecomment-1868302374

yan12125:
> > Looks like this pull request and #7229 are already integrated as part of #7026
> 
> Seems like it, but potentially CI won't be turned on this pull request yet: [#7026 (comment)](https://github.com/buildbot/buildbot/pull/7026#issuecomment-1868302374)

Thanks, I misunderstood that comment.

Now all relevant pull requests are meeged. Thank you all for the great work!

PR #7230, opened 2023-12-03
actually drop python2.7 from ci

Mic92:
This has been EOL for a long time and shouldn't be used for network services anymore

## Contributor Checklist:

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -52,7 +52,6 @@ matrix:
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.5
 
     # Worker tests on python and twisted package combinations.
-    # We support worker on Python 2.7, on which Twisted 20.3.0 is the last version that works
     - env: PYTHON=3.9 TWISTED=18.7.0 SQLALCHEMY=latest TESTS=trial_worker
 
     # Configuration when SQLite database is persistent between running tests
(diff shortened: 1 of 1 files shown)


p12tic:
Python 2.7 support will be removed on the next release after Buildbot 3.10. Then we can merge this PR.

p12tic:
> Ok. Is there a list of things left to do for 3.10? I would like to help.

I think https://github.com/buildbot/buildbot/pull/7239 was the last thing needed for 3.10. It will be released in coming days.

.bbtravis.yml
```diff
@@ -97,13 +88,7 @@ install:
     cmd: |
       set -e
       rm -rf /tmp/bbworkervenv
-      if [ "$PYTHON" = "2.7" ]; then
```
p12tic, line 100: This should have been WORKER_PYTHON...

.bbtravis.yml
```diff
@@ -49,12 +49,9 @@ matrix:
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.7
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.6
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.5
-    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=2.7
```
p12tic, line 52: For interoperability tests we should still test that old Buildbot worker can connect to newer master. We can simply install buildbot worker from pip. It would make sense to add PIP_WORKER_VERSION and install that package instead of installing from the git checkout.

p12tic: I've opened https://github.com/buildbot/buildbot/issues/7288 for this comment. This PR does not need to be held up by that.

PR #7246, opened 2023-12-04
Using legacy provider for ssl.

gl-yziquel:


mokibit:
Hi, @gl-yziquel,
this issue was only relevant when AngularJS frontend was used for Buildot UI in older versions and this workaround helped to fix it.
Now React frontend is used and AngularJS has been removed recently (https://github.com/buildbot/buildbot/pull/7287), therefore I will close this issue.

PR #7249, opened 2023-12-04
www/react: Implement case insensitive search in logs as an option

tdesveaux:
The default is still case-sensitive search.

Style/CSS is pretty janky, that's really not my skills there, so I'd welcome anyone that want to clean it up.

I may implement regex search if there is interest (on our users side, or here). Did not do it right here as there is issues with result highlighting due to assumption that highlighting should be from found index beginning + length of input (which is no longer the case with a regex). I guess there will also be issues with matches that span more than one line.

Also, I add to implement a `regexEscape` because there is none available directly in JS...
I didn't know if there was a guideline about third-party lib, but if not, it might be worth it to use one to delegate testing to it.

### DIFF
--- a/www/react-base/src/components/LogSearchField/LogSearchField.tsx
+++ b/www/react-base/src/components/LogSearchField/LogSearchField.tsx
@@ -18,26 +18,51 @@
 import './LogSearchField.scss'
 import {Ref, useState} from "react";
 import {FaChevronDown, FaChevronUp, FaSearch} from "react-icons/fa";
+import {VscCaseSensitive} from "react-icons/vsc";
 
 export type LogSearchButtonProps = {
   currentResult: number;
   totalResults: number;
-  onTextChanged: (text: string) => void;
+  onSearchInputChanged: (text: string, caseSensitive: boolean) => void;
   onPrevClicked: () => void;
   onNextClicked: () => void;
   inputRef: Ref<HTMLInputElement>;
 };
 
 export const LogSearchField = ({currentResult, totalResults,
-                                onTextChanged, onPrevClicked, onNextClicked,
+                                onSearchInputChanged, onPrevClicked, onNextClicked,
                                 inputRef}: LogSearchButtonProps) => {
   const [searchText, setSearchText] = useState<string>('');
   const [hasFocus, setHasFocus] = useState<boolean>(false);
+  const [isCaseSensitive, setIsCaseSensitive] = useState<boolean>(true);
 
   const onSearchTextChanged = (text: string) => {
     setSearchText(text);
-    onTextChanged(text);
+    onSearchInputChanged(text, isCaseSensitive);
   };
+
+  const onCaseInsensitiveToggled = () => {
+    const newValue = !isCaseSensitive;
+    setIsCaseSensitive(newValue);
+    onSearchInputChanged(searchText, newValue);
+  }
+
+  const optionButtonClass = (optionName: string, isToggled: boolean) => {
+    return [
+      'bb-log-search-field-options',
+      `bb-log-search-field-options-${optionName}`,
+      `bb-log-search-field-options-${isToggled ? 'toggled' : 'untoggled'}`,
+    ].join(' ');
+  };
+  const renderOptions = () => (
+    <>
+      <button
+        className={optionButtonClass('case', isCaseSensitive)}
+        onClick={onCaseInsensitiveToggled}>
+        <VscCaseSensitive />
+      </button>
+    </>
+  );
 
   const renderSearchNav = () => (
     <div className="bb-log-search-field-nav">
@@ -58,16 +83,19 @@ export const LogSearchField = ({currentResult, totalResults,
     }
   }
 
+  const shouldRenderOptionals = (hasFocus || searchText !== '');
+
   return (
     <form role="search" className="bb-log-search-field">
       <FaSearch className="bb-log-search-field-icon"/>
+      {shouldRenderOptionals ? renderOptions() : <></>}
       <input className="bb-log-search-field-text" type="text" value={searchText}
              ref={inputRef}
              onFocus={() => setHasFocus(true)} onBlur={() => setHasFocus(false)}
              onChange={e => onSearchTextChanged(e.target.value)}
              onKeyDown={e => onKeyDown(e)}
              placeholder="Search log"/>
-      {(hasFocus || searchText !== '') ? renderSearchNav() : <></>}
+      {shouldRenderOptionals ? renderSearchNav() : <></>}
     </form>
   );
 }
--- a/www/react-base/src/util/Regex.test.ts
+++ b/www/react-base/src/util/Regex.test.ts
@@ -1,0 +1,31 @@
+/*
+  This file is part of Buildbot.  Buildbot is free software: you can
+  redistribute it and/or modify it under the terms of the GNU General Public
+  License as published by the Free Software Foundation, version 2.
+
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+  FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+  details.
+
+  You should have received a copy of the GNU General Public License along with
+  this program; if not, write to the Free Software Foundation, Inc., 51
+  Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+  Copyright Buildbot Team Members
+*/
+
+import {regexEscape} from "./Regex";
+
+describe('Regex', () => {
+  describe('escape', () => {
+
+    it('no change', () => {
+      expect(regexEscape("this is a simple test")).toStrictEqual("this is a simple test");
+    });
+    it('escapes symbols', () => {
+      expect(regexEscape("/ \\ [ ] ( ) - ^ $ . | ? + * { }"))
+        .toStrictEqual("\\/ \\\\ \\[ \\] \\( \\) \\- \\^ \\$ \\. \\| \\? \\+ \\* \\{ \\}");
+    });
+  });
+});
--- a/newsfragments/www-case-insensitive-search.feature
+++ b/newsfragments/www-case-insensitive-search.feature
@@ -1,0 +1,1 @@
+Add support for case insensitive search within the logs.
--- a/www/react-base/src/components/LogSearchField/LogSearchField.scss
+++ b/www/react-base/src/components/LogSearchField/LogSearchField.scss
@@ -3,9 +3,9 @@
 
 .bb-log-search-field {
   position: relative;
-  width: 270px;
+  width: 300px;
   display: inline-block;
-  margin-right: 1rem;
+  margin-right: 0.5rem;
 }
 
 .bb-log-search-field-icon {
@@ -28,6 +28,28 @@
   padding: 0.5rem;
 }
 
+.bb-log-search-field-options {
+  position: absolute;
+  top: 0.4rem;
+  color: $body-color;
+  padding: 0;
+  border: none;
+  background: none;
+}
+
+.bb-log-search-field-options-toggled {
+  border: rgb(0, 0, 255);
+  background-color: rgba(0, 0, 255, 0.4);
+}
+.bb-log-search-field-options-untoggled {
+  border: none;
+  background-color: rgba(255, 255, 255, 0);
+}
+
+.bb-log-search-field-options-case {
+  margin-left: 1.75rem;
+}
+
 .bb-log-search-field-nav-button {
   border-radius: 0.25rem;
   border: 0px solid transparent;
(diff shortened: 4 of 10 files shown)


www/react-base/src/util/Regex.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import {regexEscape} from "./Regex";
+
+describe('Regex', () => {
+  describe('escape', () => {
+
+    it('no change', () => {
+      expect(regexEscape("this is a simple test")).toStrictEqual("this is a simple test");
+    });
+    it('escapes symbols', () => {
+      expect(regexEscape("/ \\ [ ] ( ) - ^ $ . | ? + * { }")).toStrictEqual("\\/ \\\\ \\[ \\] \\( \\) \\- \\^ \\$ \\. \\| \\? \\+ \\* \\{ \\}");
```
p12tic, line 27: Let's keep lines shorter than 100 chars.

p12tic:
Unfortunately regex search is much slower than indexOf. Case-insensitive search can be implemented by doing toLowerCase() on both search term and text and it's still faster than regex.

For performance results see https://perf.link/#eyJpZCI6ImdhZHpocW1pN3JhIiwidGl0bGUiOiJDb21wYXJpbmcgdGV4dCBzZWFyY2giLCJiZWZvcmUiOiJjb25zdCB0ZXh0ID0gXCJhYmNkZWFiY2RcIi5yZXBlYXQoMTAwMDApO1xuY29uc3Qgc2VhcmNoID0gXCJhXCI7XG5jb25zdCByID0gbmV3IFJlZ0V4cChzZWFyY2gsIFwiZ1wiKTtcbiIsInRlc3RzIjpbeyJuYW1lIjoidGV4dCIsImNvZGUiOiJsZXQgcG9zID0gMDtcbndoaWxlICh0cnVlKSB7XG4gICAgcG9zID0gdGV4dC5pbmRleE9mKHNlYXJjaCwgcG9zKTtcbiAgICBpZiAocG9zIDwgMCkgYnJlYWs7XG4gICAgcG9zICs9IHNlYXJjaC5sZW5ndGg7XG59IiwicnVucyI6WzU4MCw3NzQsNTQ4LDU2OSw2NzcsNzUyLDY3Nyw2NjYsNzIwLDU4MCwzOTcsNDYyLDQxOSwzNDQsNDk0LDM5NywyNjgsMjkwLDQ1MSwzMzMsMzExLDMzMywzMDEsMzAxLDI2OCwzNjUsMzExLDM2NSwzNDQsMjY4LDIwNCwyOTAsMzMzLDMyMiwyNzksMjkwLDMwMSwyOTAsMzY1LDMzMywyNTgsMjY4LDMxMSwyOTAsMzMzLDI5MCwzODcsNDE5LDI3OSwyMjUsMjA0LDI1OCwzMjIsMzAxLDE4MiwyMjUsMjE1LDI3OSwzNzYsMjY4LDMzMywzMDEsMzQ0LDI0NywyNjgsMjkwLDE4MiwyNDcsMzAxLDI2OCwxMjksMzMzLDM1NCwyNDcsMzQ0LDI5MCwzMzMsMzY1LDI3OSwzMjIsMjU4LDMwMSwzMzMsMzIyLDMwMSwyNTgsMjU4LDMzMywyNzksMzIyLDMwMSwzMDEsNDE5LDMzMywyOTAsMzU0LDM5NywzNDQsMjkwLDI2OF0sIm9wcyI6MzQzfSx7Im5hbWUiOiJyZWdleCIsImNvZGUiOiJBcnJheS5mcm9tKHRleHQubWF0Y2hBbGwocikpOyIsInJ1bnMiOls2NCw3NSw5Niw5Niw4Niw2NCw5Niw2NCw3NSw1Myw1Myw2NCw1Myw0Myw0Myw0Myw1Myw0Myw0MywzMiwzMiw0Myw0Myw0MywzMiwzMiwzMiw1MywzMiw0MywzMiwzMiwzMiw0Myw0MywzMiwzMiwzMiw0Myw0Myw0Myw1MywzMiwzMiwzMiwzMiwzMiwzMiw0MywzMiwzMiwyMSwzMiw1MywzMiwzMiwzMiwyMSw0Myw0MywzMiwzMiw0MywzMiwzMiwzMiwyMSw0Myw0Myw0MywzMiw0MywzMiw0MywzMiwyMSwyMSw0Myw0Myw0Myw0Myw0MywzMiw0Myw0MywzMiw0Myw0Myw0MywzMiw0MywzMiwzMiw0MywzMiwzMiw0Myw0Myw0Myw0M10sIm9wcyI6NDF9LHsibmFtZSI6InRleHQgbG93ZXJjYXNlIiwiY29kZSI6ImNvbnN0IGxvd2VyID0gdGV4dC50b0xvd2VyQ2FzZSgpO1xubGV0IHBvcyA9IDA7XG53aGlsZSAodHJ1ZSkge1xuICAgIHBvcyA9IGxvd2VyLmluZGV4T2Yoc2VhcmNoLCBwb3MpO1xuICAgIGlmIChwb3MgPCAwKSBicmVhaztcbiAgICBwb3MgKz0gc2VhcmNoLmxlbmd0aDtcbn0iLCJydW5zIjpbNzUyLDczMSw1MTYsNTI2LDU5MSw3NDEsNjU1LDY5OCw1MjYsNTkxLDYwMiw0OTQsNDQwLDUxNiwzMDEsMzY1LDM0NCwzNTQsMzExLDM1NCw0MzAsMzY1LDI1OCwyOTAsMjY4LDMwMSwyNTgsMjU4LDMxMSwzNTQsMjc5LDM5NywzMjIsNDA4LDI5MCwyNzksMjU4LDMyMiwzMDEsMzAxLDI2OCwyNTgsMzAxLDM0NCwyMDQsMjU4LDIzNiwyNzksMjkwLDE3MiwyNTgsMjE1LDIzNiwyMjUsMzExLDMxMSwxNzIsMjI1LDI3OSwyNTgsMjQ3LDM4NywyOTAsMzExLDM1NCwyNDcsMTgyLDI0NywyMzYsMjM2LDI5MCwzMTEsMjkwLDMyMiwyNDcsMTkzLDE3MiwyNDcsMTcyLDI1OCwzMDEsMzExLDI2OCwyMjUsMzAxLDMxMSwyNTgsMzAxLDM0NCwzMjIsMzAxLDM5NywyNTgsMzAxLDMwMSwyNjgsMzAxLDI3OSwzMzMsNDMwXSwib3BzIjozMzF9XSwidXBkYXRlZCI6IjIwMjMtMTItMDRUMjE6NTY6MTEuNDY4WiJ9.

tdesveaux:
Tried with logs from a Buildbot as a test sample (with a js script run with node, I don't think fetch is available in the perflink context?).

Difference seems reasonable to me. But, I'll leave the final decision to you and I can update the PR with `toLowerCase` if needed.

Tests are in the same order as your link, with a 4th one being a regex with the case insensitive flag.
```bash
$ node test.js 
Search: webpack
Execution time: 1701 ms
Execution time: 2130 ms
Execution time: 6604 ms
Execution time: 1651 ms
Search: notfound
Execution time: 2023 ms
Execution time: 2029 ms
Execution time: 7137 ms
Execution time: 1119 ms
```

And the script:
```js
async function main(search) {
  const response = await fetch(
    "https://buildbot.buildbot.net/api/v2/logs/1404552/contents?offset=515&limit=5000",
  );
  const json = await response.json();
  const text = json["logchunks"][0]["content"]

  const r = new RegExp(search, "g");
  const r2 = new RegExp(search, "gi");

  function test1() {
    let pos = 0;
    while (true) {
      pos = text.indexOf(search, pos);
      if (pos < 0) break;
      pos += search.length;
    }
  }

  function test2() {
    Array.from(text.matchAll(r));
  }

  function test3() {
    const lower = text.toLowerCase();
    let pos = 0;
    while (true) {
      pos = lower.indexOf(search, pos);
      if (pos < 0) break;
      pos += search.length;
    }
  }

  function test4() {
    Array.from(text.matchAll(r2));
  }

  console.log(`Search: ${search}`);
  for (const func of [test1, test2, test3, test4]) {
    const start = Date.now();

    for (let i = 0; i < 10000; ++i) {
      func();
    }

    const end = Date.now();
    console.log(`Execution time: ${end - start} ms`);
  }
}

main("webpack").then(() => main("notfound"))
```

p12tic:
Thanks for the test script. I think shorter search strings are more relevant, because people type search string symbol by symbol.

This is what I get from your script for short strings:
```
Search: a
Execution time: 4174 ms
Execution time: 22140 ms
Execution time: 27086 ms
Execution time: 30228 ms
Search: p
Execution time: 4062 ms
Execution time: 13542 ms
Execution time: 26800 ms
Execution time: 22669 ms
Search: z
Execution time: 384 ms
Execution time: 870 ms
Execution time: 24885 ms
Execution time: 7384 ms
Search: zz
Execution time: 279 ms
Execution time: 328 ms
Execution time: 23491 ms
Execution time: 3589 ms
```

On Firefox toLowerCase is less slower (10x less iterations):
```
Search: a 
Execution time: 1418 ms 
Execution time: 5082 ms 
Execution time: 4955 ms 
Execution time: 6579 ms 
starting test 
Search: p 
Execution time: 1065 ms 
Execution time: 4738 ms 
Execution time: 3459 ms 
Execution time: 3924 ms 
starting test 
Search: z 
Execution time: 77 ms 
Execution time: 143 ms 
Execution time: 4127 ms 
Execution time: 688 ms 
starting test 
Search: zz 
Execution time: 225 ms 
Execution time: 220 ms 
Execution time: 3726 ms 
Execution time: 605 ms 
```

Chrome:
```
Search: a
Execution time: 918 ms
Execution time: 2374 ms
Execution time: 2643 ms
Execution time: 2869 ms
starting test
Search: p
Execution time: 428 ms
Execution time: 1270 ms
Execution time: 2323 ms
Execution time: 2057 ms
starting test
Search: z
Execution time: 51 ms
Execution time: 104 ms
Execution time: 2521 ms
Execution time: 815 ms
starting test
Search: zz
Execution time: 40 ms
Execution time: 42 ms
Execution time: 2740 ms
Execution time: 437 ms```

The right solution is to cache the lowercased string, but I think this is not necessary for the initial PR. As a result the performance overhead of case conversion will only be paid on the first symbol of the first search on the page. Subsequent searches would be as fast as indexOf.

**Comment by samhocevar** [comment]
For all it’s worth, `toLowerCase()` offers no guarantee that the string length is preserved, so one cannot expect `indexOf()` to return a valid index to the original string. A simple example:

    > "İİİİİİİİİİ hello".indexOf("hello")
    11
    > "İİİİİİİİİİ hello".toLowerCase().indexOf("hello")
    21

**Comment by p12tic** [comment]
> For what it’s worth, toLowerCase() offers no guarantee that the string length is preserved, so one cannot expect indexOf() to return a valid index to the original string.

That means we can't use toLowerCase() unfortunately.

Interestingly, the following way to retrieve matches is around 2x faster than matchall. 

```
r.lastIndex = 0;
while ((result = r.exec(text)) !== null) {
    result.index;
}
r.lastIndex = 0;
```

The API is also more similar to how indexOf works.

Maybe we can replace text.indexOf with some kind of wrapper object which does this in the regex case:

```
nextMatch(text: string, pos: number): number {
    this.regex.lastIndex = pos;
    return this.regex.exec(text);
}
```

and this in indexOf case:

```
nextMatch(text: string, pos: number): number {
    return text.indexOf(this.search, pos);
}
```

This would allow both implementations and would still be maintainable.

**Comment by tdesveaux** [comment]
Interesting that `exec` is that much faster than `matchAll`.
I used `exec` initially but replaced it with `matchAll` to use a simpler `for ( of )`. Although I ended up with the weird `if (pos === undefined)` because the last iteration if undefined for some reason...

I'll take a stab at implementing the wrapper you suggested later in the week.

**Comment by tdesveaux** [comment]
@p12tic Implemented as you suggested to keep indexOf for classic search.

First and last commit stayed the same, I squashed the two implementing stuff in LogChunkSearch here https://github.com/buildbot/buildbot/pull/7249/commits/de688b624e296ad0aa64ec8d24f307148d481bd8

File: www/react-base/src/components/LogSearchField/LogSearchField.tsx [hunk 2]
```diff
@@ -17,28 +17,47 @@
...
+  const onCaseInsensitiveToggled = () => {
+    const newValue = !isCaseSensitive;
+    setIsCaseSensitive(newValue);
+    onSearchInputChanged(searchText, newValue);
+  }
+
+  const renderOptions = () => (
+    <>
+      <button
+        className={`bb-log-search-field-options bb-log-search-field-options-case bb-log-search-field-options-${isCaseSensitive ? "toggled" : "untoggled"}`}
```
p12tic, line 54: Too long line, let's keep length no more than 100 chars.

p12tic (approved):
Looks great except for a single nit.

tdesveaux:
Should I add a newsfragment as well?

p12tic:
Yes, a newsfragment is needed. I will add it after rebase.

PR #7251, opened 2023-12-05
www: Remove AngularJS web frontend

p12tic:


p12tic:
I don't have time to look into failing tests for now.

PR #7266, opened 2023-12-11
Add a Workers tab to BuilderView

### LINKED ISSUE #7162: React UI: Add builder's workers to BuilderView
When there is a large number of workers and builders and all worker are not assigned to all builders, it's useful to be able to see which workers are assigned to a given builder.
The information is already there on the BuildersView, but it's annoying to find one builder when the list is large.

I managed to add it roughly in the react-ui with this code: https://github.com/tdesveaux/buildbot/commit/ec6d2967fd4505315e536616e65b4e756bee6b15
image

But, IMHO, it would clutter the view a bit for an information that is not useful to everyone.

One solution that came to mind would be to add the information in a tab beside the 'Build requests' tab.
However, the tab container is part of the `BuildRequestsTable` component.

Would it be fine to extract the tabs part from `BuildRequestsTable` and move it to `BuilderView`?
This would result in a tab with Build requests and Builds information as it is now (with renaming of the tab title) and another tab with the workers list.

I'm prepared to submit a PR with proposed changes or to discuss another solution to this.

tdesveaux:
closes #7162

### DIFF
--- a/newsfragments/react-buildview-workers.feature
+++ b/newsfragments/react-buildview-workers.feature
@@ -1,0 +1,1 @@
+Add a "Workers" tab to the Builder view listing workers that are assigned to this builder (:issue:`7162`)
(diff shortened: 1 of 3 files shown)


p12tic:
Thanks! Please add a newsfragment.

PR #7267, opened 2023-12-12
migrate from python-future to six

### LINKED ISSUE #6980: buildbot worker and Python 2 (dependency on future)

eli-schwartz:
The python-future library makes some odd decisions and also does not work with modern python, ironically.

The six portability library does a much more minimal and elegant handling of all the actually-needed functionality in a significantly smaller codebase.

It is also much more popular and well-supported by tools. For example, `pyupgrade --py3-only` can automatically upgrade outdated six.moves imports and switch six.* functionality to the canonical python3 version.

Fixes #6980


a-detiste:
`text_type` = `str` you know .... just read six.py.

It's just replacing one poison by another and it's not much more complicated to do it right from the start.

https://wiki.debian.org/Python3-six-removal

Also all the `from __future__ import print_function` must go.

eli-schwartz:
> `text_type` = `str` you know .... just read six.py.
> 
> It's just replacing one poison by another and it's not much more complicated to do it right from the start.

I've read six.py a lot of times. It is not replacing one poison by another. The purpose of six.py is to provide a single-sourced, clean implementation of functionality that does not require importing sys, then checking sys.version_info, then repeating the same type-specific aliases in every source file.

As a bonus, 

> It is also much more popular and well-supported by tools. For example, `pyupgrade --py3-only` can automatically upgrade outdated six.moves imports and switch six.* functionality to the canonical python3 version.

I don't know what the purpose of your comment was. Are you trolling? Was some part of my original explanation faulty?

> https://wiki.debian.org/Python3-six-removal
> 
> Also all the `from __future__ import print_function` must go.

What you fail to mention is that this is your own personal wiki page, and most of the reports are about things that aren't *supposed* to be using python3-six, because upstream doesn't use it, because upstream has... dropped python2 support.

You cannot remove `__future__` statements from projects that want to still support python2.

You cannot speak for Debian policy either, unless you have the backing of the Debian project consensus.

Trust me, I've been nuking `six` dependencies right left and center (not in debian -- another distro). When the project doesn't use six anymore. If the project *does* use six, then I do not nuke something that is needed.

I would love if buildbot drops python2 support and this problem goes away via cleanups. But they won't do so until after the branchpoint. I would like to be able to successfully run on python3.12, however, and leaving the other PR open for a year means no python 3.12 support...

a-detiste:
Ok I first understood that Py2 support was dropped.

Yes python-future is in a much worse shape.

a-detiste:
You are right

eli-schwartz:
:tada: thank you!

PR #7275, opened 2023-12-17
worker: Add support for customizing affinity and nodeSelector fields of k8s pod spec

p12tic:


p12tic:
This has been opened twice in https://github.com/buildbot/buildbot/pull/7281.

PR #7286, opened 2023-12-19
e2e: Increase test timeout

p12tic:
This would hopefully fix the tests that were failing recently.


p12tic:
Seems the issue is somewhere else.

p12tic:
Actually it does seem that some tests take longer than 30 seconds sometimes, at least according to the test output.

PR #7311, opened 2023-12-22
POC: Check if there is bug in towncrier==21.9.0 (DO NOT MERGE)

pmisik:
Intent of this PR/commit is to check that there is unspotted bug in in towncrier==21.9.0 happens on buildbot infrastructure as I see it on my own machine

Details:
As docs-towncrier target is called twice
from docs-release
and from docs-release-spelling
it generates release notes for PR twice.
This was fixed in towncrier 22.8.0.rc1 (2022-08-28) and is reported as error

Release notes:
"The detection of duplicate release notes was fixed and recording changes of same version is no longer triggered"
https://towncrier.readthedocs.io/en/stable/release-notes.html#id81

## Contributor Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,9 @@ docs-towncrier:
 	if command -v towncrier >/dev/null 2>&1 ;\
 	then \
 	towncrier --draft | grep  'No significant changes.' || yes n | towncrier ;\
+	echo "---- Diff start ----"; \
+	git --no-pager diff --cached; \
+	echo "---- Diff end ----"; \
 	fi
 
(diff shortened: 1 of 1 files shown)


pmisik:
Theory is confirmed. 
Step `docs` (docs-release): generates https://buildbot.buildbot.net/#/builders/101/builds/2812/steps/9/logs/stdio
Step `spelling` (docs-release-spelling) duplicates it https://buildbot.buildbot.net/#/builders/101/builds/2812/steps/10/logs/stdio

PR #7325, opened 2023-12-25
Integrate ruff format

Mic92:
## Contributor Checklist:


.bbtravis.yml
```diff
@@ -181,6 +181,12 @@ script:
       export PATH=/tmp/bbvenv/bin/:$PATH
       isort --check -df `git ls-files |grep '.py$'`
 
+  - title: ruff-format
+    condition: TESTS == "ruff"
```
Mic92, line 185: Don't know what this condition does. I just copied it from the other examples.

.bbtravis.yml
```diff
@@ -181,6 +181,12 @@ script:
       export PATH=/tmp/bbvenv/bin/:$PATH
       isort --check -df `git ls-files |grep '.py$'`
```
Mic92, line 182: We can also replace pylint/isort/flake8 with ruff as it integrates the same rulesets.
Than we no longer have competing linters.

Mic92:
Since there is still a conflict between ruff and isort, I will replace isort with ruff's own sorting in the next PR.

.bbtravis.yml
Mic92: started in https://github.com/buildbot/buildbot/pull/7326

pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[tool.ruff]
+line-length = 88
+target-version = "py38"
```
Mic92, line 3: What's the lowest version still supported here?

Mic92:
Since I am getting a lot of these "line-too-long" errors now, would it be acceptable to disable this lint and just let `ruff` make sure that long lines are broken down? Most of these warnings come from strings.

p12tic:
I would like to use `--skip-string-normalization` black option because most formatting changes are for quotes (which don't really matter when considering the benefits of using black/ruff. It doesn't seem ruff has this option, does it?

Mic92:
Looks like it was added here: https://github.com/astral-sh/ruff/pull/8822

Mic92:
ruff > [v0.1.8](https://github.com/astral-sh/ruff/releases/tag/v0.1.8)

Mic92:
Just let me know your preference and I can update this pull request.

pyproject.toml
p12tic: 3.8 is correct. For worker minimum is 3.4.

p12tic: line-length should be 100.

p12tic:
> Just let me know your preference and I can update this pull request.

Let's enable `--skip-string-normalization` and convert. I don't have anything large planned that could need a backport to 3.x, so there will be little inconvenience.

pyproject.toml
Mic92: Done.

Mic92:
The big ruff format is now down automatically. You can reproduce by running:

```
ruff format .
ruff lint --fix --select "ISC" .
isort .
git commit -am 're-format with ruff'
```

Mic92:
Finally all linters seems to be happy now!

p12tic:
@Mic92 FYI in the end import order didn't need to be fixed. This was probably caused by some confusion over what's created by ruff reformat and what are later manual changes.

PR #7326, opened 2023-12-26
Replace isort with ruff

Mic92:
In order to use ruff for formatting we want consistent behavior between `ruff format` and import sorting (implemented as in the ruff linter). Also ruff should be faster in the end than isort.

Command run to do the format automatically:

```
make ruff
git commit -am 'ruff: apply isort'
```

## Contributor Checklist:


Mic92:
Looks like the CI run out of disk space?

p12tic:
Rebased and resolved conflicts

p12tic:
I had to add `known-first-party = ["buildbot", "buildbot_worker"]` so that ruff is less confused by import order outside master/buildbot and worker/buildbot_worker folders.

Mic92:
Thanks for pushing this through. Will look at the next targets for ruff.

PR #7344, opened 2023-12-30
maketarballs: removing silently failing command related to removing AngularJS

pmisik:
"/bin/sh: 1: cd: can't cd to www/build_common"

Looks like it was not removed as part in PR #7287

## Contributor Checklist:


p12tic:
Thanks!

PR #7345, opened 2023-12-30
makefile: Enable more strict mode for shell

pmisik:
- The -e option makes the shell exit immediately if any command it runs exits with a non-zero status, effectively causing the Makefile to fail if any command fails.
- The -u option treats unset variables as an error. If any variable used in a command is not set, the shell will exit with an error.

## Contributor Checklist:


p12tic:
Thanks!

PR #7389, opened 2024-01-17
Fix build summary requests UI when a line break occurs in the step's build summary.

zenoxs:
This PR fix the following issue:
The build summary requests UI is broken when a line break occurs in the step's build summary.

_Screenshot of the issue_
ui broken on line break

_Screenshot with the fix_
issue-build-bot-fixed

## Contributor Checklist:

* [not_needed ] I have updated the unit tests
* [not_needed ] I have updated the appropriate documentation

### DIFF
--- a/www/react-base/src/components/BuildSummary/BuildSummary.scss
+++ b/www/react-base/src/components/BuildSummary/BuildSummary.scss
@@ -38,6 +38,10 @@
 .bb-build-summary-step-line {
   position:relative;
 
+  .bb-build-summary-step-label {
+    display: inline-block;
+  }
+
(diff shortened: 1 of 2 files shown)


p12tic:
Actually, this doesn't work properly because the time and status fields moved to the left. I didn't notice this in the screenshots. I will revert the PR for the time being.

zenoxs:
> Actually, this doesn't work properly because the time and status fields moved to the left. I didn't notice this in the screenshots. I will revert the PR for the time being.

My bad, I didn’t noticed the side effect during my tests. I will open a new PR with the correct behaviour, sorry for that 🙏

Issue #7385, opened 2024-01-15
Improve code completion


cathaysia:
Python currently supports using tools such as typing to prompt variable types. But buildbot currently does not support this part. The experience with complete code is still not very good.
On the other hand, modules exported from buildbot.plugin also have no type information.

To improve the code completion experience, on the one hand, type information needs to be added to existing code, and on the other hand, the modules in the plugin need to be exported using reexport instead of loading them using functions.

Do you have any plans to change this, and if I submit a PR, will this kind PR be accepted?

p12tic:
I'm closing this issue as https://github.com/buildbot/buildbot/pull/7390 has been closed.

Improvements to typing support are welcome in Buildbot.

PR #7390, opened 2024-01-18
style(types): add some types hint

cathaysia:
This is a craft If this type of pr is acceptable, I will populate the content of this pr

close: #7385

### DIFF
--- a/master/buildbot/steps/__init__.py
+++ b/master/buildbot/steps/__init__.py
@@ -1,0 +1,5 @@
+from .cmake import CMake
+from .cppcheck import Cppcheck
+from .source.git import Git, GitCommit, GitPush, GitStepMixin, GitTag
+from .source.github import GitHub
+from .source.gitlab import GitLab
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -70,11 +70,11 @@ class Git(Source, GitStepMixin):
 
     def __init__(
         self,
-        repourl=None,
+        repourl: str,
         branch='HEAD',
         mode='incremental',
         method=None,
-        reference=None,
+        reference: str | None = None,
         submodules=False,
         remoteSubmodules=False,
         tags=False,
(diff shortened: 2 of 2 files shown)


master/buildbot/steps/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .cmake import CMake
```
p12tic, line 1: What is the reason why this is required? I can't think of any and this will make the steps accessible from multiple places.

p12tic (commented):
I think adding type hints is great!

master/buildbot/steps/__init__.py
cathaysia: This reexport is to facilitate importing types from buildbot. This avoids lengthy imports like from buildbot.steps.cmake import CMake. 

I like :

```python
from zorg.buildbot.builders import ClangBuilder
from zorg.buildbot.builders import FlangBuilder
from zorg.buildbot.builders import PollyBuilder
```

but not

```python
from buildbot.reporters.generators.build import BuildStartEndStatusGenerator
from buildbot.reporters.generators.buildrequest import BuildRequestGenerator
from buildbot.reporters.generators.buildset import BuildSetStatusGenerator
from buildbot.reporters.generators.worker import WorkerMissingGenerator
from buildbot.reporters.message import MessageFormatterRenderable
```

The first way seems more intuitive and easier to use.

p12tic: (sorry for late reply - took some time to double check my comment)

cathaysia: Unless buildbot.plugins has type information, I'm afraid no one will like to import plugins from here. If you don't like reexport, then we may need to find a way to add type information to buildbot.plugins

p12tic: Thanks for pointing this out. Indeed, buildbot.plugins will need improvement in this regard.

cathaysia:
Okay, I'll see how to make it better later.
________________________________

cathaysia:
hi, I close this pr due to I don't have more time support this. sorry. :(

PR #7394, opened 2024-01-21
Clarify migration documentation

pmisik:
I am not sure if these changes are correct.

Based on  [2.12.1. Upgrading to Buildbot 4.0 (not released)](https://docs.buildbot.net/latest/manual/upgrading/4.0-upgrade.html#upgrading-to-buildbot-4-0-not-released) , I assume that React based plugins will co-exist with AngularJS, so you don't plan to rename React based plugins to their original names which use AngularJS based plugins.
So, the Buildbot user needs to change the master configuration when upgrading to version 4.

## Contributor Checklist:

### DIFF
--- a/master/docs/manual/upgrading/4.0-upgrade.rst
+++ b/master/docs/manual/upgrading/4.0-upgrade.rst
@@ -22,7 +22,7 @@ The recommended upgrade procedure is as follows:
 Web frontend
 ------------
 
-Buildbot 4.0 removes the legacy web frontend written in AngularJS.
+Buildbot 4.0 replaces the AngularJS-based web frontend with a new React-based one.
 In simple Buildbot installations there is nothing that needs to be done except to install compatible versions of any www plugins that are used.
 The following plugins are maintained as part of Buildbot and can be upgraded seamlessly by just installing new, compatible version:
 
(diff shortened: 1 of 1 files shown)


p12tic:
Good catch.

The documentation is incorrect/unclear. There's no AngularJS in 4.x. The intention is that `console_view` plugin means AngularJS plugin in 3.x and React plugin in 4.x. For easier migration, a duplicate web React web frontend is provided in 3.x in `buildbot-react-console-view` and similar packages. In 4.x these are no longer needed because the regular `buildbot-console-view` package will contain React-based plugin too. However for 4.0.x the `buildbot-react-*` packages will still be built and published so that users don't get a broken configuration after upgrading to 4.0. There will be deprecation warnings shown so that users go back to the canonical `buildbot-console-view` and similar packages.

pmisik:
>However for 4.0.x the buildbot-react-* packages will still be built and published so that users don't get a broken configuration after upgrading to 4.0

Isn't this an unnecessary complication?
Does that mean you want to create another 4 directories under www/ with setup.py (which will contain a different package name, packages, entry_point)?

p12tic:
> Isn't this an unnecessary complication?
Does that mean you want to create another 4 directories under www/ with setup.py (which will contain a different package name, packages, entry_point)?

It is temporary. It will live just for several 4.0.x releases, so I don't plan to think of anything time consuming. Copying the plugin directories and changing plugin names I think is reasonable solution in this case. Git does not store identical files, so there's no downside.

The problem what I want to solve is that users should have a way to get identical Buildbot configuration working on both e.g.  3.11.0 and 4.0.0 without any additional changes. This makes upgrade process smoother as at no point the user's configuration is broken by something that's outside user's direct control.

p12tic (approved):
Thanks!

PR #7395, opened 2024-01-22
www: Add WSGI dashboards plugin for React

lephilousophe:
This PR migrates the old WSGI dashboards plugin from Angular to React while maintaining Python compatibility.
As Bootstrap version changed, the CSS of served webpages may need some updates.

There wasn't any tests in the old plugin and the existing documentation can be reused.

The configuration was needed in the setup step of the plugin to enumerate all dashboards to add in the menu so I added the `config` object to the `onBuildbotSetupPlugin` function.

## Contributor Checklist:


p12tic (approved):
This is great, thank you!

PR #7430, opened 2024-02-09
reporters: BitbucketServerCoreAPIStatusPush: Support epoch time format

vibbo:
Events produced by WAMP message queue implementation use epoch instead of `datetime.datetime` format. Calculating duration is now compatible with both formats for values of `started_at` and `complete_at`.

## Contributor Checklist:


p12tic (approved):

p12tic:
builds data in the reports should contain only time since epoch. I think this fix improves the situation, but long term we will solve the root cause.

PR #7432, opened 2024-02-12
buildbot.worker.openstack: Use find_image() instead of list_images()

cmouse:
If there are too many images, list_images() requires paging. Since we are looking for specific name in any case, use find_image() instead.

## Contributor Checklist:


p12tic:
Rebased, fixed a test and adjusted newsfragment because this is not just a change, but a fix for performance problem.

PR #7436, opened 2024-02-15
Replace flake8/pep8 with ruff

Mic92:
## Remove this paragraph

And especially:

## Contributor Checklist:


pyproject.toml
```diff
@@ -1,7 +1,15 @@
 [tool.ruff]
 line-length = 100
 target-version = "py38"
-select = ["I"]
+select = ["W", "E", "F", "I"]
+ignore = [
+   "E711", # comparison to None should be 'if cond is None:'
```
Mic92, line 6: Turns out a lot of former disabled checks are not in ruff or do not need to be ignored.

common/merge_and_pep8.sh
```diff
@@ -1,89 +0,0 @@
...
-git merge $PREPEP8
-if [[ $? != 0 ]]
-then
-    echo "Please fix the merge conflicts between your branch, and last commit before autopep8!"
-    newshell
-fi
-
-status "merging against first commit after autopep8 and take our version when there are conflicts"
-git merge $POSTPEP8
-# autopep8 takes 1h30 to run on the whole codebase, so let git resolve the obvious merge conflicts.
```
Mic92, line 54: Current ruff is faster than I can type it in the terminal. So I guess this script is no longer needed.

p12tic:
Rebased.

PR #7441, opened 2024-02-17
Update alembic from 1.11.3 to 1.13.1 with fix for mysql

pmisik:
Alembic 1.12.0 changed default for type comparison.
Since 1.12.0 this is on by default. See https://alembic.sqlalchemy.org/en/latest/changelog.html#change-1.12.0
In case of mysql, there is issue with comparison MEDIUMBLOB() vs LargeBinary(length=65536) in logchunks table.
Problem is reported by trial test `buildbot.test.integration.test_upgrade.UpgradeTestEmpty.test_emptydb_modelmatches`

```
[[('modify_type',
None,
'logchunks',
'content',
{'existing_comment': None,
'existing_nullable': True,
'existing_server_default': False},
MEDIUMBLOB(),
LargeBinary(length=65536))]]
```

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/integration/test_upgrade.py
+++ b/master/buildbot/test/integration/test_upgrade.py
@@ -122,7 +122,13 @@ class UpgradeTestMixin(db.RealDatabaseMixin, TestReactorMixin):
         def comp(engine):
             # use compare_model_to_db, which gets everything but indexes
             with engine.connect() as conn:
-                diff = compare_metadata(MigrationContext.configure(conn), self.db.model.metadata)
+                opts = None
+                if engine.dialect.name == 'mysql':
+                    # Disable type comparisons for mysql. Since 1.12.0 it is enabled by default.
+                    # https://alembic.sqlalchemy.org/en/latest/changelog.html#change-1.12.0
+                    # There is issue with comparison MEDIUMBLOB() vs LargeBinary(length=65536) in logchunks table.
+                    opts = {"compare_type": False}
+                diff = compare_metadata(MigrationContext.configure(conn, opts=opts), self.db.model.metadata)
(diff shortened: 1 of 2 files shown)


master/buildbot/test/integration/test_upgrade.py
```diff
@@ -122,7 +122,13 @@ def assertModelMatches(self):
         def comp(engine):
             # use compare_model_to_db, which gets everything but indexes
             with engine.connect() as conn:
-                diff = compare_metadata(MigrationContext.configure(conn), self.db.model.metadata)
+                opts = None
+                if engine.dialect.name == 'mysql':
+                    # Disable type comparisons for mysql. Since 1.12.0 it is enabled by default.
+                    # https://alembic.sqlalchemy.org/en/latest/changelog.html#change-1.12.0
+                    # There is issue with comparison MEDIUMBLOB() vs LargeBinary(length=65536) in logchunks table.
+                    opts = {"compare_type": False}
```
pmisik, line 130: This fixes just test but is it correct for real migration?

p12tic: It's enough if the test is fixed for now. Do you still have logs from the error in test? Could you open a bug with these?

pmisik: >  Do you still have logs from the error in test? 

It is still visible in  PR #7447
See logs
- https://buildbot.buildbot.net/#/builders/89/builds/8025
- https://buildbot.buildbot.net/#/builders/115/builds/2474

p12tic: Thanks!

Issue #7384, opened 2024-01-12
Git: Allow to auth with password/token


tdesveaux:
At the moment, Buildbot allow to provide a ssh key to handle Git authentication.
GitHub and other hosts allow to authenticate with access tokens (and even recommend it).

From a rather quick search, I only found this related issue #3137

I started implementing the feature to authenticate with said token in Buildbot on this [branch](https://github.com/buildbot/buildbot/compare/master..tdesveaux:buildbot:git/http-credentials?expand=1).
This add the `gitCredentials` and `gitCredentialStoreContent` (I'll try to find better names...) attributes to `GitStepMixin` to enable authentication with said tokens by using the [git credential](https://git-scm.com/docs/git-credential) git feature with the [`store`](https://git-scm.com/docs/git-credential-store) which store credentials in a file allowing to implement in a similar way to the ssh auth.

There is some work left to do in this branch as it's more in a POC state than anything (no tests/doc, only tested `source.Git` step, make it work for `GitPoller`, ...).

Is this a feature of interest to the Buildbot project? And is this approach acceptable?

I also saw that there might be some opportunities to refactor some related code that I could tackle while at it.
Two that come to mind:
- Maybe extract the ssh key code from `GitStepMixin` into a `GitStepWithAuthMixin` to remove the dummy declarations in `GitTag` and `GitCommit` steps.
- Maybe find a way to factorize the code a bit between `GitStepMixin` and `GitPoller`? There seems to be a bit of duplication in the ssh key handling.

Let me know if you think of other things I can look at.

p12tic:
@tdesveaux Thanks for working on this feature. It makes sense to have it in Buildbot.

I've looked into the WIP code and I didn't see any problems in the logic part. 

Just before doing a PR please replace all camelCase in code additions with snake_case - Buildbot is gradually transitioning to standard Python snake_case style and all new code should use that.

tdesveaux:
@p12tic  apologize about the ping, but I feel I need to confirm an architectural / code style choice with you.

I started to look into this by refactoring the code a bit (first by trying to move SSH stuff to `GitMixin` so that it can be used by both `GitPoller` and `GitSteps`).

However, I'm feeling more and more that it would be simpler to use composition instead of the Mixin pattern here.
I'd like your input as Mixin seems to be used accross the codebase, and might still be prefered in this case (my issues might be more about my lack of experience with Mixin and multiple inheritance in Python).

I see the following drawbacks to the Mixin inheritance used now:

[...]

Issue #7371, opened 2024-01-05
Consider to convert inlineCallbacks to async/await


pmisik:
Twisted documentation for [Inline callbacks - using ‘yield’](https://docs.twisted.org/en/stable/core/howto/defer-intro.html#inline-callbacks-using-yield) states

>Unless your code supports Python 2 (and therefore needs compatibility with older versions of Twisted), writing coroutines with the functionality described in “Coroutines with async/await” is preferred over inlineCallbacks. Coroutines are supported by dedicated Python syntax, are compatible with asyncio, and provide higher performance.

As Buildbot dropped support for Python2 there is opportunity to convert inlineCallbacks to async/await.

I’m not sure if it helps.

Some thoughts on why and how to do this are given in https://github.com/matrix-org/synapse/issues/7988#issuecomment-858836748

This is related blogpost https://patrick.cloke.us/posts/2021/06/11/converting-twisteds-inlinecallbacks-to-async/

p12tic:
Buildbot will transition to async/await at some point. One problem is that moving from inlineCallbacks to async/await changes return type of a function from defer.Deferred to coroutine. Users may be relying on additional functionality exposed by defer.Deferred() class. Thus the migration will be rather complex.

PR #7446, opened 2024-02-19
Refactor Git SSH auth handling

tdesveaux:
Related to #7384 
This allow more code re-use between GitPoller and GitSteps.

Took more time than anticipated and there might still be areas of improvement but I feel it's in a state good enough to merge.
Tried to make separated commits as much as possible.

I added typing of inlineCallbacks return types in a separate commit as they may be a bit confusing. 

## Contributor Checklist:

* [N/A?] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
    There is no public API changes


master/buildbot/changes/gitpoller.py
```diff
@@ -52,16 +57,14 @@ class GitPoller(base.ReconfigurablePollingChangeSource, StateMixin, GitMixin):
         "project",
         "pollAtLaunch",
         "buildPushesWithNoCommits",
-        "sshPrivateKey",
-        "sshHostKey",
-        "sshKnownHosts",
         "pollRandomDelayMin",
         "pollRandomDelayMax",
+        "_git_ssh_auth",
```
tdesveaux, line 62: Not sure if this plus the fact `AbstractGitSSHAuth` inherit from `ComparableMixin` will do what I expect.
Not sure how to test this, would welcome any pointers.

master/buildbot/changes/gitpoller.py
```diff
@@ -481,37 +484,86 @@ def _process_changes(self, newRev, branch):
...
+
     def _isSshPrivateKeyNeededForCommand(self, command):
-        commandsThatNeedKey = [
-            'fetch',
-            'ls-remote',
-        ]
-        if self.sshPrivateKey is not None and command in commandsThatNeedKey:
-            return True
-        return False
+        ssh_auth_warndeprecate(
```
tdesveaux, line 515: Might have gone a bit overboard with deprecation messages...
Maybe we could remove at least private method without warnings?

master/buildbot/changes/gitpoller.py
```diff
@@ -481,37 +484,86 @@ def _process_changes(self, newRev, branch):
                 src='git',
             )
 
+    @property
+    def sshPrivateKey(self):
+        ssh_auth_warndeprecate(
+            "self.sshPrivateKey",
+            "self._git_ssh_auth.ssh_private_key",
+        )
+        assert self._git_ssh_auth is not None
```
tdesveaux, line 493: I used a lot of asserts to placate type linter mostly in deprecated code

master/buildbot/changes/gitpoller.py
```diff
@@ -481,37 +484,86 @@ def _process_changes(self, newRev, branch):
...
+        return self._git_ssh_auth._get_ssh_host_key_path(ssh_data_path)
 
     @defer.inlineCallbacks
     def _dovccmd(self, command, args, path=None):
-        if self._isSshPrivateKeyNeededForCommand(command):
+        if (
+            self._git_ssh_auth is not None and
+            self._git_ssh_auth.is_ssh_private_key_needed_for_git_command(command)
```
tdesveaux, line 565: There is some checks as this one that are not great in themselves.
I intend to hide them in a `git_auth_handler` which will allow to add auth by user:password/token mentioned in #7384 in a fairly transparent manner later on.

master/buildbot/util/git.py
```diff
@@ -270,53 +348,356 @@ def checkFeatureSupport(self):
...
+    @defer.inlineCallbacks
+    def remove_ssh_private_key_if_needed(
+        self, workdir: str,
+    ) -> Generator[defer.Deferred[Any], Any, int]:
+        if not self.did_download_ssh_private_key:
+            return RC_SUCCESS
+
+        Path(
+            self._get_ssh_private_key_path(workdir)
+        ).unlink(missing_ok=True)
```
tdesveaux, line 697: Mostly used pathlib here for this `missing_ok`, can replace with the `os.unlink` in a try/except if preferred.

master/buildbot/changes/gitpoller.py
tdesveaux: I kind of want to create mixin classes that implement these deprecation so the "real" class implementation is not "polluted".

Something like:
```python
class _DeprecatedGitStepMixin:
    @property
    def sshPrivateKey(self):
        ssh_auth_warndeprecate(
            "self.sshPrivateKey",
            "self._git_auth.ssh_private_key",
        )
        return self._git_auth.ssh_private_key

...

class GitStepMixin(GitMixin, _DeprecatedGitStepMixin):
```
with `_DeprecatedGitStepMixin` moved at the end of the file or even another file altogether.

p12tic: I think this makes code actually harder to understand. Just move these to the end of the file.

In any case, the properties are not documented as an API, so you can simply remove these.

master/buildbot/util/git.py
```diff
@@ -137,66 +159,112 @@ def ensureSshKeyNewline(privateKey: str) -> str:
...
-        # we can't use the workdir for temporary ssh-related files, because
-        # it's needed when cloning repositories and git does not like the
-        # destination directory being non-empty. We have to use separate
-        # temporary directory for that data to ensure the confidentiality of it.
-        # So instead of
-        # '{path}/{to}/{workerbuilddir}/{workdir}/.buildbot-ssh-key'
-        # we put the key in
-        # '{path}/{to}/.{workerbuilddir}.{workdir}.buildbot/ssh-key'.
+    @property
+    def sshHostKey(self):
```
p12tic, line 195: No longer needed, remove. Same with other similar properties.

master/buildbot/util/git.py
```diff
@@ -137,66 +159,112 @@ def ensureSshKeyNewline(privateKey: str) -> str:
...
-        workdir = self._getSshDataWorkDir().rstrip('/\\')
+    @abc.abstractmethod
+    def _get_auth_data_workdir(self) -> str:
+        raise NotImplementedError()
 
-        if path_module.isabs(workdir):
-            parent_path = path_module.dirname(workdir)
-        else:
-            parent_path = path_module.join(self.worker.worker_basedir, path_module.dirname(workdir))
+    def _getSshDataWorkDir(self) -> str:
```
p12tic, line 214: We're moving code around, so prefer changing the function names to `snake_case`.

p12tic: Ah, actually this also can be removed.

master/buildbot/util/git.py
```diff
@@ -270,53 +345,344 @@ def checkFeatureSupport(self):
...
+            'ls-remote',
+            'push',
+            'submodule',
+        ]
+        if git_command in git_commands_that_need_auth:
+            return True
+        return False
+
+    @property
+    @abc.abstractmethod
```
p12tic, line 413: How are these abc.abstractmethod decorators useful? What hole do I have in my python knowledge? I always just override stuff in subclasses and that's it.

master/buildbot/util/git.py
```diff
@@ -270,53 +345,344 @@ def checkFeatureSupport(self):
...
 
-            if self.sshHostKey is not None:
+            if known_hosts is not None:
+                known_hosts_contents = known_hosts
+            else:
                 known_hosts_contents = getSshKnownHostsContents(host_key)
-            yield self.downloadFileContentToWorker(
-                known_hosts_path, known_hosts_contents, workdir=workdir, mode=0o400
+
+            yield self._download_file(
```
p12tic, line 499: Maybe keep name _download_file_content_to_worker?

master/buildbot/util/git.py
```diff
@@ -270,53 +345,344 @@ def checkFeatureSupport(self):
...
+            return RC_SUCCESS
+
+        Path(
+            self._get_ssh_private_key_path(workdir)
+        ).unlink(missing_ok=True)
+        Path(
+            self._get_ssh_host_key_path(workdir)
+        ).unlink(missing_ok=True)
+
+        yield defer.succeed(None)
```
p12tic, line 687: This doesn't do anything.

p12tic:
@tdesveaux Thanks a lot for this work. I added several comments and will have another look once these are addressed. Overall the PR looks good.

master/buildbot/util/git.py
tdesveaux: Wouldn't that be confusing in the context of `GitServiceAuth` which has no worker?

master/buildbot/util/git.py
tdesveaux: Linter (pylint here) warn if a subclass does not override a method marked as abstract. I can remove it if preferred.

image

master/buildbot/changes/gitpoller.py
tdesveaux: Since all that I marked with `ssh_auth_warndeprecate` are not documented, can I take this as a go on removing all of them?

master/buildbot/util/git.py
tdesveaux: I should have added a comment.
This is awkward stuff needed if we want to type the function in combination to of using `inlineCallback` (one more case were porting to async as mentioned in #7371 would be helpful).

Since `GitStepAuth` implementation need to return a `Deferred` and `inlineCallback` make using multiple far more readable, we need to use it in `AbstractGitAuth`.
Then, as `AbstractGitAuth` is tagged as `inlineCallback`, the function's return type need to be `Generator[...]`.
Which make `GitServiceAuth` need to return a `Generator` as well as be an `inlineCallback`, thus the weird useless yield...

As I said, I typing of `inlineCallback` function's return in a separate commit that can be dropped as it probably bring more confusion that help in anything.

master/buildbot/changes/gitpoller.py
p12tic: Yes.

master/buildbot/util/git.py
p12tic: Maybe it makes sense to not use typing until we migrate to async/await? Right now typing adds a lot of noise to inlineCallbacks function declarations and also requires cumbersome code.

master/buildbot/util/git.py
p12tic: Indeed, I'm wrong.

master/buildbot/util/git.py
p12tic: Many pylint warnings are excessive. E.g. see https://github.com/buildbot/buildbot/blob/master/common/pylintrc for the list of pylint warnings that are disabled.

p12tic: In my experience I haven't yet seen a bug that is prevented by `@abc.abstractmethod`. I suggest we don't use it, because it's inconsistent across Buildbot codebase anyway.

master/buildbot/changes/gitpoller.py
p12tic: ComparableMixin compares properties recursively so it does the right thing here.

tdesveaux:
@p12tic I should have addressed everything.

p12tic (approved):

p12tic:
Thanks!

PR #7458, opened 2024-02-28
Improvements to the React GUI for builders and workers

YngveNPettersen:
This set of patches have the following improvements that I would like to have included in a future 3.11+ release.

We have been using patches similar to these for several years in the Angular JS version.

The GUI patches are:

* Project configuration of number of recent builds to display in the Home App
* Show the revision property for a build in the builders and workers pages
* Improve display of build completion time 
* Let the project specify properties to use to display e.g. a build 1.2.3.4 version number instead of just build number 999 on that builder.
* Let the Tags column on the Builders page be less wide when there are no tags

Additionally:

* Fix the Makefile to work on Windows

## Contributor Checklist:

* I have updated the unit tests (no tests affected AFAICT)
* I have updated the appropriate documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ endif
 
 VENV_NAME := .venv$(VENV_PY_VERSION)
 PIP ?= $(ROOT_DIR)/$(VENV_NAME)/$(VENV_BIN_DIR)/pip
-PYTHON ?= $(ROOT_DIR)/$(VENV_NAME)/$(VENV_BIN_DIR)/python
+VENV_PYTHON ?= $(ROOT_DIR)/$(VENV_NAME)/$(VENV_BIN_DIR)/python
 YARN := $(shell which yarnpkg || which yarn)
 
 
@@ -93,7 +93,7 @@ frontend: frontend_deps
 # build frontend wheels for installation elsewhere
 frontend_wheels: frontend_deps
 	for i in pkg $(WWW_PKGS); \
-		do (cd $$i; $(PYTHON) setup.py bdist_wheel || exit 1) || exit 1; done
+		do (cd $$i; $(VENV_PYTHON) setup.py bdist_wheel || exit 1) || exit 1; done
 
 # do installation tests. Test front-end can build and install for all install methods
 frontend_install_tests: frontend_deps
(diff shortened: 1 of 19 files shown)


www/react-base/src/components/BuildsTable/BuildsTable.tsx
```diff
@@ -59,16 +59,26 @@ export const BuildsTable = observer(({builds, builders}: BuildsTableProps) => {
       )
       : <></>;
 
+    const buildStartedCompletedElement = build.complete
+      ? (
+          <span title={dateFormat(build.complete_at!)}>
+            {durationFormat(build.complete_at! - build.started_at)}
```
YngveNPettersen, line 65: Incorrect function; result of copy/paste/forget-to-modify

www/react-base/src/components/BuildsTable/BuildsTable.tsx
```diff
@@ -59,16 +59,26 @@ export const BuildsTable = observer(({builds, builders}: BuildsTableProps) => {
...
+          </span>
+      )
+      : (
+          <span title={dateFormat(build.started_at)}>
+            {durationFromNowFormat(build.started_at, now)}
```
YngveNPettersen, line 70: Incorrect function; result of copy/paste/forget-to-modify

Makefile
```diff
@@ -4,13 +4,22 @@ ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 .PHONY: docs pylint flake8 virtualenv
 
+ifeq ($(OS),Windows_NT)
+  VENV_BIN_DIR := Scripts
+  VENV_PY_VERSION ?= python
+  VENV_CREATE := python -m venv
```
Mic92, line 10: I think we can just use venv on all platforms now. Does windows not call it python3 as well? I thought this would be consistent across platforms.

YngveNPettersen: No, The windows installer are not calling the executable python3.exe, only python.exe; It is a bit irritating, but I have to copy and rename the exe each time I do an installation.

Mic92: Alright, do you know how other package manager on windows handle it? i.e. does https://www.msys2.org/ also provide a python binary that is not python2?

p12tic:
Closing as all changes have been merged via separate PRs.

PR #7512, opened 2024-04-16
buildrequestdistributor: Sort pending builders only once

vibbo:
Sorting pending builder each time the list is updated is unnecessary overhead. This can be done once the list is passed on to BuildRequestDistributor._activityLoop.

---

I was debugging some issues with BuildRequestDistributor and I find this change is an improvement of the existing implementation.
The question is, am I missing something? Is there actually any reason for the current implementation where pending builders are sorted each time the list is updated?

I will create newsfragment if the change is viable.


p12tic:
I'm not confident with merging this change. I would need to review the algorithm used.

vibbo:
> I'm not confident with merging this change. I would need to review the algorithm used.

Is there anything I can do to help now?

PR #7528, opened 2024-04-23
www/react: Mark filtered tags on Builders page in bb-builders-view-container table header row as clickable

pmisik:
Since the Badge implements onClick, hand mouse cursor highlights that it is clickable. I find this feature useful in case I need to filter a builder with multiple tags on a Buildbot instance with a huge amount of builders.

image

## Contributor Checklist:

### DIFF
--- a/www/react-ui/src/util/TagFilterManager.tsx
+++ b/www/react-ui/src/util/TagFilterManager.tsx
@@ -191,7 +191,7 @@ export class TagFilterManager {
         enabledTagsElements.push((
           <>
             <Badge variant="success"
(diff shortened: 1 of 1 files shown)


pmisik:
Unfortunately I'm not able replicate build failures on my Ubuntu 22.04.4 LTS (I tried run them 3x on my machine).
As they run on `p12-pd-` workers I speculate if there is still some issue I reported in https://github.com/buildbot/buildbot-infra/issues/269
This time I do not see visible time changes in logs that could prove mine theory so it is speculation.

p12tic (approved):

PR #7529, opened 2024-04-23
remove duplicate project description component

elbehery95:
Previously the project field was being rendered twice when viewing a project.

Now it is only rendered once and `renderDescription` no longer returns the extra `<TableHeading>Description:</TableHeading>` component since it is already rendered in the card body

Before:
image

After:
image


p12tic (approved):
Thanks for the PR.

PR #7543, opened 2024-04-30
Git steps: implement auth by user/password

tdesveaux:
This implement user/password authentication for the Git and GitPush steps (first part of #7384).

Authentication is handled with the builtin git credential flow using the [git-credential-store](https://git-scm.com/docs/git-credential-store) backend.

Credentials are stored using `git credential approve`.
An alternative implementation could directly write the credential file that git-credential-store would read, but it's not supposed to be user facing ([see doc](https://git-scm.com/docs/git-credential-store#_storage_format): "Do not view or edit the file with editors") so might be subject to change in the future.
Whereas the [git credential IOFormat](https://git-scm.com/docs/git-credential#IOFMT) is defined and should be stable.

Also, it allow to implement auth through another git-credential backend in the future if needed.

Credentials can be provided to the steps as a simple user/password tuple with `auth_credentials`, in which case, it will complete the auth form with the `repourl` the step is supposed to clone.
It's also possible to directly provide the `git credential` form through `git_credentials` (`GitCredentialOptions`). This might be needed for repositories that have submodule which are private and require a different set of credentials.

`GitCredentialOptions` also has a `use_http_path` member to override the config value of the same name. This is needed in case multiple credentials are needed for the same host (eg. GIthub repositories owner by different org/user). See [doc](https://git-scm.com/docs/gitcredentials#Documentation/gitcredentials.txt-useHttpPath).

I'll implement the same mechanism in `GitPoller` in another PR once this one is accepted as it require some more code change to provide a way for `GitServiceAuth` to run commands.

PS: Thanks for the `async_to_deferred` helper!

## Contributor Checklist:

### DIFF
--- a/master/docs/manual/configuration/steps/source_git.rst
+++ b/master/docs/manual/configuration/steps/source_git.rst
@@ -152,3 +152,13 @@ The Git step takes the following arguments:
    This may be either a :ref:`Secret` or just a string.
    `sshPrivateKey` must be specified in order to use this option.
    `sshHostKey` must not be specified in order to use this option.
+
+``auth_credentials``
+
+   (optional) An username/password tuple to use when running git for fetch operations.
+   The worker's git version needs to be at least 1.7.9.
+
+``git_credentials``
+
+   (optional) See :ref:`GitCredentialOptions`.
+   The worker's git version needs to be at least 1.7.9.
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -13,6 +13,10 @@
 #
 # Copyright Buildbot Team Members
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from twisted.internet import defer
 from twisted.internet import reactor
 from twisted.python import log
@@ -25,6 +29,12 @@ from buildbot.steps.source.base import Source
 from buildbot.steps.worker import CompositeStepMixin
 from buildbot.util.git import RC_SUCCESS
 from buildbot.util.git import GitStepMixin
+from buildbot.util.git_credential import GitCredentialOptions
+from buildbot.util.git_credential import add_user_password_to_credentials
+
+if TYPE_CHECKING:
+    from buildbot.interfaces import IRenderable
+
 
 GIT_HASH_LENGTH = 40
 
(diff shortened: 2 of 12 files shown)


p12tic:
FYI `@async_to_deferred` is only needed for externally callable APIs. Internal functions can be proper `async` because we can ensure that all call sites use either `yield` or `await` on it.

master/docs/manual/configuration/steps/source_git.rst
```diff
@@ -152,3 +152,13 @@ The Git step takes the following arguments:
    This may be either a :ref:`Secret` or just a string.
    `sshPrivateKey` must be specified in order to use this option.
    `sshHostKey` must not be specified in order to use this option.
+
+``auth_credentials``
+
+   (optional) An username/password tuple to use when running git for push operations.
```
p12tic, line 158: "for push operations" probably does not fit here, because this part is for `Git` step, not `GitPush`.

p12tic (changes requested):
Looks good, except single nit.

tdesveaux:
> FYI `@async_to_deferred` is only needed for externally callable APIs. Internal functions can be proper `async` because we can ensure that all call sites use either `yield` or `await` on it.

Since calling an async function from a deferred one (with yield) does not await the call, I felt it was less error prone to use the decorator on every async function, it could then be removed in the future when all code is async.

Other option would be to either propagate async conversion to caller (recursively) or use `defer.fromCoroutine` in the call code.

(might be helpful to set guidelines for async in #7371?).

p12tic:
> Since calling an async function from a deferred one (with yield) does not await the call, I felt it was less error prone to use the decorator on every async function, it could then be removed in the future when all code is async.

This is surprising. Supporting async functions from inlineCallbacks would be great migration strategy. I will check what's missing from it on twisted side, but this PR does not need to wait.

tdesveaux:
I was surprised as well.
I even made a unit test to try the various combination between inlineCallback / pure async and async wrapped with async_to_deferred.

The only issue was when calling an async function from an inlineCallback with yield.

p12tic (approved):

p12tic:
> I was surprised as well. I even made a unit test to try the various combination between inlineCallback / pure async and async wrapped with async_to_deferred.
> 
> The only issue was when calling an async function from an inlineCallback with yield.

FYI: https://github.com/twisted/twisted/pull/12239

PR #7544, opened 2024-04-30
Fix randomly breaking pending build requests > shows

mokibit:
This PR fixes unreliable pending build requests > shows test on webkit.


p12tic:
Seems this test is no longer a problem.

PR #7549, opened 2024-05-01
www: Make Builder related URLs take buildername

tdesveaux:
This allow to share user-friendly URLs between users.

Note that using `buildername` in APIs can be problematic, see #7552.

## Contributor Checklist:

### DIFF
--- a/www/base/src/views/BuilderView/BuilderView.tsx
+++ b/www/base/src/views/BuilderView/BuilderView.tsx
@@ -46,6 +46,7 @@ import {FaStop, FaSpinner} from "react-icons/fa";
 import {buildTopbarItemsForBuilder} from "../../util/TopbarUtils";
 import { Tab, Tabs } from "react-bootstrap";
 import { LoadingSpan } from "../../components/LoadingSpan/LoadingSpan";
+import { UrlNotFoundView } from "../UrlNotFoundView/UrlNotFoundView";
 
 const anyCancellableBuilds = (builds: DataCollection<Build>,
                               buildrequests: DataCollection<Buildrequest>) => {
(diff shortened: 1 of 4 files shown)


p12tic:
The problem with builder names in URLs is that builder names can change and then all URLs would break. I will think about how this could be solved. The PR itself looks good and would improve user experience.

tdesveaux:
> The problem with builder names in URLs is that builder names can change and then all URLs would break. I will think about how this could be solved. The PR itself looks good and would improve user experience.

I agree that it can be an issue (in addition to #7552).
I looked into this as it was a feedback from users on things they missed since upgrading from 0.8.

But, maybe I missed something, but builderid seem to not be stable either in case of name change right?
I tried with the default sample from a `create-master`, ran a build then renamed `runtests` to `run-tests` and the BuilderId changed.
So wouldn't buildernames be as faulty as builderids in urls?

p12tic:
>  So wouldn't buildernames be as faulty as builderids in urls?

You're completely right. Buildername has 1:1 relationship with builderid. Which means that there are no blockers for this PR.

tdesveaux:
Still think merging this should be on hold until #7552 is addressed as this would add onto the inconsistency of which builder can be accessed by name.

tdesveaux:
@p12tic since the issue with the data api and buildernames is fixed, there is no blocker for this AFAIK.

Let me know if you think of anything else.

www/base/src/views/BuilderView/BuilderView.tsx
```diff
@@ -140,11 +141,11 @@ export const BuilderView = observer(() => {
   })));
 
   const builder = buildersQuery.getNthOrNull(0);
-  const builds = buildsQuery.getParentCollectionOrEmpty(builderid.toString());
-  const buildrequests = buildrequestsQuery.getParentCollectionOrEmpty(builderid.toString());
-  const forceschedulers = forceSchedulersQuery.getParentCollectionOrEmpty(builderid.toString());
+  const builds = buildsQuery.getParentCollectionOrEmpty(builder?.id ?? "");
```
p12tic, line 144: This introduces another roundtrip until contents can be displayed.

p12tic (approved):

p12tic:
Sorry for a long delay. I've been assuming that there is some risk. The risk didn't exist and I understood that only after taking a deep look into the PR for the n-th time.

Issue #7527, opened 2024-04-22
GitPoller misses changes when buildbot reconfigured


doliver-pulsemining:
Our buildbot configuration builds every git commit on select branches using the SingleBranchScheduler. We regularly add or remove branches to be built and reload the config with `buildbot reconfig`.

I've been investigating a few commits that did not trigger a build in our setup. I found that when buildbot is reconfigured to pick up config changes it ignores any commit between the last poll before the reconfig and the first poll after the reconfig. 
The commit does trigger a build if buildbot is restarted.

It appears this is caused by the line [self.lastRev = {}](https://github.com/buildbot/buildbot/blob/b6a7d098471020aa4c89ebc2e4b55788a36a8392/master/buildbot/changes/gitpoller.py#L175). 
When buildbot is restarted the `activate` method loads `lastRev` from the database [self.lastRev = yield self.getState('lastRev', {})](https://github.com/buildbot/buildbot/blob/b6a7d098471020aa4c89ebc2e4b55788a36a8392/master/buildbot/changes/gitpoller.py#L215)
I think that the `reconfigService` method should also call `getState` to load the correct `lastRev` from the database.

Buildbot Version: 3.11.1
Operating System: RHEL9

mokibit:
I am looking into this.

PR #7566, opened 2024-05-05
changes: remove `GitPoller.lastRev` member

tdesveaux:
- renamed `lastRev` to `last_rev`.
- `last_rev` is now a variable of `GitPoller.poll` method.
- `last_rev` is now always read from state at each poll.
- `GitPoller.activate` is no longer a deferred.
- Minor fix to error log in `GitPoller._process_changes`. When an error occured while retrieving a rev information (author/timestamp/...) the error log would log the new HEAD rev of the branch instead of the problematic rev.
- introduced TestGitPollerBase helpers `set_last_rev` and `assert_last_rev`.

Probable fix to #7527

## Contributor Checklist:


master/buildbot/changes/gitpoller.py
```diff
@@ -226,11 +225,8 @@ def _checkGitFeatures(self):
             if has_ssh_private_key:
                 raise EnvironmentError('SSH private keys require Git 2.3.0 or newer')
 
-    @defer.inlineCallbacks
     def activate(self):
         try:
-            self.lastRev = yield self.getState('lastRev', {})
```
p12tic, line 232: There can be only a single gitpoller which updates the `lastRev` state. Storing the newest value in `self` is perfectly acceptable way to avoid queries to the database.

p12tic:
Also, there has been a lot of hairy bugs in this particular area, so I would like the changes to be as easy to review (now and later). Can the variable renames be split to at least a separate commit?

master/buildbot/changes/gitpoller.py
tdesveaux: As it's one simple read query every `pollInterval`, I felt it didn't need the (albeit small) pain of maintaining the same state value in two places.

I could change this PR to simply:
- set `lastRev` to None in ctor and `reconfigService`
- in `poll`, if `lastRev` is None, load it from DB with `getState`.

I think it would solve the issue as well (need to confirm). Let me know if you prefer this solution.

p12tic:
Rebased.

tdesveaux:
@p12tic  I think it's good now.
Only failure is react flaky smokes and some [buildbot.test.integration.test_worker_proxy](https://buildbot.buildbot.net/#/builders/89/builds/9438) which these changes should not have impacted.

PR #7571, opened 2024-05-06
react-base: Minor fix and improvement to ForceBuildModal fields

tdesveaux:
Mainly fix `FieldChoiceString` with `multiple`.

Biggest change was to untype the value stored by `ForceBuildfieldsState` as storing everything as string would be awkward in some places and I don't think it brought much to type it.

## Contributor Checklist:

### DIFF
--- a/www/react-base/src/components/ForceBuildModal/Fields/FieldInt.tsx
+++ b/www/react-base/src/components/ForceBuildModal/Fields/FieldInt.tsx
@@ -34,7 +34,7 @@ export const FieldInt = observer(({field, fieldsState}: FieldIntProps) => {
       <div className="col-sm-10">
         <input data-bb-test-id={`force-field-${field.fullName}`}
                type="text" className="form-control" value={state.value}
-               onChange={event => fieldsState.setValue(field.fullName, event.target.value)}/>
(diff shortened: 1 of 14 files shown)


p12tic (approved):
Took a while to agree that this PR makes sense after all.

p12tic:
@tdesveaux Does this PR really not need a newsfragment? It does fix something.

tdesveaux:
> @tdesveaux Does this PR really not need a newsfragment? It does fix something.

You're right, I can't remember why I thought differently.
I'll also look into adding tests on the 3 fields component this touches so there is no regression.

www/react-base/src/components/ForceBuildModal/Fields/FieldInt.tsx
```diff
@@ -34,7 +34,7 @@ export const FieldInt = observer(({field, fieldsState}: FieldIntProps) => {
       <div className="col-sm-10">
         <input data-bb-test-id={`force-field-${field.fullName}`}
                type="text" className="form-control" value={state.value}
```
tdesveaux, line 36: With the change to the state storage, maybe this should be change to `type="number"?

p12tic: Agreed

PR #7577, opened 2024-05-06
Change GitPoller tracker branch name format

### LINKED ISSUE #7061: GitPoller fails with filename too long when using Bitbucket repository tokens
We've hit an issue with GitPoller as we try to move from away from app passwords for our CI jobs to using repository tokens.  The issue seems to be that GitPoller uses `git fetch <repo_url> +<src>:<dst>` where the `<dst>` directory uses an escaped version of the full repo URL.  When using repository tokens, the URL is of the form `https://x-token-auth:ReallyLongToken@bitbucket.org/org/repo.git` and the full destination refspec becomes too long.

Is there any reason that the full URL is used in the destination refspec here:  https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/gitpoller.py#L209

Is it just creating a usefully named destination directory and if so, could we detect the use of a token and just strip it?  Or are there nicer ways of solving this?

tdesveaux:
Fixes #7061

On systems with filepath limitation (eg. on Windows, old APIs limit the number of char in a filename and a full path), current implementation could result in large path.
Removing the credential information from the url should reduce the size (as well as not having credentials in clear on disk).
Not keeping `/` between url parts (protocol, domain, owner, repo) will split into subdirs.

## Contributor Checklist:


p12tic:
Last commit message: typo in `tracker_branche identifier`

master/buildbot/changes/gitpoller.py
```diff
@@ -281,9 +282,24 @@ def _removeHeads(self, branch):
             branch = branch[11:]
         return branch
 
-    def _trackerBranch(self, branch):
-        url = urlquote(self.repourl, '').replace('~', '%7E')
-        return f"refs/buildbot/{url}/{self._trim_prefix(branch, 'refs/')}"
+    def _tracker_branch(self, branch: str) -> str:
```
p12tic, line 285: I think we should have more tests for this function if we're changing it, especially the case with password which this PR fixes.

p12tic (changes requested):
Looks good in principle, thanks! We need more tests for _tracker_branch function.

master/buildbot/changes/gitpoller.py
p12tic: The _tracker_branch function could be made staticmethod so that whole class instance does not need to be instantiated.

master/buildbot/test/unit/changes/test_gitpoller.py
```diff
@@ -2247,3 +2247,59 @@ def test_branch_and_only_tags(self):
     def test_gitbin_default(self):
         poller = yield self.attachChangeSource(gitpoller.GitPoller("/tmp/git.git"))
         self.assertEqual(poller.gitbin, "git")
+
+
+class TestGitPollerUtils(unittest.TestCase):
+    def test_tracker_ref_protos(self):
+        for url, expected_traker in [
```
p12tic, line 2254: typo

p12tic (changes requested):
LGTM except typo

p12tic (approved):

Issue #7591, opened 2024-05-10
Modernize Database layer


tdesveaux:
As per #7385 , I'm adding type information here and there as I work on other things.

One thing I'm noticing is that the database models do not have [mapped python](https://docs.sqlalchemy.org/en/14/tutorial/metadata.html#declaring-mapped-classes) classes to interact, which would help typing. Is that by choice or a legacy implementation?

If I understand the code correctly, database data is passed around as `dict` which then need tests to validate that the data is as expected with `Validator` (defined [here](https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/util/validation.py#L294))?

Some things that could be done from the top of my head:
- require [sqlalchemy >=1.4](https://github.com/buildbot/buildbot/blob/master/master/setup.py#L660) (first released on [march 2021](https://docs.sqlalchemy.org/en/20/changelog/changelog_14.html#change-1.4.0)) which is the version made to be a [transition point to SQLAlchemy 2.0](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html) (this would allow to fix deprecation warnings ignore [here](https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/__init__.py#L163-L170))
- Implement DB models as mapped classes (or make dataclasses first, then make these dataclasses mapped afterward using [Hybrid Declaration](https://docs.sqlalchemy.org/en/14/orm/dataclasses.html#orm-declarative-dataclasses))

Also, it seems DB queries are run from a threadpool, an obvious reason is that queries are synchronous. If there is no other reason, SQLAlchemy 1.4.3 (March 2021) introduced asyncio support at a [beta level](https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html#) (although there is bugfixes as recently as a critical one in [1.4.51 (Jan 2024)](https://docs.sqlalchemy.org/en/20/changelog/changelog_14.html#change-13dac36b3e17cdbe4099c761e5151fb8), so there is reason to hold off on it.

Here is as far as I got looking into it for now. As I don't have a good understanding of Buildbot's internals nor history, I wanted to get maintainers opinion on this. I'm willing to submit PRs on this (hopefully, finding a way to make them as small as possible).

p12tic:
I think what you wrote is mostly correct.

Dataclasses is a welcome addition.

[...]

PR #7592, opened 2024-05-11
Bump SQLAlchemy requirement to >1.4.0

tdesveaux:
Draft until warn when using different DBs are resolved.

Relates to #7591.

This change the minimum required version of SQLAlchemy to 1.4.0 which is a version made to prepare upgrade to 2.0.
Also bumps SQLAlchemy for CI to 1.4.52.

Also fixes the majority of `RemovedIn20Warning` warnings, leaving 3 unresolved ones as their fixes are less simple than the others.
Each warning fix is a self contained commit to ease review, but I can split this in mulitple PRs if needed.

Maybe `SQLALCHEMY_WARN_20=1` should be set in env in `master/buildbot/test/__init__.py` as well, otherwise, running trial locally without it will result in errors as the warn message is a generic one:
`Deprecated API features detected! These feature(s) are not compatible with SQLAlchemy 2.0. To prevent incompatible upgrades prior to updating applications, ensure requirements files are pinned to "sqlalchemy<2.0". Set environment variable SQLALCHEMY_WARN_20=1 to show all deprecation warnings.  Set environment variable SQLALCHEMY_SILENCE_UBER_WARNING=1 to silence this message.`

## Contributor Checklist:


master/buildbot/db/dbconfig.py
```diff
@@ -61,7 +61,8 @@ def getDb(self):
         db.model = model.Model(db)
         db.state = state.StateConnectorComponent(db)
         try:
-            self.objectid = db.state.thdGetObjectId(db_engine, self.name, "DbConfig")['id']
+            with db_engine.connect() as conn:
```
p12tic, line 64: The thread pool is to ensure that the main thread does not block. Is db_engine.connect() guaranteed not to block even for a short time?

master/buildbot/db/logs.py
```diff
@@ -427,7 +427,7 @@ def thddeleteOldLogs(conn):
 
         return self.db.pool.do(thddeleteOldLogs)
 
-    def _logdictFromRow(self, row):
-        rv = dict(row)
+    def _logdictFromRow(self, row: sa.engine.row.Row):
+        rv = dict(row._mapping)  # noqa pylint: disable=protected-access
```
p12tic, line 434: Using internal variable doesn't seem like a long-term solution. I think we should use mappings() function https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.mappings

p12tic (changes requested):
Thanks a lot. Several comments added, should be easy to address.

master/buildbot/db/logs.py
tdesveaux: This is indeed far better,
I guess I use the `_mapping` as it lead to less change, but that's a poor reason and `mappings()` is a far better alternative.
I'll update the PR.

master/buildbot/db/dbconfig.py
tdesveaux: I think this does not change the blocking aspect of this code as `thdGetObjectId` is the "implementation` with `_getObjectId` being the wrapper that will thread it.

I though direct call to `thdGetObjectId` would block the thread as well, am I wrong?

p12tic: OK, I was confused about which code path this is used in. All good, thanks.

PR #7597, opened 2024-05-12
react-base: Improve BuildView 'Rebuild' topbar action

### LINKED ISSUE #7450: Rebuild UI unclear
When you hit Rebuild there's no indication of that request succeeding. And in fact if you didn't know better you might end up clicking on Rebuild several times to try to get something going.

1. It would be good if after you click Rebuild there was some indication of success.
2. And if you then go back to the prior page it would be good if the Rebuild button was annotated in some way that makes it clear a rebuild is queued.

tdesveaux:
Provide better feedback to user with the following changes:
- Caption is now one of 'Rebuilt' (finished rebuild), 'Rebuilding' (rebuild in progress), 'Rebuild pending' (rebuild requested but not started).
- Button now redirect to the Buildrequest for the rebuild, which, in turn, will redirect to the Build if it started.
- If multiple rebuild were issued, redirection is done to the newest one (can be changed to the one closest to the Build viewed, see NOTE in code).
- When 'Rebuild' button is clicked but new Buildrequest wasn't found yet, caption is 'Rebuilding' as before, but button is a noop.

Closes #7450

## Contributor Checklist:

### DIFF
--- a/newsfragments/react-base-BuildView-improve-rebuild-ui.feature
+++ b/newsfragments/react-base-BuildView-improve-rebuild-ui.feature
@@ -1,0 +1,1 @@
+The Rebuild button on a Build's view now redirect to the Buildrequest corresponding to the latest rebuild.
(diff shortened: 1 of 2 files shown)


p12tic (approved):

p12tic:
Rebased.

PR #7598, opened 2024-05-12
typing: convert `db.sourcestamps.SsDict` to dataclass

tdesveaux:
Relates to #7591 

This converts `db.sourcestamps.SsDict` from a class inheriting from `Dict` to a dataclass.
Split `patch` information to another dataclass representing DB model.
Added a `__get_item__` method with deprecation warning for backward compatibility.

This improves developer experience, as well as allow to remove validation tests.

Note BEFORE merging, `SsDict` should be renamed. I think `SourceStampModel` would be appropriate?

## Contributor Checklist:


p12tic:
SourceStampModel sounds good.

p12tic (approved):

PR #7607, opened 2024-05-13
typing: convert db.buildrequests.BrDict to dataclass

tdesveaux:
Relates to #7591

This one was a bit more annoying.
Contains some fixes/improvements.

Not super happy about the change in `_getBuildRequestForBrdict` so let me know if you see something better.

Also, since this is planned to be in 3.12, if you create the branch, I can take care of rebase + PR against it.

## Contributor Checklist:

### DIFF
--- a/common/code_spelling_ignore_words.txt
+++ b/common/code_spelling_ignore_words.txt
@@ -204,6 +204,7 @@ buildrequestid
 buildrequestresults
 buildrequests
 buildrequestsconnectorcomponent
+buildrequestmodel
 buildresult
 buildroot
 build's
--- a/master/buildbot/data/buildrequests.py
+++ b/master/buildbot/data/buildrequests.py
@@ -29,22 +29,23 @@ if TYPE_CHECKING:
     from typing import Sequence
 
     from buildbot.data.resultspec import ResultSpec
+    from buildbot.db.buildrequests import BuildRequestModel
 
 
-def _db2data(dbdict: dict, properties: dict | None):
+def _db2data(dbmodel: BuildRequestModel, properties: dict | None):
     return {
-        'buildrequestid': dbdict['buildrequestid'],
-        'buildsetid': dbdict['buildsetid'],
-        'builderid': dbdict['builderid'],
-        'priority': dbdict['priority'],
-        'claimed': dbdict['claimed'],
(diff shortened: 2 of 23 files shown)


p12tic:
I think the change with `_getBuildRequestForBrdict` is fine. If it turns out a performance problem we can try optimizing it.

The root cause is having buildername in build request which I think should not live there. But let's not deprecate too much at a single time.

p12tic (approved):
Looks good, but some tests are broken.

p12tic:
> Also, since this is planned to be in 3.12, if you create the branch, I can take care of rebase + PR against it.

I think for now let's focus on having all pieces in the master branch first. I can easily backport your PRs if you disappear for a month, but if the conversion itself is not finished this would be way worse.

p12tic:
Also I have some doubts whether backporting to 3.12.0 is the best way to go. The database schema has changed in 4.0, so the backporting will not be trivial and the chance of bugs is increased. The cost of doing this change in e.g. 4.1.0 is just carrying deprecated `__getitem__` calls which is a small cost.

One way to go is once 4.0 is otherwise ready, create 4.0.x branch, revert the PRs that have gone in so far and cut v4.0.0. Master branch keeps going on to be 4.1.0. The cost of doing so is increased number of backports to 4.0.x, but many of the backports will need to go to 3.11.x anyway, so it's not a huge increase of effort.

tdesveaux:
> Also I have some doubts whether backporting to 3.12.0 is the best way to go. The database schema has changed in 4.0, so the backporting will not be trivial and the chance of bugs is increased. The cost of doing this change in e.g. 4.1.0 is just carrying deprecated `__getitem__` calls which is a small cost.
> 
> One way to go is once 4.0 is otherwise ready, create 4.0.x branch, revert the dataclass PRs that have gone in so far and cut v4.0.0. Master branch keeps going on to be 4.1.0. The cost of doing so is increased number of backports to 4.0.x, but many of the backports will need to go to 3.11.x anyway, so it's not a huge increase of effort.

I think pushing this to 4.x is sensible, it could indeed lead to some bugs.
Once all the db have their model merged, I'll look into spawning a test instance of our setup. Although I don't think we cover that much of Buildbot's features, we do have some traffic so that could unearth some things.

p12tic (approved):

PR #7459, opened 2024-02-28
Improvements to the React GUI for builders and workers

YngveNPettersen:
We have been using patches similar to these for several years in the Angular JS version.

The GUI patches are:

* Project configuration of number of recent builds to display in the Home App
* Show the revision property for a build in the builders and workers pages
* Improve display of build completion time
* Let the project specify properties to use to display e.g. a build 1.2.3.4 version number instead of just build number 999 on that builder.
*Let the Tags column on the Builders page be less wide when there are no tags

Additionally:

* Fix the Makefile to work on Windows

## Contributor Checklist:

* I have updated the unit tests (no tests affected AFAICT)
* I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* I have updated the appropriate documentation

PR #7613, opened 2024-05-15
www: Add a way to configure text in build links

p12tic:
This allows to customize Buildbot deployments without having to modify Buildbot code.

## Contributor Checklist:


p12tic:
Extracted from https://github.com/buildbot/buildbot/pull/7459.

PR #7620, opened 2024-05-16
[WIP] Support for Python 3.13

pmisik:
Work in progress to see what needs to be fixed to be compatible with future Python 3.13.

## Contributor Checklist:

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -44,6 +44,7 @@ matrix:
     - env: PYTHON=3.10 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
     - env: PYTHON=3.11 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
     - env: PYTHON=3.12 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
+    - env: PYTHON=3.13 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
 
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=dev_virtualenv
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
 
     env:
       # Needed to ignore specific warns, otherwise, it'll warn a generic message
--- a/master/setup.py
+++ b/master/setup.py
@@ -135,6 +135,7 @@ setup_args = {
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3.11',
         'Programming Language :: Python :: 3.12',
+        'Programming Language :: Python :: 3.13',
     ],
     'packages': [
         "buildbot",
--- a/worker/setup.py
+++ b/worker/setup.py
@@ -90,6 +90,7 @@ setup_args = {
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3.11',
         'Programming Language :: Python :: 3.12',
+        'Programming Language :: Python :: 3.13',
     ],
     'packages': [
         "buildbot_worker",


yan12125:
The first CI error seems related to a Pillow issue: https://github.com/python-pillow/Pillow/pull/8050

I also got erros in other dependencies: cffi, greenlet, ruamel.yaml.clib, scandir. A tracking issue may be worthwhile for those dependencies.

pmisik:
> A tracking issue may be worthwhile for those dependencies.

Tracking issue https://github.com/buildbot/buildbot/issues/7631

yan12125:
For Pillow, the relevant pull request is merged, and the `exec` branch is gone. Thus, `git+https://github.com/radarhere/Pillow.git@exec` can be replaced with `git+https://github.com/python-pillow/Pillow.git`.

pmisik:
> For Pillow, the relevant pull request is merged, and the exec branch is gone. Thus, git+https://github.com/radarhere/>Pillow.git@exec can be replaced with git+https://github.com/python-pillow/Pillow.git.

@yan12125  thanks for hit.
I've applied change.

Based on pillow issue tracker it seems we might expect release of compatible pillow package around July 1, 2024 https://github.com/python-pillow/Pillow/issues/8076

pmisik:
I've removed changes regarding to pillow as 10.4.0 was released and was recently merged to master.
I've also rebased this PR.

yan12125:
Hi, could you consider extracting changes related to `cgi` as a separate pull request? Adding Python 3.13 support for non-Windows first will be great.

master/buildbot/www/rest.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, contenttype):
         self.typeheader = contenttype
 
     def gettype(self):
-        mimetype, _ = cgi.parse_header(bytes2unicode(self.typeheader))
+        mimetype, _ = parse_header(bytes2unicode(self.typeheader))
```
tdesveaux, line 64: Looks like this could simply be replaced with
```python
if self.typeheader is None:
    return None
return bytes2unicode(self.typeheader).split(';', 1)[0]
```

[ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#syntax)

pmisik:
Does anybody have an idea why `buildbot.test.unit.db.test_logs.TestFakeDB.test_addLog_getLog` fails with Python 3.13?
I tried to understand it but I failed.
I did not find any reason why it fails.
I admit that python is not my primary language, so my report may be inappropriate.

tdesveaux:
> Does anybody have an idea why `buildbot.test.unit.db.test_logs.TestFakeDB.test_addLog_getLog` fails with Python 3.13? I tried to understand it but I failed. I did not find any reason why it fails. I admit that python is not my primary language, so my report may be inappropriate.

Managed to reproduce (although it's flaky).
I suspect the issue is with this: https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/fake/fakemaster.py#L165

`setServiceParent` is a deferred which is not awaited. I'll take a stab at patching `make_master` to make it await it (and everywhere it's called...).

Still, did not yet looked into it more, but a bit confused as to why only the FakeDB tests fail (specifically, only the `test_logs` one until now). Didn't see where a sqlite connection is made in them either.

tdesveaux:
#8103 should fix the test issue

p12tic:
@tdesveaux 

> I'll take a stab at patching make_master to make it await it (and everywhere it's called...).

Sorry for noticing this only now - I have the fix already implemented in a local checkout. Will create PRs sometime next week.

tdesveaux:
> Sorry for noticing this only now - I have the fix already implemented in a local checkout. Will create PRs sometime next week.

Oh, no worries, I forgot about this.
The real issue for this PR was unrelated (PR mentioned above).

pmisik:
@p12tic 
Could you please have a look at https://github.com/buildbot/metabbotcfg/pull/159?
I hope this is last issue to solve Buildbot compatibility with Python 3.13.
It seems it work on Windows now.
We need to be confident it works on Linux as well.

p12tic:
```
Error: 
  File "C:\hostedtoolcache\windows\Python\3.13.0\x64\Lib\site-packages\twisted\trial\_synctest.py", line 1071, in run
    warnings.warn_explicit(**w)
builtins.ResourceWarning: unclosed database in <sqlite3.Connection object at 0x0000028AB5C6AE30>

buildbot.test.unit.changes.test_hgpoller.TestHgPollerBranches.test_poll_regular
```

Amazing. This warning will allow us to track a number of race conditions due to unclosed databases.

p12tic (approved):

PR #7621, opened 2024-05-16
process: Fix crash when collapseRequestsFn is None

p12tic:
Fixes https://github.com/buildbot/buildbot/pull/7607/commits/8e9df31ddee62b77997e6d771a60dcda68f59b10.

@tdesveaux this is perfect illustration why a single commit should change only single thing and any unrelated changes should be at least in separate commits. Large commits mean errors become easier to miss in review.


p12tic:
Merging without CI because buildbot CI is broken now.

tdesveaux:
Absolutely, I got a bit too confident in tests to catch issues (and far too much in myself to not make dumb things).

I'll make a better effort to split code change in the future PRs. Apologize for the issue it caused.

PR #7633, opened 2024-05-19
Removal of scandir package from requirements-ci.txt

### LINKED ISSUE #7632: Removal of scandir package from requirements
`scandir` has been included in the Python 3.5 standard library as os.scandir(), and the related performance improvements to os.walk() have also been included. see https://github.com/benhoyt/scandir?tab=readme-ov-file#now-included-in-a-python-near-you

As Buildbot now requires Python 3.8 or newer we can use standard library rather than package.

pmisik:
`scandir` has been included in the Python 3.5 standard library as os.scandir(), and the related performance improvements to os.walk() have also been included. see https://github.com/benhoyt/scandir?tab=readme-ov-file#now-included-in-a-python-near-you

As Buildbot now requires Python 3.8 or newer we can use standard library rather than package.

Fixes #7632

## Contributor Checklist:

### DIFF
--- a/requirements-ci.txt
+++ b/requirements-ci.txt
@@ -102,7 +102,6 @@ responses==0.25.0
 ruamel.yaml==0.18.6
 ruamel.yaml.clib==0.2.8
 s3transfer==0.10.1
-scandir==1.10.0
 service-identity==24.1.0
 setuptools-trial==0.6.0
 singledispatch==4.1.0


p12tic:
requirements-ci.txt is used to ensure that the set of installed packages is reproducible in CI. We should ensure that scandir does not come from some transitive dependency. We can remove this package only if there are no such dependencies.

p12tic:
> If there was a transitive dependency, then scandir should appear in the pip list output.

You're right. I removed the transitive dependencies from requirements-ci.txt, installed it and it didn't bring scandir.

p12tic (approved):

PR #7674, opened 2024-06-01
test: fix SQLAlchemy 'current statement is being autocommitted'

tdesveaux:
This removes the last SQLAlchemy RemovedIn20Warning from tests.

Once merge, this allow to start work on using the `future=True` in Engine and SQLAlchemy 2.0 compatibility.

I went with simply wrapping insert/update/delete call in a transaction as it's what was happening, now it's simply obvious.
Transaction usage can be optimized/improved in the future, but I feel it should not be done here.

I also added the `DBPool.do_with_transaction` helper as I felt it was useful for most simple cases, but the commit can be dropped.

## Contributor Checklist:


tdesveaux:
Can't repro the test failure locally, but I'm using py3.11 so that might be the diff.
Could the test be re-run to check if it's "just" flaky?

p12tic (approved):

PR #7690, opened 2024-06-07
db: implement upsert

tdesveaux:
Requires #7686

This implement upsert for case where a DB entry needs to be updated, and created if it does not exists.
This allow to replace manual implementation where race condition can happen.

Only implemented for MySQL, PSQL, and SQLite.
Fallback on manual implementation if driver or DB version is not compatible.

## Contributor Checklist:


p12tic:
Rebased.

p12tic (approved):
Looks great, thanks!

PR #7693, opened 2024-06-08
tests: add a bare git repository

### LINKED ISSUE #7666: tests: Add bare git repository to test GitPoller
I introduced a bug (#7663) in #7554 where I changed a `git rev-parse` command to add `--` which changed the expected output of the command.
While this is my fault as I should have better tested my changes, there is a simple way to prevent this kind of mistakes from happening in the future.

Currently, `GitPoller` tests (and probably all other git tests) are run with tests that list the commands that are run and check it matches what's expected by patching the command that would ultimately run the command.
This mean the command is not run, so tests also specify what the output of the command is expected to be.

I propose to introduce a new set of tests for GitPoller first, then we can see where it can be implemented if appropriate, which would run the commands against a git repository which content is controlled in a similar way snapshots are used in react tests.

With Git, this can be pretty simple to setup with a bare repository.
For example, a repo could be added in `master/buildbot/test/test_git_repo/`.
Here is an log:
```bash
$ git -C master/buildbot/test/test_git_repo/ log --graph --oneline
*   2d68b0b503 (HEAD -> main) Merge branch 'feature/1'
|\  
| * 0bf12173c4 (feature/1) feature 1
* | 4de337cb51 (fix/1) Fix 1
|/  
* 17fccfdfd6 Initial
```
and a tree:
```bash
tree master/buildbot/test/test_git_repo/
master/buildbot/test/test_git_repo/
├── HEAD
├── objects
│   ├── 0b
│   │   └── f12173c47bf4572f0aa0374bf40d7ca93dcca4
│   ├── 0c
│   │   └── 34d5ebdc778bae104c9d17c1ad4382532cf80d
│   ├── 0f
│   │   └── 9c9f5015ff51b0fac92d5b27578cc93e1eb899
│   ├── 17
│   │   └── fccfdfd69e10e4c88e1b1406b9bf89e6a82216
│   ├── 2d
│   │   ├── 68b0b50304c5dcda6a628ed1c8cda15d36a16b
│   │   └── 96fa693ced7a641ab93004198ca62c4f5cdb79
│   ├── 4d
│   │   └── e337cb5132592e818b1bba8cc153d1e0961dd1
│   ├── 74
│   │   └── a9cc18106c388615c31407045708006123fd70
│   ├── 7d
│   │   └── 069cf8c836331952e4ec416715ca9092217825
│   ├── 8d
│   │   └── 5d568501f0d0f4305701598c9a6009cbdb25f7
│   ├── ef
│   │   └── 61c289021843f51ec130f1596f756a9d1c4330
│   └── fa
│       └── aae1b1924ed1b41c269a28b799c68bf638ddcb
└── refs
    └── heads
        ├── feature
        │   └── 1
        ├── fix
        │   └── 1
        └── main

17 directories, 16 files
```

As seen with the `git log` command above, the repository can be accessed with git commands that don't interact with the working dir.
It can also be used as a remote, eg. the `ls-remote` command used in `GitPoller`:
```bash
$ git ls-remote master/buildbot/test/test_git_repo/
2d68b0b50304c5dcda6a628ed1c8cda15d36a16b	HEAD
0bf12173c47bf4572f0aa0374bf40d7ca93dcca4	refs/heads/feature/1
4de337cb5132592e818b1bba8cc153d1e0961dd1	refs/heads/fix/1
2d68b0b50304c5dcda6a628ed1c8cda15d36a16b	refs/heads/main
```

If the repository need to be updated for new tests, or used in tests for `GitStep`, it can be cloned and fetch/pull/pushed.
To create the repo, I did:
- `git init --bare master/buildbot/test/test_git_repo`
- `git clone master/buildbot/test/test_git_repo /tmp/test_git_repo`
- Added commits, merges, etc in `/tmp/test_git_repo`
- still in `/tmp/test_git_repo`, `git push [branches...]`
- removed not useful files from `master/buildbot/test/test_git_repo` (git init create things like config/exclude and others that are not useful here)

tdesveaux:
This allow to implement GitPoller tests without faking git commands.

Closes #7666

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/fake/fakemaster.py
+++ b/master/buildbot/test/fake/fakemaster.py
@@ -156,7 +156,7 @@ def make_master(
         master.mq.setServiceParent(master)
     if wantDb:
         assert testcase is not None, "need testcase for wantDb"
-        master.db = fakedb.FakeDBConnector(testcase)
+        master.db = fakedb.FakeDBConnector(testcase, reactor=_reactor)
         master.db.setServiceParent(master)
     if wantData:
         master.data = fakedata.FakeDataConnector(master, testcase)
--- a/master/buildbot/test/fakedb/base.py
+++ b/master/buildbot/test/fakedb/base.py
@@ -19,10 +19,10 @@ from buildbot.data import resultspec
 class FakeDBComponent:
     data2db = {}
 
-    def __init__(self, db, testcase):
+    def __init__(self, db, testcase, reactor=None):
         self.db = db
         self.t = testcase
-        self.reactor = testcase.reactor
+        self.reactor = reactor if reactor is not None else testcase.reactor
         self.setUp()
 
     def mapFilter(self, f, fieldMapping):
--- a/master/buildbot/test/fakedb/connector.py
+++ b/master/buildbot/test/fakedb/connector.py
@@ -48,50 +48,52 @@ class FakeDBConnector(AbstractDBConnector):
     see their documentation for more.
     """
 
-    def __init__(self, testcase):
+    def __init__(self, testcase, reactor=None):
+        if reactor is None:
+            reactor = testcase.reactor
         super().__init__()
         # reset the id generator, for stable id's
         Row._next_id = 1000
         self.t = testcase
         self.checkForeignKeys = False
         self._components = []
-        self.changes = comp = FakeChangesComponent(self, testcase)
+        self.changes = comp = FakeChangesComponent(self, testcase, reactor)
         self._components.append(comp)
-        self.changesources = comp = FakeChangeSourcesComponent(self, testcase)
+        self.changesources = comp = FakeChangeSourcesComponent(self, testcase, reactor)
         self._components.append(comp)
-        self.schedulers = comp = FakeSchedulersComponent(self, testcase)
+        self.schedulers = comp = FakeSchedulersComponent(self, testcase, reactor)
         self._components.append(comp)
-        self.sourcestamps = comp = FakeSourceStampsComponent(self, testcase)
+        self.sourcestamps = comp = FakeSourceStampsComponent(self, testcase, reactor)
         self._components.append(comp)
-        self.buildsets = comp = FakeBuildsetsComponent(self, testcase)
+        self.buildsets = comp = FakeBuildsetsComponent(self, testcase, reactor)
         self._components.append(comp)
-        self.workers = comp = FakeWorkersComponent(self, testcase)
+        self.workers = comp = FakeWorkersComponent(self, testcase, reactor)
         self._components.append(comp)
-        self.state = comp = FakeStateComponent(self, testcase)
+        self.state = comp = FakeStateComponent(self, testcase, reactor)
         self._components.append(comp)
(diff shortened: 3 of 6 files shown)


p12tic:
Would it be possible to reproduce the repository state on each run? I think git commit does expose enough options and environment variables to make the repository reproducible.

p12tic (approved):
LGTM, just one question.

tdesveaux:
Not sure I understand the question.

Do you mean, declaring the repository in the code and generating it at runtime in test start via calling git with options to make sure it's idempotent vs having a bare repository committed as implemented here?

I guess it could be done, but I don't know how complex it would be to find all git conf entries and command options. I feel it would be a hassle in the long run if the worker has a git config change that we didn't notice here or for a setting that was introduced in the future.

p12tic:
> Do you mean, declaring the repository in the code and generating it at runtime in test start via calling git with options to make sure it's idempotent vs having a bare repository committed as implemented here?

It would be possible to assert git hash on each test. I think git only includes file data and metadata, author and committer info and dates.

p12tic:
For example, this sequence of commands produces git repositories with identical commit hash:

```
git init
echo 1 > test
touch -t 201401010102.03 test
git add test
GIT_AUTHOR_DATE=2014-01-01T05:06:07 GIT_COMMITTER_DATE=2014-01-01T05:06:07 git commit --author="Test <test@example.com>" -m "test"
```

git log returns:

```
commit e7360cbb049aafde7f693ab6abb4dbc685ba7fbe (HEAD -> master)
Author: Test <test@example.com>
Date:   Wed Jan 1 05:06:07 2014 +0200

    test
```

tdesveaux:
Thanks for checking the technical possibility of this.

My question was more regarding why this would be needed and/or the benefits compared to my approach in this PR.
Is the issue the addition of a bare 'fake' repository? Where you'd prefer the repository to be generated 'on-demand' in test setup?

p12tic:
> My question was more regarding why this would be needed and/or the benefits compared to my approach in this PR.

The reason why this is better than the current approach is that it's possible to review it. From the point of view of a reviewer the current PR adds a bunch of binary blobs to the repository. Anyone who wants to really understand what's happening would need to checkout the repository and check manually, which is inconvenient. This has security implications as well, as the xz vulnerability happened in a similar way. This PR may be safe, but let's consider a potential scenario when someone else comes with a PR that improves the test and one of the updated binary blobs contains malware. Therefore it's better to not rely on binary blobs at all, if possible.

tdesveaux:
I thought about the PR review side, and that's why I added this [`TEST_REPO_STATE`](https://github.com/buildbot/buildbot/blob/aafe4f683e191b7208642bd14672ecdeb23360f5/master/buildbot/test/util/git_repository.py#L30) thing (with the asset in test functions that uses it. I felt it was an ok tradeoff to be able to edit the embedded repository with know git tools.

I did not consider the security aspect though you make a good point. I'll rework the PR to use what you suggested.

p12tic (approved):

p12tic:
Thanks!

Issue #7536, opened 2024-04-29
v3.11.1 re-tagged?


yan12125:
Hi, I maintain the [buildbot](https://archlinux.org/packages/extra/any/buildbot/) package in Arch Linux. Here, several ways can be used to guarantee the integrity of sources downloaded for creating packages. For this package, I choose to use the checksum of a git archive. Specifically, the following command is used to get the checksum:
```
git archive --format tar v3.11.1 | sha256sum
```
On Apr 10, I got `18d4d5e5f9cdb5ad3f6ee5b0eea52ec83a7943c0827a1fe735a7b0b159923f9f`, while on Apr 23, the checksum became `b1895b9cd6c73b3fe132460626dbbcf6cfd3d7d14fa11cb10ea21eae0cef964b`. Is v3.11.1 re-tagged recently?

yan12125:
I noticed that diffoscope shows some differences between https://github.com/buildbot/buildbot/releases/download/v3.11.1/buildbot-v3.11.1.gitarchive.tar.gz and the archive from the git command mentioned above:
```
├── buildbot-3.11.1/master/buildbot/__init__.py
│ @@ -53,15 +53,15 @@
│      for root, _, files in os.walk(cwd):
│          for f in files:
│              m = max(os.path.getmtime(os.path.join(root, f)), m)
│      d = datetime.datetime.fromtimestamp(m, datetime.timezone.utc)
│      return d.strftime("%Y.%m.%d")
│
│
│ -def getVersionFromArchiveId(git_archive_id='1708788305  (tag: v3.11.1)'):
│ +def getVersionFromArchiveId(git_archive_id='1708788305  (tag: v3.11.1, 3.11.x)'):
│      """ Extract the tag if a source is from git archive.
│
│          When source is exported via `git archive`, the git_archive_id init value is modified
│          and placeholders are expanded to the "archived" revision:
│
│              %ct: committer date, UNIX timestamp
│              %d: ref names, like the --decorate option of git-log
├── buildbot-3.11.1/pkg/buildbot_pkg.py
│ @@ -80,15 +80,15 @@
│      for root, dirs, files in os.walk(cwd):
│          for f in files:
│              m = max(os.path.getmtime(os.path.join(root, f)), m)
│      d = datetime.datetime.fromtimestamp(m, datetime.timezone.utc)
│      return d.strftime("%Y.%m.%d")
│
│
[...]

PR #7742, opened 2024-06-28
Fix: git-archive archives are not reproducible

lf-:
What happened:
%d would expand to `(tag: v3.11.3, 3.11.x)`, but once 3.11.3 is not the latest in the 3.11 series, 3.11.x will cease applying, resulting in the archive not being reproducible.

The fix:
Use a describe specifier, which is *slightly* more sound. It still would break if someone tags an ancestor of a commit, but it should never break on tagged versions, assuming they have just one tag.

Testing conducted (without tagging the version, then after tagging it): $ git archive --format=tar.gz --prefix buildbot-src/ @ > buildbot.tar.gz $ tar xf buildbot.tar.gz
$ cd buildbot-src/master
$ python -c 'import buildbot; print(buildbot.getVersionFromArchiveId())'

## Contributor Checklist:

* [ ] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory) - seems unnecessary given it is a packaging change that should not be visible

### DIFF
--- a/master/buildbot/__init__.py
+++ b/master/buildbot/__init__.py
@@ -59,14 +59,15 @@ def mTimeVersion(init_file):
     return d.strftime("%Y.%m.%d")
 
 
-def getVersionFromArchiveId(git_archive_id='$Format:%ct %d$'):
+def getVersionFromArchiveId(git_archive_id='$Format:%ct %(describe:abbrev=10)$'):
     """Extract the tag if a source is from git archive.
 
     When source is exported via `git archive`, the git_archive_id init value is modified
     and placeholders are expanded to the "archived" revision:
 
         %ct: committer date, UNIX timestamp
-        %d: ref names, like the --decorate option of git-log
+        %(describe:abbrev=10): git-describe output, always abbreviating to 10 characters of commit ID.
+                               e.g. v3.10.0-850-g5bf957f89
 
     See man gitattributes(5) and git-log(1) (PRETTY FORMATS) for more details.
     """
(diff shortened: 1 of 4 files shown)


yan12125:
FWIW, we were hit by a similar issue (https://github.com/buildbot/buildbot/issues/7536), and we "fixed" it by disabling such substitutions altogether: https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/176/diffs

lf-:
> FWIW, we were hit by a similar issue (#7536), and we "fixed" it by disabling such substitutions altogether: https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/176/diffs

Since we tree-hash the sources and have a comically overpowered build system to use, we can actually have a post -extract command that fixes all instances of the substitution with sed, if we really wanted to. But I am hopeful this change should make that unnecessary.

Given how many of these bugs I've found when doing an audit though, I'm inclined to agree with you that we should just ban export-subst as a general rule and erase all the placeholders in every package that uses it. And our buildbot package should stop using fetchFromGitHub and instead use the gitarchive release artifact you mentioned since at least it wouldn't change.

yan12125:
> Given how many of these bugs I've found when doing an audit though, I'm inclined to agree with you that we should just ban export-subst as a general rule and erase all the placeholders in every package that uses it. And our buildbot package should stop using fetchFromGitHub and instead use the gitarchive release artifact you mentioned since at least it wouldn't change.

Yeah. In Arch Linux, several packages are affected, and only buildbot can use such `gitarchive` trick, so I proposed to fix it once for everyone in the build system.

p12tic:
I apologize in advance that it will take time to review this PR. I'm not that well versed in this part of Python packaging and I will need to study it before reviewing the PR.

p12tic (approved):

p12tic:
Rebased and ran `ruff format`.

PR #7744, opened 2024-06-28
git: fetching, and submodules - use shallowClone

grooverdan:
It isn't just the clone stage that should use shallowClone, the fetching and submodule retrieval should also use the same option.

## Contributor Checklist:


p12tic (approved):
Thanks!

PR #7747, opened 2024-07-03
Fix #7746

### LINKED ISSUE #7746: 4.0.0 reverse proxy problem
I've been running 3.11.5 with no issues, on Fedora 40 reverse proxied behind Apache, both http and ws.

I'm the Fedora package maintainer, and in the process of testing 4.0.0 to upgrade, I ran into a problem where the javascript UI elements failed to load. 

The browser debugging console revealed that several urls were incorrect.

Example: If my buildbot instance is at https://example.com/buildbot, the UI is trying to load https://example.com/plugins/waterfall_view/scripts.js, which should actually be https://example.com/buildbot/plugins/waterfall_view/scripts.js

When I create a ssh tunnel to the server and connect via http://localhost:port, the UI works as expected.

I haven't made any changes to where files are placed with the RPMs in 4.0.0 vs 3.11.5. I did use the github source tarballs since they're not on pypi, but between that and the new react ui, that's the only change.

tdesveaux:
Tested with this nginx configuration:
```nginx
	upstream bb-main {
		server 127.0.0.1:8010;
		keepalive 2;
	}

		location /subpath/ {
			proxy_pass http://bb-main/;
		}

		location /subpath/ws {
			proxy_set_header Upgrade $http_upgrade;
			proxy_set_header Connection "upgrade";
			proxy_pass http://bb-main/ws/;
			# raise the proxy timeout for the websocket
			proxy_read_timeout 6000s;
		}
```

## Contributor Checklist:

* [n/a I think?] I have updated the unit tests

### DIFF
--- a/www/base/index.html
+++ b/www/base/index.html
@@ -5,8 +5,8 @@
     <meta name="viewport" content="initial-scale=1, minimum-scale=1, user-scalable=no, maximum-scale=1, width=device-width">
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Buildbot Web UI"/>
-    <link id="bbicon" href="/icon.png" rel="icon">
-    <link href="/icon.svg" title="Buildbot" rel="fluid-icon">
+    <link id="bbicon" href="icon.png" rel="icon">
+    <link href="icon.svg" title="Buildbot" rel="fluid-icon">
     <title>Buildbot</title>
     <!-- BUILDBOT_THEME_CSS_PLACEHOLDER_BEGIN -->
     <style id="bb-style">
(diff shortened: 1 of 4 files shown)


www/base/index.html
```diff
@@ -36,9 +34,6 @@
     <noscript>You need to enable JavaScript to view this website.</noscript>
     <div id="root"></div>
   </body>
-  <script src="/browser-warning.js"></script>
```
p12tic, line 39: This one should be left as is because if browser is too old then it is not guaranteed that react code will be able to run at all.

yan12125:
Some of them should be fixed by using relative paths, like how the old UI did: https://github.com/buildbot/buildbot/blob/v3.11.5/www/base/src/app/layout.jade

p12tic (approved):

PR #7749, opened 2024-07-04
Fail fast if yarn or yarnpkg is not present.

gl-yziquel:
I tried to build buildbot on a "blank slate" machine with not much installed. In particular, with no yarn install. I noticed a failure with an "install" instruction. The YARN makefile variable was in fact set to an empty string and what should have been "yarn install" became interpolated into "install". This PR adds a check to make if "fail fast" with an error message that is more palatable that a badly interpolated "install" instruction in some shell scripting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,10 @@ PIP ?= $(ROOT_DIR)/$(VENV_NAME)/$(VENV_BIN_DIR)/pip
 VENV_PYTHON ?= $(ROOT_DIR)/$(VENV_NAME)/$(VENV_BIN_DIR)/python
 YARN := $(shell which yarnpkg || which yarn)
 
+ifeq ($(YARN),)
+  $(error "yarnpkg or yarn is not installed")
+endif
+
 WWW_PKGS := www/base www/console_view www/grid_view www/waterfall_view www/wsgi_dashboards www/badges
 WWW_EX_PKGS := www/nestedexample
 WWW_DEP_PKGS := www/plugin_support www/data-module www/ui


p12tic:
I think it would be better to add this to the rules that failed. It is possible to checkout buildbot to run unit tests, for which yarn is not needed.

gl-yziquel:
> I think it would be better to add this to the rules that failed.

I'm sorry, but I do not understand. What "rules" ?

> It is possible to checkout buildbot to run unit tests, for which yarn is not needed.

M'kay. How do I get that fail fast for just the pieces of code that need yarn ? It's for maintainance purposes on my box: I build buildbot from source, and hate hunting down cryptic error messages in build failures. In my personal opinion, it is the build system's job to check, in one way or another, for the presence of dependencies it needs.

If that check is in the wrong place, I'm fine with it. But I don't like empty variables being interpolated when they should be resolved to something that is a build dependency like yarn is.

p12tic:
> I'm sorry, but I do not understand. What "rules" ?

I meant rules in the makefile. E.g. this thing:

```
ruff:
	ruff --fix .
```

is called a rule. See e.g. https://www.gnu.org/software/make/manual/html_node/Rule-Example.html.

What is your "install" instruction? I just now realized there's no `make install`. What command do you call? Is it `make frontend`?

gl-yziquel:
> > I'm sorry, but I do not understand. What "rules" ?
> 
> I meant rules in the makefile. E.g. this thing:

Ah. Sorry. I use `just` nowadays instead of Makefiles, and the word "rules" has kind of slipped away from my mind.

As to the command I use, it is `make tarballs`. I try to package source code I install as much as possible. In fact, I have the feeling I've been reimplement some kind of toy `guix` tooling. So, yeah, I build the whole stuff as much as possible.

p12tic:
I think what is needed is a phony rule `check_for_yarn` which would fail if yarn is not present. This rule would then be added as a dependency to any rule which requires yarn. I hope make would be intelligent enough to run `check_for_yarn` only once for whole dependency graph.

gl-yziquel:
I'll keep that in mind. I don't have much time for that. So feel free to close this PR.

p12tic:
Sounds good. Thanks for the PR!

PR #7750, opened 2024-07-05
oauth2: better error if client credentials are incorrect

Mic92:
Instead of trying to access a non-existing token it's better to show the error to the user:
https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/

## Contributor Checklist:


p12tic:
Added newsfragment.

PR #7757, opened 2024-07-07
requirements: Upgrade buildbot-www to 4.0.0

p12tic:
This is needed for the unit tests to use correct www format. 3.x was AnularJS, 4.0 is React.


p12tic:
Included into https://github.com/buildbot/buildbot/pull/7756.

PR #7761, opened 2024-07-07
Fix issues in "Scheduled biweekly dependency update for week 26"

pmisik:
This PR is based on https://github.com/buildbot/buildbot/pull/7745 with these modification(s):
- updates `mypy-zope` from `1.0.4` to `1.0.5`
- fixes pylint issue about unreachable code

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/process/test_buildrequest.py
+++ b/master/buildbot/test/unit/process/test_buildrequest.py
@@ -66,6 +66,7 @@ class TestBuildRequestCollapser(TestReactorMixin, unittest.TestCase):
         def collapseRequests_fn(master, builder, brdict1, brdict2):
             # Allow all requests
             self.fail("Should never be called")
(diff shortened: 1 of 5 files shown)


p12tic:
Looks like only a small pylint-related fix is left.

pmisik:
> Looks like only a small pylint-related fix is left.

Should be fixed now.
I've chosen to suppress pylint warnings. 
If you rather prefer remove those line please let me know.

p12tic:
>  If you rather prefer remove those lines (see last two commits) please let me know.

worker/buildbot_worker/commands/fs.py needs migration to defer.inlineCallbacks

Lines master/buildbot/test/unit/process/test_buildrequest.py can indeed be removed

Feel free to create a separate PR for these issues, I think this PR does not need to be blocked for this.

Thanks!

PR #7764, opened 2024-07-08
waterfall_view: Fix waterfall squashed to a pixel high

jvivenot:
There was a confusion between builder ids and array indices.

In cases where no builder id matches an array-of-builders index, waterfall would fail to determine the minTime/maxTime range, and result in an empty (one pixel high) waterfall view.

As an example, the error occurs with a brand new master and a single builder (builderid = 1, but index = 0):

Before the fix :
Capture d’écran 2024-07-09 à 01 01 58

After the fix :
Capture d’écran 2024-07-09 à 01 01 26

## Contributor Checklist:

* I have NOT updated the unit tests. Tests are already giving proper builder ids ([1]) and not indices.
* I have NOT updated the appropriate documentation. Nothing relevant to document.


jvivenot:
To be more precise, after further tests, I noticed that the issue can appear in two different ways : 
* if no builderid happens to match any index, the whole view is squashed to a single pixel
* if some builderids happen to match some indices, the time range of the view is limited to the time range of those builders matched by chance. The view will fail to show builds outside this time frame

p12tic (approved):
Thanks!

PR #7766, opened 2024-07-09
waterfall_view: Allow float scaling_waterfall setting

jvivenot:
A float instead of integer for scaling_waterfall setting can allow for a more condensed view by allowing values below 1.

This also fixes a bug where one would hit the "zoom out" button, causing a below-1 scaling factor, then move to the setting page and back to the waterfall page, causing a squashed waterfall with a scaling factor of 0.

As an example, our buildbot install uses a waterfall view with a default scaling_waterfall factor of 0.05, showing a couple days build history among 50 or so builders.

## Contributor Checklist:

* I have *not* updated the unit tests. Nothing relevant.
* I have *not* updated the appropriate documentation. Nothing relevant

### DIFF
--- a/newsfragments/www-waterfall-scaling.change
+++ b/newsfragments/www-waterfall-scaling.change
@@ -1,0 +1,1 @@
+waterfall_view: Allow floating scaling_waterfall setting (for a more condensed view)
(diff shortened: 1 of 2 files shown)


p12tic (approved):
Thanks!

PR #7767, opened 2024-07-09
git: fetch without -t and use tags=True for tags

grooverdan:
The git fetch included -t(ags) by default. The effect of this is with shallowClone the fetch time is drasticly increased.

So the -t is removed and the Tags=True is required for the fetch to include tags.

## Contributor Checklist:


p12tic (approved):
Thanks!

PR #7777, opened 2024-07-10
Removed useless logs `fatal Exception on DB: log with slug ... already exists in this step`

slydiman:
We faced a lot of useless messages in logs like
```
Got fatal Exception on DB
builtins.KeyError: "log with slug ''some_name'' already exists in this step"
```
This patch removes them.


p12tic (changes requested):
Looks good, thanks.

Please run `ruff format master` and add a newsfragment.

slydiman:
> Please run `ruff format master` and add a newsfragment.

Done.

Issue #7759, opened 2024-07-07
4.0.1 backports


p12tic:
This issue will collect PRs that need to be backported to 4.0.x branch for 4.0.1 release.

p12tic:
https://github.com/buildbot/buildbot/pull/7755

p12tic:
https://github.com/buildbot/buildbot/pull/7758, https://github.com/buildbot/buildbot/pull/7756

p12tic:
https://github.com/buildbot/buildbot/pull/7769

p12tic:
https://github.com/buildbot/buildbot/pull/7768

p12tic:
These three need code changes to be duplicated across the duplicated www components (due to migration from react to regular naming in 4.0.x branch) https://github.com/buildbot/buildbot/pull/7766 https://github.com/buildbot/buildbot/pull/7765 https://github.com/buildbot/buildbot/pull/7764

p12tic:
https://github.com/buildbot/buildbot/pull/7747

p12tic:
https://github.com/buildbot/buildbot/pull/7777

mokibit:
https://github.com/buildbot/buildbot/pull/7747 backported by https://github.com/buildbot/buildbot/pull/7780.

mokibit:
https://github.com/buildbot/buildbot/pull/7755 backported by https://github.com/buildbot/buildbot/pull/7782.

mokibit:
https://github.com/buildbot/buildbot/pull/7758 backported by https://github.com/buildbot/buildbot/pull/7783.

mokibit:
https://github.com/buildbot/buildbot/pull/7756 backported by https://github.com/buildbot/buildbot/pull/7784.

mokibit:
https://github.com/buildbot/buildbot/pull/7769 backported by https://github.com/buildbot/buildbot/pull/7786.

mokibit:
https://github.com/buildbot/buildbot/pull/7768 backported by https://github.com/buildbot/buildbot/pull/7787.

mokibit:
https://github.com/buildbot/buildbot/pull/7766 backported by https://github.com/buildbot/buildbot/pull/7788.

mokibit:
https://github.com/buildbot/buildbot/pull/7765 backported by https://github.com/buildbot/buildbot/pull/7792.

mokibit:
[...]

PR #7780, opened 2024-07-11
4.0 backports:  Fix #7746

### LINKED ISSUE #7746: 4.0.0 reverse proxy problem

mokibit:
This PR backports https://github.com/buildbot/buildbot/pull/7747.
PR is a partial fix for https://github.com/buildbot/buildbot/issues/7759.

### DIFF
--- a/www/base/index.html
+++ b/www/base/index.html
@@ -5,8 +5,8 @@
     <meta name="viewport" content="initial-scale=1, minimum-scale=1, user-scalable=no, maximum-scale=1, width=device-width">
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Buildbot Web UI"/>
-    <link id="bbicon" href="/icon.png" rel="icon">
-    <link href="/icon.svg" title="Buildbot" rel="fluid-icon">
+    <link id="bbicon" href="icon.png" rel="icon">
+    <link href="icon.svg" title="Buildbot" rel="fluid-icon">
     <title>Buildbot</title>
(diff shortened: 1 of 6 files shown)


www/base/index.html
```diff
@@ -5,8 +5,8 @@
     <meta name="viewport" content="initial-scale=1, minimum-scale=1, user-scalable=no, maximum-scale=1, width=device-width">
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Buildbot Web UI"/>
-    <link id="bbicon" href="/icon.png" rel="icon">
```
p12tic, line 8: Also apply the same changes to base_react.

PR #7790, opened 2024-07-11
www: Fix loading of plugin settings from master

jvivenot:
Settings that are set via ui_default_config currently fail to be applied because `applyBuildbotConfig` is called before the plugin's javascript is added, and therefore before the plugin registers its settings group.

This results in errors such as :
`bad setting name Waterfall.show_old_builders: group does not exist`

In this PR, I fixed this issue by deferring the load of global and local settings for a specific group to the moment the group is added to GlobalSettings (i.e. when each plugin registers itself). 

I suppose there could be other ways to fix this. For example :
* by waiting for all plugins to be registered before global and local settings are applied once. (`pluginScript.onload` ?)
* by applying ALL global and local settings on every single plugin registration

Please let me know if the solution I implemented is unsatisfactory.

## Contributor Checklist:

* I have updated the unit tests and added a new one to test the group-specific load.
* I have not updated the appropriate documentation. Nothing relevant

### DIFF
--- a/newsfragments/www-plugin-settings-load.bugfix
+++ b/newsfragments/www-plugin-settings-load.bugfix
@@ -1,0 +1,1 @@
+Fix loading of plugins settings (e.g. from master's ui_default_config)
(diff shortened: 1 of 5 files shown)


p12tic (approved):

p12tic:
Looks great. A good addition would be to implement a e2e test (smokes-react/ directory) to protect against regressions.

PR #7806, opened 2024-07-13
[4.0] meta react packages

tdesveaux:
This remove the duplicate code for www plugins and will thus make it easier to backport fixes.

buildbot-react-x package are now 'meta' packages. Meaning, they have a dependency on the non-react package, and setup an entrypoint which target the non-react package implementation.

## Contributor Checklist:


p12tic:
I'm sorry, I can't merge this PR.

I agree that it's cleaner solution. However the current approach of crudely duplicating the code was a conscious decision. The code path of using the old plugin names will be significantly less tested, if at all. The least we want is upgrade path silently breaking and users wasting their time thinking that their configuration is wrong.

Even though the duplication commit looks like a copy, actually the code is identical to what was before the rename to reduce risk of something accidentally breaking.

The downsides of this approach are bounded. The backport PRs need changes to be duplicated - small cost. I look like stupid developer - small cost. I need to ensure in backport reviews that all changes are duplicated - small cost.

However the downsides of accidentally breaking upgrades are basically unbounded if we consider how many people may spend time scratching their heads when upgrading their Buildbots. And if no one cares to fill a bug report we won't even see it.

Sorry again. The PR looks nice and it's a pity that I can't accept it. If there were more developers actively working on Buildbot then I would do otherwise.

tdesveaux:
No worries! I was mostly curious if that would work and did not confirm you'd be interested.

I understand your reasoning, and don't think it's stupid, it's a valid trade-off.

PR #7809, opened 2024-07-13
master/reporters/github: render token for each request

Mic92:
Github token can expire i.e. if an GitHub apps are used the token needs to be renewed every hour from the installation token. By moving the renderSecrets just when the requests happens, we provide an interface for users to provide a new refresh GITHUB_TOKEN.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/reporters/github.py
+++ b/master/buildbot/reporters/github.py
@@ -15,6 +15,7 @@
 
 
 import re
+from typing import Dict
 
 from twisted.internet import defer
 from twisted.python import log
@@ -70,7 +71,7 @@ class GitHubStatusPush(ReporterBase):
         generators=None,
         **kwargs,
     ):
-        token = yield self.renderSecrets(token)
+        self.token = token
         self.debug = debug
         self.verify = verify
         self.verbose = verbose
@@ -89,7 +90,7 @@ class GitHubStatusPush(ReporterBase):
         self._http = yield httpclientservice.HTTPClientService.getService(
             self.master,
             baseURL,
-            headers={'Authorization': 'token ' + token, 'User-Agent': 'Buildbot'},
+            headers={'User-Agent': 'Buildbot'},
             debug=self.debug,
             verify=self.verify,
         )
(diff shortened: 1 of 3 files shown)


Mic92:
We switched from GITHUB_TOKEN to github apps, which has security advantages but it is also more convenient to maintain in a bigger organization. However this also required us to make github token refreshable. I choose the secret render function here in the hopes it would also benefit other users. i.e. https://developer.hashicorp.com/vault/docs/sync/github can write out a GITHUB_TOKEN with vault agent.

If you find this interface not suitable. We also have an alternative patch that replaces token with a Callable (see: https://github.com/nix-community/buildbot-nix/blob/ffcb6b8869a540c4784724425495f6474b6f976c/nix/0001-master-reporters-github-render-token-for-each-reques.patch)

master/buildbot/reporters/github.py
```diff
@@ -241,6 +259,7 @@ def sendMessage(self, reports):
                     context=context,
                     issue=issue,
                     description=description,
+                    props=props,
```
p12tic, line 262: At this point props is required to be not None. We can make this argument not optional.

master/buildbot/reporters/github.py
```diff
@@ -109,6 +111,15 @@ def _create_default_generators(self):
             ),
         ]
 
+    @defer.inlineCallbacks
+    def _get_auth_header(self, props: Union[Properties, None]) -> dict[str, str]:
+        if props:
+            token = yield props.render(self.token)
+        else:
+            token = yield self.renderSecrets(self.token)
```
p12tic, line 119: I think it's enough to just do `self.renderSecrets(self.token)`. A secret is just a special renderable that gathers data from secret providers. So a renderable that refreshes GITHUB_TOKEN is should work fine here.

Am I missing anything?

Mic92: The normal render function doesn't have access to the project we are about to build. The idea is that we can use the GITHUB_TOKEN for the project rather than a global GITHUB_TOKEN. I am open for other suggestions.

master/buildbot/reporters/github.py
Mic92: Done.

p12tic (approved):

p12tic:
Rebased.

ThomasDevoogdt:
Is there any reason why this is not part of v4.0.2?

PR #7810, opened 2024-07-14
Drop future dependency

yan12125:
Usage of the future library was already dropped in https://github.com/buildbot/buildbot/pull/6997

## Contributor Checklist:


p12tic (approved):

yan12125:
CI logs do not show `future` brought in as a transitive dependency, so removing it from requirements is safe.

Also, test failures look irrelevant.

PR #7812, opened 2024-07-14
smokes-react: Add an e2e test to prevent regression on #7790

jvivenot:
As suggested by @p12tic in #7790 , this adds an end-to-end test to prevent regression on the loading of ui_default_config settings for plugins such as Waterfall.

## Contributor Checklist:

* I have added an e2e test.
* [ ] _I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)_
* [ ] _I have updated the appropriate documentation_

### DIFF
--- a/smokes-react/master.cfg
+++ b/smokes-react/master.cfg
@@ -124,7 +124,10 @@ c['www'] = {
         "console_view": {},
         "grid_view": {},
     },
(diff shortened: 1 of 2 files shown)


p12tic (approved):
Thanks a lot for getting back to this.

PR #7813, opened 2024-07-14
config: ensure buildbotURL and titleURL have a trailing '/'

tdesveaux:
Since it was required in the doc but without config checks, I found it easier to ensure it there.

## Contributor Checklist:

### DIFF
--- a/master/docs/manual/configuration/global.rst
+++ b/master/docs/manual/configuration/global.rst
@@ -306,12 +306,11 @@ Three basic settings describe the buildmaster in status reports:
 
 :bb:cfg:`title` is a short string that will appear at the top of this buildbot installation's home page (linked to the :bb:cfg:`titleURL`).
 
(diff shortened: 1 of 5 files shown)


master/docs/manual/configuration/global.rst
```diff
@@ -306,12 +306,11 @@ Three basic settings describe the buildmaster in status reports:
 
 :bb:cfg:`title` is a short string that will appear at the top of this buildbot installation's home page (linked to the :bb:cfg:`titleURL`).
 
-:bb:cfg:`titleURL` is a URL string that must end with a slash (``/``).
```
p12tic, line 309: I think it makes sense to keep this, otherwise users will all make an error of first trying with URL without a slash.

p12tic: Ah, the slash is added automatically, sorry.

p12tic (approved):

Issue #7774, opened 2024-07-10
Spaces are trimmed when showing shell logs


vit9696:
Consider a `ShellCommand` that prints something like this:

```
[  1/ 10] Test1                      \x1b[32;1mOK\x1b[0m
[  2/ 10] Test2                      \x1b[32;1mOK\x1b[0m
[  3/ 10] Test3                      \x1b[32;1mOK\x1b[0m
…
[ 10/ 10] Test10                     \x1b[32;1mOK\x1b[0m
```

New React UI will print:

```
[ 1/ 10] Test1 OK
[ 2/ 10] Test2 OK
[ 3/ 10] Test2 OK
…
[ 10/ 10] Test10 OK
```

While the expected output is:

```
[  1/ 10] Test1                       OK
[  2/ 10] Test2                       OK
[  3/ 10] Test3                       OK
…
[ 10/ 10] Test10                      OK
```

The issue is caused by using `span` in formatting without defining `white-space` policy. If one adds `white-space: pre;` it works correctly again.

mokibit:
@p12tic, 
issue was fixed in https://github.com/buildbot/buildbot/pull/7816.

vit9696:
This one is indeed fixed in 4.0.2. Thank you.

PR #7816, opened 2024-07-15
www: Do not trim spaces in logs

mokibit:
This PR does not allow trimming spaces in logs as original log lines should be preserved.
PR fixes issue https://github.com/buildbot/buildbot/issues/7774.


p12tic (approved):
Thanks!

PR #7861, opened 2024-07-29
Enable pyugprade in ruff

Mic92:
## Remove this paragraph

And especially:

## Contributor Checklist:

### DIFF
--- a/common/download_release.py
+++ b/common/download_release.py
@@ -9,9 +9,9 @@ import yaml
 
 def download(session, url, fn):
     if os.path.exists(fn):
-        print('Removing old file {}'.format(fn))
+        print(f'Removing old file {fn}')
         os.unlink(fn)
-    print('Downloading {} from {}'.format(fn, url))
+    print(f'Downloading {fn} from {url}')
     with open(fn, 'wb') as f:
         r = session.get(url, stream=True)
         r.raise_for_status()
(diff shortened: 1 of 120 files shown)


p12tic:
I would split the upgrades into multiple commits each with the same type of upgrade. E.g. fix exception types, remove explicit encoding and so on, potentially with a short explanation, why the upgrade can be done. This way, if the PR causes any problems down the road, it will be easier to bisect.

p12tic:
This approach would also allow to bisect the current problem with the tests.

Mic92:
Here is the script that I used:

```
#!/usr/bin/env bash

# https://docs.astral.sh/ruff/rules/#pyupgrade-up
declare -A rules=(
  ["UP001"]=useless-metaclass-type,
  ["UP003"]=type-of-primitive,
  ["UP004"]=useless-object-inheritance,
  ["UP005"]=deprecated-unittest-alias,
  ["UP006"]=non-pep585-annotation,
  ["UP007"]=non-pep604-annotation,
  ["UP008"]=super-call-with-parameters,
  ["UP009"]=utf8-encoding-declaration,
  ["UP010"]=unnecessary-future-import,
  ["UP011"]=lru-cache-without-parameters,
  ["UP012"]=unnecessary-encode-utf8,
  ["UP013"]=convert-typed-dict-functional-to-class,
  ["UP014"]=convert-named-tuple-functional-to-class,
  ["UP015"]=redundant-open-modes,
  ["UP017"]=datetime-timezone-utc,
  ["UP018"]=native-literals,
  ["UP019"]=typing-text-str-alias,
  ["UP020"]=open-alias,
  ["UP021"]=replace-universal-newlines,
  ["UP022"]=replace-stdout-stderr,
  ["UP023"]=deprecated-c-element-tree,
  ["UP024"]=os-error-alias,
  ["UP025"]=unicode-kind-prefix,
  ["UP026"]=deprecated-mock-import,
  ["UP028"]=yield-in-for-loop,
  ["UP029"]=unnecessary-builtin-import,
  ["UP030"]=format-literals,
  ["UP031"]=printf-string-formatting,
  ["UP032"]=f-string,
  ["UP033"]=lru-cache-with-maxsize-none,
  ["UP034"]=extraneous-parentheses,
  ["UP035"]=deprecated-import,
  ["UP036"]=outdated-version-block,
  ["UP037"]=quoted-annotation,
  ["UP038"]=non-pep604-isinstance,
  ["UP039"]=unnecessary-class-parentheses,
  ["UP040"]=non-pep695-type-alias,
  ["UP041"]=timeout-error-alias,
)

for rule in "${!rules[@]}"; do
  ruff check --select "$rule" --fix --unsafe-fixes .
  ruff format .
  if git diff --quiet; then
    echo "No changes needed for rule $rule"
    continue
  fi
  git commit -am "ruff: apply rule $rule

See https://docs.astral.sh/ruff/rules/${rules[$rule]}
"
done
```

p12tic:
This is great, thanks

PR #7864, opened 2024-07-30
deps: bump twisted min version to 21.2.0

tdesveaux:
Since Buildbot requires Python3.8 at least,
it makes sense to require the first version of twisted that supports it.

ref: https://github.com/twisted/twisted/blob/twisted-21.2.0/NEWS.rst "Python 3.8 is now tested and supported. (#9955)"

## Contributor Checklist:


p12tic (approved):
Thanks!

PR #7887, opened 2024-08-02
www: Some improvements to V2RootResource.renderRest

tdesveaux:
Content-Length header was not correctly calculated (based on unicode str length rather than bytes, cf. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length).

Made some other chances while at it:
- Include Content-Length on non-HEAD requests
- Use JSONEncoder to chunk write the response

## Contributor Checklist:


p12tic:
I think this needs a release note. Other than that, looks great, thanks!

p12tic (approved):

PR #7892, opened 2024-08-03
Declare support for Pypy

p12tic:
Likely pypy worked with previous releases already. Only few adjustments in the tests were needed to pass whole testsuite.

## Contributor Checklist:


p12tic:
Seems that pypy is becoming dead :/

PR #7912, opened 2024-08-08
buildbot_worker/runprocess.py: Use windows Job API

mifo123:
If process dies, there can still be sub-processes which can not be killed (on windows platform). This can cause problems with tests running on buildbot.

Attach spawned process to `JobObject` so if process dies, we can still kill all its child processes. Now we do not have to kill process (which does not have to exists) but we terminate whole JobObject, where are all processes of the main process (included).

## Contributor Checklist:

### DIFF
--- a/master/docs/manual/configuration/steps/shell_command.rst
+++ b/master/docs/manual/configuration/steps/shell_command.rst
@@ -26,6 +26,13 @@ You can also specify the command with a single string, in which case the string 
 On Windows, commands are run via ``cmd.exe /c`` which works well.
 However, if you're running a batch file, the error level does not get propagated correctly unless you add 'call' before your batch file's name: ``cmd=['call', 'myfile.bat', ...]``.
 
+``ShellCommand`` includes all sub-processes created by the command in ``JobObject``. This ensures
+that all child processes are managed together with the parent process. When the main command is
+terminated, all sub-processes are also terminated automatically, preventing any orphaned processes.
+This enhancement aligns the behavior of Windows systems with POSIX systems, where similar process
+management has been in place.
+
+
 The :bb:step:`ShellCommand` arguments are:
 
 ``command``
--- a/worker/buildbot_worker/runprocess.py
+++ b/worker/buildbot_worker/runprocess.py
@@ -47,6 +47,10 @@ from buildbot_worker.exceptions import AbandonChain
 
 if runtime.platformType == 'posix':
     from twisted.internet.process import Process
+if runtime.platformType == 'win32':
+    import win32api
+    import win32con
+    import win32job
 
 
 def win32_batch_quote(cmd_list, unicode_encoding='utf-8'):
@@ -396,6 +400,7 @@ class RunProcess(object):
         self.killTimer = None
         self.keepStdout = keepStdout
         self.keepStderr = keepStderr
+        self.job_object = None
 
         assert usePTY in (
             True,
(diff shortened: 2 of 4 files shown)


worker/buildbot_worker/runprocess.py
```diff
@@ -748,6 +780,11 @@ def cleanUp(self, hit):
             # blocks here until the process has terminated, while closing
             # stderr. This is weird.
             self.pp.transport.loseConnection()
+        elif runtime.platformType == 'win32':
+            if hasattr(self, 'job_object'):
+                pass
```
p12tic, line 785: Probably should be deleted?

worker/buildbot_worker/runprocess.py
```diff
@@ -748,6 +780,11 @@ def cleanUp(self, hit):
             # blocks here until the process has terminated, while closing
             # stderr. This is weird.
             self.pp.transport.loseConnection()
+        elif runtime.platformType == 'win32':
+            if hasattr(self, 'job_object'):
```
p12tic, line 784: Better set job_object to None initially and then check for that, instead of using hasattr.

worker/buildbot_worker/runprocess.py
```diff
@@ -810,15 +847,23 @@ def sendSig(self, interruptSignal):
 
     def _taskkill(self, pid, force):
         try:
-            if force:
-                cmd = "TASKKILL /F /PID {0} /T".format(pid)
-            else:
-                cmd = "TASKKILL /PID {0} /T".format(pid)
-
+            cmd = (
```
p12tic, line 850: Previous initialization was easier to understand.

worker/buildbot_worker/runprocess.py
```diff
@@ -810,15 +847,23 @@ def sendSig(self, interruptSignal):
...
+                "TASKKILL /F /PID {0} /T".format(pid)
+                if force
+                else "TASKKILL /PID {0} /T".format(pid)
+            )
+            pr_info = win32job.QueryInformationJobObject(
+                self.job_object, win32job.JobObjectBasicProcessIdList
+            )
+            if runtime.platformType == 'win32' and hasattr(self, 'job_object'):
```
p12tic, line 858: QueryInformationJobObject and self.job_object are used couple of lines above without any checks.

worker/buildbot_worker/runprocess.py
mifo123: Sorry, I totally forgot about this.

worker/buildbot_worker/runprocess.py
mifo123: Sorry, I totally forgot about this.

master/docs/manual/configuration/steps/shell_command.rst
```diff
@@ -26,6 +26,9 @@ You can also specify the command with a single string, in which case the string
 On Windows, commands are run via ``cmd.exe /c`` which works well.
 However, if you're running a batch file, the error level does not get propagated correctly unless you add 'call' before your batch file's name: ``cmd=['call', 'myfile.bat', ...]``.
 
+``ShellCommand`` includes all sub-processes created by the command in ``JobObject``. This ensures that all child processes are managed together with the parent process. When the main command is terminated, all sub-processes are also terminated automatically, preventing any orphaned processes. This enhancement aligns the behavior of Windows systems with POSIX systems, where similar process management has been in place.
```
p12tic, line 29: Please wrap at 100 chars per line. I do understand that the surrounding text does not use this convention.

worker/buildbot_worker/runprocess.py
```diff
@@ -814,11 +851,20 @@ def _taskkill(self, pid, force):
                 cmd = "TASKKILL /F /PID {0} /T".format(pid)
             else:
                 cmd = "TASKKILL /PID {0} /T".format(pid)
-
+            if runtime.platformType == 'win32' and self.job_object is not None:
```
p12tic, line 854: _taskkill can only be called on win32 platform. Maybe it's worth renaming _taskkill to _win32_taskkill and then removing the above check.

worker/buildbot_worker/runprocess.py
```diff
@@ -814,11 +851,20 @@ def _taskkill(self, pid, force):
...
+                pr_info = win32job.QueryInformationJobObject(
+                    self.job_object, win32job.JobObjectBasicProcessIdList
+                )
+                if force or len(pr_info) < 2:
+                    win32job.TerminateJobObject(self.job_object, 1)
+            self.log_msg("terminating job object with pids {0}".format(str(pr_info)))
```
p12tic, line 860: Let's use f-strings for new code.

worker/buildbot_worker/runprocess.py
mifo123: Ok, so I renamed the method and remove repetitive check for platformtype.

p12tic (approved):

p12tic:
LGTM, please squash the fixup commits.

PR #7924, opened 2024-08-15
buildbot_worker: Add `maxLines` parameter to shell command.

mifo123:
- Introduced `maxLines` parameter to shell command and passed it to to worker runprocess.
 - Implemented logic to count lines and terminate process if it exceeds `maxLines`.
 - Added failure reason to summary output.
 - Updated tests to cover `maxLines` parameter functionality.
 - Documented the `maxLines` parameter in the project documentation.

## Contributor Checklist:


worker/buildbot_worker/runprocess.py
```diff
@@ -754,6 +760,21 @@ def doMaxTimeout(self):
         self.send_update([("failure_reason", "timeout")])
         self.kill(msg)
 
+    def _checkMaxLines(self, data):
+        self.line_count += len(re.findall(r"\r\n|\r|\n", data))
```
p12tic, line 764: I think it makes sense to not count lines if the limit is not set.

master/buildbot/process/buildstep.py
```diff
@@ -281,6 +281,7 @@ def __init__(self, **kwargs):
         self._acquiringLocks = []
         self.stopped = False
         self.timed_out = False
+        self.max_lines_reach = False
```
p12tic, line 284: max_lines_reached would be more readable name

p12tic:
Could you please tell why you needed this parameter? Is it to protect master from too large number of lines from broken commands on worker?

p12tic:
For new code please use `snake_case` convention even if it does not match surrounding code. Eventually whole buildbot will use `snake_case`.

mifo123:
> Could you please tell why you needed this parameter? Is it to protect master from too large number of lines from broken commands on worker?

Yes, that's right. In the company where I work, we often encounter the problem of a faulty command going haywire and sending millions of lines in an infinite loop.

p12tic:
Nice. 

I noticed a thing I missed. Please see changes to command_version variable in https://github.com/buildbot/buildbot/commit/76fbf79d10332b0b3e87f616839a8546285a5fcc#diff-ae09855c6e4135391a59ce95753b1652f7b5c346e81465241299b1011de1881d.

We need to do something similar here too so that master knows capabilities of the worker. Just 2 lines of changes.

Also, in case you publish `fixup!` commits in order for me to see what has changed since previous PR iteration, this is not necessary because github UI shows me changes between force pushes. This way I won't need to wait for you to squash the commits and merge the PR as soon as it is good.

mifo123:
> Nice.
> 
> I noticed a thing I missed. Please see changes to command_version variable in [76fbf79#diff-ae09855c6e4135391a59ce95753b1652f7b5c346e81465241299b1011de1881d](https://github.com/buildbot/buildbot/commit/76fbf79d10332b0b3e87f616839a8546285a5fcc#diff-ae09855c6e4135391a59ce95753b1652f7b5c346e81465241299b1011de1881d).
> 
> We need to do something similar here too so that master knows capabilities of the worker. Just 2 lines of changes.
> 
> Also, in case you publish `fixup!` commits in order for me to see what has changed since previous PR iteration, this is not necessary because github UI shows me changes between force pushes. If there are no such commits I won't need to wait for you to squash the commits and merge the PR as soon as it is good.

I hope that I've understood correctly.

Also, i rebased so there will be no fixup. Thank you for mention about this functionality

pmisik:
> > Could you please tell why you needed this parameter? Is it to protect master from too large number of lines from broken commands on worker?
> 
> Yes, that's right. In the company where I work, we often encounter the problem of a faulty command going haywire and sending millions of lines in an infinite loop.

I think Buildbot should be distributed with rationally security limits.
Maybe in future releases it would be worth considering to change the unlimited number of lines of logs to some concrete value (e.g. 1000000 lines). This restriction actually protects several master resources such as CPU, memory, network traffic and disk space of the database.

When this problem occurred to us, we were "lucky" because it consumed memory on our master very quickly and the master crashed.
This can happen to anyone using Buildbot.

RazvanLiviuVarzaru:
Hi!

This is indeed very useful. However, I’ve observed that the primary factor affecting performance is the number of updates per second received by the master through the remote_update method.

Even in a simpler setup, such as a single builder echoing 1,000,000 lines to a single master, the CPU remains at 100% most of the time. In a real-world scenario, you typically don’t have clients producing logs that quickly, but having multiple builders adds to the number of updates per second, keeping the master busy.

In a multi-master setup, when this happens, it often slows down so much that it loses the connection to CrossBar, and restarting the service becomes the only option, which is usually inconvenient for users.

I hope that in future versions, you’ll have the opportunity to investigate a throttling or rate-limiting mechanism.

Issue #5091, opened 2019-11-25
Announcement bar


pmisik:
Hi

I’m not sure if this should be default Buildbot functionality or it should be implemented as plugin.
So I filled issue to consult what is proper design from Buildbot architecture/maintainers perspective.

We use bleeding edge Buildbot slightly modified with custom_templates in company I work for.
From time to time we face planed/unplanned IT infrastructure issues in some region.
At the moment we try to notify developers by email if there is some infrastructure issue/outage.
I think it may be annoying for developers if they do not interact with build system at the moment of outage to receive “spam” email. 

I think it would be awesome if there would be possibility to display such information on the top of Buildbot page in such cases similar as there is div for incompatible browser.

What do you think about it and what is your suggestion to implement it?

Should it be “core feature” of Buildbot or plugin?

Thank You 

Palo

tardyp:
Hi, I think it is a great idea.

I even think we had that in the first version in dojo.js back in the days.
It should not be complicated to implement.

You can put it just behind connectionstatus component.
https://github.com/buildbot/buildbot/blob/b0875d0d017d25665511ad14f5d64887d4865cd8/www/base/src/app/index.jade#L7

Would you like to make a PR?

uglycoyote:
Our approach to this has been to use a custom plugin, a completely sideband thing implemented with Mongo/Python/Flask/Eve, but it would be cool to see something like this as a built-in part of buildbot.  

Ours a bit more informal than "announcements", it's more like "post-it notes" where any user is free to post a note and/or remove old ones, which helps with communication and avoiding multiple people looking in to the same problem.  (notes are often of the form "I'm investigating XYZ problem" with a name and timestamp).

PR #7948, opened 2024-08-22
[WIP] announcement bar creation

mifo123:
Add announcement bar with new endpoint '/messagesinfo' which will take *.txt file from specified folder c['messageInfoDir']. FE will shows all those messages.

Is this approach ok, or it should be done via some DB or any other way?

## Contributor Checklist:


mifo123:
Here are some screenshots of FE:

image

image

p12tic:
I have looked into this PR and need more time to think whether the current design is good. I agree that the feature makes sense.

pmisik:
Relates to #5091

mifo123:
Hi @p12tic, did you have time to look at this? Thanks.

PR #7980, opened 2024-09-02
Stop using twisted.internet.defer.returnValue

cjwatson:
Python 2 support was removed from the worker in buildbot 4.0.0, and `defer.returnValue` was only needed in Python 2; in Python 3, a simple `return` is fine.

`twisted.internet.defer.returnValue` is deprecated as of Twisted 24.7.0.

## Contributor Checklist:

### DIFF
--- a/common/validate.sh
+++ b/common/validate.sh
@@ -81,14 +81,6 @@ check_long_lines() {
 }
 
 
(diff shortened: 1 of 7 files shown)


p12tic:
Thanks!

p12tic (approved):

PR #7981, opened 2024-09-03
www/base: port tests to @testing-library/react

tdesveaux:
tests used react-test-renderer which has been deprecated: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-react-test-renderer

Ported to @testing-library recommended by React and was already in www/base dependencies

react-test-renderer is still used in www/ui for FixedSizeList.
Porting these tests is not 'trivial', and I'm unsure how to tackle this since it's a vendored dependency.

## Contributor Checklist:

* [x] I have updated the unit tests
* [n/a] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)
* [n/a] I have updated the appropriate documentation

Issue #7773, opened 2024-07-10
util.BooleanParameter checkbox is not checked in UI when default=True


vit9696:
For the following code snippet:

```py
    param_test = util.BooleanParameter(
        name='is_test',
        label="Enable test (uncheck for no test)",
        default=True,
    )
```

The form will not have `is_test` checkbox enabled when running the plan. It will be disabled and will result in `null` in the plan properties. This is a regression somewhere in late 3.x or 4.0. Possibly also caused by React view.

tdesveaux:
Same as #7772, was fixed by #7571, but not backported.

vit9696:
I have just upgraded to 4.0.2, yet the issue is still present.

tdesveaux:
https://github.com/buildbot/buildbot/pull/7832 backported the fix, but only applied it to `react-base` and not `base`.

@vit9696 using the `base_react` www plugin should fix the issue for you (you'll need to install the `buildbot-www-react` python package).

```python
c["www"] = {
    "port": 8010,
    "plugins": {"base_react": True},
}
```

vit9696:
* `base_react` plugin is deprecated, while we can migrate back to it, it will disappear sooner or later.
* `base_react` plugin does not work correctly either: when multiple entries are selected, clicking on one results in this entry being deselected, instead of becoming the only selected one. This violates the standard select behaviour: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select_multiple

pp1e:
Have the same problem on 4.0.2 version

vit9696:
Tested, still broken on 4.0.3.

vit9696:
Tested, works on 4.1.0. Most likely 4.0.x is still broken.

PR #7983, opened 2024-09-03
www/base: fix field choice string multiple selection

tdesveaux:
based on #7981 

[field choice usage](https://github.com/user-attachments/assets/7a0a8649-4ac1-43f4-88a7-85b3d950ab9e)

## Contributor Checklist:

### DIFF
--- a/www/base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
+++ b/www/base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
@@ -22,29 +22,45 @@ import {observer} from "mobx-react";
 import {FieldBase} from "./FieldBase";
 import {Form} from "react-bootstrap";
 import {Tooltip} from 'react-tooltip'
+import Select, { ActionMeta, PropsValue, SingleValue, MultiValue } from 'react-select';
+import CreatableSelect from 'react-select/creatable';
 
 type FieldChoiceStringProps = {
   field: ForceSchedulerFieldChoiceString;
   fieldsState: ForceBuildModalFieldsState;
 }
 
+interface SelectOption {
+  readonly value: string,
+  readonly label: string,
+}
+
+const ValueToSelectOption = (value: string): SelectOption => { return {value: value, label: value} };
+
 export const FieldChoiceString = observer(({field, fieldsState}: FieldChoiceStringProps) => {
   const state = fieldsState.fields.get(field.fullName)!;
+  if (field.multiple && !Array.isArray(state.value)) {
+    fieldsState.setValue(field.fullName, state.value ? [state.value] : []);
+  }
 
-  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    if (!field.multiple) {
-      fieldsState.setValue(field.fullName, event.target.value);
-    }
-    else {
-      const currentValue = new Set(state.value);
-      if (currentValue.has(event.target.value)) {
-        currentValue.delete(event.target.value);
-      }
-      else {
-        currentValue.add(event.target.value);
-      }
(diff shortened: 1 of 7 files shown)


newsfragments/www-fix-forcefield-choicestring-multiple-selection.bugfix
```diff
@@ -0,0 +1 @@
+Fix ``ChoiceStringParameter`` fields with ``multiple`` to match standard selection behaviour (would deselect a selected item on click, now unselect all except this item).
```
p12tic, line 1: Wouldn't this break multiple selections? The description could be improved a bit if that's not the case.

tdesveaux: I made the change in reaction to this comment: https://github.com/buildbot/buildbot/issues/7773#issuecomment-2273690076

Since it referenced w3school, I assumed it was a convention (I'm really not a frontend person, so I have no base knowledge here).

That's interesting that your link does not follow this, however, the widget has checkboxes and highlighting which provide correct feedback on this behavior to the user.
With the simple select, I think it should follow the basic standards.

Or, I can rework this to use the `react-select` component that you okayed [here](https://github.com/buildbot/buildbot/pull/7984#issuecomment-2377213812).

p12tic: Understood. The example on w3schools is I think less intuitive because the only way to select multiple options is to use Ctrl button. Most likely this does not work on touch-based interfaces.

I think a proper component such as `react-select` is better.

tdesveaux: react-select is a bit annoying to work with, but doable.
I added a video of usage.

Note that this also make it possible to combine `multiple` and `strict`.
Should I add a newsfragment? This limitation was not clearly documented before, not did it generate a config error, it'd just not display the text input box

p12tic (approved):

newsfragments/www-fix-forcefield-choicestring-multiple-selection.bugfix
p12tic: >  Should I add a newsfragment? This limitation was not clearly documented before, not did it generate a config error, it'd just not display the text input box

Yes, please do. It's still a user visible change even if the previous behavior wasn't documented.

www/base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
```diff
@@ -22,29 +22,47 @@ import {observer} from "mobx-react";
...
+import Select, { ActionMeta, PropsValue, SingleValue, MultiValue } from 'react-select';
+import CreatableSelect from 'react-select/creatable';
 
 type FieldChoiceStringProps = {
   field: ForceSchedulerFieldChoiceString;
   fieldsState: ForceBuildModalFieldsState;
 }
 
+interface SelectOption
+{
```
p12tic, line 34: I think the brace should be on the line above, like `type` a few lines above.

www/base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
```diff
@@ -22,29 +22,47 @@ import {observer} from "mobx-react";
...
+  }
+
+  const onChange = (
+    newValue: PropsValue<SelectOption>,
+    _actionMeta: ActionMeta<SelectOption>
+  ) => {
+    fieldsState.setValue(
+      field.fullName,
+      (
+        field.multiple ?
```
p12tic, line 54: I think something like:

```
condition ? b
                  : c;
```

would be more readable in this case. I was genuinely confused for a bit.

PR #7984, opened 2024-09-03
www: add worker bulk action modal

tdesveaux:
image

Simple select component is pretty bare bone and does not provide features parity with Angular version ([un]select-all, search).

I looked at some existing components for multi select, but they all seem not maintained:
- [react-select](https://github.com/JedWatson/react-select): (last release close to one year ago)
- [react-multi-select-component](https://github.com/hc-oss/react-multi-select-component): (last release 2 years ago)
- [multiselect-react-dropdown](https://github.com/srigar/multiselect-react-dropdown): (last release 2 years ago)

I found that bigger component libraries that contain a widget that would help are kept up to date.
Candidates:
- https://primereact.org/multiselect/
- https://mui.com/material-ui/react-select/#multiple-select
- https://coreui.io/react/docs/forms/multi-select/

These libraries should provide other components that could be used elsewhere in the UI and plugins.

Leaving the decision whether to add one and which one to Buildbot's maintainer team.
Waiting on feedback to un-draft this PR. This could be merged as-is as a first step.

## Contributor Checklist:

### DIFF
--- a/www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
+++ b/www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
@@ -1,0 +1,106 @@
+/*
+  This file is part of Buildbot.  Buildbot is free software: you can
+  redistribute it and/or modify it under the terms of the GNU General Public
+  License as published by the Free Software Foundation, version 2.
+
+  This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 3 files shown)


p12tic:
@tdesveaux In terms of other libraries, I think let's go with react-select. That one seems to be the most mature one. Incidentally it had a release just last week.

Additionally, it doesn't seem to be unmaintained, just a case of a mature product and some maintainers no longer active. See e.g. https://github.com/JedWatson/react-select/issues/5911#issuecomment-2158685812.

Issue #7852, opened 2024-07-24
ansi log escaping seems to break at line 500 and later


q66:
When you have a log with ansi colors in it, the escapes stop being interpreted at line 500 and after in some manner.

I found that when I saw our firefox build output tends to be kinda broken, e.g. like https://build.chimera-linux.org/#/builders/5/builds/232/steps/5/logs/pkg_contrib_firefox_128_0_2-r0 - you can see some of the sequences are not being interpreted, sometimes the whole output just becomes blue.

I tried launching a local buildmaster where I just cat'd a portion of the log file. I noticed that things kept looking broken until I removed enough lines that the first broken line started before line 500, at which point it started rendering correctly again. The content of the lines removed does not seem to matter.

This is with the new react UI. With the old angular UI we had the same breakage; however, it seemed to be less affected as some stuff that previously rendered okay no longer renders okay now, but that's likely a separate regression. I haven't yet tracked down a good sample for that.

q66:
this seems to be due to downloadOverscanRowCount=500 in `www/base/src/components/LogViewer/LogViewer.tsx`, but i have no idea what a proper fix would look like...

q66:
this does not seem to fully map from certain number; setting it to less than 2500 will cut off coloring in log viewer at the number, but setting it to 2500 and more will result in entire log being properly colored - i'm not sure why exactly that is, but maybe this information is helpful?

PR #7996, opened 2024-09-17
www: fix ansi color reset in log view

q66:
If we just return cssClasses, we are returning the previous instance which may be filled with existing colors. That means a reset sequence `CSI_PREFIX[m`, which has zero ansi classes, will result in the previous class set being used, rather than resetting the colors. This is in contrast to `CSI_PREFIX[0m`, which correctly empties the class set.

That said, in full log view, the escaper still gives up after 500something lines and starts printing unescaped sequences, e.g. like https://build.chimera-linux.org/#/builders/2/builds/1148/steps/5/logs/pkg_main_glycin-loaders_1_1_1-r0 - this is issue https://github.com/buildbot/buildbot/issues/7852 and appears to be unrelated.

### DIFF
--- a/www/base/src/util/AnsiEscapeCodes.tsx
+++ b/www/base/src/util/AnsiEscapeCodes.tsx
@@ -75,7 +75,7 @@ export function stripAnsiSgrEntry(ansiEntry: string): string {
 
 export function ansiSgrClassesToCss(ansiClasses: string[], cssClasses: {[key: string]: boolean}) {
   if (ansiClasses.length === 0) {
-    return cssClasses;
+    return {};
   }
 
   const fgbg: {[key: string]: string} = {'38': 'fg', '48': 'bg'};


q66:
There is also another separate issue and that is when a color is set overriding the previous color without performing an explicit reset, that is unhandled here (the previous color will be shown because it will take on both classes and the browser will take the first one)

q66:
i think the fix for that would be to change

```
    for (let i of ansiClasses) {
      if ((i === '39') || (i === '0')) { // "color reset" code and "all attributes off" code
        cssClasses = {};
      } else {
        cssClasses[`ansi${i}`] = true;
      }
    }
```

to always only have a single color class at a time, rather than adding more of them?

q66:
a naive way to address that is probably like

```
index a4896c52f..651a910f5 100644
--- a/www/base/src/util/AnsiEscapeCodes.tsx
+++ b/www/base/src/util/AnsiEscapeCodes.tsx
@@ -92,6 +92,17 @@ export function ansiSgrClassesToCss(ansiClasses: string[], cssClasses: {[key: st
       if ((i === '39') || (i === '0')) { // "color reset" code and "all attributes off" code
         cssClasses = {};
       } else {
+        // if we're setting ansi color, reset other color classes if we have them
+        let j = parseInt(i);
+        if (j >= 30 && j <= 37) {
+            for (j = 30; j <= 37; ++j) {
+                delete cssClasses[`ansi${i}`];
+            }
+        } else if (j >= 40 && j <= 47) {
+            for (j = 40; j <= 47; ++j) {
+                delete cssClasses[`ansi${i}`];
+            }
+        }
         cssClasses[`ansi${i}`] = true;
       }
     }
```

i don't really like it though? any ideas?

p12tic (changes requested):
Looks good, thanks.

Could you add a comment saying something like this: "According to ECMA-48 standard empty parameter set is interpreted as zero, which is a color reset code"?

Also, a unit test is needed. Please check out www/base/src/util/AnsiEscapeCodes.test.tsx which contains tests for parseEscapeCodesToClasses().

q66:
do you have any comments wrt the fix for the above issue that i did not include in the patch?

p12tic:
> do you have any comments wrt the fix for the above issue that i did not include in the patch?

Sorry for missing the question the first time.

I think the map can be replaced with the following object:

```
{
    fg: string | null; // current foreground class (one set via 38 code)
    bg: string | null; // current background class (one set via 48 code)
    simpleColor: string | null; // current simple color class (one set via 30..37 codes)
    classes: {[key: string]: boolean}; // the rest of classes
}
```

Then clearing colors would be quite simple and efficient. Also, this would fix support for 38 code clearing previous 48 code setting and vice-versa which made it impossible to set both background and foreground at the same time.

Do I miss anything?

q66:
looks fine, i'm not very familiar with typescript though...

p12tic:
@q66 If you aren't planning to finish this PR then I think someone else could take over as it's pretty straightforward what needs to be done for someone well versed in typescript.

p12tic:
This PR has been merged as part of https://github.com/buildbot/buildbot/pull/8180. Thanks a lot for investigation and providing preliminary fix.

PR #8004, opened 2024-09-23
www: compress Rest API responses

tdesveaux:
## Contributor Checklist:


p12tic (approved):
In the future it makes sense to support Brotli too.

tdesveaux:
> In the future it makes sense to support Brotli too.

It would be failry easy (zstd as well), but would require an extra dependency.
I think it's the reason Twisted only implement a GZip encoding class.

tdesveaux:
Convert to draft, adding Brotli/zstd support is pretty trivial, so might as well do it here.

tdesveaux:
Removed the config to control compression level. Everything uses the defaults, maybe the brotli/zstd ones should be bumped higher?

brotli and zstd are optional and can be installed with the corresponding extras when installing Buildbot.

p12tic (approved):
Great, thanks!

PR #8016, opened 2024-09-26
e2e: Disable WebKit runner due to internal bug in WebKit

p12tic:
WebKit contains a race condition that occurs under certain conditions.

https://bugs.webkit.org/show_bug.cgi?id=280366

It so happens that Buildbot internal CI which runs under gvisor manages to reproduce the issue quite often.

There exists a pull request that should address the issue.

https://github.com/WebKit/WebKit/pull/34268

For the time being the browser is disabled. It can be re-enabled once the fix lands into a version of WebKit distributed by Playwright.

### DIFF
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -40,10 +40,14 @@ export default defineConfig({
       use: { ...devices['Desktop Firefox'] },
     },
 
-    {
+    // Webkit is disabled due to an internal bug that occurs when tests are run under gvisor.
+    // https://bugs.webkit.org/show_bug.cgi?id=280366.
+    // https://github.com/WebKit/WebKit/pull/34268 should fix it, once it's in WebKit version
+    // used by playwright, webkit can be re-enabled.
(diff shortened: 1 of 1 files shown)


e2e/playwright.config.ts
```diff
@@ -40,10 +40,14 @@ export default defineConfig({
...
     },
 
-    {
+    // Webkit is disabled due to an internal bug that occurs when tests are run under gvisor.
+    // https://bugs.webkit.org/show_bug.cgi?id=280366.
+    // https://github.com/WebKit/WebKit/pull/34268 should fix it, once it's in WebKit version
+    // used by playwright, webkit can be re-enabled.
+    /*{
       name: 'webkit',
       use: { ...devices['Desktop Safari'] },
```
mcatanzaro, line 49: BTW this WebKitGTK bug does not affect Safari (the code is not compiled at all), so I suspect the test name is misleading... but perhaps you already know this.

PR #8017, opened 2024-09-26
db: add support for zstd/brotli for log compression

tdesveaux:
Changing the default compression to zstd if it's available before falling back to gzip is not really necessary, the commit can be dropped.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -24,6 +24,7 @@ from twisted.internet import defer
 from twisted.python import log
 
 from buildbot.db import base
+from buildbot.db.compression import BrotliCompressor
 from buildbot.db.compression import BZipCompressor
 from buildbot.db.compression import CompressorInterface
 from buildbot.db.compression import GZipCompressor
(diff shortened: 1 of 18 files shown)


master/buildbot/db/logs.py
```diff
@@ -157,7 +155,7 @@ def thdGetLogLines(conn) -> str:
             for row in conn.execute(q):
                 # Retrieve associated "reader" and extract the data
                 # Note that row.content is stored as bytes, and our caller expects unicode
-                data = self.COMPRESSION_BYID[row.compressed]["read"](row.content)
+                data = self.COMPRESSION_BYID.get(row.compressed, RawCompressor).read(row.content)
```
p12tic, line 158: We should probably throw an exception if the decompressor is unavailable.

master/buildbot/test/unit/scripts/test_cleanupdb.py
```diff
@@ -36,6 +36,22 @@
 except ImportError:
     hasLz4 = False
 
+try:
+    import zstandard
+
+    [zstandard]
+    hasZstd = True
```
p12tic, line 43: snake_case for new code. I do understand that we use camelCase just a few lines above, but there's no need to make future cleanup larger.

p12tic (changes requested):
Needs several small changes, but overall looks great, thank you!

p12tic (approved):

p12tic:
@tdesveaux Looks like this build https://buildbot.buildbot.net/#/builders/129/builds/666 is a valid test failure.

p12tic:
Thanks!

PR #8039, opened 2024-09-29
Ruff: apply assert-raises-exception lint

Mic92:
see https://docs.astral.sh/ruff/rules/assert-raises-exception/#assert-raises-exception-b017


master/buildbot/test/unit/util/test_git.py
```diff
@@ -155,7 +155,8 @@ def test_throws_when_wrapper_not_given(self):
 
         command = []
         env = {}
-        with self.assertRaises(Exception):
+        with self.assertRaises(RuntimeError):
+            self.setupGit()
```
Mic92, line 159: Found one bug in the test suite while doing this.

p12tic:
Thanks!

PR #8040, opened 2024-09-29
ruff: apply rule useless-expression (B018)

Mic92:
See https://docs.astral.sh/ruff/rules/useless-expression

### DIFF
--- a/master/buildbot/config/master.py
+++ b/master/buildbot/config/master.py
@@ -377,7 +377,7 @@ class MasterConfig(util.ComparableMixin):
             try:
                 import lz4  # pylint: disable=import-outside-toplevel
 
-                [lz4]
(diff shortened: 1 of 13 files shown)


master/buildbot/config/master.py
```diff
@@ -397,7 +397,7 @@ def copy_str_or_callable_param(name):
             try:
                 import brotli  # pylint: disable=import-outside-toplevel
 
-                [brotli]
+                _ = brotli
```
Mic92, line 400: Also my lsp recognizes this better as an unused value. Also looks less weird to me as a reader. First time seeing this it looked like SyntaxError. Also doesn't allocate a list.

p12tic (approved):

PR #8041, opened 2024-09-29
ruff: apply rule cached-instance-method (B019)

Mic92:
See https://docs.astral.sh/ruff/rules/cached-instance-method

### DIFF
--- a/master/buildbot/data/base.py
+++ b/master/buildbot/data/base.py
@@ -72,7 +72,7 @@ class ResourceType:
             pathPatterns[i] = pp
         self.eventPaths = pathPatterns
 
-    @functools.lru_cache(1)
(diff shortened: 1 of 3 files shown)


master/buildbot/data/base.py
```diff
@@ -72,7 +72,7 @@ def compileEventPathPatterns(self):
             pathPatterns[i] = pp
         self.eventPaths = pathPatterns
 
-    @functools.lru_cache(1)
+    @functools.lru_cache(1)  # noqa: B019
```
Mic92, line 75: I looked at the cases and given there is not an unbound number of instances, this memory leak is fine. But in general people should be aware of this issue and review if their usage is acceptable.

PR #8042, opened 2024-09-29
ruff: apply rule B020 + B023 

Mic92:


master/buildbot/spec/indent.py
```diff
@@ -5,7 +5,9 @@
 import sys
 
 spaces = re.compile("^ +")
-for fn in sys.argv[1:]:
+
+
+def indent_file(fn: str) -> None:
     lines = []
```
Mic92, line 11: Moving this into a function makes this a local variable instead of a loop variable.

Mic92: Also everything is now type checked.

p12tic (approved):

PR #8045, opened 2024-09-29
ruff: apply rule B904 and add bugbear rules to ci check

Mic92:


worker/buildbot_worker/scripts/create_worker.py
```diff
@@ -201,7 +201,7 @@ def createFile(path, file, contents):
         try:
             os.mkdir(path)
         except OSError as exception:
-            raise CreateWorkerError(f"error creating directory {path}: {exception.strerror}")
+            raise CreateWorkerError(f"error creating directory {path}") from exception
```
Mic92, line 204: The error message doesn't need to be repeated here:

```
class Foo(Exception):
    pass

def main() -> None:
    try:
        open("non_existent_file.txt")
    except FileNotFoundError as e:
        raise Foo("File not found") from e

if __name__ == "__main__":
    main()
```

```
  File "/tmp/foo.py", line 6, in main
    open("non_existent_file.txt")
FileNotFoundError: [Errno 2] No such file or directory: 'non_existent_file.txt'

The above exception was the direct cause of the following exception:

  File "/tmp/foo.py", line 11, in <module>
    main()
  File "/tmp/foo.py", line 8, in main
    raise Foo("File not found") from e
Foo: File not found
```

Issue #5380, opened 2020-06-22
owner error: builtins.TypeError: unhashable type: 'list'


buck-E:
Not sure if I'm doing something wrong or a bug.  I have a scheduler that worked until I added a properties value with owners.

Defined like `DOC_SCHEDULER = schedulers.SingleBranchScheduler(`

```
properties = {
        'owner': ['my@email.com']
    }
```

But when I run `buildbot-worker restart` the job doesn't finish.  In the UI it's not red or green but purple (have not seen that before) and the HTML shows this error:

```
builtins.TypeError: unhashable type: 'list'
/usr/lib/python3/dist-packages/buildbot/process/build.py:334 in startBuild
333        try:
334            self.setupBuild()  # create .steps
335        except Exception:
/usr/lib/python3/dist-packages/buildbot/process/build.py:486 in setupBuild
485        # gather owners from build requests
486        owners.update({r.properties['owner'] for r in self.requests
487                       if "owner" in r.properties})
/usr/lib/python3/dist-packages/buildbot/process/build.py:486 in <setcomp>
485        # gather owners from build requests
486        owners.update({r.properties['owner'] for r in self.requests
487                       if "owner" in r.properties})
builtins.TypeError: unhashable type: 'list'
```

Is this a bug or did I misunderstand how to apply the information in this doc: https://docs.buildbot.net/latest/manual/configuration/schedulers.html?highlight=owners

tardyp:
owner property shall be a string, not a list. it shall contain only one owner

buck-E:
https://docs.buildbot.net/latest/manual/configuration/reporters.html?highlight=owner:

```
You can also add interested users by setting the owners build property to a list of users in the scheduler constructor
```

```
The owner property may be of particular interest, as its contents (as a list) will be added to the list of “interested users”

...

    properties = {
        'owner': ['zorro@example.com', 'silver@example.com']
    })

owner and owners talk about *lists* and even have an example.

buck-E:
I'm not trying to be rude, you know.  I'm just very confused.  The docs are clear, but they are not what you're saying, and the example doesn't work.

buck-E:
@tardyp you see why your comment doesn't answer this right?  please see my reply when you have time.

tardyp:
I think the doc is wrong. and the previous behaviour was wrong. This is why I fixed it when working on buildbot 1.0. I forgot to update the doc :-}

Owner is singular. this does not make sense that it is possible to add a list

[...]

PR #8066, opened 2024-10-01
Fix Scheduler documentation regarding the "owner" property

mattock:
This fixes https://github.com/buildbot/buildbot/issues/5380

### DIFF
--- a/master/docs/manual/configuration/schedulers.rst
+++ b/master/docs/manual/configuration/schedulers.rst
@@ -135,14 +135,14 @@ There are several common arguments for schedulers, although not all are availabl
 ``properties`` (optional)
 
     This is a dictionary specifying properties that will be transmitted to all builds started by this scheduler.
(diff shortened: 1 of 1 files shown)


p12tic:
@mattock Thanks for the pull request. As I understand, list didn't work, right? I think this behavior is a bug and we should fix it (not necessarily you).

mattock:
@p12tic a list does not work. According to @tardyp in https://github.com/buildbot/buildbot/issues/5380 a list is not supposed to work:

> I think the doc is wrong. and the previous behaviour was wrong. This is why I fixed it when working on buildbot 1.0. I forgot to update the doc :-} Owner is singular. this does not make sense that it is possible to add a list

If I interpret all this correctly then fixing the documentation is the proper fix.

p12tic:
@mattock Many thanks for pointing this out.

I think this PR needs a release note in newsfragments directory, like in this commit 4c21a0667c3080480481936d5b269e066ca528c6.

p12tic:
This PR has been landed via https://github.com/buildbot/buildbot/pull/8187. Thanks a lot for contributing the fix.

PR #8073, opened 2024-10-02
www: fix ChoiceStringParameter when required and no default value

terencehonles:
This change fixes a regression where a ``ChoiceStringParameter`` requires a user selected value (required, but no default value), but the force build form incorrectly displays the first choice as being selected and does not submit it to the backend which causes a validation error.

## Contributor Checklist:

### DIFF
--- a/newsfragments/www-forcefield-choicestring-no-default.bugfix
+++ b/newsfragments/www-forcefield-choicestring-no-default.bugfix
@@ -1,0 +1,4 @@
+Fixes a regression where a ``ChoiceStringParameter`` requires a user selected
+value (no default value), but the force build form incorrectly displays the
+first choice as being selected and does not submit it to the backend which
+causes a validation error.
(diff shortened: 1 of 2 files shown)


terencehonles:
In case this needs to be backported, this is what I used (and tested):

```diff
diff --git a/www/react-base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx b/www/react-base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
index f6896218a..e64194b4d 100644
--- a/www/react-base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
+++ b/www/react-base/src/components/ForceBuildModal/Fields/FieldChoiceString.tsx
@@ -52,6 +52,10 @@ export const FieldChoiceString = observer(({field, fieldsState}: FieldChoiceStri
         <Form.Control data-bb-test-id={`force-field-${field.fullName}`}
                       as="select" multiple={field.multiple} value={state.value}
                       onChange={onChange}>
+          {!field.multiple && !field.default && !field.choices.includes(field.default)
+            ? <option>Select an option</option>
+            : <></>
+          }
           {field.choices.map(choice => (<option key={`force-field-${field.fullName}-option-${choice}`}>{choice}</option>))}
         </Form.Control>
         { !field.strict && !field.multiple
```

p12tic (approved):

PR #8049, opened 2024-09-29
ruff: enable RUF lints

Mic92:
https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf

Depends on https://github.com/buildbot/buildbot/pull/8045

PR #8078, opened 2024-10-05
[WIP] replace defer.inlineCallbacks with async/await [part 1]

Mic92:
Depends on https://github.com/buildbot/buildbot/pull/8049


Mic92:
Depends on https://github.com/buildbot/buildbot/pull/8049

p12tic:
There are several challenges with converting defer.inlineCallbacks to async/await:
 - It cannot be easily done incrementally in Buildbot codebase. I would be really unhappy if a bug is bisected to async/await change and it's single 1000-line commit instead of multiple commits that can be bisected further.
 - It changes the API exposed to users. I expect that most users would just yield the returned deferreds in their inlineCallbacks decorated functions. Unfortunately before Twisted 24.7.0 a inlineCallbacks couldn't yield a coroutine. Even if this is solved coroutine still hasn't the member functions of a Deferred.
 
 My plan was to wait until dust created by the 4.0 release settles. Then the solution would be something like this:
  - A decorator that would return a subclass of Deferred that would raise deprecation warnings if addCallbacks or other member functions are called.
  - The decorator would be applied to all documented public APIs
  - The rest of code would be changed to async/await.

Mic92:
Sounds good. 
I went ahead and automated step 3 "The rest of code would be changed to async/await.".
This works for the majority of the code base using redbaron. But I had some syntax parsing issues for a small number of files, which I now manually formatted.
After that I created 1 commit per changed file.

p12tic:
@Mic92 I think one place where we could already replace inlineCallbacks with async/await is all tests. Note that buildbot/master/buildbot/test/reactor.py and buildbot/master/buildbot/test/steps.py are exposed to end users and should not be changed just yet.

Maybe this will make unit tests a little bit faster and increase developer productivity a bit.

Mic92:
I think I want to unlock more typing first before doing scary async refactorings.

PR #8080, opened 2024-10-06
ruff: apply rule RUF015

Mic92:


p12tic (approved):
Ugh, I feel the old version is more readable. However, seems that this is Python standard now, so I will need to adapt..

PR #8081, opened 2024-10-06
ruff: apply rule RUF017

Mic92:


p12tic (changes requested):
Old version is more readable in this particular case.

PR #8085, opened 2024-10-06
ruff: apply rule RUF100

Mic92:


www/nestedexample/buildbot_nestedexample/__init__.py
```diff
@@ -1,3 +1,6 @@
+import functools
```
p12tic, line 1: Seems like unnecessary change

Mic92: Could be an artifact of taking commits apart.

Mic92: fixed.

Issue #7992, opened 2024-09-09
OAuth2Auth disables HTTPS certificate validation


LinAGKar:
By default, urllib3 will verify certificates when making HTTPS requests. But `OAuth2Auth` disables certificate verification by setting `s.verify = None` in https://github.com/buildbot/buildbot/blob/b81dc55c308184f51f0bcf55b6a7d848fe8a103a/master/buildbot/www/oauth2.py#L135 (unless the user modifies the `sslVerify` field), which means the verification requests are insecure by default, and leads to warnings in the log like:

```
/usr/lib/python3.6/site-packages/urllib3/connectionpool.py:993: InsecureRequestWarning: Unverified HTTPS request is being made to host 'example.com'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
```

PR #8086, opened 2024-10-06
auth: Allow default HTTPS request certificate verification

mokibit:
Fixes https://github.com/buildbot/buildbot/issues/7992.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/oauth2.py
+++ b/master/buildbot/www/oauth2.py
@@ -81,13 +81,13 @@ class OAuth2Auth(auth.AuthBase):
     tokenUriAdditionalParams = {}
     loginUri = None
     homeUri = None
-    sslVerify = None
 
-    def __init__(self, clientId, clientSecret, autologin=False, **kwargs):
+    def __init__(self, clientId, clientSecret, autologin=False, ssl_verify: bool = True, **kwargs):
         super().__init__(**kwargs)
         self.clientId = clientId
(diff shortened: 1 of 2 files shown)


master/buildbot/www/oauth2.py
```diff
@@ -81,13 +81,13 @@ class OAuth2Auth(auth.AuthBase):
     tokenUriAdditionalParams = {}
     loginUri = None
     homeUri = None
-    sslVerify = None
 
-    def __init__(self, clientId, clientSecret, autologin=False, **kwargs):
+    def __init__(self, clientId, clientSecret, autologin=False, sslVerify=True, **kwargs):
```
Mic92, line 85: Not mandatory for the PR but nice to have:

```suggestion
    def __init__(self, clientId: str, clientSecret: str, autologin: bool = False, sslVerify: bool = True, **kwargs: Any) -> None:
```

Mic92: Would also need a import of Any at the top.

mokibit: Thanks for suggestions, I improved the commit.
I only introduced the type for the argument which I added (ssl_verify).  Specifying other types could potentially break the code as I did not specifically analyze the types of those arguments. Maybe in next PR.

p12tic (approved):

PR #8088, opened 2024-10-06
buildbot: enable mypy in CI

Mic92:


p12tic (approved):
Looks great, thanks for taking initiative in this area.

PR #8096, opened 2024-10-08
Fix various mypy errors in buildbot-master [part 3]

Mic92:
Depends on https://github.com/buildbot/buildbot/pull/8093

### DIFF
--- a/master/buildbot/db/migrations/env.py
+++ b/master/buildbot/db/migrations/env.py
@@ -12,7 +12,7 @@ config = context.config
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
-fileConfig(config.config_file_name)
+fileConfig(str(config.config_file_name))
 
 target_metadata = model.Model.metadata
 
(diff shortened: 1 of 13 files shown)


master/buildbot/db/migrations/env.py
```diff
@@ -12,7 +12,9 @@
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
-fileConfig(config.config_file_name)
+fileConfig(str(config.config_file_name))
+
+# Argument 1 to "fileConfig" has incompatible type "Union[str, PathLike[str], None]"; expected "Union[Union[str, bytes, PathLike[str], PathLike[bytes]], IO[str], RawConfigParser]"
```
p12tic, line 17: Can this line be split into two?

p12tic (approved):

master/buildbot/db/migrations/env.py
Mic92: It actually this shouldn't be there.

Mic92: Removed.

PR #8097, opened 2024-10-08
Fix various mypy errors in buildbot-master [part 4]

Mic92:
Depends on https://github.com/buildbot/buildbot/pull/8096


Mic92:
CI looks odd in those one. However works in https://github.com/buildbot/buildbot/pull/8098

PR #8100, opened 2024-10-08
www: paginate WorkersView

tdesveaux:
Also add an input box to search on worker name

## Contributor Checklist:

### DIFF
--- a/www/base/src/views/WorkersView/WorkersView.tsx
+++ b/www/base/src/views/WorkersView/WorkersView.tsx
@@ -25,6 +25,7 @@ import {
   Master,
   Worker,
   useDataAccessor,
+  useDataApiDynamicQuery,
   useDataApiQuery
 } from "buildbot-data-js";
 import {WorkerActionsModal} from "../../components/WorkerActionsModal/WorkerActionsModal";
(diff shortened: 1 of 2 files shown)


www/base/src/views/WorkersView/WorkersView.tsx
```diff
@@ -75,28 +81,46 @@ const getBuildsForWorkerMap = (workersQuery: DataCollection<Worker>,
...
-    useState<boolean>(false);
+  const [paginatedWorkers, paginationElement] = makePagination(
+    currentPage, setCurrentPage,
+    settings.getIntegerSetting("Workers.page_size"),
+    filteredWorkers
+  );
+
+  const paginatedWorkerIds = paginatedWorkers.map(w => w.workerid).sort();
+  const buildsQuery = useDataApiDynamicQuery(paginatedWorkerIds, () => {
+    // wait for workersQuery to be resolved to avoid querying withou
```
p12tic, line 109: typo

www/base/src/views/WorkersView/WorkersView.tsx
```diff
@@ -153,6 +184,11 @@ buildbotSetupPlugin((reg) => {
       name: 'show_old_workers',
       caption: 'Show old workers',
       defaultValue: false
+    }, {
+      type:'integer',
```
p12tic, line 188: spocing

www/base/src/views/WorkersView/WorkersView.tsx
```diff
@@ -75,28 +81,46 @@ const getBuildsForWorkerMap = (workersQuery: DataCollection<Worker>,
...
+    filteredWorkers
+  );
+
+  const paginatedWorkerIds = paginatedWorkers.map(w => w.workerid).sort();
+  const buildsQuery = useDataApiDynamicQuery(paginatedWorkerIds, () => {
+    // wait for workersQuery to be resolved to avoid querying withou
+    // workerid filter
+    if (!workersQuery.isResolved()) {
+      const col = new DataCollection<Build>();
+      col.resolved = true;
```
p12tic, line 113: I think col.resolved is not needed here. Until workersQuery is not resolved, builds query is not resolved either.

tdesveaux: If the 'real' query condition is never matched, the DataCollection will always be unresolved without doing this.
Which mean, in cases where `isResolved` is used to display a `LoadingSpan`, it will always wrongly be marked as 'loading'.

It's probably irrelevant here (with the current code anyway), but it became something of an habit to avoid the issue mentioned above.

p12tic: Thanks for explaining.

p12tic (approved):

PR #8107, opened 2024-10-10
Require Twisted 22.1 or newer

p12tic:
No distributions package Twisted between 21.2.0 (inclusive) and 22.1.0 (exclusive), thus this change will not introduce difficult compatibility problems for the end users (e.g. when using distribution Python packages). Users using virtualenv will be able to upgrade to newer Twisted easily, if they even use Twisted that old.

This change was necessary because newer Twisted is required for typing changes.

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -37,8 +37,8 @@ matrix:
 
     # include "ci" string into the name of the status that is eventually submitted to Github, so
     # that the codecov.io service would wait until this build is finished before creating report.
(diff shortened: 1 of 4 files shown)


p12tic:
cc @Mic92 FYI this PR fixes test breakage introduced in https://github.com/buildbot/buildbot/pull/8101/commits/b561631fa93ae7b34658544c0f5613998a772274.

Mic92:
Is it because of `defer.Deferred[str]`?

p12tic:
Probably, didn't check in detail. Twisted 21.7 has a release note about fixing Deferred typing, presumably before that it was broken.

PR #8108, opened 2024-10-10
www: Increase workers page pagination to 100 items

p12tic:
25 introduces too much friction when navigating

No release note necessary because this is new feature.


p12tic:
cc @tdesveaux

PR #8114, opened 2024-10-11
fix all remaining type errors except in unit tests [part 8]

Mic92:


Mic92:
If the rest of the CI also goes green, I believe this is ready for review: @p12tic

p12tic:
@Mic92 Thanks a lot!

p12tic:
Oh, I noticed that this PR changes the worker. I think let's hold off from that because worker should support old Python versions as well. This wasn't tested in tests and is currently broken unfortunately.

p12tic:
Looks like worker works on old python as well, great.

Mic92:
To avoid regressions, we should have a different pyproject.toml with a different minimum python version.

p12tic:
> To avoid regressions, we should have a different pyproject.toml with a different minimum python version for the worker code.

Good idea, thanks.

p12tic:
I will delay this PR a bit until I release 4.1.0 today or tomorrow.

p12tic:
I removed the `ruff: enable flake8-future-annotations (FA) rules` commit because I think we should attribute it to @tdesveaux and merge his PR instead. I will make sure the merged PR contains combination of whatever I removed and what's in #8117.

PR #8115, opened 2024-10-11
ci: remove python3.8, which is eol

Mic92:
- also see https://devguide.python.org/versions/
- python3.8 doesn't support subscribed types, which makes it difficult to use with typing
- many Linux distribution won't offer this python version anymore, because it no longer receives any security fixes
- even Debian oldstable has python3.9 as default

### DIFF
--- a/.bbtravis.yml
+++ b/.bbtravis.yml
@@ -39,8 +39,7 @@ matrix:
     - env: PYTHON=3.9 TREQ=22.2.0 TWISTED=22.1.0 SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
     - env: PYTHON=3.9 TREQ=22.2.0 TWISTED=22.4.0 SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=1.4.52 NUM_CPU=2 TESTS=trial
-    - env: PYTHON=3.8 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=ci/coverage
-    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
+    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=ci/coverage
     - env: PYTHON=3.10 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
     - env: PYTHON=3.11 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
     - env: PYTHON=3.12 TWISTED=latest SQLALCHEMY=latest NUM_CPU=2 TESTS=trial
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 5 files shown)


tdesveaux:
Not related to dropping support for py3.8.

> python3.8 doesn't support subscribed types, which makes it difficult to use with typing

I see no reason not to use `from __future__ import annotations` (https://docs.python.org/3/library/__future__.html https://peps.python.org/pep-0563/). It makes type hinting easier to work with.

Mic92:
One thing to consider when supporting these old python versions is, that one is basically expect from contributors to install unmaintained, potentially insecure libraries.

Mic92:
> Not related to dropping support for py3.8.
> 
> > python3.8 doesn't support subscribed types, which makes it difficult to use with typing
> 
> I see no reason not to use `from __future__ import annotations` (https://docs.python.org/3/library/__future__.html https://peps.python.org/pep-0563/). It makes type hinting easier to work with.

The only disadvantage I found so far is that it might break libraries that try to introspect on type hints. Likely not used in this project. i.e. if you get member field types from a dataclass it would than return a string instead of the type object.

tdesveaux:
> The only disadvantage I found so far is that it might break libraries that try to introspect on type hints. Likely not used in this project. i.e. if you get member field types from a dataclass it would than return a string instead of the type object.

From https://peps.python.org/pep-0563/#resolving-type-hints-at-runtime

> For code that uses type hints, the typing.get_type_hints(obj, globalns=None, localns=None) function correctly evaluates expressions back from its string form. Note that all valid code currently using __annotations__ should already be doing that since a type annotation can be expressed as a string literal.

Since:
```python
def func() -> "int":
    return 0
```
is the same as:
```python
from __future__ import annotations

def func() -> int:
    return 0
```

Since string types are already used ([example here](https://github.com/buildbot/buildbot/blob/5e9c38f8948f622521420a25f5f250d53f273fe2/master/buildbot/interfaces.py#L148)), any library that does not correctly introspect would be already broken.

There are other advantages to the annotations future, but I'll stop here as it's not really the subject of this PR.
I'll open an issue or PR to continue on this discussion.

worker/setup.py
```diff
@@ -84,11 +84,6 @@ def make_release_tree(self, base_dir, files):
         'Topic :: Software Development :: Build Tools',
         'Topic :: Software Development :: Testing',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
```
p12tic, line 87: Buildbot worker should still support the oldest Python release possible. People run worker on all kinds of weird and old machines.

p12tic:
I would weakly prefer to keep python 3.8, because Ubuntu 20.04 LTS still supports it. Maybe using `from __future__ import annotations` is enough to make python 3.8 not as painful?

Mic92:
I guess the annotations stuff is enough to continue to support version-impaired distributions for now.

Mic92:
But how long do we want to keep supporting this? Ubuntu 20.04 goes EOL in 2030-04, I hope to use newer python features before that.

p12tic:
In reality buildbot 4.1 will be released sometime this month and buildbot 4.2 will be released around end of the year. So after that we can already merge this PR.

p12tic:
Master parts of this PR could go in, but the branch has been removed :/

Mic92:
I will try to restore.

Mic92:
@p12tic feel free edit as you want. Might be short of time till end of the year for this line of work.

Mic92:
Rebased.

PR #8117, opened 2024-10-11
ruff: enable flake8-future-annotations (FA) rules

tdesveaux:
EDIT: #8114 should be merged first

Related discussion started in https://github.com/buildbot/buildbot/pull/8115#issuecomment-2406915799

Using `from __future__ import annotations` with typing add a lot of convenience:
- no need to quote annotations, even for forward declaration
- use basic types from stdlib instead of from typing (`List` -> `list`)
- ` | ` instead of `Union[]`
- ` | None` instead of `Optional`

I noticed the worker still supports py3.5 in setup.py. Is it a simple missed deprecation or a conscious choice?
This would prevent the merge of this PR as the import was introduced in py3.7.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/mail.py
+++ b/master/buildbot/changes/mail.py
@@ -17,6 +17,8 @@
 Parse various kinds of 'CVS notify' email.
 """
 
+from __future__ import annotations
+
 import calendar
 import datetime
 import re
--- a/master/buildbot/changes/mail.py
+++ b/master/buildbot/changes/mail.py
@@ -29,7 +29,6 @@ from email.utils import mktime_tz
 from email.utils import parseaddr
 from email.utils import parsedate_tz
 from typing import ClassVar
-from typing import Optional
 from typing import Sequence
 
 from twisted.internet import defer
@@ -47,7 +46,7 @@ class MaildirSource(MaildirService, util.ComparableMixin):
 
     compare_attrs: ClassVar[Sequence[str]] = ("basedir", "pollInterval", "prefix")
     # twisted is marked as typed, but doesn't specify this type correctly
-    name: Optional[str] = 'MaildirSource'  # type: ignore[assignment]
+    name: str | None = 'MaildirSource'  # type: ignore[assignment]
 
     def __init__(self, maildir, prefix=None, category='', repository=''):
         super().__init__(maildir)
--- a/master/buildbot/changes/manager.py
+++ b/master/buildbot/changes/manager.py
@@ -13,6 +13,8 @@
 #
 # Copyright Buildbot Team Members
 
+from __future__ import annotations
+
 from typing import Optional
 
 from buildbot.process.measured_service import MeasuredBuildbotServiceManager
--- a/master/buildbot/changes/manager.py
+++ b/master/buildbot/changes/manager.py
@@ -15,12 +15,10 @@
 
 from __future__ import annotations
 
-from typing import Optional
-
 from buildbot.process.measured_service import MeasuredBuildbotServiceManager
 
 
 class ChangeManager(MeasuredBuildbotServiceManager):
-    name: Optional[str] = "ChangeManager"  # type: ignore[assignment]
+    name: str | None = "ChangeManager"  # type: ignore[assignment]
     managed_services_name = "changesources"
     config_attr = "change_sources"
--- a/master/buildbot/data/graphql.py
+++ b/master/buildbot/data/graphql.py
@@ -19,7 +19,6 @@ import asyncio
 import functools
 import textwrap
 from types import ModuleType
-from typing import Optional
 
 from buildbot.asyncio import AsyncIOLoopWithTwisted
 from buildbot.asyncio import as_deferred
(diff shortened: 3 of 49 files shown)


Mic92 (approved):
I guess overall this avoids some import dependency cycle this way.

Mic92:
If you PR was created automatically, I would prefer to have https://github.com/buildbot/buildbot/pull/8114 merged first, to avoid having to fix merge conflicts.

tdesveaux:
Not automatic, but pretty trivial, so I don't mind waiting for the merge and handling the rebase.

p12tic:
> I noticed the worker still supports py3.5 in setup.py. Is it a simple missed deprecation or a conscious choice?

Yes, this is conscious choice. We cannot use modern python on the worker.

p12tic (approved):

tdesveaux:
> I think this PR is ok to merge, isn't it. It's still in draft.

was waiting on information about python version on the worker.
If the worker still need to support py<3.7, this can't be merged as is I think.

tdesveaux:
Well, a simple copy-paste of the current github job to run tests is not enough.
I'll look into setting up CI tests (at least github), and make a PR of it before un-drafting this.

p12tic:
We already have tests for older python, just not on Windows:

```
    # Worker-master interoperability tests
    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.9
    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.8
    - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.7
```

Maybe we can remove type annotations on worker and call it a day?

tdesveaux:
Indeed, missed that.
Looks like these tests are faulty though.

I tried to run the unit tests in a docker python:3.5 image, and installation fails:
```bash
root@8ea255511ee5:/buildbot# venvbb/bin/pip install -e worker
Obtaining file:///buildbot/worker
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/buildbot/worker/setup.py", line 29, in <module>
        from buildbot_worker import version
      File "/buildbot/worker/buildbot_worker/__init__.py", line 47
        return f"{major}.{minor}.{patch}.dev{dev}"
                                                 ^
    SyntaxError: invalid syntax
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /buildbot/worker/
```

Issue is with the step ```bash
set -e
rm -rf /tmp/bbworkervenv
python$PYTHON -m venv /tmp/bbworkervenv
/tmp/bbworkervenv/bin/pip install -e worker
```

using `$PYTHON` instead of `$WORKER_PYTHON`

**Comment by p12tic** [comment]
Right, these don't run python tests, but run integration test of sorts. See `            worker_python_version = os.environ.get("WORKER_PYTHON", None)` in `master/buildbot/test/util/integration.py`.

**Comment by p12tic** [comment]
But you're right, bbworkerenv is set incorrectly.

**Comment by p12tic** [comment]
From https://github.com/buildbot/buildbot/pull/8119 it's rather clear that the worker has been broken lately.

**Comment by Mic92** [comment]
Btw also with the import it will still not work for python38

```
def callFromThread(self, callable: Callable[..., Any], *args: object, **kwargs: object) -> None:  # type: ignore[override]
builtins.TypeError: 'ABCMeta' object is not subscriptable
```

I dropped those definitions now in my pull request and enabled the ruff linting rule.

Mic92:
It get even worse.

For new python version we need this: builtins.DeprecationWarning: Callable must be used as Callable[[arg, ...], result]. (This will raise a TypeError in Python 3.10.)

Mic92:
Ok, I think I fixed it.

tdesveaux:
> It get even worse.
> 
> For new python version we need this: builtins.DeprecationWarning: Callable must be used as Callable[[arg, ...], result]. (This will raise a TypeError in Python 3.10.)

I think the issue is with import for Callable from `collections`, which does not support type subsribing in py38.
You should import it from `typing`

Mic92:
> > It get even worse.
> > For new python version we need this: builtins.DeprecationWarning: Callable must be used as Callable[[arg, ...], result]. (This will raise a TypeError in Python 3.10.)
> 
> I think the issue is with import for Callable from `collections`, which does not support type subsribing in py38. You should import it from `typing`

Ok. I think we have another linter that currently want this import.

p12tic (approved):

p12tic:
I will merge this PR because tests pass and another PR depends on it.

p12tic:
Just noticed "EDIT: https://github.com/buildbot/buildbot/pull/8114 should be merged first". Agreed.

p12tic:
Turns out https://github.com/buildbot/buildbot/pull/8114 cannot be merged without putting this PR on top. Thus I will merge this PR first and then fix the fallout in that other PR.

PR #8119, opened 2024-10-11
Use correct python versions for worker tests

p12tic:


.bbtravis.yml
```diff
@@ -55,7 +55,11 @@ matrix:
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.6
     - env: PYTHON=3.9 TWISTED=latest SQLALCHEMY=latest TESTS=interop WORKER_PYTHON=3.5
 
-    # Worker tests on python and twisted package combinations.
+    # Worker tests on older python and twisted package combinations.
+    - env: PYTHON=3.5 TREQ=22.2.0 TWISTED=21.2.0 SQLALCHEMY=latest TESTS=trial_worker
```
Mic92, line 59: Why does buildbot_worker needs to python3.5 support? Even redhat 6 had python3.6 and that is no longer supported since June.
Also in this case wouldn't you use buildbot from your distribution instead from upstream? What criteria is this decision based on?

.bbtravis.yml
```diff
@@ -55,7 +55,11 @@ matrix:
...
+    - env: PYTHON=3.6 TREQ=22.2.0 TWISTED=21.2.0 SQLALCHEMY=latest TESTS=trial_worker
```
Mic92, line 60: Same question for this one.

.bbtravis.yml
p12tic: For worker the criteria is "for as long as reasonably possible". In this PR I simply restore the tests that should have been running, but didn't. I didn't yet look into what's reasonable to support right now. 

My initial testing showed that e.g. python3.5 is not able to connect to pypi without throwing SSL errors, so this version is certainly no longer reasonable to support. For python3.6 you also make a convincing argument. So probably 3.7 is where it makes sense to draw the line.

Overall buildbot-worker is rather small project, so I think it's not too problematic to support ancient python versions and whatever other weirdness.

Mic92: My main concern was that early version had not good support for async/await syntax, which would migration harder later. I also forgot where the cut-off point was where it got reasonable.

p12tic: Genuinely asking: in what way the migration would be harder. It is true that we would need to postpone migration for the worker, but master does not share any code with the worker, so migration could go full speed there.

PR #8128, opened 2024-10-12
don't allow new untyped modules

Mic92:
depends on https://github.com/buildbot/buildbot/pull/8127


Mic92:
As this would make writing new modules quite hard,
We can have a new list, where modules go that call to untyped functions: https://github.com/buildbot/buildbot/pull/8131/commits/0707aad196e5625fff01c8e5907d105f7985a04d#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R153

This ensures we don't regress on type checking too much.

Mic92:
However it's also not super unreasonable for new code to add types to all methods as they get called. If you don't understand what types the functions you use returns, you might have a problem anyway.

p12tic:
I think let's wait until at least a majority of Buildbot code is typed. Then we can revisit. I feel this would add a significant burden on new contributions when Buildbot itself doesn't pay it yet.

PR #8131, opened 2024-10-13
Add type checks to all methods in buildbot.process.buildstep

Mic92:
Depends on https://github.com/buildbot/buildbot/pull/8128

### DIFF
--- a/master/buildbot/process/buildstep.py
+++ b/master/buildbot/process/buildstep.py
@@ -18,12 +18,15 @@ from __future__ import annotations
 import inspect
 import sys
 from typing import TYPE_CHECKING
+from typing import Any
 from typing import Callable
 from typing import ClassVar
 from typing import Sequence
+from typing import cast
 
 from twisted.internet import defer
 from twisted.internet import error
+from twisted.internet.base import ReactorBase
 from twisted.python import deprecate
 from twisted.python import log
 from twisted.python import versions
@@ -51,7 +54,7 @@ from buildbot.process import results
 from buildbot.process.locks import get_real_locks_from_accesses
 
 # (WithProperties used to be available in this module)
-from buildbot.process.properties import WithProperties
+from buildbot.process.properties import WithProperties  # noqa: F401
 from buildbot.process.results import ALL_RESULTS
 from buildbot.process.results import CANCELLED
 from buildbot.process.results import EXCEPTION
@@ -68,8 +71,13 @@ from buildbot.util import flatten
 from buildbot.util.test_result_submitter import TestResultSubmitter
 
 if TYPE_CHECKING:
+    from buildbot.interfaces import IBuildStep
+    from buildbot.master import BuildMaster
     from buildbot.process.build import Build
+    from buildbot.process.log import StreamLog
+    from buildbot.util.twisted import InlineCallbacksType
     from buildbot.worker.base import AbstractWorker
+    from buildbot.worker.protocols.base import Connection
 
 
 class BuildStepFailed(Exception):
@@ -96,16 +104,16 @@ class _BuildStepFactory(util.ComparableMixin):
 
     compare_attrs: ClassVar[Sequence[str]] = ('factory', 'args', 'kwargs')
 
-    def __init__(self, step_class, *args, **kwargs):
+    def __init__(self, step_class: type[BuildStep], *args: Any, **kwargs: Any) -> None:
         self.step_class = step_class
         self.args = args
         self.kwargs = kwargs
 
-    def buildStep(self):
+    def buildStep(self) -> BuildStep:
         try:
             step = object.__new__(self.step_class)
             step._factory = self
-            step.__init__(*self.args, **self.kwargs)
+            step.__init__(*self.args, **self.kwargs)  # type: ignore[misc]
             return step
         except Exception:
             log.msg(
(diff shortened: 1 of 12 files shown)


Mic92:
Will have a look at ci failures later

master/buildbot/process/buildstep.py
```diff
@@ -421,8 +437,9 @@ def getResultSummary(self):
         return {'step': stepsumm}
 
     @defer.inlineCallbacks
-    def getBuildResultSummary(self):
-        summary = yield self.getResultSummary()
+    def getBuildResultSummary(self) -> Generator[defer.Deferred[Any], Any, dict[str, str]]:
```
p12tic, line 440: This is a bit hard to understand, especially to new contributors to Buildbot.

I wonder if a better path would be to convert the function to async and decorate it using `async_to_deferred` to preserve `Deferred` return type. Then the function return type hint would be simply `-> dict[str, str]`.

Am I missing anything?

Mic92: To do this conversion there is still some things that are unclear to me:

* I often see a yield on an object that is not a deferred one. If I convert this to await it will be a runtime error. However if do not convert this to an await, is there potential other code that might override the function in a subclass and makes it async?
* Also how does twisted deals with deferred objects that are never yield? With await these futures might be never executed. In my own project I had deferred object that had never been yielded and there were fine somehow. Without having types everywhere, this seems very dangerous because now I have to check the whole code base if there is any method call that is not deferred.

With just adding types so far it seems relatively safe to not change the behavior too much. However with async I have the feeling, I can potentially introduce many nasty bugs.

p12tic: > I often see a yield on an object that is not a deferred one. If I convert this to await it will be a runtime error. However if do not convert this to an await, is there potential other code that might override the function in a subclass and makes it async?

Thanks, I haven't realized that. Both ways possible - Deferred to non-Deferred and vice-versa. We'll need something like the following in all places where we await something that user may extend.

```
async def maybe_async(c):
    if inspect.isawaitable(c) or isinstance(c, defer.Deferred):
        return await c
    return c
```

> Also how does twisted deals with deferred objects that are never yield? With await these futures might be never executed.

Twisted still executes them. However if we decorate the function with `async_to_deferred` it will still be awaited in all cases and bug will not happen.

Is it possible to define some type that instead of `Generator[defer.Deferred[Any], Any, dict[str, str]]` one would write e.g. `DeferredGenerator[dict[str, str]]`? That would alleviate my concerns.

tdesveaux: > Thanks, I haven't realized that. Both ways possible - Deferred to non-Deferred and vice-versa. We'll need something like the following in all places where we await something that user may extend.

Twisted already has the helper `defer.maybeDeferred` that does exactly this.

> Is it possible to define some type that instead of Generator[defer.Deferred[Any], Any, dict[str, str]] one would write e.g. DeferredGenerator[dict[str, str]]? That would alleviate my concerns.

It is possible, see https://github.com/buildbot/buildbot/issues/7371#issuecomment-1956564405

p12tic: > Twisted already has the helper defer.maybeDeferred that does exactly this.

Yeah, the only reason why I wrote maybe_async was not to go through Deferred if the argument is coroutine already.

> It is possible, see https://github.com/buildbot/buildbot/issues/7371#issuecomment-1956564405

Cool, I haven't read that comment until now.

tdesveaux: > Yeah, the only reason why I wrote maybe_async was not to go through Deferred if the argument is coroutine already.

While it would be better, wouldn't it be a bit unsafe until Buildbot require Twisted min version to be 24.7.0 due to [this issue that you patched](https://github.com/twisted/twisted/pull/12239)?

p12tic: https://github.com/twisted/twisted/pull/12239 is about users doing `yield awaitable`. Our plan would be to decorate all functions at API boundaries with `async_to_deferred`. Accordingly, API users would never get a coroutine, but a normal Deferred.

Mic92: @p12tic what are api boundaries actually? All functions without a _?

p12tic: All documented functions. If one cannot find function in the documentation, then it's not an exposed API.

Mic92:
@tdesveaux btw. https://docs.python.org/3.14/whatsnew/3.14.html has deprecated the `from __future__ import annotations` so we will need to get rid of it eventually.

master/buildbot/process/buildstep.py
p12tic: @Mic92 I re-read this discussion and it seems that while we covered a lot of topics, what in the end is important is that using `InlineCallbacksType[dict[str, int]]` instead of `Generator[defer.Deferred[Any], Any, dict[str, str]]` would make this PR good to go. The rest of suggestions were either incorrect or could be applied in the future.

InlineCallbacksType would be defined as follows (taken from https://github.com/buildbot/buildbot/issues/7371#issuecomment-1956564405 by @tdesveaux)

```
from typing import Any
from typing import Generator
from typing import TypeVar

T = TypeVar('T')
InlineCallbacksType = Generator[defer.Deferred[Any], Any, T]
```

terencehonles:
> @tdesveaux btw. https://docs.python.org/3.14/whatsnew/3.14.html has deprecated the `from __future__ import annotations` so we will need to get rid of it eventually.

Just a drive by, but it's deprecated because it's the default at that point, and when 3.13 reaches EoL it can be removed. That's quite a bit of time...

Mic92:
Will revisit this later.

p12tic:
This one can be open, because it doesn't need a lot improvements to land.

p12tic:
Rebased.

Mic92:
Thanks!

PR #8137, opened 2024-10-16
Fix CommandToString.test_list_with_objects

cjwatson:
59c3ccd0f69e9d2b5ed72310c85158c73ef9e6af changed the behaviour of `command_to_string` here.

This went unnoticed because the test file had been shadowed by a package.  I disentangled this.

## Contributor Checklist:


p12tic:
Thanks. @cjwatson How do you notice such things? You have a superpower :)

cjwatson:
Debian's CI [noticed it for me](https://ci.debian.net/packages/a/ansible-core/testing/i386/53117033/) - if you wade through the extremely noisy output enough, that is!

cjwatson:
A Debian package build noticed it for me, as a side-effect of somehow managing to run tests with `PYTHONPATH` set to include both the package being built and an installed version of it.  I was very confused for a while!

PR #8140, opened 2024-10-18
 db: LogsConnectorComponent.compressLog run compression outside DB thread

tdesveaux:
So, this is a relatively big one.

We upgraded our instance to 4.1 and enabled log compression with zstd. I was even greedy and modified the compression level from 9 to 22 as usually, we were more pressured on DB IO than CPU.

I did notice before that compression occurred in a DB thread, so took a mental note. Now, the time spent on compression is pretty significant since we host the master on a relatively weak host.

Anyway, our instance has the DB thread pool locked by compression tasks, leading to the webUI not responding or being overall sluggish.

I'll wait for feedback on this to do the same for `getLogLines`, and `thdSplitAndAppendChunk`.

Did not add a newsfragment. It's neither a bug nor feature, but might be worth to mention it anyway in a misc or change?

PS: the fact the trailing `\n` is removed from logchunks makes it really awkward to work with in most of the code. But I don't see an easy way to change this, as it would need a DB migration that rewrite all chunks...

## Contributor Checklist:

* [?] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)

### DIFF
--- a/master/buildbot/db/compression/zstd.py
+++ b/master/buildbot/db/compression/zstd.py
@@ -15,6 +15,7 @@
 
 from __future__ import annotations
 
+from buildbot.db.compression.protocol import CompressObjInterface
 from buildbot.db.compression.protocol import CompressorInterface
 
 try:
@@ -29,10 +30,38 @@ class ZStdCompressor(CompressorInterface):
     name = "zstd"
     available = HAS_ZSTD
 
+    COMPRESS_LEVEL = 9
+
     @staticmethod
     def dumps(data: bytes) -> bytes:
-        return zstandard.compress(data, level=9)
+        return zstandard.compress(data, level=ZStdCompressor.COMPRESS_LEVEL)
 
     @staticmethod
     def read(data: bytes) -> bytes:
-        return zstandard.decompress(data)
+        # data compressed with streaming APIs will not
+        # contains the content size in it's frame header
+        # which is expected by ZstdDecompressor.decompress
+        # use ZstdDecompressionObj instead
+        # see: https://github.com/indygreg/python-zstandard/issues/150
+        decompress_obj = zstandard.ZstdDecompressor().decompressobj()
+        return decompress_obj.decompress(data) + decompress_obj.flush()
+
+    class CompressObj(CompressObjInterface):
+        def __init__(self) -> None:
+            # zstd compressor is safe to re-use
+            # Note that it's not thread safe
+            self._compressor = zstandard.ZstdCompressor(level=ZStdCompressor.COMPRESS_LEVEL)
+            self._create_compressobj()
+
+        def _create_compressobj(self) -> None:
+            self._compressobj = self._compressor.compressobj()
+
+        def compress(self, data: bytes) -> bytes:
+            return self._compressobj.compress(data)
+
+        def flush(self) -> bytes:
+            try:
+                return self._compressobj.flush(flush_mode=zstandard.COMPRESSOBJ_FLUSH_FINISH)
+            finally:
+                # recreate compressobj so this instance can be re-used
+                self._create_compressobj()
--- a/master/buildbot/db/connector.py
+++ b/master/buildbot/db/connector.py
@@ -46,7 +46,9 @@ from buildbot.db import test_results
 from buildbot.db import users
 from buildbot.db import workers
 from buildbot.util import service
+from buildbot.util.deferwaiter import DeferWaiter
 from buildbot.util.sautils import get_upsert_method
+from buildbot.util.twisted import async_to_deferred
 
 upgrade_message = textwrap.dedent("""\
 
@@ -148,6 +150,7 @@ class DBConnector(AbstractDBConnector):
         self._engine = enginestrategy.create_engine(db_url, basedir=self.basedir)
         self.upsert = get_upsert_method(self._engine)
         self.pool = pool.DBThreadPool(self._engine, reactor=self.master.reactor, verbose=verbose)
+        self._db_tasks_waiter = DeferWaiter()
 
         # make sure the db is up to date, unless specifically asked not to
         if check_version:
@@ -162,6 +165,19 @@ class DBConnector(AbstractDBConnector):
                     log.msg(l)
                 raise exceptions.DatabaseNotReadyError()
 
+    @async_to_deferred
+    async def _shutdown(self) -> None:
+        """
+        Called by stopService, except in test context
+        as most tests don't call startService
+        """
+        await self._db_tasks_waiter.wait()
+
+    @defer.inlineCallbacks
+    def stopService(self):
+        yield self._shutdown()
+        yield super().stopService()
+
     def _doCleanup(self):
         """
         Perform any periodic database cleanup tasks.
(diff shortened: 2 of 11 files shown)


p12tic:
A couple of comments before I do full review:

> Did not add a newsfragment. It's neither a bug nor feature, but might be worth to mention it anyway in a misc or change?

Yeah, misc makes sense. change is for things that affect some kind of API or UI exposed to users.

> PS: the fact the trailing \n is removed from logchunks makes it really awkward to work with in most of the code. But I don't see an easy way to change this, as it would need a DB migration that rewrite all chunks...

Agreed.

master/buildbot/db/compression/zstd.py
```diff
@@ -29,10 +30,36 @@ class ZStdCompressor(CompressorInterface):
...
     @staticmethod
     def dumps(data: bytes) -> bytes:
-        return zstandard.compress(data, level=9)
+        return zstandard.compress(data, level=ZStdCompressor.COMPRESS_LEVEL)
 
     @staticmethod
     def read(data: bytes) -> bytes:
-        return zstandard.decompress(data)
+        # not too sure why, but zstandard.decompress do not work with
+        # data compressed with `ZstdCompressionObj`
```
p12tic, line 42: I think this is pretty serious problem and indicates some kind of a bug. Could we compare the bytes emitted by ZstdCompressionObj and compare them to what zstandard.compress emits?

master/buildbot/db/compression/zstd.py
```diff
@@ -29,10 +30,36 @@ class ZStdCompressor(CompressorInterface):
...
+        # `ZstdDecompressionObj` does and can also decompress data from `zstandard.compress`
+        decompress_obj = zstandard.ZstdDecompressor().decompressobj()
+        return decompress_obj.decompress(data) + decompress_obj.flush()
+
+    class CompressObj(CompressObjInterface):
```
p12tic, line 47: zstandard recomments against using CompressObjInterface due to performance reasons https://python-zstandard.readthedocs.io/en/latest/compressor.html#zstdcompressionobj . Was the streaming API proposed by them considered? Not suggesting to change anything in this PR because it's still an improvement.

p12tic:
> This will probably hurt performance in simple scenarios (master instances without much to do).

Please fix my understanding. With the changes in this PR we basically introduce a small penalty to the overall log compression throughput and latency because there's multiple roundtrips to the DB thread instead of one.

However, on lightly loaded master this wouldn't be noticeable anyway because there are free CPU resources for this additional work that needs to be done. For the end user arrival of logs some milliseconds later shouldn't be visible either. Do I miss anything?

I'd say that splitting work on the DB thread into multiple smaller work items should strictly improve user responsiveness because other work has chance to get processed faster.

Also, roundrips to the database (as opposed to the DB thread) are likely to have way more impact on latency. In this case we remove one roundtrip to the database by optimizing calculation of how many bytes the recompression saved. So maybe the PR actually even improves performance in cases there aren't too many log chunk groups that are compressed separately.

p12tic (approved):
Overall looks good. 

Please ensure the commit title line lengths are less than 72 characters. 72 characters is the accepted best practice (some argue for 50 character titles) and many tools are configured to show that many characters of the commit title.

master/buildbot/db/compression/zstd.py
tdesveaux: I interpreted the note as this API is for compatibility wit Python's stdlib compression APIs, so use it as such.
If possible, prefer other APIs.

Here, we could use `ZstdCompressionChunker`, and either implement the compress/flush calls joining the iterable returned.
Or change the Interface return type to `Generator[bytes, None, None]`, but that would change all implementation that do not benefit from it.

So I used this compatibility layer since it felt it made more sense for this use case.

tdesveaux:
> > This will probably hurt performance in simple scenarios (master instances without much to do).
> 
> Please fix my understanding where I'm incorrect. ...

Yes, sorry for my poor phrasing.
It's exactly as you said, in a context of a whole master, performances should be overall better (fairly sure of this, but it'd need to be profiled to confirm).
Only impact should be when taking into account ONLY the call to `compressLog`, and not on performance but on throughput as you said.

master/buildbot/db/compression/zstd.py
tdesveaux: Sure, I'll dig into this more to at least leave a better comment.
I only did a quick search on this in zstandard issue which did not lead to a concrete reason for this.

tdesveaux: Found an issue that explained the problem, updated the comment with a link.

tdesveaux:
Leaving as draft for now as I have some concern about lifetime of things that could impact the changes made here (due to my lack of knowledge on twisted ThreadPool and Buildbot's lifetime of services).

p12tic:
Rebased just to see if changes in master branch helps with regards to crashes.

p12tic:
Database layer (e.g. the same instance that holds the thread pool) could have a DeferWaiter instance to which all calls to log compression are added. Then during database system shutdown we would wait until the compression actions finish. Would that address the problem?

tdesveaux:
> Database layer (e.g. the same instance that holds the thread pool) could have a DeferWaiter instance to which all calls to log compression are added. Then during database system shutdown we would wait until the compression actions finish. I think this should address the problem.

Perfect, thanks!

master/buildbot/db/connector.py
```diff
@@ -148,6 +149,8 @@ def setup(self, check_version=True, verbose=True):
         self._engine = enginestrategy.create_engine(db_url, basedir=self.basedir)
         self.upsert = get_upsert_method(self._engine)
         self.pool = pool.DBThreadPool(self._engine, reactor=self.master.reactor, verbose=verbose)
+        self._db_tasks_waiter = DeferWaiter()
+        self.master.reactor.addSystemEventTrigger('before', 'shutdown', self._db_tasks_waiter.wait)
```
p12tic, line 153: I think `stopService()` is enough. It is a bug to not wait for `stopService()` to finish before shutting down the reactor.

tdesveaux:
I'm pretty confused as to why it's consistently the same 4 tests that fails, but I can't find what's unique to them. They seem to be the same as all other tests on steps...

```
buildbot.test.unit.test_download_secret_to_worker.TestDownloadFileSecretToWorkerCommand.testBasic
buildbot.test.unit.test_contrib_buildbot_cvs_mail.TestBuildbotCvsMail.test_buildbot_cvs_mail_with_unknown_opt_exits_with_error
buildbot.test.unit.test_contrib_buildbot_cvs_mail.TestBuildbotCvsMail.test_buildbot_cvs_mail_without_cvsroot_opt_exits_with_error
buildbot.test.unit.test_steps_git_diffinfo.TestDiffInfo.test_empty_diff
```

master/buildbot/db/connector.py
tdesveaux: I'm not particularly happy with this.
Lifetime of services are pretty confusing and all over the place.

`DBThreadPool` (and compression pool now) is stopped on the `during, shutdown` event of the reactor.

But this is "stopped" on `DBConnector.stopService`.
I also had to make a `_shutdown` method since, in tests, services are not started nor then stopped.

I feel it'd be more consistent to either:
- `self._db_tasks_waiter.wait()` on `before, shutdown` of the reactor
- add start/stop methods to `DBConnectorComponent` / `DBThreadPool` and call them from the `DBConnector` `startService` / `stopService`.

PR #8141, opened 2024-10-18
bbtravis: fix treq pinned version for twisted 22.4.0 backward compat test

tdesveaux:
treq 23.8.0 does not exist on pypi, nor on their github release.

Use latest version that does not require twisted 22.10.0.

## Contributor Checklist:


p12tic:
Ugh, I think I was thinking about Twisted here.

PR #8148, opened 2024-10-21
test: Make fakemaster.make_master() return a Deferred

p12tic:
This resolves a potential race condition and makes it possible to add more complex functionality into make_master.


p12tic:
cc @tdesveaux

PR #8150, opened 2024-10-21
Tear down TestReactorMixin explicitly, not using addCleanup()

p12tic:
`addCleanup` runs before `tearDown`. Tearing down TestReactorMixin in `addCleanup` functions results in database being unavailable in `tearDown`, so any shutdown code there will most likely fail. This PR introduces a way to control shutdown of TestReactorMixin explicitly so that it's possible to run it as the very last thing.

## Contributor Checklist:


p12tic:
cc @tdesveaux - any shutdown problems you saw may be caused by the problem this PR fixes.

PR #8157, opened 2024-10-23
db: appendLog/getLogLines compression outside DB thread

tdesveaux:
Follow #8140, same work, on remaining methods.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/db/test_logs.py
+++ b/master/buildbot/test/unit/db/test_logs.py
@@ -136,14 +136,16 @@ class Tests(interfaces.InterfaceTests):
             'another line',
             'yet another line',
         ]
+
+        def _join_lines(lines: list[str]):
+            return ''.join(e + '\n' for e in lines)
+
         for first_line in range(0, 7):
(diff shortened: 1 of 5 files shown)


master/buildbot/test/unit/db/test_logs.py
```diff
@@ -136,14 +136,16 @@ def checkTestLogLines(self):
             'another line',
             'yet another line',
         ]
+
+        def _join_lines(lines: list[str]):
```
p12tic, line 141: I think this should be a separate commit. I got confused that the behavior is being changed, where in fact it is not.

p12tic (changes requested):
Thanks for the PR. It looks great, I just had a single nit.

p12tic (approved):

tdesveaux:
@p12tic FYI, deployed this on a master that sees a lot of jobs running at the same time with fairly large logs.

It resulted in a bottleneck on the main thread staying at 100% usage, so there is an issue with this.
I'll investigate this week, but in the mean time, this should not be released.

PR #8158, opened 2024-10-24
Misc fixes and improvements

tdesveaux:
Some random smaller patches I did while working on other stuff.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/workerforbuilder.py
+++ b/master/buildbot/process/workerforbuilder.py
@@ -86,7 +86,7 @@ class AbstractWorkerForBuilder:
             self.worker.buildFinished(self)
 
     @defer.inlineCallbacks
-    def attached(self, worker, commands):
+    def attached(self, worker: AbstractWorker, commands):
         """
         @type  worker: L{buildbot.worker.Worker}
         @param worker: the Worker that represents the worker as a whole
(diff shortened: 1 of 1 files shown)


master/buildbot/process/workerforbuilder.py
```diff
@@ -221,7 +222,8 @@ def attached(self, worker, commands):
         return super().attached(worker, commands)
 
     def substantiate(self, build):
-        return self.worker.substantiate(self, build)
+        if self.worker:
+            return self.worker.substantiate(self, build)
```
p12tic, line 226: I think we should keep appropriate return when the condition is not matched, probably `defer.succeed(None)`

master/buildbot/process/build.py
```diff
@@ -577,37 +578,26 @@ def addStepsAfterLastStep(self, step_factories):
         # Add the new steps to the end.
         self.steps.extend(self.setupBuildSteps(step_factories))
 
-    def getNextStep(self):
+    def getNextStep(self) -> IBuildStep | None:
         """This method is called to obtain the next BuildStep for this build.
-        When it returns None (or raises a StopIteration exception), the build
+        When it returns None, the build
         is complete."""
```
p12tic, line 585: This line could be merged in the previous line

master/buildbot/process/build.py
```diff
@@ -577,37 +578,26 @@ def addStepsAfterLastStep(self, step_factories):
         # Add the new steps to the end.
         self.steps.extend(self.setupBuildSteps(step_factories))
 
-    def getNextStep(self):
+    def getNextStep(self) -> IBuildStep | None:
         """This method is called to obtain the next BuildStep for this build.
-        When it returns None (or raises a StopIteration exception), the build
```
p12tic, line 582: I think it makes sense to add to commit description that the behavior of handling StopIteration exception is not documented, thus it's fine to consider this internal API and change it without deprecation.

p12tic (changes requested):
LGTM, but I had several easy to address comments.

master/buildbot/process/workerforbuilder.py
tdesveaux: Good catch.
Used `defer.succeed(False)` since `worker.substantiate` returns a `Deferred[bool]`.

p12tic (approved):

PR #8163, opened 2024-10-29
workers bulk actions: Search workers via regex

mifo123:
I modified the 'MultipleWorkersActionsModal' component and fixed the functionality for selecting all workers (selecting all filtered workers). I also added the option to search using regex.

image

### DIFF
--- a/www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
+++ b/www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
@@ -19,7 +19,7 @@ import {observer} from "mobx-react";
 import {useState} from "react";
 import {Button, Modal} from "react-bootstrap";
 import {Worker} from "buildbot-data-js";
-import Select, { ActionMeta, MultiValue } from 'react-select';
+import Select, { ActionMeta, MultiValue, InputActionMeta  } from 'react-select';
 
 type MultipleWorkersActionsModalProps = {
   workers: Worker[];
@@ -44,6 +44,7 @@ export const MultipleWorkersActionsModal = observer(({workers, preselectedWorker
   const [errors, setErrors] = useState<string|null>(null);
   const [reasonText, setReasonText] = useState<string>("");
   const [selectedWorkers, setSelectedWorkers] = useState<Worker[]>(preselectedWorkers);
+  const [menuIsOpen, setMenuIsOpen] = useState<boolean>();
 
   const stopDisabled = selectedWorkers.length <= 0 || selectedWorkers.every(w => w.connected_to.length === 0)
   const pauseDisabled = selectedWorkers.length <= 0 || selectedWorkers.every(w => w.paused)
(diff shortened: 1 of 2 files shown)


www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
```diff
@@ -65,6 +66,18 @@ export const MultipleWorkersActionsModal = observer(({workers, preselectedWorker
     }
   }
 
+  const onInputChange = (
+    inputValue: string,
+    { action, prevInputValue }: InputActionMeta
+  ) => {
+    if (action === 'input-change') return inputValue;
```
p12tic, line 133: Please add explicit braces.

www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
```diff
@@ -65,6 +66,18 @@ export const MultipleWorkersActionsModal = observer(({workers, preselectedWorker
...
+    if (action === 'menu-close') {
+      if (prevInputValue) setMenuIsOpen(true);
```
p12tic, line 135: Please add explicit braces. Maybe `setMenuIsOpen(prevInputValue ? true : undefined)` could also be concise yet readable enough?

www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
```diff
@@ -36,6 +37,42 @@ const workerToSelectOption = (worker: Worker): SelectOption => {
   return { value: worker, label: worker.name }
 }
 
+const Control = (props: any) => {
+  const { useRegexSearch, handleToggleClick, selectRef } = props;
```
p12tic: Better not use `use` as a prefix for variables, because React convention is to `use` for hooks. https://legacy.reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook

www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
```diff
@@ -36,6 +37,42 @@ const workerToSelectOption = (worker: Worker): SelectOption => {
   return { value: worker, label: worker.name }
 }
 
+const Control = (props: any) => {
```
p12tic, line 41: Control should be a more explicit name, it's hard to figure out what exactly `Control` refers to.

p12tic, line 41: Props should be typed. Probably makes sense to check what props `components.Control` accepts and then add additional members to it.

www/base/src/components/MultipleWorkersActionsModal/MultipleWorkersActionsModal.tsx
```diff
@@ -38,23 +39,43 @@ const workerToSelectOption = (worker: Worker): SelectOption => {
...
+
+  const [ isHoveredRegexSearch, setIsHoveredRegexSearch ] = useState(false);
+
+  const [isHoveredSelectAll, setIsHoveredSelectAll] = useState(false);
+  const [isActiveSelectAll, setIsActiveSelectAll] = useState(false);
 
   return (
     <components.Control {...props}>
-      <div style={{ display: "flex", alignItems: "center", width: "100%" }}>
+      <div className="flex-container">
```
p12tic, line 51: Probably should be named `bb-workers-actions-modal-control-container`. `flex-container` is too generic and may clash with other libraries.

p12tic (changes requested):
Looks good overall, I had comments mostly about stylistic issues.

Unrelatedly, please prefix commit messages with `www: ` as they change code in that area exclusively.

p12tic (approved):

PR #8164, opened 2024-10-29
Fixes issue #8162

### LINKED ISSUE #8162: [buildbot-worker] [WIN32] pywintypes.error: (5, 'AssignProcessToJobObject', 'Access is denied.')
It is possible that spawned process already finished before calling `AssignProcessToJobObject`. 
As result `_spawnProcess` fails with exception:
```
2024-10-26 13:16:47+0300 [Broker,client] error in RunProcess._startCommand
          File "c:\buildworker\venv\lib\site-packages\buildbot_worker\pb.py", line 323, in remote_startCommand
            d = self.protocol_command.command.doStart()
          File "c:\buildworker\venv\lib\site-packages\buildbot_worker\commands\base.py", line 162, in doStart
            d = defer.maybeDeferred(self.start)
          File "c:\buildworker\venv\lib\site-packages\twisted\internet\defer.py", line 212, in maybeDeferred
            result = f(*args, **kwargs)
          File "c:\buildworker\venv\lib\site-packages\buildbot_worker\commands\shell.py", line 50, in start
            d = self.command.start()
          File "c:\buildworker\venv\lib\site-packages\buildbot_worker\runprocess.py", line 461, in start
            self._startCommand()
          File "c:\buildworker\venv\lib\site-packages\buildbot_worker\runprocess.py", line 582, in _startCommand
            self.process = self._spawnProcess(
          File "c:\buildworker\venv\lib\site-packages\buildbot_worker\runprocess.py", line 641, in _spawnProcess
            win32job.AssignProcessToJobObject(self.job_object, pHandle)
        pywintypes.error: (5, 'AssignProcessToJobObject', 'Access is denied.')
```

simple test code:
```
import win32job
import win32api
import win32con
import subprocess


def create_job_object():
        job = win32job.CreateJobObject(None, "")
        extented_info = win32job.QueryInformationJobObject(
            job, win32job.JobObjectExtendedLimitInformation
        )
        extented_info['BasicLimitInformation']['LimitFlags'] = (
            win32job.JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE
            | win32job.JOB_OBJECT_TERMINATE_AT_END_OF_JOB
        )
        win32job.SetInformationJobObject(
            job, win32job.JobObjectExtendedLimitInformation, extented_info
        )
        return job

process = subprocess.Popen(["cmd.exe", "/c", "echo", "123"])

process.wait()

pHandle = win32api.OpenProcess(win32con.PROCESS_ALL_ACCESS, False, int(process.pid))
job_object = create_job_object()
win32job.AssignProcessToJobObject(job_object, pHandle)

```

mifo123:
Fix Access is denied error on WIN32 when assigning processes to job objects. On Windows, attempting to assign an already-exited process to a job object fails with an Access is denied error. This change checks if the process is still active before calling AssignProcessToJobObject, avoiding attempts to assign exited processes, which resolves the error.

`win32process.GetExitCodeProcess(pHandle) == win32con.STILL_ACTIVE:` will check if return code is 259 (macro STILL_ACTIVE) which indicate that process does not exit yet, and it is safe to assign in to the job.

## Contributor Checklist:

### DIFF
--- a/worker/buildbot_worker/runprocess.py
+++ b/worker/buildbot_worker/runprocess.py
@@ -54,6 +54,7 @@ if runtime.platformType == 'win32':
     import win32api
     import win32con
     import win32job
+    import win32process
 
 
 def win32_batch_quote(cmd_list, unicode_encoding='utf-8'):
@@ -638,10 +639,10 @@ class RunProcess:
                     processProtocol, executable, args, env, path, usePTY=usePTY
                 )
             pHandle = win32api.OpenProcess(win32con.PROCESS_ALL_ACCESS, False, int(process.pid))
-
-            # use JobObject to group subprocesses
-            self.job_object = self._create_job_object()
-            win32job.AssignProcessToJobObject(self.job_object, pHandle)
+            if win32process.GetExitCodeProcess(pHandle) == win32con.STILL_ACTIVE:
+                # use JobObject to group subprocesses
+                self.job_object = self._create_job_object()
+                win32job.AssignProcessToJobObject(self.job_object, pHandle)
             return process
 
         # use the ProcGroupProcess class, if available
--- a/newsfragments/8162.bugfix
+++ b/newsfragments/8162.bugfix
@@ -1,0 +1,1 @@
+Fixes an `Access is denied` error on WIN32 when calling `AssignProcessToJobObject`. (:issue:`8162`)
--- a/worker/buildbot_worker/test/unit/test_runprocess.py
+++ b/worker/buildbot_worker/test/unit/test_runprocess.py
@@ -554,6 +554,21 @@ class TestRunProcess(BasedirMixin, unittest.TestCase):
         d = s._spawnAsBatch(s.pp, s.command, "args", tempEnviron, "path", False)
         return d
 
+    @defer.inlineCallbacks
+    @compat.skipUnlessPlatformIs("win32")
+    def test_assign_exited_process(self):
+        s = runprocess.RunProcess(
+            0,
+            ["cmd.exe", "/c", "exit"],
+            self.basedir,
+            'utf-8',
+            self.send_update,
+        )
+        yield s.start()
+
+        # Assert that the process completed successfully
+        self.assertTrue(('rc', 0) in self.updates, self.show())
+
     def test_spawnAsBatchCommandString(self):
         return self._test_spawnAsBatch("dir c:/", "cmd.exe")
 


worker/buildbot_worker/runprocess.py
```diff
@@ -638,10 +639,10 @@ def _spawnProcess(
                     processProtocol, executable, args, env, path, usePTY=usePTY
                 )
             pHandle = win32api.OpenProcess(win32con.PROCESS_ALL_ACCESS, False, int(process.pid))
-
-            # use JobObject to group subprocesses
-            self.job_object = self._create_job_object()
-            win32job.AssignProcessToJobObject(self.job_object, pHandle)
+            if win32process.GetExitCodeProcess(pHandle) == win32con.STILL_ACTIVE:
```
tdesveaux, line 642: might be worth it to look into the `CREATE_SUSPENDED` flag to handle cases where a command will directly spawn a process and exit (although this is good for a quick fix).

mifo123: is it even possible with reactor.spawnProcess to use flags? If no, it can be very challenging rewrite this part to win32process.

tdesveaux: If [this](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/_dumbwin32proc.py#L191-L194) is the implementation used by twisted for process on Windows (not sure, only did a quick search), it doesn't look like it.
Probably not event possible to subclass it since everything is done in the init...

mifo123: I also did not find anything. So is this sufficient fix or need to be rework some other way?

p12tic:
Would it be possible to explain how this commit fixes the problem (here and in commit description)? Without reading Windows documentation it's difficult to understand what's happening which means that every person who wants to understand the change in the git history will need to open Windows documentation and read enough of it to understand the context.

mifo123:
> Would it be possible to explain how this commit fixes the problem (here and in commit description)? Without reading Windows documentation it's difficult to understand what's happening which means that every person who wants to understand the change in the git history will need to open Windows documentation and read enough of it to understand the context.

Sure, commit message updated, description in this PR as well.

p12tic:
> Sure, commit message updated, description in this PR as well.

Thanks a lot!

I rebased the commit and added more descriptive commit title as well.

p12tic (approved):

PR #8166, opened 2024-10-31
db: fix logs compression tasks blocking main thread

tdesveaux:
As mentioned [here](https://github.com/buildbot/buildbot/pull/8157#issuecomment-2447845229), this fix a pretty serious issue introduced in #8157.

Running a function which is a Generator in a thread, will result in the Generator work being done on the consumer thread.

Solution here is to use a queue to communicate between the threads. Capped at 1 element so that, if for some reason, the DB thread is not quick enough to flush the queue, the compression thread will only have one chunk in memory at a time.

Annoyingly, in tests, the (~`NullPool`~ -> )`NonThreadPool` is used, which run the workload on the current thread, cause a deadlock, leading to the uncapping of the queue conditioned on `get_is_in_unit_tests`.

I could modify (~`NullPool`~ -> )`NonThreadPool` to be have a single thread (maybe one per test case to handle `trial -jX` runs). (EDIT: integration tests do not use `NonThreadPool`, so we can rely on them)

## Contributor Checklist:

### DIFF
--- a/master/buildbot/db/logs.py
+++ b/master/buildbot/db/logs.py
@@ -18,14 +18,19 @@ from __future__ import annotations
 import dataclasses
 import io
 import os
+import threading
+from functools import partial
 from typing import TYPE_CHECKING
+from typing import cast
 
 import sqlalchemy as sa
 from twisted.internet import defer
 from twisted.internet import threads
 from twisted.python import log
 from twisted.python import threadpool
+from twisted.python.failure import Failure
 
+from buildbot.config.master import get_is_in_unit_tests
 from buildbot.db import base
 from buildbot.db.compression import BrotliCompressor
 from buildbot.db.compression import BZipCompressor
@@ -45,6 +50,7 @@ if TYPE_CHECKING:
     from typing import TypeVar
 
     from sqlalchemy.engine import Connection as SAConnection
+    from twisted.internet.interfaces import IReactorThreads
     from typing_extensions import ParamSpec
 
     _T = TypeVar('_T')
@@ -293,11 +299,15 @@ class LogsConnectorComponent(base.DBConnectorComponent):
                     line_idx += 1
 
         async for chunk_first_line, _, compressed, content in _iter_chunks_batched():
-            for line in await self._defer_to_compression_pool(
-                _iter_uncompress_lines,
-                chunk_first_line,
-                compressed,
-                content,
+            async for line in _async_iter_on_pool(
+                partial(
+                    _iter_uncompress_lines,
+                    chunk_first_line=chunk_first_line,
+                    compressed=compressed,
+                    content=content,
+                ),
+                reactor=self.master.reactor,
+                provider_threadpool=self._compression_pool,
             ):
                 yield line
 
@@ -406,9 +416,9 @@ class LogsConnectorComponent(base.DBConnectorComponent):
 
             return compressed_chunk, compressor_id, len(lines)
 
-        async def _thd_iter_chunk_compress(
+        def _thd_iter_chunk_compress(
             content: str,
-        ) -> AsyncGenerator[tuple[bytes, int, int], None]:
+        ) -> Generator[tuple[bytes, int, int], None]:
             """
             Split content into chunk delimited by line-endings.
             Try our best to keep chunks smaller than MAX_CHUNK_SIZE
@@ -469,9 +479,13 @@ class LogsConnectorComponent(base.DBConnectorComponent):
             compressed_chunk,
             compressed_id,
             chunk_lines_count,
-        ) in await self._defer_to_compression_pool(
-            _thd_iter_chunk_compress,
-            content,
+        ) in _async_iter_on_pool(
+            partial(
+                _thd_iter_chunk_compress,
+                content=content,
+            ),
+            reactor=self.master.reactor,
+            provider_threadpool=self._compression_pool,
         ):
             last_line = chunk_first_line + chunk_lines_count - 1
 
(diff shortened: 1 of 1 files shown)


master/buildbot/db/logs.py
```diff
@@ -750,3 +759,71 @@ def _model_from_row(self, row):
...
+) -> AsyncGenerator[_T, None]:
+    """
+    Utility to have a simple function returning an AsyncGenerator
+    when the workload yielding should be executed on a thread.
+    """
+    # create a single element queue as to
+    # occupy a thread of the pool
+    # avoiding too many compressed chunks in memory awaiting DB insert
+    # use 0 (unlimited) in tests as there isn't really a threadpool / reactor running
+    queue_maxsize = 1 if not get_is_in_unit_tests() else 0
```
p12tic, line 780: This function will be used for decompressing lines as well, right?

If yes, the queue_maxsize should come as an argument so that it would be possible to customize it on case by case basis. Higher queue size will be useful when the relative size of each item is small. In such case it is likely that performance would be significantly reducing due to threads blocking one another. The sender thread would repeatedly block on the queue being full and receiving thread would repeatedly block on the queue being empty. Only when sending and receiving threads are perfectly synchronized the amount of such blocks will be small.

tdesveaux: > This function will be used for decompressing lines as well, right?

yes, it's used in both compression (appendLog) and decompression (iter_log_lines, getLogLines) contexts.

> the queue_maxsize should come as an argument

I think it makes sense to provide flexibility for sure.

> The sender thread would repeatedly block on the queue being full and receiving thread would repeatedly block on the queue being empty. Only when sending and receiving threads are perfectly synchronized the amount of such blocks will be small.

At the moment, the thread that put in queue will block when it has done work and is ready to put in queue.
So as soon as consumer thread get the pending item from the queue, the producer thread will add the next item and go on to work on the next item.
Rationale here was to limit memory usage, at the cost of throughput.

A higher size would make sense to improve throughput, just not sure how to profile a valid use case to tweak it.

I used this setup to do some basic profiling when working on the fix: https://github.com/tdesveaux/buildbot-repro/tree/perf-regression-from-8157

(unrelated note, `RemoteCommand.split_line`, called from `RemoteCommand.remote_update` is blocking the mainthread in `LineBoundaryFinder.append`'s call to `re.Pattern.sub`, which should probably moved to a thread).

p12tic: > A higher size would make sense to improve throughput, just not sure how to profile a valid use case to tweak it.

I think it's reasonable to use gut feeling in this case. Log chunks are likely large, so in such case the queue size can be something like 10 (the extra memory size will still be less than 10 megabytes, so can still be ignored, as buildbot master is unlikely to have less than 1GB available). For something small like lines we can use larger number like 10000.

tdesveaux: https://github.com/buildbot/buildbot/pull/8166/commits/32ee89e35eda623bd6207f2b24265562e601404d

went with slightly different values, and commented on the rationale behind them.

tdesveaux:
@p12tic sorry, made some changes to improve usage of `_async_iter_on_pool`: https://github.com/buildbot/buildbot/compare/5db428c7fe61f10b5b6977d0fe849ba509e450a6..4bcf80f52cc016ff3095963e532ead2192148f31

Main thing was to go back to take a simple `Generator` and hide internal queue communication.
I think it provide a better and simpler interface for callers.

`_async_iter_on_pool` is now pretty agnostic to usage in `db/logs`, maybe I should move it to it's own `buildbot/util/` file?

master/buildbot/db/logs.py
```diff
@@ -469,9 +481,20 @@ def _truncate_line(line: bytes) -> bytes:
...
+                content=content,
+            ),
+            reactor=self.master.reactor,
+            provider_threadpool=self._compression_pool,
+            # In theory, memory usage could grow to:
+            # MAX_CHUNK_SIZE * max_backlog PER thread (capped by _compression_pool.maxthreads)
+            # with:
+            #   - MAX_CHUNK_SIZE = 64KB
+            #   - max_backlog = 100
+            # ~6MB per threads
```
p12tic, line 496: typo. threads -> thread

p12tic (approved):

p12tic:
> _async_iter_on_pool is now pretty agnostic to usage in db/logs, maybe I should move it to it's own buildbot/util/ file?

Either is fine, please do as you find better.

p12tic:
Tests fail on python 3.12 with:

```
Error: 
  File "C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages\twisted\internet\defer.py", line 2010, in _inlineCallbacks
    result = context.run(
  File "C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages\twisted\python\failure.py", line 549, in throwExceptionIntoGenerator
    return g.throw(self.value.with_traceback(self.tb))
  File "d:\a\buildbot\buildbot\master\buildbot\process\log.py", line 83, in addRawLines
    yield self.lock.run(lambda: self.master.data.updates.appendLog(self.logid, lines))
  File "C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages\twisted\internet\defer.py", line 2010, in _inlineCallbacks
    result = context.run(
  File "C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages\twisted\python\failure.py", line 549, in throwExceptionIntoGenerator
    return g.throw(self.value.with_traceback(self.tb))
  File "d:\a\buildbot\buildbot\master\buildbot\data\logs.py", line 144, in appendLog
    res = yield self.master.db.logs.appendLog(logid=logid, content=content)
  File "C:\hostedtoolcache\windows\Python\3.12.7\x64\Lib\site-packages\twisted\internet\defer.py", line 2014, in _inlineCallbacks
    result = context.run(gen.send, result)
  File "d:\a\buildbot\buildbot\master\buildbot\db\logs.py", line 480, in appendLog
    async for (
builtins.TypeError: cannot unpack non-iterable NoneType object

buildbot.test.integration.test_steps_configurable.BuildbotTestCiTest.test_buildbot_ci
```

tdesveaux:
> Tests fail on python 3.12 with:

Yeah, looks to be something flaky.
Don't see an obvious cause, and am failing to reproduce locally.

I'll dig into it some more.

p12tic:
Let's see if this was just one accident or we will see more failures in CI.

p12tic:
After all the intermittent failure is rather frequent, e.g. https://buildbot.buildbot.net/#/builders/89/builds/16570 and https://buildbot.buildbot.net/#/builders/89/builds/16573

tdesveaux:
Still no luck on local reproduction.
I'll setup a local VM with hardware constraints to see if it helps.

Still no idea where a None could come from staring at the code.

PR #8195, opened 2024-11-04
Fix issues in "Scheduled biweekly dependency update for week 44"

pmisik:
This PR is based on #8193 with these modifications:
* removed `Update mypy from 1.12.1 to 1.13.0` due to mypy-zope upgrade issue - see https://github.com/Shoobx/mypy-zope/pull/124
* conditional update of werkzeug due to 3.1.0 dropped support for Python 3.8 see https://github.com/pallets/werkzeug/blob/main/CHANGES.rst#version-310

## Contributor Checklist:


p12tic:
I think it makes sense to extract ruff upgrade to separate commit because it wants to reformat >50 files.

pmisik:
> I think it makes sense to extract ruff upgrade to separate PR because it wants to reformat >50 files.

I agree. I'll do it.

PR #8203, opened 2024-11-07
Use standard URL syntax for Git steps to enable the use of a dedicated SSH port

fabianjuelich:
- Added function for converting from SCP style to URL syntax and passing the repository address to it when setting up the Git step
- Added a missing default value in the documentation of the 'branch' parameter for the source step 'Git'
- Fixed typo in 'newsfragments' README

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -83,6 +83,7 @@ class Git(Source, GitStepMixin):
     def __init__(
         self,
         repourl=None,
+        port=22,
         branch='HEAD',
         mode='incremental',
         method=None,
@@ -111,6 +112,7 @@ class Git(Source, GitStepMixin):
         self.branch = branch
         self.method = method
         self.repourl = repourl
+        self.port = port
         self.reference = reference
         self.retryFetch = retryFetch
         self.submodules = submodules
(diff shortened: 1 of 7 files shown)


master/buildbot/util/git.py
```diff
@@ -66,6 +66,14 @@ def getSshCommand(keyPath, knownHostsPath):
     return ' '.join(command)
 
 
+def scpStyleToUrlSyntax(address, port=22, scheme='ssh'):
```
p12tic, line 69: use snake_case

master/buildbot/steps/source/git.py
```diff
@@ -111,6 +112,7 @@ def __init__(
         self.branch = branch
         self.method = method
         self.repourl = repourl
+        self.port = port
```
p12tic, line 115: This probably needs to be added to `GitStepMixin` as evidenced by the tests that are failing.

master/buildbot/steps/source/git.py
```diff
@@ -632,6 +635,7 @@ def __init__(
     ):
         self.workdir = workdir
         self.repourl = repourl
+        self.port = port,
```
p12tic, line 638: Extra comma. This is actually a tuple.

PR #8224, opened 2024-11-25
www: some REST response improvements

tdesveaux:
We have a Builder that never had a success.

Upon user visiting the build page, call to `builds/{buildid}/changes` would try to load about the full history of changes for the (long-lived) project.
Response would result in a huge list of changes, that needed to be json-encoded/compressed, all that on the main thread.
Since this would take too long, request would timeout, but the master would keep working on it.

So this PR does two things:
- check that the connection is still alive between json encode chunks, so work can be aborted on lost connection
- defer json-encode to thread to avoid blocking the main thread

## Contributor Checklist:

### DIFF
--- a/master/buildbot/www/resource.py
+++ b/master/buildbot/www/resource.py
@@ -69,7 +69,7 @@ class Resource(resource.Resource):
         if writeError is None:
             writeError = writeErrorDefault
         try:
-            d = _callable(request)
+            d = defer.maybeDeferred(_callable, request)
         except Exception as e:
             d = defer.fail(e)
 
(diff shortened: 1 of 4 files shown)


master/buildbot/www/resource.py
```diff
@@ -69,7 +69,7 @@ def writeErrorDefault(msg, errcode=400):
         if writeError is None:
             writeError = writeErrorDefault
         try:
-            d = _callable(request)
+            d = defer.maybeDeferred(_callable, request)
```
p12tic, line 72: seems like a good candidate for rewrite to `defer.inlineCallbacks` or coroutine. Not for this PR though.

p12tic (approved):

PR #8243, opened 2024-12-05
www: Fix crash in build view when clicking on changes tab

p12tic:
The changeid field is insufficient to populate ChangesTable.

cc @tdesveaux


tdesveaux:
@p12tic missed the mention, sorry about this issue :grimacing: 

IIRC I was trying to only query fields necessary to avoid loading comments / files of changes since they can be quite large in our use case.

PR #8260, opened 2024-12-10
Remove GraphQL support

p12tic:
3 years passed since initial code has been implemented, but unfortunately this wasn't completed. The Buildbot web frontend still uses HTTP queries with JSON results. Supporting GraphQL properly and reaping its benefits will require almost complete web frontend rewrite which is unlikely to happen.

No release note has been added because GraphQL support has never been officially released.


p12tic:
cc @tardyp 

It's unfortunate that this work couldn't be used... Maybe someday with renewed interest in Buildbot we can resurrect the code.

PR #8263, opened 2024-12-10
Update twisted from 24.10.0 to 24.11.0

pmisik:
## Contributor Checklist:

### DIFF
--- a/requirements-ci.txt
+++ b/requirements-ci.txt
@@ -73,7 +73,7 @@ six==1.16.0
 SQLAlchemy==2.0.36
 treq==24.9.1;  python_version >= "3.9"
 treq==22.2.0;  python_version < "3.9" # pyup: ignore
-Twisted==24.10.0
+Twisted==24.11.0
 txaio==23.1.1
 txrequests==0.9.6
 typing_extensions==4.12.2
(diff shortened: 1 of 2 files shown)


pmisik:
Do not merge yet.
It looks like mypy check is incorrectly implemented.
See errors in https://buildbot.buildbot.net/#/builders/148/builds/575/steps/9/logs/stdio

worker/buildbot_worker/test/fake/reactor.py
```diff
@@ -83,7 +83,7 @@ def run(self) -> None:
     def running(self) -> bool:
         raise NotImplementedError()
 
-    def resolve(self, name: str, timeout: Sequence[int]) -> defer.Deferred[str]:
+    def resolve(self, name: str, timeout: Sequence[int] = (1, 3, 11, 45)) -> defer.Deferred[str]:
```
p12tic, line 86: We should ignore this line in terms of typing because mypy will fail if other version of Twisted is installed (e.g. on developer machine).

p12tic:
>  It looks like mypy check is incorrectly implemented.

I think this PR should be able to go in still? It's not like it created these problems.

pmisik:
> I think this PR should be able to go in still? It's not like it created these problems.

This is correct. 
I just put warning as I was surprised this PR got `All checks have passed` as I have seen errors in mypy errors on master while I "fixed" workers mypy errors.

It looks like https://github.com/buildbot/buildbot/blob/master/.bbtravis.yml#L186-L189
```
    cmd:  |
      export PATH=/tmp/bbvenv/bin/:$PATH
      (cd ./master && mypy --config-file ../pyproject.toml buildbot)
      (cd ./worker && mypy --config-file ../pyproject.toml buildbot_worker)
```
is evaluated in `or` manner while it should be valuated in `and` manner

p12tic:
Rebased.

PR #8275, opened 2024-12-19
[DONTNOD] Display build summary label as flex to fix display issue

tdesveaux:
#23


tdesveaux:
My bad, was support to target our fork

PR #8278, opened 2024-12-25
Don't apply monkey patches on generic start

lephilousophe:
Since commit d6b698e, the monkey patches were applied on every installs.

This implies that set_is_in_unit_tests(True) was called even when not it in unit tests.

Because it believes being in tests, db.logs._async_iter_on_pool is importing NonThreadPool
from a submodule in the test module.
This runs the test __init__ code which sets up warnings as error by default.

As a consequence, any Deprecation warning will trigger an exception 
instead of just being shown in logs.

This is causing troubles in current v4.2.0 with out of tree modules (which use deprecated code, granted).
A backport would be great.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/monkeypatches/__init__.py
+++ b/master/buildbot/monkeypatches/__init__.py
@@ -72,8 +72,9 @@ def patch_config_for_unit_tests():
     set_is_in_unit_tests(True)
 
 
-def patch_all():
+def patch_all(for_tests=False):
+    if for_tests:
+        patch_testcase_timeout()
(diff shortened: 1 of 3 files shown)


p12tic:
I think it would make sense to bring back `for_tests` argument and keep enabling `patch_servicechecks` and `patch_decorators` on normal runs.

lephilousophe:
Done.

p12tic (approved):

p12tic:
I added release note and also copied PR description to the commit comment.

Issue #8285, opened 2025-01-10
Skipped Builds


Mic92:
Just now it's possible skip build steps, but it would be cool to actually skip builds instead.
Reason is that even if all build steps are skipped, it still needs to reach out to the worker which can take 2s. Also when using large build matrices it's not visible without clicking on a build that all builds have been skipped.

We would like to implement the feature by adding a `doBuildIf` predicate in the same way we have a `doStepIf`.

p12tic:
Fixed in #8288

PR #8288, opened 2025-01-10
Add `Build.skipBuildIf`

MagicRB:
## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -34,6 +34,7 @@ from buildbot.process.results import CANCELLED
 from buildbot.process.results import EXCEPTION
 from buildbot.process.results import FAILURE
 from buildbot.process.results import RETRY
+from buildbot.process.results import SKIPPED
 from buildbot.process.results import SUCCESS
 from buildbot.process.results import WARNINGS
 from buildbot.process.results import computeResultAndTermination
@@ -320,6 +321,10 @@ class Build(properties.PropertiesMixin):
         finishes. This Deferred is guaranteed to never errback."""
         self.workerforbuilder = workerforbuilder
         self.conn = None
+        if self.builder.config is not None:
+            self.do_build = self.builder.config.do_build_if(self)
+        else:
+            self.do_build = True
 
         worker = workerforbuilder.worker
         assert worker is not None
@@ -329,7 +334,7 @@ class Build(properties.PropertiesMixin):
         self.workername = worker.workername
         self.worker_info = worker.info
 
-        log.msg(f"{self}.startBuild")
+        log.msg(f"{self}.startBuild{' skipped' if not self.do_build else ''}")
 
         # TODO: this will go away when build collapsing is implemented; until
         # then we just assign the build to the first buildrequest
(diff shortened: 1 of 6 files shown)


master/buildbot/process/build.py
```diff
@@ -395,99 +400,103 @@ def startBuild(self, workerforbuilder: AbstractWorkerForBuilder):
...
+                    self._locks_acquire_step.stepid, SUCCESS, False
+                )
 
         yield self.master.data.updates.setBuildStateString(self.buildid, 'building')
 
+        if self.skipBuild:
+            self.results = SKIPPED
+
         # start the sequence of steps
         self.startNextStep()
```
p12tic, line 501: I think we shouldn't attempt to execute steps. Since `.conn` is not set, `startNextStep()` will execute `allStepsDone` immediately. Do we even want steps to be created at all?

Mic92:
Related to this one: https://github.com/buildbot/buildbot/issues/8285

master/buildbot/process/build.py
Mic92: Probably we shouldn't create steps. This also should be lighter on the database.

MagicRB: I was hoping we could get them to show up the UI just for consistency, but yes not much point beyond "it shows up"

master/buildbot/process/build.py
```diff
@@ -396,110 +404,118 @@ def startBuild(self, workerforbuilder: AbstractWorkerForBuilder):
         yield self.master.data.updates.setBuildStateString(self.buildid, 'starting')
         yield self.master.data.updates.generateNewBuildEvent(self.buildid)
 
-        try:
-            self.setupBuild()  # create .steps
-        except Exception:
-            yield self.buildPreparationFailure(Failure(), "setupBuild")
-            yield self.buildFinished(['Build.setupBuild', 'failed'], EXCEPTION)
-            return
+        if not self.skipBuild:
```
p12tic, line 407: I think it makes sense to have the following here:

```
if self.skipBuild:
   ...
   return
```

This short circuit will make the rest of the code unchanged and easier to understand. We get a little bit of duplication (10 lines?), but this code path is so different that the usual reasons to eliminate duplication don't apply anymore.

p12tic: Maybe it makes sense to do this just before `self._preparation_step` is setup, for the same reasons.

p12tic:
Looks like my previous review failed to identify a number of possible improvements. Sorry about that. I was focused into the changes to Build side, not how it's configured.

My further comments:

 - Please use `snake_case` for new code.
 
 - Steps have `doStepIf`. I think our property could be named `do_build_if` to keep consistency.
 
 - BuildFactories currently behave essentially as a list of build steps now. The configuration lives in a BuilderConfig. I think it makes sense to move `do_build_if` there as well. The the line `self.skipBuild = self.skipBuildIf(self)` would become `self.do_build = self.builder.config.do_build_if(self)` or similar.
 
- There won't be need for `setSkipBuildIf` anymore? A static configuration will be enough.

- Integration test is missing. Easiest would be to add one to `master/buildbot/test/integration/test_custom_buildstep.py`. Add `builder_kwargs` argument to `create_config_for_step` and maybe use this test as a base for your test:

```
    @defer.inlineCallbacks
    def test_step_raising_buildstepfailed_in_start(self):
        builder_id = yield self.create_config_for_step(FailingCustomStep())

        yield self.do_test_build(builder_id)
        yield self.assertBuildResults(1, results.FAILURE)
 ```

MagicRB:
I implemented your suggestions

PR #8315, opened 2025-02-04
Use `/run/current-system/sw/bin/tail` instead of `/usr/bin/tail` on N…

MagicRB:
## Contributor Checklist:


master/buildbot/scripts/logwatcher.py
```diff
@@ -110,6 +110,8 @@ def start(self):
             tailBin = "/usr/xpg4/bin/tail"
         elif platform.system().lower() == 'haiku' and os.path.exists('/bin/tail'):
             tailBin = "/bin/tail"
+        elif os.path.exists('/etc/nixos'):
+            tailBin = "/run/current-system/sw/bin/tail"
```
p12tic, line 114: I think we should also check for existence of `/run/current-system/sw/bin/tail`, like in checks above.

PR #8408, opened 2025-05-01
process: Add a mutable env attribute to Build

tdesveaux:
Allow steps to change the running env for the remaining of the Build's lifetime.

For example, a step that create a Python virtual environment in the worker tmp location can update the build's env so that subsequent steps use it without the need to configure them for this case.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -15,9 +15,11 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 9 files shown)


p12tic (approved):
Thanks a lot!

p12tic:
Rebased, hopefully tests pass this time.

PR #8409, opened 2025-05-02
worker: add `path_cls` as an alternative to `path_module`

tdesveaux:
The `path_module` usage is not playing nicely with typing currently being deployed in Buildbot.

A possibility would be to type `AbstractWorker.path_module` using a [module protocol](https://typing.python.org/en/latest/spec/protocol.html#modules-as-implementations-of-protocols), but it would need to be maintained here (although the API is stable, so wouldn't be that much work in the long term).

Another way done here is to use `pathlib.PurePath` implementations.
Usage is mostly the same as before, give or take some `str()` wrapping as `PathLike` are not directly compatible with `str` expected in functions (but requiring `PathLike` when a path is expected would improve safety further).

eg. `worker.path_module.join(a, b)` becomes `worker.path_cls(a, b)`.

There is other side benefits in that the `PurePath` classes does not implement methods that are invalid in this context.

For example, there is usages of `worker.path_module.abspath` and `worker.path_module.normpath`.
Both are unvalid usages as:
- [`os.path.normpath`](https://docs.python.org/3/library/os.path.html#os.path.normpath). [used here](https://github.com/tdesveaux/buildbot/blob/57d0b0a73852dfcddcbea42d37a5437888852b07/master/buildbot/steps/source/p4.py#L307)
    > This string manipulation may change the meaning of a path that contains symbolic links.
- [`os.path.abspath`](https://docs.python.org/3/library/os.path.html#os.path.abspath) will use `os.getcwd()` for non-absolute paths (which would use the master workdir). [used here](https://github.com/tdesveaux/buildbot/blob/57d0b0a73852dfcddcbea42d37a5437888852b07/master/buildbot/steps/source/base.py#L233-L234)

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/build.py
+++ b/master/buildbot/process/build.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 from functools import reduce
+from pathlib import PurePath
 from typing import TYPE_CHECKING
 from typing import cast
 
@@ -284,6 +285,7 @@ class Build(properties.PropertiesMixin):
     def setupWorkerForBuilder(self, workerforbuilder: AbstractWorkerForBuilder):
         assert workerforbuilder.worker is not None
         self.path_module = workerforbuilder.worker.path_module
+        self.path_cls: type[PurePath] | None = workerforbuilder.worker.path_cls
         self.workername = workerforbuilder.worker.workername
         self.worker_info = workerforbuilder.worker.info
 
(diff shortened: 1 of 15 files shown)


p12tic (approved):

p12tic:
Let's convert whole Buildbot to pathlib.

tdesveaux:
> Let's convert whole Buildbot to pathlib.

I can do another PR to deprecate usage of `path_module` if you want.

PR #8410, opened 2025-05-03
Upgrade some dependencies (mypy, ruff)

tdesveaux:
Looks like pyup-bot is no longer working for some time now. Last PR was #8237.

## Contributor Checklist:


tdesveaux:
Circle CI fails as mypy 1.15 no longer supports Python3.8.
Should be fine to upgrade it to 3.9.

Mypy will still that types are compatible with 3.8 due to `tool.mypy.python_version` in the pyproject (although, it could be bumped as well since it's been EOL for some time.

Ref. https://mypy-lang.blogspot.com/2025/02/mypy-115-released.html "Drop Support for Python 3.8"

reneleonhardt:
Wow, thank you for all the update work! 👍

p12tic (approved):

PR #8411, opened 2025-05-03
Ruff: enable typechecking

tdesveaux:
depends on #8410

Enable ruff [flake8-type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc)

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/base.py
+++ b/master/buildbot/changes/base.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
 from typing import Any
 
 from twisted.internet import defer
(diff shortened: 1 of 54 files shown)


p12tic (changes requested):
LGTM, but let's ignore TC006 rule. It introduces inconsistency itself, as some types will be quoted, others not. I don't think we have many cases where `cast` needs string type definition by itself.

tdesveaux:
> LGTM, but let's ignore TC006 rule. It introduces inconsistency itself, as some types will be quoted, others not. I don't think we have many cases where `cast` needs string type definition by itself.

There is a performance (not sure how much) benefit to enforcing string types in cast calls so that generics are not unnecessarily resolved at runtime.
Also, casts of generic types fails in older python versions, and using a string prevent the failures (I saw this case in my other PR typing `buildbot_worker`, I'll try to find the exact issue later).

Another benefit is to move some imports into "TYPE_CHECKING" block since they would not be needed at runtime.

So since it make sens to stringy some types in cast, enforcing all casts to use string types would be more consistent.

tdesveaux:
@p12tic extract enabling TC006 to another PR so discussion does not block the other rules -> #8418

p12tic (approved):

PR #8413, opened 2025-05-04
.github: add dependabot configuration

tdesveaux:
NOTE: feel free to close without any discussion as it's not a change validated by maintainers

Relates to https://github.com/buildbot/buildbot/pull/8410#issue-3037447217

dependabot is already used for JS dependencies (https://github.com/buildbot/buildbot/pulls?q=author%3Aapp%2Fdependabot+created%3A%3C%3D2025-05-04+).

Adding the following `.github/dependabot.yml` would setup checks for JS/Python/Docker/GithubActions.

I added `groups` so that major upgrade would have their own PR, but minor/patch are batched into a single PR to avoid noise.

It looks like dependabot honor `# pyup: ignore` (https://github.com/tdesveaux/buildbot/pull/20/files)

I don't think there is other pitfalls this change could have, let me know if you think of any that need investigation.

## Contributor Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,48 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "cron"
+      cronjob: "0 0 1,15 * *"  # every 1st and 15th of the month
+    groups:
+      minor-upgrades:
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "npm"
+    directories:
+      - "e2e/"
+      - "www/**"
+    schedule:
+      interval: "cron"
+      cronjob: "0 0 1,15 * *"  # every 1st and 15th of the month
+    groups:
+      minor-upgrades:
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "docker"
+    directories:
+      - "**/*"
+    schedule:
+      interval: "cron"
+      cronjob: "0 0 1,15 * *"  # every 1st and 15th of the month
+    groups:
+      minor-upgrades:
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "cron"
+      cronjob: "0 0 1,15 * *"  # every 1st and 15th of the month
+    groups:
+      minor-upgrades:
+        update-types:
+          - "minor"
+          - "patch"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,45 @@
...
+      interval: "weekly"
+    groups:
+      minor-upgrades:
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "github-actions"
+    directories:
+      - ".github/**/*"
```
reneleonhardt, line 38: / should be enough for Dependabot:
```suggestion
    directory: "/"
```

tdesveaux:
Unfortunately, it doesn't catch that it should upgrade the runs-on of CIDB jobs, which uses a deprecated ubuntu version.

https://github.com/buildbot/buildbot/actions/runs/14853441473/job/41701261486
image

reneleonhardt:
There is no `.github/dependabot.yml` in https://github.com/buildbot/buildbot/tree/master/.github
How did Dependabot create those PRs?

By the way, postgres:12 was declared EOL half a year ago:
https://endoflife.date/postgresql

ubuntu-20.04 has been deprecated a long time ago, so it was removed 2 weeks ago:
https://github.com/actions/runner-images/pull/11748

Your CI runs on windows-2019 which will de deprecated this year.

Why not update manually now to get runnable workflows again? 🤔

tdesveaux:
>  How did Dependabot create those PRs?

You can enable Dependabot in a repository settings, and it will run with a default configuration.

> Why not update manually now to get runnable workflows again? 🤔

I'd say it's outside the scope of this PR. My comment only noted that Dependabot doesn't catch this issue. (Also, weird choice on Github's part to skip the build instead of failing).

.github/dependabot.yml
```diff
@@ -0,0 +1,44 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
p12tic, line 6: better biweekly. Weekly was too frequent

p12tic (changes requested):
One small nit, but great stuff otherwise, thanks!

.github/dependabot.yml
tdesveaux: Dependabot doesn't seem to have `biweekly` ([ref](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#interval)). So used a cronjob instead.

p12tic (approved):

tdesveaux:
@p12tic dependabot should have ran, it did on my fork at least.

Here is what's configured in the settings on my fork if that helps track why (in Settings > Security > Advanced Security):
image

PR #8414, opened 2025-05-05
buildbot_worker: add type annotations

tdesveaux:
Should probably wait on #8411 for ruff checks, and #8410 since I used mypy 1.15.0 locally.

I started by ... doing it by hand, but quickly lost my mind and went looking for auto-typing tools for at least the basic stuff.

Ended up with a process of:
- [autotyping](https://pypi.org/project/autotyping/) with `--aggressive` for basic simple typing.

- [MonkeyType](https://pypi.org/project/MonkeyType)
  - `monkeytype run `which trial` -x buildbot_worker` as the runtime 'source' for types
  - `monkeytype apply {MODULE}`

- then manually post-process stuff like `InlineCallbackType`
- and finally, a loop of `ruff format && ruff check && mypy worker/buildbot_worker/`, fixing remaining errors

(NOTE: I also tried [auto-type-annotate](https://pypi.org/project/auto-type-annotate/) but gave up as it needs a basic mypy to run without errors).

As for changes, I limited myself (as much as possible) to NOT change the code.
Using `cast` or `assert` (when the code would also fails without the assert) when needed. Introducing some basic type aliases when it made sense.
Added some `# [TODO|FIXME]: ` when encountering an obvious issue.

It's not perfect at all and probably introduce typing mistakes, but my goal was to setup a base typing so that improvement can be then added.
It seems acceptable to me as annotations have no runtime impacts.

Apologize for the lack of consistency between the commits (some update only tests, some do class+tests, some do multiple classes, ...). The whole process was somewhat mentally taxing, BUT feel free to bring up anything that would improve the review experience (split in multiple PRs, re-split commits, or anything else).

## Contributor Checklist:


p12tic:
amazing work. I think it makes sense to split the PR into smaller changes which can land one by one.

p12tic:
Rebased.

worker/buildbot_worker/commands/transfer.py
```diff
@@ -162,7 +181,7 @@ def _writeBlock(self):
             if self.stderr is None:
                 self.stderr = f'Maximum filesize reached, truncating file \'{self.path}\''
                 self.rc = 1
-            data = ''
+            data = b''
```
p12tic, line 184: should live in separate commit

worker/buildbot_worker/commands/utils.py
```diff
@@ -44,14 +44,12 @@ def getCommand(name):
 
 # this just keeps pyflakes happy on non-Windows systems
 if runtime.platformType != 'win32':
-    WindowsError: type | None = RuntimeError
```
p12tic, line 47: Should live in separate commit

p12tic (approved):

tdesveaux:
PR split into multiple to ease review.
PRs can be found here: https://github.com/buildbot/buildbot/pulls?q=is%3Apr+author%3Atdesveaux+buildbot_worker%3A+add+type+annotations+part

PR #8418, opened 2025-05-10
ruff: provide context on ignore TC006 runtime-cast-value

tdesveaux:
requires #8411

https://docs.astral.sh/ruff/rules/runtime-cast-value/

extracted from #8411 to discuss whether it's beneficial or not.

See:
- https://github.com/buildbot/buildbot/pull/8411#pullrequestreview-2828142875
- https://github.com/buildbot/buildbot/pull/8411#issuecomment-2866422547

## Contributor Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,9 @@ target-version = "py38"
         "RUF012", # mutable-class-default
         "RUF017", # quadratic-list-summation (only used in one case, where didn't like the outcome: https://github.com/buildbot/buildbot/pull/8081)
 
+        # See: https://github.com/buildbot/buildbot/pull/8418#issuecomment-2869007629
+        # quoted types in cast should be considered on a case by case basis
+        # rather than enforced.
         "TC006", # runtime-cast-value
     ]
 


p12tic:
Rebased

p12tic:
In reply to https://github.com/buildbot/buildbot/pull/8411#issuecomment-2866422547:

I don't like TC006 because it's ugly. Some IDE's won't support the code completions in the types within string within cast. They are inconsistent with all the annotations which don't use strings. With regards to performance benefits - I think if we start thinking about performance in such places, then Python was wrong language for the job :-) A lot of code will be 10-100x faster if it was rewritten to Rust or C++.

So I think it's best to write the simplest and most readable code without thinking about performance too much and revisit only if certain code path is deemed a hot path.

Regarding failures on older Python versions, I agree that types in strings are appropriate for buildbot worker because it supports older Python.

tdesveaux:
I agree the performance is bit of a non argument, it seemed to be the base reason of this rule, so mentioned it.

A bigger argument IMHO is to remove the need to import some types from runtime, moving them to TYPE_CHECKING blocks.
So, since some of the casts need to be strings for this reason or because of Generics in older pythons. Forcing all casts to use string for consistency make sense.

But I understand your argument, I'll update the PR to simply comment why the rule is ignored.

p12tic:
I agree that there _are_ arguments for using string. But it's ugly. I don't know, maybe I just don't like it personally and will need time to adjust. I didn't like black formatting at the start either :-) We will see. Enabling TC006 and reformatting the codebase is something like 2 minutes of work, so we can defer it until we see what's the final impact once the codebase is completely typed.

tdesveaux:
> I agree that there _are_ arguments for using string. But it's ugly. I don't know, maybe I just don't like it personally and will need time to adjust. I didn't like black formatting at the start either :-) We will see. Enabling TC006 and reformatting the codebase is something like 2 minutes of work, so we can defer it until we see what's the final impact once the codebase is completely typed.

No worries, I think your opinion is fair, and ultimately, you are the maintainer and have to live with these changes, so it's 100% your call.

I updated the changes to only add a comment explaining why this rule is disabled. Let me know if you have feedback on it.

PR #8442, opened 2025-05-14
.github: add release workflow

### LINKED ISSUE #8412: CI: port AppVeyor/CircleCI jobs to GithubActions
Relates to https://github.com/buildbot/buildbot/pull/8410#issuecomment-2848669606

There is multiple configurations to maintain when bumping the minimum python version.
Removing these two would be helpful.

### AppVeyor

Was added to have Windows workers (https://github.com/buildbot/buildbot/pull/2019)
Also runs coverage.

Both should be trivial to replace with a GHA

### CircleCI

Was added to build the docs: (https://github.com/buildbot/buildbot/pull/2013)
Also does the release process.

Both can be done in GHA (or maybe Buildbot itself?)

If maintainers ok this, I should be able to do it.

tdesveaux:
On publish of a new Release, this will be triggered to build tarballs and pyinstaller worker for linux and windows. They will be added to the release that triggered them.

This remove appveyor and circleci jobs.

Closes #8412 

The release process might be even more automated, by triggering on a new tag and creating the release.
The pypi upload could also be trivially added here.

This remove the codecov from appveyor and circleci, but is codecov still really used?

This also add back the pyinstaller build of buildbot-worker for windows.

Here is an example run on my fork: https://github.com/tdesveaux/buildbot/releases/tag/v99.9.9
Related action: https://github.com/tdesveaux/buildbot/actions/runs/15014550509
The release was simply created by adding a title and tag, the action built and added the files.

## Contributor Checklist:


RELEASING.rst
```diff
@@ -54,16 +54,15 @@ Then run:
 
 This will create the required tags, make documentation, copy it to bbdocs repo and push everything.
 
-Step 5: Draft a new release and wait for CircleCi to create release tarballs
+Step 5: Draft a new release and wait for GithubAction to create release tarballs
```
p12tic, line 57: Github Action (add space)

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+          cache-dependency-path: |
+            requirements-pip.txt
+            requirements-ci.txt
+            worker/setup.py
+      - run: python -m pip install -r requirements-pip.txt
+      - run: python -m pip install -r requirements-ci.txt
+      - run: python -m pip install pyinstaller==6.13.0
+      - run: pyinstaller pyinstaller/buildbot-worker.spec
+      - run: mv "dist/buildbot-worker${{ matrix.ext }}" "dist/${{ env.WORKER_BIN_NAME }}"
+      - run: trial --reporter=text --rterrors buildbot.test.integration.interop
```
p12tic, line 52: Better not run tests, because once tag is created, the release _must_ happen one way or another. The person doing release should hopefully look into the results of tests before creating the tag. So the only thing this step does is introducing potential instability

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+          cache: 'pip'
+          cache-dependency-path: |
+            requirements-pip.txt
+            requirements-ci.txt
+            master/setup.py
+            worker/setup.py
+      - run: python -m pip install -r requirements-pip.txt
+      - run: python -m pip install -r requirements-ci.txt
+      # needed by maketarballs
+      - run: pipx install virtualenv
```
p12tic, line 79: Why not `python -m pip install virtualenv` like in previous steps?

p12tic (changes requested):
Thanks a lot for looking into this.

p12tic:
I think we should not remove circle CI configuration completely for now, just make it not contain any jobs. I cannot turn off CircleCI because I may need to release from older Buildbot branches.

.github/workflows/release.yml
tdesveaux: To split the 'real' dependencies from the build dependencies.
If virtualenv require a version incompatible with what Buildbot requires, it would cause issues.
Since `pipx` install in an isolated environment, it prevent this.

I forgot to do the same for pyinstaller in the other job though.

p12tic (approved):

.github/workflows/release.yml
p12tic: Yes, please do so. Thanks!

Issue #8374, opened 2025-03-08
Connection to old buildbot slave no longer works


mojca:
After upgrading the buildbot master to 4.2.1, the old build slaves (using "native" system packages on Debian 9 and Solaris 10) are no longer able to connect.

Here is part of the log from the slave:
![Image](https://github.com/user-attachments/assets/dd98f2b8-7cf1-46c7-95b0-0f8dad4c14b9)

mojca:
The code that is throwing the error seems to have been added in https://github.com/twisted/twisted/commit/2ac17b0efd15f828214bb54ac5dee8a99b4df75e.

mojca:
@pythonspeed: do you have any suggestion what to check or modify here?

Note that I'm not a developer of `buildbot`, so maybe another developer could chime in. I believe that the code is asking the client that tries to connect for `getWorkerInfo` which is present in newer version (`buildbot-worker`), but absent in older versions of buildbot clients (`buildbot-slave` which later had to be renamed to avoid unsuitable terms in opensource). With older versions of twisted this was handled properly. If `getWorkerInfo` failed, another command was called, but with the latest version of twisted the same code no longer works.

p12tic:
Ugh, I guess Buildbot will need start patching that specific part of Twisted.

p12tic:
@mojca Could you check if https://github.com/buildbot/buildbot/pull/8450 fixes your problem?

PR #8450, opened 2025-05-15
monkeypatches: Work around old workers not being able to connect

p12tic:
The cause of the error is this new code in Twisted:

https://github.com/twisted/twisted/commit/2ac17b0efd15f828214bb54ac5dee8a99b4df75e#diff-3ee4b290d7221a2e082e7c4c040b64321158a7fdb26bfdb5aefebe4c54075e78R488-R492

Old workers may not send parents attribute in their state, thus they will fail to connect. The same problem may happen on newer workers as well, but it will only be triggered when an exception is sent to the master.

Fixes: https://github.com/buildbot/buildbot/issues/8374

### DIFF
--- a/master/buildbot/monkeypatches/__init__.py
+++ b/master/buildbot/monkeypatches/__init__.py
@@ -72,9 +72,35 @@ def patch_config_for_unit_tests():
     set_is_in_unit_tests(True)
 
 
+@onlyOnce
+def patch_twisted_failure():
+    try:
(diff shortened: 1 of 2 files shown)


tdesveaux:
Setup a reproduction: https://github.com/tdesveaux/buildbot-repro/tree/issue/8374

Can confirm this works

p12tic:
Thanks for reproduction project.

PR #8459, opened 2025-05-23
typing: worker protocols manager

tdesveaux:
Another fairly large one but I think it's better to not split the fixes from the typing, which in turn mean keeping the tests typing as well.

## Contributor Checklist:

### DIFF
--- a/master/buildbot/worker/protocols/manager/msgpack.py
+++ b/master/buildbot/worker/protocols/manager/msgpack.py
@@ -415,12 +415,12 @@ class BuildbotWebSocketServerProtocol(WebSocketServerProtocol):
 class Dispatcher(BaseDispatcher):
     DUMMY_PORT = 1
 
-    def __init__(self, config_portstr: int, portstr: str) -> None:
-        super().__init__(portstr)
+    def __init__(self, config_port: str | int) -> None:
+        super().__init__(config_port=config_port)
         try:
-            port = int(config_portstr)
+            port = int(config_port)
         except ValueError as e:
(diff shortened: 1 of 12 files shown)


master/buildbot/worker/protocols/manager/msgpack.py
```diff
@@ -129,8 +127,7 @@ def initialize(self) -> InlineCallbacksType[None]:
         finally:
             eventually(dispatcher.master.initLock.release)
 
-    @defer.inlineCallbacks
-    def call_update(self, msg: dict[str, Any]) -> InlineCallbacksType[None]:
+    def call_update(self, msg: dict[str, Any]) -> Deferred[None]:
```
p12tic, line 130: This should _really_ stay inlineCallbacks just for consistency, because with Deferreds it's easy to lose thrown exceptions. Now if usage is not consistent, then it's very easy not to return a Deferred where one should have been returned. Best to just use inlineCallbacks always.

p12tic: My point was that `remote_update_msgpack` does not return Deferred _yet_, but could conceivably do so in the future like other `remote_` methods. I think it makes sense to code defensively until we migrated to async/await and got warnings whenever an awaitable was not awaited. 

I think I remember noticing the exact same problem in one of this class methods during development, this is why I'm cautious here.

p12tic: Agreed.

p12tic (approved):

PR #8463, opened 2025-05-30
process.buildrequest: deprecate Pascal Case fields

tdesveaux:
left the `.id`  property as it's clear. Maybe should remove `brid`?

## Contributor Checklist:

### DIFF
--- a/newsfragments/buildrequest-fields.removal
+++ b/newsfragments/buildrequest-fields.removal
@@ -1,0 +1,1 @@
+The ``submittedAt``, and ``waitedFor`` fields of ``process.BuildRequest`` have been deprecated in favor of their snake case equivalent (``submitted_at``, and ``waited_for``).
--- a/master/buildbot/process/botmaster.py
+++ b/master/buildbot/process/botmaster.py
@@ -151,7 +151,7 @@ class BotMaster(service.ReconfigurableServiceMixin, service.AsyncMultiService, L
                     for build in list(builder.building):
                         # if build is waited for then this is a sub-build, so
                         # no need to retry it
-                        if sum(br.waitedFor for br in build.requests):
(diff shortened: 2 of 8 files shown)


newsfragments/buildrequest-fields.removal
```diff
@@ -0,0 +1 @@
+The ``submittedAt``, and ``waitedFor`` fields of ``process.BuildRequest`` have been deprecated in favor of their snake case equivalent (``submitted_at``, and ``waited_for``).
```
p12tic, line 1: Also add this as an instruction to master/docs/manual/upgrading/5.0-upgrade.rst

p12tic:
LGTM, small nit

p12tic:
> Maybe should remove brid?

Let's leave this until we have fully type-safe codebase. There may be a lot of nonobvious uses.

tdesveaux:
> > Maybe should remove brid?
> 
> Let's leave this until we have fully type-safe codebase. There may be a lot of nonobvious uses.

Well, I added `brid` in https://github.com/buildbot/buildbot/commit/e4bdc1484b43a43aeee949799bea36fa3a44532b#diff-6f3ba8ac94dcd0b4e6155965cff3529ec11fba99ae81826d7993584fa7c55296L258 and now, I wonder if it's not better to stay on `id` here.

p12tic:
> Well, I added brid in https://github.com/buildbot/buildbot/commit/e4bdc1484b43a43aeee949799bea36fa3a44532b#diff-6f3ba8ac94dcd0b4e6155965cff3529ec11fba99ae81826d7993584fa7c55296L258 and now, I wonder if it's not better to stay on id here.

Ok, then I agree that we can remove brid.

PR #8464, opened 2025-05-30
Use twisted new(-ish) logger

tdesveaux:
The newer Twisted logger allow to specify a level as well as namespace, making identifying logs origin easier.

## Contributor Checklist:


p12tic:
> The newer Twisted logger

For the record, the new logger is more than 10 years old at this point.

PR #8466, opened 2025-05-31
Bump min py version in some places

tdesveaux:
Drop of support for Python 3.8 was missed in a few places.

Also, the mypy conf need to be split between master and worker since they target two different version.

Or could worker also require Py3.9?

I understand the need to support older distribution that can't have newer Python versions, but since older workers can connect to the master, this should be sufficient as worker code doesn't change much?
Any other reason I'm missing?

## Contributor Checklist:

### DIFF
--- a/master/buildbot/changes/bitbucket.py
+++ b/master/buildbot/changes/bitbucket.py
@@ -21,7 +21,6 @@ from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
 from typing import ClassVar
-from typing import Sequence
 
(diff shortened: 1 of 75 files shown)


p12tic:
> Also, the mypy conf need to be split between master and worker since they target two different version.

Yes.

> Or could worker also require Py3.9?

No, because we still want to be able to let people upgrade in cases such as security vulnerabilities. Most of the time nobody uses the capability, but when they need it, they really need it.

p12tic (approved):
Thanks!

Issue #8503, opened 2025-06-06
Add properties parameter to DataConnector.get method


RaymondBakerIBM:
https://docs.buildbot.net/latest/developer/data.html#buildbot.data.connector.DataConnector.get

Using the rest api we can specify the properties we want like this
api/v2/builds?property=propKey1&property=propKey2 (returns the properties that match the given keys)

but the get method linked above doesn't provide a way to do that.

PR #8504, opened 2025-06-06
Add properties to data get method

RaymondBakerIBM:
## Allow user to specify properties in the data.get method to keep this interface at parity with the rest api

associated issue https://github.com/buildbot/buildbot/issues/8503

## Contributor Checklist:

### DIFF
--- a/master/buildbot/data/connector.py
+++ b/master/buildbot/data/connector.py
@@ -127,9 +127,16 @@ class DataConnector(service.AsyncService):
     def getResourceType(self, name):
         return getattr(self.rtypes, name, None)
 
-    def get(self, path, filters=None, fields=None, order=None, limit=None, offset=None):
+    def get(
+        self, path, filters=None, fields=None, order=None, limit=None, offset=None, properties=None
+    ):
         resultSpec = resultspec.ResultSpec(
-            filters=filters, fields=fields, order=order, limit=limit, offset=offset
+            filters=filters,
+            fields=fields,
+            order=order,
+            limit=limit,
+            offset=offset,
+            properties=[resultspec.Property(b"property", "eq", properties)] if properties else None,
         )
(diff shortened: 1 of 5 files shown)


p12tic (changes requested):
Thanks a lot! Looks good from the first sight. 

Could we add a more involved test? E.g. see master/buildbot/test/unit/data/test_connector.py. Extend patchFooListPattern store things with extra properties and add a test that shows how these properties are included into result, while regular filter without properties in test_get_filters wouldn't include the properties.

RaymondBakerIBM:
Hi thanks for the quick review and tip, though I'm not sure how the properties param can be tested in test_connector.py

Since [resultSpec.apply](https://github.com/buildbot/buildbot/blob/master/master/buildbot/data/resultspec.py#L398) doesn't filter based on properties and the dummy [endpoint.get](https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/data/test_connector.py#L155) method is just returning a list of dictionaries it wouldn't be representative of the actual code. For example the [BuildsEndpoint.get](https://github.com/buildbot/buildbot/blob/master/master/buildbot/data/builds.py#L106) method which is doing a separate query if it sees properties in the resultSpec.

So without adding some logic into patchFooListPattern to popProperties and apply property filters separately. I'm not sure how I would provide coverage.

It does seem like there is coverage for properties param in these files.
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/data/test_builds.py#L100
https://github.com/buildbot/buildbot/blob/master/master/buildbot/test/unit/data/test_buildrequests.py#L102

When you have the time let me know how you'd like me to proceed thanks.

p12tic:
Thanks for reply. I think the current PR code is good enough, I just need time to figure out how best to test it.

PR #8595, opened 2025-06-16
Don't use the -v option for cp on AIX when running cpdir

RaymondBakerIBM:
Remove the -v option from cp when running the cpdir command on AIX. This should allow AIX to use buildbot.steps.source.git.Git with method copy.

I don't see tests relating to the solaris check, so I assume that a new unit test isn't required. Let me know if I'm wrong. 

## Contributor Checklist:


p12tic (approved):

p12tic:
Looks great, thanks!

Issue #8678, opened 2025-09-27
UI crash when searching in paginated log with ANSI code


cedric-roussel:
Buildbot version: 4.3.0

Reported console error:
```js
chunk-XR3JNTW2.js?v=6974948c:28374 Uncaught TypeError: Cannot read properties of undefined (reading 'firstPos')
    at addOverlayToCssClasses (LineCssClasses.ts:101:17)
    at overlaySearchResultsOnLine (LogChunkSearch.ts:351:24)
    at LogTextManager.getRenderedLineContent (LogTextManager.ts:463:39)
    at LogViewerTextLineRenderer (LogViewerTextLineRenderer.tsx:56:18)
    at children (LogViewerText.tsx:195:13)
```

A builder to reproduce, prints enough ANSI-colored lines so pagination kicks:
```python
from buildbot.plugins import util, steps

f = util.BuildFactory()
f.addStep(steps.ShellCommand(command=[
    "bash","-lc",
    r'''for i in $(seq 1 20000); do printf "\033[1;3$((i%7))mLINE_%06d\033[0m\n" "$i"; done;'''
]))

c['builders'].append(util.BuilderConfig(
    name="ansi-flood-shell",
    workernames=["YOUR-WORKER"],
    factory=f
))
```

When we open the log and start typing in search, more logs are loaded:
GET /api/v2/logs/4467095/contents?offset=528&limit=5000
The first line of log from that response, should be appended in log viewer, but it replaces the original first one.
<img width="1266" height="128" alt="Image" src="https://github.com/user-attachments/assets/a8b20558-1750-40c8-87ab-f8be4f490220" />

Because addOverlayToCssClasses  uses the positions from the original first line, it will either highlight the new line randomly, or crash depending on content length.

p12tic:
Fixed in https://github.com/buildbot/buildbot/pull/8679.

PR #8679, opened 2025-09-28
WIP: LogChunkParsing: revert to use absolute row number

cedric-roussel:
Fix for #8678 

After #8267 the cache is relative; keys should not be combined without reindexing
Invalidate the cache to force a lazy rebuild on next render

## Contributor Checklist:


cedric-roussel:
Hi, @p12tic @mokibit, buildbot is so great, thank you !
While looking at this issue I'm a bit puzzled by latest changes in [LogChunkParsing](https://github.com/buildbot/buildbot/blob/master/www/base/src/util/LogChunkParsing.tsx).

`basic` test expects global indexes
`basicForceTextNoEscapes `test expects chunk-relative indexes

I first thought that's my problem, we should pick one and stick to it.
But eventually got lost... I'm not sure I really have understood the full picture.

I fixed my runtime crashes by invalidating the CSS class map after chunk merge...
Unit tests green, and manual testing for #7852 and #8678 are both looking good to me.

Let me know what you think ?

PR #8705, opened 2025-10-20
process: allow to collapse build requests when the builder object is not defined.

clopez:
On asymmetric multi-master configurations where one master runs the Web UI and the other master works with the builders and workers, it is possible to configure the Web UI master with an empty list of builders and workers.

However, the code to collapse the build requests runs on the Web UI master and if this list is empty then it can't find the builder object to evaluate if a custom collapse function is defined for it, so it fails to collapse anything.

Instead of doing that, it should continue and use the globally defined collapse function or the default collapse function.

This patch moves the code to evaluate getCollapseRequestsFn to the class BuildRequestCollapser allowing it to work even when the builder object is not defined.

This issue was observed on the WebKit's Buildbot CI deployment and there it was work-arounded by defining also the builders/workers for the Web UI master. See https://bugs.webkit.org/show_bug.cgi?id=300920

## Contributor Checklist:

### DIFF
--- a/master/buildbot/process/builder.py
+++ b/master/buildbot/process/builder.py
@@ -28,7 +28,6 @@ from twisted.python import log
 from buildbot import interfaces
 from buildbot.data import resultspec
 from buildbot.interfaces import IRenderable
-from buildbot.process import buildrequest
 from buildbot.process import workerforbuilder
 from buildbot.process.build import Build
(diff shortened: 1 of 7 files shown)


p12tic:
Rebased, added missing typing as we now require it.

clopez:
thanks!

PR #8707, opened 2025-10-26
Fix deprecation warning from alembic >= 1.16.0

cjwatson:
Following
https://github.com/sqlalchemy/alembic/commit/9913fdb8e20262ccbc34f715b044d36b4be8b5ad, alembic produces a deprecation warning if we don't set `path_separator`. This caused test failures in Debian, as seen in
https://bugs.debian.org/1117391.

## Contributor Checklist:

* [N/A] I have updated the unit tests

### DIFF
--- a/master/buildbot/db/migrations/alembic.ini
+++ b/master/buildbot/db/migrations/alembic.ini
@@ -8,7 +8,8 @@ script_location = %(here)s/
 file_template = %%(rev)s_%%(year)d-%%(month).2d-%%(day).2d_%%(slug)s
 
 # sys.path path, will be prepended to sys.path if present.
-# defaults to the current working directory.
+# defaults to the current working directory.  for multiple paths, the path separator
+# is defined by "path_separator" below.
 prepend_sys_path = .
 
 # timezone to use when rendering the date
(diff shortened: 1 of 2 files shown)


cjwatson:
This might conceivably help with #8695, although I can't decipher the CI failures there.

sa2ajj:
i wonder what this may mean: https://github.com/buildbot/buildbot/actions/runs/18822960886/job/53701296994?pr=8707#step:4:83

cjwatson:
I'm afraid I have no useful experience of Windows.  The current version of pywin32 appears to be 311 so I guess you could try just bumping that?  I don't really want to do that myself since it's clearly unrelated to my change and I don't want to commit something I don't understand in somebody else's project.

https://buildbot.buildbot.net/#/builders/151/builds/895 is also apparently an issue, but the logs appear blank to me so I've no idea what's going on there.

sa2ajj:
@cjwatson, i did not realise :) long time no see in a different forum.

`pywin32` seems to be pinned :( i'd need to review the situation, however this PR should go in

thalinda-oss (approved):

Issue #8711, opened 2025-11-07
GerritEventLogPoller broken in 4.3.0


flichtenheld:
We use GerritEventLogPoller and I tried to upgrade from 3.11.9 to 4.3.0. But GerritEventLogPoller broke.
It just logs:

```
while executing <bound method GerritHttpEventLogPollerConnector.do_poll of <buildbot.changes.gerritchangesource.GerritHttpEventLogPollerConnector object at 0x70db1566b6a0>>
        Traceback (most recent call last):
          File "/buildbot_venv/lib/python3.9/site-packages/buildbot/util/poll.py", line 57, in _run
            yield self.fn(self.instance)
          File "/buildbot_venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 2285, in unwindGenerator
            gen = f(*args, **kwargs)
        builtins.TypeError: do_poll() takes 1 positional argument but 2 were given
```

I debugged the error. The problem is actually in how it uses util.Poller. It calls `util.poll.Poller(self._connector.do_poll, self, self.master.reactor)` and then Poller calls `self.fn(self.instance)`. But  `self._connector.do_poll` is already bound to an object, so self is given as a second argument. I think something like `util.poll.Poller(type(self._connector).do_poll, self._connector, self.master.reactor)` would work. Or change the interface of Poller to not require the instance argument. If you let me know which solution you prefer I can try to create a PR.

flichtenheld:
Fixed the potential patch description. In addition to the first argument, the second argument to `Poller()` is also wrong since it should be `self._connector` not `self`.

flichtenheld:
Updates from my tests. With this change it works somewhat. But it produces a lot of duplicated change source events. AFAICT from the source this is because in 3.11.x the de-duplication happened in GerritChangeSourceBase but in 4.3.0 that is gone. There is some de-duplication in GerritChangeSource but none in GerritEventLogPoller. But de-deduplication is important for the poll mechanism since it only updates the timestamp to the ts of the last received event. So until there is a new event it will see that event every time it polls. So if that last event is a valid change that will not be filtered by any other means it will be added again and again to the database.

virsworld:
Hey @flichtenheld, is this issue still outstanding? I'd like to contribute a fix as part of a course I'm taking. I would have the PR up by December 1st.

flichtenheld:
[...]

PR #8717, opened 2025-11-28
GerritEventLogPoller Fix (Bounded method passed -> Unbounded method passed)

virsworld:
Fixed the issue described in #8711. The solution was suggested by [flichtenheld](https://github.com/flichtenheld) and I plan on tackling the issue of duplication in the ``GerritEventLogPoller`` (mentioned in the issue) in a separate PR.

**Description of problem (essentially what is mentioned in the issue)**
1. An instance of Poller was being created in gerritchangesource.py where the argument passed for the parameter fn was a bounded method. 
2. When creating the same instance of Poller the argument for instance was one of ``GerritEventLogPoller`` object. 

The first issue eventually causes a TypeError in the _run method of Poller since the bounded method passes ``self`` implicitly causing there to be two arguments being passed to fn instead of one. The second issue is that instead of passing an instance of ``GerritEventLogPoller`` we should be passing an instance of ``GerritHttpEventLogPollerConnector`` since ``do_poll`` (the bounded method in question) calls the method ``poll`` on the instance passed to it. Turns out ``poll`` is a method of ``GerritHttpEventLogPollerConnector`` and not ``GerritEventLogPoller``.

**Solution Implemented**
As suggested in the issue by [flichtenheld](https://github.com/flichtenheld) the for 1. instead of passing a bounded method an unbounded method is passed. For 2. instead of passing an instance of ``GerritEventLogPoller`` (``self``), an instance of ``GerritHttpEventLogPollerConnector`` (``self._connector``) is passed.

**Unit Test Added**
The unit test was added to test_gerritchangesource.py and essentially tries to trigger the TypeError described in the issue. I have tested it with the version with the change made in this PR and it was indeed failing with a TypeError (and of course passing with the change).

## Contributor Checklist:

* [x] I have updated the appropriate documentation (Don't think this is needed in this case)

### DIFF
--- a/master/buildbot/changes/gerritchangesource.py
+++ b/master/buildbot/changes/gerritchangesource.py
@@ -1017,7 +1017,9 @@ class GerritEventLogPoller(GerritChangeSourceBase):
             on_lines_received_cb=self._lines_received,
         )
         yield self._connector.setup()
-        self._poller = util.poll.Poller(self._connector.do_poll, self, self.master.reactor)
+        self._poller = util.poll.Poller(
+            type(self._connector).do_poll, self._connector, self.master.reactor
+        )
 
     def getFiles(self, change: str, patchset: str) -> defer.Deferred:
         return self._connector.get_files(change, patchset)
--- a/master/buildbot/test/unit/changes/test_gerritchangesource.py
+++ b/master/buildbot/test/unit/changes/test_gerritchangesource.py
@@ -1083,6 +1083,33 @@ class TestGerritEventLogPoller(
         self.assertEqual('GerritEventLogPoller:gerrit', self.changesource.name)
 
     @defer.inlineCallbacks
+    def test_poller_uses_correct_signature(self):
+        yield self.master.db.insert_test_data([
+            fakedb.Object(
+                id=self.OBJECTID,
+                name='GerritEventLogPoller:gerrit',
+                class_name='GerritEventLogPoller',
+            )
+        ])
+        yield self.newChangeSource()
+
+        self._http.expect(
+            method='get',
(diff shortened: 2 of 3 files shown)


sa2ajj:
i wonder if we have any guidelines about the language: "added" vs "add", 'fixes" vs "fix", "reformatted" vs "reformat"

my personal view is that a *proposed* change does not do anything, but proposes a change, hence **my** expectation would be that :smile:

anyway, lgtm. i am checking the CI results...

virsworld:
> i wonder if we have any guidelines about the language: "added" vs "add", 'fixes" vs "fix", "reformatted" vs "reformat"
> 
> my personal view is that a _proposed_ change does not do anything, but proposes a change, hence **my** expectation would be that 😄
> 
> anyway, lgtm. i am checking the CI results...

Would "Fixes GerritEventLogPoller [...]" be more appropriate? :thinking:

sa2ajj:
using an infinitive is always more than welcome: your change is not in (yet), hence it describes what is going to happen when it is :smile:

sa2ajj:
(please prefer cosmetic changes -- such as handling empty lines -- as separate commits)

sa2ajj:
(to re-iterate: "Fixes" / "Re-formatted" are all like they did take place... not yet)

sa2ajj:
@virsworld did you look at the failure? something that needs to fixed otherwise or is it your change? please do not say "yes" :)

virsworld:
> @virsworld did you look at the failure? something that needs to fixed otherwise or is it your change? please do not say "yes" :)

You mean the 3 failures in the CI pipeline right? I did look at them and they were not related to my change.

PR #8740, opened 2026-01-23
Drop Python 3.9 support

tdesveaux:
It's been EOL since Nov 2025 (https://devguide.python.org/versions/). Both Debian stable and Ubuntu 22.04 (oldest LTS still on standard support) do not use it.

# Contributor Checklist:

### DIFF
--- a/requirements-cidocs.txt
+++ b/requirements-cidocs.txt
@@ -1,5 +1,6 @@
 Sphinx==7.4.7;  python_version == "3.9" # pyup: ignore
-Sphinx==8.2.3;  python_version >= "3.10"
+Sphinx==8.1.3;  python_version == "3.10" # pyup: ignore
+Sphinx==8.2.3;  python_version >= "3.11"
 sphinx-jinja==2.0.2
 sphinx-rtd-theme==3.0.2
 sphinxcontrib-applehelp==2.0.0


tdesveaux:
@aj062 https://github.com/buildbot/buildbot/pull/8778#issuecomment-3928063087

> I'd like to raise a concern about dropping Python 3.9 in this PR. While it reached upstream EOL in October 2025, I think removing it in a 4.x minor release is little early:
>
>    macOS Tahoe (26.3) still ships Python 3.9.6 by default. Many Mac-based CI environments rely on this.
    RHEL 9 / CentOS Stream 9 ship Python 3.9 as the system Python and are supported until 2032 - exactly the kind of enterprise environment where Buildbot commonly lives for lot of users.
    Amazon Linux 2023 also defaults to Python 3.9.
    Python 3.9 still accounts for a significant share of PyPI downloads.
>
> Dropping a supported Python version is effectively a breaking change to minimum system requirements. Users upgrading from 4.3.x to 4.4.x wouldn't expect their runtime to suddenly be unsupported. This seems like a major version bump (Buildbot 5.0), where users explicitly opt into breaking changes and can plan accordingly. This is especially important given that there hasn't been a buildbot release for a while since version 4.3.0 and users might want to update to 4.4.x to get all the fixes/improvements made since then in an easy manner.

I didn't see a written policy regarding Buildbot's Python version support, but from the past, removing support for a Python version that's no longer maintained is accepted (see [Buildbot 4.3.0 Release notes](https://buildbot.readthedocs.io/en/latest/relnotes/index.html#deprecations-and-removals) ).

In any case, I'll leave maintainers decide the policy and whether to drop or not Python 3.9.

aj062:
> In any case, I'll leave maintainers decide the policy and whether to drop or not Python 3.9.

Thanks for the context! Fair point that 3.8 was dropped in 4.3.0 - though Python 3.9 is a different situation in practice. Python 3.8 wasn't the default system Python on any  major platform still in support when it was dropped. Python 3.9, on the other hand, remains the default on current macOS Tahoe (26.3), RHEL 9 (supported until 2032), CentOS Stream  
  9, Amazon Linux 2023 etc. That's a significantly larger installed base that would be affected.

For precedent within Buildbot itself, dropping Python 2 - which had a similarly large installed base -  was treated as a breaking change and shipped in the 3.0 major version bump.

Just something for maintainers to weigh when deciding. Either way, appreciate your work and efforts on improving buildbot for everyone.

tdesveaux:
I made a hasty choice on dropping Python 3.9 support when the underlying goal was to fix CI jobs (failing due to py3.9 not available in Github Actions `setup-python` and some pinned dependencies version dropping support for it).

I opened #8783 to keep track of OS that still uses Python 3.9 for when the subject of dropping it is considered again in the future.

Thanks @aj062 for the feedback.

PR #8742, opened 2026-01-28
Fix ruff issues that appeared after it's update

tdesveaux:
EDIT: Since #8788 fixed ruff issues, leaving this as it implements some hash functions. Feel free to close


p12tic:
Thanks for fixing this properly

PR #8743, opened 2026-01-31
pb: use abortConnection() instead of transport buffer hack

Mic92:
I stumbled over this while working on https://github.com/twisted/twisted/pull/12568

The PB connection's loseConnection() called transport.loseConnection() and then manually cleared internal Twisted transport state (dataBuffer/offset) to avoid waiting for the write buffer to drain, since the remote end is unlikely to ACK the data and we'd wait for a TCP timeout otherwise.

abortConnection() is the supported Twisted API for immediate close (available since Twisted 11.1) and does the same thing properly: it drops buffered data, unregisters from the reactor, and schedules connectionLost via callLater — which correctly fires notifyOnDisconnect handlers. The old _closeSocket() approach that was tried before the hack broke notifyOnDisconnect, but abortConnection() doesn't have that problem.

# Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/worker/test_protocols_pb.py
+++ b/master/buildbot/test/unit/worker/test_protocols_pb.py
@@ -150,7 +150,7 @@ class TestConnection(TestReactorMixin, unittest.TestCase):
 
         self.assertEqual(conn.keepalive_timer, None)
         assert conn.mind is not None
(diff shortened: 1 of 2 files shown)


Mic92:
Depends on https://github.com/buildbot/buildbot/pull/8740 for python 3.9 removal.

Issue #8369, opened 2025-03-03
BB v4 displays blank pages if Workers.showWorkerBuilders used in ui_default_config


marrak:
In BB documentation there is example:
c['www']['ui_default_config'] = {
    'Builders.buildFetchLimit': 500,
    'Workers.showWorkerBuilders': True,
}
in https://docs.buildbot.net/latest/manual/configuration/www.html
However after I migrated our BB from 3.11.9 to 4.0 every BB site is blank.
After showWorkerBuilders was removed BB pages are displayed in v4 properly.

aj062:
I am noticing this as well. This also logs following error in the browser console: `bad setting name Workers.showWorkerBuilders: setting does not exist`

If this is not supported anymore, it would be good to officially mention that it in the documentation.

https://docs.buildbot.net/4.3.0/manual/configuration/www.html still specify `'Workers.showWorkerBuilders': True` in the example for `ui_default_config`.

cc @tardyp @p12tic

tdesveaux:
#8771 fixed the crash in UI code, and #8820 will be used to track documentation task of these settings.

PR #8771, opened 2026-02-16
www: fix GlobalSettings.getSettingItem for unknown selector

tdesveaux:
Relates to #8369

# Contributor Checklist:

> remove items that are not applicable


p12tic:
@tdesveaux If you use `fixes <issue number>` in PR description then after merging the issue would be closed automatically. Less work for us :-)

tdesveaux:
Hum, I didn't want to do it here as `Workers.showWorkerBuilders` is documented (in an example only) but no longer exists in the React frontend.
I think I wanted to leave it to later choose between removing from doc or implement it.

PR #8773, opened 2026-02-17
www: Fix log copy on Chromium browser missing line breaks

### LINKED ISSUE #8704: [Buildbot 4 Regression] Line breaks don't persist when copying from stdio viewer
Copying text from the stdio viewer on buildbot 4 runs doesn't preserve line breaks. 

This doesn't reproduce while using buildbot-www version 3.11.9, but reproduces with buildbot-www version 4.0.0 (keeping same buildbot v4.0.0 in both cases).

Steps to reproduce:

- Go to a buildbot build like: https://buildbot.buildbot.net/#/builders/118/builds/4596
- Identify any build step with logs, click on "View all x lines" , it will open entire logs
- Select multiple lines and copy them
- Paste them in a text editor (preferably one which use plain-text)

Expected behavior:
- Text should be pasted in multiple lines, just as it was displayed in buildbot logs

Actual behavior:
- Entire multi-line text is pasted in a single line (making it extremely unreadable)

Large logs are often copied and pasted in various places like bug reports, messages etc. and need to preserve line breaks for readability.

Note that I noticed that this issue doesn't reproduce while pasting that copied text in apps like Notes, Mail on macOS (they probably use rich text fields), but reproduces with various apps like vim, VS code, BBEdit etc. (probably using plain text fields).

tdesveaux:
Fixes #8704

Firefox and Chromium works differently on this.
Firefox adds the line breaks to keep consistency with how the elements are displayed.

# Contributor Checklist:

### DIFF
--- a/newsfragments/fix-chromium-log-copy.bugfix
+++ b/newsfragments/fix-chromium-log-copy.bugfix
@@ -1,0 +1,1 @@
+Fixed Log copy on Chromium browser would not have line breaks (:issue:8704)
--- a/www/base/src/components/LogViewer/LogTextManager.ts
+++ b/www/base/src/components/LogViewer/LogTextManager.ts
@@ -433,7 +433,7 @@ export class LogTextManager {
     const lineIndexInChunk = index - chunk.firstLine;
     const lineType = chunk.lineTypes[lineIndexInChunk];
     const lineStartInChunk = chunk.textLineBounds[lineIndexInChunk];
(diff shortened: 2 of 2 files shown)


tdesveaux:
@p12tic  FYI, I think this broke the log view (eg. https://buildbot.buildbot.net/#/builders/122/builds/4898/steps/8/logs/stdio `[1A[2K[43/44] [firefox] › tests/worker.spec.ts:24:7 › ...`).
Will look into it this afternoon.

tdesveaux:
Nevermind, search highlighting is offset on lines with invisible char, but it was the case before as well.

PR #8778, opened 2026-02-18
Support for Python 3.14

tdesveaux:
Declare support for Py3.14, and fix related issues

Requires https://github.com/buildbot/metabbotcfg/pull/172

# Contributor Checklist:

### DIFF
--- a/master/buildbot/test/unit/reporters/test_irc.py
+++ b/master/buildbot/test/unit/reporters/test_irc.py
@@ -267,11 +267,14 @@ class TestIrcStatusBot(unittest.TestCase):
         self.assertNotIn('c0', b.channels)
 
         if platform.python_implementation() != 'PyPy':
-            self.assertEqual(sys.getrefcount(u), 2)  # local, sys
+            py_ver_str = platform.python_version_tuple()[:2]
+            py_ver = (int(py_ver_str[0]), int(py_ver_str[1]))
+            ref_counts = (2, 3, 2) if py_ver <= (3, 13) else (1, 2, 1)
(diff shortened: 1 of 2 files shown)


aj062:
@tdesveaux  Thanks for the work on Python 3.14 support!

I'd like to raise a concern about dropping Python 3.9 in this PR. While it reached upstream EOL in October 2025, I think removing it in a 4.x minor release is little early:

  - macOS Tahoe (26.3) still ships Python 3.9.6 by default. Many Mac-based CI environments rely on this.
  - RHEL 9 / CentOS Stream 9 ship Python 3.9 as the system Python and are supported until 2032 - exactly the kind of enterprise environment where Buildbot commonly lives for lot of users.
  - Amazon Linux 2023 also defaults to Python 3.9.
  - Python 3.9 still accounts for a significant share of PyPI downloads.

Dropping a supported Python version is effectively a breaking change to minimum system requirements. Users upgrading from 4.3.x to 4.4.x wouldn't expect their runtime to suddenly be unsupported. This seems like a major version bump (Buildbot 5.0), where users explicitly opt into breaking changes and can plan accordingly. This is especially important given that there hasn't been a buildbot release for a while since version 4.3.0 and users might want to update to 4.4.x to get all the fixes/improvements made since then in an easy manner.

The Python 3.14 additions in this PR are independently valuable and could land without the 3.9 removal. Could the drop be deferred to a future 5.x release instead?

tdesveaux:
This PR is built on top of #8740, which is the one dropping Python 3.9 support (needed as it fixes CI jobs). I'll link to your comment and answer there.

PR #8782, opened 2026-02-20
www: Fix favicon missing due to double slash in icon URL path

### LINKED ISSUE #8722: [Buildbot 4 Regression] Favicon missing on most pages
The favicon path is incorrectly generated for most pages (there is an extra `/`) resulting in a missing favicon.

For example, if you go to https://buildbot.buildbot.net/#/builders/107/builds/5696, the favicon is missing. According to Chrome's developer tools, neither icon.svg nor icon.png can be loaded:

Image

Note that there are two slashes in the path, highlighted in yellow.

The path for the favicon is correct for some pages. For example, https://buildbot.buildbot.net/#/about

I suspect the problem is here:
https://github.com/buildbot/buildbot/blob/591c42114069987cc3328bd2fb96e908360b753e/www/ui/src/util/FavIcon.ts#L40-L41

If `buildbotUrl` always ends with a slash, which I think is the case, then an extra slash is appended.

t0k0shi:
buildbotUrl is guaranteed to end with '/' by the backend (copy_str_url_param_with_trailing_slash), but FavIcon.ts concatenated  '/icon.png' and '/icon.svg', producing double slashes like 'http://host//icon.png' which results in 404.                                               Remove the leading '/' from icon path literals so that the URL is correctly formed. 
 Fixes #8722                                                                                                                                  

# Contributor Checklist: 
 [x] I have updated the unit tests                                                                                                          
 [x] I have created a file in the `newsfragments` directory (and read the `README.txt` in that directory)

### DIFF
--- a/newsfragments/favicon-double-slash.bugfix
+++ b/newsfragments/favicon-double-slash.bugfix
@@ -1,0 +1,1 @@
+Fixed favicon missing on most pages due to double slash in URL path (:issue:`8722`).
--- a/www/ui/src/util/FavIcon.test.tsx
+++ b/www/ui/src/util/FavIcon.test.tsx
@@ -1,0 +1,127 @@
+/*
+  This file is part of Buildbot.  Buildbot is free software: you can
+  redistribute it and/or modify it under the terms of the GNU General Public
+  License as published by the Free Software Foundation, version 2.
+
+  This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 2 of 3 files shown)


p12tic:
@t0k0shi Prettier test fails

t0k0shi:
> @t0k0shi Prettier test fails

@p12tic  Thanks for the heads-up! I pushed a formatting fix in 5e737f5.

PR #8787, opened 2026-02-22
Upgrade Yarn usage to v4

tdesveaux:
Requires https://github.com/buildbot/metabbotcfg/pull/175

I don't think this needs a newsfragment?

nodeLinker PnP does not work first hand, will look into it more once this is merged.
Workspace setup might be useful as well.

# Contributor Checklist:

### DIFF
--- a/www/base/package.json
+++ b/www/base/package.json
@@ -83,7 +83,7 @@
     "less": "^4.3.0",
     "prettier": "^3.5.3",
     "process": "^0.11.10",
-    "rollup-plugin-visualizer": "^6.0.3",
+    "rollup-plugin-visualizer": ">=6.0.3 <=6.0.5",
     "sass": "^1.56.0",
     "vite": "~5.4.12",
     "vite-plugin-checker": "^0.9.0",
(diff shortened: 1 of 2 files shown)


p12tic:
I think we can add a newsfragment because this is relevant to web plugin developers.

p12tic:
let's see if this PR makes everything green

tdesveaux:
I don't see the `COREPACK_ENABLE_DOWNLOAD_PROMPT` envvar, so I assume it's not using the latest image?

tdesveaux:
@p12tic  still looking into this or can I take over?

p12tic:
@tdesveaux not looking anymore, but by now I think we should migrate off yarn. Current problem is that yarn creates lockfile hashes that are different between two machines. This is completely not acceptable.

It seems that they generate a zip of the package and take a hash of that, leading to various consistency problems, e.g. https://github.com/yarnpkg/berry/issues/6105.

The only reason to use yarn was that it supported local packages better `"buildbot-data-js": "link:../data-module"` in package.json. It seems like npm supports equivalent since npm 7 which is 5 years old at this point. Maybe better idea would be migrate to that.

tdesveaux:
Well that’s annoying.

I’ll look into porting to npm.

p12tic:
> I’ll look into porting to npm.

thanks a lot!

tdesveaux:
Replaced by #8791

PR #8796, opened 2026-02-25
www: Fix /config missing `user_any_access_allowed` field

tdesveaux:
This would break frontend dev server setup (`npm run start` in `www/base`) as it uses `/config` instead of the injection as in production.

# Contributor Checklist:


p12tic:
> This would break frontend dev server setup (npm run start in www/base)

Can we still have special case for dev server setup to work?

tdesveaux:
> > This would break frontend dev server setup (npm run start in www/base)
> 
> Can we still have special case for dev server setup to work?

Poor phrasing on my part. It's currently broken and this PR fixes it.

PR #9034, opened 2026-05-07
SecretInPass fix failure on warning and support multiline secrets

RaymondBakerIBM:
Allow SecretInPass to support multiline secrets, useful for ssh keys etc 
ie the `sshPrivateKey` param in the git step https://docs.buildbot.net/latest/manual/configuration/steps/source_git.html

Don't fail to retrieve secret if there is stderr in the output, since generally that is just a warning.

example stderr when pass still successfully retrieves secret:  lock files disapearing mid call (this shouldn't flunk the step as the secret is avail)
```
gpg: error opening lockfile '<Home>/.gnupg/pubring.kbx.lock': No such file or directory
gpg: lockfile disappeared
```

# Contributor Checklist:

### DIFF
--- a/master/buildbot/secrets/providers/passwordstore.py
+++ b/master/buildbot/secrets/providers/passwordstore.py
@@ -72,7 +72,7 @@ class SecretInPass(SecretProviderBase):
                 ['pass', entry],
                 env=self._env,
                 collect_stderr=True,
-                stderr_is_error=False, # usually this is true
+                stderr_is_error=False,
             )
             if stderr:
                 log.warn("Got stderr while accessing 'pass {entry}': {stderr}", entry=entry, stderr=stderr)
--- a/master/buildbot/secrets/providers/passwordstore.py
+++ b/master/buildbot/secrets/providers/passwordstore.py
@@ -27,6 +27,8 @@ from twisted.internet import defer
 from buildbot import config
 from buildbot.secrets.providers.base import SecretProviderBase
 from buildbot.util import runprocess
+from twisted.logger import Logger
+log=Logger()
 
 if TYPE_CHECKING:
     from buildbot.util.twisted import InlineCallbacksType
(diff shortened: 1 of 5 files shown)


master/buildbot/secrets/providers/passwordstore.py
```diff
@@ -65,15 +67,25 @@ def get(self, entry: str) -> InlineCallbacksType[str | None]:
...
+            if stderr:
+                log.warn("Got stderr while accessing 'pass {entry}': {stderr}", entry=entry, stderr=stderr)
             if rc != 0:
+                log.error("Got RC != 0 accessing 'pass {entry}' RC = {rc}", entry=entry, rc=rc)
                 return None
-            return output.decode("utf-8", "ignore").splitlines()[0]
-        except OSError:
+
+            secret = "\n".join(output.decode("utf-8", "ignore").strip().splitlines())
+            if not secret:
```
tdesveaux, line 84: If I understand correctly, this would return None even if a (empty) secret was retrieved?
Any reason to treat empty as None?

master/buildbot/secrets/providers/passwordstore.py
```diff
@@ -65,15 +67,25 @@ def get(self, entry: str) -> InlineCallbacksType[str | None]:
...
             )
+            if stderr:
+                log.warn("Got stderr while accessing 'pass {entry}': {stderr}", entry=entry, stderr=stderr)
             if rc != 0:
+                log.error("Got RC != 0 accessing 'pass {entry}' RC = {rc}", entry=entry, rc=rc)
                 return None
-            return output.decode("utf-8", "ignore").splitlines()[0]
-        except OSError:
+
+            secret = "\n".join(output.decode("utf-8", "ignore").strip().splitlines())
```
tdesveaux: Not sure it's our place to "sanitize" leading/trailing whiespaces and line endings here.
Is there an argument to do it?

master/buildbot/secrets/providers/passwordstore.py
RaymondBakerIBM: The original code handles errors such as OSError by returning None. The logic here is there is no secret that is an empty string so it's an error.

master/buildbot/secrets/providers/passwordstore.py
RaymondBakerIBM: You are right, I was just doing this to remove the trailing newline that pass adds to it's output. I've changed this to use rstrip('\r\n') instead

PR #9057, opened 2026-05-25
Improved reporters handling, especially in case of a lot reporters (50+)

slydiman:
The current implementation of getPreviousBuild() may cause catastrophic performance degradation in high-load systems.
Handled the case when the previous build is not completed yet (if few workers are used).
We will significantly improve performance in case of many reporters (with the mode `problem`, especially with logs) if allow the first call getDetailsForBuilds() to completely request the necessary data (and cache) just once.

- Implemented db.builds.getPrevBuild() and db.builds.getNextBuild() to get the prev/next build by a single SQL request with minimal overheads.
- Filter the prev/next build by the scheduler name to ignore manually triggered builds.
- Avoid exceptions when getPreviousBuild() returned None.
- Cache everything in getDetailsForBuilds() and do not call getDetailsForBuilds() in parallel.

### DIFF
--- a/master/buildbot/config/master.py
+++ b/master/buildbot/config/master.py
@@ -237,6 +237,7 @@ class MasterConfig(util.ComparableMixin):
             "logfileName": 'http.log',
         }
         self.services = {}
+        self.reporters_scheduler_filter = None
 
     _known_config_keys = set([
         "buildbotNetUsageData",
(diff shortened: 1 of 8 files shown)


tanwar-div:
Hi, I'm not a reviewer but I read your changes, some of them are performance enhancement but some are just increasing (or we maybe decreasing) readability, especially the build object having all the data in itself instead of requiring zip.

I just have one question, why after 2 weeks the pr is still not merged if all the tests have passed

Thanks

slydiman:
> some are just increasing (or we maybe decreasing) readability, especially the build object having all the data in itself instead of requiring zip

Please note that currently getDetailsForBuilds() always requests everything for each reporter and each generator. I changed getDetailsForBuilds() to request only the missing data, so a big zip to connect everything together is not acceptable.

slydiman:
Ping

Issue #9031, opened 2026-05-04
buildbot-worker==4.3.0 via PyPI won't work with Python 3.6


RazvanLiviuVarzaru:
**Environment**: 
- SUSE Enterprise Linux 15, Service Pack 7
- Python: 3.6
- buildbot-worker via pip

**Cause**: `Annotations` is not a `future` feature in Python 3.6: https://docs.python.org/3.6/library/__future__.html
**Suggested fix**: minimum required python version in the worker `project.toml`.
**Temporary workaround**: install an older buildbot-worker but it would be nice if we let the package manager decide which is the latest compatible.

```bash
Traceback (most recent call last):
  File "/usr/bin/buildbot-worker", line 4, in <module>
    from buildbot_worker.scripts.runner import run
  File "/usr/lib/python3.6/site-packages/buildbot_worker/scripts/runner.py", line 19
    from __future__ import annotations
    ^
SyntaxError: future feature annotations is not defined
ERROR: buildbot-worker not installed
````

hasufell:
Same on rockylinux 8. First time trying buildbot... after 3 minutes of following the main documentation. Already stuck.

> This tutorial is all about instant gratification and the five minute experience: in five minutes we want to convince you that this project works, and that you should seriously consider spending time learning the system.

I think this type of installation experience makes people lose interest pretty quickly.

PR #9066, opened 2026-06-06
worker: declare min required python version

tdesveaux:
It was the case at least since the introduction
of __future__ import annotations (ref. https://github.com/buildbot/buildbot/issues/9031).

### DIFF
--- a/worker/pyproject.toml
+++ b/worker/pyproject.toml
@@ -7,6 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "buildbot-worker"
+requires-python = ">=3.7"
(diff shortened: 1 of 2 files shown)


worker/pyproject.toml
```diff
@@ -7,6 +7,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "buildbot-worker"
+requires-python = ">=3.7"
```
sa2ajj, line 10: do i understand right that this and the referenced PR must go in together?

hasufell: What referenced PR?

PR #9072, opened 2026-06-13
SetPropertyFromCommand should preserve command output and results

sa2ajj:
## Summary
- preserve property updates from command output even when the command exits non-zero
- return `cmd.results()` so `decodeRC` mappings propagate from `SetPropertyFromCommand`
- document the updated failure and result semantics

## Testing
- `python -m twisted.trial buildbot.test.unit.steps.test_shell.SetPropertyFromCommand`


sa2ajj:
apparently i need to understand how the modern PRs are handled

PR #9074, opened 2026-06-14
git: apply partial clone filters when fetching

### LINKED ISSUE #9058: [BUG] apply partial clone filters to existing repos on fetch
When using the Git source step with filters=["blob:none"] and a persistent build directory (e.g. mode="full", method="fresh"), the filter is only effective on the initial clone. Once a repo exists in the build directory, all subsequent builds go through the fetch path, which has no awareness of the filters parameter. The repo remains a full clone forever.

This means the filters parameter only helps if the build directory is deleted and re-cloned, which essentially rarely happens. Workers that have been running for any length of time get no benefit from configuring filters on the master.

On the CPython buildbots that I maintain, that has a significant impact as each repo blob (without using filters) is around 800mb with 8 builders and 4 active branches this can rack up to a lot of wasted disk space. And while I can clean them up, I don't expect all the buildbot admins to regularly clean up the automated git fetches. 

So proposing that when filters is configured and an existing repo is reused, the Git step should configure the repo for partial clone (via git config remote.origin.promisor and remote.origin.partialclonefilter) before fetching, so that subsequent fetches respect the filter.

Some more context: https://github.com/python/buildmaster-config/pull/729

I can work on this if you'd think that'd be a useful feature.

stratakis:
Partial clone filters, while they were applied on initial cloning, it was not the case for later fetches, hence existing repos would grow overtime potentially to full clones, occupying disk space.

Existing repositories are configured as promisor partial clones before fetching and pass the configured filters to git fetch as well. This keeps the update path the same with the clone path.

Supported for git >= 2.27

Fixes #9058

# Contributor Checklist:

> remove items that are not applicable

### DIFF
--- a/master/buildbot/steps/source/git.py
+++ b/master/buildbot/steps/source/git.py
@@ -42,6 +42,7 @@ if TYPE_CHECKING:
 
 
 GIT_HASH_LENGTH = 40
+COMBINE_FILTER_RESERVED_CHARS = frozenset('~!@#$^&*()[]{}\\;",<>?\'+%')
 
 
 def isTrueOrIsExactlyZero(v: Any) -> bool:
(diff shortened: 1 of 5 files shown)


sa2ajj:
ty

Issue #7192, opened 2023-10-27
Cannot get values ​​from a long JSON string (more than 4096 bytes)


ivanov17:
Hello again. I need to parse JSON file generated on the worker to get several properties. Unfortunately, the file contains a very long JSON string. 

To do this I use the `SetPropertyFromCommand` step which performs `cat` command on the worker and then try to parse the output using a lambda function placed in the `extract_fn` parameter:

```python
steps.SetPropertyFromCommand(
    name='location-properties',
    description=['getting', 'locations'],
    command=['cat', 'file.json'],
    extract_fn=lambda rc, stdout, stderr: dict((k, v) for k, v in list(
        json.loads(stdout.strip()).items()) if k in (
            'key1', 'key2', 'key3')),
    logEnviron=False, haltOnFailure=True
)
```

When executing the step I got an exception:

```python
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/twisted/internet/defer.py", line 892, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "/usr/lib/python3.11/site-packages/twisted/internet/defer.py", line 1792, in gotResult
    _inlineCallbacks(r, gen, status, context)
  File "/usr/lib/python3.11/site-packages/twisted/internet/defer.py", line 1693, in _inlineCallbacks
    result = context.run(
  File "/usr/lib/python3.11/site-packages/twisted/python/failure.py", line 518, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
--- <exception caught here> ---
  File "/usr/lib/python3.11/site-packages/buildbot/process/buildstep.py", line 486, in startStep
    self.results = yield self.run()
  File "/usr/lib/python3.11/site-packages/twisted/internet/defer.py", line 1697, in _inlineCallbacks
    result = context.run(gen.send, result)
  File "/usr/lib/python3.11/site-packages/buildbot/steps/shell.py", line 131, in run
    new_props = self.extract_fn(cmd.rc,
  File "/data/master.cfg", line 626, in <lambda>
    json.loads(stdout.strip()).items()) if k in (
  File "/usr/lib64/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
[...]

PR #9082, opened 2026-06-19
worker: fix data loss in LineBoundaryFinder when splitting long lines

aj062:
LineBoundaryFinder splits log output longer than max_line_length into chunks before sending it to the master. The implementation had few bugs that silently dropped most of any line longer than 8192 (2 * max_line_length):

1. text[first_position : self.max_line_length - 1] in the splitting while loops used incorrect slice end. After the first iteration, subsequent slice produced an empty string. emitted.

2. The first while loop advanced first_position by max_line_length, but each chunk only consumes max_line_length - 1 source characters (the trailing '\n' is appended, not from the source). The corresponding loop in the partial-line code path already used the correct max_line_length - 1 increment, so this also caused 1 character of data loss per iteration on lines that contained an embedded newline.

3. The second while loop did not increment ret_line_count, so when it produced more than one chunk the line_times list returned to the caller had fewer entries than ret_indexes.

Each split chunk now contains max_line_length - 1 contiguous source characters plus an appended '\n', and consumers that reassemble via str.splitlines() + ''.join() recover the original line bit-for-bit (see issue #7192).

### DIFF
--- a/newsfragments/7192.bugfix
+++ b/newsfragments/7192.bugfix
@@ -1,0 +1,10 @@
+Fix data loss in worker-side log line splitting. ``LineBoundaryFinder``
+in the worker had two slicing bugs that caused most of any log line
+longer than ``2 * max_line_length`` (4096 bytes by default) to be
+silently dropped before being sent to the master, surfacing as broken
+output for build steps that capture long single-line content such as
+JSON parsed via ``SetPropertyFromCommand`` (:issue:`7192`). After this
+fix, each split chunk contains ``max_line_length - 1`` contiguous source
+characters plus an appended newline and consumers that reassemble via
+``str.splitlines()`` followed by ``''.join()`` recover the original line
+intact.
--- a/worker/buildbot_worker/test/util/test_lineboundaries.py
+++ b/worker/buildbot_worker/test/util/test_lineboundaries.py
@@ -135,5 +135,51 @@ class LBF(unittest.TestCase):
         self.assertEqual(self.lbf.append('123456789012', 4.0), None)
         self.assertEqual(self.lbf.flush(), ('5123456789012\n', [13], [3.0]))
 
+    def test_long_line_split_no_data_loss(self) -> None:
+        """A long terminated line is split into chunks that reassemble bit-for-bit.
+
+        Regression test for https://github.com/buildbot/buildbot/issues/7192:
+        prior to the fix the splitting code's slice end was an absolute index,
+        so iterations after the first emitted empty content (just '\\n') and
+        most of the line was silently dropped. With max_line_length=20 a
+        50-char line splits into chunks of 19 + 19 + 12 source chars.
+        """
+        long_line = 'x' * 50
+        out = self.lbf.append(long_line + '\n', 1.0)
+        self.assertEqual(
+            out,
+            (
+                'x' * 19 + '\n' + 'x' * 19 + '\n' + 'x' * 12 + '\n',
+                [19, 39, 52],
+                [1.0, 1.0, 1.0],
+            ),
+        )
+        assert out is not None
+        self.assertEqual(''.join(out[0].splitlines()), long_line)
+
+    def test_long_partial_line_split_no_data_loss(self) -> None:
+        """A long unterminated line is split via the partial-line path without data loss."""
+        long_partial = 'p' * 50
+        out = self.lbf.append(long_partial, 1.0)
+        self.assertEqual(
+            out,
+            ('p' * 19 + '\n' + 'p' * 19 + '\n', [19, 39], [1.0, 1.0]),
+        )
+        flushed = self.lbf.flush()
+        self.assertEqual(flushed, ('p' * 12 + '\n', [12], [1.0]))
+        assert out is not None and flushed is not None
+        self.assertEqual(''.join((out[0] + flushed[0]).splitlines()), long_partial)
+
+    def test_long_json_line_round_trip(self) -> None:
+        """Reassembling worker-split chunks via splitlines + ''.join recovers the original.
+
+        Mirrors the consumer pattern used by SetPropertyFromCommand-style
+        callers in https://github.com/buildbot/buildbot/issues/7192.
+        """
+        payload = '{"data":"' + 'A' * 5000 + '"}'
+        out = self.lbf.append(payload + '\n', 1.0)
+        assert out is not None
+        self.assertEqual(''.join(out[0].splitlines()), payload)
+
     def test_empty_flush(self) -> None:
         self.assertEqual(self.lbf.flush(), None)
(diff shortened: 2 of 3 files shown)


aj062:
FYI @tdesveaux @p12tic @tardyp 
For additional context: we hit this bug in WebKit's EWS after upgrading buildbot-worker from 2.10.5 to 4.3.0 (https://bugs.webkit.org/show_bug.cgi?id=302654). 

WebKit's layout-test results are emitted as a single JSON line that routinely exceeds 4096 characters, and after the upgrade the worker began corrupting it, which broke result parsing on a large number of builds.

With this fix, the split chunks reassemble bit-for-bit and the issue is resolved on our side. Happy to link this to https://github.com/buildbot/buildbot/issues/8996 as well if that's the right tracking issue.

tdesveaux:
I had a hunch that this would insert some line endings at the max line length boundary and this test seem to confirm it:

```patch
diff --git i/master/buildbot/test/integration/interop/test_setpropertyfromcommand.py w/master/buildbot/test/integration/interop/test_setpropertyfromcommand.py
index fb2e3a6333..6f8d606089 100644
--- i/master/buildbot/test/integration/interop/test_setpropertyfromcommand.py
+++ w/master/buildbot/test/integration/interop/test_setpropertyfromcommand.py
@@ -16,7 +16,7 @@

 from __future__ import annotations

-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, ClassVar
 from typing import Any

@@ -73,3 +73,36 @@ class SetPropertyFromCommandPB(SetPropertyFromCommand):

 class SetPropertyFromCommandMsgPack(SetPropertyFromCommand):
     proto = "msgpack"
+
+
+class SetPropertyFromLongCommand(RunMasterBase):
+    # line longer than worker's max_line_lentgth (defaults to 4096)
+    LONG_LINE: ClassVar[str] = "A" * 5000
+
+    async def setup_config(self) -> None:
+        c = {}
+        from buildbot.plugins import schedulers  # noqa: PLC0415
+        from buildbot.plugins import steps  # noqa: PLC0415
+        from buildbot.plugins import util  # noqa: PLC0415
+
+        c['schedulers'] = [schedulers.ForceScheduler(name="force", builderNames=["testy"])]
+
+        f = util.BuildFactory()
+        f.addStep(steps.SetPropertyFromCommand(property="test", command=["echo", self.LONG_LINE]))
+        c['builders'] = [util.BuilderConfig(name="testy", workernames=["local1"], factory=f)]
+
+        await self.setup_master(c)
+
+    async def test_long_line_has_no_extra_newlines(self) -> None:
+        await self.setup_config()
+        build = await self.doForceBuild(wantProperties=True)
+
+        self.assertEqual(build['properties']['test'], (self.LONG_LINE, 'SetPropertyFromCommand Step'))
+
+
+class SetPropertyFromLongCommandPB(SetPropertyFromLongCommand):
+    proto = "pb"
+
+
+class SetPropertyFromLongCommandMsgPack(SetPropertyFromLongCommand):
+    proto = "msgpack"
``` 


So unfortunately, I don't think this is a proper fix as it would corrupt any data (in a less bad way than previous truncation, but still).

aj062:
@tdesveaux Thanks for the review, appreciate that. You're right, and it's a real problem worth fixing. It just isn't a problem this PR creates, or one it tries to solve.

That synthetic newline comes from `lineboundaries.py:71` and has been there since f4c869e1 (2022). Your test fails on `master` today for the same reason, and by a wider margin. So the choice isn't "corruption vs no corruption", it's "extra newline plus severe data loss" vs "extra newline only".

**This PR fixes the data loss.** That's a separate bug in the same function, and by far the more destructive of the two: today a 100 KB log line reaches the master as 5.7 KB.

| line length | chars preserved on master | with this PR |
|---:|---:|---:|
| 4,096 | 4,095 | 4,096 |
| 5,000 | 4,999 | 5,000 |
| 8,192 | 4,095 | 8,192 |
| 9,000 | 4,903 | 9,000 |
| 20,000 | 7,711 | 20,000 |
| 100,000 | 5,791 | 100,000 |

```python
from buildbot_worker.util.lineboundaries import LineBoundaryFinder
NRE = r'(\r\n|\r(?=.)|\033\[u|\033\[[0-9]+;[0-9]+[Hf]|\033\[2J|\x08+)'

for n in (4096, 5000, 8192, 9000, 20000, 100000):
    records = LineBoundaryFinder(4096, NRE).append("A" * n + "\n", 1.0)[0].splitlines()
    print(n, len(records), sum(map(len, records)))
```

The record count is identical in every row, so this PR does not insert a single newline that `master` doesn't already insert. It only stops data loss. It also makes the workaround from #7192 work for the first time: `''.join(stdout.splitlines())` now reconstructs the line byte for byte, where today it silently returns truncated data.

The data loss is a blocker for us, so I'd like to land this fix on its own. Removing the extra newline needs a protocol change (on both the msgpack and PB payloads), so the worker can tell the master which newlines it inserted. Happy to work on that as a follow-up PR.

tdesveaux:
I think the simple solution would be to simply remove the max_line_length from worker side processing.

We might event remove the whole `LineBoundaryFinder` eventually as the master does not use it's processing AFAIK (worker side pb only send text, master side msgpack processing only forward text). But let's not do that here, I'll need to investigate further.

So removing the worker-side `max_line_length` will fix your issue properly.
Master will still truncate the lines, but that's. something that should be taken care of at another time.

Proper fix for the comment "too big lines will fill master's memory, and slow down the UI too much." should be to send the text as chunks, which can be done at another time after looking into the metadata worker's `LineBoundaryFinder` send.

PS: sorry I took this long to properly consider this.

aj062:
@tdesveaux  Thanks for looking into the details. 
As you suggested, I prototyped removing max_line_length from the worker. The deletion is clean and fixes the data loss, but it then hits Twisted's PB limit: with no splitting, a single long line becomes one update message, and banana.SIZE_LIMIT (640 KB) rejects it. Reproduced with a real 976,823-char line through BufferManager.

Not theoretical for us, e.g.: https://ews-build.webkit.org/#/builders/1/builds/154047/steps/22/logs/json is ~1 MB.

It's exposed because buffer_manager.append only splits at newline indexes (buffer_manager.py:149-194), so one long line can't be chunked. Pre-3.6.0 the worker chunked at 128 KB byte granularity specifically because of this limit. So removing max_line_length drops the splitting but keeps the line accumulation, and nothing guards the ceiling anymore.

So byte-level chunking in buffer_manager looks necessary alongside it. PB handles that fine since the master still runs its own LBF. But msgpack would need the master-side assembly that 3.6.0 removed: add_stdout_lines now goes straight to _on_whole_lines (process/log.py:231) and assumes a trailing \n, where 3.5.0 ran an LBF per stream.

Is restoring that the direction you'd want? It would break new-worker/old-master on msgpack, though it's still msgpack_experimental_v7.

PR #9103, opened 2026-07-16
fix `BuildbotService` not passing keyword-argument `name` to subclasses

terencehonles:
This fixes creating services such as `OldBuildCanceller`, which defines `checkConfig` with `name` as an argument, effectively requiring `name` to be a positional argument.

# Contributor Checklist:


terencehonles:
I updated the change to raise a deprecation warning instead of crashing services that don't have a `checkConfig` that accepts a `name`. This shouldn't be raised from buildbot itself, but I'm unsure how to migrate services such as `ReconfigurableBaseScheduler` (part of the [previous test failures](https://github.com/buildbot/buildbot/actions/runs/29490479882/job/87595119394?pr=9103)). Is it OK to just accept a kwargs list or should I only accept the `name` argument?

PR #9161, opened 2026-09-02
A shared Git object cache across builders on the same worker

### LINKED ISSUE #9059: RFE: support shared object cache across builders on the same worker
This is a step further (or an extra more aggressive feature?) compared to https://github.com/buildbot/buildbot/issues/9058

So every build directory on a buildbot worker maintains an independent full clone of the repository. For large repos, this means massive duplication as the same objects are stored once per builder. A worker with 32 active build directories for a repo like CPython (~800MB per .git/objects) uses ~25GB of disk for git alone.

See for some relevant context: https://github.com/python/buildmaster-config/pull/729 

The Git step could be able to maintain a single shared bare mirror per repository on each worker, and have all builders reference it through git's alternates mechanism. New clones would use git clone --reference to the shared mirror, and existing build directories would have their alternates configured to use it.

I pressume the proper location for the mirror would be the worker's base directory, outside any builder's directory so it's accessible to all builders on the worker.

This would reduce per-worker storage from X full copies to one shared copy plus negligible per-builder overhead.

This will create a single point of failure though, if the repo is corrupted for example, but should be recoverable with clobberOnFailure=True without manual intervention.

I assume this feature is kinda complicated, but myself maintaining 20 buildbots for CPython, disk space has always been an issue, so I think having something like that would be beneficial for projects like CPython.

I'd be willing to work on this assuming this is something that you'd consider.

stratakis:
# Contributor Checklist:

Fixes #9059

The Git step can now maintain one bare object cache per repository on each worker, shared by every builder on that worker through Git's alternates mechanism. New clones attach with `git clone --reference`, and existing build directories get a Buildbot-managed alternates entry. N builders of one large repository go from N full object stores to one. For the CPython case with its objects being around 800mb, when a worker has 8 builders across 4 branches, this feature will save 25GB of disk space as build dirs are recreated.

There are four standalone preparatory fixes first, then the feature in layers (parameter and identity, cache lifecycle, checkout integration, custom-path ownership, Windows support, integration tests, docs). Every commit passed lint, mypy and its test target in isolation, so hopefully it should easier to review.

## Design points

* The cache key is the credential-free normalized `repourl`, or an explicit path given as a string value. Equivalent spellings such as `git@host:repo` and `ssh://git@host/repo` share one cache.
* Credentials never reach the cache's on-disk configuration, and cache commands run with maintenance and auto-GC disabled, isolated hooks, and repository-scoped environment stripped.
* Failure semantics: a cache that cannot be prepared, updated or validated is kept in place, the reason is reported in the build log, and the build continues cache-free. Buildbot never deletes a cache that checkouts may borrow from.
* Windows is fully supported, including UNC cache paths and network sources.

## Scope

* Cross-fork sharing is unsupported. Fork-based flows such as GitLab merge requests get one cache per fork.
* No GC as a fetch-triggered GC could delete objects another checkout still borrows. Growth is documented but retention could be a future enhancement.
* Submodule object databases and LFS payloads are not shared.
* `clobberOnFailure` recreates checkouts, but a broken cache is preserved and disabled, never repaired or removed automatically. Retiring should be done by an admin of the machine.

## Testing

Due to me planning to deploy it on the CPython fleet which can save me many GB of disk space from my workers, I ran extenstive validation in as many OSes I could, including real master/worker connections across various Python versions. Full validation summary with the platform, Git-version and real-worker matrices: https://gist.github.com/stratakis/a29ee88951f8f23da6ec9be91b3dcb75

In short: the 678-test batch passes on Linux glibc/musl, NixOS, FreeBSD, OpenIndiana, Windows and macOS across CPython 3.9-3.14. Real Git binaries from 2.11 (where the feature correctly stays disabled) through 2.55. Around 520 real builds over PB master/worker pairs in both directions.

## Follow-ups

While working on the RFE I found several pre-existing issues (worker CLI and spawn encoding on Windows, the docs build with the pinned Sphinx 9.1.0, credential logging in GitPoller, and a few more), which I will file separately. Depending on how a deployment might go, this could be enhanced with things such as integrity-check cost controls, fewer warm-path round-trips on the cache, and a retention policy.

Full disclosure: Since I've started working on this feature I've utilized various AI models to help me out, mainly the tests and the whole matrix testing but also drafting code and uncovering edge cases etc. These included codex with gpt5.5 to the latest 5.6 models, claude with opus 4.6, 4.8 and the latest and finalized review rounds were from the Fable model.


stratakis:
Moving to draft for now as it seems I've found some small issues here and there.

PR #9166, opened 2026-09-03
Decode StringFileWriter incrementally to fix hangs on invalid/split UTF-8

### LINKED ISSUE #3982: UTF-8 decoding error with 0.9.15.post1 worker on Python 2.7
While running both worker and master with Buildbot 0.9.15.post1 and Python 2.7, I sometimes got backtraces below.

```
018-03-07T02:00:54+0100 [HangCheckProtocol,client]   using PTY: False
2018-03-07T02:07:54+0100 [-] Unhandled Error
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/base.py", line 1243, in run
      self.mainLoop()
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/base.py", line 1252, in mainLoop
      self.runUntilCurrent()
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/base.py", line 878, in runUntilCurrent
      call.func(*call.args, **call.kw)
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/task.py", line 239, in __call__
      d = defer.maybeDeferred(self.f, *self.a, **self.kw)
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 150, in maybeDeferred
      result = f(*args, **kw)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 164, in poll
      self.command.addLogfile(self.name, data)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 763, in addLogfile
      self._addToBuffers(('log', name), data)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 739, in _addToBuffers
      self._sendBuffers()
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 721, in _sendBuffers
      self._sendMessage(msg)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 667, in _sendMessage
      msg = self._collapseMsg(msg)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 653, in _collapseMsg
      m = bytes2unicode(m, self.builder.unicode_encoding)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/compat.py", line 87, in bytes2unicode
      return text_type(x, encoding, errors)
    File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
      return codecs.utf_8_decode(input, errors, True)
  exceptions.UnicodeDecodeError: 'utf8' codec can't decode bytes in position 9998-9999: unexpected end of data

2018-03-07T02:10:19+0100 [-] sending app-level keepalive
2018-03-07T02:18:28+0100 [-] command finished with signal None, exit code 0, elapsedTime: 1054.394399
2018-03-07T02:18:28+0100 [-] unexpected error in processEnded
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/process.py", line 985, in maybeCallProcessEnded
      self.reapProcess()
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/process.py", line 350, in reapProcess
      self.processEnded(status)
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/_baseprocess.py", line 52, in processEnded
      self.maybeCallProcessEnded()
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/process.py", line 987, in maybeCallProcessEnded
      _BaseProcess.maybeCallProcessEnded(self)
    File "/usr/local/lib/python2.7/dist-packages/twisted/internet/_baseprocess.py", line 64, in maybeCallProcessEnded
      proto.processEnded(Failure(reason))
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 250, in processEnded
      self.command.finished(sig, rc)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 774, in finished
      w.stop()
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 129, in stop
      self.poll()
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 164, in poll
      self.command.addLogfile(self.name, data)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 763, in addLogfile
      self._addToBuffers(('log', name), data)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 739, in _addToBuffers
      self._sendBuffers()
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 721, in _sendBuffers
      self._sendMessage(msg)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 667, in _sendMessage
      msg = self._collapseMsg(msg)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/runprocess.py", line 653, in _collapseMsg
      m = bytes2unicode(m, self.builder.unicode_encoding)
    File "/usr/local/lib/python2.7/dist-packages/buildbot_worker/compat.py", line 87, in bytes2unicode
      return text_type(x, encoding, errors)
    File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
      return codecs.utf_8_decode(input, errors, True)
  exceptions.UnicodeDecodeError: 'utf8' codec can't decode byte 0x99 in position 0: invalid start byte
```

This happens on very log-intensive jobs (verbose buildroot output). After such backtrace occurred, the job is hanging unfinished indefinitely (DockerLatentWorker).

I'm unable to reproduce such issue while running 0.9.10 worker using the same master instance (ie. 0.9.15.post1). I reported a similar issue in #3702 few month ago but it was maybe not related to initial reporter issue.

alexreinking:
## Summary

`StringFileWriter.remote_write()` decodes each transfer chunk (`blocksize=32*1024` by default) independently as UTF-8. If a multi-byte character straddles a chunk boundary — or the data isn't valid UTF-8 at all (e.g. mixed-encoding log output, per #3982's discussion) — this raises `UnicodeDecodeError` from inside a Twisted PB remote-message handler, rather than through buildbot's normal command-failure path. The step's deferred is never fired, so the build hangs forever, still holding any worker locks it had acquired.

We hit this in production: a `CTest` step that uploads ctest's XML results via `getFileContentFromWorker()` (which uses `StringFileWriter`) hung for multiple days once a multi-byte character in the XML landed exactly on a 32KB chunk boundary, starving other builds that needed the same worker lock. Root-caused with a byte-exact reproduction against a real production log.

## Fix

Decode incrementally with `codecs.getincrementaldecoder('utf-8')(errors='replace')`, as suggested by @rodrigc in the discussion on the stale-closed #4621, pointing at the same pattern already used for stdout/stderr decoding in `RunProcessPP` (`buildbot_worker/runprocess.py`). An incremental decoder buffers an incomplete trailing byte sequence across `remote_write()` calls instead of raising on it, and `errors='replace'` means it never raises on genuinely invalid input either — a silent days-long hang is worse than a stray `U+FFFD` in an uploaded log. `remote_close()` flushes the decoder with `final=True` so a truncated final sequence is resolved too.

This mirrors #4621's approach but takes the reviewer's suggested implementation strategy (`getincrementaldecoder`) instead of the ad hoc `_data_tail`-retry approach that PR struggled with.

## Test plan

- [x] Added `testMultiByteCharacterSplitAcrossChunks`, `testInvalidUtf8IsReplacedNotRaised`, and `testTruncatedFinalCharacterIsReplacedOnClose` to `test_remotetransfer.py`.
- [x] `python -m twisted.trial buildbot.test.unit.process.test_remotetransfer buildbot.test.unit.steps.test_worker buildbot.test.unit.steps.test_source_cvs` — 71 tests, all pass (covers both real consumers of `StringFileWriter`: `getFileContentFromWorker` and CVS's `_sourcedirIsUpdatable`).
- [x] `ruff check` / `ruff format --check` pass.

Fixes #3982


master/buildbot/process/remotetransfer.py
```diff
@@ -23,11 +23,11 @@
 import shutil
 import tarfile
 import tempfile
+from codecs import getincrementaldecoder
```
tdesveaux, line 26: I'd prefer the use of higher level helpers of the io package instead of codecs here.

You can see usage here for example: https://github.com/buildbot/buildbot/blob/61f2f4118098dbbd22cf9212e2b470cea7c505ba/master/buildbot/db/logs.py#L283-L289

master/buildbot/process/remotetransfer.py
```diff
@@ -190,12 +190,25 @@ class StringFileWriter(base.FileWriterImpl):
 
     def __init__(self) -> None:
         self.buffer = ""
+        # Data arrives in fixed-size chunks (see FileWriterImpl blocksize), so a multi-byte
```
tdesveaux: Comments are a bit too verbose, this hurts readability.

tdesveaux (changes requested):
Good catch and thanks for the fix.
Fix is conceptually good for me, just some minor improvements.

alexreinking:
@tdesveaux -- Thanks for the review! I tried switching to `io.TextIOWrapper`, but it breaks the incremental `café` test case I added (see an isolated example below). The issue is that `BytesIO` has no way to signal "no data yet" vs. "stream is finished". `TextIOWrapper` treats hitting the current write position as EOF and immediately flushes the incomplete trailing bytes as replacement characters, rather than holding them for the next write. 

The difference between `logs.py` and this is that it decodes one complete chunk in a single pass, whereas in `remotetransfer.py` we're receiving bytes a block at a time and it can split a perfectly valid multi-byte character across that boundary. I modeled this PR after how `RunProcessPP` works (see `buildbot_worker/runprocess.py`), since it deals with the same network issues. I'd like to keep `getincrementaldecoder` here, unless you see a way to get the right behavior out of `io` that I'm missing.

Here's the repro I tried:

```python
import io
buf = io.BytesIO()
reader = io.TextIOWrapper(buf, encoding='utf-8', errors='replace')
data = 'café'.encode()
pos = 0
result = ''
for i in range(len(data)):
    buf.write(data[i:i+1])
    buf.seek(pos)
    result += reader.read()
    pos = buf.tell()
print(repr(result))  # 'caf\ufffd\ufffd' -- should be 'café'
```

master/buildbot/process/remotetransfer.py
alexreinking: Trimmed!

master/buildbot/process/remotetransfer.py
alexreinking: That doesn't seem to have the right behavior when a valid multi-byte character gets split across packets.

tdesveaux:
Sorry, I was unclear.

I meant something like this:
```patch
diff --git i/master/buildbot/process/remotetransfer.py w/master/buildbot/process/remotetransfer.py
index 7e15040e45..5bd10d381c 100644
--- i/master/buildbot/process/remotetransfer.py
+++ w/master/buildbot/process/remotetransfer.py
@@ -189,10 +189,16 @@ class StringFileWriter(base.FileWriterImpl):
     """
 
     def __init__(self) -> None:
-        self.buffer = ""
+        self._buffer = BytesIO()
+
+    @property
+    def buffer(self) -> str:
+        return self._buffer.getvalue().decode(errors='replace')
 
     def remote_write(self, data: str | bytes) -> None:  # type: ignore[override]
-        self.buffer += bytes2unicode(data)
+        if not isinstance(data, bytes):
+            data = data.encode(errors='replace')
+        self._buffer.write(data)
 
     def remote_close(self) -> None:  # type: ignore[override]
         pass
```


Note the conversion from not bytes to bytes as I don't think this ever receive a str data (and should not if it does).
But still keeping it like this until I can check if it's possible or not.

alexreinking:
@tdesveaux — Thanks! That does seem to work. I needed to add an encoding setter to the `buffer` property, too. There were some errors that came from trying to assign a string to the buffer when I ran the test suite.

tdesveaux (approved):

PR #9170, opened 2026-09-08
Fix GitLab OAuth group pagination

### LINKED ISSUE #9169: GitLabAuth only gets the first 20 groups for user, due to pagination
I have GitLab auth set up similar to this in master.cfg, in order to only allow access to people who are members of the "examplegroup" group in GitLab.

```python
from buildbot.plugins import util

c = BuildmasterConfig = {}

c['www'] = {
    'port': 8010,
    'plugins': {},
    'auth': util.GitLabAuth(
        gitlab_auth['secret'],
        gitlab_auth['id'],
        gitlab_auth['secret'],
        autologin=True,
    ),
    'authz': util.Authz(
        allowRules=[
            util.AnyEndpointMatcher(role='examplegroup'),   # the group in gitlab we want to grant access to
            util.AnyEndpointMatcher(role='xxend-of-listxx', defaultDeny=True),  # deny access to all others
        ],
        roleMatchers=[
            util.RolesFromGroups(),
        ],
    ),
}
```

However, in [getUserInfoFromOAuthClient](https://github.com/buildbot/buildbot/blob/61f2f4118098dbbd22cf9212e2b470cea7c505ba/master/buildbot/www/oauth2.py#L428), GitLabAuth just makes one call to `/groups` with no arguments. But the [groups API at GitLab is paginated](https://docs.gitlab.com/api/groups/#list-all-groups), and only returns 20 groups by default. This means that if a given user is a member of more than 20 groups, `examplegroup` might not be returned, leading to the user not being a allowed access despite being a member of this group. To fix this, GitLabAuth needs to fetch the full group list.

KundanDhage:
## Summary

Fixes #9169.

`GitLabAuth` only fetched the first page of the GitLab `/groups` API, which defaults to 20 groups. As a result, users belonging to more than 20 groups could have their required group missing from the response and be incorrectly denied access by Buildbot's role-based authorization.

This change updates `GitLabAuth` to:

* Request up to 100 groups per page.
* Follow GitLab's `X-Next-Page` response header.
* Fetch all pages until there are no more groups.
* Combine the groups from all pages before performing role matching.

## Testing

* Python compilation passed for the changed files.
* Added/updated tests to verify that multiple GitLab group pages are fetched correctly.


master/buildbot/www/oauth2.py
```diff
@@ -427,7 +437,13 @@ def __init__(self, instanceUri: str, clientId: str, clientSecret: str, **kwargs:
 
     def getUserInfoFromOAuthClient(self, c: requests.Session) -> dict[str, Any]:
         user = self.get(c, "/user")
-        groups = self.get(c, "/groups")
+        groups = []
+        page = 1
+        while page:
+            page_groups, headers = self.getWithHeaders(c, f"/groups?per_page=100&page={page}")
```
fm9390: Would it make sense to follow the Link header’s rel="next" URL here? The [GitLab pagination docs](https://docs.gitlab.com/api/rest/#pagination-link-header) recommend following these links instead of constructing page URLs.

KundanDhage: Thanks for the suggestion. I updated the pagination logic to follow GitLab’s `Link` header with `rel="next"` instead of constructing page URLs from `X-Next-Page`. The linked absolute URL is requested directly without prepending the resource endpoint, and the test now covers a realistic `Link` header containing multiple relations.

master/buildbot/www/oauth2.py
```diff
@@ -159,14 +159,24 @@ def createSessionFromToken(self, token: dict[str, Any]) -> requests.Session:
...
         if ret.status_code >= 400:
             msg = f'OAuth2 session: error accessing resource {path}: {ret.status_code}'
             extra_info = ret.headers.get('www-authenticate', None)
             if extra_info:
                 msg += f' www-authenticate: {extra_info}'
             raise Error(503, msg.encode('utf-8'))
-        return ret.json()
+        return ret
+
+    def getWithHeaders(
```
tdesveaux: I'd say there is not much use to this function.
Your change of adding `getResponse` is sufficient and should be ok to expect caller to manipulate the Reponse object as they see fit.

master/buildbot/www/oauth2.py
```diff
@@ -159,14 +159,24 @@ def createSessionFromToken(self, token: dict[str, Any]) -> requests.Session:
         return s
 
     def get(self, session: requests.Session, path: str) -> Any:
+        ret = self.getResponse(session, path)
+        return ret.json()
+
+    def getResponse(self, session: requests.Session, path: str) -> requests.Response:
```
tdesveaux: Please use snake_case for new APIs, we're trying to transition to it (however slowly...).

master/buildbot/www/oauth2.py
```diff
@@ -427,7 +445,13 @@ def __init__(self, instanceUri: str, clientId: str, clientSecret: str, **kwargs:
 
     def getUserInfoFromOAuthClient(self, c: requests.Session) -> dict[str, Any]:
         user = self.get(c, "/user")
-        groups = self.get(c, "/groups")
+        groups = []
+        url = f"{self.resourceEndpoint}/groups?per_page=100"
```
tdesveaux: I'm not familiar with Gitlab, so raising this as a question.

From the docs: https://docs.gitlab.com/api/groups/#list-all-groups

Maybe `all_available=false` should be set to avoid gitlab admin having access to all groups? 
`active` and `archived` might be worth considering as well.

master/buildbot/www/oauth2.py
```diff
@@ -427,7 +445,13 @@ def __init__(self, instanceUri: str, clientId: str, clientSecret: str, **kwargs:
...
+        while url:
+            page_groups, headers = self.getWithHeadersFromUrl(c, url)
+            groups.extend(page_groups)
+            next_url = self._parse_next_link(headers)
+            url = next_url or ""
```
tdesveaux: small nit, I'd prefer a clear `if next_url is None: break` here.

tdesveaux (commented):
Looks good overall, some small things.

Thanks for this!

master/buildbot/www/oauth2.py
KundanDhage: Agreed : removed getWithHeaders/getWithHeadersFromUrl entirely. getUserInfoFromOAuthClient now calls get_response directly and reads .json() and .headers off the response itself, so callers manipulate the Response object as you suggested.

master/buildbot/www/oauth2.py
KundanDhage: Done , renamed to get_response to match the project's snake_case convention for new APIs.

master/buildbot/www/oauth2.py
KundanDhage: Good catch. Added all_available=false explicitly to the request URL. Without it, GitLab defaults all_available to true for administrators, which would return every accessible group on the instance rather than just the groups the user is actually a member of that could cause incorrect role matches for admin accounts. Left active/archived out for now since those weren't required, but happy to add them if we want archived-group memberships excluded from role matching.

master/buildbot/www/oauth2.py
KundanDhage: Updated , replaced url = next_url or "" with an explicit if next_url is None: break, so the loop termination doesn't rely on "" being falsy.

master/buildbot/www/oauth2.py
```diff
@@ -436,6 +448,14 @@ def getUserInfoFromOAuthClient(self, c: requests.Session) -> dict[str, Any]:
             "groups": [g["path"] for g in groups],
         }
 
+    @staticmethod
+    def _parse_next_link(headers: requests.structures.CaseInsensitiveDict[str]) -> str | None:
```
fm9390: Optional simplification: could we use response.links.get("next", {}).get("url") instead of a custom parser? requests.Response already provides parsed links.

fm9390 (commented):
I reviewed the pagination changes and ran buildbot.test.unit.www.test_oauth locally: 22 tests passed, 4 E2E tests skipped. Following the Link header addresses my original suggestion. I left one optional suggestion to simplify the parsing.

master/buildbot/www/oauth2.py
KundanDhage: Thanks for the suggestion. I replaced the custom `Link` header parser with `response.links.get("next", {}).get("url")`, using Requests’ built-in Link-header parsing. The test was updated to cover the parsed `next` link as well.

fm9390: Thanks for simplifying this and updating the test as well!

tdesveaux (approved):
LGTM, just need CI passing to merge

KundanDhage:
> LGTM, just need CI passing to merge

@tdesveaux Fixed the Ruff formatting issue in `test_oauth.py` and pushed the correction in commit `ae4b3ee2a`. CI should pass now.

tdesveaux (approved):

